PR #72, opened 2015-03-20
Fix comparison of `sfdisk` output.

crawford:
```
$ sudo sfdisk -q -c /dev/sdb 1
sfdisk: --id is deprecated in favour of --part-type
 7
```

Both the leading and trailing whitespace needs to be trimmed from
" 7\n" in order for the comparison to succeed.

### DIFF
--- a/waagent
+++ b/waagent
@@ -2145,7 +2145,7 @@ def RunGetOutput(cmd,chk_err=True):
     """
     LogIfVerbose(cmd)
     try:                                     
-        output=subprocess.check_output(cmd,stderr=subprocess.STDOUT,shell=True)
+        output=subprocess.check_output(cmd,shell=True)
     except subprocess.CalledProcessError,e :
         if chk_err :
             Error('CalledProcessError.  Error Code is ' + str(e.returncode)  )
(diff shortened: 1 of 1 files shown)


crawford:
Actually, this isn't sufficient. Both stdout and stderr are being captured.

crawford:
Updated. This now discards stderr. I have no idea what side effects this will have, but it matches the documentation now.

yuezh:
@crawford I understand the first commit. But why the second is needed?

yuezh:
@crawford I'm a little bit concerned about the second patch. Would you please keep it for CoreOS first? A better solution would be replace "id" check with "part-type". I would do it in next version.

yuezh:
Hi, Alex

I've create a commit which changes the log level to warning.
https://github.com/Azure/WALinuxAgent/commit/0bf5284cd73a07f2bd2006c20189d63c995d6229

In next version, we will fix this issue.

Thanks
Yue

crawford:
@yuezh yeah, I was concerned about it as well. I wasn't sure if all of your supported platforms had "--id" or "--part-type", so I didn't change the flags.

You're solution looks good.

PR #73, opened 2015-03-24
Allow provisioning with existing system user

tmm1:
I used `azure vm create --no-ssh-password --user admin` with a VM image that already had an admin user created on it with the uid 501. The VM failed to come up fully, failing on the following provisioning error:

```
2015/03/16 22:26:47 ERROR:CreateAccount: admin is a system user. Will not set password.
2015/03/16 22:26:47 ERROR:Provisioning image FAILED Failed to set password for system user: admin (0x06).
```

Since I am not requesting a password to be set, it does not make sense for provisioning to fail here. I have updated the code in this PR to avoid the failure for this case.

### DIFF
--- a/waagent
+++ b/waagent
@@ -1838,7 +1838,7 @@ class FreeBSDDistro(AbstractDistro):
             pass
         if uidmin == None:
             uidmin = 100
-        if userentry != None and userentry[2] < uidmin:
+        if userentry != None and userentry[2] < uidmin and password != None:
             Error("CreateAccount: " + user + " is a system user. Will not set password.")
             return "Failed to set password for system user: " + user + " (0x06)."
         if userentry == None:
@@ -2241,7 +2241,7 @@ def CreateAccount(user, password, expiration, thumbprint):
         pass
     if uidmin == None:
         uidmin = 100
-    if userentry != None and userentry[2] < uidmin:
+    if userentry != None and userentry[2] < uidmin and password != None:
         Error("CreateAccount: " + user + " is a system user. Will not set password.")
         return "Failed to set password for system user: " + user + " (0x06)."
     if userentry == None:
--- a/waagent
+++ b/waagent
@@ -1095,7 +1095,7 @@ class CoreOSDistro(AbstractDistro):
             pass
         if uidmin == None:
             uidmin = 100
-        if userentry != None and userentry[2] < uidmin and userentry[2] != self.CORE_UID:
+        if userentry != None and userentry[2] < uidmin and userentry[2] != self.CORE_UID and password != None:
             Error("CreateAccount: " + user + " is a system user. Will not set password.")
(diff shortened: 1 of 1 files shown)


yuezh:
The reason you encounter this error is that there is already a system account 'admin' in this image. Do you really want to overwrite all login info for that account? 
The error message may not be very clear. The really information it tried to deliver is "Failed to overwrite system account: admin"

tmm1:
Yes I would like to re-use the existing account and only add the new ssh key to it. This is how it works on Amazon ec2 for example.

tmm1:
The existing code will not try to overwrite the account.

Since `userentry != None`, it will skip creation here: https://github.com/tmm1/WALinuxAgent/blob/system-user-no-password/waagent#L1109.

Since `password == None`, it will only add a sudo entry here: https://github.com/tmm1/WALinuxAgent/blob/system-user-no-password/waagent#L1114

Then it will make sure `/home/admin/.ssh` is created (already exists, so no error), and set authorized_keys file: https://github.com/tmm1/WALinuxAgent/blob/system-user-no-password/waagent#L1132

So if you provision using `--no-ssh-password` with existing account, it will work as expected.

In fact, this is already the behavior when user's id is matching `CORE_UID` (500). But for my admin user, the uid is 501 so I need this PR.

imikushin:
+1
Also makes sense for [RancherOS](https://github.com/rancherio/os) provisioning.

tmm1:
I've confirmed this is working as expected for us.

```
2015/04/15 00:11:49 Provisioning image using OVF settings in the DVD.
2015/04/15 00:11:49 Disabled SSH password-based authentication methods.
2015/04/15 00:11:49 CreateAccount: admin already exists. Will update password.
2015/04/15 00:11:49 Created user account: admin
2015/04/15 00:11:49 Deploy public key:C71FA1E985FC11670FC72BBA2EBDB2B9E0A52BFE
2015/04/15 00:11:49 Ovf XML process finished
```

ilmerovingio:
+1

yuezh:
As so many people want to by-pass this check. We will consider to add an option to disable it in next release. if you want to disable this for your distro, you could derive a distro class and override this method just like CoreOS did.

hglkrijger:
@yuezh, did we ever add this as a feature?
@tmm1, sorry for the delay on this PR, it seems reasonable to me.
@ahmetalpbalkan - any objections?

ahmetalpbalkan:
if I recall correctly, I think this is already implemented. The agent should be updating existing user's password. This behavior was missing(/changed) in waagent 2.1 (I remember discovering that) and it was added back.

@hglkrijger forwarding you a thread about this now

hglkrijger:
It sounds like we do the right thing in the 2.1 version of the agent. @tmm1, please take a look and see if those changes work for you.

PR #74, opened 2015-03-27
fix "compile" errors

imikushin:
fix "compile" errors

- HttpDelete is never called: probably that is why its signature is messed;
- WALAEvent.Save must be throwing exception from exception throwing code :D

### DIFF
--- a/waagent
+++ b/waagent
@@ -2686,8 +2686,8 @@ class Util(object):
     def HttpPut(self, url, data, headers=None, maxRetry=3):
         return self.HttpRequest("PUT", url, data, headers, maxRetry)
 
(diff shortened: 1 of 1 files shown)


waagent
```diff
@@ -2686,8 +2686,8 @@ class Util(object):
     def HttpPut(self, url, data, headers=None, maxRetry=3):
         return self.HttpRequest("PUT", url, data, headers, maxRetry)
 
-    def HttpDelete(url, data, headers=None, maxRetry=3):
```
imikushin: `data` is not used in `DELETE` requests

yuezh:
Good catch. Thx

PR #75, opened 2015-03-31
Docker container

imikushin:
Dockerize WALinuxAgent adding a couple config options on the way.

### DIFF
--- a/waagent
+++ b/waagent
@@ -5876,7 +5876,7 @@ def main():
     if len(sys.argv) == 1:
         sys.exit(Usage())
 
-    LoggerInit('/var/log/waagent.log','/dev/console')
+    LoggerInit('/dev/stdout',None)
     global LinuxDistro
     LinuxDistro=DistInfo()[0]
     
--- a/waagent
+++ b/waagent
@@ -5880,7 +5880,7 @@ def main():
     if len(sys.argv) == 1:
         sys.exit(Usage())
 
-    LoggerInit('/dev/stdout',None)
+    LoggerInit('/var/log/waagent.log','/dev/console')
     global LinuxDistro
     LinuxDistro=DistInfo()[0]
     
@@ -5922,8 +5922,8 @@ def main():
     if logfile is not None:
         myLogger.file_path = logfile
     logconsole = Config.get("Logs.Console")
-    if logconsole is not None and logconsole.lower().startswith("y"):
-        myLogger.con_path = "/dev/console"
+    if logconsole is not None and logconsole.lower().startswith("n"):
+        myLogger.con_path = None
     verbose = Config.get("Logs.Verbose")
     if verbose != None and verbose.lower().startswith("y"):
         myLogger.verbose=True
--- a/waagent
+++ b/waagent
@@ -2962,12 +2962,13 @@ class ConfigurationProvider(object):
     """
     Parse amd store key:values in waagent.conf
     """
-    def __init__(self):
+    def __init__(self, walaConfigFile):
         self.values = dict()
         if 'MyDistro' not in globals():
             global MyDistro
             MyDistro = GetMyDistro()
-        walaConfigFile = MyDistro.getConfigurationPath()
+        if walaConfigFile is None:
+            walaConfigFile = MyDistro.getConfigurationPath()
         if os.path.isfile(walaConfigFile) == False:
             raise Exception("Missing configuration in {0}".format(walaConfigFile))
         try:
@@ -5890,6 +5891,7 @@ def main():
     if MyDistro == None :
         sys.exit(1)
     args = []
+    conf_file = None
     global force
     force = False
     for a in sys.argv[1:]:
@@ -5902,6 +5904,8 @@ def main():
             myLogger.verbose = True
         elif re.match("^([-/]*)force", a):
             force = True
+        elif re.match("^(?:[-/]*)conf=.+", a):
+            conf_file = re.match("^(?:[-/]*)conf=(.+)", a).groups()[0]
         elif re.match("^([-/]*)(setup|install)", a):
             sys.exit(MyDistro.Install())
         elif re.match("^([-/]*)(uninstall)", a):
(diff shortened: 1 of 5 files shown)


imikushin:
Is anyone interested in this?

ibuildthecloud:
FYI, these are changes we needed to get RancherOS running on Azure.  I would say this PR is still a work in progress and we'll keep updating it as we continue to get RancherOS running properly.

yuezh:
Hi, @imikushin. @ibuildthecloud mentioned you guys are still working on this PR. Would you please ping me when its done.

imikushin:
@yuezh, hi! it's pretty much done, actually. Works fine on RancherOS. It's also pretty generic, so it can be used to provision any Linux with Docker on Azure.

yuezh:
Ok. I will run some test and merge it.

crawford:
@ibuildthecloud what are the command line options with which you are invoking this Docker container?

imikushin:
@crawford This container essentially runs waagent in the context of the host OS. So it needs access to host devices, network and processes, as well as all the necessary files in `/etc` and `/var`. I don't know your host environment, but it is generic and can run on any Linux if it can access the necessary resources. You'll probably end up with something like this:
```sh
docker run -d --name=waagent \
  --privileged \
  --restart=always \
  --net=host \
  --ipc=host \
  --pid=host \
  -v=/home:/home \
  -v=/dev:/host/dev \
  -v=/etc/passwd:/etc/passwd \
  -v=/etc/shadow:/etc/shadow \
  -v=/etc/group:/etc/group \
  -v=/etc/sudoers:/etc/sudoers \
  -v=/lib/modules:/lib/modules \
  -v=/lib/firmware:/lib/firmware \
  -v=/var/run:/var/run \
  -v=/var/log:/var/log \
  -v=/var/lib/waagent:/var/lib/waagent \
  waagent \
  -daemon -verbose
```

In RancherOS we're using compose syntax to run system services (well, and user services, too) and `waagent` service looks like this (https://github.com/imikushin/os-azure/blob/master/azure.yml):
```yml
    waagent:
      privileged: true
      restart: always
      labels:
      - io.rancher.os.scope=system
      volumes_from:
      - all-volumes
      volumes:
      - /var/lib/waagent:/var/lib/waagent
      links:
      - console
      net: host
      ipc: host
      pid: host
      image: waagent
      command: -daemon -verbose
```

crawford:
@imikushin Thank you. That's roughly in line with what I expected.

waagent
```diff
@@ -5876,7 +5880,7 @@ def main():
     if len(sys.argv) == 1:
         sys.exit(Usage())
 
-    LoggerInit('/var/log/waagent.log','/dev/console')
+    LoggerInit('/dev/stdout',None)
```
yuezh: Hi, @imikushin. What's the purpose of the change of this line? This will break our test.

imikushin: The purpose was to make less assumptions about the environment (like availability of `/var/log/waagent.log` or `/dev/console`). These settings are later read from the config and set using `myLogger` var. Default behaviour doesn't change, because these settings are the same in the config file (`Logs.File=/var/log/waagent.log` and `Logs.Console=y`). 

Background: Docker container main process usually doesn't log to a file or serial console. Instead, they just log to stdout and docker decides what to do with it (e.g. log to syslog or you can read these logs with `docker logs` command).

If it does break tests, I want to know which ones, so I can fix.

yuezh: I see. It make sense to disable it in docker container. However, would you please change the default behavior to 'y'. Otherwise, when existing user upgrade agent to this new version, they would lost log.

waagent
```diff
@@ -5909,8 +5916,14 @@ def main():
         else:
             args.append(a)
     global Config
-    Config = ConfigurationProvider()
+    Config = ConfigurationProvider(conf_file)
     
+    logfile = Config.get("Logs.File")
+    if logfile is not None:
```
yuezh: if log file is None, set it to /var/log/waagnt

waagent
```diff
@@ -5909,8 +5916,14 @@ def main():
...
+        myLogger.file_path = logfile
+    logconsole = Config.get("Logs.Console")
+    if logconsole is not None and logconsole.lower().startswith("y"):
```
yuezh: If Logs.Console is not set, default value is 'y'

waagent
yuezh: Hi, @imikushin. Do you think it's OK if I change the default behavior?

waagent
imikushin: They wouldn't, because: 
- the log is still being written to stdout
- the hardcoded value `/var/log/waagent.log` was just migrated to waagent.conf which is a part of this software
- the change is documented in the README

waagent
imikushin: The setting remains the same, only now the value isn't hardcoded, but explicitly provided by the config. 

I'm setting the default to stdout, because otherwise it's really hard to find where the log gets written without digging in this source file, which is, frankly, quite hard to do for a normal human being.

waagent
imikushin: Again, it is set by the config. And if the user tries to run waagent from a command line, the serial console is unnecessary if the log is already written to stdout.

yuezh: Hi, @imikushin I was talking about upgrade scenario. When existing user upgrade to new version, I can't overwrite existing config. That's the reason I need to keep the console log as default setting.

imikushin: That's exactly what you're doing: the installation procedure replaces the old config with the new one, moving the existing config to `/etc/waagent.conf.old`
From README:
```
-install: Manual installation of the agent
     * Checks the system for required dependencies
     * Creates the SysV init script (/etc/init.d/waagent), the logrotate
       configuration file (/etc/logrotate.d/waagent) and configures the image
       to run the init script on boot
     * Writes sample configuration file to /etc/waagent.conf
     * Any existing configuration file is moved to /etc/waagent.conf.old
     * Detects kernel version and applies the VNUMA workaround if necessary
     * Moves udev rules that may interfere with networking
       (/lib/udev/rules.d/75-persistent-net-generator.rules,
        /etc/udev/rules.d/70-persistent-net.rules) to /var/lib/waagent/
```

yuezh: That step is for manual installation. In factor, most users will use rpm or dpkg to install the agent. For example, on SuSE, we can't overwrite the config file.
All I'm asking is to keep the change only for RancherOS and docker based distribution.

imikushin:
@yuezh I've made the changes you requested.

PR #79, opened 2015-04-30
Fixed ReplaceStringInFile clears file when pattern to replace is not found

jorrit:
When `ReplaceStringInFile` is invoked (for instance when using `-serialconsole`) and the string or pattern to be replaced is not found, the entire file is cleared.

This PR fixes that and adds two unit tests.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -14,6 +14,8 @@
 #
 
 import unittest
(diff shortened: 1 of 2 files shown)


jorrit:
Is there anything I can do to make this PR acceptable?

PR #83, opened 2015-05-15
Add an option to decode CustomData from Base64

AbelHu:


jasonzio:
Same change needs to be applied at line 1146 in waagent (to handle CoreOS), no?

AbelHu:
CoreOS always decode CustomData. If CoreOS want to use the configure to enable it, they should set the default value to y and then remove the method in line 1146.

Sent by my Windows Phone
发件人: Jason Zions<mailto:notifications@github.com>
发送时间: ‎2015/‎5/‎16 2:24
收件人: Azure/WALinuxAgent<mailto:WALinuxAgent@noreply.github.com>
抄送: Abel Hu<mailto:Chou.Hu@microsoft.com>
主题: Re: [WALinuxAgent] Add an option to decode CustomData from Base64 (#83)


Same change needs to be applied at line 1146 in waagent (to handle CoreOS), no?

PR #86, opened 2015-06-02
Fix for stuck "Waiting for network." on 32-bit Linux

peteroptional:
Please accept this pull request to add support for 32-bit Linux platforms to the waaagent script. The only change is the size of the structure used to read the available interfaces depends on executing platform.

Thanks!


yuezh:
Thanks for sending the PR. I will run some test and merge it.

PR #96, opened 2015-07-27
Adding FAQ file

SuperScottz:


yuezh:
Would be better to put it in to WIKI page

PR #98, opened 2015-07-31
Add feature to disable password reset and user creation

evan2645:
Introduce config variable `Agent.DisableUserManipulation`, which neuters
`CreateAccount`, and registers a relevant error when invoked. This option is
useful for those who wish to revoke VM user management rights from the
Azure Portal. It should be noted that `CreateAccount` is used during
provisioning, so only set this to `y` once the VM has been created and
fully provisioned.

This functionality is disabled by default for backwards compatibility.

### DIFF
--- a/README
+++ b/README
@@ -161,6 +161,7 @@ LBProbeResponder=y
 Logs.Verbose=n
 OS.RootDeviceScsiTimeout=300
 OS.OpensslPath=None
+Agent.DisableUserManipulation=n
 
 The various configuration options are described in detail below. Configuration
 options are of three types : Boolean, String or Integer. The Boolean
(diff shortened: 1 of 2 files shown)


theckman:
:+1:

yuezh:
Hi, Evan. Did you test the E2E scenario on Azure portal?

evan2645:
@yuezh not sure what that is. Can you provide a link and/or docs so that I can further investigate? Google doesn't find much

yuezh:
Hi, Even

If user creatation is disabled, how should user create a user account on that VM?

And the same comment with another PR. Would you please send the PR agains 2.1 branch? We will soon switch to the new branch.

Thanks

PR #99, opened 2015-08-03
setup.py should print an error before exiting

evan2645:
Looks like this was missed


yuezh:
Hi, Evan.

The maste branch is a very old version. Please reopen a PR to the branch 2.1.

Thanks
Yue

PR #100, opened 2015-08-06
fix error when active resource disk in BSD

ryanmiao:
The error report is:
  File "/tmp/bsd_activate_resource_disk.py", line 12, in <module>
    Config=waagent.ConfigurationProvider()
TypeError: __init__() takes exactly 2 arguments (1 given)

ConfigurationProvider is changed to accept two arguments from
commit 88e8d5f47, but it doesn't update ConfigurationProvider
in bsd_activate_resource_disk.py.

Signed-off-by: Jincheng Miao <jincheng.miao@gmail.com>


yuezh:
Good catch :)

PR #101, opened 2015-08-07
Add feature to disable password reset and user creation 21

evan2645:
New PR against 2.1 branch to replace https://github.com/Azure/WALinuxAgent/pull/98

This introduces a knob named `Agent.DisableUserManipulation`. It disables the agent's ability to:
* Create new users
* Modify existing users

The intent is to prevent any possibility that someone who has gained unauthorized access to the Azure control panel could, as a result, gain unauthorized access to the VMs running there. With that in mind, destructive actions are fine, but actions which could allow someone to access the VM are not OK.

I realize that these functions are critical for provisioning, and have documented that fact. The default is `n`, so functionality will remain the same unless the knob is turned. Since the default is `n`, provisioning images will work, and the user has the option to disable this functionality after the VM is initially provisioned.

I am unsure as to the state of this project. It looks like the 'new' agent is not ready yet, and that the 'old' monolithic agent is still present and in use. I have patched the monolithic agent, and have also taken a stab at applying a similar patch to the new hotness. If you can provide some guidance here, I will update the docs to reflect the current state.

Given the nature and intention of the change, it is possible that this knob encompasses more than just user creation/modification in the future... I am open to changing the name to `Agent.Harden` or something similar.

### DIFF
--- a/README
+++ b/README
@@ -359,6 +359,13 @@ HttpProxy.Host=None
 HttpProxy.Port=None
 Type: String Default: None
 
+Agent.DisableUserManipulation:
+Type: Boolean Default: n
+
+If set, `waagent` will not create or update local users. This means you will also
+lose the ability to reset user passwords. If you wish to set this to `y`, be aware
+that it will likely affect provisioning.
+
 If set, agent will use proxy server to access internet
 
 APPENDIX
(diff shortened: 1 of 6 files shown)


hglkrijger:
@evan2645, sorry for the delay on this PR. If you would still like to submit, please target the 2.1 branch and resubmit. We are about to close the 2.0 branch and do some cleanup here.

PR #103, opened 2015-08-18
Snappy support

lizzha:


azurelinuxagent/distro/ubuntu/loader.py
```diff
@@ -22,11 +22,25 @@
...
     from  azurelinuxagent.distro.ubuntu.osutil import Ubuntu1204OSUtil, \
                                                       UbuntuOSUtil, \
-                                                      Ubuntu14xOSUtil
+                                                      Ubuntu14xOSUtil, \
+                                                      UbuntuSnappyOSUtil
+    import os
+    import azurelinuxagent.utils.fileutil as fileutil
+
+    # TODO: https://bugs.launchpad.net/snappy/+bug/1481086
+    def is_snappy():
```
yuezh: Please check azurelinuxagent.metadata. You could use DISTRO_FULL_NAME to detect snappy.

azurelinuxagent/metadata.py
```diff
@@ -70,24 +71,12 @@ def get_distro():
 Add this walk arround for detecting Snappy Ubuntu Core temporarily, until ubuntu 
 fixed this bug: https://bugs.launchpad.net/snappy/+bug/1481086
 """
-def which(program):
```
lizzha: there's a problem of this method.
On a normal Ubuntu 15.04 VM, user could install package snappy-tools to build snap package. If snappy-tools is installed, which("snappy") will return true, and it will consider the VM as a snappy VM

yuezh: Make sense.

PR #114, opened 2015-10-19
- Handle ValueError

rjschwei:
+ If/when the distribution name is empty the __import__ function throws
    a ValueError because an attemot is made to import a module represented
    by the empty string. This generates a traceback insteda of falling back
    to teh default loader.
- Do not manage the exception object caught
  + The exception object is not used for any purpose in the exception handler
    thus we do not need to assign it and manage it


yuezh:
Good catch.

PR #121, opened 2015-11-02
RHUI code(only for redhat)

andyzhangx:
The rhui code would only run in Redhat distro. It would do the following steps:
1.       Get region info from /var/lib/waagent/ExtensionsConfig*
2.       Execute a script in /tmp/ dir to decide whether yum install RHUI package

### DIFF
--- a/waagent
+++ b/waagent
@@ -2693,6 +2693,53 @@ def DeviceForIdePort(n):
 class HttpResourceGoneError(Exception):
     pass
 
+def DoInstallRHUIRPM():
+    """
+    Install RHUI RPM according to VM region
+    """
+    rhuiRPMinstalled = os.path.exists(LibDir + "/rhuirpminstalled")
+    if rhuiRPMinstalled:
+        return
+    else:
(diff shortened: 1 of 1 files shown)


waagent
```diff
@@ -5720,6 +5767,10 @@ class Agent(Util):
             if goalState.ExtensionsConfig != None :
                 goalState.ExtensionsConfig.ReportHandlerStatus()
             
+            global LinuxDistro
+            if LinuxDistro == "redhat":
+                DoInstallRHUIRPM()
```
yuezh: Putting this line here means, it will be executed every 25 seconds.

andyzhangx: [andy] Yes, but actully it will ony run once, see function DoInstallRHUIRPM(): 
 + rhuiRPMinstalled = os.path.exists(LibDir + "/rhuirpminstalled")

2701  + if rhuiRPMinstalled:

2702  + return

2703  + else:

2704  + SetFileContents(LibDir + "/rhuirpminstalled", "")

PR #128, opened 2015-11-23
Use fallocate instead of dd for swapfile creation

### LINKED ISSUE #127: Creating a swapfile using dd is extremely slow, please use fallocate on Linux
@butangmucat told me today that his Azure instance takes 10 mins to generate a swapfile on each boot. Most sane Linux admins, however, people prefer to use [fallocate(1)](http://linux.die.net/man/1/fallocate) to do the work of creating a non-sparse file, especially on IO-speed-challanged conditions like Azure. This also reduces disk wear (trivial to say, isn't it?).

The biggest difference of using `dd` and `fallocate` is that the former creates a file with some given contents, while the latter only allocates some space for a file. Since the space is allocated (therefore the file is not sparse), it's still effective for swaps.

https://github.com/Azure/WALinuxAgent/blob/2.0/waagent#L545 exhibits this behavior:
```Python
swap = Config.get("ResourceDisk.EnableSwap")
if swap == None or swap.lower().startswith("n"):
    return
sizeKB = int(Config.get("ResourceDisk.SwapSizeMB")) * 1024
if os.path.isfile(mountpoint + "/swapfile") and os.path.getsize(mountpoint + "/swapfile") != (sizeKB * 1024):
    os.remove(mountpoint + "/swapfile")
if not os.path.isfile(mountpoint + "/swapfile"):
    Run("dd if=/dev/zero of=" + mountpoint + "/swapfile bs=1024 count=" + str(sizeKB))
    Run("mkswap " + mountpoint + "/swapfile")
if not Run("swapon " + mountpoint + "/swapfile"):
    Log("Enabled " + str(sizeKB) + " KB of swap at " + mountpoint + "/swapfile")
else:
    Error("ActivateResourceDisk: Failed to activate swap at " + mountpoint + "/swapfile")
```

See also https://stackoverflow.com/questions/257844/quickly-create-a-large-file-on-a-linux-system.

Resolves #127.

Arthur2e5:
This saves time and disk life. Fix #127.

You may want to cherry-pick this commit to all other branches.

`fallocate` comes from the `util-linux` (or `util-linux-ng`) package, and is widely adopted across distros. It appears that you have already listed `util-linux` as a dependency of waagent in your control and spec files.

### DIFF
--- a/waagent
+++ b/waagent
@@ -2579,14 +2579,15 @@ class FreeBSDDistro(AbstractDistro):
         if os.path.isfile(filename):
             os.remove(filename)
         
-        fn_sh = ShellEscape(filename)
+        fn_sh = ShellQuote(filename)
         
         if nbytes < max_blocksize:
             exitcode = Run(dd_cmd.format(bs=bytes, fn=fn_sh, n=1))
         else:
-            exitcode = Run(dd_cmd.format(bs=max_blocksize, fn=filename, n=bytes/max_blocksize))
+            exitcode = Run(dd_cmd.format(bs=max_blocksize, fn=fn_sh, n=bytes/max_blocksize))
             exitcode <<= 8 # lshift so we have space for another exitcode
-            exitcode += Run(dd.cmd.format(bs=nbytes%max_blocksize, fn=fn_sh, n=1))
+            if nbytes%max_blocksize:
+                exitcode += Run(dd_cmd.format(bs=nbytes%max_blocksize, fn=fn_sh, n=1))
         return exitcode
 
     def GetHome(self):
@@ -2729,13 +2730,16 @@ def GetLineStartingWith(prefix, filepath):
             return line
     return None
 
-def ShellEscape(wordList):
+def ShellQuote(wordList):
     """
-    Escape a list or tuple of strings for Unix Shell as words, using the
+    Quote a list or tuple of strings for Unix Shell as words, using the
     byte-literal single quote.
     
     The resulting string is safe for use with ``shell=True`` in ``subprocess``,
-    and in ``os.system``. ``assert shlex.split(ShellEscape(wordList)) == wordList``.
+    and in ``os.system``. ``assert shlex.split(ShellQuote(wordList)) == wordList``.
+    
+    See POSIX.1:2013 Vol 3, Chap 2, Sec 2.2.2:
+    http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_02_02
     """
     if not isinstance(wordList, (tuple, list)):
         wordList = (wordList,)
--- a/waagent
+++ b/waagent
@@ -586,11 +586,11 @@ class AbstractDistro(object):
         
         See also:
           - os.posix_fallocate(filename,0,bytes) # py 3.3+
-          - Run('mkfile {nbytes} {fn}'.format(fn=filename, nbytes=nbytes)) # even slower than dd, xfs_mkfile on Linux
+          - Run('mkfile {nbytes} {fn}'.format(fn=filename, nbytes=nbytes)) # BSD, xfs_mkfile (slow) on Linux
         """
         if os.path.isfile(filename)
             os.remove(filename)
-        return Run(ShellEscape(['fallocate', '-l', str(bytes), filename]))
+        return Run(['fallocate', '-l', str(bytes), filename])
     
     def GetHome(self):
         return GetHome()
@@ -2735,7 +2735,7 @@ def ShellEscape(wordList):
     byte-literal single quote.
     
     The resulting string is safe for use with ``shell=True`` in ``subprocess``,
-    and ``assert shlex.split(ShellEscape(wordList)) == wordList``.
+    and in ``os.system``. ``assert shlex.split(ShellEscape(wordList)) == wordList``.
     """
     if not isinstance(wordList, (tuple, list)):
         wordList = (wordList,)
@@ -2755,12 +2755,18 @@ def RunGetOutput(cmd, chk_err=True, log_cmd=True):
     """
     Wrapper for subprocess.check_output.
     Execute 'cmd'.  Returns return code and STDOUT, trapping expected exceptions.
-    Reports exceptions to Error if chk_err parameter is True
+    Reports exceptions to Error if chk_err parameter is True.
+    
+    When 'cmd' is a list or a tuple instead of a string, the process ``cmd[0]`` with
+    args ``cmd[1:]`` is started. This gives a safer and more determistic behavior
+    by avoiding all kinds of expansions. Otherwise, the shell will handle the
+    splitting, as well as processing of syntax elemenets like 'if', '&&' and globs.
     """
     if log_cmd:
-        LogIfVerbose(cmd)
+        LogIfVerbose(str(cmd))
     try:
-        output=subprocess.check_output(cmd,stderr=subprocess.STDOUT,shell=True)
+        output=subprocess.check_output(cmd,stderr=subprocess.STDOUT,
+                                       shell=(not isinstance(cmd,(tuple, list))))
     except subprocess.CalledProcessError,e :
         if chk_err and log_cmd:
             Error('CalledProcessError.  Error Code is ' + str(e.returncode)  )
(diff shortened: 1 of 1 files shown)


Arthur2e5:
Grr, so I need to keep a dd for BSD. Implementing that 4M blocksize one in hope of better performance…

waagent
```diff
@@ -550,7 +550,7 @@ class AbstractDistro(object):
         if os.path.isfile(mountpoint + "/swapfile") and os.path.getsize(mountpoint + "/swapfile") != (sizeKB * 1024):
             os.remove(mountpoint + "/swapfile")
         if not os.path.isfile(mountpoint + "/swapfile"):
-            Run("dd if=/dev/zero of=" + mountpoint + "/swapfile bs=1024 count=" + str(sizeKB))
+            self.fallocate(mountpoint + "/swapfile", sizeKB * 1024)
```
Arthur2e5: Now I feel this one without any method things is better:
```Python
# fn = mountpoint + "/swapfile"
# We can also try using some fail-and-fallback methods here, like catching the 127 (command not found) return from fallocate and…
if sys.version_info >= (3,3):
  os.posix_fallocate(fn, 0, sizeKB * 1024) 
else if IsLinux():
  Run('fallocate -l ' + str(bytes) + ' ' + filename)
else:
  Run("dd if=/dev/zero of=" + fn + " bs=" + str(16384 * 1024) + " count=" + str(sizeKB / 16384))
  if sizeKB % 16384:
    Run("dd if=/dev/zero of=" + fn + " bs=" + str(sizeKB % 16384 * 1024) + " count=1")
```
Well, at least we don't need as much checks as for a real method~~, and all kinds of mess should be as short as possible so people are not that likely to see them~~.

waagent
```diff
@@ -2087,6 +2098,24 @@ class FreeBSDDistro(AbstractDistro):
...
+        """
+        blocksize = 16 * 1024 * 1024
+        if type(bytes) is str:
+           bytes = int(bytes)
+
+        # We want to make the return value look like real.
+        exitcode = Run("dd if=/dev/zero of=" + filename + " bs=" + str(blocksize) + " count=" + str(bytes/blocksize))
+        if bytes%blocksize:
+            # Finish up using a single block.
+            exitcode = Run("dd if=/dev/zero of=" + filename + " bs=" + str(bytes%blocksize) + " count=1")
```
jasonzio: Unless you use "append" this will simply overwrite the first 16MB of the file with more zeroes and not achieve your goal. It would be better to actually pre-compute the correct number of blocks (rounding up if necessary) and just call dd once.

Arthur2e5: Yes.. and the only clean `oflag` stuff seems like a GNUism, defeating the (potential but hard-to-find) point of providing FreeBSD compatibility. I don't really like the rounding up idea, since it messes up with the original design of providing kb-level options.

Besides `oflag`, another candidate in dd,`seek`, looks tricky too unless we fall back to the original `bs=1024` (or the dd default and common 'logical sector' size `bs=512`) design for filling the last blocks. Doing some `>>` operations in the shell command can have our appending job done, but we are not actually using `dd`'s precise blockfile processing abilities on our output then.

And still that question -- should I really write a method like this? To align with the `dd` behavior I added some redundant checks (at least in our original context) for file existence.

(And a correction: 'simply overwrite the first 16MB [or less] of the file with more zeroes, [truncates the file to that size,] and not achieve your goal).

Arthur2e5: Anyway, let me round it to 512-byte blocks first.

Arthur2e5: Ah never mind, I've decided to do it the shell way (x

waagent
```diff
@@ -570,6 +570,17 @@ class AbstractDistro(object):
...
 
+    def fallocate(self,filename,bytes):
+        """
+        Create a non-sparse file of that size. Overwrite existing file.
+        
+        Python 3.3+: os.posix_fallocate(filename,0,bytes)
+        I am too lazy to write sys.version_info >= (3,3) here and for BSD. No, no more wrapping methods.
+        """
+        if os.path.isfile(filename)
+            os.remove(filename)
```
jasonzio: If you do this, you're not *overwriting* the existing file; you're deleting it and recreating it. Those are two distinct things. (It matters if some process has an open file handle; overwrite leaves the handle pointing to a file that is still linked, but remove/create leaves that old handle pointing to a file that will be deleted as soon as the handle is closed.)

Arthur2e5: Ah…… yes you are right. I should change the comments.

Arthur2e5:
Completely forgot this. Will update, uh, in a week.

Arthur2e5:
Ugh… almost forgot this PR again. Just here again to write some TODO…

For BSD I can try `mkfile`. Its Linux equivalent is `xfs_mkfile`.

https://stackoverflow.com/a/257849/3770260

Arthur2e5:
Updated, and all rebased.

Not using mkfile since it looks terribly slow.

Arthur2e5:
@jasonzio Please review.

waagent
```diff
@@ -2551,6 +2564,31 @@ class FreeBSDDistro(AbstractDistro):
...
+        
+        fn_sh = ShellEscape(filename)
+        
+        if nbytes < max_blocksize:
+            exitcode = Run(dd_cmd.format(bs=bytes, fn=fn_sh, n=1))
+        else:
+            exitcode = Run(dd_cmd.format(bs=max_blocksize, fn=filename, n=bytes/max_blocksize))
+            exitcode <<= 8 # lshift so we have space for another exitcode
+            exitcode += Run(dd.cmd.format(bs=nbytes%max_blocksize, fn=fn_sh, n=1))
```
jasonzio: First: typo. Should be Run(dd_cmd.format(....)) ; you have a dot between "dd" and "cmd". How did this work when you tested it?

Second: You only need to do this second running of the dd command if nbytes%max_blocksize is non-zero.

waagent
```diff
@@ -2691,6 +2729,19 @@ def GetLineStartingWith(prefix, filepath):
...
+    Escape a list or tuple of strings for Unix Shell as words, using the
+    byte-literal single quote.
+    
+    The resulting string is safe for use with ``shell=True`` in ``subprocess``,
+    and in ``os.system``. ``assert shlex.split(ShellEscape(wordList)) == wordList``.
+    """
+    if not isinstance(wordList, (tuple, list)):
+        wordList = (wordList,)
+
+    return " ".join("'{0}'".format(s.replace("'","'\\''") for s in wordList))
```
jasonzio: Sorry, this just isn't sufficient. Suppose one of the elements of wordlist was `foo\'bar`. This substitution turns that into `foo\\'bar` which will be interpreted very differently (and in a breaking way) by the shell.

waagent
```diff
@@ -2551,6 +2564,31 @@ class FreeBSDDistro(AbstractDistro):
...
+        max_blocksize = 64 * 1024 ** 2
+        dd_cmd = "dd if=/dev/zero bs={bs} count={n} conv=notrunc of={fn}"
+        
+        if not type(nbytes) is int:
+           nbytes = int(nbytes)
+        
+        if os.path.isfile(filename):
+            os.remove(filename)
+        
+        fn_sh = ShellEscape(filename)
```
jasonzio: You don't need the power of the shell, so don't set yourself up for that heartache. Build this as an array of strings. Since you need to do it multiple times, build a function that takes blocksize, count, and filename as parameters and returns the array of strings to pass to Run().

waagent
```diff
@@ -2551,6 +2564,31 @@ class FreeBSDDistro(AbstractDistro):
...
         return retcode,out
 
+    def mkfile(self,filename,nbytes):
+        """
+        Create a non-sparse file of that size. Replace existing file. BSD-compatible dd version.
+        
+        For sizes < 64M, a single-pass dd is used. Otherwise use two-pass dd.
+        """
+        max_blocksize = 64 * 1024 ** 2
+        dd_cmd = "dd if=/dev/zero bs={bs} count={n} conv=notrunc of={fn}"
```
jasonzio: This should have the smell of "sql injection attack" when it comes to a root-privileged program running a child process.

waagent
Arthur2e5: No. `foo\'bar` will be escaped to `'foo\'\''bar'` in this scheme.

waagent
Arthur2e5: The only unsafe part should be `fn`. A non-int `bs` will be killed by the `int`.

Arthur2e5: Oops, used `filename` on line 2587 by mistake.

waagent
Arthur2e5: Looks like `ShellEscape` is a terrible name. How about `ShellQuote`?

hglkrijger:
@jasonzio - any further comments on this one?

@Arthur2e5 - we'll be doing some branch cleanup soon, and will need to port this over to 2.1

waagent
```diff
@@ -2551,6 +2564,32 @@ class FreeBSDDistro(AbstractDistro):
...
+        
+        if not type(nbytes) is int:
+           nbytes = int(nbytes)
+        
+        if os.path.isfile(filename):
+            os.remove(filename)
+        
+        fn_sh = ShellQuote(filename)
```
Arthur2e5: And this quoting does solve the security issue.

Arthur2e5:
@hglkrijger Rebasing should be very trivial.

jasonzio:
LGTM (once rebased)

Arthur2e5:
Finally got some time to really look up the 2.1 source. It looks like there is no BSD-specific distro support now, so it's kind of up to you to decide if `mkfile` should go inside `fileutil` (adds an annoying `shellutil` dep) or `osutil`. Instead of making it distro-dependent, it might also make some sense to let the function do some fallback on its own, e.g.:

```Python


def mkfile(self,filename,nbytes):
    """
    Create a non-sparse file of that size. Deletes and replaces existing file.
    
    To allow efficient execution, fallocate will be tried first. This includes
    ``os.posix_fallocate`` on Python 3.3+ (unix) and the ``fallocate`` command
    in the popular ``util-linux{,-ng}`` package.
    
    A dd fallback will be tried too. When size < 64M, perform single-pass dd.
    Otherwise do two-pass dd.
    """
    dd_maxbs = 64 * 1024 ** 2
    dd_cmd = "dd if=/dev/zero bs={0} count={1} conv=notrunc of={2}"

    if not isinstance(nbytes, int):
        nbytes = int(nbytes)
    
    if nbytes < 0:
        raise ValueError(nbytes)

    if os.path.isfile(filename)
        os.remove(filename)

    # os.posix_fallocate
    if sys.version_info >= (3,3):
        # Probable errors:
        #  - OSError: Seen on Cygwin, libc notimpl?
        #  - AttributeError: What if os.posix_fallocate is gone?
        with open(filename, 'w') as f:
            try:
                os.posix_fallocate(f.fileno(), 0, nbytes)
                return 0
            except:
                # Not confident with this thing, just keep trying...
                pass
    
    # fallocate command
    fn_sh = shellutil.quote((filename,))
    ret = shellutil.run(u"fallocate -l{0} -- {1}".format(bytes, filename))
    if ret != 127: # 127 = command not found
        return ret
    
    # dd fallback
    blocks = int(nbytes/dd_maxbs)
    if blocks > 0:
        ret = shellutil.run(dd_cmd.format(dd_maxbs, fn_sh, blocks)) << 8
    
    remains = int(nbytes%dd_maxbs)
    if remains > 0:
        ret += shellutil.run(dd_cmd.format(remains, fn_sh, 1)) << 8

    return ret
```

hglkrijger:
osutil and resourceDisk both seem reasonable locations to me. I personally like the idea of having the base version try do this kind of fallback behavior. Is the double call to os.posix_fallocate intentional?

Arthur2e5:
No, it's a copy-paste-from-notepad accident.

PR #130, opened 2015-11-27
remove the print e in LogToCon

andyliuliming:
remove the print e, when logging to the c
onsole, because when IOError occurs, the print would also rais a IOError


thomas1206:
LGTM

PR #136, opened 2015-12-11
Fix config for updateMode.

karataliu:
1. Should check 'updateMode' under 'handlerManifest' node.
2. The default value for 'updateMode' looks like to be True in v2.0, defaulting to false might cause beaking changes for extensions.


karataliu:
Close this since it was fixed inhttps://github.com/Azure/WALinuxAgent/commit/5bfc685856180c6c9650fbccff299ef6bc289932

PR #176, opened 2016-04-13
[2.0] Fix prefix & tgtDir, closes #173

### LINKED ISSUE #173: [2.0] Weird prefix behaviour in setup.py
[The line which is in question](https://github.com/Azure/WALinuxAgent/blob/WALinuxAgent-2.0.18/setup.py#L81). It sets the prefix to `/` when the prefix path has a slash at the end. If it does not have a slash at the end, it adds one.

Shouldn't it be:

```
if not prefix:
    prefix = '/'
elif prefix[-1] != '/':
    prefix += '/'
```
? Same is for `tgtDir`.

z3ntu:
See #173 for more info.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -75,14 +75,17 @@ class InstallData(install):
         init = self.init_system
         prefix = self.prefix
         tgtDir = self.root
-        if prefix and prefix[-1] != '/':
-            prefix += '/'
-        else:
+
+        if not prefix:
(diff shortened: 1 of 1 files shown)


hglkrijger:
seems cleaner to me. @ahmetalpbalkan - any objections?

ahmetalpbalkan:
no objections here.

PR #182, opened 2016-05-03
Initial commit to allow for persistent ifnames

szarkos:
- Allow persistent ifnames via udev if OS.PersistentNetNames is unset or
'n'


szarkos:
Hold this commit for signoff

hglkrijger:
@szarkos, should we port this to the 2.1 branch when it's moved to master?

szarkos:
This definitely needs to be ported to 2.1.

PR #183, opened 2016-05-03
Fixes a race condition with kernel partition read and resource disk mount

sriramsa:
Problem: 
On CentOS 7.2 waagent ends up formatting the resource disk on every reboot. 

Running the parted command seems to trigger an async partition re-read in the kernel that doesn't seem to have completed when we issue a mount. If the re-read hasn't completed, the mount of the resource partition fails and waagent ends up formatting the drive and mounting again. 

You can simulate by creating a Azure CentOS 7.2 image and rebooting. Or on CentOS 7.2 try 
```
$ parted /dev/sdb print; blkid 
$ blkid 
```

Solution:
Is to do trigger a partition re-read sequentially via 'sfdisk -R <partition>' and then attempt mount again. This works and have tested on CentOS 7.2.

### DIFF
--- a/waagent
+++ b/waagent
@@ -602,13 +602,24 @@ class AbstractDistro(object):
                 if existingFS == "7" and fs != "ntfs":
                     Run("sfdisk -c " + device + " 1 83")
                     Run("mkfs." + fs + " " + partition)
+
+            Log("Mounting Resource Disk (" + partition + ").")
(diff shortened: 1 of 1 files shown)


hglkrijger:
lgtm, @ahmetalpbalkan - any objections?

ahmetalpbalkan:
lgtm

Issue #189, opened 2016-05-23
waagent should create proper version for LinuxDiagnostic


LiweiPeng:
In a VM I deployed on 5/23/2016 Ubuntu 14.04 (5/17/2016 release), with waagent 2.0.16 and LinuxDiagnostics (LAD) 2.3.9, I found misleading version information in the LAD extension.log file. see details below.

What's in the log: the version is "1.0" in HandlerEnvironment.json.
What's expected from a user point of view: as a user of LAD, when looking at extension.log, I am expecting to see something like "Microsoft.OSTCExtensions.LinuxDiagnostic-2.3" or "Microsoft.OSTCExtensions.LinuxDiagnostic-2.3.9"  because that's the version I am using.

**Log file:**
2016/05/23 17:31:28 [Microsoft.OSTCExtensions.LinuxDiagnostic-1.0]sequence number is 0
2016/05/23 17:31:28 [Microsoft.OSTCExtensions.LinuxDiagnostic-1.0]setting file path is/var/lib/waagent/Microsoft.OSTCExtensions.LinuxDiagnos
tic-2.3.9/config/0.settings2

**Related files:**
	HandlerEnvironment.json
[{  "name": "Microsoft.OSTCExtensions.LinuxDiagnostic", "seqNo": "0", "version": 1.0,  …

	HandlerManifest.json
[{
  "name": "LinuxAzureDiagnostic",
  "version": 1.0,

**> azure vm extension list -n LinuxDiagnostic**
info:    Executing command vm extension list
+ Getting extensions
data:    Publisher                       : Microsoft.OSTCExtensions
data:    Name                            : LinuxDiagnostic
data:    Version                         : 2.3

LiweiPeng:
This issue seems from the LinuxDiagnostics code side. So close this one from waagent.

hosungsmsft:
Reopening this, as it's confirmed that this is a waagent issue (as described in https://github.com/Azure/azure-linux-extensions/issues/140 ). Will send a PR for a fix soon.

ahmetalpbalkan:
@LiweiPeng the "version" key reported in HandlerEnvironment.json is not the version of the extension handler but the schema of the HandlerEnvironment.json file itself.

ahmetalpbalkan:
that said I'm not sure if the extensions model states that the agent is responsible for giving a way to find out the extension handler version. ideally an extension binary should know its own version.

PR #191, opened 2016-05-25
Write correct extension version in HandlerEnvironment.json

hosungsmsft:
Addressing issue #189.

### DIFF
--- a/waagent
+++ b/waagent
@@ -4572,7 +4572,7 @@ class ExtensionsConfig(object):
 
                 SetFileContents(root +"/config/" + seqNo +".settings",  config )
                 #create HandlerEnvironment.json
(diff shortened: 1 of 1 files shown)


ahmetalpbalkan:
@hosungsmsft that 1.0 is indicating the v1.0 of the handlerenvironment schema; not the extension handler version.

image

hosungsmsft:
Thanks @ahmetalpbalkan . Then I guess we'll have to fix the following line in the extension's handler util. Will look from that direction, and close this later as such.

https://github.com/Azure/azure-linux-extensions/blob/master/Utils/HandlerUtil.py#L178

ahmetalpbalkan:
@hosungsmsft yes, and erroring out if `version <> 1.*` might be the best practice to apply.

ahmetalpbalkan:
@hosungsmsft also please open PRs from your forked repo, otherwise it causes new unattended branches in Azure repo which needs to be cleaned up by humans.

PR #193, opened 2016-05-26
Remove deprecated --init-system flag from setup.py

### LINKED ISSUE #177: [2.1] "--init-system" gets ignored completely
It is called "deprecated" (not removed completely), but still completely ignored.

ahmetalpbalkan:
Apparently the value of this flag was already ignored and
only a warning was getting written although it is documented as
_deprecated_.

Fixes #177

cc: @hglkrijger @szarkos

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -124,8 +124,6 @@ def get_data_files(name, version, fullname):
 
 class install(_install):
     user_options = _install.user_options + [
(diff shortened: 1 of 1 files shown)


hglkrijger:
lgtm, we'll need to cherry pick this after the branch cleanup

Issue #196, opened 2016-05-30
Hostname is NULL after reboot FreeBSD vm in Azure


KylieLiang:
The hostname of a vm of  FreeBSD 10.3 with waagent 2.1.4  is NULL after reboot it  in Azure.
Could you please have a look at it?  The agent should configure the right hostname when booting, right?

 (The hostname is NULL after reboot/restart it.)

KylieLiang:
Dong did investigation. The host name is not persisted after rebooting. We’ll write hostname to ‘/etc/hostname’, which Linux will read and set hostname. But FreeBSD will read ‘/etc/rc.conf’ only, the result is a FreeBSD VM would lost its Azure VM hostname after rebooting.

Dong did a private fix for FreeBSD 10.3 which will be published in Azure marketplace. https://github.com/karataliu/WALinuxAgent/tree/freebsd21 
The test for above fix is in the progress. Once it is done, could it be included in Agent 2.1.5? Thank you.

KylieLiang:
We tested above fix with FreeBSD 10.3 and confirmed this bug is fixed. 
Could you please ensure this fix is captured by 2.1.5? Thank you.

hglkrijger:
closing, as the fix is merged.

PR #198, opened 2016-05-31
[2.1] Support persistent hostname for FreeBSD.

karataliu:
FreeBSD uses '/etc/rc.conf' instead of '/etc/hostname' for hostname configuration. The following change makes hostname persistent after rebooting on FreeBSD.

### DIFF
--- a/azurelinuxagent/distro/freebsd/osutil.py
+++ b/azurelinuxagent/distro/freebsd/osutil.py
@@ -30,7 +30,10 @@ class FreeBSDOSUtil(DefaultOSUtil):
         self._scsi_disks_timeout_set = False
 
     def set_hostname(self, hostname):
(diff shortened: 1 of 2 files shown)


karataliu:
Related issue: https://github.com/Azure/WALinuxAgent/issues/196

brendandixon:
@karataliu I'm a little uneasy with this change. I agree that, for FreeBSD, /etc/rc.conf is the correct location. But that file may contain other settings as well, including a previously set hostname. It may be better if the code scanned for and updated the hostname line.

karataliu:
Yeah, /etc/rc.conf also supports other settings, and a config line would override previous one with identical config name. So the append approach would work. 

Anyway it's better to make the file more clear. I've updated the PR by adding the logic for updating the existing hostname line.

PR #202, opened 2016-06-01
Separate provision agent and guest agent

brendandixon:
This PR introduces the self-update feature as well as more cleanly separating various functionality.


ahmetalpbalkan:
which part updates the freebsd support?

brendandixon:
@ahmetalpbalkan Various bits update FreeBSD. If you "git log --pretty=oneline" and find the split point from 2.1.3, you can see the individual commits included in this. It's what I had to do...

azurelinuxagent/ga/update.py
```diff
@@ -0,0 +1,278 @@
...
+class UpdateHandler(object):
+
+    def __init__(self):
+        self.osutil = get_osutil()
+        self.protocol_util = get_protocol_util()
+        self.last_etag = None
+        self.agents = []
+        self.error_record = {}
+
+    def run(self):
```
ahmetalpbalkan: this method probably should be full of log statements.

azurelinuxagent/ga/update.py
```diff
@@ -0,0 +1,278 @@
...
+
+            if not latest_agent.is_downloaded():
+                try:
+                    latest_agent.download()
+                except Exception as e:
+                    raise UpdateError(u"Download failed", e)
+
+        devnull = open(os.devnull, 'w')
+        try:
+            child = subprocess.Popen([agent_bin, 'run-exthandlers'],
```
ahmetalpbalkan: it looks like we relaunch the `waagent run-exthandlers` with `2>/dev/null 1>/dev/null`. I'm not sure if this losing output is a good idea. Does the agent log itself?

azurelinuxagent/ga/update.py
brendandixon: @ahmetalpbalkan The invoked methods generally log a good deal. (I assume you mean "run" vs. the __init__ method.)

azurelinuxagent/ga/update.py
brendandixon: @ahmetalpbalkan Yes, the agent sets up its own logger.

ahmetalpbalkan: @brendandixon I wonder how the pidfile is managed if waagent is executed twice, and it appears like there might be a write/write conflict resulting in loss of data in waagent.log

azurelinuxagent/ga/update.py
ahmetalpbalkan: I mean the steps in pydoc below are clear but while they are executing, `waagent -update` will actually start the extension handler. that is kind of unexpected to me, and I bet to other people as well. we should probably say log statements like

    No guest agent update applied, starting extension handler

azurelinuxagent/agent.py
```diff
@@ -74,15 +123,24 @@ def main():
...
+                agent.deprovision(force, deluser=True)
+            elif command == "provision":
+                agent.provision()
+            elif command == "deprovision":
+                agent.deprovision(force, deluser=False)
+            elif command == "register-service":
+                agent.register_service()
+            elif command == "daemon":
+                agent.daemon()
+            elif command == "update":
```
ahmetalpbalkan: is `waagent -update` the new entrypoint if we want to have self-update + GA functionality?

azurelinuxagent/ga/update.py
brendandixon: @ahmetalpbalkan You are correct. The current code integrates update as a pre-step to extension handling. We could add flags to tweak that.

azurelinuxagent/ga/update.py
brendandixon: @ahmetalpbalkan daemon/main.py contains a check_pid method that looks for / exits if it finds an existing pidfile (whose value corresponds to a running process) and exits if so. I believe it could be more robust.

azurelinuxagent/agent.py
brendandixon: @ahmetalpbalkan No, -daemon remains the proper entry point. daemon/main.py is the main loop.

Issue #146, opened 2016-01-13
CustomData not handled correctly when Provisioning.DecodeCustomData=n


chkp-zegman:
When you set:
Provisioning.DecodeCustomData=n
The following code raises an exception:
SetFileContents(LibDir + '/CustomData', bytearray(MyDistro.translateCustomData(self.CustomData)))
The reason is that:
self.CustomData is a unicode string.
The value returned from MyDistro.translateCustomData is the same unicode string.
bytearray on a unicode string raises an exception.
As a result, no CustomData is written to the disk.

chkp-zegman:
A suggested change is to call bytearray with a 2nd encoding argument thus:
SetFileContents(LibDir + '/CustomData', bytearray(MyDistro.translateCustomData(self.CustomData), 'utf-8'))

hglkrijger:
fix merged into develop

PR #203, opened 2016-06-01
Encoding fix [2.0.x]

hglkrijger:
In case we ever need it, porting fix from issue #146 to the 2.0 agent in the 2.1 branch; note that in the 2.1 agent itself, fileutil.write_file uses utf-8 encoding by default, therefore the same bug should not exist.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,9 @@ var/
 .installed.cfg
 *.egg
 
+# PyCharm
+.idea/
+
 # PyInstaller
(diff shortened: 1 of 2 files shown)


ahmetalpbalkan:
It would be squash into one commit before merging. Merge commits are not very preferable in PRs.

hglkrijger:
sure thing, updated

ahmetalpbalkan:
Lgtm

Issue #169, opened 2016-04-05
[WAagent 2.1] Should evaluate ssh pub key file before using it


karataliu:
Should evaluate the pub key file before generate thumbprint, but now it looks at the private key file.

1. Check private key (should be .pub)
https://github.com/Azure/WALinuxAgent/blob/5bfc685/azurelinuxagent/distro/ubuntu/provision.py#L91

2. Generate thumbprint:
https://github.com/Azure/WALinuxAgent/blob/78d9a9d/azurelinuxagent/distro/default/provision.py#L86

ahmetalpbalkan:
@karataliu I am not sure if I understand this. 

Are you saying `get_ssh_host_key_thumbprint` and `wait_for_ssh_host_key` are somewhat related?

It looks like `wait_for_ssh_host_key` can work just fine by looking at the private key. I don't see an issue.

hglkrijger:
@karataliu, does the [referenced commit](https://github.com/hglkrijger/WALinuxAgent/commit/26c972a3796639ed95d4c3555d923ca3504bbd66) implement what you had in mind?

ahmetalpbalkan:
@hglkrijger I'm not sure if that change should be made. The keys and public keys are always generated together and `wait_for_ssh_host_key` is just used as an indicator to find out if cloud-init is completed. The change essentially does not change anything.

hglkrijger:
@ahmetalpbalkan, I *think* the concern behind the original issue was that we check for the existence of one file and then load another. Although we expect them to be generated together, it makes sense to check for the file we are going to load, doesn't it?

ahmetalpbalkan:
@hglkrijger hmm I thought that function was computing the thumbprint of user-provided OpenSSH public key. 

I just realized we actually have some logic that sends the ssh host key thumbprint in the 'provisioned' signal to the Wire Server. 

I am not sure why we do it, if this is actually necessary or cloud-init does the same as well. We should probably understand these first. In any case, it makes sense to apply the change.

szarkos:
The agent pushes the ssh host key fingerprint to the fabric so that it is available to the user via the portal or via the API (i.e. the classic Get Deployment API).  They'll want to verify this when they connect to the VM via SSH for the first time.  The original agent behavior was to push the fingerprint just once at runtime, but in some cases the host key did not exist yet (i.e. cloud-init may not have created it yet), and so wait_for_ssh_host_key() was added.

[...]

PR #205, opened 2016-06-01
Race condition [2.1]

hglkrijger:
this is related to a race condition which may exist if we check for the existence of the key file, but then try open the corresponding pub; instead check that the pub file exists before using it. addresses #169.


ahmetalpbalkan:
Could you please squash into one commit either before merging or while merging through GitHub UI?

hglkrijger:
sure, updated

ahmetalpbalkan:
lgtm

PR #206, opened 2016-06-01
Graceful failure when deploying keys [2.1]

hglkrijger:
check whether files exist before trying to chmod/chown/chcon. partially addresses #104; 2.1 does already support non-RSA keys.

### DIFF
--- a/bin/waagent2.0
+++ b/bin/waagent2.0
@@ -228,6 +228,9 @@ class AbstractDistro(object):
         Returns exit result.
         """
         if self.isSelinuxSystem():
+            if not os.path.exists(path):
+                Error("Path does not exist: {0}".format(path))
+                return 1
             return Run('chcon ' + cn + ' ' + path)
         
     def setHostname(self,name):
@@ -2336,7 +2339,10 @@ def ChangeOwner(filepath, user):
     except:
         pass
     if p != None:
-        os.chown(filepath, p[2], p[3])
+        if not os.path.exists(filepath):
+            Error("Path does not exist: {0}".format(filepath))
+        else:
+            os.chown(filepath, p[2], p[3])
 
 def CreateDir(dirpath, user, mode):
     """
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -194,7 +194,7 @@ class DefaultOSUtil(object):
         """
         path, thumbprint, value = pubkey
         if path is None:
-            raise OSUtilError("Publich key path is None")
+            raise OSUtilError("Public key path is None")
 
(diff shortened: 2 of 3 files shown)


azurelinuxagent/distro/default/osutil.py
```diff
@@ -261,7 +261,11 @@ def set_selinux_context(self, path, con):
         Returns exit result.
         """
         if self.is_selinux_system():
-            return shellutil.run('chcon ' + con + ' ' + path)
+            if not os.path.exists(path):
+                logger.error("Path does not exist: {0}".format(path))
+                return 1
+            else:
```
ahmetalpbalkan: `else` unnecessary

bin/waagent2.0
```diff
@@ -228,7 +228,11 @@ class AbstractDistro(object):
         Returns exit result.
         """
         if self.isSelinuxSystem():
-            return Run('chcon ' + cn + ' ' + path)
+            if not os.path.exists(path):
+                Error("Path does not exist: {0}".format(path))
+                return 1
+            else:
```
ahmetalpbalkan: `else` unnecessary

bin/waagent2.0
```diff
@@ -2336,7 +2340,10 @@ def ChangeOwner(filepath, user):
     except:
         pass
     if p != None:
-        os.chown(filepath, p[2], p[3])
+        if not os.path.exists(filepath):
+            Error("Path does not exist: {0}".format(filepath))
+        else:
```
ahmetalpbalkan: `else` unnecessary

hglkrijger:
updated with feedback and squashed

azurelinuxagent/distro/default/osutil.py
```diff
@@ -261,6 +261,9 @@ def set_selinux_context(self, path, con):
         Returns exit result.
         """
         if self.is_selinux_system():
+            if not os.path.exists(path):
```
ahmetalpbalkan: I feel like this checks should not be inside the `set_selinux_context`/`chcon`/`chown` methods. Why don't we check the file before they get called? It does not make much sense to call multiple operations on a file that does not exist, plus this patch adds non-obvious logic to method names like chown/chmod.

hglkrijger: These methods are called from many locations (especially chmod), and adding checks everywhere is both messy and error-prone since any future calls may not enforce those checks. As there is no side-effect from the additional code, but rather the unwanted side-effect of failure is avoided, I feel the change is reasonable.

ahmetalpbalkan: Fair, we discussed this with Brendan a bit as well. There are arguments for both, but we can live with these changes. So LGTM.

PR #208, opened 2016-06-01
Debian patch [2.1]

hglkrijger:
port of debian patch disable_udev_overrides; brings develop in line with production version 2.1.3

/cc @szarkos


ahmetalpbalkan:
Lgtm

PR #209, opened 2016-06-02
- Implementation for RDMA driver handler

rjschwei:
+ This functions as a base class. Distributions that support RDMA must
    implement a sub class and at least implement install_driver().


ahmetalpbalkan:
@rjschwei Could you please send a PR to `develop` branch? Sorry for the inconvenience.

rjschwei:
Based on phone call with Steve and Dan earlier today I was under the impression that repo reorg was completed and PRs should be sent to master??

ahmetalpbalkan:
Unfortunately we decided `develop` to be the tip and `master` to be a more stable(?) branch. I personally have an argument that in this case `develop` should be the default branch, your PR has just proved my point.

PR #210, opened 2016-06-02
Implementation for RDMA driver handler

rjschwei:
This functions as a base class. Distributions that support RDMA must implement a sub class and at least implement install_driver(). How RDMA handling needs to be integrated into the overall process flow of the agent is still an open discussion point. However, the implementation of the handler itself should not require much discussion.

### DIFF
--- a/azurelinuxagent/distro/default/rdma.py
+++ b/azurelinuxagent/distro/default/rdma.py
@@ -1,6 +1,6 @@
 # Windows Azure Linux Agent
 #
-# Copyright 2016 Microsoft Corporation, SUSE LLC, Robert Schweikert
+# Copyright 2016 Microsoft Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -32,7 +32,7 @@ class RDMAHandler(object):
         base_kernel_err_msg = 'Kernel does not provide the necessary '
         base_kernel_err_msg += 'information or the hv_kvp_daemon is not '
         base_kernel_err_msg += 'running.'
-        if not os.path.exists(driver_info_source):
+        if not os.path.isfile(driver_info_source):
             error_msg = 'Source file "%s" does not exist. '
             error_msg += base_kernel_err_msg
             logger.error(error_msg % driver_info_source)
@@ -54,11 +54,11 @@ class RDMAHandler(object):
             logger.error(error_msg % driver_info_source)
             return
 
-    def initialize_driver(self):
+    def load_driver_module(self):
         """Load the kernel driver, this depends on the proper driver
            to be installed with the install_driver() method"""
-        driver_module_name = 'modprobe hv_network_direct'
-        result = os.system(driver_module_name)
+        driver_module_name = 'hv_network_direct'
+        result = os.system('modprobe %s' driver_module_name)
         if result != 0:
             error_msg = 'Could not load "%s" kernel module. '
             error_msg += 'Run "modprobe %s" as root for more details'
@@ -69,6 +69,6 @@ class RDMAHandler(object):
 
     def install_driver(self):
         """Install the driver. This is distribution specific and must
-           be overwritted in the child implementation."""
+           be overwritten in the child implementation."""
 
         raise Exception('RDMAHandler.install_driver not implemented')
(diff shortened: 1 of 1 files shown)


ahmetalpbalkan:
The change looks good to me. Do you mind rebasing it to resolve merge conflicts introduced by a recent PR? Thanks.

azurelinuxagent/distro/default/rdma.py
```diff
@@ -0,0 +1,74 @@
...
+            error_msg = 'Could not load "%s" kernel module. '
+            error_msg += 'Run "modprobe %s" as root for more details'
+            logger.error(error_msg % (driver_module_name, driver_module_name))
+            return
+
+        return True
+
+    def install_driver(self):
+        """Install the driver. This is distribution specific and must
+           be overwritted in the child implementation."""
```
ahmetalpbalkan: overwritten*

azurelinuxagent/distro/default/rdma.py
```diff
@@ -0,0 +1,74 @@
...
+            return NdDriverVersion
+        else:
+            error_msg = 'NdDriverVersion not found in "%s"'
+            logger.error(error_msg % driver_info_source)
+            return
+
+    def initialize_driver(self):
+        """Load the kernel driver, this depends on the proper driver
+           to be installed with the install_driver() method"""
+        driver_module_name = 'modprobe hv_network_direct'
```
ahmetalpbalkan: this would be better named as `driver_probe_command`

azurelinuxagent/distro/default/rdma.py
```diff
@@ -0,0 +1,74 @@
...
+class RDMAHandler(object):
+    def __get_rdma_version(self):
+        """Retrieve the firmware version information from the system.
+           This depends on information provided by the Linux kernel."""
+
+        driver_info_source = '/var/lib/hyperv/.kvp_pool_0'
+        base_kernel_err_msg = 'Kernel does not provide the necessary '
+        base_kernel_err_msg += 'information or the hv_kvp_daemon is not '
+        base_kernel_err_msg += 'running.'
+        if not os.path.exists(driver_info_source):
```
ahmetalpbalkan: a `os.path.isfile` check would be better. it contains the `exists` check as well.

azurelinuxagent/distro/default/rdma.py
```diff
@@ -0,0 +1,74 @@
+# Windows Azure Linux Agent
+#
+# Copyright 2016 Microsoft Corporation, SUSE LLC, Robert Schweikert
```
ahmetalpbalkan: It appears like our guidance on this is: new source files should follow the repo standard. Could you keep this the same with the rest of the repo?

azurelinuxagent/distro/default/rdma.py
rjschwei: On 06/03/2016 06:13 PM, Ahmet Alp Balkan wrote:
>> +            logger.error(error_msg % driver_info_source)
>> +            return
>> +
>> +        r = re.search("NdDriverVersion\0+(\d\d\d\.\d)", lines)
>> +        if r:
>> +            NdDriverVersion = r.groups()[0]
>> +            return NdDriverVersion
>> +        else:
>> +            error_msg = 'NdDriverVersion not found in "%s"'
>> +            logger.error(error_msg % driver_info_source)
>> +            return
>> +
>> +    def initialize_driver(self):
>> +        """Load the kernel driver, this depends on the proper driver
>> +           to be installed with the install_driver() method"""
>> +        driver_module_name = 'modprobe hv_network_direct'
> 
> this would be better named as `driver_probe_command`

Actually not, an implementation of a child and specialization by a
distribution implementation my prefer the "insmod" command.
"load_driver_module" would be another option and possibly is more
accurate, will use that name.


-- 
Robert Schweikert                   MAY THE SOURCE BE WITH YOU
Public Cloud Architect                         LINUX
rjschwei@suse.com
IRC: robjo

azurelinuxagent/distro/default/rdma.py
rjschwei: On 06/03/2016 06:28 PM, Ahmet Alp Balkan wrote:
>> @@ -0,0 +1,74 @@
>> +# Windows Azure Linux Agent
>> +#
>> +# Copyright 2016 Microsoft Corporation, SUSE LLC, Robert Schweikert
> 

All done, ready for merge


-- 
Robert Schweikert                   MAY THE SOURCE BE WITH YOU
Public Cloud Architect                         LINUX
rjschwei@suse.com
IRC: robjo

ahmetalpbalkan:
@rjschwei Thanks for the pull request. Will you be contributing a driver for SUSE as well?

PR #211, opened 2016-06-03
Improved logic to find non-lo interfaces

### LINKED ISSUE #188: Need logic improvement to find non loop back interface
Source: https://github.com/Azure/WALinuxAgent/blob/2.0/waagent
Function related to the query: GetFirstActiveNetworkInterfaceNonLoopback()
 
The function in question creates a socket to query the list of network interfaces. The function is returning the first one that isn’t lo. If there is a preferred NIC set in its config, it will prefer that one if it’s present. 
 
But what happens, if customer defines an alias to the loop back interface? Does it create any issues because alias like lo:1 is considered as non-loop back interface by that function? In general, I haven’t seen cases where customer creates an alias to loop back interface. I wonder, why would they do that? But, if customer does indeed create an alias to loopback interfaces, how can we avoid any issues if persists?

ahmetalpbalkan:
This will treat interface names such as 'lo:1' as loopback
as well. The ultimate solution is to use sysfs+procfs in this
method to check the flags and find the ipv4 addr. However sticking
to an extended check for now.

Fixes #188.

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -430,7 +430,7 @@ class DefaultOSUtil(object):
         sock = buff.tostring()
         for i in range(0, struct_size * expected, struct_size):
             iface=sock[i:i+16].split(b'\0', 1)[0]
-            if iface == b'lo':
+            if iface == b'lo' or iface.startswith('lo:'):
                 continue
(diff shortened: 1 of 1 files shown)


jasonzio:
With that caveat... lgtm.

PR #216, opened 2016-06-06
Updates error messages

### LINKED ISSUE #199: Fix error message for Provisioning Agent
In version 2.0.18 of waagent, if the provisioning agent isn't enabled, the user will be unable to install extensions and they'll get this error in their logs.
`Unable to send status reports`
This comes from the following code in waagent
```
        try:
            uri=GetNodeTextData(self.Extensions[0].getElementsByTagName("StatusUploadBlob")[0]).replace('&amp;','&')
        except:
            Error('Error parsing ExtensionsConfig.  Unable to send status reports')
            return None
```

This error message should be updated to reflect the underlying error (e.g. "Unable to send status reports. Is provisioning agent installed?"). 
Further, an identical error is emitted from the same function if the goal state incarnation cannot be found. These error messages should be different to simplify troubleshooting.
```
        try:
            incarnation=self.Extensions[0].getAttribute("goalStateIncarnation")
        except:
            Error('Error parsing ExtensionsConfig.  Unable to send status reports')
            return None
```

hglkrijger:
Updates error messages to distinguish between them.

Fixes #199


ahmetalpbalkan:
LGTM.

PR #217, opened 2016-06-06
Port over retry logic

hglkrijger:
The 2.0 agent contains retry logic when restarting the network interface, [here](https://github.com/Azure/WALinuxAgent/blob/b05e389088d5a6d741dd4dfd7d05624bad09e297/bin/waagent2.0#L474-480).

This ports the logic over and adds a simple unit test.

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -516,8 +516,18 @@ class DefaultOSUtil(object):
                                       'send host-name',
                                       'send host-name "{0}";'.format(hostname))
 
-    def restart_if(self, ifname):
-        shellutil.run("ifdown {0} && ifup {1}".format(ifname, ifname))
+    def restart_if(self, ifname, retries=3, wait=5):
+        retry_limit=retries+1
+        for attempt in range(1, retry_limit):
+            return_code=shellutil.run("ifdown {0} && ifup {0}".format(ifname))
+            if return_code == 0:
+                return
+            logger.warn("failed to restart {0}: return code {1}".format(ifname, return_code))
+            if attempt < retry_limit:
+                logger.info("retrying in {0} seconds".format(wait))
+                time.sleep(wait)
+            else:
+                logger.warn("exceeded restart retries")
 
     def publish_hostname(self, hostname):
         self.set_dhcp_hostname(hostname)
--- a/tests/common/osutil/__init__.py
+++ b/tests/common/osutil/__init__.py
@@ -1,0 +1,16 @@
+# Copyright 2014 Microsoft Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


azurelinuxagent/common/osutil/default.py
```diff
@@ -516,8 +516,15 @@ def set_dhcp_hostname(self, hostname):
                                       'send host-name',
                                       'send host-name "{0}";'.format(hostname))
 
-    def restart_if(self, ifname):
-        shellutil.run("ifdown {0} && ifup {1}".format(ifname, ifname))
+    def restart_if(self, ifname, retries=3, wait=5):
+        for attempt in range(1, retries+1, 1):
```
ahmetalpbalkan: `,1)` is not necessary

tests/distro/test_default.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2014 Microsoft Corporation
```
ahmetalpbalkan: I'm not familiar with python testing much but shouldn't this file be `tests/common/osutil/test_default.py`?

azurelinuxagent/common/osutil/default.py
```diff
@@ -516,8 +516,15 @@ def set_dhcp_hostname(self, hostname):
...
+            return_code=shellutil.run("ifdown {0} && ifup {1}".format(ifname, ifname))
+            if return_code == 0:
+                return
+            logger.warn("failed to restart {0}: return code {1}".format(ifname, return_code))
+            if attempt < retries:
+                logger.info("retrying in {0} seconds".format(wait))
+                time.sleep(wait)
```
ahmetalpbalkan: a log message saying `exceeded retries` would be nice too.

ahmetalpbalkan: I wonder why they didn't save and return the result of `restart_if` in the first place. Does `shellutil.run` throw and exception (I guess not)? 

In any case it might be helpful to see that this method returns an exception or some error indicator when retries are exceeded and the agent fails accordingly. I'm not sure if it's OK to continue in this particular failure mode.

ahmetalpbalkan:
Ideally we could write a method named `run_with_retries` which takes a command, or a lambda function and would isolate the app logic from the retry concern. I'm assuming this is like a one-time thing, so it is fine.

azurelinuxagent/common/osutil/default.py
```diff
@@ -516,8 +516,15 @@ def set_dhcp_hostname(self, hostname):
                                       'send host-name',
                                       'send host-name "{0}";'.format(hostname))
 
-    def restart_if(self, ifname):
-        shellutil.run("ifdown {0} && ifup {1}".format(ifname, ifname))
+    def restart_if(self, ifname, retries=3, wait=5):
+        for attempt in range(1, retries+1, 1):
+            return_code=shellutil.run("ifdown {0} && ifup {1}".format(ifname, ifname))
```
jasonzio: `return_code=shellutil.run("ifdown {0} && ifup {0}".format(ifname))` would be clearer.

tests/distro/test_default.py
hglkrijger: for clarity yes it should, though it would be found by the test runner either way; moved it over.

azurelinuxagent/common/osutil/default.py
hglkrijger: sadly there is no emoji, so `<facepalm/>`; fixed.

ahmetalpbalkan:
LGTM

PR #219, opened 2016-06-07
Fixes for SCVMM

### LINKED ISSUE #185: 2.1 scvmm detection and support broken
It seems scvmm detection and integration is completely broken now. 
Daemon calls for:

```
        if conf.get_detect_scvmm_env():
            if self.distro.scvmm_handler.run():
                return
```

While `ScvmmHandler` class does not have method `run()` at all.
When I corrected for that I got error:
`UbuntuOSUtil object has no attribute 'get_dvd_mount_point'`

hglkrijger:
- correct scvmm references
- add unit tests for scvmm detection
- ensure all matching devices are mounted and examined for scvmm configuration
- exit once vmm startup script is found and executed
- fixes #185

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -280,25 +280,27 @@ class DefaultOSUtil(object):
 
 
     def get_dvd_device(self, dev_dir='/dev'):
-        patten=r'(sr[0-9]|hd[c-z]|cdrom[0-9]|cd[0-9])'
-        for dvd in [re.match(patten, dev) for dev in os.listdir(dev_dir)]:
+        pattern=r'(sr[0-9]|hd[c-z]|cdrom[0-9]|cd[0-9])'
+        for dvd in [re.match(pattern, dev) for dev in os.listdir(dev_dir)]:
             if dvd is not None:
                 return "/dev/{0}".format(dvd.group(0))
         raise OSUtilError("Failed to get dvd device")
 
-    def mount_dvd(self, max_retry=6, chk_err=True):
-        dvd = self.get_dvd_device()
-        mount_point = conf.get_dvd_mount_point()
+    def mount_dvd(self, max_retry=6, chk_err=True, dvd_device=None, mount_point=None):
+        if dvd_device is None:
+            dvd_device = self.get_dvd_device()
+        if mount_point is None:
+            mount_point = conf.get_dvd_mount_point()
         mountlist = shellutil.run_get_output("mount")[1]
-        existing = self.get_mount_point(mountlist, dvd)
+        existing = self.get_mount_point(mountlist, dvd_device)
         if existing is not None: #Already mounted
-            logger.info("{0} is already mounted at {1}", dvd, existing)
+            logger.info("{0} is already mounted at {1}", dvd_device, existing)
             return
         if not os.path.isdir(mount_point):
             os.makedirs(mount_point)
 
         for retry in range(0, max_retry):
-            retcode = self.mount(dvd, mount_point, option="-o ro -t udf,iso9660",
+            retcode = self.mount(dvd_device, mount_point, option="-o ro -t udf,iso9660",
                                  chk_err=chk_err)
             if retcode == 0:
                 logger.info("Successfully mounted dvd")
@@ -310,8 +312,9 @@ class DefaultOSUtil(object):
         if chk_err:
             raise OSUtilError("Failed to mount dvd.")
 
-    def umount_dvd(self, chk_err=True):
-        mount_point = conf.get_dvd_mount_point()
+    def umount_dvd(self, chk_err=True, mount_point=None):
+        if mount_point is None:
+            mount_point = conf.get_dvd_mount_point()
         retcode = self.umount(mount_point, chk_err=chk_err)
         if chk_err and retcode != 0:
             raise OSUtilError("Failed to umount dvd.")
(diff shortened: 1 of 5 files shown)


azurelinuxagent/common/osutil/default.py
```diff
@@ -280,15 +280,17 @@ def conf_sshd(self, disable_password):
...
+        pattern=r'(sr[0-9]|hd[c-z]|cdrom[0-9]|cd[0-9])'
+        for dvd in [re.match(pattern, dev) for dev in os.listdir(dev_dir)]:
             if dvd is not None:
                 return "/dev/{0}".format(dvd.group(0))
         raise OSUtilError("Failed to get dvd device")
 
-    def mount_dvd(self, max_retry=6, chk_err=True):
-        dvd = self.get_dvd_device()
-        mount_point = conf.get_dvd_mount_point()
+    def mount_dvd(self, max_retry=6, chk_err=True, dvd=None, mount_point=None):
```
ahmetalpbalkan: maybe a variable name better than `dvd` would be better? like `dvd_device`?

azurelinuxagent/daemon/scvmm.py
```diff
@@ -34,21 +35,34 @@ class ScvmmHandler(object):
     def __init__(self):
         self.osutil = get_osutil()
 
-    def detect_scvmm_env(self):
+    def detect_scvmm_env(self, dev_dir='/dev/'):
```
ahmetalpbalkan: `/dev` would be better.

azurelinuxagent/daemon/scvmm.py
```diff
@@ -34,21 +35,34 @@ class ScvmmHandler(object):
...
+
+        # cycle through all available /dev/sr*|hd*|cdrom*|cd* looking for the scvmm configuration file
         mount_point = conf.get_dvd_mount_point()
-        found = os.path.isfile(os.path.join(mount_point, VMM_CONF_FILE_NAME))
-        if found:
-            self.start_scvmm_agent()
-        else:
-            self.osutil.umount_dvd(chk_err=False)
+        for dvds in [re.match(r'(sr[0-9]|hd[c-z]|cdrom[0-9]?|cd[0-9]+)', dev) for dev in os.listdir(dev_dir)]:
+            if dvds is None:
```
ahmetalpbalkan: I think `dvds is None:` will not hold true inside the loop

ahmetalpbalkan: oh wait I just realized what you're doing. try using `filter` function.

hglkrijger:
addressed feedback - @ahmetalpbalkan, is this what you had in mind for `filter`?

azurelinuxagent/daemon/scvmm.py
```diff
@@ -34,21 +35,35 @@ class ScvmmHandler(object):
...
+        self.osutil.try_load_atapiix_mod()
+
+        # cycle through all available /dev/sr*|hd*|cdrom*|cd* looking for the scvmm configuration file
         mount_point = conf.get_dvd_mount_point()
-        found = os.path.isfile(os.path.join(mount_point, VMM_CONF_FILE_NAME))
-        if found:
-            self.start_scvmm_agent()
-        else:
-            self.osutil.umount_dvd(chk_err=False)
+        for devices in filter(self.isNotEmpty, [re.match(r'(sr[0-9]|hd[c-z]|cdrom[0-9]?|cd[0-9]+)', dev) for dev in os.listdir(dev_dir)]):
```
ahmetalpbalkan: hmm kind of.. probably can be done without `self.isNotEmpty`. let me try.

ahmetalpbalkan: you just need to do this:

    filter(None.__ne__, ...)

and it will eliminate `None` values from the list

ahmetalpbalkan: also you can do the `group(0)` in the `for` line:

    for device in [d.group(0) in filter(...

hglkrijger: That is a Python 3 feature. The additional method seems reasonably clean to me.

ahmetalpbalkan: @hglkrijger we only support python 3, though?

I find extra method not so clean, you can also do `lambda x: x is not None` in the filter.

azurelinuxagent/common/osutil/default.py
```diff
@@ -322,7 +325,13 @@ def eject_dvd(self, chk_err=True):
         if chk_err and retcode != 0:
             raise OSUtilError("Failed to eject dvd: ret={0}".format(retcode))
 
-    def load_atappix_mod(self):
+    def try_load_atapiix_mod(self):
+        try:
+            self.load_atapiix_mod()
+        except:
+            logger.warn("could not load ATAPI driver")
```
ahmetalpbalkan: we are effectively swallowing the meaningful exception message(s) from `load_atapiix_mod` here.

hglkrijger: yes, we are, since we can still continue without it, hence the level of warning. if you think it's important we can add verbose logs to the load method as well?

azurelinuxagent/daemon/scvmm.py
hglkrijger: updated

azurelinuxagent/common/osutil/default.py
hglkrijger: updated

hglkrijger:
addressed feedback

azurelinuxagent/common/osutil/default.py
```diff
@@ -322,7 +325,13 @@ def eject_dvd(self, chk_err=True):
         if chk_err and retcode != 0:
             raise OSUtilError("Failed to eject dvd: ret={0}".format(retcode))
 
-    def load_atappix_mod(self):
+    def try_load_atapiix_mod(self):
+        try:
+            self.load_atapiix_mod()
+        except Exception as e:
+            logger.warn("Could not load ATAPI driver: {0}".format(e.message))
```
ahmetalpbalkan: I believe you'll get `AttributeError: 'Exception' object has no attribute 'message'` here?

hglkrijger: It works in Python 2, but you are right, this was deprecated in Python 3. Fixed - thanks for catching

hglkrijger:
@ahmetalpbalkan - any more feedback, or can we merge this one?

ahmetalpbalkan:
LGTM

PR #221, opened 2016-06-07
Update agent config for CoreOS

crawford:
These are the settings that ship with CoreOS currently. We have our own [copy of this config](https://github.com/coreos/coreos-overlay/blob/master/app-emulation/wa-linux-agent/files/waagent.conf), but we might as well try to keep them in sync.


ahmetalpbalkan:
LGTM. I believe we keep our own for just documentation/reference purposes.

PR #226, opened 2016-06-08
Add monitoring of fstab changes

CraigWiand:
Microsoft CSS is requesting additions to the Azure Linux agent to
validate fstab file changes prior to a VM reboot.  We believe this will
help prevent no-boot conditions brought about by incorrect or malformed
fstab files.  Please find more info in the 1-pager below

[fstab checking_public_1-pager.docx](https://github.com/Azure/WALinuxAgent/files/305715/fstab.checking_public_1-pager.docx)

### DIFF
--- a/azurelinuxagent/distro/default/env.py
+++ b/azurelinuxagent/distro/default/env.py
@@ -21,6 +21,8 @@ import os
 import socket
 import threading
 import time
+import datetime
+import azurelinuxagent.utils.shellutil as shellutil
 import azurelinuxagent.logger as logger
 import azurelinuxagent.conf as conf
 
@@ -38,6 +40,8 @@ class EnvHandler(object):
         self.hostname = None
         self.dhcpid = None
         self.server_thread=None
+        self.lastNotice = datetime.datetime.min
+        self.fstabModifiedStamp = None
 
     def run(self):
         if not self.stopped:
@@ -45,6 +49,7 @@ class EnvHandler(object):
             self.stop()
 
         self.stopped = False
+        self.fstabModifiedStamp = shellutil.run_get_output("stat /etc/fstab -c %Y")
         logger.info("Start env monitor service.")
         self.distro.dhcp_handler.conf_routes()
         self.hostname = socket.gethostname()
@@ -66,6 +71,7 @@ class EnvHandler(object):
             if conf.get_monitor_hostname():
                 self.handle_hostname_update()
             self.handle_dhclient_restart()
+            self.handle_fstab_update()
             time.sleep(5)
 
     def handle_hostname_update(self):
@@ -94,6 +100,22 @@ class EnvHandler(object):
            self.distro.dhcp_handler.conf_routes()
            self.dhcpid = newpid
 
+     def handle_fstab_update(self):
+        # Only try to mount if the fstab has been modified. This avoids background mounts when admins are working on a unmounted partition
+        # If we find an error, backoff and only check every minute to give time to correct it and avoid filling the log
+        fstabCurrentModifiedStamp = shellutil.run_get_output("stat /etc/fstab -c %Y")
+        if (fstabCurrentModifiedStamp != self.fstabModifiedStamp) and (datetime.datetime.now() > self.lastNotice + datetime.timedelta(seconds=60)):
+            ret, output = shellutil.run_get_output("mount -av")
+            if ret != 0:
+                # Notify the logged on users who just modified the file to take action, immediately and then periodically thereafter...
+                notice =  "[AZURE AGENT] Current fstab settings are invalid.  Please correct the /etc/fstab file content before rebooting using the error information in /var/log/waagent.log"
+                shellutil.run_get_output("wall " + notice)
+                self.lastNotice = datetime.datetime.now()
+            else:
+                # no errors during the mount, avoid any warnings
+                self.fstabModifiedStamp = fstabCurrentModifiedStamp
+                logger.info("fstab file was modified and passed mount validation.")
+		   
     def stop(self):
         """
         Stop server comminucation and join the thread to main thread.


azurelinuxagent/distro/default/env.py
```diff
@@ -66,6 +71,7 @@ def monitor(self):
             if conf.get_monitor_hostname():
                 self.handle_hostname_update()
             self.handle_dhclient_restart()
+            self.handle_fstab_update()
```
ahmetalpbalkan: Method description needs updating as well now that this functionality is added.

ahmetalpbalkan:
@CraigWiand thanks, I will now review. It will be a bit complicated for us to port this onto our main development branch (`"develop"`). So if you could actually make this change on top of `develop` branch and send the PR against it (on github PR UI), it would be great. That would require closing this PR and opening a new one once you rebase your changes. If this is a lot for you to do, please let us know.

azurelinuxagent/distro/default/env.py
```diff
@@ -38,13 +40,16 @@ def __init__(self, distro):
...
+        self.lastNotice = datetime.datetime.min
+        self.fstabModifiedStamp = None
 
     def run(self):
         if not self.stopped:
             logger.info("Stop existing env monitor service.")
             self.stop()
 
         self.stopped = False
+        self.fstabModifiedStamp = shellutil.run_get_output("stat /etc/fstab -c %Y")
```
ahmetalpbalkan: please use `os.stat()` or `os.path.getmtime()`

azurelinuxagent/distro/default/env.py
```diff
@@ -94,6 +100,22 @@ def handle_dhclient_restart(self):
            self.distro.dhcp_handler.conf_routes()
            self.dhcpid = newpid
 
+     def handle_fstab_update(self):
```
ahmetalpbalkan: Please stick to 80-character wide lines. If you get notices about this, please run `pep8` command on the source code.

azurelinuxagent/distro/default/env.py
```diff
@@ -94,6 +100,22 @@ def handle_dhclient_restart(self):
...
+        # Only try to mount if the fstab has been modified. This avoids background mounts when admins are working on a unmounted partition
+        # If we find an error, backoff and only check every minute to give time to correct it and avoid filling the log
+        fstabCurrentModifiedStamp = shellutil.run_get_output("stat /etc/fstab -c %Y")
+        if (fstabCurrentModifiedStamp != self.fstabModifiedStamp) and (datetime.datetime.now() > self.lastNotice + datetime.timedelta(seconds=60)):
+            ret, output = shellutil.run_get_output("mount -av")
+            if ret != 0:
+                # Notify the logged on users who just modified the file to take action, immediately and then periodically thereafter...
+                notice =  "[AZURE AGENT] Current fstab settings are invalid.  Please correct the /etc/fstab file content before rebooting using the error information in /var/log/waagent.log"
+                shellutil.run_get_output("wall " + notice)
```
ahmetalpbalkan: I suggest surrounding the message with `"`s:

    run_get_output("wall \"{0}\".format(notice))

azurelinuxagent/distro/default/env.py
```diff
@@ -94,6 +100,22 @@ def handle_dhclient_restart(self):
...
 
+     def handle_fstab_update(self):
+        # Only try to mount if the fstab has been modified. This avoids background mounts when admins are working on a unmounted partition
+        # If we find an error, backoff and only check every minute to give time to correct it and avoid filling the log
+        fstabCurrentModifiedStamp = shellutil.run_get_output("stat /etc/fstab -c %Y")
+        if (fstabCurrentModifiedStamp != self.fstabModifiedStamp) and (datetime.datetime.now() > self.lastNotice + datetime.timedelta(seconds=60)):
+            ret, output = shellutil.run_get_output("mount -av")
+            if ret != 0:
+                # Notify the logged on users who just modified the file to take action, immediately and then periodically thereafter...
+                notice =  "[AZURE AGENT] Current fstab settings are invalid.  Please correct the /etc/fstab file content before rebooting using the error information in /var/log/waagent.log"
```
ahmetalpbalkan: where are you logging something at waagent.log? I see the `output` above is an unused variable. Might be helpful to use it in an `logger.error` perhaps?

azurelinuxagent/distro/default/env.py
```diff
@@ -38,13 +40,16 @@ def __init__(self, distro):
         self.hostname = None
         self.dhcpid = None
         self.server_thread=None
+        self.lastNotice = datetime.datetime.min
+        self.fstabModifiedStamp = None
```
ahmetalpbalkan: `fstabModTime` is better.

azurelinuxagent/distro/default/env.py
```diff
@@ -94,6 +100,22 @@ def handle_dhclient_restart(self):
            self.distro.dhcp_handler.conf_routes()
            self.dhcpid = newpid
 
+     def handle_fstab_update(self):
+        # Only try to mount if the fstab has been modified. This avoids background mounts when admins are working on a unmounted partition
+        # If we find an error, backoff and only check every minute to give time to correct it and avoid filling the log
+        fstabCurrentModifiedStamp = shellutil.run_get_output("stat /etc/fstab -c %Y")
+        if (fstabCurrentModifiedStamp != self.fstabModifiedStamp) and (datetime.datetime.now() > self.lastNotice + datetime.timedelta(seconds=60)):
```
ahmetalpbalkan: parentheses are not necessary.

also isn't `shellutil.run_get_output` returning `ret, output`? It appears like you're saving a `(ret,output)` in `self.fstabCurrentModifiedStamp` and comparing these tuples (pairs) each time? if you use `os.stat()` or `os.path.getmtime()` that problem will disappear.

azurelinuxagent/distro/default/env.py
jasonzio: Single and double quotes are interchangeable string delimiters.
`run_get_output('wall "{0}"'.format(notice))`

azurelinuxagent/distro/default/env.py
```diff
@@ -94,6 +100,22 @@ def handle_dhclient_restart(self):
...
+            ret, output = shellutil.run_get_output("mount -av")
```
ahmetalpbalkan: I think you need to do `mount --fake` here as well? Maybe I just `umount`ed something, and by running an actual `mount` command effectively you are mounting that thing back?  `mount --fake`  will validate the syntax but will not actually check devices etc, so I'm not sure if it fits your task, however there's this side effect of this call unless you call it with fake (it might be fine as it would happen on restarts anyway).

azurelinuxagent/distro/default/env.py
hglkrijger: When I tried this locally on Ubuntu 14.04, `wall` would not accept input as a parameter, only from a file or stdin, which matches the man page. Perhaps something like `run_get_output('echo "{0}" | wall'.format(notice))` is necessary here?

jasonzio: @hglkrijger you are correct about wall getting the message-to-display from stdin. Using the pipe syntax means the shell gets a crack at interpreting the text, so quoting becomes far more problematic. If you have a utility function that starts a subprocess and lets you write data to its stdin, I strongly recommend you use that.

If you have to use run_get_output, you'll need to run through `notice` and insert escapes for any single-quote characters, then wrap that in single quotes. That is...

```
quotedNotice = escapeSingleQuotes(notice)
run_get_output("echo '{0}' | wall".format(quotedNotice))
```

azurelinuxagent/distro/default/env.py
CraigWiand: I believe that run_get_output() will log to waagent.log when it sees a non-zero return code.  However, I saw at least one instance in testing it where I didn't receive the logging I expected.  I will add an explicit call to logger.error.  thanks.

azurelinuxagent/distro/default/env.py
CraigWiand: I evaluated it initially, but sadly `mount --fake` doesn't do the level of validation that I believe is necessary to catch certain conditions.  I try to illustrate this in the attached script. If you run `fstabTests.sh` and `fstabTests.sh "-f"` you will see that several of the scenarios are not caught.
[fstabTests.txt](https://github.com/Azure/WALinuxAgent/files/314265/fstabTests.txt)

In the scenario you describe above, simply doing an `umount` shouldn't cause it to remount as the `mount` code is only triggered when fstab is modified.  That said, if an admin `umount` _and_ changed fstab then that would indeed cause the unmounted partition to be mounted.  However, if the fstab is valid this should be a 1-time occurrence.  It seems like an edge case with minimal impact but I am open to feedback on this if folks feel differently.  This also goes to the principal in the [1-pager](https://github.com/Azure/WALinuxAgent/files/305715/fstab.checking_public_1-pager.docx) of 
>"A VM could rebooted at any time, hence the fstab file should be valid at all times."

ahmetalpbalkan:
Closing as another PR is opened.

PR #229, opened 2016-06-14
Add monitoring of fstab changes  (develop branch)

CraigWiand:
I am creating this new PR request based on the great feedback I received from the [previous version](https://github.com/Azure/WALinuxAgent/pull/226).  As before, Microsoft CSS is requesting additions to the Azure Linux agent to validate fstab file changes prior to a VM reboot.  We believe this will help prevent no-boot conditions brought about by incorrect or malformed fstab files.  Please find more info in the [1-pager here](https://github.com/Azure/WALinuxAgent/files/305715/fstab.checking_public_1-pager.docx)

With this updated PR I attempt to incorporate the items in the list below from the prior [conversation] (https://github.com/Azure/WALinuxAgent/pull/226).  I will also attempt to comment directly on some there.  Tagging the prior reviewers for their additional input: @ahmetalpbalkan @jasonzio  @hglkrijger

Thanks!

```
-if you could actually make this change on top of develop branch and send the PR against it (on github PR UI), it would be great
-fstabModTime is better.
-please use os.stat() or ()os.path.getmtime
-monitor method description needs updating as well now that this functionality is added.
-Please stick to 80-character wide lines. If you get notices about this, please run pep8 command on the source code
-parentheses are not necessary. - if (fstabCurrentModifiedStamp != self.fstabModifiedStamp)
-Single and double quotes are interchangeable string delimiters.   run_get_output('wall "{0}"'.format(notice))
-Perhaps something like  run_get_output('echo "{0}" | wall'.format(notice))  is necessary here?
-you are correct about wall getting the message-to-display from stdin. Using the pipe syntax means the shell gets a crack at interpreting the text, so quoting becomes far more problematic.
```

### DIFF
--- a/azurelinuxagent/daemon/env.py
+++ b/azurelinuxagent/daemon/env.py
@@ -21,6 +21,8 @@ import os
 import socket
 import threading
 import time
+import datetime
+import azurelinuxagent.common.utils.shellutil as shellutil 
 import azurelinuxagent.common.logger as logger
 import azurelinuxagent.common.conf as conf
 from azurelinuxagent.common.osutil import get_osutil
@@ -44,6 +46,8 @@ class EnvHandler(object):
         self.hostname = None
         self.dhcpid = None
         self.server_thread=None
+        self.lastNotice = datetime.datetime.min
+        self.fstabModTime = None 
 
     def run(self):
         if not self.stopped:
@@ -51,6 +55,7 @@ class EnvHandler(object):
             self.stop()
 
         self.stopped = False
+        self.fstabModTime = os.path.getmtime("/etc/fstab")
         logger.info("Start env monitor service.")
         self.dhcp_handler.conf_routes()
         self.hostname = socket.gethostname()
@@ -62,6 +67,7 @@ class EnvHandler(object):
     def monitor(self):
         """
         Monitor dhcp client pid and hostname.
+        Montor for changes in fstab, warn if content is invalid
         If dhcp clinet process re-start has occurred, reset routes.
         """
         while not self.stopped:
@@ -72,6 +78,7 @@ class EnvHandler(object):
             if conf.get_monitor_hostname():
                 self.handle_hostname_update()
             self.handle_dhclient_restart()
+            self.handle_fstab_update() 
             time.sleep(5)
 
     def handle_hostname_update(self):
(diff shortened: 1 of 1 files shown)


ahmetalpbalkan:
No need to create a new PR every time. you just need to `git push -f` to your branch. That will preserve the history on comments we left and gives us ability to check what's fixed and how.

ahmetalpbalkan:
LGTM

azurelinuxagent/daemon/env.py
```diff
@@ -100,6 +107,33 @@ def handle_dhclient_restart(self):
...
+        """
+        Look for changes in fstab, mount if the fstab has new modified stamp 
+        If error, only re-check every minute to give an admin
+         time to correct it and avoid filling the log
+        """
+        fstabCurrentModifiedStamp = os.path.getmtime("/etc/fstab")
+        if fstabCurrentModifiedStamp != self.fstabModTime and \
+                            datetime.datetime.now() >  \
+                            self.lastNotice + datetime.timedelta(seconds=60):
+            ret, output = shellutil.run_get_output("mount -av")
```
hglkrijger: I am concerned about the fact that now the agent monitoring has a side effect. I saw your comment on the other PR, but when I run the tests with -avf I get the same results as with -av. I think the mount options should be changed to include -f.

ahmetalpbalkan: @hglkrijger it does have a side effect of mounting only the points umounted since the last (re)boot, which is something that's going to happen on the next reboot (which could happen at anytime) anyway. The main problem is we're trying so hard to prevent people to shoot themselves in the foot. As a result of this side effect, we're really going to annoy someone who's trying to umount something in the fstab by remounting it just after the fstab change. However this does not really seem even like a rare use case to me.

Based on Craig's comment from other PR, --fake does not really raise issues for most of the failure modes that could lead to boot issues.

hglkrijger: @ahmetalpbalkan, there is no valid reason I see for annoying users, even if it is a small percentage. As I mentioned above, the tests with --fake give me the same results as without.

CraigWiand: @hglkrijger - I agree with your sentiment about minimizing annoyance for users, but I do not see the same results in using -avf. I tried with the following versions provisioned from Azure VM marketplace images:
- RHEL 7.2
- CentOS 7
- Ubuntu 14.4
- Ubuntu 15.10
- Ubuntu 16.04

Using [fstabTests](https://github.com/Azure/WALinuxAgent/files/314265/fstabTests.txt), with the exception of Ubuntu 14.04 (which returned a success code for all tests) the invalid GUID was caught with -avf, but not the _invalid device name_ or _malformed entry_ test.

I am happy to test with your distro+version, but given these results I fear the change would have limited benefit using -avf in lieu of -av.  For example from CentOS 7:
```
# ./fstabTest.sh -avf
=============================================
Incorrect UUID
/                        : ignored
mount: can't find UUID=1111111-1111-1111-1111-111111111111
Error code: 32
=============================================
Invalid device name - no sdf1 on this computer
/                        : ignored
/testpartion             : successfully mounted
Error code: 0
=============================================
malformed entry - same as above but reversed device and partition name
/                        : ignored
/dev/sdf1                : successfully mounted
Error code: 0
=============================================
Additional entry - just to validate
/                        : ignored
/testpartion             : successfully mounted
Error code: 0

# ./fstabTest.sh -av
=============================================
Incorrect UUID
/                        : ignored
mount: can't find UUID=1111111-1111-1111-1111-111111111111
Error code: 32
=============================================
Invalid device name - no sdf1 on this computer
/                        : ignored
mount: special device /dev/sdf1 does not exist
Error code: 32
=============================================
malformed entry - same as above but reversed device and partition name
/                        : ignored
mount: mount point /dev/sdf1 does not exist
Error code: 32
=============================================
Additional entry - just to validate
/                        : ignored
/testpartion             : successfully mounted
Error code: 0

```

hglkrijger: I was using 14.04, which explains it. If we cannot avoid side effects, then at least we are improving things for more people than we are annoying, so lgtm.

hglkrijger: I have been thinking about this, and perhaps we should explore other options. For example, if an entry in fstab is bad, it only matters if it does not fail the boot process. Perhaps we should be monitoring for things like 'nobootwait' in Ubuntu, or the equivalents, to prevent boot issues.

hglkrijger:
We have had some discussions via email which would suggest a different mechanism for this - closing the PR for now.

PR #235, opened 2016-06-15
Actually kill the extension processes after timeout

### LINKED ISSUE #234: [2.1] extension enable invocation is time unbounded
I just created an ubuntu 16.04 VM and added docker extension with no config.

Then I changed HandlerManifest.xml `"enable"` key to a new `scripts/count.sh` (chmod +x) which does:

```sh
#!/bin/bash
set -e

rm -f /var/log/count.log
exec > >(tee -ia /var/log/count.log)
exec 2>&1

for i in `seq 1 3000`; do
	echo $i - $(date) && sleep 1
done

echo Done
```

pstree:

```
  ├─python3 /usr/sbin/waagent -daemon
  │   ├─sh -c /var/lib/waagent/Microsoft.Azure.Extensions.DockerExtension-1.1.1606092330/scripts/count.sh
  │   │   └─count.sh /var/lib/waagent/Microsoft.Azure.Extensions.DockerExtension-1.1.1606092330/scripts/count.sh
  │   │       ├─count.sh /var/lib/waagent/Microsoft.Azure.Extensions.DockerExtension-1.1.1606092330/scripts/count.sh
  │   │       │   └─tee -ia /var/log/count.log
  │   │       └─sleep 1
```

It's supposed to timeout after printing for 900 seconds, however it keeps running according to pstree and the logs.

I believe the timeout is missing/buggy here.

ahmetalpbalkan:
Fixes #234.

Now the agent actually polls on the subprocesses it started.

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -599,7 +599,7 @@ class ExtHandlerInstance(object):
             raise ExtensionError("Failed to launch: {0}, {1}".format(cmd, e))
 
         retry = timeout / 5
-        while retry > 0 and child.poll == None:
+        while retry > 0 and child.poll != None:
             time.sleep(5)
             retry -= 1
         if retry == 0:
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -439,7 +439,7 @@ class ExtHandlerInstance(object):
         self.set_operation(WALAEventOperation.Enable)
 
         man = self.load_manifest()
-        self.launch_command(man.get_enable_command())
+        self.launch_command(man.get_enable_command(), timeout=300)
         self.set_handler_state(ExtHandlerState.Enabled)
         self.set_handler_status(status="Ready", message="Plugin enabled")
 


ahmetalpbalkan:
anyone has ideas why the tests are failing, the failure is very cryptic to me.

hglkrijger:
I think the problem is in TestOSUtil.test_restart, which has a bad patch. I submitted a PR, hopefully that fixes it.

ahmetalpbalkan:
@hglkrijger hmm rebased but I think the same error remains?

ahmetalpbalkan:
@hglkrijger tests are still failing on tip. Any ideas if this is noise?

hglkrijger:
I had a couple test runs go through ok, are you sure these aren't valid failures? They could be bad tests too, haven't looked closely.

hglkrijger:
Took a look at the first failures, do this file need to be mocked out?

```
157 can't open file '/tmp/TestExtension_aYqjNQ/OSTCExtensions.ExampleHandlerLinux-1.0.0/sample.py': [Errno 2] No such file or directory
```

ahmetalpbalkan:
let me undo the changes and see if this actually normally passes

ahmetalpbalkan:
@hglkrijger Looks like a test issue. It could be about the first error indeed... (it's hard to tell how come this little change could possibly raise that error)

I verified this functionality manually and it works as expected. I looked at the tests, it is a bit hard to dig in, I bet it is a test issue. Tests must have written for behavior that matches the buggy case. About to pull my hair...

ahmetalpbalkan:
let's hold off onto this, we should fix it at some point though.

hglkrijger:
closing, as this is replaced by #303

Issue #212, opened 2016-06-03
Use architected interfaces to determine IPv4 addresses of interfaces, identify loopback NICs, etc.


jasonzio:
Use sysfs, procfs, or other standard architected interfaces to check NIC flags for IFF_LOOPBACK, find the ipv4 addresses of NICs, etc. See [Issue 188](https://github.com/Azure/WALinuxAgent/issues/188) for more details.

rginjupalli:
Based on the ideas pointed in [Issue 188](https://github.com/Azure/WALinuxAgent/issues/188), 

- Do a kernel version check and availability of /sys to validate flags are available
- If above check fails, use the current logic ( if iface == b'lo' or iface.startswith('lo:'): )

hglkrijger:
this has been merged, closing the issue.

Issue #190, opened 2016-05-24
Incorrect host route set for 168.63.129.16 on machines with multiple NICs (CentOS 7)


rubeon:
Machines deployed with 2 NICs get a route to 168.63.129.16 set on second NIC with no gateway, 

The routing table winds up looking something like this:

```

Destination     Gateway         Genmask         Flags   MSS Window  irtt Iface
0.0.0.0         10.108.20.1     0.0.0.0         UG        0 0          0 eth0
0.0.0.0         10.108.20.1     0.0.0.0         UG        0 0          0 eth0
10.108.2.0      0.0.0.0         255.255.255.0   U         0 0          0 eth1
10.108.20.0     0.0.0.0         255.255.255.0   U         0 0          0 eth0
168.63.129.16   0.0.0.0         255.255.255.255 UH        0 0          0 eth1
168.63.129.16   0.0.0.0         255.255.255.255 UH        0 0          0 eth1

```
This prevents DNS, DHCP, and other stuff from working correctly. Deleting routes to 168.63.129.16 resolves the issue (if you can ever login to the machine).

The following script reproduces this issue:

```bash
#!/bin/bash

backendRGName="cr-bin"
vmNamePrefix="test-vm-72"
location="northeurope"
vmSize="Standard_D3_v2"
nic1Name="${vmNamePrefix}-default-nic"
nic2Name="${vmNamePrefix}-backnet-nic"
publisher="OpenLogic"
offer="centos"
sku="7.2"
version="latest"
prmStorageAccountName="crbin2542"
osDiskName="${vmNamePrefix}-boot"
username="centos"
password="MyP455w0_d"
vnetName="test-vnet"
subnet1="default"
subnet2="backnet"
# create the nics
azure network nic create -g $backendRGName \
		--location $location  \
		--subnet-name $subnet1  \
		--subnet-vnet-name $vnetName \
		--name $nic1Name

azure network nic create -g $backendRGName \
		--location $location  \
		--subnet-name $subnet2  \
		--subnet-vnet-name $vnetName \
		--name $nic2Name


azure vm create --resource-group $backendRGName \
        --name $vmNamePrefix$suffixNumber \
        --location $location \
        --vm-size $vmSize \
        --nic-names $nic1Name,$nic2Name \
        --os-type linux \
        --image-urn $publisher:$offer:$sku:$version \
        --storage-account-name $prmStorageAccountName \
        --storage-account-container-name test-vhds \
[...]

PR #237, opened 2016-06-16
Multi-nic improvements

hglkrijger:
Several distinct pieces here, so I have left them as separate commits for the PR. 

- update loopback detection to use SIOCGIFFLAGS and IFF_LOOPBACK
- examine the routing table for known wireserver ip, then skip the dhcp lookup and route manipulation
- ensure the first selected interface is the primary, which has the default route
- parse the wireserver from the dhcp leases file on disk; distro-specific, this is the ubuntu implementation with common parsing logic
- unit tests for each scenario
- addresses #188, #190, and #212

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -400,7 +400,7 @@ class DefaultOSUtil(object):
     def get_mac_addr(self):
         """
         Convienience function, returns mac addr bound to
-        first non-loobback interface.
+        first non-loopback interface.
         """
         ifname=''
         while len(ifname) < 2 :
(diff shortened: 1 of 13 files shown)


azurelinuxagent/common/osutil/default.py
```diff
@@ -440,14 +442,138 @@ def get_first_if(self):
             logger.warn(('SIOCGIFCONF returned more than {0} up '
                          'network interfaces.'), expected)
         sock = buff.tostring()
+        primary = self.get_primary_interface()
         for i in range(0, struct_size * expected, struct_size):
             iface=sock[i:i+16].split(b'\0', 1)[0]
-            if iface == b'lo' or iface.startswith('lo:'):
+            if len(iface) == 0 or self.is_loopback(iface) or iface != primary:
+                # test the next one
```
ahmetalpbalkan: might be nice to add logs so we can see from the logs what happened in this loop (only if this has a risk of going bad).

azurelinuxagent/common/osutil/default.py
```diff
@@ -440,14 +442,138 @@ def get_first_if(self):
...
+        """
+        OS specific, this should return the decoded endpoint of
+        the wireserver from option 245 in the dhcp leases file
+        if it exists on disk.
+        :return: The endpoint if available, or None
+        """
+        return None
+
+    @staticmethod
+    def get_endpoint_from_leases_path(pathname):
```
ahmetalpbalkan: pathname --> pathglob or pathpattern is more appropriate.

ahmetalpbalkan:
looks legit to me.

.travis.yml
```diff
@@ -9,6 +9,7 @@ python:
 install:
   #- pip install .
   #- pip install -r requirements.txt
+  - pip install --upgrade pip
```
ahmetalpbalkan: we should probably pip freeze and restore the `-r` command above sometime and disable the `install` ones below this.

azurelinuxagent/common/osutil/default.py
hglkrijger: sure, will add some logging

azurelinuxagent/common/osutil/default.py
hglkrijger: sure, will update

PR #239, opened 2016-06-16
Support package upgrade in for the RDMA driver

rjschwei:
+ A different version of the rdma driver may need to be installed when
    a VM gets rebooted as it may be deployed on different HW that has a
    different firmware version

### DIFF
--- a/azurelinuxagent/distro/default/rdma.py
+++ b/azurelinuxagent/distro/default/rdma.py
@@ -26,8 +26,8 @@ import azurelinuxagent.common.utils.shellutil as shellutil
 
 
 class RDMAHandler(object):
-    def __init__(self):
-        self.driver_module_name = 'hv_network_direct'
+
+    driver_module_name = 'hv_network_direct'
 
     def __get_rdma_version(self):
         """Retrieve the firmware version information from the system.
@@ -67,7 +67,7 @@ class RDMAHandler(object):
             error_msg = 'Could not load "%s" kernel module. '
             error_msg += 'Run "modprobe %s" as root for more details'
             logger.error(
-                error_msg % (driver_module_name, self.driver_module_name)
+                error_msg % (self.driver_module_name, self.driver_module_name)
             )
             return
 
@@ -79,8 +79,15 @@ class RDMAHandler(object):
 
         raise Exception('RDMAHandler.install_driver not implemented')
 
-    def remove_driver_module(self):
-        """Force remove the kernel driver"""
-        # Force remove the loaded driver
-        # Potential failure if the module is currently not loaded is ignored
-        shellutil.run('rmmod --force %s' % self.driver_module_name)
+    def is_driver_loaded(self):
+        """Check if the network module is loaded in kernel space"""
+        cmd = 'lsmod | grep %s' % self.driver_module_name
+        status, loaded_modules = shellutil.run_get_output(cmd)
+        if loaded_modules:
+            return True
+
+    def reboot_system(self):
+        """Reboot the system. This is required as the kernel module for
+           the rdma driver cannot be unloaded with rmmod"""
+        shellutil.run('shutdown -r now')
+        
--- a/azurelinuxagent/distro/default/rdma.py
+++ b/azurelinuxagent/distro/default/rdma.py
@@ -89,5 +89,5 @@ class RDMAHandler(object):
     def reboot_system(self):
         """Reboot the system. This is required as the kernel module for
            the rdma driver cannot be unloaded with rmmod"""
+        logger.info('System reboot')
         shellutil.run('shutdown -r now')
-        
--- a/azurelinuxagent/distro/default/rdma.py
+++ b/azurelinuxagent/distro/default/rdma.py
@@ -21,10 +21,14 @@ Handle packages and modules to enable RDMA for IB networking
 
 import os
 import re
-import azurelinuxagent.logger as logger
+import azurelinuxagent.common.logger as logger
+import azurelinuxagent.common.utils.shellutil as shellutil
 
 
 class RDMAHandler(object):
+    def __init__(self):
+        self.driver_module_name = 'hv_network_direct'
+
     def __get_rdma_version(self):
         """Retrieve the firmware version information from the system.
            This depends on information provided by the Linux kernel."""
@@ -58,12 +62,13 @@ class RDMAHandler(object):
     def load_driver_module(self):
         """Load the kernel driver, this depends on the proper driver
            to be installed with the install_driver() method"""
-        driver_module_name = 'hv_network_direct'
-        result = os.system('modprobe %s' % driver_module_name)
+        result = shellutil.run('modprobe %s' % self.driver_module_name)
         if result != 0:
             error_msg = 'Could not load "%s" kernel module. '
             error_msg += 'Run "modprobe %s" as root for more details'
-            logger.error(error_msg % (driver_module_name, driver_module_name))
+            logger.error(
+                error_msg % (driver_module_name, self.driver_module_name)
+            )
             return
 
         return True
@@ -73,3 +78,9 @@ class RDMAHandler(object):
            be overwritten in the child implementation."""
 
         raise Exception('RDMAHandler.install_driver not implemented')
+
+    def remove_driver_module(self):
+        """Force remove the kernel driver"""
+        # Force remove the loaded driver
+        # Potential failure if the module is currently not loaded is ignored
+        shellutil.run('rmmod --force %s' % self.driver_module_name)
--- a/azurelinuxagent/distro/suse/rdma.py
+++ b/azurelinuxagent/distro/suse/rdma.py
@@ -67,6 +67,9 @@ class SUSERDMAHandler(RDMAHandler):
             # Unloading the particular driver with rmmod does not work
             # We have to reboot after the new driver is installed
             if self.is_driver_loaded():
+                info_msg = 'Currently loaded driver does not match the '
+                info_msg += 'Firmware implementation, rebbot is required.'
+                logger.info(info_msg)
                 requires_reboot = True
             cmd = zypper_remove % package_name
             shellutil.run(cmd)
--- a/azurelinuxagent/distro/suse/rdma.py
+++ b/azurelinuxagent/distro/suse/rdma.py
@@ -34,11 +34,11 @@ class SUSERDMAHandler(RDMAHandler):
             error_msg += 'no driver will be installed.'
             logger.error(error_msg)
             return
-        zypper_install = 'zypper -n in'
-        zypper_remove = 'zypper -n rm'
-        zypper_search = 'zypper se -s'
+        zypper_install = 'zypper -n in %s'
+        zypper_remove = 'zypper -n rm %s'
+        zypper_search = 'zypper se -s %s'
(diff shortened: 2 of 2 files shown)


ahmetalpbalkan:
manually fixing the CLA label

azurelinuxagent/distro/suse/rdma.py
```diff
@@ -48,22 +51,35 @@ def install_driver(self):
...
+                    logger.info(info_msg % (package_name, version))
+                    return True
+                if installed == 'i':
+                    driver_package_installed = True
+
+        # If we get here the driver package is installed but the
+        # version doesn't match or no package is installed
+        if driver_package_installed:
+            self.remove_driver_module()
+            cmd = '%s &s' % (zypper_remove, package_name)
```
ahmetalpbalkan: %s?

azurelinuxagent/distro/suse/rdma.py
```diff
@@ -35,10 +35,13 @@ def install_driver(self):
             logger.error(error_msg)
             return
         zypper_install = 'zypper -n in'
+        zypper_remove = 'zypper -n rm'
```
ahmetalpbalkan: it would be good to add `%s` to commands up here.

ahmetalpbalkan:
Change looks good to me generally. Left a few comments.

@rjschwei does suse use  `waagent.conf` config keys:
	• OS.CheckRdmaDriver 
	• OS.UpdateRdmaDriver
	• OS.RdmaRepository
	• OS.PackageExtraConfig

azurelinuxagent/distro/default/rdma.py
```diff
@@ -21,10 +21,14 @@
...
 import os
 import re
-import azurelinuxagent.logger as logger
+import azurelinuxagent.common.logger as logger
+import azurelinuxagent.common.utils.shellutil as shellutil
 
 
 class RDMAHandler(object):
+    def __init__(self):
+        self.driver_module_name = 'hv_network_direct'
```
ahmetalpbalkan: can we declare this just under `class RDMAHander`

azurelinuxagent/distro/suse/rdma.py
```diff
@@ -35,10 +35,13 @@ def install_driver(self):
...
         zypper_search = 'zypper se -s'
```
ahmetalpbalkan: maybe extract these `zypper` commands just below `class RDMAHandler` (and have trailing `%s` so we can format easier)?

azurelinuxagent/distro/default/rdma.py
```diff
@@ -73,3 +78,9 @@ def install_driver(self):
            be overwritten in the child implementation."""
 
         raise Exception('RDMAHandler.install_driver not implemented')
+
+    def remove_driver_module(self):
+        """Force remove the kernel driver"""
+        # Force remove the loaded driver
```
ahmetalpbalkan: cosmetic: this line is not very necessary

ahmetalpbalkan:
also one general comment: `install_driver()` is getting long, if we could split that into multiple functions and clarify the flow in the  `install_driver()` better, that would be more favorable.

thanks @rjschwei.

rjschwei:
On 06/16/2016 07:08 PM, Ahmet Alp Balkan wrote:
> Change looks good to me generally. Left a few comments.
> 
> @rjschwei does suse use  `waagent.conf` config keys:
> 	• OS.CheckRdmaDriver 
> 	• OS.UpdateRdmaDriver
> 	• OS.RdmaRepository
> 	• OS.PackageExtraConfig 
> 

No, these are unnecessary. I was planning on introducing a new variable

OS.EnableRDMA

We can skip the variable and simply look at /var/lib/hyperv/.kvp_pool_0
and do everything automatically, but I am leaning towards having the
config variable to enable the mechanism.


-- 
Robert Schweikert                   MAY THE SOURCE BE WITH YOU
Public Cloud Architect                         LINUX
rjschwei@suse.com
IRC: robjo

azurelinuxagent/distro/default/rdma.py
rjschwei: On 06/16/2016 07:11 PM, Ahmet Alp Balkan wrote:
>>  
>>  
>>  class RDMAHandler(object):
>> +    def __init__(self):
>> +        self.driver_module_name = 'hv_network_direct'
> 
> can we declare this just under `class RDMAHander`

Yes, it's a style choice. I generally prefer using __init__ but if you'd
rather me drop __init__ I certainly have no problem doint it the other
way. From a language perspective I am not certain if there is actually a
difference.

rjschwei: OK after refreshing my memory a bit I agree that a class member is a nicer choice for the driver_module name

azurelinuxagent/distro/suse/rdma.py
rjschwei: Trailing %s in command set up makes sense. Having played around a bit with refactoring the whole method and ending up in a tangled mess I see no advantage to have the commands as class or instance members, thus I think we should leave them as they are with the added "%s"

rjschwei:
Agreed that install_driver() is getting a bit lengthy. I did fiddle around with it to see if I could find clean lines of refactoring the code but I ended up in a tangled mess that I did not like at all. That is not to say that refactoring along nice lines is not possible just that I have not found a nice solution to the problem yet. THus I'd prefer to keep the method in place as is.

rjschwei:
Testing changes....

rjschwei:
OK, changes ready for review.

I needed to make a change to the "package upgrade" path. It turns out that this particular module will not remove with the "rmmod" command. Thus is the module is loaded into kernel space but an upgrade is needed we have no choice but to reboot the system.

Implemented the reboot condition. The question is, is it safe to do this at this point, or should this be handled at the end of the agent workflow?

ahmetalpbalkan:
> Implemented the reboot condition. The question is, is it safe to do this at this point, or should this be handled at the end of the agent workflow?

Hi @rjschwei, any ideas why the old code in 2.0 branch have did not have any `reboot`s?

azurelinuxagent/distro/suse/rdma.py
```diff
@@ -34,11 +34,14 @@ def install_driver(self):
             error_msg += 'no driver will be installed.'
             logger.error(error_msg)
             return
-        zypper_install = 'zypper -n in'
-        zypper_search = 'zypper se -s'
+        zypper_install = 'zypper -n in %s'
+        zypper_remove = 'zypper -n rm %s'
+        zypper_search = 'zypper se -s %s'
```
ahmetalpbalkan: In the old code we had:

     self.command_to_get_rdma_package_info = self.rpm_path + " -q " + self.rdma_package_name

which would just return the package name like `microsoft-hyper-v-rdma-4.1.0.142-20160323.x86_64`. Isn't this a better fit to determine the installed package?

ahmetalpbalkan:
> No, these are unnecessary. I was planning on introducing a new variable
> 
> OS.EnableRDMA
> 
> We can skip the variable and simply look at /var/lib/hyperv/.kvp_pool_0
> and do everything automatically, but I am leaning towards having the
> config variable to enable the mechanism.

@rjschwei regarding these variables I mentioned earlier, do we know that customers were not using them in their custom images at all? Because it had pre-post package install script hooks, extra repostories and packages to install. 

If we don't provide the same keys, it might mean regression for the new users.

Here is the logic I extracted from 2.0:

```
read config keys from waagent.conf
	• OS.CheckRdmaDriver 
	• OS.UpdateRdmaDriver
	• OS.RdmaRepository
	• if OS.PackageExtraConfig: read from JSON:
			§ PreTask (file path)
			§ PostTask (file path)
			§ ExtraRepositories (array)
			§ Packages (array)

if $OS.RdmaRepository:
	distro.addPackageRepository(...)

exec $PreTask  [continue on failure]

if $OS.UpdateRdmaDriver:
	distro.rdmaUpdate() [continue on failure]
	
foreach $ExtraRepositories:
	distro.addPackageRepository(...)

foreach $Packages:
	distro.refreshRepository()
	distro.installPackage(...)  [continue on failure]

exec $PostTask  [continue on failure]

if $OS.CheckRdmaDriver:
	res = distro.checkRDMA()
        print res
```

rjschwei:
No idea why the 2.0 code did not have the reboot, in the case where a VM would have been relaunched  on HW with a different FW implementation the 2.0 branch would have left the user in a state where IB would not have been usable as the wrong driver would have been loaded in the kernel.

azurelinuxagent/distro/suse/rdma.py
rjschwei: Well it is not necessarily about finding what is installed, it is also about finding what is available. The "zypper se -s" call gives us both. We have to cover 3 conditions:

1.) No driver installed => install the proper driver version from the available packages, thus I have to find all the avialbale packages
2.) Driver installed and matching version => do nothing (this case could be covered by the rpm call from the 2.0 branch
3.) Driver installed but the FW version does not match => uninstall the driver and find the proper version from all the available packages

All cases are covered by the information zypper returns. Only 1 case is covered by the information the rpm call returns.

rjschwei:
So the repository is already built into the image and if it doe not work because the system sits in a VNET without Internet access then we have the local cache of packages that is alos built into the image. This makes OS.RdmaRepository obsolete. 

With the new logic the driver automatically gets updated when needed, this makes OS.UpdateRdmaDriver obsolete.

The use of OS.CheckRdmaDriver is replaced by the new setting OS.EnableRDMA with the new setting being more explicit in name as it indicates that the RDMA driver will actually be enabled. The previous name only indicates that the driver would be checked.

OS.PackageExtraConfig is IMHO not related to RDMA, if it is then from my perspective it is intermingling of different concepts. I can see that there may be value in having OS.PackageExtraConfig, but I would argue that any functionality hidden behind this should be executed no matter what and that the switch is superfluous.

ahmetalpbalkan:
@rjschwei sounds good. I realized CentOS did not really use these flags either. I am assuming they're added there for users to extend. Let's leave them out for now and see what happens.

azurelinuxagent/distro/suse/rdma.py
```diff
@@ -48,22 +51,42 @@ def install_driver(self):
...
+                result = shellutil.run(cmd)
+                if result:
+                    error_msg = 'Failed install of package "%s" '
+                    error_msg += 'from available repositories.'
+                    logger.error(error_msg % complete_name)
+                msg = 'Successfully installed "%s" from '
+                msg += 'configured repositories'
+                logger.info(msg % complete_name)
+                self.load_driver_module()
+                if requires_reboot:
```
ahmetalpbalkan: can we please add some logging around the reboot. it is quite a critical operation for us?

azurelinuxagent/distro/suse/rdma.py
ahmetalpbalkan: @rjschwei sounds good.

ahmetalpbalkan:
I guess most of the `install_driver()` logic is going to be the same for CentOS. including checking for FW version, matching it against installed (if any), upgrading it etc. The installing from `/opt/microsoft/rdma` is going to be the same as well. Only the actual commands and the way we parse versions etc will differ.

I guess we can generalize that method more later on.

ahmetalpbalkan:
LGTM

PR #240, opened 2016-06-17
Workaround for race condition between azurelinuxagent and scvmmagent

emsi:
This fixes a race condition between azurelinuxagent  and scvmmagent when systemd is used.
The problem is due to the fact that systemd restarts walinuxagent once it finishes its job. The sleep delays the restart and leaves scvmmagent  enough time for finishing its job.

### DIFF
--- a/azurelinuxagent/daemon/scvmm.py
+++ b/azurelinuxagent/daemon/scvmm.py
@@ -21,6 +21,7 @@ import re
 import os
 import sys
 import subprocess
+import time
 import azurelinuxagent.common.logger as logger
 import azurelinuxagent.common.conf as conf
 from azurelinuxagent.common.osutil import get_osutil
@@ -70,4 +71,5 @@ class ScvmmHandler(object):
         if self.detect_scvmm_env():
             self.start_scvmm_agent()
             logger.info("Exiting")
+            time.sleep(300)
             sys.exit(0)


emsi:
The alternative fix could be to modify systemd `walinuxagent.service` file and set the service type to `oneshot`:

```
[Unit]
Description=Azure Linux Agent

After=network-online.target cloud-final.service
Wants=network-online.target sshd.service sshd-keygen.service cloud-final.service

[Service]
Type=oneshot
ExecStart=/usr/bin/python3 /usr/sbin/waagent -daemon
Restart=always

[Install]
WantedBy=multi-user.target
```

I'm not sure though it that wouldn't break some other functionality on Azure ad waagent wouldn't be restarted on exit.

ahmetalpbalkan:
@emsi I was not familiar with this code path. Does svmm have its own systemd unit? If we set `Restart=on-failure` instead of `always`, would that do the job?

azurelinuxagent/daemon/scvmm.py
```diff
@@ -70,4 +71,5 @@ def run(self):
         if self.detect_scvmm_env():
             self.start_scvmm_agent()
```
jasonzio: Why are you calling self.start_scvmm_agent() a second time? If detect_scvmm_env returned true, then that function already called start_scvmm_agent (back on line 53).

hglkrijger: Good catch @jasonzio - this was a bad merge on my part, will fix with #241

emsi: @ahmetalpbalkan scvmm does not (currently) have systemd unit. It definitely should.

emsi: Setting `Restart=on-failure` is not working. Most probably scvmmagent is not cleanly exiting and waagent loops.

PR #244, opened 2016-06-17
Agent Self-Update Feature

brendandixon:
The Agent will monitor for more recent versions and update itself
(without impacting the installed version). If the new Agent fails
(in a very bad way or consistently), the code rolls back to the prior
version.


azurelinuxagent/ga/update.py
```diff
@@ -16,266 +16,399 @@
...
+            agent_name = latest_agent.name()
 
         devnull = open(os.devnull, 'w')
         try:
             child = subprocess.Popen([agent_bin, 'run-exthandlers'], 
                                       stdout=devnull, stderr=devnull)
+            ret = child.wait()
+            if ret == None:
+                ret = 1
+            if ret != 0:
```
hglkrijger: it looks like the scvmm logic should work fine with this; any concerns there?

tests/ga/test_update.py
```diff
@@ -15,122 +15,711 @@
...
+        return
+
+    def test_ensure_latest_agent_skips_if_too_frequent(self):
+        conf.get_autoupdate_frequency = Mock(return_value=10000)
+        self.update_handler.last_attempt_time = time.time()
+        self.assertFalse(self._test_ensure_latest_agent())
+        return
+
+    def test_ensure_latest_agent_skips_when_etag_matches(self):
+        self.update_handler.last_etag = 42
```
hglkrijger: :+1:

hglkrijger:
looks good to me

azurelinuxagent/ga/update.py
brendandixon: @hglkrijger Good question. I had not considered the impact on scvmm. Could you fill me in what I need to be aware of?

PR #247, opened 2016-06-19
Fix sudo asking for password

rubeon:
On current latest version, Agent fails to properly set sudoers as NOPASSWD when password auth is disabled.

This patch fixes that.


jasonzio:
lgtm

hglkrijger:
hi @rubeon, thanks for this - the fix already exists in the develop branch though, so I will close this PR.

PR #250, opened 2016-06-20
Host plugin channel

hglkrijger:
Separate commits here since one is purely reformatting.

- add a protocol handler for the host plugin channel
- add apis for /versions, /status and /vmAgentLog
- make use of the /status api to report vm status as a fallback
- adds unit tests

Note that the /vmAgentLog is not called or unit tested yet.

### DIFF
--- a/azurelinuxagent/common/protocol/hostplugin.py
+++ b/azurelinuxagent/common/protocol/hostplugin.py
@@ -23,6 +23,7 @@ from azurelinuxagent.common.utils import textutil
 HOST_PLUGIN_PORT = 32526
 URI_FORMAT_GET_API_VERSIONS = "http://{0}:{1}/versions"
 URI_FORMAT_PUT_VM_STATUS = "http://{0}:{1}/status"
+URI_FORMAT_PUT_LOG = "http://{0}:{1}/vmAgentLog"
 API_VERSION = "2015-09-01"
 
 
@@ -79,3 +80,30 @@ class HostPluginProtocol(object):
                 logger.error("put VM status returned status code [{0}]".format(response.status))
         except HttpError as e:
             logger.error("put VM status failed with [{0}]".format(e))
+
+    def put_vm_log(self, content, container_id, deployment_id):
+        """
+        Try to upload the given content to the host plugin
+        :param deployment_id: the deployment id, which is obtained from the goal state (tenant name)
+        :param container_id: the container id, which is obtained from the goal state
+        :param content: the binary content of the zip file to upload
+        :return:
+        """
+        if not self.is_initialized:
+            self.initialize()
+        if not self.is_available:
+            logger.error("host plugin channel is not available")
+            return
+        if content is None or container_id is None or deployment_id is None:
+            logger.error("invalid arguments passed: [{0}], [{1}], [{2}]".format(content, container_id, deployment_id))
+            return
+        url = URI_FORMAT_PUT_LOG.format(self.endpoint, HOST_PLUGIN_PORT)
+
+        headers = {"x-ms-vmagentlog-deploymentid": deployment_id, "x-ms-vmagentlog-containerid": container_id}
+        logger.info("put VM log at [{0}]".format(url))
+        try:
+            response = restutil.http_put(url, content, headers)
+            if response.status != httpclient.OK:
+                logger.error("put log returned status code [{0}]".format(response.status))
+        except HttpError as e:
+            logger.error("put log failed with [{0}]".format(e))
--- a/azurelinuxagent/common/protocol/hostplugin.py
+++ b/azurelinuxagent/common/protocol/hostplugin.py
@@ -36,22 +36,28 @@ class HostPluginProtocol(object):
         self.api_versions = None
         self.endpoint = endpoint
 
-    def initialize(self):
-        self.api_versions = self.get_api_versions()
-        self.is_available = self.api_versions is not None and API_VERSION in self.api_versions
-        self.is_initialized = True
+    def ensure_initialized(self):
+        if not self.is_initialized:
+            self.api_versions = self.get_api_versions()
+            self.is_available = API_VERSION in self.api_versions
+            self.is_initialized = True
+        return self.is_available
 
     def get_api_versions(self):
-        url = URI_FORMAT_GET_API_VERSIONS.format(self.endpoint, HOST_PLUGIN_PORT)
+        url = URI_FORMAT_GET_API_VERSIONS.format(self.endpoint,
+                                                 HOST_PLUGIN_PORT)
         logger.info("getting API versions at [{0}]".format(url))
         try:
             response = restutil.http_get(url)
             if response.status != httpclient.OK:
-                logger.error("get API versions returned status code [{0}]".format(response.status))
-                return None
+                logger.error(
+                    "get API versions returned status code [{0}]".format(
+                        response.status))
+                return []
             return response.read()
         except HttpError as e:
             logger.error("get API versions failed with [{0}]".format(e))
+            return []
 
     def put_vm_status(self, status_blob, sas_url):
         """
@@ -59,9 +65,7 @@ class HostPluginProtocol(object):
         :param sas_url: the blob SAS url to pass to the host plugin
         :type status_blob: StatusBlob
         """
-        if not self.is_initialized:
-            self.initialize()
-        if not self.is_available:
+        if not self.ensure_initialized():
             logger.error("host plugin channel is not available")
             return
         if status_blob is None or status_blob.vm_status is None:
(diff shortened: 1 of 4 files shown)


azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -0,0 +1,109 @@
...
+            logger.error("get API versions failed with [{0}]".format(e))
+
+    def put_vm_status(self, status_blob, sas_url):
+        """
+        Try to upload the VM status via the host plugin /status channel
+        :param sas_url: the blob SAS url to pass to the host plugin
+        :type status_blob: StatusBlob
+        """
+        if not self.is_initialized:
+            self.initialize()
```
brendandixon: @hglkrijger I like this general pattern, but the code is more robust if you push the conditional into the called method. For example, I often write code such as:

```python
def ensure_initialized(self):
      if self.is_initialized:
          return
      self.is_initialized = True
      return

def worker(self):
      self.ensure_initialized()
      return
```

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -0,0 +1,109 @@
...
+            raise ProtocolError("Host plugin endpoint not provided")
+        self.is_initialized = False
+        self.is_available = False
+        self.api_versions = None
+        self.endpoint = endpoint
+
+    def initialize(self):
+        self.api_versions = self.get_api_versions()
+        self.is_available = self.api_versions is not None and API_VERSION in self.api_versions
+        self.is_initialized = True
```
brendandixon: @hglkrijger The check for `is not None` is not necessary unless you expect `self.api_versions` may possibly (and very oddly) contain `None`.

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -0,0 +1,109 @@
...
+    def get_api_versions(self):
+        url = URI_FORMAT_GET_API_VERSIONS.format(self.endpoint, HOST_PLUGIN_PORT)
+        logger.info("getting API versions at [{0}]".format(url))
+        try:
+            response = restutil.http_get(url)
+            if response.status != httpclient.OK:
+                logger.error("get API versions returned status code [{0}]".format(response.status))
+                return None
+            return response.read()
+        except HttpError as e:
```
brendandixon: @hglkrijger Is this an API call we may need to retry (say, if we receive an errant 503 or 500 or the network times out)? I know we've seen such with wireserver.

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -0,0 +1,109 @@
...
+            logger.error("host plugin channel is not available")
+            return
+        if status_blob is None or status_blob.vm_status is None:
+            logger.error("no status data was provided")
+            return
+        url = URI_FORMAT_PUT_VM_STATUS.format(self.endpoint, HOST_PLUGIN_PORT)
+        status = textutil.b64encode(status_blob.vm_status)
+        headers = {"x-ms-version": API_VERSION}
+        blob_headers = [{'headerName': 'x-ms-version', 'headerValue': status_blob.__storage_version__},
+                        {'headerName': 'x-ms-blob-type', 'headerValue': status_blob.type}]
```
brendandixon: @hglkrijger These lines do not follow the PEP 8 standard (at least as I understand it): They should be one value per line if they cannot all fit on one line.

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -0,0 +1,109 @@
...
+        data = json.dumps({'requestUri': sas_url, 'headers': blob_headers, 'content': status}, sort_keys=True)
+        logger.info("put VM status at [{0}]".format(url))
+        try:
+            response = restutil.http_put(url, data, headers)
+            if response.status != httpclient.OK:
+                logger.error("put VM status returned status code [{0}]".format(response.status))
+        except HttpError as e:
```
brendandixon: @hglkrijger Ditto to my comments above on the need to retry. Perhaps you could introduce to this class (and so sad that we cannot, at this time, make it global) an HTTP helper that accepts a cooked request and verb but handles all retries etc.

azurelinuxagent/common/protocol/hostplugin.py
brendandixon: @hglkrijger In looking at this further, I suggest that the proposed `ensure_initialized` return a boolean and that the caller return if the method returns `False`. Then I modify my proposal above to the following:

```python
def ensure_initialized(self):
    if not self.is_initialized:
        self.api_versions = self.get_api_versions()
        self.is_available = API_VERSION in self.api_versions
        self.is_initialized = True
    return self.is_available
```
That way all checks to `self.is_initialized` and `self.is_available` are contained to a single method. Calling code is simplified as well.

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -0,0 +1,109 @@
...
+            return
+        url = URI_FORMAT_PUT_LOG.format(self.endpoint, HOST_PLUGIN_PORT)
+
+        headers = {"x-ms-vmagentlog-deploymentid": deployment_id, "x-ms-vmagentlog-containerid": container_id}
+        logger.info("put VM log at [{0}]".format(url))
+        try:
+            response = restutil.http_put(url, content, headers)
+            if response.status != httpclient.OK:
+                logger.error("put log returned status code [{0}]".format(response.status))
+        except HttpError as e:
```
brendandixon: @hglkrijger Ditto again on the possible need to retry.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -314,45 +321,51 @@ def __init__(self, client):
...
     def upload(self, url):
-        #TODO upload extension only if content has changed
+        # TODO upload extension only if content has changed
         logger.verb("Upload status blob")
-        blob_type = self.get_blob_type(url)
-
-        data = self.to_json()
+        upload_successful = False
+        self.type = self.get_blob_type(url)
+        self.data = self.to_json()
```
brendandixon: @hglkrijger Is there a benefit to making these instance variables vs. method variables?

azurelinuxagent/common/protocol/hostplugin.py
hglkrijger: it would if the response status != OK, then the getter returns None

azurelinuxagent/common/protocol/hostplugin.py
hglkrijger: will update

azurelinuxagent/common/protocol/hostplugin.py
hglkrijger: the restutil method retries 3 times by default, which I thought was enough

azurelinuxagent/common/protocol/hostplugin.py
hglkrijger: will update, for some reason pycharm uses 120

tests/protocol/test_hostplugin.py
```diff
@@ -0,0 +1,68 @@
...
+import azurelinuxagent.common.protocol.restapi as restapi
+
+wireserver_url = "168.63.129.16"
+sas_url = "http://sas_url"
+api_versions = '["2015-09-01"]'
+
+
+class TestHostPlugin(AgentTestCase):
+    def test_fallback(self):
+        with patch.object(wire.HostPluginProtocol, "put_vm_status") as patch_put:
```
brendandixon: @hglkrijger Do we need a test to ensure it's not engaged when (e.g., test_does_not_fallback)?

brendandixon:
@hglkrijger Other than my comments above, it looks fine.

azurelinuxagent/common/protocol/hostplugin.py
brendandixon: @hglkrijger I did not know that. All good then.

azurelinuxagent/common/protocol/hostplugin.py
brendandixon: @hglkrijger Bug then in the getter. Getters should always return legal values, otherwise calling code gets heavily littered with those kinds of checks. Getter should return an empty list in that case.

azurelinuxagent/common/protocol/hostplugin.py
hglkrijger: `restutil.http_request` handles this today

azurelinuxagent/common/protocol/wire.py
hglkrijger: yes, that way the fallback logic in wire.py can simply pass the status_blob instance along to the host plugin instead of type, status and storage version.

azurelinuxagent/common/protocol/hostplugin.py
hglkrijger: will update

tests/protocol/test_hostplugin.py
hglkrijger: good call, will add

PR #252, opened 2016-06-21
Corrected FlexibleVersion comparison of prerelease tags

brendandixon:


hglkrijger:
lgtm

Issue #248, opened 2016-06-19
Incorrect sudoers file generated for deployments with disabled admin-password, requiring password


rubeon:
Hi,

latest image `openlogic:CentOS:7.2:7.2.20160615` creates an issue where sudo asks for a password, even when password is disabled. On image `openlogic:CentOS:7.2:7.2.20160308` this was not the case.

The `sudoers` file for these two deployments is generated differently:

```bash
# diff /etc/sudoers.d/waagent /mnt/centos-72-latest-nopass/etc/sudoers.d/waagent
1c1
< centos ALL = (ALL) NOPASSWD: ALL
---
> centos ALL = (ALL) NOPASSWD
```

I've created a pull request to address this:

https://github.com/Azure/WALinuxAgent/pull/247

szarkos:
Thanks for reporting this.  We're lucky to have such a great community to help watch our backs like this.

I've confirmed the issue and we've pulled the 7.2 image for now.  We will push a new one tomorrow to resolve this issue.  Agent team, can you please review and merge this PR, and I will add the patch to the agent RPM.

szarkos:
It appears this issue was already fixed with this patch: https://github.com/Azure/WALinuxAgent/commit/78d9a9d7b1740f4c68d29888c1600f303c95f68a

This is in v2.1.4 and is already in the 'develop' branch, so that's good.  The 'master' branch snapped to v2.1.3 which lacks the fix.

Thanks,
Steve

szarkos:
New CentOS images v20160620 have been published.  Can you please give it a shot?

This image reverts the Linux Agent to v2.0.18 from the Openlogic repo, but we include the "NM_CONTROLLED=no" in the CentOS 7.2 image which I think was the key piece that you needed.

Thanks,
Steve

rubeon:
Hi, Steve

I've deployed machines with the `7.2.2010620` image, and both issues are resolved.

Thanks for your help!
Eric

PR #254, opened 2016-06-22
Fix NOPASSWD sudoers line in osutil.py by adding : ALL 

ashikv:
Without ": ALL" after NOPASSWD user without password is unable to use sudo like in [issue 248](https://github.com/Azure/WALinuxAgent/issues/248)


ahmetalpbalkan:
@ashikv Sorry for the inconvenience. It appears like this is fixed in the upcoming release branch (`develop`):

https://github.com/Azure/WALinuxAgent/blob/97cdc3e6285e2fe77f0826aaf1cadd7db34b80b5/azurelinuxagent/common/osutil/default.py#L134

PR #255, opened 2016-06-22
Integrate RDMA handling

rjschwei:
Initial proposal to integrate the RDMA handler into the primary code flow.

- Given directory names such as ga and pa I was not certain where to place the handler
- Tried to follow the established factory pattern implementation
- Not yet tested, will test when we agree on the structure


ahmetalpbalkan:
@rjschwei apparently there's a broken test. Other than that it looks fine to me. cc @hglkrijger to comment/merge.

azurelinuxagent/pa/rdma/factory.py
```diff
@@ -0,0 +1,33 @@
...
+#
+
+import azurelinuxagent.common.logger as logger
+
+from azurelinuxagent.common.version import DISTRO_NAME
+from azurelinuxagent.common.rdma import RDMAHandler
+from .suse import SUSERDMAHandler
+
+
+def get_rdma_handler(distro_name=DISTRO_NAME, distro_version=DISTRO_VERSION):
```
hglkrijger: everything which is under common/osutil has already been through the appropriate factory. let's keep a single set of these checks, and move this code under there. I would also suggest we move rdma.py to common/rdma.py rather than pa.

rjschwei: I did move the base class rdma.py into common In pa is the SUSE specialization just like there are OS specializations in provisioning and deprovisioning. Did I miss something or am I misunderstanding the comment?

rjschwei:
OK, tested works on SLES as expected.

One question remains, but could probably be address in a separate pull request. The exception that gets thrown in the RDMAHandler base class from install_driver stops the execution of the agent at this point if RDMA is enabled but there is no distro implementation. The question is whether it is preferred to throw the exception as is in install_driver and catch it in main.py i.e. use:

if conf.enable_rdma():
  try:
            self.rdma_handler.install_driver()
  except:
    pass

Or write an error from install_driver() in the base class and simply return False? With the current implementation a mis-configuration may leave the VM in an undesirable state.

Thoughts?

ahmetalpbalkan:
@rjschwei I believe the preference would be to keep the agent running, this must be the case in the v2.0 implementation as well (the code path for RDMA on unsupported distros simply did not get executed at all).

I think we will also make sure that only the HPC images we publish will have `OS.EnableRdma` config set. Without this key, none of the RDMA code paths should be invoked at all, so we're good.

azurelinuxagent/pa/rdma/factory.py
hglkrijger: I missed the rdma -> common move, thanks for that. The factory in pa is really for the distros that need special provisioning logic. It looks like the driver install is done from the daemon, which would be invoked at run time, not provisioning time, so it seems more appropriate that this be done in osutil?

azurelinuxagent/common/conf.py
```diff
@@ -80,6 +80,9 @@ def load_conf_from_file(conf_file_path, conf=__conf__):
         raise AgentConfigError(("Failed to load conf file:{0}, {1}"
                                 "").format(conf_file_path, err))
 
+def enable_rdma(conf=__conf__):
+    return conf.get_switch("OS.EnableRDMA", False)
```
ahmetalpbalkan: so this is a new key, can you please update the `waagent.conf` and the `README` please.

it also means we are removing support for:

* OS.CheckRdmaDriver 
* OS.UpdateRdmaDriver
* OS.RdmaRepository
* OS.PackageExtraConfig

keys as well. Therefore we should update the waagent.conf and the README ideally.

ahmetalpbalkan: also maybe move this method a bit lower?

rjschwei: - Updated the README.md file to document OS.EnableRDMA
- I could not find any evidence of the existing (2.0 branch) variables thus I did not refer or comment on those
- I placed the method in somewhat alpha order (e before g) am happy to move it anywhere else in the file, just didn't see any intrinsic organization of method placement

ahmetalpbalkan: sounds good thanks.

ahmetalpbalkan:
LGTM on my side. ping @hglkrijger

azurelinuxagent/pa/rdma/factory.py
rjschwei: Hmm, since this is a handler I would not intuitively look in osutil and looking at the implementations of modules in osutil I feel my intuition is confirmed.

If there is a desire to have the various phases separated at the top level then I guess rdma directory should just move out of "pa". (pa == provisioning-agent??) From that point it woudl be nice if these directory named convey some meaning.

My thought is that the RDMA handlers do not belong into osutils and should stand on their own alongside provisioning and potentially other "workflows" within the agent.

hglkrijger:
@rjschwei, I chatted with @ahmetalpbalkan about this offline, we are good to take the PR as is, and then make any additional structural changes which might be needed. Thanks!

PR #259, opened 2016-06-23
Update RDMA IB interface from SharedConfig

ahmetalpbalkan:
This change ports https://github.com/Azure/WALinuxAgent/blob/adfb590ad3904e261da5605b36c3924afa22b2c5/waagent#L4239-L4324 and https://github.com/Azure/WALinuxAgent/blob/adfb590ad3904e261da5605b36c3924afa22b2c5/waagent#L4197-L4203 of v2.0 code to v2.1.

The code flow did not change. In the GoalState parsing, we parse SharedConfig for the first time we and spin up a thread to:

* parse `rdmaMacAddress`/`rdmaIPv4Address` from SharedConfig.xml
* update `dat.conf` with rdma IPv4 addr
* wait for `/dev/hvnd_rdma`
* write IPv4/MAC addrs to `/dev/hvnd_rdma`
* find network interface name for infiniband
* `ifconfig up` && `ifconfig [...] $IPv4Addr/16`

On errors, we throw exceptions but we catch them, so we do not crash the agent (keeping the current behavior).

Mostly manually tested and it seems to be working fine on CentOS.

**The only problem is** it looks like we run the agent twice:
```
  ├─python -u /usr/sbin/waagent -daemon
  │   ├─python /usr/sbin/waagent run-exthandlers
```

and if I add this change to GoalState processing, the code is being executed twice in separate processes. How do we prevent this? (Ideally this should be done in `daemon` and not `exthandlers`)

@rjschwei I encourage you to pick this up and test it on  SUSE. (although the issue above should be fixed, the code is testable.) I also appreciate the code reviews!

Here's how logs should look like (in journalctl, waagent.log will have duplications due to issue above):

```diff


Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.678873 INFO Azure Linux Agent Version:2.1.5.rc0
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.680079 INFO OS: centos 7.1.1503
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.680768 INFO Python: 2.7.5
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.681222 INFO Run daemon
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.681819 ERROR RDMAHandler.install_driver not implemented
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.682168 INFO Activate resource disk
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.688779 INFO Resource disk /dev/sdb1 is already mounted
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.689844 INFO Clean protocol
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.690377 INFO Start env monitor service.
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.692337 INFO Configure routes
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.692666 INFO Gateway:None
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.693016 INFO Routes:None
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.707830 INFO Detect protocol endpoints
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.709284 INFO Clean protocol
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.710388 INFO WireServer endpoint is not found. Rerun dhcp handler
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.711386 INFO test for route to 168.63.129.16
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.714847 INFO route to 168.63.129.16 exists
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.715688 INFO Wire server endpoint:168.63.129.16
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.720836 INFO Fabric preferred wire protocol version:2015-04-05
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.723281 INFO Wire protocol version:2012-11-30
Jun 23 00:08:38 centos python[49540]: 2016/06/23 00:08:38.724916 WARNING Server prefered version:2015-04-05
+Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.850724 INFO RDMA capabilities are enabled in configuration
+Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.852225 INFO Found RDMA details. IPv4=172.16.1.20 MAC=00:15:5D:33:FF:1D
+Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.852406 INFO RDMA: starting device processing in the background.
+Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.852936 INFO Updating DAPL configuration file
Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.853899 INFO Event: name=WALA, op=HeartBeat, message=
+Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.854191 INFO RDMA: trying /etc/dat.conf
+Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.855335 INFO RDMA: DAPL config not found at /etc/dat.conf
+Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.856497 INFO RDMA: trying /etc/rdma/dat.conf
+Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.857421 INFO RDMA: DAPL config is at: /etc/rdma/dat.conf
+Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.858466 INFO RDMA: DAPL configuration is updated
+Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.859412 INFO RDMA: waiting for device /dev/hvnd_rdma
+Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.859845 INFO RDMA: device ready
+Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.860166 INFO RDMA: Updating device with configuration: rdmaMacAddress="00:15:5D:33:FF:1D" rdmaIPv4Address="172.16.1.20"
+Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.864575 INFO RDMA: Updated device with IPv4/MAC addr successfully
+Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.864915 INFO RDMA: will update the network interface with IPv4/MAC
+Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.869903 INFO RDMA: network interface found: eth1
+Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.870169 INFO RDMA: bringing network interface up
+Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.873663 INFO RDMA: configuring IPv4 addr and netmask on interface
+Jun 23 00:08:42 centos python[49540]: 2016/06/23 00:08:42.876738 INFO RDMA: network address and netmask configured on interface
```

### DIFF
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -684,7 +684,7 @@ class WireClient(object):
                     else:
                         logger.info("RDMA capabilities are enabled in configuration")
                         rdma_device_configured = True # do not run again
-                        self.write_ip_on_rdma()
+                        self.setup_rdma_dev()
                 else:
                     logger.info("RDMA capabilities are not enabled, skipping")
 
@@ -884,7 +884,7 @@ class WireClient(object):
             "x-ms-guest-agent-public-x509-cert": cert
         }
 
-    def write_ip_on_rdma(self):
+    def setup_rdma_dev(self):
         logger.verbose("Parsing SharedConfig XML contents for RDMA details")
         xml_doc = parse_doc(self.shared_conf.xml_text)
         if xml_doc is None:
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -24,7 +24,6 @@ from azurelinuxagent.common.future import httpclient, bytebuffer
 from azurelinuxagent.common.utils.textutil import parse_doc, findall, find, findtext, \
     getattrib, gettext, remove_bom, get_bytes_from_pem
 import azurelinuxagent.common.utils.fileutil as fileutil
-from azurelinuxagent.common.rdma import RDMADeviceHandler
 from azurelinuxagent.common.utils.cryptutil import CryptUtil
 from azurelinuxagent.common.protocol.restapi import *
 from azurelinuxagent.common.protocol.hostplugin import HostPluginProtocol
@@ -54,9 +53,6 @@ ENDPOINT_FINE_NAME = "WireServer"
 SHORT_WAITING_INTERVAL = 1  # 1 second
 LONG_WAITING_INTERVAL = 15  # 15 seconds
 
-# rdma_device_configured variable limits RDMA device updating code within the
-# GoalState loop to be executed only once
-rdma_device_configured = False
 
 class UploadError(HttpError):
     pass
@@ -646,8 +642,6 @@ class WireClient(object):
         self.ext_conf = ExtensionsConfig(xml_text)
 
     def update_goal_state(self, forced=False, max_retry=3):
-        global rdma_device_configured
-
         uri = GOAL_STATE_URI.format(self.endpoint)
         xml_text = self.fetch_config(uri, self.get_header())
         goal_state = GoalState(xml_text)
@@ -677,17 +671,6 @@ class WireClient(object):
                 self.update_shared_conf(goal_state)
                 self.update_certs(goal_state)
                 self.update_ext_conf(goal_state)
-
-                if conf.enable_rdma():
-                    if rdma_device_configured:
-                        logger.info("RDMA device already configured, skipping")
-                    else:
-                        logger.info("RDMA capabilities are enabled in configuration")
-                        rdma_device_configured = True # do not run again
-                        self.setup_rdma_dev()
-                else:
-                    logger.info("RDMA capabilities are not enabled, skipping")
-
                 return
             except WireProtocolResourceGone:
                 logger.info("Incarnation is out of date. Update goalstate.")
@@ -884,37 +867,6 @@ class WireClient(object):
             "x-ms-guest-agent-public-x509-cert": cert
         }
 
-    def setup_rdma_dev(self):
-        logger.verbose("Parsing SharedConfig XML contents for RDMA details")
-        xml_doc = parse_doc(self.shared_conf.xml_text)
-        if xml_doc is None:
-            logger.error("Could not parse SharedConfig XML document")
-            return
-        instance_elem = find(xml_doc, "Instance")
-        if not instance_elem:
-            logger.error("Could not find <Instance> in SharedConfig document")
-            return
-
-        rdma_ipv4_addr = getattrib(instance_elem, "rdmaIPv4Address")
-        if not rdma_ipv4_addr:
-            logger.error("Could not find rdmaIPv4Address attribute on Instance element of SharedConfig.xml document")
-            return
-
-        rdma_mac_addr = getattrib(instance_elem, "rdmaMacAddress")
-        if not rdma_mac_addr:
-            logger.error("Could not find rdmaMacAddress attribute on Instance element of SharedConfig.xml document")
-            return
-
-        rdma_mac_addr=self.format_mac_addr(rdma_mac_addr)
-        logger.info("Found RDMA details. IPv4={0} MAC={1}".format(rdma_ipv4_addr, rdma_mac_addr))
-        RDMADeviceHandler(rdma_ipv4_addr, rdma_mac_addr).start()
-
-    def format_mac_addr(self, addr):
-        """
-        Formats a MAC addr like 00155D33FF1D into "00:15:5D:33:FF:1D"
-        """
-        return ':'.join([addr[i:i+2] for i in range(0, len(addr), 2)])
-
 class VersionInfo(object):
     def __init__(self, xml_text):
         """
@@ -1014,16 +966,19 @@ class HostingEnv(object):
 
 class SharedConfig(object):
     """
-    Holds SharedConfig XML data
+    parse role endpoint server and goal state config.
     """
-    xml_text=None
 
     def __init__(self, xml_text):
-        logger.verbose("Loading SharedConfig.xml")
-        self.xml_text=xml_text
+        logger.verbose("Load SharedConfig.xml")
+        self.parse(xml_text)
 
-    def parse(self):
-       pass
+    def parse(self, xml_text):
+        """
+        parse and write configuration to file SharedConfig.xml.
+        """
+        # Not used currently
+        return self
 
 class Certificates(object):
     """
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -24,6 +24,7 @@ from azurelinuxagent.common.future import httpclient, bytebuffer
 from azurelinuxagent.common.utils.textutil import parse_doc, findall, find, findtext, \
     getattrib, gettext, remove_bom, get_bytes_from_pem
 import azurelinuxagent.common.utils.fileutil as fileutil
+from azurelinuxagent.common.rdma import RDMADeviceHandler
 from azurelinuxagent.common.utils.cryptutil import CryptUtil
 from azurelinuxagent.common.protocol.restapi import *
 from azurelinuxagent.common.protocol.hostplugin import HostPluginProtocol
@@ -53,6 +54,9 @@ ENDPOINT_FINE_NAME = "WireServer"
 SHORT_WAITING_INTERVAL = 1  # 1 second
 LONG_WAITING_INTERVAL = 15  # 15 seconds
 
+# rdma_device_configured variable limits RDMA device updating code within the
+# GoalState loop to be executed only once
+rdma_device_configured = False
 
 class UploadError(HttpError):
     pass
@@ -642,6 +646,8 @@ class WireClient(object):
         self.ext_conf = ExtensionsConfig(xml_text)
 
     def update_goal_state(self, forced=False, max_retry=3):
+        global rdma_device_configured
+
         uri = GOAL_STATE_URI.format(self.endpoint)
         xml_text = self.fetch_config(uri, self.get_header())
         goal_state = GoalState(xml_text)
@@ -671,6 +677,17 @@ class WireClient(object):
                 self.update_shared_conf(goal_state)
                 self.update_certs(goal_state)
                 self.update_ext_conf(goal_state)
+
+                if conf.enable_rdma():
+                    if rdma_device_configured:
+                        logger.info("RDMA device already configured, skipping")
+                    else:
+                        logger.info("RDMA capabilities are enabled in configuration")
+                        rdma_device_configured = True # do not run again
+                        self.write_ip_on_rdma()
+                else:
+                    logger.info("RDMA capabilities are not enabled, skipping")
+
                 return
             except WireProtocolResourceGone:
                 logger.info("Incarnation is out of date. Update goalstate.")
@@ -867,6 +884,33 @@ class WireClient(object):
             "x-ms-guest-agent-public-x509-cert": cert
         }
 
+    def write_ip_on_rdma(self):
+        logger.verbose("Parsing SharedConfig XML contents for RDMA details")
+        xml_doc = parse_doc(self.shared_conf.xml_text)
+        if xml_doc is None:
+            logger.error("Could not parse SharedConfig XML document")
+            return
+        instance_elem = find(xml_doc, "Instance")
+        if not instance_elem:
+            logger.error("Could not find <Instance> in SharedConfig document")
+            return
+
+        rdma_ipv4_addr = getattrib(instance_elem, "rdmaIPv4Address")
+        if not rdma_ipv4_addr:
+            logger.error("Could not find rdmaIPv4Address attribute on Instance element of SharedConfig.xml document")
+
+        rdma_mac_addr = getattrib(instance_elem, "rdmaMacAddress")
+        if not rdma_mac_addr:
+            logger.error("Could not find rdmaMacAddress attribute on Instance element of SharedConfig.xml document")
+        rdma_mac_addr=self.format_mac_addr(rdma_mac_addr)
+        logger.info("Found RDMA details. IPv4={0} MAC={1}".format(rdma_ipv4_addr, rdma_mac_addr))
+        RDMADeviceHandler(rdma_ipv4_addr, rdma_mac_addr).start()
+
+    def format_mac_addr(self, addr):
+        """
+        Formats a MAC addr like 00155D33FF1D into "00:15:5D:33:FF:1D"
+        """
+        return ':'.join([addr[i:i+2] for i in range(0, len(addr), 2)])
 
 class VersionInfo(object):
     def __init__(self, xml_text):
@@ -967,20 +1011,16 @@ class HostingEnv(object):
 
 class SharedConfig(object):
     """
-    parse role endpoint server and goal state config.
+    Holds SharedConfig XML data
     """
+    xml_text=None
 
     def __init__(self, xml_text):
-        logger.verbose("Load SharedConfig.xml")
-        self.parse(xml_text)
+        logger.verbose("Loading SharedConfig.xml")
+        self.xml_text=xml_text
 
-    def parse(self, xml_text):
-        """
-        parse and write configuration to file SharedConfig.xml.
-        """
-        # Not used currently
-        return self
-
+    def parse(self):
+       pass
 
 class Certificates(object):
     """
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -898,10 +898,13 @@ class WireClient(object):
         rdma_ipv4_addr = getattrib(instance_elem, "rdmaIPv4Address")
         if not rdma_ipv4_addr:
             logger.error("Could not find rdmaIPv4Address attribute on Instance element of SharedConfig.xml document")
+            return
 
         rdma_mac_addr = getattrib(instance_elem, "rdmaMacAddress")
         if not rdma_mac_addr:
             logger.error("Could not find rdmaMacAddress attribute on Instance element of SharedConfig.xml document")
+            return
+
         rdma_mac_addr=self.format_mac_addr(rdma_mac_addr)
         logger.info("Found RDMA details. IPv4={0} MAC={1}".format(rdma_ipv4_addr, rdma_mac_addr))
         RDMADeviceHandler(rdma_ipv4_addr, rdma_mac_addr).start()
--- a/azurelinuxagent/common/rdma.py
+++ b/azurelinuxagent/common/rdma.py
@@ -186,7 +186,7 @@ class RDMADeviceHandler(object):
 
     @staticmethod
     def wait_rdma_device(path, timeout_sec, check_interval_sec):
-        logger.info("RDMA: waiting for device={0} timeout={0}s".format(path, timeout_sec))
+        logger.info("RDMA: waiting for device={0} timeout={1}s".format(path, timeout_sec))
         total_retries = timeout_sec/check_interval_sec
         n = 0
         while n < total_retries:
--- a/azurelinuxagent/common/rdma.py
+++ b/azurelinuxagent/common/rdma.py
@@ -234,4 +234,3 @@ class RDMADeviceHandler(object):
         if eths is None or len(eths) == 0:
             raise Exception("ifname with mac: {0} not found".format(mac))
         return eths[-1]
->>>>>>> Update RDMA IB interface from SharedConfig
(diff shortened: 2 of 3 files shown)


rjschwei:
Tested.

The code does configure the additional network:

~> ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default 
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 00:0d:3a:10:01:3b brd ff:ff:ff:ff:ff:ff
    inet 10.2.0.5/24 brd 10.2.0.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::20d:3aff:fe10:13b/64 scope link 
       valid_lft forever preferred_lft forever
3: eth1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 00:15:5d:33:ff:35 brd ff:ff:ff:ff:ff:ff
    inet 172.16.1.44/16 brd 172.16.255.255 scope global eth1
       valid_lft forever preferred_lft forever
    inet6 fe80::215:5dff:fe33:ff35/64 scope link 
       valid_lft forever preferred_lft forever

The driver is loaded as expected:

~> lsmod | grep hv
hv_network_direct     109186  0 
ib_core                96404  11 rdma_cm,ib_cm,ib_sa,iw_cm,ib_mad,ib_ucm,ib_umad,hv_network_direct,ib_uverbs,rdma_ucm,ib_ipoib
ib_addr                18923  4 rdma_cm,ib_core,hv_network_direct,rdma_ucm
hv_netvsc              39767  0 
hv_utils               28477  0 
hv_storvsc             22171  2 
hv_vmbus              544328  7 hyperv_keyboard,hv_netvsc,hid_hyperv,hv_utils,hyperv_fb,hv_storvsc,hv_network_direct
scsi_mod              244624  5 sg,libata,sd_mod,sr_mod,hv_storvsc

Thus things appear to work as expected.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -867,6 +884,33 @@ def get_header_for_cert(self):
...
+            logger.error("Could not find rdmaMacAddress attribute on Instance element of SharedConfig.xml document")
+        rdma_mac_addr=self.format_mac_addr(rdma_mac_addr)
+        logger.info("Found RDMA details. IPv4={0} MAC={1}".format(rdma_ipv4_addr, rdma_mac_addr))
+        RDMADeviceHandler(rdma_ipv4_addr, rdma_mac_addr).start()
+
+    def format_mac_addr(self, addr):
+        """
+        Formats a MAC addr like 00155D33FF1D into "00:15:5D:33:FF:1D"
+        """
+        return ':'.join([addr[i:i+2] for i in range(0, len(addr), 2)])
```
rjschwei: Hmm I am wondering why all the rdma handling would be don ein a class named "UploadError". The class name would imply that this class would handle errors when something fails to upload.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -867,6 +884,33 @@ def get_header_for_cert(self):
             "x-ms-guest-agent-public-x509-cert": cert
         }
 
+    def write_ip_on_rdma(self):
```
rjschwei: Given that this code does not actually write any data to the device, I think the methos is misnamed. setup_rdma_dev() may be more appropriate

azurelinuxagent/common/protocol/wire.py
ahmetalpbalkan: @rjschwei oh no that must be a github UI illusion. this should be the GoalState class. This is identical to what 2.0 branch used to do

azurelinuxagent/common/rdma.py
```diff
@@ -90,3 +96,133 @@ def reboot_system(self):
...
+    def __init__(self, ipv4_addr, mac_addr):
+        self.ipv4_addr = ipv4_addr
+        self.mac_addr = mac_addr
+
+    def start(self):
+        """
+        Start a thread in the background to process the RDMA tasks and returns.
+        """
+        logger.info("RDMA: starting device processing in the background.")
+        threading.Thread(target=self.process).start()
```
rjschwei: Running this in parallel feels a bit like premature optimization.

azurelinuxagent/common/protocol/wire.py
ahmetalpbalkan: you're right

azurelinuxagent/common/rdma.py
ahmetalpbalkan: @rjschwei this is the same behavior as v2.0, if we don't do background, this thread may block up to 2 minutes (so for instance, your PRs need to be executed to set up the RDMA drivers, so the device comes up and I can write the IP to there)... so I think we'll keep it.

azurelinuxagent/common/rdma.py
```diff
@@ -90,3 +96,133 @@ def reboot_system(self):
...
+        logger.info("RDMA: waiting for device {0}".format(path))
+        total_retries = timeout_sec/check_interval_sec
+        n = 0
+        while n < total_retries:
+            if os.path.exists(path):
+                logger.info("RDMA: device ready")
+                return
+            logger.info(
+                "RDMA: device not ready, sleep {0}s".format(check_interval_sec))
+            time.sleep(check_interval_sec)
```
rjschwei: Adding potentialy 120 messages to the log is not necessarily useful, maybe only every 10th or 20th message should actually be written.

azurelinuxagent/common/rdma.py
```diff
@@ -90,3 +96,133 @@ def reboot_system(self):
...
+
+    def start(self):
+        """
+        Start a thread in the background to process the RDMA tasks and returns.
+        """
+        logger.info("RDMA: starting device processing in the background.")
+        threading.Thread(target=self.process).start()
+
+    def process(self):
+        RDMADeviceHandler.update_dat_conf(dapl_config_paths, self.ipv4_addr)
```
rjschwei: We should probably not write the dat configuration prior to knowing that the device will show up, thus I think update_dat_conf should be called after wait_rdma_device

azurelinuxagent/common/rdma.py
ahmetalpbalkan: Heh, I was actually reducing interval to 3s. The 120s apparently is very very conservative and it should probably pass right away (what our folks told me) –and a lil' more logging is probably never bad? (I'm a logging fan) 

Perhaps we can switch this to verbose. But I wanna test this with real world image first and see how it goes. It probably is not going to print 120, but when it does, we know things are bad and we'll be thankful(?)

azurelinuxagent/common/rdma.py
```diff
@@ -90,3 +96,133 @@ def reboot_system(self):
...
+        RDMADeviceHandler.wait_rdma_device(
+            self.rdma_dev, self.device_check_timeout_sec, self.device_check_interval_sec)
+        RDMADeviceHandler.write_rdma_config_on_device(
```
rjschwei: Should be "_to_" instead of "_on_". In general language we refer to writing to a device. Or it could also be set_network_dev_config()

azurelinuxagent/common/rdma.py
ahmetalpbalkan: @rjschwei I just ported the logic from v2.0 codebase and it was like this. Let me ask our team as well.

azurelinuxagent/common/rdma.py
```diff
@@ -90,3 +96,133 @@ def reboot_system(self):
...
+            ipv4_addr)
+        return re.sub(old, new, cfg)
+
+    @staticmethod
+    def write_rdma_config_on_device(path, ipv4_addr, mac_addr):
+        data = RDMADeviceHandler.get_rdma_config(ipv4_addr, mac_addr)
+        logger.info(
+            "RDMA: Updating device with configuration: {0}".format(data))
+        with open(path, "w") as c:
+            c.write(data)
```
rjschwei: "c" is the device, why not name it "dev" or "rdma_dev" more explicit names are nice even in short context spans.

ahmetalpbalkan: I'll probably just rename it to `f` (more conventional in this file block?) I just copypasted some parts so didn't notice the naming heh

azurelinuxagent/common/rdma.py
```diff
@@ -90,3 +96,133 @@ def reboot_system(self):
...
+        logger.info("RDMA: Updated device with IPv4/MAC addr successfully")
+
+    @staticmethod
+    def get_rdma_config(ipv4_addr, mac_addr):
```
rjschwei: I would say this is more of a "generate" method then a "get" method. The method does not get data from somewhere, it generates the data given into the proper format. I think "get_rdma_device_config" would be a more descriptive name. Also as the parameter name ip_addr is probably a more forward looking choice. It is only a matter of time before IPv6 will be supported and then the parameter has to be renamed and if it is forgotten it will be confusing.

azurelinuxagent/common/rdma.py
```diff
@@ -90,3 +96,133 @@ def reboot_system(self):
...
+
+        if_name=RDMADeviceHandler.get_interface_by_mac(mac_addr)
+        logger.info("RDMA: network interface found: {0}", if_name)
+        logger.info("RDMA: bringing network interface up")
+        if shellutil.run("ifconfig {0} up".format(if_name)) != 0:
+            raise Exception("Could not bring up RMDA interface: {0}".format(if_name))
+
+        logger.info("RDMA: configuring IPv4 addr and netmask on interface")
+        addr = '{0}/{1}'.format(ipv4_addr, netmask)
+        if shellutil.run("ifconfig {0} {1}".format(if_name, addr)) != 0:
```
rjschwei: I am wondering if we need to worry about older distributions. Presumably the kernel module only works on reasonably new distributions that also support the "ip" commands. Given that "ifconfig" is considered somewhat antiquated it might be nice to swich over to us "ip" instead of "ifconfig"

azurelinuxagent/common/protocol/wire.py
rjschwei: OK, I expanded the code in the diff still looked as if it was part of "UploadError", Lets consider it a GitHub artifact.

azurelinuxagent/common/rdma.py
ahmetalpbalkan: @rjschwei you're right about method naming, I'll fix that.

I think stating ipv4 for now better because we indeed read the IP addr from a field called `rdmaIPv4Address` and passing to a field called `rdmaIPv4Address`. I prefer being precise about the variable name if that's what the type of the data it holds is.

azurelinuxagent/common/rdma.py
ahmetalpbalkan: this is also how it was done in v2.0 codebase. I think the RDMA distros we support have `ifconfig -a`, so I say we keep it. If we ever get a new image that this doesn't work, it will be visibly broken and will be easy to fix.

azurelinuxagent/common/protocol/wire.py
ahmetalpbalkan: hehe gotta expand more (took me like 15 clicks)

azurelinuxagent/common/rdma.py
```diff
@@ -90,3 +96,133 @@ def reboot_system(self):
...
+            raise Exception("Could set addr to {1} on {0}".format(if_name, addr))
+        logger.info("RDMA: network address and netmask configured on interface")
+
+    @staticmethod
+    def get_interface_by_mac(mac):
```
ahmetalpbalkan: I actually tried to do this via `/proc/net` but I failed. If anyone can think of a cleaner way, that'd be better.

azurelinuxagent/common/rdma.py
rjschwei: The effect, at least on me, when I look through logs that have the same message many times in a row is that I start to glance over things and then stuff that follows gets missed. It is exactly in the case where there is a problem that we do not want to trigger that behavior in the log reviewer. If there is a problem we'd potentially log 120 messages of "RDMA: device not ready...." If we instead would log only 3 or 4 but include the wait interval then we get the same effect, i.e. it is obvious that there is a problem, but we do not roll the log viewers eyes into the back of their head.

So something like:
RDMA: device not ready current wait time X, sleep Y

That way X starts a say 20, while Y remains at 1, this indicates that we print every 20th message and that there is something going on.

azurelinuxagent/common/rdma.py
rjschwei: UNderstood, but we can make improvements while porting ;)

ahmetalpbalkan: @rjschwei I hear from my team that

>It doesn’t matter which step you do first.
>
>If we know this is running in Azure HPC, all those need to be updated to make the whole stack work.

I don't know what this file does to be honest. If you are sure your suggestion is a better way, we can implement it.

azurelinuxagent/common/rdma.py
ahmetalpbalkan: I'm switching this to `.verbose` instead.

ahmetalpbalkan:
Pushed some more commits fixing code review comments.

azurelinuxagent/common/rdma.py
rjschwei: On 06/23/2016 01:18 PM, Ahmet Alp Balkan wrote:
>> +    ipv4_addr = None
>> +    mac_adr = None
>> +
>> +    def __init__(self, ipv4_addr, mac_addr):
>> +        self.ipv4_addr = ipv4_addr
>> +        self.mac_addr = mac_addr
>> +
>> +    def start(self):
>> +        """
>> +        Start a thread in the background to process the RDMA tasks and returns.
>> +        """
>> +        logger.info("RDMA: starting device processing in the background.")
>> +        threading.Thread(target=self.process).start()
>> +
>> +    def process(self):
>> +        RDMADeviceHandler.update_dat_conf(dapl_config_paths, self.ipv4_addr)
> 
> @rjschwei I hear from my team that
> 
>> It doesn’t matter which step you do first.
>>
>> If we know this is running in Azure HPC, all those need to be updated to make the whole stack work.
> 
> I don't know what this file does to be honest. If you are sure your suggestion is a better way, we can implement it.
> 

The config files that are being set up configure the OFED driver for the
IB interface.

ahmetalpbalkan:
> The only problem is it looks like we run the agent twice (-daemon and -exthandlers) and if I add this change to GoalState processing, the code is being executed twice in separate processes. How do we prevent this?

@hglkrijger @brendandixon I need help with this problem above and code reviewing this.

brendandixon:
@ahmetalpbalkan The path is as follows:

Parent process:
- Agent#main launches the daemon
- daemon invokes UpdateHandler#run_latest
- UpdateHandler#run_latest spawns the latest agent with "run-exthandlers"

Child process:
- Agent#main launches UpdateHandler#run
- UpdateHandler#run processes goal state

I am confused. Where are you placing the code such that it's executed twice? In the above model, it should occur only once.

ahmetalpbalkan:
@brendandixon code is at `update_goal_state` function, which is supposed to update files on disk with  Goal State retrieved from file server. This method is simply called from both child & parent processes (otherwise my code wouldn't be invoked twice).

This also means that the XML files we write on disk for diagnostics are subject to race condition or corruption as well.

brendandixon:
@ahmetalpbalkan That placement feels incorrect. It's making RDMA handling a side-effect of other work. Why not raise it to a top-level method called just after provisioning completes?

brendandixon:
@ahmetalpbalkan And, no, there is no race condition because of the ordering I described above. While there are multiple processes, their ordering is very clean.

ahmetalpbalkan:
@brendandixon Goal State is retrieved after the provisioning is completed. Where else can I happily put my code where I have access to the `SharedConfig.xml` contents and get invoked from only one process?

ahmetalpbalkan:
@brendandixon and I'm afraid you're wrong, there is indeed a race condition. SharedConfig.xml is written to the disk **by two processes** within 200ms.

2016/06/24 19:46:14.043846 INFO -------------> WRITE TO SHAREDCONFIG.XML pid=57858
2016/06/24 19:46:14.200146 INFO -------------> WRITE TO SHAREDCONFIG.XML pid=57873

just edit `azurelinuxagent/common/protocol/wire.py` and add this:

```diff
    def update_shared_conf(self, goal_state):
+        import os
+        logger.info('-------------> WRITE TO SHAREDCONFIG.XML pid=%d' % os.getpid())
```

azurelinuxagent/common/protocol/wire.py
```diff
@@ -671,6 +677,17 @@ def update_goal_state(self, forced=False, max_retry=3):
...
                 self.update_certs(goal_state)
                 self.update_ext_conf(goal_state)
+
+                if conf.enable_rdma():
+                    if rdma_device_configured:
+                        logger.info("RDMA device already configured, skipping")
+                    else:
+                        logger.info("RDMA capabilities are enabled in configuration")
+                        rdma_device_configured = True # do not run again
+                        self.setup_rdma_dev()
```
hglkrijger: I would flip these two lines, in case something in `setup_rdma_dev` throws

hglkrijger:
apart from the one comment, looks good to me

azurelinuxagent/common/protocol/wire.py
ahmetalpbalkan: @hglkrijger if we get another call to Goal State code path, then we may run into race condition where we have two threads writing to the /dev/hvnd_rdma device.

RDMA flow is unfortunately just fragile and if something fails here it is unlikely that it will pass second time. I'm not sure flipping this will do any good to us.

hglkrijger: fair enough

ahmetalpbalkan:
@brendandixon moved the RDMA device setup logic outside the goal state loop. We now look at the SharedConfig.xml file on the disk (written by the processes earlier) to parse the info and use it to setup the RDMA before the main loop.

brendandixon:
@ahmetalpbalkan What guarantee does the code have that SharedConfig.xml will be written beforehand? Is it a side-effect of earlier provisioning code?

ahmetalpbalkan:
@brendandixon based on my observation, the SharedeConfig.xml gets written by the same (parent) process earlier to the disk. I think it is a side-effect of something (probably not the provisioning?).

brendandixon:
@ahmetalpbalkan I'm concerned that it's written by the monitor.py code...where we just removed the side-effect. If you absolutely have to have a SharedConfig.xml file, then, *maybe*, the code should reside inside the UpdateHandler#run loop (as much as I hate to suggest it...).

ahmetalpbalkan:
@brendandixon hmm I rebased the code on top of your fix, and SharedConfig.xml was still being written. Do you want me to move it back to the GoalState handling?

brendandixon:
@ahmetalpbalkan No. If you've rebased and it's working (on a clean machine without residual files), then let's keep it as is.

ahmetalpbalkan:
@brendandixon I will test again just to make sure.

ahmetalpbalkan:
@brendandixon just tested again to make sure the code includes your arragement and I removed the *.xml files to see that the parent writes the SharedConfig.xml and then this process gets called.

We've fair amount of logs in the RDMA code and I went back and added logs to other pieces of RDMA code I have not authored myself as well. When something goes wrong, it'll be easy to reveal from the logs.

PR #261, opened 2016-06-23
Add more logs to RDMA drivers

ahmetalpbalkan:
Just adding more logs and `RDMA:` prefixes where applicable in case something goes wrong and we need troubleshooting.

@rjschwei I appreciate your reviews on suse/rdma.py

### DIFF
--- a/azurelinuxagent/common/rdma.py
+++ b/azurelinuxagent/common/rdma.py
@@ -92,7 +92,7 @@ class RDMAHandler(object):
     def reboot_system(self):
         """Reboot the system. This is required as the kernel module for
            the rdma driver cannot be unloaded with rmmod"""
-        logger.info('Rebooting system.')
+        logger.info('RDMA: Rebooting system.')
         ret = shellutil.run('shutdown -r now')
         if ret != 0:
-            logger.error('Failed to reboot the system')
+            logger.error('RDMA: Failed to reboot the system')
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/rdma.py
```diff
@@ -82,11 +83,16 @@ def is_driver_loaded(self):
...
+            logger.info('RDMA: module loaded.')
             return True
+        logger.info('RDMA: module not loaded.')
 
     def reboot_system(self):
         """Reboot the system. This is required as the kernel module for
            the rdma driver cannot be unloaded with rmmod"""
-        logger.info('System reboot')
-        shellutil.run('shutdown -r now')
+        logger.info('Rebooting system.')
```
rjschwei: Should probably add the "RDMA:" prefix to this message as well so we know the reboot was initiated from the RDMA code

azurelinuxagent/common/rdma.py
```diff
@@ -82,11 +83,16 @@ def is_driver_loaded(self):
...
+        ret = shellutil.run('shutdown -r now')
+        if ret != 0:
+            logger.error('Failed to reboot the system')
```
rjschwei: Dito "RDMA:" prefix is probably a good idea

rjschwei:
Other than the potentially missing "RDMA:" prefixes in two messages this is IMHO good to go

hglkrijger:
lgtm

PR #262, opened 2016-06-23
rdma: make RDMAHander.__get_rdma_version static

ahmetalpbalkan:
little syntactic sugar updates. cc: @rjschwei

### DIFF
--- a/azurelinuxagent/pa/rdma/suse.py
+++ b/azurelinuxagent/pa/rdma/suse.py
@@ -28,7 +28,7 @@ class SUSERDMAHandler(RDMAHandler):
     def install_driver(self):
         """Install the appropriate driver package for the RDMA firmware"""
 
-        fw_version = self._RDMAHandler__get_rdma_version()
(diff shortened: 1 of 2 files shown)


azurelinuxagent/pa/rdma/suse.py
```diff
@@ -28,7 +28,7 @@ class SUSERDMAHandler(RDMAHandler):
     def install_driver(self):
         """Install the appropriate driver package for the RDMA firmware"""
 
-        fw_version = self._RDMAHandler__get_rdma_version()
+        fw_version = RDMAHandler.get_rdma_version()
```
rjschwei: I know you don't like the name mangling but this kind of defeats the purpose which was that the version info should only be accessed by an instance of this class or it's children. Python has no other way to communicate this other than using __ as prefix which causes the name-mangling. In the end it is somewhat immaterial but it invites sloppy programming by potentially mis-using the interface in other places.

hglkrijger:
lgtm

Issue #238, opened 2016-06-16
Agent Should Not Install Internal Version of Extensions


boumenot:
An internal test version of CustomScriptForLinux was published on 6/16/2016.  The latest version of CustomScriptForLinux is 1.5.2.0 and the test version was 1.6.1337.  The test version was deployed on Ubuntu 16.04 VMs, but was not deployed onto Ubuntu 14.04 VMs.  IOW, the issue appears to be against v2.1 not v2.0.

The guest agent received the following manifest.

```xml
<?xml version="1.0" encoding="utf-8"?>
<PluginVersionManifest xmlns:i="http://www.w3.org/2001/XMLSchema-instance">
  <Plugins>
    <!-- removed all versions previous to 1.5.2.0 -->
    <Plugin>
      <Version>1.5.2.0</Version>
      <Uris>
        <Uri>https://ardfeextsy3prdstr01.blob.core.windows.net/b01058962be54ceca550a390fa5ff064/Microsoft.OSTCExtensions__CustomScriptForLinux__1.5.2.0</Uri>
        <Uri>https://zrdfepirv2sy3prdstr03a.blob.core.windows.net/b01058962be54ceca550a390fa5ff064/Microsoft.OSTCExtensions__CustomScriptForLinux__1.5.2.0</Uri>
        <Uri>https://zrdfepirv2sy3prdstr04.blob.core.windows.net/b01058962be54ceca550a390fa5ff064/Microsoft.OSTCExtensions__CustomScriptForLinux__1.5.2.0</Uri>
      </Uris>
    </Plugin>
  </Plugins>
  <InternalPlugins>
    <Plugin>
      <Version>1.5.0.0</Version>
      <Uris>
        <Uri>https://ardfeextsy3prdstr01.blob.core.windows.net/b01058962be54ceca550a390fa5ff064/Microsoft.OSTCExtensions__CustomScriptForLinux__1.5.0.0</Uri>
        <Uri>https://zrdfepirv2sy3prdstr03a.blob.core.windows.net/b01058962be54ceca550a390fa5ff064/Microsoft.OSTCExtensions__CustomScriptForLinux__1.5.0.0</Uri>
        <Uri>https://zrdfepirv2sy3prdstr04.blob.core.windows.net/b01058962be54ceca550a390fa5ff064/Microsoft.OSTCExtensions__CustomScriptForLinux__1.5.0.0</Uri>
      </Uris>
    </Plugin>
    <Plugin>
[...]

PR #263, opened 2016-06-23
Corrected extension version selection and stdout/stderr redirection.

### LINKED ISSUE #260: [2.1] journald logs incomplete
```
  ├─python -u /usr/sbin/waagent -daemon
  │   ├─python /usr/sbin/waagent run-exthandlers
```

Ever since we split the agent into this model, only the logs from the parent process shows up in the journald logs. However /var/log/waagent.log is complete. It might be a bad exec where we perhaps don't pass the stdout/stderr pipe correctly.

OR maybe it's just a missing `-u` again...

brendandixon:
Corrects extension version selection (issue #238).
Corrects handling of stdout / stderr (issue #260).

### DIFF
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -80,10 +80,9 @@ class UpdateHandler(object):
             agent_bin = latest_agent.get_agent_bin()
             agent_name = latest_agent.name()
 
-        devnull = open(os.devnull, 'w')
         try:
             child = subprocess.Popen([agent_bin, 'run-exthandlers'], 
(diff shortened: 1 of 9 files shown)


azurelinuxagent/ga/update.py
```diff
@@ -83,7 +83,7 @@ def run_latest(self):
         devnull = open(os.devnull, 'w')
```
ahmetalpbalkan: variable unused

ahmetalpbalkan:
@brendandixon generally LGTM. next time please try to send separate PR for independent issue fixes.

Issue #232, opened 2016-06-15
Shelling out to 'cat file | grep | awk' not needed


nathanleclaire:
In various places the agent calls out to the shell asking to run commands like `cat /proc/cpuinfo | grep ... | awk '{print $3;}'`.  The agent is written in Python, a full scripting language is at its disposal already.  Just read from the file.

The agent should read from these files directly using `os.open` and iterating over the relevant information.  Otherwise the current code will remain extremely fragile.

PR #264, opened 2016-06-23
Remove call to grep

hglkrijger:
Note one of the commits is just pep8 formatting fixes. 

Remove the call to grep and change to a open/read.

### DIFF
--- a/azurelinuxagent/common/dhcp.py
+++ b/azurelinuxagent/common/dhcp.py
@@ -82,9 +82,10 @@ class DhcpHandler(object):
         route_exists = False
         logger.info("test for route to {0}".format(KNOWN_WIRESERVER_IP))
         try:
+            # grep -c returns 1 if the count is 0, so ignore the return code
             route = shellutil.run_get_output(
-                "grep -c {0} /proc/net/route".format(
-                    KNOWN_WIRESERVER_IP_ENTRY))
+                "grep -c {0} /proc/net/route".format(KNOWN_WIRESERVER_IP_ENTRY),
+                chk_err=False)
             # route[0]: (int) return code
             # route[1]: (str) output
             if route is not None and route[0] == 0 and int(route[1]) > 0:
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/dhcp.py
```diff
@@ -80,23 +79,30 @@ def wireserver_route_exists(self):
         This is true when running in a virtual network.
         :return: True if a route to KNOWN_WIRESERVER_IP exists.
         """
-        route_exists=False
+        route_exists = False
         logger.info("test for route to {0}".format(KNOWN_WIRESERVER_IP))
         try:
-            route = shellutil.run_get_output("grep -c {0} /proc/net/route".format(KNOWN_WIRESERVER_IP_ENTRY))
+            # grep -c returns 1 if the count is 0, so ignore the return code
+            route = shellutil.run_get_output(
```
ahmetalpbalkan: `_, route := ...`

jasonzio: Why shell out to grep when python is perfectly capable of opening a file, reading it, and counting the lines that contain a fixed string? Since you actually only care about "0" vs. "greater than 0", doing it yourself within python will be faster than shelling out to grep and reading the entire file.

ahmetalpbalkan: yeah, same feedback was given in #232

hglkrijger: The feedback in #232 seems to be based on resiliency, while that above on performance. Since this check is executed once on a relatively small file, I'm not sure performance is enough of a concern to warrant a change (note that this is not code being introduced with this PR). With regards to resiliency, I agree with the feedback that parsing the content of a file should be done with Python itself, but for a simple check like this I am not sure I agree.

Let me know if you folks feel strongly on this one, and we can chat more.

jasonzio: I feel strongly that scripting languages (like python) that handle grep-like functionality in a three-line function should not be shelling out to grep. There are multiple reasons for this:
- performance
- reduced memory footprint
- reduced attack surface
- semantic transparency ("What does grep -c do?" vs count=CountMatchingLinesInFile(filename))

ahmetalpbalkan:
LGTM

nathanleclaire:
👍

PR #265, opened 2016-06-24
Converted update process to rely on HandlerManifest

brendandixon:


ahmetalpbalkan:
can we create the .zip on the fly? I'm not a big fan of checking in binaries to source tree.

brendandixon:
@ahmetalpbalkan Not for now. It's best if left as it is. Actually checking in a binary is a good thing in this case. Otherwise we'd need to ensure (somehow) the test binary really matches that which we produce for production. By checking in a production binary we know they're the same.

brendandixon:
@ahmetalpbalkan @hglkrijger Can we merge this?

azurelinuxagent/ga/update.py
```diff
@@ -100,6 +120,7 @@ def run_latest(self):
                 latest_agent.mark_failure(is_fatal=True)
 
         sys.exit(0)
+        return
```
hglkrijger: is this return necessary?

hglkrijger:
lgtm

PR #271, opened 2016-06-27
Dhcp caching

hglkrijger:
When we generalize and capture an image, dhcp options exist on disk. When that image is used to create a vm, communication with the cached endpoint will fail, so we need to allow for both cache invalidation and multiple valid but distinct entries in the leases file. 

- fix bug in cache lookup logic - take the the last valid entry, not the first
- allow skipping the cache in failure mode

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -73,7 +73,7 @@ class DefaultOSUtil(object):
         userentry = self.get_userentry(username)
         uidmin = None
         try:
-            uidmin_def = fileutil.get_line_startingwith("UID_MIN", 
+            uidmin_def = fileutil.get_line_startingwith("UID_MIN",
                                                         "/etc/login.defs")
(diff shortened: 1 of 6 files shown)


azurelinuxagent/common/osutil/default.py
```diff
@@ -570,13 +572,21 @@ def get_endpoint_from_leases_path(pathglob):
                                 expire_date = datetime.datetime.strptime(expire_string, FORMAT_DATETIME)
                                 if expire_date > datetime.datetime.utcnow():
                                     expired = False
+                                    logger.info("entry is not expired")
+                                else:
+                                    logger.info("entry is expired")
```
ahmetalpbalkan: I suggest adding some prefix indicating this is in the context of dhcp. info level will totally get surfaced in the logs by default.

hglkrijger: updated the logging to be more clear

Issue #272, opened 2016-06-28
Children process is leaked


ahmetalpbalkan:
(CentOS 6.5 with the develop branch of the agent but I'm not sure if the problem is limited to CentOS)

It appears like when the waagent service is stopped (`service waagent stop`) we're leaking the extension handler child process. Likely a signal forwarding issue.

> Sitenote: it also appears like I can run multiple `-daemon`s manually as well, what happened to our pidfile checking?

Here's pstree output from commands:

### service waagent start

```
  ├─python,30086 /usr/sbin/waagent -daemon
  │   ├─python,30108 /usr/sbin/waagent run-exthandlers
  │   ├─{python},30104
  │   └─{python},30106
```

### service waagent stop

> child process gets parented to init

```
$ pstree -a -p
init,1
  ...
  ├─python,30108 /usr/sbin/waagent run-exthandlers
  ...
```

### service waagent start

two instances of `run-exthandlers`

```
init,1
  ...
  ├─python,30108 /usr/sbin/waagent run-exthandlers
  ├─python,30159 /usr/sbin/waagent -daemon
  │   ├─python,30182 /usr/sbin/waagent run-exthandlers
  │   ├─{python},30178
  │   └─{python},30180
```

### service waagent stop

Another run-exthandlers parented to init.

```
init,1
  ...
  ├─portreserve,1297
  ├─python,30108 /usr/sbin/waagent run-exthandlers
  ├─python,30182 /usr/sbin/waagent run-exthandlers
  ├─rpc.statd,1365
```

### service waagent start

```
init,1
  ...
  ├─python,30108 /usr/sbin/waagent run-exthandlers
  ├─python,30182 /usr/sbin/waagent run-exthandlers
  ├─python,30213 /usr/sbin/waagent -daemon
  │   ├─python,30234 /usr/sbin/waagent run-exthandlers
  │   ├─{python},30230
  │   └─{python},30232
```

### service waagent stop

```
init,1
  ...
  ├─portreserve,1297
  ├─python,30108 /usr/sbin/waagent run-exthandlers
  ├─python,30182 /usr/sbin/waagent run-exthandlers
  ├─python,30234 /usr/sbin/waagent run-exthandlers
  ├─rpc.statd,1365
```

so this can keep going on...

brendandixon:
@ahmetalpbalkan Hmm, DaemonHandler#main still checks for the pidfile. The child process does not overwrite that file (since it does not call DaemonHandler#main).

[...]

PR #281, opened 2016-06-29
Ensure child process is terminated #272

brendandixon:
Ensure the child process is terminated. #272 


hglkrijger:
looks ok to me

ahmetalpbalkan:
I verified the fix works.

PR #288, opened 2016-07-04
Move RDMA pkg installation after provisioning

ahmetalpbalkan:
Move RDMA package installation after provisioning as it _may_ reboot the VM after installing the packages which will cause provisioning CDROM to disappear after the first reboot and hence the provisioning will 100% fail on HPC images.


ahmetalpbalkan:
Merging without CR as this was discussed internally.

PR #292, opened 2016-07-06
Fix the debian packaging

ralt:


ahmetalpbalkan:
@ralt Hello, could you please explain the change?

ralt:
@ahmetalpbalkan without it, `dpkg-buildpackage` fails. The first error is when it tries to copy `debian/tmp/waagent`, since this file doesn't exist. The 2nd one is when it tries to copy the `README` file which doesn't exist. Finally, the resulting package did not have the python library included since `usr` wasn't mentioned in the `install` file.

The `install` file supports a short-hand way to specify the folders to put in the resulting package, and since the build process will put everything in `usr` and `etc`, we can just specify these.

Try it! Run `dpkg-buildpackage` and see what gets copied to the package in `debian/tmp`.

ahmetalpbalkan:
cc: @szarkos

ahmetalpbalkan:
waiving CLA requirement on this one, it's short.

hglkrijger:
lgtm

Issue #290, opened 2016-07-06
2.1 auto update fails to find extracted contents


ahmetalpbalkan:
As promised, trying out the agent self-update functionality... I've just published a dummy package `Microsoft.OSTCLinuxAgent.Test` `2.1.5.1` which is just supposed to create a file on the filesystem.

It is currently available in `Southeast Asia` region. If you add the following to `/etc/waagent.conf`:

```
AutoUpdate.GAFamily=Test
AutoUpdate.Enabled=y
``` 

it should be picking up this package. My package contains: `HandlerManifest.json` and `bin/main.py`. They got extracted right under /var/lib/waagent/:

<details>
<summary>/var/lib/waagent/HandlerManifest.json</summary>
```
[{
    "name": "ExampleHandlerLinux",
    "version": 1.0,
    "handlerManifest": {
        "installCommand": "",
        "uninstallCommand": "",
        "updateCommand": "",
        "enableCommand": "python -u bin/main.py",
        "disableCommand": "",
        "rebootAfterInstall": false,
        "reportHeartbeat": false
    }
}]
```
</details>

<details>
<summary>/var/lib/waagent/bin/main.py</summary>
```
import os
import time

if __name__ == '__main__':
	print('Starting the dummy agent.')
	with open('/var/lib/waagent/dummy-agent', 'w+') as f:
		print('Dummy agent: created the sentinel file')
	print('Dummy agent: sleeping 24h')
	time.sleep(60*60*24)
	print ('Dummy agent: exiting peacefully')
```
</details>

The logs are indicating that they cannot find the HandlerManifest.json file, but it is right there.

My suspicion is there is some issue with extraction. I say this because the package contents everything got extracted right under `/var/lib/waagent`; not `/var/lib/waagent/Microsoft.OSTCLinuxAgent.Test-2.1.5.1/`.

Also, when update fails, the agent crashes and goes into an infinite restart loop. I hope this is also desirable / by-design.

Here are the logs:

```
2016/07/06 01:49:15.949734 INFO Azure Linux Agent Version:2.1.5.rc3
2016/07/06 01:49:15.962067 INFO OS: ubuntu 16.04
2016/07/06 01:49:15.981018 INFO Python: 3.5.1
2016/07/06 01:49:15.983117 INFO Run daemon
2016/07/06 01:49:15.988401 INFO No RDMA handler exists for distro='Ubuntu' version='16.04'
2016/07/06 01:49:16.002901 INFO Clean protocol
[...]

PR #294, opened 2016-07-06
Corrected self-update to preserve errors and improved telemetry (#290)

brendandixon:


ahmetalpbalkan:
Manually verified. Fixes #290 .

ahmetalpbalkan:
LGTM. Kicked off another build as it appears like the previous one got stuck..

PR #296, opened 2016-07-07
Improved telemetry and enabled re-spawning of child process

brendandixon:

### DIFF
--- a/azurelinuxagent/common/event.py
+++ b/azurelinuxagent/common/event.py
@@ -71,11 +71,11 @@ class EventLogger(object):
         except IOError as e:
             raise EventError("Failed to write events to file:{0}", e)
 
-    def add_event(self, name, op="", is_success=True, duration=0, version="1.0",
+    def add_event(self, name, op="", is_success=True, duration=0, version=AGENT_VERSION,
                   message="", evt_type="", is_internal=False):
         event = TelemetryEvent(1, "69B669B9-4AF8-4C50-BDC4-6006FA76E975")
         event.parameters.append(TelemetryEventParam('Name', name))
-        event.parameters.append(TelemetryEventParam('Version', version))
+        event.parameters.append(TelemetryEventParam('Version', str(version)))
         event.parameters.append(TelemetryEventParam('IsInternal', is_internal))
         event.parameters.append(TelemetryEventParam('Operation', op))
         event.parameters.append(TelemetryEventParam('OperationSuccess', 
@@ -92,7 +92,7 @@ class EventLogger(object):
 
 __event_logger__ = EventLogger()
 
-def add_event(name, op="", is_success=True, duration=0, version="1.0",
+def add_event(name, op="", is_success=True, duration=0, version=AGENT_VERSION,
               message="", evt_type="", is_internal=False,
               reporter=__event_logger__):
     log = logger.info if is_success else logger.error
@@ -102,7 +102,7 @@ def add_event(name, op="", is_success=True, duration=0, version="1.0",
         logger.warn("Event reporter is not initialized.")
         return
     reporter.add_event(name, op=op, is_success=is_success, duration=duration,
-                       version=version, message=message, evt_type=evt_type,
+                       version=str(version), message=message, evt_type=evt_type,
                        is_internal=is_internal)
 
 def init_event_logger(event_dir, reporter=__event_logger__):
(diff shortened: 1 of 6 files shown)


azurelinuxagent/common/event.py
```diff
@@ -71,11 +71,11 @@ def save_event(self, data):
         except IOError as e:
             raise EventError("Failed to write events to file:{0}", e)
 
-    def add_event(self, name, op="", is_success=True, duration=0, version="1.0",
+    def add_event(self, name, op="", is_success=True, duration=0, version=AGENT_VERSION,
```
ahmetalpbalkan: I think version should be a required argument. In the past we've had issues with people calling add_event without paying close attention. So making the default AGENT_VERSION may actually be causing some extension telemetry to be reported with agent version.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -131,7 +132,7 @@ def run(self):
             msg = u"Exception retrieving extension handlers: {0}".format(
                 ustr(e))
             logger.warn(msg)
-            add_event(name="WALA", is_success=False, message=msg)
+            add_event(AGENT_NAME, version=CURRENT_VERSION, is_success=False, message=msg)
```
ahmetalpbalkan: I guess this will change it from WALA to WALinuxAgent. Not sure if there's any piece of telemetry/alerting hooked up to this string, just reminding.

brendandixon: @ahmetalpbalkan Correct. I made this change after speaking with @boumenot.

azurelinuxagent/common/event.py
brendandixon: @ahmetalpbalkan Agreed. It is too large a change for now. The prior value was "1.0"...an obviously incorrect value.

ahmetalpbalkan:
LGTM

azurelinuxagent/ga/update.py
```diff
@@ -148,26 +135,49 @@ def run_latest(self):
...
+                    is_success=True,
+                    message=msg)
         except Exception as e:
             msg = u"Agent {0} launch failed with command '{1}' failed with exception: {2}".format(
                 agent_name,
                 agent_cmd,
                 ustr(e))
             logger.warn(msg)
-            add_event(u"WALA", is_success=False, message=msg)
+            add_event(
```
boumenot: Consider introducing a more compact, and easier to read helper for this.  You're spending a lot of vertical space unnecessarily.

  log_enable(is_success, msg)

add_event is oddly named. 😜

azurelinuxagent/ga/update.py
```diff
@@ -270,7 +293,7 @@ def _ensure_latest_agent(self):
         if len(manifests) == 0:
             msg = u"Incarnation {0} has no agent family {1} updates".format(etag, family)
             logger.info(msg)
-            add_event(u"WALA", message=msg)
+            add_event(AGENT_NAME, version=CURRENT_VERSION, message=msg)
```
boumenot: e.g. These no-op events could have a name with better semantic meaning.

brendandixon: @boumenot Are you suggesting an "op=xxxx"? Certainly. Any suggestions?

azurelinuxagent/ga/update.py
brendandixon: @boumenot Considered and rejected for now. It should be done across the codebase and not just in this file. That change was too large, so I continued the pattern used elsewhere.

azurelinuxagent/ga/update.py
boumenot: I was not suggesting an op=xxxx, I was more suggesting a method that does *not* take an op to better convey that no op is not associated with the event.  I'm trying to distinguish between logging messages, and logging events.  This code does the latter, and I cannot think of a good way to meaningfully distinguish.

Spitballing... 
Maybe call it log_(info|warn|error)_event(...)

It seems like 286 and 296 would be very chatty especially considering they are events.  How frequently is this code evaluated?  Is there any value in having them as events?  Is there value is creating an event for INFO messages, my gut says probably not otherwise they wouldn't be INFO.

brendandixon: @boumenot I favor spewing as much data as we possibly can and then filtering at reporting. Again, for now, I've adopted the existing pattern. Let's work with this and change we get better data. Ok?

boumenot: Of course.

PR #300, opened 2016-07-07
Added package building script

brendandixon:


setup.py
```diff
@@ -168,15 +168,14 @@ def run(self):
...
-setuptools.setup(name=AGENT_NAME,
-                 version=AGENT_VERSION,
-                 long_description=AGENT_DESCRIPTION,
-                 author= 'Yue Zhang, Stephen Zarkos, Eric Gable',
-                 author_email = 'walinuxagent@microsoft.com',
-                 platforms = 'Linux',
-                 url='https://github.com/Azure/WALinuxAgent',
-                 license = 'Apache License Version 2.0',
-                 packages=find_packages(exclude=["tests"]),
-                 cmdclass = {
```
hglkrijger: is removing cmdclass intentional?

brendandixon: @hglkrijger Ouch! Good catch. No, that was not. It was a leftover from early fiddling.

makepkg.py
```diff
@@ -0,0 +1,75 @@
...
+        "uninstallCommand": "",
+        "updateCommand": "",
+        "enableCommand": "{1}",
+        "disableCommand": "",
+        "rebootAfterInstall": False,
+        "reportHeartbeat": False
+    }}
+}}'''
+
+ENABLE_CMD = "python -c 'from azurelinuxagent import agent; agent.main([\"-run-exthandlers\"]);' {0}"
```
ahmetalpbalkan: hmm doesn't something like `python foo.egg -run-exthandlers` work?

brendandixon: @ahmetalpbalkan Sadly nope. I've tried. Unless you can find a trick...

brendandixon: @ahmetalpbalkan It complains that "__main__.py" could not be found even though it is present. Any ideas?

ahmetalpbalkan: @brendandixon I'm just not sure if we're missing any preliminary setup (like logging to certain files or some other stuff) by invoking `agent.main` directly instead of having the args parser run and pick up `-run-exthandlers` etc.

If you're confident that these two methods of launching it has no difference, this works for me (worst case, we can always auto-update) otherwise we should probably try to make that syntax work... It'd be cleaner.

PR #303, opened 2016-07-08
Explicitly specify enable timeout as 5m

hglkrijger:
- set timeout and correct condition check
- replaces #235

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -667,15 +667,17 @@ class ExtHandlerInstance(object):
         base_dir = self.get_base_dir()
         try:
             devnull = open(os.devnull, 'w')
-            child = subprocess.Popen(base_dir + "/" + cmd, shell=True,
-                                     cwd=base_dir, stdout=devnull)
+            child = subprocess.Popen(base_dir + "/" + cmd,
+                                     shell=True,
+                                     cwd=base_dir,
+                                     stdout=devnull)
         except Exception as e:
             #TODO do not catch all exception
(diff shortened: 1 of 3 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -674,7 +674,7 @@ def launch_command(self, cmd, timeout=300):
             raise ExtensionError("Failed to launch: {0}, {1}".format(cmd, e))
 
         retry = timeout / 5
-        while retry > 0 and child.poll == None:
+        while retry > 0 and child.poll is None:
```
brendandixon: @hglkrijger I still don't believe this code is correct. `child.poll` is a method. It returns immediately after setting `child.returncode`. If `child.returncode is None` is true, then the child is still active. Otherwise it is terminated.

hglkrijger: You are correct in that this should be a method call, but it returns the returncode, not sets it. There are some issues with the tests as well though, so I will look into that next.

hglkrijger:
So I think I have run this down now. The main change after correcting the method call was to remove the unit test patch for time.sleep, which was skipping the necessary wait period for the unit test to run. The way these tests are written is to actually kick off a script (sample.py) and wait for that process to terminate. If time.sleep is patched, the process is still running when the retries are complete. The resulting effect is for the tests to take much longer (~1.5 minutes) so I reduced the sleep time to 1 second, which should still be reasonable at runtime as well. I think it would be better to fix these tests, but I think that is outside of the scope of this change.

ahmetalpbalkan:
LGTM thanks.

PR #310, opened 2016-07-08
Prevent cycling agent

brendandixon:
This change blacklists an agent the restarts too often in a five minute window.


hglkrijger:
lgtm

PR #315, opened 2016-07-12
Set the umask prior to file creation

edneville:


ahmetalpbalkan:
It appears like this commit changes waagent2.1 _and_ waagent2.0 but submitted to develop branch (where we develop 2.1 and do not change 2.0).

I think we can merge this as we merged #316 and took the fix to `archive/2.0` branch already. Extra fix to `bin/waagent2.0` seems harmless as we ship 2.0 with 2.1 so that VM extensions can use some common functionality that is not related to this change.

@hglkrijger wdyt?

hglkrijger:
lgtm

PR #295, opened 2016-07-06
Set the umask prior to file creation

edneville:

PR #316, opened 2016-07-12
Set the umask prior to file creation

edneville:


ahmetalpbalkan:
previously LGTMed at #295. I will submit PR to fix this on 2.1 (to develop branch)

PR #320, opened 2016-07-13
rdma: kvpd install fix for centos

ahmetalpbalkan:
Added some logic to detect and install kvp daemon if missing, for CentOS.

### DIFF
--- a/azurelinuxagent/common/rdma.py
+++ b/azurelinuxagent/common/rdma.py
@@ -86,8 +86,7 @@ class RDMAHandler(object):
 
         driver_info_source = '/var/lib/hyperv/.kvp_pool_0'
         base_kernel_err_msg = 'Kernel does not provide the necessary '
-        base_kernel_err_msg += 'information or the hv_kvp_daemon is not '
-        base_kernel_err_msg += 'running.'
+        base_kernel_err_msg += 'information or the kvp daemon is not running.'
         if not os.path.isfile(driver_info_source):
             error_msg = 'RDMA: Source file "%s" does not exist. '
             error_msg += base_kernel_err_msg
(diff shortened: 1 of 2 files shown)


ahmetalpbalkan:
So far this has been mostly manually tested. There are some issues with centos65-hpc images but it should not be pertinent to this. Sending upstream so we can test from `develop` branch directly.

ahmetalpbalkan:
can we merge?

hglkrijger:
could you close and reopen? the build seems to be in limbo.

ahmetalpbalkan:
@hglkrijger for that I suggest using `git commit --amend --no-edit; git push -f`. will do this.

azurelinuxagent/common/rdma.py
```diff
@@ -110,6 +109,26 @@ def get_rdma_version():
...
+        if exitcode != 0:
+            raise Exception('RDMA: ps -ef failed: %s' % ps_out)
+        for n in  kvp_daemon_names:
+            if n in ps_out:
+                logger.info('RDMA: kvp daemon (%s) is running' % n)
+                return True
+            else:
+                logger.verbose('RDMA: kvp daemon (%s) is not running' % n)
+        return False
+
```
rjschwei: Tying this to RDMA creates, IMHO, a very tight coupling between RDMA and something that appears to be generic. The kvp daemon doe not only exist to provide RDMA information, I think. It appears to me that this code should live somewhere else. Given that there is work going on for PCI pass through I would expect that the kvp daemon will also be involved there. It feels like this code should live in code located in azurelinuxagent/common/utils/.

Also there is another option for generalization in that the call that gets the process list and error checking should probably live in a method called get_running_processes() such that it can be reused for other things we will be looking for.

ahmetalpbalkan: Fair points, I just ported the code do the same thing from waagent v2.0. It pretty much sounds like in the long term we need to find a better strategy to do RDMA, so I'm trying not to change many things and get back to v2.0 logic as close as possible.

PR #324, opened 2016-07-14
add defaults for self-update to config file

hglkrijger:
- add defaults for self-update to config file

### DIFF
--- a/config/coreos/waagent.conf
+++ b/config/coreos/waagent.conf
@@ -92,3 +92,7 @@ OS.OpensslPath=None
 
 # Enable RDMA management and set up, should only be used in HPC images
 # OS.EnableRDMA=y
+
+# Enable or disable self-update, default is enabled
+AutoUpdate.Enabled=y
+AutoUpdate.GAFamily=Prod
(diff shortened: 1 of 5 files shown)


imduffy15:
Just out of curiousity, what is the rational behind auto update enabled by default? This sounds somewhat terrifying something running a production workload.

config/coreos/waagent.conf
```diff
@@ -92,3 +92,7 @@ OS.OpensslPath=None
 
 # Enable RDMA management and set up, should only be used in HPC images
 # OS.EnableRDMA=y
+
+# Enable or disable self-update, default is enabled
+# AutoUpdate.GAFamily=Prod
```
ahmetalpbalkan: hmm we should probably comment out the config keys if we specify the default value. also, can you please move AutoUpdate.Enabled above the AutoUpdate.GAFamily?

brendandixon:
@imduffy15 Calling the feature self-update is misleading and is historical. What we've actually implemented is treating Azure goal state processing within the agent as itself being an extension. This allows us to keep the key parts of the agent quite stable (e.g. provisioning) while adjusting to changes in the various internal Azure protocols.

Further, to be safe, the feature closely monitors updates to the extension processing code (much better than we've had previously) and falls back to the installed version if errors accumulate.

We expect the feature, and this has been borne out in our testing, to make the agent more stable.

ahmetalpbalkan:
LGTM

PR #325, opened 2016-07-14
fix setup_rdma_device call

ahmetalpbalkan:
I don't know how this ever happened but I started to see logs like 

    ERROR Error setting up rdma device: setup_rdma_device() takes exactly 1 argument (0 given)

turns out I moved this method out of a different file/class recently and forgot to import stuff and remove `self`.

I think we can take this to the 2.1.5 release, as it fixes things and guarded behind `OS.EnableRDM=y` config key.

### DIFF
--- a/azurelinuxagent/common/rdma.py
+++ b/azurelinuxagent/common/rdma.py
@@ -24,9 +24,12 @@ import re
 import time
 import threading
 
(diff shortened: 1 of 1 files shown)


hglkrijger:
lgtm, I agree it should be safe to take

PR #329, opened 2016-07-15
Fix #323

### LINKED ISSUE #323: [2.1.3/2.1.5] CustomScript fails to report status
It appears like when CustomScriptExtension fails on Ubuntu 16.04 (has agent v2.1.3 but this behavior reproes on agent v2.1.5 as well), the status does not get propagated to Azure Portal/API ever:

image

The command to repro the error:

    azure vm extension set <GROUP> <VM> CustomScriptForLinux Microsoft.OSTCExtensions '1.5' --public-config '{}'

this causes an error because public config is missing a required field.

So far what I manually verified is:

- Ubuntu14.04 (agent v2.0.14) works just fine and failure reported just fine to the API.
- Ubuntu16.04 stock image (agent v2.1.3) does not work
- Ubuntu16.04 stock image with updated agent (v2.1.5-develop) does not work
- `.status` files reported on 14.04 is [here][1] and 16.04 is [here][2]. (I can't spot a difference, ordering of fields is probably because I used jq to indent)
- Status gets propagated to the portal/API ([screenshot][3]) if extension succeeds.

[1]: https://gist.github.com/anonymous/d31ba60a99752eb895f8b5a9f0e506e2
[2]: https://gist.github.com/anonymous/21ef5025f1525d7b3635c1f13ceccabc
[3]: https://cl.ly/1k1J0r1n251U/Image%202016-07-13%20at%203.10.48%20PM.png

This suggests that something broken in 2.1.x agents regarding reporting the failure status.

@boumenot have you observed this before?
@hglkrijger do you know if the failure case is covered by the BVT tests?
@brendandixon any ideas?

brendandixon:
Fix #323 

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -453,6 +453,7 @@ class ExtHandlerInstance(object):
         version = self.ext_handler.properties.version
         add_event(name=self.ext_handler.name, version=version, message=message, 
                   op=self.operation, is_success=is_success)
+        self.logger.info(message)
 
     def download(self):
         self.logger.info("Download extension package")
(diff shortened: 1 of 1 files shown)


hglkrijger:
lgtm, will start testing

brendandixon:
@hglkrijger Let's test. But I'm doing more examination as well. We may hold this one.

PR #338, opened 2016-07-22
Updates to makepkg

hglkrijger:
- minor refactoring changes
- add a manifest.xml to the root of the zip; this will be used in extension publishing

### DIFF
--- a/makepkg.py
+++ b/makepkg.py
@@ -25,11 +25,35 @@ MANIFEST = '''[{{
     }}
 }}]'''
 
+PUBLISH_MANIFEST = '''<?xml version="1.0" encoding="utf-8" ?>
+<ExtensionImage xmlns="http://schemas.microsoft.com/windowsazure"  xmlns:i="http://www.w3.org/2001/XMLSchema-instance">
+  <!-- WARNING: Ordering of fields matter in this file. -->
+  <ProviderNameSpace>Microsoft.OSTCLinuxAgent</ProviderNameSpace>
+  <Type>Test</Type>
+  <Version>{0}</Version>
+  <Label>Microsoft Azure Guest Agent for Linux IaaS</Label>
+  <HostingResources>VmRole</HostingResources>
+  <MediaLink></MediaLink>
+  <Description>Microsoft Azure Guest Agent for Linux IaaS</Description>
+  <IsInternalExtension>true</IsInternalExtension>
+  <Eula>https://github.com/Azure/WALinuxAgent/blob/2.1/LICENSE.txt</Eula>
+  <PrivacyUri>https://github.com/Azure/WALinuxAgent/blob/2.1/LICENSE.txt</PrivacyUri>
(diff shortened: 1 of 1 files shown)


hglkrijger:
@boumenot, @brendandixon, please take a look.

makepkg.py
```diff
@@ -25,51 +25,81 @@
...
+  <!-- WARNING: Ordering of fields matter in this file. -->
+  <ProviderNameSpace>Microsoft.OSTCLinuxAgent</ProviderNameSpace>
+  <Type>Test</Type>
+  <Version>{0}</Version>
+  <Label>Microsoft Azure Guest Agent for Linux IaaS</Label>
+  <HostingResources>VmRole</HostingResources>
+  <MediaLink></MediaLink>
+  <Description>Microsoft Azure Guest Agent for Linux IaaS</Description>
+  <IsInternalExtension>true</IsInternalExtension>
+  <Eula>https://github.com/Azure/WALinuxAgent/blob/2.1/LICENSE.txt</Eula>
```
ahmetalpbalkan: should we just make this /master/ ?

makepkg.py
```diff
@@ -25,51 +25,81 @@
...
+  <PrivacyUri>https://github.com/Azure/WALinuxAgent/blob/2.1/LICENSE.txt</PrivacyUri>
+  <HomepageUri>https://github.com/Azure/WALinuxAgent</HomepageUri>
+  <IsJsonExtension>true</IsJsonExtension>
+  <CompanyName>Microsoft</CompanyName>
+  <SupportedOS>Linux</SupportedOS>
+  <Regions>South Central US</Regions>
```
ahmetalpbalkan: maybe parametrize this as well so that global rollout can be done? I'm not sure how they do global rollout for a version when I tried dropping this element or leaving it empty, I got errors.

ahmetalpbalkan:
mostly LGTM

makepkg.py
hglkrijger: there are a number of commits in develop which need to be merged up; until we do that I think I will continue working here. besides, we still need to figure out the whole publishing workflow, and today it uses the develop branch (I think).

makepkg.py
hglkrijger: yes, we probably need to do that eventually, but as a first step I am working on a test publishing build which afaik does no region updates.

makepkg.py
brendandixon: @hglkrijger I believe we should shift focus to master. If work remains on develop, let's merge there and then PR into master.

ahmetalpbalkan: it's not about focus or anything. these links get shown on Azure Portal etc. master is more unlikely to go 404 than 2.1.

hglkrijger: alright, if we are ready to move to master I can do that in a separate PR and then lock the branch. ok?

brendandixon: @hglkrijger Let's move to master. I have a PR coming that is already based on master.

PR #341, opened 2016-07-25
Add supported_dists kwarg to distro detection

nathanleclaire:
cc @ahmetalpbalkan 

Distro detection won't work in some cases without it.  Even though there's a valid `/etc/os-release` file in Alpine the Python code checks to see if the distro is in `supported_dists` kwarg before returning the information.

Signed-off-by: Nathan LeClaire <nathan.leclaire@gmail.com>

### DIFF
--- a/azurelinuxagent/common/version.py
+++ b/azurelinuxagent/common/version.py
@@ -31,7 +31,8 @@ def get_distro():
         release = re.sub('\-.*\Z', '', ustr(platform.release()))
         osinfo = ['freebsd', release, '', 'freebsd']
     elif 'linux_distribution' in dir(platform):
-        osinfo = list(platform.linux_distribution(full_distribution_name=0))
+        osinfo = list(platform.linux_distribution(full_distribution_name=0,
+            supported_dists=platform._supported_dists+('alpine',)))
         full_name = platform.linux_distribution()[0].strip()
         osinfo.append(full_name)
     else:


nathanleclaire:
cc @hglkrijger also

ahmetalpbalkan:
looks good however now every time we want to add support for some exotic distro, now we have to edit this list... I also know that we have forks (such as kali linux or nixos), this change will break them. 

I wish we could get the list and just append `alpine` there but it doesn't seem possible?

nathanleclaire:
> looks good however now every time we want to add support for some exotic distro, now we have to edit this list... I also know that we have forks (such as kali linux or nixos), this change will break them.

You already have this list, it's just not explicitly stated.  From `platform` in Python standard library the signature is:

```
def linux_distribution(distname='', version='', id='',
                       supported_dists=_supported_dists,
                       full_distribution_name=1):
```

and `platform._supported_dists` is:

```
('SuSE', 'debian', 'fedora', 'redhat', 'centos', 'mandrake', 'mandriva', 'rocks', 'slackware', 'yellowdog', 'gentoo', 'UnitedLinux', 'turbolinux')
```

I could append `alpine` (and I guess `coreos`? I'm not sure if / how CoreOS detection is working without that today, maybe slightly different version of `platform.linux_distribution`) to this list if desired.  At any rate though the "default provisoner" will always take over if the distro cannot be detected.

szarkos:
IIRC CoreOS works because of this patch: https://github.com/coreos/coreos-overlay/blob/master/dev-lang/python-oem/files/python-2.7.6-add_os_release_support.patch

Which Alex pulled from http://bugs.python.org/issue17762.

ahmetalpbalkan:
@szarkos wow great input thanks. Yes I just verified it returns `platform` module returns `('CoreOS', '1068.8.0', 'coreos')` on coreos. 

@nathanleclaire I believe appending to  the `platform._supported_dists` on the runtime is going to be a better approach.

nathanleclaire:
The referenced CoreOS patch wasn't merged into upstream so I'm a bit surprised they're still carrying it.  I'm not sure I understand it 100% correctly since it seems to add support for parsing `/etc/os-release` which works fine in my instance but it also somehow seems to short-circuit the `supported_dists` tuple intention (unless they add `coreos` in another patch).  BTW, be advised that [upstream Python is deprecating `platform.linux_distribution()` entirely](http://bugs.python.org/issue1322#msg207427).

> @nathanleclaire I believe appending to the platform._supported_dists on the runtime is going to be a better approach.

Maybe I'll do that in the Alpine OS util then?

ahmetalpbalkan:
@nathanleclaire I'm not sure if we'll be able to discover it is alpine osutil unless we run this step prior to that.

nathanleclaire:
> @nathanleclaire I'm not sure if we'll be able to discover it is alpine osutil unless we run this step prior to that.

Hm, OK.  Where to put the append then?

ahmetalpbalkan:
> Hm, OK. Where to put the append then?

just before we call `platform.linux_distribution()`?

nathanleclaire:
How about we pass the appended list to that call instead? Saves us from modifying some state

ahmetalpbalkan:
@nathanleclaire that would be better, you're right.

nathanleclaire:
Cool I'll implement that

nathanleclaire:
How does this look?

ahmetalpbalkan:
@nathanleclaire sorry for lots of back and forth but why do you need to add coreos? Apparently it'll work without appending it?

nathanleclaire:
@ahmetalpbalkan Oh, I figured it was just good thing to add but I can remove it if desired.  If they ever move off of their non-upstreamed patch and back to regular Python, this issue might well come back to bite them too.

ahmetalpbalkan:
@nathanleclaire I bet that issue will make itself very clear on coreos if they ever move-off of that. So let's just remove and perhaps add a little note?

nathanleclaire:
Sure I can do that

nathanleclaire:
Done.

ahmetalpbalkan:
Travis appears to be stuck. LGTM.

ahmetalpbalkan:
@nathanleclaire can you please`git commit --amend --no-edit && git push -f <...>`, travis is stuck.

nathanleclaire:
Sure

nathanleclaire:
Rebase

nathanleclaire:
@ahmetalpbalkan Mergey mergey?

ahmetalpbalkan:
@hglkrijger LGTY?

nathanleclaire:
Thanks!

PR #344, opened 2016-07-29
rdma: log exception thrown in device processing

ahmetalpbalkan:
Exceptions thrown from rdma processing thread running in
the background were not logged properly.


hglkrijger:
lgtm

PR #345, opened 2016-07-29
updating versioning instructions in maintenance document

hglkrijger:
- updating versioning instructions in maintenance document


brendandixon:
@hglkrijger LGTM

PR #346, opened 2016-08-01
Mount Options feature

### LINKED ISSUE #192: Set Resource Disk mount flag to nodev
Currently `ActivateResourceDisk` mounts the resource disk with no flags. It would be preferrable to have the nodev flag set, either by default or as configured by `/etc/waagent.conf`, since leaving this unset is a bit insecure and triggers vulnerability reports when running credentialed vulnerability scans (eg Nexpose).

I would prefer to not need to manage `/etc/fstab` to work around this.

This is similar to, but not the same as https://github.com/Azure/WALinuxAgent/issues/113

I am willing to write the code if it is likely to be merged.

luthes:
This is intended to close #192. Factored out the mount string to make it easier to test. Tested, both with unittest and manually, the mount string. Was only tested in CentOS 7.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ Provisioning.PasswordCryptSaltLength=10
 ResourceDisk.Format=y
 ResourceDisk.Filesystem=ext4
 ResourceDisk.MountPoint=/mnt/resource
-ResourceDisk.MountOptions=n
+ResourceDisk.MountOptions=None
 ResourceDisk.EnableSwap=n
 ResourceDisk.SwapSizeMB=0
 LBProbeResponder=y
@@ -320,9 +320,10 @@ Type: String Default: /mnt/resource
 This specifies the path at which the resource disk is mounted.
 
 ResourceDisk.MountOptions:
-Type: String Default: n  
+Type: String Default: None  
 
-This specifies disk mount options.
+Specifies disk mount options to be passed to the mount -o command. This is a comma
+separated list of values, ex. 'nodev,nosuid'. See mount(8) for details.
 
 ResourceDisk.EnableSwap:
 Type: Boolean Default: n
--- a/README.md
+++ b/README.md
@@ -183,6 +183,7 @@ Provisioning.PasswordCryptSaltLength=10
 ResourceDisk.Format=y
 ResourceDisk.Filesystem=ext4
 ResourceDisk.MountPoint=/mnt/resource
+ResourceDisk.MountOptions=n
 ResourceDisk.EnableSwap=n
 ResourceDisk.SwapSizeMB=0
 LBProbeResponder=y
@@ -317,6 +318,11 @@ ResourceDisk.MountPoint:
 Type: String Default: /mnt/resource
 
 This specifies the path at which the resource disk is mounted.
+
+ResourceDisk.MountOptions:
+Type: String Default: n  
+
+This specifies disk mount options.
 
 ResourceDisk.EnableSwap:
 Type: Boolean Default: n
--- a/bin/waagent2.0
+++ b/bin/waagent2.0
@@ -126,7 +126,7 @@ Provisioning.MonitorHostName=y          # Monitor host name changes and publish 
 ResourceDisk.Format=y                   # Format if unformatted. If 'n', resource disk will not be mounted.
 ResourceDisk.Filesystem=ext4            # Typically ext3 or ext4. FreeBSD images should use 'ufs2' here.
 ResourceDisk.MountPoint=/mnt/resource   #
-ResourceDisk.MountOptions=None             # Takes a comma seperated list of mount options, check man pages.
+ResourceDisk.MountOptions=None          # Comma-seperated list of options, check man(8) for valid options.
 ResourceDisk.EnableSwap=n               # Create and use swapfile on resource disk.
 ResourceDisk.SwapSizeMB=0               # Size of the swapfile.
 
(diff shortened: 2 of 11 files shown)


hglkrijger:
looks ok to me. @luthes, please sign the CLA. @ahmetalpbalkan, any other feedback?

README.md
```diff
@@ -183,6 +183,7 @@ Provisioning.PasswordCryptSaltLength=10
 ResourceDisk.Format=y
 ResourceDisk.Filesystem=ext4
 ResourceDisk.MountPoint=/mnt/resource
+ResourceDisk.MountOptions=None
```
ahmetalpbalkan: what part of the code does detect the special string `"None"`

nbirnel: azurelinuxagent/daemon/resourcedisk/default.py
line 147 `if mount_options is not None:`

hglkrijger: I think it should be `if mount_options is not None and mount_options != "None"`. You should also have a test case for this, i.e. `"None"` rather than `None`

bin/waagent2.0
```diff
@@ -517,8 +518,9 @@ class AbstractDistro(object):
             fs = Config.get("ResourceDisk.Filesystem")
             if fs == None:
                 fs = "ext3"
-
-            partition = device + "1"
+            mountOpts = Config.get("ResourceDisk.MountOpts")
+            if mountOpts == None:
```
hglkrijger: `if mountOpts is None or mountOpts == "None"`

README.md
ahmetalpbalkan: @hglkrijger is right. I believe in this case you need to check against str value `"None"`.

@nbirnel why don't you leave the value empty as `ResourceDisk.MountOptions=` then you can simply check it as `if not mountOpts:`

luthes:
The way we handle `None` reflects current code conventions,
```
def get_httpproxy_host(conf=__conf__):
    return conf.get("HttpProxy.Host", None)
```
Load handles the magic string `None`.

```
def load(self, content):
    if not content:
        raise AgentConfigError("Can't not parse empty configuration")
    for line in content.split('\n'):
        if not line.startswith("#") and "=" in line:
            parts = line.split()[0].split('=')
            value = parts[1].strip("\" ")
            if value != "None":
                self.values[parts[0]] = value
            else:
                self.values[parts[0]] = None
```

ahmetalpbalkan:
ah I was not aware of that. In that case, generally LGTM.

@luthes I noticed you updated the `bin/waagent2.0` file as well, we no longer touch to that file (unless you're using 2.0.x version of the waagent –in that case we need to accept a change to the archival branches, which we most probably prefer not to). So if you could just revert this file, it would be great.

luthes:
Thanks Ahmet. I've reverted the changes to bin/waagent2.0.

hglkrijger:
@ahmetalpbalkan, we would need to take changes to waagent2.0 in the case where something which references that file (like extensions) would also need to be aware of the change. In this case, I don't think it's relevant though.

@luthes, the only feedback I have is could you please add the new entry in the other conf files as well? Also, were you able to test this out with a generalized image, or does that still need to be done?

ahmetalpbalkan:
I don't see why extensions would need to be aware of mount option key.

nbirnel:
@hglkrijger, I'll ask @luthes to change the other conf files.

I did manual testing on CentOS 7 only. Is that adequate, along with the unit tests, or would you like to see other OS's manually tested? If so, which ones?

luthes:
Config files are fixed. Like @nbirnel said, we haven't done any manual testing on anything outside of Cent 7.

hglkrijger:
@luthes, did you push? I am not seeing the new commit. 

I think the only other testing we would want to do is Ubuntu 1404. If you folks have the bandwidth that would be great, but if not I can take it.

luthes:
Forgot to push from home, fixed. As for Ubuntu testing, @nbirnel will test 14.04, unsure of when though.

nbirnel:
@hglkrijger I attempted to manually test on Ubuntu 14.04, but I hit issues on building the package. See #380 for details. Do you want to test Ubuntu, or give me some guidance on a workaround to build the .deb?

nbirnel:
@hglkrijger I fixed #378, #380, and #382, and have submitted PRs for them. I still don't have a workaround or a fix for #386, and so am unable to test on Ubuntu 14.04.

hglkrijger:
@nbirnel, thanks for taking a look at this. I think we can solve the other packaging issues separately, but testing from source is usually easiest. For 1404 I usually do something like this:
```
git clone https://github.com/<repo>/WALinuxAgent
cd WALinuxAgent
git checkout <fix>
wget https://bootstrap.pypa.io/ez_setup.py -O - | python
python setup.py install
service walinuxagent restart
```
For testing provisioning fixes, we then need to capture this image and generalize, so:
```
waagent -deprovision
```
Then from outside the VM, using the Azure CLI:
```
azure vm deallocate -g <resource group> -n <vm name>
azure vm generalize -g <resource group> -n <vm name>
azure vm capture <resource group> <vm name> -t <template name>
```
Does that help?

nbirnel:
@hglkrijger how would you feel about me passing the manual Ubuntu testing over to you? This is a bit more down the rabbit hole for us than we expected, since we really only intend to use this on CentOS.

hglkrijger:
That's fine. Not sure when I can get to it though, perhaps next week,

hglkrijger:
@ahmetalpbalkan, @brendandixon - I am going to be porting #183 over to master, which means I will be testing this area of the code. Any objections to taking this PR now?

ahmetalpbalkan:
Sounds good.

PR #348, opened 2016-08-05
- Remove eject_dvd

rjschwei:
+ The eject command is only useful for physical HW as the DVD/CDROM drive
    tray can be moved out via software control. In the cloud we have no
    physical drive thus issuing an eject command is superfluous. The DVD
    drive gets unmounted when no longer needed, this is sufficient to remove
    the device from the system. This also removes a packaging dependency for
    Linux systems.

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -321,12 +321,6 @@ class DefaultOSUtil(object):
         if chk_err and retcode != 0:
             raise OSUtilError("Failed to umount dvd.")
 
-    def eject_dvd(self, chk_err=True):
-        dvd = self.get_dvd_device()
-        retcode = shellutil.run("eject {0}".format(dvd))
-        if chk_err and retcode != 0:
-            raise OSUtilError("Failed to eject dvd: ret={0}".format(retcode))
-
     def try_load_atapiix_mod(self):
         try:
             self.load_atapiix_mod()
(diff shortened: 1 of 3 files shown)


hglkrijger:
@rjschwei, the initial set of config we provide to the VM for provisioning comes from the attached drive, so while there is no physical drive, we still make use of this code.

rjschwei:
@hglkrijger unless the Azure framework actually monitors the SCSI bus command that "eject" sends issuing the command is superfluous, all it does is introduce a dependency. Once the drive is unmounted which happens when the code calls umount_dvd() it no longer exists from a kernel perspective.

So the question really is, does the Azure framework monitor the SCSI bus commands? and if so what does it do with the config drive after "eject"?

szarkos:
This is a defense-in-depth feature that ensures that the provisioning ISO does not stick around post-provisioning.

Hyper-V actually does have the concept of ejecting the DVD.  Unmounting the DVD does not automatically eject it, and so it could feasibly be remounted if needed until you eject it either from the Host or by running something like 'eject /dev/sr0' from the guest (fyi, 'eject -t' doesn't work in case you were wondering).

Typically, once the agent signals to the Azure fabric that it has completed provisioning the platform will instruct the Host to disconnect/eject the ISO from the VM.  However, we have seen that ejecting the ISO from the Host side does not always work properly with some Linux kernels.  Thus why this code exists.

We have some fixes on the platform side that will enable us to force the issue for uncooperative kernels.  This also may not be a big issue on SLES since one would need to be root anyway to remount the DVD device.  For the near term, however, I would recommend keeping this code path if possible.

rjschwei:
Thanks for the explanation.

Issue #347, opened 2016-08-04
[2.1.5]waagent fails to create swapfile in RHEL-7 if fstype is ext3


yuxisun1217:
Platform: RHEL-7.3 alpha build
Kernal version: 3.10.0-481.el7.x86_64

In WALA-2.1.5, it uses "fallocate" instead of "dd" to create swapfile. Check the manual, it seems that fallocate is not supported by ext3 filesystem after kernel-2.6.31.
man fallocate
"As of the Linux Kernel v2.6.31, the fallocate system call is supported by the btrfs, ext4, ocfs2, and xfs filesystems."

Steps to Reproduce:
1. Prepare a RHEL7.3 VM on Azure. Set the following lines in the /etc/waagent.conf:
ResourceDisk.Format=y
ResourceDisk.Filesystem=ext3
ResourceDisk.EnableSwap=y
ResourceDisk.SwapSizeMB=1024
2. Reboot the VM.
3. Check the swapfile:
ls -l /mnt/resource|grep swapfile
-rw-r--r--. 1 root root     0 Aug  4 14:17 swapfile
4. #free -m
              total        used        free      shared  buff/cache   available
Mem:           6968         180        6623           8         164        6605
Swap:             0           0           0

Logs:
/var/log/waagent:
2016/08/04 14:17:26.365987 ERROR run cmd 'fallocate -l 1073741824 '/mnt/resource/swapfile'' failed
2016/08/04 14:17:26.379904 ERROR Error Code:1
2016/08/04 14:17:26.389545 ERROR Result:fallocate: /mnt/resource/swapfile: fallocate failed: Operation not supported

2016/08/04 14:17:26.420215 ERROR run cmd 'mkswap /mnt/resource/swapfile' failed
2016/08/04 14:17:26.434877 ERROR Error Code:1
2016/08/04 14:17:26.444640 ERROR Result:mkswap: error: swap area needs to be at least 40 KiB

Usage:
 mkswap [options] device [size]

Options:
 -c, --check               check bad blocks before creating the swap area
 -f, --force               allow swap size area be larger than device
 -p, --pagesize SIZE       specify page size in bytes
 -L, --label LABEL         specify label
 -v, --swapversion NUM     specify swap-space version number
 -U, --uuid UUID           specify the uuid to use
 -V, --version             output version information and exit
 -h, --help                display this help and exit

2016/08/04 14:17:26.524109 ERROR run cmd 'swapon /mnt/resource/swapfile' failed
[...]

PR #350, opened 2016-08-05
correct logic for dd fallback

hglkrijger:
when fallocate fails for any reason we should fall back to dd, not just when fallocate is not found. also adds unit test. fix for #347


brendandixon:
@hglkrijger LGTM

PR #352, opened 2016-08-08
Delete resolv.conf during deprovisioning

### LINKED ISSUE #351: [2.1.5]waagent doesn't delete /etc/resolv.conf during deprovisioning
Description of problem:
After deprovisioning, the /etc/resolv.conf is not deleted.

WALinuxAgent-2.1.5

RHEL Version:
RHEL-7.3 internal build

How reproducible:
100%

Steps to Reproduce:
1. Prepare a RHEL7.3 VM on Azure. Check /etc/resolv.conf:
\# ls -l /etc/resolv.conf
-rw-r--r--. 1 root root 125 Aug  8 10:18 /etc/resolv.conf
2. Deprovision the VM:
\# waagent -deprovision and input 'y' to execute deprovisioning.
3. Check the messages:
WARNING! The waagent service will be stopped.
WARNING! All SSH host key pairs will be deleted.
WARNING! Cached DHCP leases will be deleted.
WARNING! root password will be disabled. You will not be able to login as root.
Do you want to proceed (y/n)y
4. Check the /etc/resolv.conf
\# ls -l /etc/resolv.conf
-rw-r--r--. 1 root root 125 Aug  8 10:18 /etc/resolv.conf

Actual results:
There's no WARNING message to show that /etc/resolv.conf file will be deleted.
The /etc/resolv.conf file is not deleted after deprovisioning.

Expected results:
There's a WARNING message to show that /etc/resolv.conf file will be deleted.
The /etc/resolv.conf file is deleted after deprovisioning.

Additional info:
1. In WALinuxAgent-2.0.16 the /etc/resolv.conf is deleted during deprovisioning.
2. It seems that there's no code to delete /etc/resolv.conf in azurelinuxagent/pa/deprovision/default.py.

hglkrijger:
This behavior is missing from 2.1.x but did exist in 2.0.x. We have special logic in place for Ubuntu, but we should always remove resolv.conf during deprovisioning.

- correctly delete /etc/resolv.conf during deprovisioning
- add unit tests
- fixes #351 

/cc @szarkos @yuxisun1217 @ahmetalpbalkan @brendandixon

### DIFF
--- a/azurelinuxagent/pa/deprovision/default.py
+++ b/azurelinuxagent/pa/deprovision/default.py
@@ -74,6 +74,11 @@ class DeprovisionHandler(object):
         files_to_del = ['/root/.bash_history', '/var/log/waagent.log']
         actions.append(DeprovisionAction(fileutil.rm_files, files_to_del))
 
+    def del_resolv(self, warnings, actions):
+        warnings.append("WARNING! /etc/resolv.conf will be deleted.")
+        files_to_del = ["/etc/resolv.conf"]
+        actions.append(DeprovisionAction(fileutil.rm_files, files_to_del))
+
     def del_dhcp_lease(self, warnings, actions):
         warnings.append("WARNING! Cached DHCP leases will be deleted.")
         dirs_to_del = ["/var/lib/dhclient", "/var/lib/dhcpcd", "/var/lib/dhcp"]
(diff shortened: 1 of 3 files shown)


brendandixon:
@hglkrijger LTGM

szarkos:
Looks good.

As future improvement we might consider that other distros could potentially use resolvconf and just attempt to detect the difference in the generic DeprovisionHandler.del_resolv().  The Ubuntu-specific class and ubuntu.py would then not be needed at all.

hglkrijger:
Thanks @szarkos - I was not sure whether the Ubuntu-specific logic was required, or whether we could use this same logic in the generic `del_resolv`. If you think it's worth changing now, I could do that?

hglkrijger:
Will leave this in place for now.

PR #354, opened 2016-08-10
rdma: reorder device processing steps back to v2.0

ahmetalpbalkan:
With this change, RDMA integration tests look as follows:

image


hglkrijger:
lgtm

PR #371, opened 2016-08-17
revive #193, which fixes #177

### LINKED ISSUE #177: [2.1] "--init-system" gets ignored completely

hglkrijger:
Not sure how it happened, but it seems the fix for #177 which was in #193 was not merged, just closed (apologies, looks like my fault). This is the same fix.


ahmetalpbalkan:
LGTM.

Issue #369, opened 2016-08-17
[2.1.5] The configuration parameters may need to be updated in waagent.conf


yuxisun1217:
This is an issue that is separated from #360 to record the configuration parameters we may need to update.

Role.ConfigurationConsumer (Useless in wala-2.1.5. Need to be removed)
Role.TopologyConsumer (Useless in wala-2.1.5. Need to be removed)
Role.StateConsumer (If we confirm that wala will not add this feature, we can remove this)
AutoUpdate.GAFamily (Due to brendandixon's explanation, this parameters cannot be removed from waagent.conf now, so may I suggest to add a comment for this in the waagent.conf that tell customers not to edit this value?)

hglkrijger:
these have been merged

PR #372, opened 2016-08-18
update configuration files for #369

hglkrijger:
update configuration files for #369

### DIFF
--- a/config/alpine/waagent.conf
+++ b/config/alpine/waagent.conf
@@ -2,17 +2,6 @@
 # Windows Azure Linux Agent Configuration
 #
 
-# Specified program is invoked with the argument "Ready" when we report ready status
-# to the endpoint server.
-Role.StateConsumer=None
-
-# Specified program is invoked with XML file argument specifying role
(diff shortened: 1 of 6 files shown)


config/alpine/waagent.conf
```diff
@@ -69,3 +58,9 @@ OS.RootDeviceScsiTimeout=300
 
 # If "None", the system default version is used.
 OS.OpensslPath=None
+
+# Enable or disable goal state processing auto-update, default is enabled
+# AutoUpdate.Enabled=y
```
ahmetalpbalkan: why are these commented? e.g. ubuntu/waagent.conf has

```

# Enable or disable self-update, default is enabled
AutoUpdate.Enabled=y
AutoUpdate.GAFamily=Prod

```

hglkrijger: I changed them all to be commented out, based on the issues which were opened and resulting discussion.

ahmetalpbalkan:
ah ok, LGTM

Issue #388, opened 2016-08-24
WALA 2.1.3 does not pick up substatus messages from extensions


krkhan:
```
root@UbuntuTestP1VM:/var/lib/waagent/Microsoft.Azure.Security.AzureDiskEncryptionForLinux-0.1.0.999172# cat status/2.status | python -m json.tool
[
    {
        "status": {
            "code": "0",
            "formattedMessage": {
                "lang": "en-US",
                "message": "https://ubuntutestp1keyvault.vault.azure.net/secrets/c31796bf-b7c9-462a-bc47-c8f291e589fe/262a43212a5b44db8e6b108a8bea9473"
            },
            "name": "Microsoft.Azure.Security.AzureDiskEncryptionForLinux",
            "operation": "EnableEncryption",
            "status": "success",
            "substatus": [
                {
                    "code": "0",
                    "formattedMessage": {
                        "lang": "en-US",
                        "message": "{\"os\": \"VMRestartPending\", \"data\": \"NotEncrypted\"}"
                    },
                    "name": "Microsoft.Azure.Security.AzureDiskEncryptionForLinux",
                    "operation": "EnableEncryption",
                    "status": "success"
                }
            ]
        },
        "timestampUTC": "2016-08-24T00:46:24Z",
        "version": "1.0"
    }
]
```

No substatus in Resource Explorer:

<img width="438" alt="capture" src="https://cloud.githubusercontent.com/assets/697967/17941370/df81e5ca-69e7-11e6-95bc-7632a2499686.PNG">

brendandixon:
@krkhan Thank you. We'll address this in our next release.

PR #390, opened 2016-08-25
Fix issue Azure/WALinuxAgent#388

krkhan:


hglkrijger:
lgtm

brendandixon:
@hglkrijger Tests would be good. I did verify the code against the Windows Guest Agent. This is the correct format.

hglkrijger:
@brendandixon, we can also take this now if we need to and add tests later.

brendandixon:
@hglkrijger @krkhan I'm fine with that. It's a small change. It matches the Windows code. Oddly, I do not see corresponding code in the waagent2.0 file. Perhaps there was a format change and the v2.1.x branch missed full implementation?

brendandixon:
@hglkrijger Will our pipeline catch errors here? That is, if an extension writes status, will the pipeline detect that the agent correctly forwards it?

hglkrijger:
@brendandixon - not sufficiently. Our automation today checks the local settings file for the expected status, but does not check that it is forwarded out. We can certainly add more rigorous checks in the upcoming replacement pipeline.

Issue #391, opened 2016-08-26
[2.1.5] python setup.py install --register-service doesn't start service


yuxisun1217:
Hi,

I notice that in the v2.1.5, the "python setup.py install --register-service" doesn't start waagent service any more, which does start waagent service in the v2.1.4. Perhaps it's because the "Agent(False).register_service()" change to "get_osutil().register_agent_service()" in the setup.py script. And the get_osutil().register_agent_service function only register the service but not start the service. 
Is it the expected result? Thanks!

hglkrijger:
Fix is merged, closing

PR #397, opened 2016-08-30
update setup.py to start service on register-service

hglkrijger:
When calling `waagent --register-service` we register _and_ start the service, see [here](https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/agent.py#L88-L95). 

However, when we call `python setup.py --install register-service` we simply register. These two should be equivalent, and since v2.1.4 started the service as well, choosing that behavior as correct.

- starts the service when calling register-service on install
- addresses #391 

/cc @brendandixon


brendandixon:
lgtm

Issue #364, opened 2016-08-15
[2.1.5] Upload BlockBlob doesn't use proxy, and a risk of HttpProxy


yuxisun1217:
Hi,

In 2.1.5, it seems that the put block/page blob request doesn't use proxy(wire.py Line 390 and 410), which do use proxy in 2.0.16. Is it a change or an issue? I'm not quite sure if it is necessary to use proxy.

And there might be a risk of the connection between WALA and Azure Server(168.63.129.16). If I configure a Network Security Group, add rules like this:
![screenshot from 2016-08-15 19-32-12](https://cloud.githubusercontent.com/assets/19280786/17663420/04339c9e-631f-11e6-897b-1f43c51a814d.png)
The rules only allow proxy server(172.20.0.254:3128) and deny all others. So that the WALA fails to connect to Azure Server even if enable HttpProxy(Because some of the wala http requests never use proxy). I know this is not a good policy but at least the Azure Portal doesn't forbidden customers to do that and even has no warning messages on UI...
Is there any reason that cannot let all the wala http request use proxy if enable HttpProxy in waagent.conf?(Perhaps when proxy service and the client VM are in different regions? I guess)

brendandixon:
@yuxisun1217 Thank you for finding this. We'll address it in v2.1.7.

hglkrijger:
There are some additional details in the Bugzilla bug, listing here as well:

# Description of problem:
"HttpProxy" configuration in /etc/waagent.conf doesn't work. The waagent connect to Azure Server directly.

Version-Release number of selected component (if applicable):
WALinuxAgent-2.1.5 (upstream, package by tester)

RHEL Version:
RHEL-7.3-20160729.1

How reproducible:
100%

# Steps to Reproduce:
1. Prepare a VM on Azure with squid installed. Prepare a RHEL7.3 VM on Azure. These 2 VMs are in the same Virtual Network. Make sure that the client VM can connect to the Internet through the squid proxy, such as:

`wget -e "http_proxy=[squid VM private IP]:3128" http://www.baidu.com`

Can download index.html successfully.

2. Change HttpProxy configurations in /etc/waagent.conf at the client VM:
```
HttpProxy.Host=[squid VM private IP]
HttpProxy.Port=3128
```

3. restart waagent service

`systemctl restart waagent`

3. Catch packages at the squid VM and the client VM:

[...]

PR #399, opened 2016-08-30
Use proxy for storage calls

hglkrijger:
- ensure all calls to `call_storage_service` use the http proxy
- pep8 fixes and cleanup
- addresses #364 

/cc @brendandixon

### DIFF
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -16,13 +16,16 @@
 #
 # Requires Python 2.4+ and Openssl 1.0+
 
+import json
+import os
+import re
 import time
 import xml.sax.saxutils as saxutils
 import azurelinuxagent.common.conf as conf
 from azurelinuxagent.common.exception import ProtocolNotFoundError
 from azurelinuxagent.common.future import httpclient, bytebuffer
-from azurelinuxagent.common.utils.textutil import parse_doc, findall, find, findtext, \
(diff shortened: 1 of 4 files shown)


azurelinuxagent/common/protocol/wire.py
```diff
@@ -369,10 +369,14 @@ def get_blob_type(self, url):
...
-                'x-ms-version': self.__class__.__storage_version__
-            })
+            resp = self.client.call_storage_service(
+                restutil.http_head,
+                url,
+                {
+                    "x-ms-date": timestamp,
+                    'x-ms-version': self.__class__.__storage_version__
+                },
+                chk_proxy=True)
```
brendandixon: @hglkrijger Could we add this to the call_storage_service method directly? That way, it will always be present vs. having to ensure every caller passes.

brendandixon:
@hglkrijger I added a line comment. Could we forcibly add the chi_proxy flag within the call_storage_service method?

hglkrijger:
@brendandixon, sure. Initially that route seemed a bit heavy-handed to me, but after thinking about it some more I think it's the right call. Updated the PR.

jasonzio:
Now that chk_proxy is forced on, you don't need to supply it in all the calls.

hglkrijger:
@jasonzio, updated the PR.

jasonzio:
lgtm

hglkrijger:
thanks

PR #402, opened 2016-08-31
update the force option for mkfs.xfs

### LINKED ISSUE #400: [2.1.6] Fail to set resource disk to xfs type in RHEL-7
Hi,

Do we support xfs type for resource disk now? Because xfs type is the default file system type in RHEL-7.
If we want to support xfs type, it seems that mkfs.xfs command doesn't have -F option(Line128 in azurelinuxagent/daemon/resourcedisk/default.py: shellutil.run("mkfs." + fs + " " + partition + " -F")). I think we can use '-f' if the type is xfs. Thanks!

WALA version: 2.1.6
OS: RHEL-7.3 internal build

Reproduce steps:
1. Set ResourceDisk.Filesystem=xfs in waagent.conf. 
2. Deprovision this VM. Capture an image. Create a new VM base on this image.

Actual Result:
Cannot format /dev/sdb1 to xfs type. 
Error logs in waagent.log:
2016/08/31 16:23:55.060711 ERROR run cmd 'mkfs.xfs /dev/sdb1 -F' failed
2016/08/31 16:23:55.084195 ERROR Error Code:1
2016/08/31 16:23:55.089074 ERROR Result:mkfs.xfs: invalid option -- 'F'
unknown option -F

hglkrijger:
The force option for xfs is different than the other filesystem types. 

- check for xfs and update the force option accordingly
- fixes #400 

/cc @brendandixon

### DIFF
--- a/azurelinuxagent/daemon/resourcedisk/default.py
+++ b/azurelinuxagent/daemon/resourcedisk/default.py
@@ -106,7 +106,10 @@ class ResourceDiskHandler(object):
             raise ResourceDiskError("Could not determine partition info for "
                                     "{0}: {1}".format(device, ret[1]))
 
-        mkfs_string = "mkfs.{0} {1} -F".format(fs, partition)
+        force_option = 'F'
+        if fs == 'xfs':
+            force_option = 'f'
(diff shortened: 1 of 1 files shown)


brendandixon:
lgtm

Issue #361, opened 2016-08-12
503 Error Handling


danielsollondon:
Hi, can we consider changing the code to forever ignore (with sleeps) 503s from wireserver (and only wireserver). Wireserver can be assumed to be there by the agent. 

thanks.

brendandixon:
We should change wire.py#call_wireserver to sleep the same LONG_WAITING_INTERVAL we do for httpclient.FORBIDDEN.

hglkrijger:
merged, closing

PR #405, opened 2016-09-02
handle wireserver 503s more gracefully

hglkrijger:
- when receiving a 503, sleep instead of throwing
- retry more often when updating health, up to ~15 minutes
- addresses #361 

/cc @brendandixon @danielsollondon


jasonzio:
lgtm

PR #413, opened 2016-09-07
Metadata certificate download support added.

SRIKKANTH:


azurelinuxagent/common/protocol/metadata.py
```diff
@@ -26,16 +26,22 @@
 import azurelinuxagent.common.logger as logger
 import azurelinuxagent.common.utils.restutil as restutil
 import azurelinuxagent.common.utils.textutil as textutil
+from azurelinuxagent.common.utils.textutil import parse_doc, findall, find, findtext, \
+    getattrib, gettext, remove_bom, get_bytes_from_pem
```
brendandixon: @SRIKKANTH We've tended to import that package name and then reference methods within the package by prepending the name. Could we adopt that pattern here as well (e.g., `textutil.parse_doc` vs. `parse_doc`).

SRIKKANTH: I have fixed it please merge these changes.

brendandixon:
lgtm

PR #414, opened 2016-09-07
Exclude current version from latest

### LINKED ISSUE #387: [2.1.6] The updated WALA will be used after restart waagent twice even if its version is lower than the local version
Description of problem:
If enable AutoUpdate, WALA will always use the updated WALA after restart waagent service twice, even if its version is echo to or lower than the local version.

WALinuxAgent-2.1.6

RHEL Version:
RHEL-7.3 internal build

How reproducible:
100%

Steps to Reproduce:
1. Prepare a RHEL7 VM on Azure East US location(ASM mode) with WALA-2.1.6 installed. Enable AutoUpdate in the /etc/waagent.conf:
\# AutoUpdate.Enabled=y
\# AutoUpdate.GAFamily=Prod
2. restart waagent service
\# systemctl restart waagent
3. Wait until the new WALA package is downloaded.
(Now WALA still use the local version)
\# ps aux|grep -E "waagent|WALinux"
python -u /usr/sbin/waagent -run-exthandlers
4. restart waagent service again
\# systemctl restart waagent
5. Check the -run-exthandler process
\# ps aux|grep -E "waagent|WALinux"
python -u bin/WALinuxAgent-2.1.6-py2.7.egg -run-exthandlers

Actual results:
WALA uses the updated version(2.1.6), not the local version(2.1.6).

Expected results:
WALA uses the local version because the new version is not higher than the local version.

Additional info:
1. If change the local version to 2.1.7 in the /usr/lib/python2.7/site-packages/azurelinuxagent/common/version.py, this issue also exists.
2. If I delete the /var/lib/waagent/WALinuxAgent-2.1.6/ and then restart waagent service, it will use the local version(python -u /usr/sbin/waagent -run-exthandlers)

hglkrijger:
- ensure current agent is excluded from get_latest_update
- adds unit test
- minor refactor and reformat
- fixes #387 

/cc @brendandixon

### DIFF
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -236,7 +236,7 @@ class UpdateHandler(object):
                     logger.info("Goal state agent {0} was orphaned -- exiting", CURRENT_AGENT)
                     break
 
-                if self._ensure_latest_agent():
+                if self._upgrade_available():
                     if len(self.agents) > 0:
                         logger.info(
                             u"Agent {0} discovered {1} as an update and will exit",
@@ -287,7 +287,7 @@ class UpdateHandler(object):
     def get_latest_agent(self):
         """
         If autoupdate is enabled, return the most current, downloaded,
-        non-blacklisted agent (if any).
+        non-blacklisted agent which is not the current version (if any).
         Otherwise, return None (implying to use the installed agent).
         """
 
@@ -295,7 +295,10 @@ class UpdateHandler(object):
             return None
         
         self._load_agents()
-        available_agents = [agent for agent in self.agents if agent.is_available]
+        available_agents = [agent for agent in self.agents
+                            if agent.is_available
+                            and agent.version != FlexibleVersion(AGENT_VERSION)]
+
         return available_agents[0] if len(available_agents) >= 1 else None
 
     def _emit_restart_event(self):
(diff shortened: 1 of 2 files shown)


azurelinuxagent/ga/update.py
```diff
@@ -287,15 +287,18 @@ def forward_signal(self, signum, frame):
...
         """
 
         if not conf.get_autoupdate_enabled():
             return None
         
         self._load_agents()
-        available_agents = [agent for agent in self.agents if agent.is_available]
+        available_agents = [agent for agent in self.agents
+                            if agent.is_available
+                            and agent.version != FlexibleVersion(AGENT_VERSION)]
```
brendandixon: Could this be ">"? I prefer stronger predicates.

azurelinuxagent/ga/update.py
```diff
@@ -312,7 +315,7 @@ def _emit_restart_event(self):
         self._set_sentinal() 
         return
 
-    def _ensure_latest_agent(self, base_version=CURRENT_VERSION):
+    def _upgrade_available(self, base_version=CURRENT_VERSION):
```
brendandixon: Huh! Didn't like my name! :)

hglkrijger: I found it a little confusing :) once I changed the name it seemed more readable to me..

PR #415, opened 2016-09-08
Fix the issue of failing to find resource disk for FreeBSD

ccic:
FreeBSD VM gets a UUID from host which informs where resource
disk locates. But that UUID string was not exposed from VM kernel,
as a result, the agent used a hard code "2:1:0" as the location of
resource disk. That should be updated after FreeBSD VM kernel
exposed UUID through "sysctl dev.storvsc" command.

But considering compatible with previous FreeBSD VM, the hard code
logic cannot be removed, and they are left there as a fallback.


ministat:
I have finished cla-requirement.

hglkrijger:
@ministat, seems the CLA has not gone through. Could you try again?

ccic:
I

PR #416, opened 2016-09-08
Report ready when provisioning is disabled

### LINKED ISSUE #411: [2.1.6] How does "Provisioning.Enabled=n" work now?
Hi,

In 2.1.6, if I set "Provisioning.Enabled=n" in the waagent.conf, run "waagent -deprovision", capture a image with "I've run waagent -deprovision" checked, then create a VM base on this image, the VM status is always "Running(Provisioning)" and after a long time becomes "Running(Provisioning Timeout)". And when I ssh into this VM, there's no "/var/lib/waagent/provisioned" file generated. Is it true?
I remember that in WALA-2.0.16, the VM can be created successfully. The status becomes "Running" and the "/var/lib/waagent/provisioned" file is generated.
In my opinion, if I set "Provisioning.Enabled=n", it means I'll use other tools to provision this VM. So I think wala should put the "provisioned" file there and the VM status should become "Running" without warning. Is it right? Thank you very much!

WALA version: 2.1.6
RHEL version: RHEL-6.8

Steps:
1. Prepare a VM on ASM mode. Set "Provisioning.Enabled=n" into the /etc/waagent.conf
2. Run "waagent -deprovision -force"
3. Deallocate this VM. Capture an image with "I've run waagent -deprovision" checked.
4. Create a VM base on this image.
5. Check the VM status. Check if the /var/lib/waagent/provisioned file is generated.

Actual result:
The VM status is Running(Provisioning) and after a long time becomes "Running(Provisioning Timeout)"
There's no /var/lib/waagent/provisioned file inside this VM.

hglkrijger:
We should allow the vm to report ready and mark provisioning complete when `Provisioning.Enabled=n`.

- write `provisioned` file and report `Ready` when provisioning is disabled in the default provider
- pep8 cleanup
- fixes #411 

/cc @brendandixon

### DIFF
--- a/azurelinuxagent/pa/provision/default.py
+++ b/azurelinuxagent/pa/provision/default.py
@@ -25,26 +25,26 @@ from azurelinuxagent.common.future import ustr
 import azurelinuxagent.common.conf as conf
 from azurelinuxagent.common.event import add_event, WALAEventOperation
 from azurelinuxagent.common.exception import ProvisionError, ProtocolError, \
-                                             OSUtilError
+    OSUtilError
 from azurelinuxagent.common.protocol.restapi import ProvisionStatus
 import azurelinuxagent.common.utils.shellutil as shellutil
 import azurelinuxagent.common.utils.fileutil as fileutil
 from azurelinuxagent.common.osutil import get_osutil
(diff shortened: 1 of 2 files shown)


brendandixon:
lgtm

PR #421, opened 2016-09-13
stop pidof from spamming the logs

### LINKED ISSUE #417: Spams journal with `pidof dhlient` related messages
We use our own image but we have the agent installed.

Since we are using systemd-networkd with its own DHCP stack, there is no dhclient being used.

How do we stop the log spam?  Is it safe to kill and disable this stuff?  it is there another solution? 

Relevant:

https://github.com/Azure/WALinuxAgent/issues/195#issuecomment-245983434

hglkrijger:
We should support scenarios when dhcpd is not present, and not spam logs with the same warnings over and over. 

- set `chk_err=False` on all calls to `pidof`
- only allow one warning if dhcps is not running
- fixes #417 

/cc @brendandixon

### DIFF
--- a/azurelinuxagent/common/osutil/alpine.py
+++ b/azurelinuxagent/common/osutil/alpine.py
@@ -29,7 +29,7 @@ class AlpineOSUtil(DefaultOSUtil):
         return True
 
     def get_dhcp_pid(self):
-        ret = shellutil.run_get_output("pidof udhcpc")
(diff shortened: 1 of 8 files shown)


brendandixon:
lgtm

truckee:
There is a post [here](https://community.spiceworks.com/topic/1580345-azure-ubuntu-vm-waagent-errors)  that describes a solution/workaround. What worked for me was `$ mkdir -p  /usr/share/sendmail/dynamic` and restarting networking.

PR #423, opened 2016-09-13
Removed unwanted debug messages.

SRIKKANTH:


brendandixon:
lgtm

hglkrijger:
@SRIKKANTH, any idea why the following warnings are shown?
```
2016/09/13 10:18:08.168792 WARNING Unknown property: certlist.certificates.storeLocation
2016/09/13 10:18:08.238270 WARNING Unknown property: certlist.certificates.storeName
```

SRIKKANTH:
Not sure why those warnings are showing up.  These properties are not defined and used in metadata.py.

hglkrijger:
@SRIKKANTH, not defined and not used, or not defined but used? I would suggest you dig into it, as the warnings are appended to the log every 30s. I imagine all that needs to happen is for these to be added to the `certificates` structure.

Issue #425, opened 2016-09-14
walinuxagent.service should not Want or After on cloud-final.service


smoser:
As filed in launchpad at [bug 1623570](https://bugs.launchpad.net/bugs/1623570) we've come across some issues with the After and Wants in walinux-agent.service.

This was noticed due to change to cloud-init for [bug 1576692](http://bugs.launchpad.net/bugs/1576692).  However, it is not clear why either of those were added to the unit.

brendandixon:
Fixed by #425.

PR #426, opened 2016-09-14
walinuxagent.service: drop cloud-final from Wants, After on Ubuntu (#425)

smoser:
There is no reason for walinuxagent.service to attempt to start up
cloud-final.service if it is running, so remove it from Wants.

There is also no reason for it to need to run After cloud-final, so
remove that also.   The need for this is because cloud-init was starting
walinux.service during cloud-init.service, which would cause unresolvable
cyclic dependencies.

### DIFF
--- a/init/ubuntu/walinuxagent.service
+++ b/init/ubuntu/walinuxagent.service
@@ -7,8 +7,8 @@
 [Unit]
 Description=Azure Linux Agent
 
-After=network-online.target cloud-final.service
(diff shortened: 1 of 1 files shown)


szarkos:
Hey Scott, I posted a comment about this on the Launchpad bug: https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/1623570/comments/4.  What are your thoughts on using the builtin approach starting with Yakkety?

PR #427, opened 2016-09-14
remove code which disables selinux

### LINKED ISSUE #367: [2.1.5] Agent disables SELinux
The agent, temporarily, disables SELinux during provisioning (so that it may make updates to critical files). A bette approach is to whitelist the agent.

hglkrijger:
- tested on RHEL 6.7 and RHEL 7.2
- fixes #367 

/cc @brendandixon @szarkos


brendandixon:
@hglkrijger Were these lines wholly unnecessary? What will not occur on an SELinux distribution without them?

PR #432, opened 2016-09-16
Cleanup SUSE config

bear454:
re: [BSC#997475 - waagent.conf in python-azure-agent package should use ext4 for ephemeral disk](https://bugzilla.suse.com/show_bug.cgi?id=997475) filed by @szarkos 

While correcting the SUSE config, a few other omissions were noted.

WRT to 1d4d383, this change is currently being made as part of image generation; moving to the config file reduces the number of times we need to touch `/etc/waagent.conf`.

### DIFF
--- a/config/suse/waagent.conf
+++ b/config/suse/waagent.conf
@@ -70,6 +70,21 @@ OS.OpensslPath=None
 # Detect Scvmm environment, default is n
 # DetectScvmmEnv=n
 
+#
+# Lib.Dir=/var/lib/waagent
+
+#
+# DVD.MountPoint=/mnt/cdrom/secure
+
+#
+# Pid.File=/var/run/waagent.pid
+
+#
+# Extension.LogDir=/var/log/azure
+
+#
+# Home.Dir=/home
+
 # Enable RDMA management and set up, should only be used in HPC images
 # OS.EnableRDMA=y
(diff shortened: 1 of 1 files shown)


szarkos:
Hi All,

What's holding up this commit?  If it's the AutoUpdate.Enabled=n change then we could possible revert that part, but clean up the rest.

Steve

config/suse/waagent.conf
```diff
@@ -70,11 +70,26 @@ OS.OpensslPath=None
...
+
+#
+# Home.Dir=/home
+
 # Enable RDMA management and set up, should only be used in HPC images
 # OS.EnableRDMA=y
 
 # Enable or disable goal state processing auto-update, default is enabled
-# AutoUpdate.Enabled=y
+AutoUpdate.Enabled=n
```
brendandixon: @bear454 We need to leave `AutoUpdate.Enabled` set to true (aka "y"). It is best for Azure customers.

brendandixon:
@szarkos Just other things going on and have not looked. I agree with reverting the change to `AutoUpdate.Enabled`. Once that's done we can merge.

bear454:
Okay...

brendandixon:
@bear454 Looks good.

Issue #443, opened 2016-09-23
Send additional headers when using HostGAPlugin


jinhyunr:
HostGAPlugin is used to report status and download extensions when network configuration on a VM is blocking VM Agent to do so via default mechanism. 

With the security guideline, When Agent need to utilize HostGAPlugin, it must send additional headers along with requests.

PR #444, opened 2016-09-23
Send additional headers when using HostGAPlugin (#443)

jinhyunr:
Taking care of "Send additional headers when using HostGAPlugin  #443"
Added necessary headers when sending put vm status request to HostGAPlugin.

Added UTs


jinhyunr:
Did not realize change in Get api versions. Will make another commit..

jinhyunr:
Spoke with Zhidong offline and confirmed 'Get API version' is still using GET instead of PUT. As it still requires a new header to be sent along with, refactored the code little.

hglkrijger:
let's hold off actually merging until we are done with this release, though.

PR #447, opened 2016-09-26
Add version hooks for F5 products

caphrim007:
F5's products do not use the common /etc/product-release format for
platform version information. A different method is needed via parsing
the /VERSION file.

I was not sure if /VERSION was too broad of a thing to check for (for
instance if it might crop up in other platforms) so instead I check
for the /shared/vadc which is definitely an F5-ism.

This patch adds a small method to parse the relevant content from that
file and put it in a list that is returned for later use by the rest
of the version.py lib.


caphrim007:
if we can hold on merging this for a day or so, i need to review a fix offline before I am confident this addresses the various f5 platforms.

PR #451, opened 2016-09-27
fix proxy port type

### LINKED ISSUE #364: [2.1.5] Upload BlockBlob doesn't use proxy, and a risk of HttpProxy

hglkrijger:
- ensure the proxy port is an integer
- logging improvements
- pep8 fixes
- fixes #364 

/cc @brendandixon

### DIFF
--- a/azurelinuxagent/common/conf.py
+++ b/azurelinuxagent/common/conf.py
@@ -24,10 +24,12 @@ import os
 import azurelinuxagent.common.utils.fileutil as fileutil
 from azurelinuxagent.common.exception import AgentConfigError
 
+
 class ConfigurationProvider(object):
     """
     Parse amd store key:values in /etc/waagent.conf.
     """
+
     def __init__(self):
         self.values = dict()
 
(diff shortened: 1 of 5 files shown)


brendandixon:
lgtm

Issue #452, opened 2016-09-29
Add Overprovisioning support


jinhyunr:
Upon VMSS overprovisioning, more than necessary VMs may get created. In this case, VM Extension handler handling on a VM should be waited until the VM is chosen to be a 'survivor'(if more than necessary VMs are created, surplus VMs will be removed). This can be determined by polling InVMArtifactsProfile blob.

PR #453, opened 2016-09-29
PR for #452, Add Overprovisioning support

jinhyunr:


azurelinuxagent/common/protocol/wire.py
```diff
@@ -764,6 +770,7 @@ def get_ext_manifest(self, ext_handler, goal_state):
                                                goal_state.incarnation)
         local_file = os.path.join(conf.get_lib_dir(), local_file)
         xml_text = self.fetch_manifest(ext_handler.versionUris)
+        print(xml_text)
```
hglkrijger: should this be a logger.verbose?

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -59,6 +62,40 @@ def get_api_versions(self):
...
+
+            response = restutil.http_get(url, headers)
+            if response.status != httpclient.OK:
+                logger.error(
+                    "get Extension Artifact returned status code [{0}]".format(
+                        response.status))
+                return None
+            resp_body = response.read()
+            if resp_body:
+                return remove_bom(bytearray(resp_body)).decode('utf-8')
```
hglkrijger: I generally recommend a single return statement for a method, so there is a clear flow of control.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -911,6 +917,43 @@ def get_header_for_cert(self):
             "x-ms-guest-agent-public-x509-cert": cert
         }
 
+    def get_host_plugin(self):
+        if self.host_plugin is None:
+            self.host_plugin = HostPluginProtocol(self.endpoint, self.get_goal_state())
+        return self.host_plugin
```
hglkrijger: do we need to support changes in the endpoint or goal state through the lifetime of the host plugin instance?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1272,3 +1320,32 @@ def _handle_packages(self, packages, isinternal):
...
+    * encryptedApplicationProfile (optional)
+    '''
+
+    def __init__(self, in_vm_artifacts_profile_json):
+        if in_vm_artifacts_profile_json and not in_vm_artifacts_profile_json.isspace():
+            self.__parse(in_vm_artifacts_profile_json)
+
+    def __parse(self, json_str):
+        # trim null and whitespaces
+        trimmed = json_str.rstrip(' \t\r\n\0')
```
hglkrijger: might want to put this in textutil, could be handy elsewhere

azurelinuxagent/common/protocol/wire.py
jinhyunr: Forgot to remove.. Thanks-

azurelinuxagent/common/protocol/wire.py
jinhyunr: Talked with Zhidong and apparently we should support GoalState change as RoleConfig name will change along with GoalState incarnation number.

azurelinuxagent/common/protocol/wire.py
jinhyunr: Will remove.

azurelinuxagent/common/protocol/wire.py
jinhyunr: Talked with Zhidong and apparently we should support GoalState change as RoleConfig name will change along with GoalState incarnation number.

jinhyunr: Yes, we need to support changes in goal state (I guess change in HostPlugin endpoint does not really make much sense) and it is done in Update_goal_state()

Every preset interval, ExtHandlersHandler.run(),
```js
     def run(self):
         self.ext_handlers, etag = None, None
         try:
         except ProtocolError as e:
```
gets executed which triggers protocol.get_ext_handlers(), 
```js
    def get_ext_handlers(self):
         logger.verbose("Get extension handler config")
         # Update goal state to get latest extensions config
         self.client.update_goal_state()
         goal_state = self.client.get_goal_state()
         ext_conf = self.client.get_ext_conf()
         # In wire protocol, incarnation is equivalent to ETag
         return ext_conf.ext_handlers, goal_state.incarnation
```
then WireClient.update_goal_state(). 

```js
@@ -702,6 +706,8 @@ def update_goal_state(self, forced=False, max_retry=3):
                  self.update_shared_conf(goal_state)
                  self.update_certs(goal_state)
                  self.update_ext_conf(goal_state)
 +                if self.host_plugin is not None:
 +                    self.host_plugin.goal_state = goal_state
                  return
```

This will effectively update goal_state ref of host_plugin.

azurelinuxagent/common/protocol/hostplugin.py
jinhyunr: done

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -22,19 +22,24 @@
...
 URI_FORMAT_PUT_LOG = "http://{0}:{1}/vmAgentLog"
 API_VERSION = "2015-09-01"
+HEADER_CONTAINER_ID = "x-ms-containerid"
+HEADER_VERSION = "x-ms-version"
+HEADER_HOST_CONFIG_NAME = "x-ms-host-config-name"
 
 
 class HostPluginProtocol(object):
-    def __init__(self, endpoint):
+    def __init__(self, endpoint, goal_state):
```
brendandixon: I think I prefer that we pass the container_id and role_instance_config_name vs. the entire goal state. The more data passed, the more tightly the code is bound together. It may not matter much here, but it's a good practice to keep the binding between objects limited to what they need to know.

jinhyunr: done

PR #455, opened 2016-09-29
Added once-daily reporting

brendandixon:

### DIFF
--- a/azurelinuxagent/common/osutil/alpine.py
+++ b/azurelinuxagent/common/osutil/alpine.py
@@ -29,7 +29,7 @@ class AlpineOSUtil(DefaultOSUtil):
         return True
 
     def get_dhcp_pid(self):
-        ret = shellutil.run_get_output("pidof udhcpc", chk_err=False)
+        ret = shellutil.run_get_output('pidof dhcpcd', chk_err=False)
         return ret[1] if ret[0] == 0 else None
 
     def set_ssh_client_alive_interval(self):


azurelinuxagent/ga/monitor.py
```diff
@@ -177,11 +180,35 @@ def collect_and_send_events(self):
         except ProtocolError as e:
             logger.error("{0}", e)
 
+    def send_daily_heartbeat(self):
+        """
+        Send the daily heartbeat containing the installed and current agent versions.
+        Send once during the first 30 minutes of the day.
+        """
+        if datetime.datetime.time() <= datetime.time(0, minute=30):
```
hglkrijger: What if the VM is provisioned for less than 24 hours?

azurelinuxagent/ga/monitor.py
```diff
@@ -177,11 +180,35 @@ def collect_and_send_events(self):
...
+                logger.warn("Exception sending daily heartbeat on {0}: {1}",
+                    today.strftime(ISOFORMAT),
+                    e)
+
     def daemon(self):
         last_heartbeat = datetime.datetime.min
         period = datetime.timedelta(minutes=30)
         while True:
             if (datetime.datetime.now() - last_heartbeat) > period:
+                send_daily_heartbeat()
```
hglkrijger: Wouldn't it be better to send the daily heartbeat once and then wait for a 24 hour period?

azurelinuxagent/ga/monitor.py
```diff
@@ -177,11 +180,35 @@ def collect_and_send_events(self):
...
+    def send_daily_heartbeat(self):
+        """
+        Send the daily heartbeat containing the installed and current agent versions.
+        Send once during the first 30 minutes of the day.
+        """
+        if datetime.datetime.time() <= datetime.time(0, minute=30):
+            try:
+                msg = u"{0}:{1}".format(
+                    INSTALLED_VERSION,
+                    CURRENT_VERSION)
```
hglkrijger: Would the python version be interesting here? What about extensions?

brendandixon: I consider it, but is a variable amount of data. The underlying goal is to improve basic telemetry so we know what versions are running.

azurelinuxagent/ga/monitor.py
brendandixon: I thought about that. In fact, I had originally written a timestamp to disk so that it was emitted only once every 24 hours. But the small error we have with the simpler code did not, in my mind, justify the increased complexity of maintaining the timestamp.

azurelinuxagent/ga/monitor.py
brendandixon: See above. :)

azurelinuxagent/ga/monitor.py
hglkrijger: Ok. Are we concerned at all about all VMs emitting telemetry within the same 30 minute window?

brendandixon: @hglkrijger I wasn't...but maybe we should be. I'll re-evaluate emitting once-per-day based on a timestamp.

PR #459, opened 2016-10-07
Add version hooks for F5 products

caphrim007:
F5's products do not use the common /etc/product-release format for
platform version information. A different method is needed via parsing
the /VERSION file.

I was not sure if /VERSION was too broad of a thing to check for (for
instance if it might crop up in other platforms) so instead I check
for the /shared/vadc which is definitely an F5-ism.

This patch adds a small method to parse the relevant content from that
file and put it in a list that is returned for later use by the rest
of the version.py lib.

### DIFF
--- a/azurelinuxagent/common/version.py
+++ b/azurelinuxagent/common/version.py
@@ -39,10 +39,15 @@ def get_distro():
         osinfo = platform.dist()
 
     # The platform.py lib has issue with detecting oracle linux distribution.
-    # Merge the following patch provided by oracle as a temparory fix.
+    # Merge the following patch provided by oracle as a temporary fix.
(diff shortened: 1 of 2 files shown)


caphrim007:
@brendandixon @hglkrijger any comments on this?

brendandixon:
@caphrim007 It looks good, but we should have tests. While much of the early code lacks tests, we're trying to ensure rigorous testing of all new code.

caphrim007:
@brendandixon absolutely. Lemme go find out where to put those tests. Thanks!

caphrim007:
@brendandixon ready for review

brendandixon:
@caphrim007 LGTM.

PR #475, opened 2016-10-31
Adds tests for iWorkflow and BIG-IQ

caphrim007:
This patch adds remaining tests for iWorkflow and BIG-IQ platforms


brendandixon:
@caphrim007 LGTM.

PR #476, opened 2016-10-31
Don't write IP/MAC to device running newer driver version 4.1 or newer

longlimsft:
Since RDMA driver 4.1 we are moving to a different framework to probe ND device. The device for exposing MAC/IP is removed from kernel and no longer needed.

### DIFF
--- a/azurelinuxagent/common/rdma.py
+++ b/azurelinuxagent/common/rdma.py
@@ -199,10 +199,24 @@ class RDMADeviceHandler(object):
     def process(self):
         try:
             RDMADeviceHandler.update_dat_conf(dapl_config_paths, self.ipv4_addr)
-            RDMADeviceHandler.wait_rdma_device(
-                self.rdma_dev, self.device_check_timeout_sec, self.device_check_interval_sec)
-            RDMADeviceHandler.write_rdma_config_to_device(
-                self.rdma_dev, self.ipv4_addr, self.mac_addr)
+
+            skip_rdma_device = False;
+            retcode,out = shellutil.run_get_output("modinfo hv_network_direct")
+            if retcode == 0 :
+                version = re.search("version:\s+(\d+)\.(\d+)\.(\d+)\D", out, re.IGNORECASE)
+                if version :
+                    v1 = version.groups(0)[0]
+                    v2 = version.groups(0)[1]
+                    if int(v1+v2)>40 :
+                        logger.info("Skip setting /dev/hvnd_rdma on 4.1 or later")
+                        skip_rdma_device = True;
+
+            if not skip_rdma_device :
+                RDMADeviceHandler.wait_rdma_device(
+                    self.rdma_dev, self.device_check_timeout_sec, self.device_check_interval_sec)
+                RDMADeviceHandler.write_rdma_config_to_device(
+                    self.rdma_dev, self.ipv4_addr, self.mac_addr)
+
             RDMADeviceHandler.update_network_interface(self.mac_addr, self.ipv4_addr)
         except Exception as e:
             logger.error("RDMA: device processing failed: {0}".format(e))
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/rdma.py
```diff
@@ -199,10 +199,24 @@ def start(self):
...
-                self.rdma_dev, self.ipv4_addr, self.mac_addr)
+
+            skip_rdma_device = False;
+            retcode,out = shellutil.run_get_output("modinfo hv_network_direct")
+            if retcode == 0 :
+                version = re.search("version:\s+(\d+)\.(\d+)\.(\d+)\D", out, re.IGNORECASE)
+                if version :
+                    v1 = version.groups(0)[0]
+                    v2 = version.groups(0)[1]
+                    if int(v1+v2)>40 :
```
ahmetalpbalkan: what's this 40 @longlimsft ?

ahmetalpbalkan: if you're trying to check v>4.1 this is not a good way to do it. v0.101 will pass this check too

azurelinuxagent/common/rdma.py
```diff
@@ -199,10 +199,24 @@ def start(self):
...
         try:
             RDMADeviceHandler.update_dat_conf(dapl_config_paths, self.ipv4_addr)
-            RDMADeviceHandler.wait_rdma_device(
-                self.rdma_dev, self.device_check_timeout_sec, self.device_check_interval_sec)
-            RDMADeviceHandler.write_rdma_config_to_device(
-                self.rdma_dev, self.ipv4_addr, self.mac_addr)
+
+            skip_rdma_device = False;
+            retcode,out = shellutil.run_get_output("modinfo hv_network_direct")
+            if retcode == 0 :
```
ahmetalpbalkan: extra spaces before `:`s in your patch

azurelinuxagent/common/rdma.py
```diff
@@ -199,10 +199,24 @@ def start(self):
     def process(self):
         try:
             RDMADeviceHandler.update_dat_conf(dapl_config_paths, self.ipv4_addr)
-            RDMADeviceHandler.wait_rdma_device(
-                self.rdma_dev, self.device_check_timeout_sec, self.device_check_interval_sec)
-            RDMADeviceHandler.write_rdma_config_to_device(
-                self.rdma_dev, self.ipv4_addr, self.mac_addr)
+
+            skip_rdma_device = False;
```
ahmetalpbalkan: some `;`s in your patch, not needed

azurelinuxagent/common/rdma.py
longlimsft: I think this should be fine, we check against 4.0. The version numbers can only go up in future.

longlimsft: I think this should be fine, we check against 4.0. The version numbers can only go up in future.

longlimsft: I think this should be fine, we check against 4.0. The version numbers can only go up in future.

azurelinuxagent/common/rdma.py
longlimsft: Will fix this

azurelinuxagent/common/rdma.py
longlimsft: Will fix this.

azurelinuxagent/common/rdma.py
ahmetalpbalkan: @longlimsft it's very sloppy. is it harder to write a longer if statement?

hglkrijger:
@ahmetalpbalkan, @longlimsft - any updates on this?

ahmetalpbalkan:
This currently needs fixes from Long.

longlimsft:
I will request a new pull request to address comments.

PR #478, opened 2016-11-03
Fix resource disk partition format detection

### LINKED ISSUE #457: "sfdisk" on Ubuntu 16.04 breaks filesystem detection
**How to reproduce:** 
Deploy an Ubuntu 16.04 machine on Azure, set

> ResourceDisk.Format=y

It doesn't format the resource disk as documented

**Problem:**
"sfdisk -c" emits deprecated warning

> sfdisk: --id is deprecated in favour of --part-type

Furthermore, "sfdisk --part-type" output with leading space " 7" therefore rstrip won't work.

https://github.com/Azure/WALinuxAgent/blob/47442f76a98e2fd647bf8d3099d5354529dfa03b/azurelinuxagent/daemon/resourcedisk/default.py

**Impact:**
Resource disk with ntfs filesystem is not detected and mkfs-ed, and mounted as-is

bryteise:
With recent versions of sfdisk the -c command is deprecated and
--part-type is recommended to be used instead. This use also requires
the output to be stripped instead of rstripped due to a leading space.

This resolves #457 (it is used as a patch for Clear Linux's WALinuxAgent currently).

### DIFF
--- a/azurelinuxagent/daemon/resourcedisk/default.py
+++ b/azurelinuxagent/daemon/resourcedisk/default.py
@@ -129,12 +129,12 @@ class ResourceDiskHandler(object):
                 shellutil.run(mkfs_string)
         else:
             logger.info("GPT not detected, determining filesystem")
-            ret = shellutil.run_get_output("sfdisk -q -c {0} 1".format(device))
-            ptype = ret[1].rstrip()
+            ret = shellutil.run_get_output("sfdisk -q --part-type {0} 1".format(device))
+            ptype = ret[1].strip()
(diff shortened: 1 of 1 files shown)


hglkrijger:
@bryteise, thanks for the patch. How recently is this parameter supported? We are running on a lot of older distros as well. It might make sense to try one and if that fails then try the other.

bryteise:
@hglkrijger util-linux updated sfdisk to change the option in 2014. See https://github.com/karelzak/util-linux/commit/e36fb07a.

Issue #481, opened 2016-11-07
Ignore InVMArtifactsProfile.OnHold once it becomes False


jinhyunr:
Once Agent receives false flag of InVMArtifactsProfile.OnHold, it should never put extension handling on hold: OnHold happens only at the very beginning of a VM life cycle.

hglkrijger:
fixed with #482

PR #482, opened 2016-11-07
Ignore InVMArtifactsProfile.OnHold once it becomes false (#481)

jinhyunr:
InVMArtifactsProfile.OnHold is used to support overprovisioning feature of CRP VMSS. Thus, the flag is meaningful only at the very beginning of the VM lifecycle and once it becomes False, Agent should never put extension processing on hold. However, under certain circumstance like VM losing an outbound internet connection, the current Agent suffers from making the correct decision as it loses access to VMArtifactsProfile when internet is not available.

The PR is needed to resolve the issue.

After the PR, Agent makes the decision based on the following conditions:
1. If VMArtifactsProfileBlobUrl is null/empty, do not hold
2. If WasOnHoldEverFalse file exists, do not hold
3. If VMArtifactsProfile is null, hold
4. If VMArtifactsProfile is not null, honor the onHold flag.

### DIFF
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -49,6 +49,7 @@ EXT_CONF_FILE_NAME = "ExtensionsConfig.{0}.xml"
 MANIFEST_FILE_NAME = "{0}.{1}.manifest.xml"
 TRANSPORT_CERT_FILE_NAME = "TransportCert.pem"
 TRANSPORT_PRV_FILE_NAME = "TransportPrivate.pem"
+WAS_ON_HOLD_EVER_SET_TO_FALSE = "OnHoldSetToFalse"
 
 PROTOCOL_VERSION = "2012-11-30"
 ENDPOINT_FINE_NAME = "WireServer"
@@ -142,6 +143,10 @@ class WireProtocol(Protocol):
     def get_in_vm_artifacts_profile(self):
         logger.verbose("Get In-VM Artifacts Profile")
         return self.client.get_in_vm_artifacts_profile()
+
+    def is_ext_handling_on_hold(self):
+        logger.verbose("Check if extension handling is on hold")
+        return self.client.is_extension_handling_on_hold()
 
     def report_provision_status(self, provision_status):
         validate_param("provision_status", provision_status, ProvisionStatus)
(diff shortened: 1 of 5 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -195,9 +195,7 @@ def handle_ext_handlers(self, etag=None):
             logger.verbose("No ext handler config found")
             return
 
-        in_vm_artifacts_profile = self.protocol.get_in_vm_artifacts_profile()
-        if in_vm_artifacts_profile and \
-           in_vm_artifacts_profile.is_extension_handlers_handling_on_hold():
+        if self.protocol.is_ext_handling_on_hold():
```
hglkrijger: Does this method need to be implemented or stubbed out in the Azure Stack / metadata protocol as well?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -925,6 +942,30 @@ def get_host_plugin(self):
...
+                # depends on the existence of WAS_ON_HOLD_EVER_SET_TO_FALSE file to determine
+                # whether to hold extension handling
+                is_extension_handling_on_hold = not os.path.isfile(was_on_hold_ever_set_to_false)
+            else:
+                is_extension_handling_on_hold = in_vm_artifacts_profile.is_extension_handling_on_hold()
+
+        if not is_extension_handling_on_hold and not os.path.isfile(was_on_hold_ever_set_to_false):
+            # If extension handling is not on hold, create WAS_ON_HOLD_EVER_SET_TO_FALSE file
+            # Once on hold is set to false, WALA should never hold extension handling in the current VM
+            fileutil.write_file(was_on_hold_ever_set_to_false, "")
```
hglkrijger: should probably add more logging throughout this method?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -925,6 +942,30 @@ def get_host_plugin(self):
                                                   goal_state.role_instance_config_name)
         return self.host_plugin
 
+    def is_extension_handling_on_hold(self):
+        is_extension_handling_on_hold = False
+        ext_conf = self.ext_conf
+        was_on_hold_ever_set_to_false = os.path.join(conf.get_lib_dir(), WAS_ON_HOLD_EVER_SET_TO_FALSE)
```
hglkrijger: Maybe just me, but this name confuses me, Can we pick something else?

hglkrijger:
Discussed offline, this is on hold so closing the PR for now.

Issue #486, opened 2016-11-08
wire.py#decode_config needs to handle empty strings


brendandixon:
wire.py#decode_config invokes `remove_bom` which assumes the string has content. Either `decode_config` or, maybe better, `remove_bom` must first ensure the string is long-enough to even contain a BOM.

hglkrijger:
fixed with #493

PR #487, opened 2016-11-09
wire.py#decode_config needs to handle empty strings (#486)

jinhyunr:
1. Add is_str_none_or_whitespace to textutil for string emptiness check
2. Update remove_bom to properly handle empty/none/whitespace string

### DIFF
--- a/azurelinuxagent/common/protocol/hostplugin.py
+++ b/azurelinuxagent/common/protocol/hostplugin.py
@@ -70,7 +70,7 @@ class HostPluginProtocol(object):
         if not self.ensure_initialized():
             logger.error("host plugin channel is not available")
             return
-        if artifact_url is None or artifact_url.isspace():
+        if textutil.is_str_none_or_whitespace(artifact_url):
             logger.error("no extension artifact url was provided")
             return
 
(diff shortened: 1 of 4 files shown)


brendandixon:
@jinhyunr What if the string contains too few legitimate characters (e.g., "a")? Won't the code still fail?

jinhyunr:
```python
def remove_bom(c):
    if not is_str_none_or_whitespace(c) and \
       str_to_ord(c[0]) > 128 and \
       str_to_ord(c[1]) > 128 and \
       str_to_ord(c[2]) > 128:
        c = c[3:]
    return c
```

in case of 'a', str_to_ord(c[0]) > 128 becomes False, so it will simply return 'a' as it is.

brendandixon:
@jinhyunr I think you are missing my point: I can construct a string that is too short that will cause the method to throw an exception. My point is, that the fix evaluates for types of strings (e.g., all whitespace), where the exception occurs because the length is too short. The code must ensure sufficient characters exist.

hglkrijger:
@jinhyunr - looks like a merge conflict with the hotfix, sorry. please take a look when you can.

PR #492, opened 2016-11-16
Correct the version check in decide_version

hglkrijger:
When extensions have multiple major versions available, initial installation will fail due to the major version upgrade allowed check. 
 
- correct the installed version check 
- check the string length in `remove_bom`, I hit this in my repro testing


jinhyunr:
LGTM

hglkrijger:
This should actually go to a hotfix branch first, closing the PR for now

PR #493, opened 2016-11-17
Decide version hotfix

hglkrijger:
When extensions have multiple major versions available, initial installation will fail due to the major version upgrade allowed check.

- correct the installed version check
- check the string length in `remove_bom`, I hit this in my repro testing

This PR is against the hotfix/2.2.0 branch.

/cc @brendandixon @jinhyunr


brendandixon:
@hglkrijger LGTM. Merge when it passes.

PR #499, opened 2016-11-30
Adds waagent config for bigip

caphrim007:
Includes minor tweaks for waagent code to ensure that it works
as expected on bigip platforms.

### DIFF
--- a/config/bigip/waagent.conf
+++ b/config/bigip/waagent.conf
@@ -1,0 +1,70 @@
+#
+# Windows Azure Linux Agent Configuration
+#
+
+# Specified program is invoked with the argument "Ready" when we report ready status
+# to the endpoint server.
+Role.StateConsumer=None
+
+# Specified program is invoked with XML file argument specifying role
+# configuration.
+Role.ConfigurationConsumer=None
+
+# Specified program is invoked with XML file argument specifying role topology.
+Role.TopologyConsumer=None
+
+# Enable instance creation
+Provisioning.Enabled=y
+
+# Password authentication for root account will be unavailable.
+Provisioning.DeleteRootPassword=y
+
+# Generate fresh host key pair.
+Provisioning.RegenerateSshHostKeyPair=y
+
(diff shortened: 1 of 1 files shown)


caphrim007:
@brendandixon ready for review

config/bigip/waagent.conf
```diff
@@ -0,0 +1,70 @@
...
+Lib.Dir=/shared/vadc/azure/waagent/
+
+# Specify location of sshd config file on BIG-IP
+OS.SshdConfigPath=/config/ssh/sshd_config
+
+# Disable RDMA management and set up
+OS.EnableRDMA=n
+
+# Enable or disable self-update
+AutoUpdate.Enabled=n
```
brendandixon: @caphrim007 We need to leave `AutoUpdate.Enable` set to `y`.

caphrim007: we can't have auto-update happen on bigip. it might update waagent to a version we have no tested

brendandixon: @caphrim007 Hmm, we're taking a pretty strong line with vendors that `AutoUpdate.Enabled` *must* be set to `y`. In the main, it gives Azure customers a better experience.

It's difficult for me to see how we could change the Agent to break on BigIP as the dependencies of the agent will not change.

caphrim007: well, the code refactor from 2.0.10 to 2.1.3 broke it and if there are new methods added that must be overridden by tmsh commands on our platform, that would break it too.

brendandixon: @caphrim007 I appreciate that. But that was because the team, from whom we inherited the code, did not so much as refactor as they rewrote. Much was untested before they deployed broadly. Our path forward is cleaner and includes much, much more rigorous testing.

caphrim007:
@brendandixon ready for review

caphrim007:
@brendandixon ping

PR #501, opened 2016-12-03
Fix resource disk partition format detection (#478)

jinhyunr:
sfdisk --part-type is not available for sfdisk version lower than 2.26 (i.e RHEL 6.8) 
In this case, sfdisk -c should be used instead.

### DIFF
--- a/azurelinuxagent/daemon/resourcedisk/default.py
+++ b/azurelinuxagent/daemon/resourcedisk/default.py
@@ -129,12 +129,12 @@ class ResourceDiskHandler(object):
                 shellutil.run(mkfs_string)
         else:
             logger.info("GPT not detected, determining filesystem")
-            ret = shellutil.run_get_output("sfdisk -q --part-type {0} 1".format(device))
+            ret = self.change_partition_type(suppress_message=True, option_str="{0} 1".format(device))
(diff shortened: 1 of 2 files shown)


azurelinuxagent/daemon/resourcedisk/default.py
```diff
@@ -174,6 +174,24 @@ def mount_resource_disk(self, mount_point, fs):
...
+            As --part-type is supported in sfdisk 2.26+,
+            use -c for lower versions
+        """
+
+        command_to_use = '--part-type'
+        # sfdisk version string is in the following formats:
+        # 1. "sfdisk from util-linux 2.27.1"
+        # 2. "sfdisk (util-linux-ng 2.17.2)"
+        err_code, version_str = shellutil.run_get_output("sfdisk -v")
+        if err_code == 0 and version_str.split()[-1].strip('(), ') < '2.26':
```
hglkrijger: I think if you are going to rely on this type of check you would need to ensure the version string matches first, as it might change in the future, so it might be better to do this kind of check with a regex.

hglkrijger:
@jinhyunr please take a look at the merge conflict as well

PR #503, opened 2016-12-05
We made a change in the kernel device, Agent should skip writing to hvnd device on module version 4.1 or newer

longlimsft:


azurelinuxagent/common/rdma.py
```diff
@@ -199,10 +199,24 @@ def start(self):
...
         try:
             RDMADeviceHandler.update_dat_conf(dapl_config_paths, self.ipv4_addr)
-            RDMADeviceHandler.wait_rdma_device(
-                self.rdma_dev, self.device_check_timeout_sec, self.device_check_interval_sec)
-            RDMADeviceHandler.write_rdma_config_to_device(
-                self.rdma_dev, self.ipv4_addr, self.mac_addr)
+
+            skip_rdma_device = False
+            retcode,out = shellutil.run_get_output("modinfo hv_network_direct")
+            if retcode == 0:
```
ahmetalpbalkan: what if `retcode != 0` ? should that throw exception?

bin/waagent2.0
```diff
@@ -3457,9 +3457,21 @@ class RdmaHandler(object):
         """
         Write config string to /dev/hvnd_rdma
         """
-        Log("Set /dev/hvnd_rdma")
-        self.wait_rdma_dev()
-        self.write_rdma_dev_conf()
+        skip_rdma_device = False
```
ahmetalpbalkan: I'm not sure why this file was edited. We didn't do any rdma changes to this code for a long time. Now we expect all the RDMA "supported" distros to be on 2.1.x or above, so let's not edit this file.

azurelinuxagent/common/rdma.py
```diff
@@ -199,10 +199,24 @@ def start(self):
...
+                version = re.search("version:\s+(\d+)\.(\d+)\.(\d+)\D", out, re.IGNORECASE)
+                if version:
```
ahmetalpbalkan: is there a valid case that `version:` won't be in `out`? if not should we throw or log an error and return for the `else` of this `if`?

azurelinuxagent/common/rdma.py
longlimsft: if retcode != 0, we just go through the normal behavior of writing to device file. The code will do the same thing before. Not sure if we want to give up and throw an exception here? In theory this modinfo command should never fail, unless something really bad happened on the system.

azurelinuxagent/common/rdma.py
longlimsft: It's possible version will not show up in older LIS releases, or in a build in kernel. Those are older versions of driver and we simply write to the device to initialize it.

bin/waagent2.0
longlimsft: So every time I do a "setup.py install", this file ended up getting installed to /usr/sbin/waagent2.0 on my test machine. I have to change it to make my system work. If this change is not necessary on 2.1.X, we can ignore this change.

azurelinuxagent/common/rdma.py
ahmetalpbalkan: Sounds good.

bin/waagent2.0
ahmetalpbalkan: @longlimsft I'd rather keep the file unchanged... If your machine runs waagent2.0, you're not testing the RDMA logic that's currently supported. Also this code is kept here for purposes other  than RDMA. That's why we can't really sign off a change to the old code that's not supported/used in RDMA scenarios.

If you're on ubuntu16.04 or something, use `python3 setup.py install`.

longlimsft: This change is mainly to support CentOS 6.x, 7.x and SUSE. I'm testing on CentOS 6.7 that's why I have to change this file. Is 2.1.x supposed to work on CentOS 6.7?

azurelinuxagent/common/rdma.py
longlimsft: How about I add a log for this error? The reason why I want the code to proceed is that this is minimal disruption to the older code logic.

ahmetalpbalkan: Yes, please add a log line so that when things go wrong, the on-call person can see that something is out of place and they can grep the error in the codebase.

bin/waagent2.0
ahmetalpbalkan: @hglkrijger any ideas about this question? ^

hglkrijger: @longlimsft 2.1.x is the only supported version across the board.

hglkrijger: Sorry I was not clear before - 2.2.1 (`/bin/waagent`) is the current version in master, originally based on the 2.1.x branch where it diverged from 2.0.x (`/bin/waagent2.0`). So you are working with the right version now.

longlimsft:
Will create a new pull request based on reviews.

PR #505, opened 2016-12-06
Host plugin updates

hglkrijger:
These changes allow the agent to function normally even while outbound connectivity to the public internet is blocked, via NSG for example. 

- enables host plugin downloads for extension and agent manifests and packages
- fixes an issue uploading the extension health via the host plugin
- fixes issues whereby the agent manifest and package are downloaded multiple times
- refactor and cleanup 
- adds unit tests

/cc @jinhyunr @brendandixon

### DIFF
--- a/azurelinuxagent/common/protocol/hostplugin.py
+++ b/azurelinuxagent/common/protocol/hostplugin.py
@@ -29,7 +29,8 @@ API_VERSION = "2015-09-01"
 HEADER_CONTAINER_ID = "x-ms-containerid"
 HEADER_VERSION = "x-ms-version"
 HEADER_HOST_CONFIG_NAME = "x-ms-host-config-name"
-
+HEADER_ARTIFACT_LOCATION = "x-ms-artifact-location"
+HEADER_ARTIFACT_MANIFEST_LOCATION = "x-ms-artifact-manifest-location"
 
 class HostPluginProtocol(object):
     def __init__(self, endpoint, container_id, role_config_name):
@@ -41,6 +42,7 @@ class HostPluginProtocol(object):
         self.endpoint = endpoint
         self.container_id = container_id
         self.role_config_name = role_config_name
+        self.manifest_uri = None
 
     def ensure_initialized(self):
         if not self.is_initialized:
@@ -52,21 +54,24 @@ class HostPluginProtocol(object):
     def get_api_versions(self):
         url = URI_FORMAT_GET_API_VERSIONS.format(self.endpoint,
                                                  HOST_PLUGIN_PORT)
-        logger.info("getting API versions at [{0}]".format(url))
+        logger.verbose("getting API versions at [{0}]".format(url))
+        return_val = []
         try:
-            headers = { HEADER_CONTAINER_ID: self.container_id }
+            headers = {HEADER_CONTAINER_ID: self.container_id}
             response = restutil.http_get(url, headers)
             if response.status != httpclient.OK:
                 logger.error(
                     "get API versions returned status code [{0}]".format(
                         response.status))
-                return []
-            return response.read()
+            else:
+                return_val = ustr(remove_bom(response.read()), encoding='utf-8')
+
         except HttpError as e:
             logger.error("get API versions failed with [{0}]".format(e))
-            return []
 
-    def get_extension_artifact_url_and_headers(self, artifact_url, artifact_manifest_url = None):
+        return return_val
+
+    def get_artifact_request(self, artifact_url, artifact_manifest_url=None):
         if not self.ensure_initialized():
             logger.error("host plugin channel is not available")
             return
@@ -74,20 +79,23 @@ class HostPluginProtocol(object):
             logger.error("no extension artifact url was provided")
             return
 
-        url = URI_FORMAT_GET_EXTENSION_ARTIFACT.format(self.endpoint, HOST_PLUGIN_PORT)
+        url = URI_FORMAT_GET_EXTENSION_ARTIFACT.format(self.endpoint,
+                                                       HOST_PLUGIN_PORT)
         headers = {HEADER_VERSION: API_VERSION,
                    HEADER_CONTAINER_ID: self.container_id,
                    HEADER_HOST_CONFIG_NAME: self.role_config_name,
-                   "x-ms-artifact-location": artifact_url}
+                   HEADER_ARTIFACT_LOCATION: artifact_url}
+
         if artifact_manifest_url is not None:
-            headers["x-ms-artifact-manifest-location"] = artifact_manifest_url
+            headers[HEADER_ARTIFACT_MANIFEST_LOCATION] = artifact_manifest_url
 
         return url, headers
 
-    def put_vm_status(self, status_blob, sas_url):
+    def put_vm_status(self, status_blob, sas_url, config_blob_type=None):
         """
         Try to upload the VM status via the host plugin /status channel
         :param sas_url: the blob SAS url to pass to the host plugin
+        :param config_blob_type: the blob type from the extension config
         :type status_blob: StatusBlob
         """
         if not self.ensure_initialized():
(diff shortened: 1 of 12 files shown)


azurelinuxagent/common/protocol/restapi.py
```diff
@@ -282,16 +282,16 @@ def get_ext_handlers(self):
     def get_ext_handler_pkgs(self, extension):
         raise NotImplementedError()
 
-    def get_in_vm_artifacts_profile(self):
+    def get_artifacts_profile(self):
```
jinhyunr: Not sure if in_vm_artifacts_profile equals to artifacts_profile.

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -96,26 +104,28 @@ def put_vm_status(self, status_blob, sas_url):
...
-                        {'headerName': 'x-ms-blob-type',
-                         'headerValue': status_blob.type}]
-        data = json.dumps({'requestUri': sas_url, 'headers': blob_headers,
-                           'content': status}, sort_keys=True)
         try:
-            response = restutil.http_put(url, data, headers)
+            url = URI_FORMAT_PUT_VM_STATUS.format(self.endpoint, HOST_PLUGIN_PORT)
+            logger.verbose("Posting VM status to host plugin")
+            status = textutil.b64encode(status_blob.data)
+            blob_type = status_blob.type if status_blob.type else config_blob_type
```
jinhyunr: I think it might be better to use get_blob_type() here and let the function handle this fallback.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -618,25 +628,42 @@ def call_storage_service(http_req, *args, **kwargs):
...
+                uri, headers = host.get_artifact_request(version.uri)
+                response = self.fetch(uri, headers)
+                if not response:
+                    logger.info("Retry fetch in {0} seconds",
+                                LONG_WAITING_INTERVAL)
+                    time.sleep(LONG_WAITING_INTERVAL)
+                else:
+                    host.manifest_uri = version.uri
+                    logger.verbose("Manifest downloaded successfully from host plugin")
+            if response:
```
jinhyunr: else?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -618,25 +628,42 @@ def call_storage_service(http_req, *args, **kwargs):
...
+            response = self.fetch(version.uri)
+            if not response:
+                logger.verbose("Manifest could not be downloaded, falling back to host plugin")
+                host = self.get_host_plugin()
+                uri, headers = host.get_artifact_request(version.uri)
+                response = self.fetch(uri, headers)
+                if not response:
+                    logger.info("Retry fetch in {0} seconds",
+                                LONG_WAITING_INTERVAL)
+                    time.sleep(LONG_WAITING_INTERVAL)
```
jinhyunr: How does the retry work here?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -802,8 +829,17 @@ def check_wire_protocol_version(self):
     def upload_status_blob(self):
         ext_conf = self.get_ext_conf()
         if ext_conf.status_upload_blob is not None:
-            if not self.status_blob.upload(ext_conf.status_upload_blob):
-                self.get_host_plugin().put_vm_status(self.status_blob, ext_conf.status_upload_blob)
+            uploaded = False
+            try:
+                uploaded = self.status_blob.upload(ext_conf.status_upload_blob)
+            except (HttpError, ProtocolError) as e:
+                # errors have already been logged
```
jinhyunr: If we let StatusBlob to use blobtype from extensionConfig as a fallback when it cannot retrieve one by making a request, status_blob.upload() should not throw HttpError/ProtocolError.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -923,52 +959,31 @@ def get_host_plugin(self):
...
+            logger.info("Getting the artifacts profile")
+            profile = self.fetch(blob)
+
+            if profile is None:
+                logger.warn("Download failed, falling back to host plugin")
+                host = self.get_host_plugin()
+                uri, headers = host.get_artifact_request(blob)
+                profile = self.decode_config(self.fetch(uri, headers))
+
+            if profile and not profile.isspace():
```
jinhyunr: use textutil.is_str_none_or_whitespace() instead?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1349,13 +1373,13 @@ class InVMArtifactsProfile(object):
...
     * encryptedHealthChecks (optional)
     * encryptedApplicationProfile (optional)
-    '''
+    """
 
-    def __init__(self, in_vm_artifacts_profile_json):
-        if not textutil.is_str_none_or_whitespace(in_vm_artifacts_profile_json):
-            self.__dict__.update(parse_json(in_vm_artifacts_profile_json))
+    def __init__(self, artifacts_profile):
+        if artifacts_profile and not artifacts_profile.isspace():
```
jinhyunr: any reason for reverting it?

azurelinuxagent/common/protocol/restapi.py
hglkrijger: Could you elaborate?

jinhyunr: I think there are just so many similar names which contains artifacts or profile(i.e. extension artifacts) so I am just worried that there might be another thing called 'artifact profile' different from 'in_vm_artifacts_profile'

azurelinuxagent/common/protocol/hostplugin.py
hglkrijger: I agree this could be improved, but I think the correct change is to remove `get_blob_type` since the `HEAD` call is not actually necessary anymore. For this PR I did not want to make more changes than necessary, but if you feel we should include that let me know.

azurelinuxagent/common/protocol/wire.py
hglkrijger: `fetch` calls `http_get` which has retries built-in.

azurelinuxagent/common/protocol/wire.py
hglkrijger: else we move on to the next uri, line 632.

azurelinuxagent/common/protocol/wire.py
hglkrijger: Even if we did that the actual upload would subsequently fail and still throw the protocol error.

azurelinuxagent/common/protocol/wire.py
hglkrijger: Sure, updated

azurelinuxagent/common/protocol/wire.py
hglkrijger: Nope, bad merge - thanks for catching.

azurelinuxagent/common/protocol/restapi.py
hglkrijger: We can always change it, but at the moment this is the only artifact profile. It just seems unnecessary to qualify it as `in_vm_`.

hglkrijger: Discussed offline, leaving as is.

azurelinuxagent/common/protocol/hostplugin.py
hglkrijger: Discussed offline, I am going to add an event to be emitted which we can use to determine how reliable the given value is before we switch over to it.

azurelinuxagent/common/protocol/wire.py
hglkrijger: Discussed offline

azurelinuxagent/common/protocol/wire.py
hglkrijger: Discussed offline

azurelinuxagent/common/protocol/wire.py
hglkrijger: Discussed offline

brendandixon:
@hglkrijger LGTM. Nice.

PR #508, opened 2016-12-08
Don't write IP/MAC to device running newer RDMA driver version 4.1 or…

longlimsft:
… newer


ahmetalpbalkan:
@longlimsft please sign up on aka.ms/azuregithub with your github & corp accounts for CLA validation.

PR #509, opened 2016-12-08
more logs to prevent unnecessary confusion in case of sfdisk --part-type failure (#478)

jinhyunr:


brendandixon:
lgtm

azurelinuxagent/daemon/resourcedisk/default.py
```diff
@@ -182,14 +182,16 @@ def change_partition_type(self, suppress_message, option_str):
...
+        err_code, output = shellutil.run_get_output(input, chk_err=False)
 
         # fall back to -c
         if err_code != 0:
+            logger.info("sfdisk with --part-type failed with error_code {0}. Retry with -c.", err_code)
             command_to_use = '-c'
             input = "sfdisk {0} {1} {2}".format(command_to_use, '-f' if suppress_message else '', option_str)
             err_code, output = shellutil.run_get_output(input)
 
+        logger.info("sfdisk completed with return code {0} and output {1}", err_code, output)
```
hglkrijger: I think we need a check here. If the return code is 0, logger.info("sfdisk completed successfully"), otherwise logger.error("...")

PR #542, opened 2017-01-10
ensure config file is updated correctly

### LINKED ISSUE #538: [2.2.0+]Fail to send hostname to DHCP server during provisioning in RHEL-7
Description of problem:
In RHEL-7, if NetworkManager is enabled, the hostname is not sent to DHCP server during provisioning, so that the VM internal DNS is not set to the DNS server.

WALinuxAgent-2.2.0+

RHEL Version:
RHEL-7.3
NetworkManager-1.4.0-12

How reproducible:
100%

Steps to Reproduce:
1. Create a RHEL-7.3 VM on Azure with WALA-2.2.0+ installed(NetworkManager enabled)
2. Check if send hostname to DHCP server:
~]# cat /var/lib/NetworkManager/dhclient-eth0.conf|grep send
There's no "send host-name" line.
3. Check DNS:
1). Get domain-name(like wala73220e.i6.internal.cloudapp.net): cat /etc/resolv.conf|grep search
2). Get hostname(like wala73220e): cat /etc/hostname
3). Check DNS:
~]# nslookup wala73220e.wala73220e.i6.internal.cloudapp.net
Server:                168.63.129.16
Address:        168.63.129.16#53

** server can't find wala73220e.wala73220e.i6.internal.cloudapp.net: NXDOMAIN

Actual Result:
The hostname is not sent to the DHCP server
The domain name is not added into DNS server

Expect Result:
The hostname is sent to the DHCP server, such as:
~]# cat /var/lib/NetworkManager/dhclient-eth0.conf |grep send
send host-name "wala73220e"; # added by NetworkManager

The domain name is added into DNS server, such as:
~]# nslookup wala73220e.wala73220e.i6.internal.cloudapp.net
Server:                168.63.129.16
Address:        168.63.129.16#53

Name:        wala73220e.wala73220e.i6.internal.cloudapp.net
Address: 10.0.0.5

Additional Info:
There're 2 root causes:

1. WALA-2.2.0 use "hostname" and write "/etc/hostname" file to set hostname, which cannot send hostname message into dbus. NetworkManager get hostname from dbus, then add "send host-name" into /var/lib/NetworkManager/dhclient-eth0.conf

2. WALA-2.2.0 write "DHCP_HOSTNAME=hostname" into /etc/sysconfig/network-scripts/ifcfg-eth0, but there's no '\n' after this line. So that NetworkManager cannot read it.

NetworkManager sends the hostname to the DHCP server using the following strategy:
- if the ifcfg file specifies a DHCP_HOSTNAME, that value is sent to the DHCP server

- if DHCP_HOSTNAME is not set, NM uses the system hostname set through hostnamectl, but only if it is not '(none)', 'localhost', 'localhost6' or 'localhost.localdomain'. If it is one of these, no hostname is sent

According to the NetworkManager developer suggestion, the best way to have NetworkManager send the hostname is to simply set a valid system hostname using 'hostnamectl set-hostname myhostname' and reactivate the connection with ifup.
The DHCP_HOSTNAME is only a way to override the system configured hostname.

So we suggest to:
1. use "hostnamectl set-hostname" command to set hostname in RHEL-7.
2. remove the "DHCP_HOSTNAME". If you need to use this parameter, please add '\n' after it.

hglkrijger:
In `fileutil.update_conf_file` we do not append a final newline, which can cause values to become unreadable. Also, when updating a value, we can potentially add multiple newline characters. This change addresses both cases.

- fixes #538 

/cc @brendandixon @jinhyunr

### DIFF
--- a/azurelinuxagent/common/utils/fileutil.py
+++ b/azurelinuxagent/common/utils/fileutil.py
@@ -140,7 +140,7 @@ def update_conf_file(path, line_start, val, chk_err=False):
     if not os.path.isfile(path) and chk_err:
         raise IOError("Can't find config file:{0}".format(path))
     conf = read_file(path).split('\n')
-    conf = [x for x in conf if not x.startswith(line_start) and x is not None and len(x) > 0]
+    conf = [x for x in conf if x is not None and len(x) > 0 and not x.startswith(line_start)]
     conf.append(val)
     write_file(path, '\n'.join(conf) + '\n')
 
--- a/azurelinuxagent/common/utils/fileutil.py
+++ b/azurelinuxagent/common/utils/fileutil.py
@@ -140,9 +140,9 @@ def update_conf_file(path, line_start, val, chk_err=False):
     if not os.path.isfile(path) and chk_err:
         raise IOError("Can't find config file:{0}".format(path))
     conf = read_file(path).split('\n')
-    conf = [x for x in conf if not x.startswith(line_start)]
+    conf = [x for x in conf if not x.startswith(line_start) and x is not None and len(x) > 0]
     conf.append(val)
-    write_file(path, '\n'.join(conf))
+    write_file(path, '\n'.join(conf) + '\n')
 
 def search_file(target_dir_name, target_file_name):
     for root, dirs, files in os.walk(target_dir_name):
--- a/tests/utils/test_file_util.py
+++ b/tests/utils/test_file_util.py
@@ -15,13 +15,12 @@
 # Requires Python 2.4+ and Openssl 1.0+
 #
 
-from tests.tools import *
 import uuid
-import unittest
-import os
-import sys
-from azurelinuxagent.common.future import ustr
+
 import azurelinuxagent.common.utils.fileutil as fileutil
+from azurelinuxagent.common.future import ustr
+from tests.tools import *
+
 
 class TestFileOperations(AgentTestCase):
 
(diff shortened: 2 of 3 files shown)


tests/utils/test_file_util.py
```diff
@@ -117,5 +116,66 @@ def test_get_all_files(self):
 
         self.assertEqual(set(expected_files), set(actual_files))
 
+    @patch('os.path.isfile')
+    def test_update_conf_file(self, _):
+        new_file = "\
+DEVICE=eth0\n\
```
jinhyunr: is this indentation intentional?

azurelinuxagent/common/utils/fileutil.py
```diff
@@ -140,9 +140,9 @@ def update_conf_file(path, line_start, val, chk_err=False):
     if not os.path.isfile(path) and chk_err:
         raise IOError("Can't find config file:{0}".format(path))
     conf = read_file(path).split('\n')
-    conf = [x for x in conf if not x.startswith(line_start)]
+    conf = [x for x in conf if not x.startswith(line_start) and x is not None and len(x) > 0]
```
jinhyunr: shouldn't 'x is not None' come before x.startswith() ?

tests/utils/test_file_util.py
hglkrijger: yes.. whitespace is not stripped when provided to the patch.

azurelinuxagent/common/utils/fileutil.py
hglkrijger: yep, thanks!

PR #543, opened 2017-01-10
support NetworkManager leases

### LINKED ISSUE #539: [2.2.0+]dhclient lease file is not removed during deprovisioning when using NetworkManager
Description of problem:
In RHEL-7, the dhclient lease file under /var/lib/NetworkManager/ folder is not removed during deprovisioning.

WALinuxAgent-2.2.0+

RHEL Version: 
RHEL-7.3

How reproducible:
100%

Steps to Reproduce:
1. Prepare a RHEL-7.3 VM with WALA-2.2.0+ installed(NetworkManager enabled)
2. Run "waagent -deprovision". It shows "WARNING! Cached DHCP leases will be deleted."
3. Check dhclient lease file:
ls /var/lib/NetworkManager/dhclient-*.lease

Actual Result:
The dhclient lease file is not removed.

Expect Result:
The dhclient lease file is removed.

Additional Info:
Currently WALA only remove all the files under /var/lib/dhclient/dhclient/. But if we use NetworkManager, the dhclient lease file is like this: /var/lib/NetworkManager/dhclient-5fb06bd0-0bb0-7ffb-45f1-d6edd65f3e03-eth0.lease

hglkrijger:
- fixes #539 

/cc @jinhyunr @brendandixon

### DIFF
--- a/azurelinuxagent/common/osutil/redhat.py
+++ b/azurelinuxagent/common/osutil/redhat.py
@@ -117,5 +117,11 @@ class RedhatOSUtil(Redhat6xOSUtil):
         DefaultOSUtil.openssl_to_openssh(self, input_file, output_file)
 
     def get_dhcp_lease_endpoint(self):
-        # centos7 has this weird naming with double hyphen like /var/lib/dhclient/dhclient--eth0.lease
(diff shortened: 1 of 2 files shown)


jinhyunr:
LGTM

PR #545, opened 2017-01-11
Support offline hostname update

### LINKED ISSUE #540: [2.2.0+] Cannot monitor hostname when self-updating
Hi,

When wala is self-updating(or restarting, or any scenario that the run-exthandler process is not running), if I change hostname, then when the run-exthandler process is running again, the new hostname cannot be sent to DHCP server. 
It seems that when the run-exthandler process start, first it set self.hostname as current hostname in the env.py, then it begins to monitor the hostname changes. So if the hostname is changed before the run-exthandler process starting, the hostname will not be sent to the DHCP server.

Do we support this scenario? Thanks!

WALA version: 2.2.0+
Reproduce steps:
1. Enable AutoUpdate, restart waagent service
2. When the run-exthandler process is exit, change hostname
~]# hostnamectl set-hostname newhostname
3. Wait for the run-exthandler process started. When it started, check DNS:
~]# nslookup newhostname.domain-name

hglkrijger:
- fixes #540 

/cc @jinhyunr @brendandixon

### DIFF
--- a/azurelinuxagent/common/conf.py
+++ b/azurelinuxagent/common/conf.py
@@ -98,6 +98,10 @@ def get_lib_dir(conf=__conf__):
     return conf.get("Lib.Dir", "/var/lib/waagent")
 
 
+def get_published_hostname(conf=__conf__):
+    return os.path.join(get_lib_dir(conf), 'published_hostname')
+
+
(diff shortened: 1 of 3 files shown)


brendandixon:
litmus

Issue #532, opened 2017-01-06
DISTRO_ID should be used instead of DISTRO_NAME


crawford:
Right now, a lot of the functionality switches implementations based on a mangled version of the [NAME](https://www.freedesktop.org/software/systemd/man/os-release.html#NAME=) field from `/etc/os-release`. This isn't safe to do because this field is intended for consumption by the user. The correct field to use is the [ID](https://www.freedesktop.org/software/systemd/man/os-release.html#ID=) field. This is actually meant for consumption by scripts and OS vendors will make an effort not to screw with this too often.

hglkrijger:
So we took a closer look at `platform.linux_distribution` and the `DISTRIB_ID` regex is what populates the `DISTRO_NAME` field. So while the nomenclature is not right in the Python libs, we are already using the `ID` field from `/etc/os-release`. You can see that we already match values like `coreos` and `ubuntu` here: https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/common/osutil/factory.py#L54

crawford:
This isn't correct. I don't see where `platform.linux_distribution` uses the `DISTRIB_ID` regex (looking at https://hg.python.org/cpython/file/2.7/Lib/platform.py), but either way, because Container Linux is not in the list of supported distributions, `full_distribution_name=0` has no effect and the full name will be returned.  [`get_distro()`](https://github.com/Azure/WALinuxAgent/blob/764b1ba4ae2a4a01cc9af9408ed972b461e25e8d/azurelinuxagent/common/version.py#L59) is what determines the `DISTRO_NAME` and it is using the first element in the tuple returned by [`platform.linux_distribution()`](https://docs.python.org/2/library/platform.html#platform.linux_distribution), while I think it should instead be using the third. Using the short name is an okay workaround at first glance, but it is not equivalent to the ID. For reference, this is what python returned on Container Linux before we introduced our workaround for this issue:

```
>>> linux_distribution()
('Container Linux by CoreOS', '1284.2.0', 'coreos')
>>> linux_distribution(full_distribution_name=0)
('Container Linux by CoreOS', '1284.2.0', 'coreos')
```

PR #548, opened 2017-01-13
Use DISTRO_ID instead of DISTRO_NAME (#532)

jinhyunr:


azurelinuxagent/common/osutil/factory.py
```diff
@@ -32,11 +32,11 @@
 from .alpine import AlpineOSUtil
 from .bigip import BigIpOSUtil
 
-def get_osutil(distro_name=DISTRO_NAME, distro_version=DISTRO_VERSION,
-               distro_full_name=DISTRO_FULL_NAME):
-    if distro_name == "clear linux software for intel architecture":
+
+def get_osutil(distro_id=DISTRO_CODE_NAME, distro_version=DISTRO_VERSION, distro_full_name=DISTRO_FULL_NAME):
+    if distro_id == "clear linux software for intel architecture":
```
hglkrijger: we need to determine what these values are before we can switch over

azurelinuxagent/daemon/main.py
```diff
@@ -57,7 +57,7 @@ def __init__(self):
 
     def run(self):
         logger.info("{0} Version:{1}", AGENT_LONG_NAME, AGENT_VERSION)
-        logger.info("OS: {0} {1}", DISTRO_NAME, DISTRO_VERSION)
+        logger.info("OS: {0} {1}", DISTRO_CODE_NAME, DISTRO_VERSION)
```
hglkrijger: we should log both

hglkrijger:
closing since we already use distro_id: see #532

PR #551, opened 2017-01-13
Add support for Match blocks in sshd config

### LINKED ISSUE #544: agent breaks ssh configuration
The agent indiscriminately adds

 PasswordAuthentication no
ChallengeResponseAuthentication no

to the end of sshd_config. However, if the configuration file happens to contain a "Match" block at the end then the added configuration options become part of the match block creating a syntax error in the configuration file and thus preventing sshd from running. This results in a running instance that the user cannot access.

hglkrijger:
We do not support `Match` blocks in sshd config, and if they occur anywhere but the end of the file, we can potentially break the config rules. In addition, we do not have unit tests for conf_sshd. These ensure the correct behavior.

- fixes #544 

/cc @jinhyunr @brendandixon

### DIFF
--- a/azurelinuxagent/common/utils/textutil.py
+++ b/azurelinuxagent/common/utils/textutil.py
@@ -229,8 +229,8 @@ def set_ssh_config(config, name, val):
         if config[i].startswith(name) and match_start == no_match:
             config[i] = "{0} {1}".format(name, val)
             found = True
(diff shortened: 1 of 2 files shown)


brendandixon:
lgtm

hglkrijger:
@brendandixon - note I pushed some additional changes, if you want to take a look.

PR #552, opened 2017-01-14
Instead of adding values in sysinfo directly to event, if a value

### LINKED ISSUE #520: Ensure events only contain VM info from the agent
Extensions may write events and, in some case, may include VM information values (such as ContainerID) they should not. This can result in events containing duplicate PARAMs which may confuse off-board telemetry collectors.

`monitor.py#collect_and_send_events` should, after parsing the event, should purge from the event any fields it intends to add from its `sysinfo` object.

jinhyunr:
already exist in the event, overwrite it to prevent duplicates.

This fixes #520.

### DIFF
--- a/azurelinuxagent/ga/monitor.py
+++ b/azurelinuxagent/ga/monitor.py
@@ -162,7 +162,7 @@ class MonitorHandler(object):
 
             try:
                 event = parse_event(data_str)
-                event.parameters.extend(self.sysinfo)
+                self.add_sysinfo(event)
                 event_list.events.append(event)
             except (ValueError, ProtocolError) as e:
                 logger.warn("Failed to decode event file: {0}", e)
@@ -193,3 +193,13 @@ class MonitorHandler(object):
             except Exception as e:
                 logger.warn("Failed to send events: {0}", e)
             time.sleep(60)
+
+    def add_sysinfo(self, event):
+        event_param_dict = {}  # key: TelemetryEventParam.name, val: idx of the EventParam in event.parameters
+        for i in range(len(event.parameters)):
+            event_param_dict[event.parameters[i].name] = i
+        for e in self.sysinfo:
+            if e.name in event_param_dict:
+                event.parameters[event_param_dict[e.name]] = e
+            else:
+                event.parameters.append(e)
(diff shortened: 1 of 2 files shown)


azurelinuxagent/ga/monitor.py
```diff
@@ -193,3 +193,13 @@ def daemon(self):
             except Exception as e:
                 logger.warn("Failed to send events: {0}", e)
             time.sleep(60)
+
+    def add_sysinfo(self, event):
```
jinhyunr: O(n) time by wasting space :p

brendandixon: @jinhyunr Worrying about performance is more than a little premature.

azurelinuxagent/ga/monitor.py
```diff
@@ -193,3 +193,13 @@ def daemon(self):
...
+        event_param_dict = {}  # key: TelemetryEventParam.name, val: idx of the EventParam in event.parameters
+        for i in range(len(event.parameters)):
```
brendandixon: @jinhyunr "value" vs. "val" and "index" vs. "idx". In comments we must be clear...

azurelinuxagent/ga/monitor.py
```diff
@@ -193,3 +193,13 @@ def daemon(self):
...
+            event_param_dict[event.parameters[i].name] = i
+        for e in self.sysinfo:
+            if e.name in event_param_dict:
+                event.parameters[event_param_dict[e.name]] = e
+            else:
+                event.parameters.append(e)
```
brendandixon: @jinhyunr This feels complex to me. Why not keep it cleaner and simpler: Purge values from `event` that occur in `sysinfo` and then extend the `event` with `sysinfo`. Such logic is harder to get wrong (it entails fewer decisions and, hence, fewer places to be incorrect). Performance is certainly not a concern: We should always code the simplest, correct solution first and then, as driven by data, tweak for performance. (Architectural performance is another matter.)

azurelinuxagent/ga/monitor.py
```diff
@@ -193,3 +193,10 @@ def daemon(self):
             except Exception as e:
                 logger.warn("Failed to send events: {0}", e)
             time.sleep(60)
+
+    def add_sysinfo(self, event):
+        sysinfo_names = [v.name for v in self.sysinfo]
+        for param in event.parameters:
+            if param.name in sysinfo_names:
+                event.parameters.remove(param)
```
hglkrijger: we might want some debug logging here, to follow up with extension owners at some point.

PR #553, opened 2017-01-16
Typo is log message

boumenot:
prefered -> preferred


brendandixon:
@boumenot A man after my own heart. Thank you. LGTM.

PR #560, opened 2017-01-19
Add telemetry event for host plugin usage

### LINKED ISSUE #557: Enrich Provision Event with Additional Telemetry
Please consider including in the Provision telemetry event the way in which the status blob was updated, e.g. public vs. host.

hglkrijger:
We need some way of determining how frequently the host plugin is being used without sending telemetry on every request. This change adds a new event when we initialize the host plugin, which happens only on usage, and only once (until restart). We can fine tune as needed.

- add new telemetry event
- minor cleanup and reformatting
- fixes #557 

/cc @boumenot @brendandixon @jinhyunr


brendandixon:
lgtm

PR #567, opened 2017-01-21
CoreOS fix

### LINKED ISSUE #532: DISTRO_ID should be used instead of DISTRO_NAME

hglkrijger:
- minor cleanup
- add `distro_code_name` which is populated from the `ID` field
- detect CoreOS via `distro_code_name`
- fixes #532 

/cc @jinhyunr @brendandixon @crawford

### DIFF
--- a/azurelinuxagent/common/osutil/factory.py
+++ b/azurelinuxagent/common/osutil/factory.py
@@ -17,9 +17,7 @@
 
 import azurelinuxagent.common.logger as logger
 from azurelinuxagent.common.utils.textutil import Version
-from azurelinuxagent.common.version import DISTRO_NAME, DISTRO_VERSION, \
-                                     DISTRO_FULL_NAME
-
+from azurelinuxagent.common.version import *
 from .default import DefaultOSUtil
 from .clearlinux import ClearLinuxUtil
 from .coreos import CoreOSUtil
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/osutil/factory.py
```diff
@@ -17,64 +17,73 @@
...
     elif distro_name == "freebsd":
         return FreeBSDOSUtil()
+
     elif distro_name == "bigip":
         return BigIpOSUtil()
+
     else:
-        logger.warn("Unable to load distro implementation for {0}.", distro_name)
-        logger.warn("Use default distro implementation instead.")
+        logger.warn("Unable to load distro implementation for {0}. Using "
```
jinhyunr: Does this work? I've seen it failing sometimes and had to use %s instead.. :(

azurelinuxagent/common/osutil/factory.py
```diff
@@ -17,64 +17,73 @@
...
         return CoreOSUtil()
+
     if distro_name == "suse":
-        if distro_full_name=='SUSE Linux Enterprise Server' and \
-           Version(distro_version) < Version('12') or \
-           distro_full_name == 'openSUSE' and \
-           Version(distro_version) < Version('13.2'):
+        if distro_full_name == 'SUSE Linux Enterprise Server' \
+                and Version(distro_version) < Version('12') \
+                or distro_full_name == 'openSUSE' and Version(distro_version) < Version('13.2'):
```
jinhyunr: Any reason for this format?

hglkrijger: No, just easier to read.

azurelinuxagent/common/osutil/factory.py
hglkrijger: Yes, our logging class supports this behavior. Note the usage is unchanged, just combining the logs into one line.

PR #577, opened 2017-01-27
Ignore return code 32

### LINKED ISSUE #571: Partition sync
The bug from #183 has resurfaced, we need to investigate:
```
(000005)Could not mount /dev/sdc1 after syncing partition table: 32	WALinuxAgent-2.2.0	debian	8.6
(000005)Could not mount /dev/sdc1 after syncing partition table: 32	WALinuxAgent-2.2.0	debian	8.7
```

hglkrijger:
When a disk is already mounted, we might receive return code 32. We can ignore this without raising an error.

- fixes #571 

/cc @jinhyunr @brendandixon

### DIFF
--- a/azurelinuxagent/daemon/resourcedisk/default.py
+++ b/azurelinuxagent/daemon/resourcedisk/default.py
@@ -152,8 +152,11 @@ class ResourceDiskHandler(object):
                                              partition,
                                              mount_point)
         logger.info("Mount resource disk [{0}]", mount_string)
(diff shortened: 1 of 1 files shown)


takekazuomi:
This is great fix.

I had resource disk mount problem in CentOS 7.3. 

```
uname -r 
3.10.0-514.6.2.el7.x86_64
```

In this kernel version, mount command return 32 with "special device /dev/sdc1 does not exist" message. waagent 2.2.5 dose not work.

I'm waiting new release of waagent.

thanks

PR #581, opened 2017-01-30
Extension name in status file set by a extension publisher is not honored

### LINKED ISSUE #580: Extension name a publisher sets in status file is not honored
A publisher can set extension name in extension status file according to Extension publisher doc:

> status\name – This property is optional. This property can be used by handlers to point to the VM workload name that are being managed by the handler.

However, the current version of the agent does not honor the name; rather it sets the field with extension handler name.

jinhyunr:
Check if status file contains extension name.
a. If it does, honor the name
b. If it does not, use extension handler name instead

- fixes #580 

Added UT.

### DIFF
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -274,14 +274,16 @@ def ext_substatus_to_v1(sub_status_list):
     return status_list
 
 
-def ext_status_to_v1(ext_name, ext_status):
+def ext_status_to_v1(ext_handler_name, ext_status):
     if ext_status is None:
         return None
     timestamp = time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime())
     v1_sub_status = ext_substatus_to_v1(ext_status.substatusList)
+    if not ext_status.name:
+        ext_status.name = ext_handler_name
     v1_ext_status = {
         "status": {
-            "name": ext_name,
+            "name": ext_status.name,
             "configurationAppliedTime": ext_status.configurationAppliedTime,
             "operation": ext_status.operation,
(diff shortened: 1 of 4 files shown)


azurelinuxagent/common/protocol/wire.py
```diff
@@ -274,14 +274,16 @@ def ext_substatus_to_v1(sub_status_list):
...
 
 
-def ext_status_to_v1(ext_name, ext_status):
+def ext_status_to_v1(ext_handler_name, ext_status):
     if ext_status is None:
         return None
     timestamp = time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime())
     v1_sub_status = ext_substatus_to_v1(ext_status.substatusList)
+    if not ext_status.name:
+        ext_status.name = ext_handler_name
```
hglkrijger: I would be a bit concerned about changing the behavior like this. Specifically:
- it has been this way for quite a while; are we breaking more people by reverting to the older behavior?
- I am not a fan of overloading a single field; let's pass back both `ext_status.name` and `ext_handler_name`
- perhaps we can handle both points above by keeping `name` the same and passing back a new field, say `extensionname` or similar

@brendandixon - what do you think?

brendandixon: @jinhyunr @hglkrijger The document is very squishy regarding this field. It is an open field for use by the extension to convey the "workload":

_status\name_ – This property is optional. This property can be used by handlers to point to the VM workload name that are being managed by the handler.

The examples does not an extension name in any form.

I suggest that we review the Windows agent on this and align behavior since that is our long-term goal.

hglkrijger:
We can address this in a future release. Closing for now.

PR #587, opened 2017-02-09
Fix cache-incarnation sync

### LINKED ISSUE #585: waagent cannot handle hostname change and dhcp client restart
We are facing an issue where waagent fails in specific regions ( mostly US-west) and works in other ( East -Asia etc)

## Symptom
The failed VMs show this message in waagent.log

````
2017/02/07 00:50:46.059925 WARNING Exception retrieving agent manifests: (000008)(000009)HTTP GET failed
2017/02/07 00:50:46.060142 ERROR Event: name=WALinuxAgent, op=Download, message=Exception retrieving agent manifests: (000008)(000009)HTTP GET failed
2017/02/07 00:50:46.061256 INFO Wire server endpoint:168.63.129.16
2017/02/07 00:50:46.062024 WARNING Exception retrieving extension handlers: (000008)(000009)HTTP GET failed
2017/02/07 00:50:46.062175 ERROR Event: name=WALinuxAgent, op=, message=Exception retrieving extension handlers: (000008)(000009)HTTP GET failed
2017/02/07 00:50:49.347846 INFO EnvMonitor: Detected dhcp client restart. Restoring routing table.
2017/02/07 00:50:49.348137 INFO Configure routes
2017/02/07 00:50:49.348611 INFO Gateway:None
2017/02/07 00:50:49.348788 INFO Routes:None
2017/02/07 00:51:11.101978 WARNING Exception retrieving extension handlers: (000008)/var/lib/waagent/ExtensionsConfig.2.xml is missing.
2017/02/07 00:51:11.106286 ERROR Event: name=WALinuxAgent, op=, message=Exception retrieving extension handlers: (000008)/var/lib/waagent/ExtensionsConfig.2.xml is missing.
````

## System details
````
# waagent -version
WALinuxAgent-2.2.2 running on centos 6.8
Python: 2.6.6
Goal state agent: 2.2.2
````

## More details
It appears that in every instance of failed waagent start , there is a dhcp client restart and hostname change

````
2017/02/07 00:50:45.029505 INFO EnvMonitor: Detected hostname change: None -> bktestcluster-masternode
...
2017/02/07 00:50:49.347846 INFO EnvMonitor: Detected dhcp client restart. Restoring routing table.
````

Attached is a good log and a bad log for you to compare
[good-waagent.txt](https://github.com/Azure/WALinuxAgent/files/756746/good-waagent.txt)
[bad-waagent.txt](https://github.com/Azure/WALinuxAgent/files/756745/bad-waagent.txt)

## Our solution template 
Link to our solution template is here
https://github.com/striim/azure-solution-template

hglkrijger:
Recently we added code to store the published hostname on disk, in order to detect a hostname change when the agent is not running. When the agent upgrades at runtime across this boundary, we can initiate a hostname change immediately after provisioning. This resets the network adapter, which can cause the manifest or goalstate download to fail. 

In some cases, particularly when installing an extension at provision time, this leads to a state where the agent has recorded the `Incarnation` as `n`, but `ExtensionsConfig.{n}.xml` does not yet exist. 

This PR contains several updates:

- prevent the network adapter from resetting, and causing the initial failure
- make sure the incarnation is only updated after the cache is stored, in case networking fails for another reason
- reduce the http verbose logging, so debugging is a little easier
- in the case of NSG, ensure we force update the goal state
- update the version to 2.2.5 for publishing
- fixes #585 

/cc @brendandixon @jinhyunr

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -749,9 +749,16 @@ class DefaultOSUtil(object):
         fileutil.write_file(conf.get_published_hostname(), contents=hostname)
 
     def get_hostname_record(self):
-        record = None
-        if os.path.exists(conf.get_published_hostname()):
-            record = fileutil.read_file(conf.get_published_hostname())
+        hostname_record = conf.get_published_hostname()
+        if not os.path.exists(hostname_record):
+            # this file is created at provisioning time with agents >= 2.2.3
+            hostname = socket.gethostname()
+            logger.warn('Hostname record does not exist, '
+                        'creating [{0}] with hostname [{1}]',
+                        hostname_record,
+                        hostname)
+            self.set_hostname_record(hostname)
+        record = fileutil.read_file(hostname_record)
         return record
 
     def del_account(self, username):
@@ -765,7 +772,7 @@ class DefaultOSUtil(object):
         return base64.b64decode(data).decode('utf-8')
 
     def get_total_mem(self):
-        # Get total memory in bytes and divide by 1024**2 to get the valu in MB.
+        # Get total memory in bytes and divide by 1024**2 to get the value in MB.
         return os.sysconf('SC_PAGE_SIZE') * os.sysconf('SC_PHYS_PAGES') / (1024**2)
 
     def get_processor_cores(self):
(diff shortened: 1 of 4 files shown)


brendandixon:
LGTM

Issue #586, opened 2017-02-09
Enable support for including other config files in waagent.conf


OddBloke:
When building Ubuntu images, we want to be able to ship a config file in the walinuxagent package without auto-update enabled, but enable auto-update within the image build process.

To avoid conflicts on upgrade, this would need to be done in a separate config file, so we would like to be able to include config files within waagent.conf (probably from /etc/waagent.conf.d/*, but it should be configurable).

OddBloke:
(We're going to look at submitting a PR for this from Canonical's side.)

sil2100:
Related to this: would everyone here be fine if in Ubuntu we, by using this 'include' functionality, add the /etc/waagent.conf.d/ directory as an official additional configuration directory to our packaging?

brendandixon:
@OddBloke @sil2100 I have mixed feelings about it. Most of the configuration settings are not that useful; they remain *mostly* for historical reasons. However, mis-setting them could cause customers problems. I *strongly* prefer that customers not alter configuration settings, except the HTTP proxy settings, unless they have a very clear, well-understood reason for doing so. Therefore, having an extensible mechanism, while nice in one sense, to me seems to open the door to more possible errors.

sil2100:
I understand your concern completely. But the solution proposed by OddBloke seems a safe thing to do to get us unblocked from the Ubuntu POV. This way there would be tools in place that aren't enabled by default (like, there would be no new directory for configuration for everyone), but can allow us to add it only through our packaging.

Of course, we can also think about just hard-coding an additional directory for overrides like these instead (we did discuss that possibility in our team as well as it could work-around some of our other problems), but not sure if that would change the situation.

sil2100:
Is there anything we can do to move this forward? Personally this (or hard-coding an additional configuration path) would be perfect for us as this way we could finally safely (from the formal POV) enable auto-updates. The release of 2.2.4 on Ubuntu is blocked on having this resolved as our plan was to include the auto-update change along with the release.

elyscape:
@brendandixon:
>I _strongly_ prefer that customers not alter configuration settings, except the HTTP proxy settings, unless they have a very clear, well-understood reason for doing so.

[...]

PR #590, opened 2017-02-13
Add support for splitting configurations into separate files

sil2100:
Add a new config option 'include' that allows to include config options from a separate file or directory. Add a test suite for the conf part. Issue: #586 .

### DIFF
--- a/azurelinuxagent/common/conf.py
+++ b/azurelinuxagent/common/conf.py
@@ -49,7 +49,7 @@ class ConfigurationProvider(object):
                 if not os.path.exists(path):
                     raise AgentConfigError(
                         "Included configuration path does "
-                        "not exist: {}".format(path))
+                        "not exist: {0}".format(path))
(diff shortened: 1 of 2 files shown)


sil2100:
This is a very simple approach. It might be good to add some checks for infinite include loops in configuration files.

sil2100:
This is now good. Seems like I got used to Python3 too much.

hglkrijger:
@sil2100 - could you sign the CLA?

msftgits:
Hi, I am closing and re-opening this PR to bump the CLA bot. Sorry for the inconvenience!

hglkrijger:
@brendandixon @jinhyunr any other feedback?

hglkrijger:
@sil2100 I believe we came to the conclusion on our call last week that this was not needed.

sil2100:
Yes, although we still think it would be a valuable addition. But it's not blocking us anymore, yes.

PR #592, opened 2017-02-14
Fix a reboot issue when updating RDMA driver on ND host update

longlimsft:


hglkrijger:
@longlimsft could you please link your corp account?

shirgall:
Is this blocked on a paperwork issue?

hglkrijger:
@shirgall the CLA is one thing, we also need to test the change.

@rjschwei - could you take a look as well?

shirgall:
@hglkrijger The LIS team is set up to test the change, and we are happy to help.

hglkrijger:
@shirgall @longlimsft our test run passed, please sort out the CLA while we wait for @rjschwei to review as well.

azurelinuxagent/pa/rdma/suse.py
```diff
@@ -93,8 +93,7 @@ def install_driver(self):
                 msg = 'RDMA: Successfully installed "%s" from '
                 msg += 'configured repositories'
                 logger.info(msg % complete_name)
-                self.load_driver_module()
-                if requires_reboot:
+                if not self.load_driver_module() or requires_reboot:
```
rjschwei: This will lead to an endless reboot loop and the user cannot even diagnose what is wrong. and no bug can be filed If for whatever reason the driver will not load.

Lets say we make a mistake and ens up with an API incompatibility in the package which would cause modprobe to fail. Rebooting the system is certainly not going to help the situation. On reboot we'll try again and then again on the next reboot.....

azurelinuxagent/common/rdma.py
```diff
@@ -133,14 +133,14 @@ def load_driver_module(self):
         """Load the kernel driver, this depends on the proper driver
            to be installed with the install_driver() method"""
         logger.info("RDMA: probing module '%s'" % self.driver_module_name)
-        result = shellutil.run('modprobe %s' % self.driver_module_name)
+        result = shellutil.run('modprobe --first-time %s' % self.driver_module_name)
```
longlimsft: load_driver_module is only used when installing/updating drivers (install_rdma_drivers). If the RDMA driver is already loaded, it's wrong that we don't do anything, because the newly installed driver is not loaded at this point.

azurelinuxagent/pa/rdma/suse.py
longlimsft: It will not trigger the next reboot, because the driver package is already before the next reboot. On the next reboot, we'll simply return not doing anything:

                if fw_version in version and installed == 'i':
                    info_msg = 'RDMA: Matching driver package "%s-%s" '
                    info_msg += 'is already installed, nothing to do.'
                    logger.info(info_msg % (package_name, version))
                    return True

longlimsft: Actual I mean the 2nd reboot (not the next reboot: 1st reboot). The 1st reboot is needed to properly load (or attempt to load) the driver. The 2nd reboot will not happen because the driver package is already installed.

longlimsft:
I'll create a new pull request. Close this one.

PR #594, opened 2017-02-24
Fix a reboot issue when updating RDMA driver on ND host update

longlimsft:


szarkos:
Bump?

hglkrijger:
@rjschwei - any further comments on this?

rjschwei:
My gut still tells me that using "--first-time" can/will get us into a situation where we "break" the system unnecessarily for the user, meaning an error condition is generated when we should not generate such an error condition. However, given that I have no argumentative/logical proof of this and it is not fair to block code changes based on gut-feel I'll have to live with the nagging feeling. We can always fix it when the issue shows up.

brendandixon:
@rjschwei Fair point. But that sounds like a rare (?) case and, after the reboot, all will be good, no?

In any case, LGTM.

rjschwei:
@brendandixon yes, probably rare. Not sure if the reboot we eliminate the need for manual intervention.

PR #603, opened 2017-03-03
retry get thumbprint if key is deleted

### LINKED ISSUE #591: Timing Issue with cloud-init and host keys
Under some circumstances, provisioning can fail waiting for cloud-init to regenerate SSH host keys:
```
2017/02/08 23:03:23.712479 INFO Wait for ssh host key to be generated.
2017/02/08 23:03:23.756726 ERROR run cmd 'ssh-keygen -lf /etc/ssh/ssh_host_rsa_key.pub' failed
2017/02/08 23:03:23.762243 ERROR Error Code:255
2017/02/08 23:03:23.766393 ERROR Result:ssh-keygen: /etc/ssh/ssh_host_rsa_key.pub: No such file or directory
 
2017/02/08 23:03:23.775621 ERROR Provision failed: (000004)**Failed to generate ssh host key: ret=255, out= ssh-keygen: /etc/ssh/ssh_host_rsa_key.pub: No such file or directory**
```
Waagent checks the thumbprint of the host rsa pub file to validate that the key is regenerated. If the file doesn't exist, it waits up to 5 minutes for it to be generated. Based on the logs, the sequence of events seems to be

1. Waagent confirms existence of the file
2. Cloud-init deletes the file and begins generating a new file
3. Waagent tries and fails to get the thumbprint because the file doesn't exist

One or both of the following should be implemented:
1. Use more caution when exiting from this function with error. If the thumbprint operation finds that the file doesn't exist (when it was just confirmed to exist by other means), we should try again.
2. If the expectation is that the SSH host keys are deleted before provisioning, then waagent -deprovision should delete them.

hglkrijger:
When cloud-init provisions the vm in Ubuntu, we wait for the ssh key to be present on disk. It is possible to hit a race condition where we fail to get the thumbprint before the key is regenerated. If this happens we should retry.

- fixes #591 

/cc @brendandixon @jinhyunr

### DIFF
--- a/azurelinuxagent/pa/provision/ubuntu.py
+++ b/azurelinuxagent/pa/provision/ubuntu.py
@@ -89,8 +89,14 @@ class UbuntuProvisionHandler(ProvisionHandler):
         path = '/etc/ssh/ssh_host_{0}_key.pub'.format(keypair_type)
         for retry in range(0, max_retry):
             if os.path.isfile(path):
-                return self.get_ssh_host_key_thumbprint(keypair_type)
+                logger.info("ssh host key found at: {0}".format(path))
+                try:
+                    thumbprint = self.get_ssh_host_key_thumbprint(keypair_type)
+                    logger.info("Thumbprint obtained from : {0}".format(path))
+                    return thumbprint
+                except ProvisionError:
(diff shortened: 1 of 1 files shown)


brendandixon:
LGTM

PR #624, opened 2017-03-24
Data disks udev rule

paulmey:
Adds matching rules for all scsi disks other than root/resource disk
Data disks will show up as /dev/disks/azure/datadisks/lun*, matching the LUN's specified on attach/deploy

Also fixed the test-out matching, since vendor and model were completely ignored

### DIFF
--- a/config/66-azure-storage.rules
+++ b/config/66-azure-storage.rules
@@ -6,8 +6,17 @@ LABEL="azure_disk"
 # The resource/resource has GUID of 0001 as the second value
 ATTRS{device_id}=="?00000000-0000-*", ENV{fabric_name}="root", GOTO="azure_names"
 ATTRS{device_id}=="?00000000-0001-*", ENV{fabric_name}="resource", GOTO="azure_names"
-ENV{DEVTYPE}=="partition", PROGRAM="/bin/sh -c 'readlink /sys/class/block/%k/../device|cut -d: -f4'", ENV{fabric_name}="datadisks/lun$result", GOTO="azure_names"
-PROGRAM="/bin/sh -c 'readlink /sys/class/block/%k/device|cut -d: -f4'", ENV{fabric_name}="datadisks/lun$result", GOTO="azure_names"
+# Wellknown SCSI controllers
+ATTRS{device_id}=="{f8b3781a-1e82-4818-a1c3-63d806ec15bb}", ENV{fabric_scsi_controller}="scsi0", GOTO="azure_datadisk"
+ATTRS{device_id}=="{f8b3781b-1e82-4818-a1c3-63d806ec15bb}", ENV{fabric_scsi_controller}="scsi1", GOTO="azure_datadisk"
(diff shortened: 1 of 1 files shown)


paulmey:
Don't merge yet, we're still testing on all distros...

szarkos:
I think the use of $ID_PATH is not going to be very consistent across distributions.  In some distributions we do not see ID_PATH configured, or the formatting varies.  We might want to revert to using something like "/sys/block/%k/device" as you were in an earlier version.

Running "udevadm info --query=property --name=/dev/sdc | grep ID_PATH" I see the following results on various endorsed distros:

SLES 11:
    ID_PATH=vmbus-f8b3781b:1e82:4818:a1c3:63d806ec15bb-lun-0

CentOS 6:
    ID_PATH=scsi-0:0:0:0

CentOS 7:
    <undefined>

Ubuntu 14.04:
    <undefined>

Ubuntu 16.04:
    ID_PATH=acpi-VMBUS:01-scsi-0:0:0:0

The lack of ID_PATH or any by-path links in CentOS 7 seems to be a bug and related to this issue: https://access.redhat.com/solutions/1565863

config/66-azure-storage.rules
```diff
@@ -1,12 +1,12 @@
...
-ATTRS{ID_MODEL}!="Virtual_Disk", GOTO="azure_end"
+ACTION=="add|change", SUBSYSTEM=="block", ENV{ID_VENDOR}=="Msft", ENV{ID_MODEL}=="Virtual_Disk", GOTO="azure_disk"
+GOTO="azure_end"
 
+LABEL="azure_disk"
 # Root has a GUID of 0000 as the second value
 # The resource/resource has GUID of 0001 as the second value
 ATTRS{device_id}=="?00000000-0000-*", ENV{fabric_name}="root", GOTO="azure_names"
 ATTRS{device_id}=="?00000000-0001-*", ENV{fabric_name}="resource", GOTO="azure_names"
+PROGRAM="/bin/sh -c 'echo $ID_PATH|cut -d: -f5'", ENV{fabric_name}="datadisks/lun$result", GOTO="azure_names"
```
szarkos: The following seems to work more consistently across distros and versions:

KERNEL=="sr*", GOTO="azure_end"
PROGRAM="/bin/sh -c 'readlink /sys/block/%k/device|cut -d: -f4'", ENV{fabric_name}="datadisks/lun-$result", GOTO="azure_names"

paulmey:
That will need a bit of work to also work with partitions... I'll be back

paulmey:
@szarkos check it out. I've changed it so it now works for:

- CentOS-6-8
- CentOS-7-3
- Debian-8
- openSUSE-Leap-42-2
- RHEL-6-8
- RHEL-7-3
- SLES-11-SP4
- SLES-12-SP2
- UbuntuServer-12-04-5-LTS
- UbuntuServer-14-04-5-LTS
- UbuntuServer-16-04-LTS

paulmey:
Okay, one more change coming. People want us to prepare for usage of multiple controllers.

PR #634, opened 2017-03-29
Improve RDMA package version matching

szarkos:
The pattern matching for the RPM naming is too strict:
  - We actually have a 4.1.3.2 bugfix release now, but pattern only allows 3-digit naming
  - We may have multi-digit version numbers in the future


szarkos:
@longlimsft can you please look at this before we merge?

longlimsft:
Looking good.

szarkos:
@hglkrijger since Long approves I think we can merge this at your leasure.

Thanks,
Steve

PR #635, opened 2017-03-29
Pass environment variables to subprocesses

baude:
In cases where subprocesses are used to call python processes
it would be helpful to pass on the environment variables used for
the agent itself.  In particular, this is helpful for running
extensions from the agent in system containers.


hglkrijger:
@baude, would you mind signing the CLA?

PR #638, opened 2017-03-30
NSG improvements

### LINKED ISSUE #630: Host plugin fallback should be sticky
If we detect the connection to storage is not available, we should keep this state and continue using the host plugin.

hglkrijger:
- ensure use of host plugin channel is sticky
- additional logging
- unit test updates
- fixes #630 

/cc @brendandixon @jinhyunr

### DIFF
--- a/azurelinuxagent/common/protocol/hostplugin.py
+++ b/azurelinuxagent/common/protocol/hostplugin.py
@@ -38,6 +38,8 @@ MAXIMUM_PAGEBLOB_PAGE_SIZE = 4 * 1024 * 1024  # Max page size: 4MB
 
 
 class HostPluginProtocol(object):
+    _is_default_channel = False
+
     def __init__(self, endpoint, container_id, role_config_name):
         if endpoint is None:
             raise ProtocolError("HostGAPlugin: Endpoint not provided")
@@ -49,6 +51,14 @@ class HostPluginProtocol(object):
         self.deployment_id = None
         self.role_config_name = role_config_name
         self.manifest_uri = None
+
+    @staticmethod
+    def is_default_channel():
+        return HostPluginProtocol._is_default_channel
+
+    @staticmethod
+    def set_default_channel(is_default):
+        HostPluginProtocol._is_default_channel = is_default
 
     def ensure_initialized(self):
         if not self.is_initialized:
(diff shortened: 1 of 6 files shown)


azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -50,6 +52,14 @@ def __init__(self, endpoint, container_id, role_config_name):
         self.role_config_name = role_config_name
         self.manifest_uri = None
 
+    @staticmethod
+    def is_default_channel():
+        return HostPluginProtocol._is_default_channel
+
+    @staticmethod
+    def set_default_channel(is_default):
```
jinhyunr: If we want to set this to True all the time(other than UT cases), do we need the argument?

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -266,8 +276,13 @@ def _base64_encode(self, data):
...
         return s
     
-    def _read_response_error(self, response):
+    @staticmethod
+    def read_response_error(response):
+        if response is None:
+            return ''
         body = remove_bom(response.read())
+        if body is None:
+            return ''
```
brendandixon: This change loses the status / reason when the body is absent. Do we want that? It seems they may be useful for other errors.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -649,6 +657,9 @@ def fetch_manifest(self, version_uris):
                 else:
                     host.manifest_uri = version.uri
                     logger.verbose("Manifest downloaded successfully from host plugin")
+                    if not HostPluginProtocol.is_default_channel():
+                        logger.info("Setting host plugin as default channel")
+                        HostPluginProtocol.set_default_channel(True)
```
brendandixon: You may want to push this into the HostPluginProtocol: It should become the default channel any time it is used and succeeds. Once default, it stays default, even if subsequent errors occur.

azurelinuxagent/common/protocol/hostplugin.py
hglkrijger: Not at the moment, but we do need it for UTs, and we may want to switch this flag off in the future.

azurelinuxagent/common/protocol/hostplugin.py
hglkrijger: Thanks, will update

azurelinuxagent/common/protocol/wire.py
hglkrijger: This is ok for get_versions and put_vm_log (which we are not using yet) but the complication is that we use a common _fetch or _download outside of host plugin and simply update the url with the get_artifact_request call. At the moment that is the primary scenario too, so we cannot reliably detect host plugin usage from within HostPluginProtocol.

Issue #641, opened 2017-03-31
Heartbeat key should be checked first


hglkrijger:
We should check if the heartbeat key exists before referencing it.

hglkrijger:
fix merged

Issue #639, opened 2017-03-31
Agent should perform case-insensitive extension state comparisons


brendandixon:
The case of the strings at exthandler.py#152-155 disagrees with checks made at exthandler.py#226-231. We should make case-insensitive or case-consistent checks.

Issue #637, opened 2017-03-30
Event cleanup


hglkrijger:
Investigate options for cleaning up old events, in the case an older agent crashed and left behind excess events on disk.

hglkrijger:
fix merged

PR #642, opened 2017-03-31
Bug fixes

### LINKED ISSUE #640: Agent crash should log traceback
When the extension handler crashes we get a generic 'agent crashed', we should log a traceback here.

hglkrijger:
- additional logging for errors
- handle heartbeat key errors correctly
- handle extension state checks case-insensitively
- cleanup older events, keeping the latest 1000
- adds unit tests for event rollover
- fixes #640, #641, #639, #637

### DIFF
--- a/tests/common/test_event.py
+++ b/tests/common/test_event.py
@@ -68,7 +68,7 @@ class TestEvent(AgentTestCase):
         init_event_logger(tmp_evt)
 
         for i in range(0, 2000):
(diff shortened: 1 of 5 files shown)


tests/common/test_event.py
```diff
@@ -0,0 +1,90 @@
+# Copyright 2014 Microsoft Corporation
```
jinhyunr: 2017..? :p

PR #644, opened 2017-04-03
add basic 32 bit support to interface detection

### LINKED ISSUE #636: WALinuxAgent 32bit support
I understand Linux 32bit isn't official supported, but if possible could we make a small change to help the support.

In /common/osutil/default.py  there is this
  def get_first_if(self)
       struct_size=40

In /common/osutil/bigip.py there is this
        python_arc = platform.architecture()[0]
        if python_arc == '64bit':
            struct_size = 40  # for 64bit the size is 40 bytes
        else:
            struct_size = 32  # for 32bit the size is 32 bytes

If default.py had the same code as bigip.py then it runs fine on 32 bit.

hglkrijger:
- fixes #636

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -18,6 +18,7 @@
 
 import multiprocessing
 import os
+import platform
 import re
 import shutil
 import socket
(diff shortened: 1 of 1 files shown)


azurelinuxagent/common/osutil/default.py
```diff
@@ -420,7 +421,12 @@ def get_first_if(self):
         """
         iface=''
         expected=16 # how many devices should I expect...
-        struct_size=40 # for 64bit the size is 40 bytes
+
+        # for 64bit the size is 40 bytes
+        # for 32bit the size is 32 bytes
+        python_arc = platform.architecture()[0]
+        struct_size = 40 if python_arc == '64bit' else 32
```
jinhyunr: do we know if this architecture str format will stay the same as '64bit' ?

PR #663, opened 2017-04-11
reset channel if host plugin put status fails

hglkrijger:
There are two known cases where using host plugin may fail today, but particularly in multi-nic vms we need to protect against the case where a transient issue leads to a non-nsg vm switching over to this channel permanently. I put this safety valve in the put vm status method only, because potentially valid failures in the download channel could regress the actual nsg scenarios.

### DIFF
--- a/azurelinuxagent/common/protocol/hostplugin.py
+++ b/azurelinuxagent/common/protocol/hostplugin.py
@@ -183,6 +183,8 @@ class HostPluginProtocol(object):
             report_event(op=WALAEventOperation.ReportStatus,
                          is_success=False,
                          message=message)
+            logger.warn("HostGAPlugin: resetting default channel")
+            HostPluginProtocol.set_default_channel(False)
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -183,6 +183,8 @@ def put_vm_status(self, status_blob, sas_url, config_blob_type=None):
             report_event(op=WALAEventOperation.ReportStatus,
                          is_success=False,
                          message=message)
+            logger.warn("HostGAPlugin: resetting default channel")
+            HostPluginProtocol.set_default_channel(False)
```
brendandixon: @hglkrijger Do we need similar logic where the code downloads artifacts?

hglkrijger: Chatted offline

Issue #653, opened 2017-04-06
Please support distros managed by purely systemd, e.g. Archlinux.


enihcam:
Please support distros managed by systemd, e.g. Archlinux.

Concretely, there are three comments:

1. Before falling back to default, please detect if the distro is managed by systemd.
`
2017/04/06 15:57:46.294102 WARNING Unable to load distro implemetation for arch.
2017/04/06 15:57:46.312828 WARNING Use default distro implemetation instead.
2017/04/06 15:57:46.314404 INFO No RDMA handler exists for distro='arch' version=''
`

2. Archlinux does not have debian tools like ifup and ifdown.
`
2017/04/06 15:58:24.812463 VERBOSE run cmd 'ifdown eth0 && ifup eth0'
2017/04/06 15:58:24.825142 ERROR run cmd 'ifdown eth0 && ifup eth0' failed
2017/04/06 15:58:24.834340 ERROR Error Code:127
2017/04/06 15:58:24.841365 ERROR Result:/bin/sh: ifdown: command not found
`

3. DHCP is managed by systemd in Archlinux.
`
2017/04/06 15:58:49.945059 WARNING Dhcp client is not running. 
2017/04/06 15:58:49.949842 VERBOSE run cmd 'pidof dhclient'
2017/04/06 15:58:49.968002 ERROR run cmd 'pidof dhclient' failed
2017/04/06 15:58:49.968590 ERROR Error Code:1
2017/04/06 15:58:49.970555 ERROR Result:
`

Thanks!

Attached detailed log:
[Archlinux.b49e9e4a-8f0f-4eec-82ed-78b47de3c7b5.serialconsole.zip](https://github.com/Azure/WALinuxAgent/files/901971/Archlinux.b49e9e4a-8f0f-4eec-82ed-78b47de3c7b5.serialconsole.zip)

enihcam:
I found that WALinuxAgent 'supports' Archlinux, but it just does not work. :(

hglkrijger:
@enihcam we are happy to accept a PR to add this support, but have no plans to add this in the near term.

enihcam:
@hglkrijger Can I send a pull-request to fix this? 😄

hglkrijger:
@enihcam yes, please do.

PR #120, opened 2015-10-29
Support arch distro

karataliu:
notes:
1. python2 has some issue in platform.dist() for detecting arch (py3 works), adding supported_dists explicitly;
2. Current default service manager is based on systemd;
3. python version was default to python3, updated the init script cmd here.

PR #670, opened 2017-04-15
Add Arch distro support

enihcam:
Add Arch distro support. This fixes the issues #653 and #120 .

Looking forward to official Arch images in Azure!

www.archlinux.org


config/arch/waagent.conf
```diff
@@ -0,0 +1,100 @@
...
+# Extension.LogDir=/var/log/azure
+
+#
+# Home.Dir=/home
+
+# Enable RDMA management and set up, should only be used in HPC images
+# OS.EnableRDMA=y
+
+# Enable or disable goal state processing auto-update, default is enabled
+AutoUpdate.Enabled=n
```
hglkrijger: I recommend you enable this

bin/waagent2.0
```diff
@@ -1117,7 +1117,148 @@ class oracleDistro(redhatDistro):
     def __init__(self):
         super(oracleDistro, self).__init__()
 
+############################################################
```
hglkrijger: The 2.0 library should not need these changes? Typically we do not update this code any more since this version of the agent is no longer supported.

enihcam:
@hglkrijger Thanks for code review. I fixed it according to your suggestion.

PR #685, opened 2017-04-29
Change option order for mkfs. Alpine is being picky

jgreat:


brendandixon:
@jgreat LGTM. Merging.

Issue #687, opened 2017-05-01
Enable customer-supported cloud-init


brendandixon:
We should add a configuration option that allows customers to use cloud-init. While we would not support their provisioning, the switch would tell the agent to pause before retrieving goal state (much like we do for Ubuntu today) until cloud-init completes.

Issue #604, opened 2017-03-05
v2.2.x - Need new OS.SshDir in waagent.conf


u6006:
Need this new key to make waagent work in customized Linux VM with ssh conf-dir in a different directory from '/etc/ssh/'. As far as I can tell, three things need to be done:

- add in the new OS.SshDir to waagent.conf and glue logic to conf.py
- revert/modify the hidden OS.SshdConfigPath to comply with OS.SshDir
- correct all hardcoded '/etc/ssh/*' in code

u6006:
My bad. s/waagent.log/waagent.conf/. Darn the Saturday night beer.

Issue #531, opened 2017-01-05
Location of configuration file should be editable


baude:
The location of the waagent.conf file should be able to be changed at runtime.  In other words, allow to override /etc/waagent.conf.  This will help with the containerization and more advanced usage of the agent.

brendandixon:
This will entail moving to use argparse and adding additional switches.

PR #692, opened 2017-05-02
Addressed issues #687, #604, and #531

brendandixon:
Address issues #687, #604, and #531.


brendandixon:
@hglkrijger Not sure. This will preserve the Ubuntu behavior and lay the groundwork for the next distribution. Issues will come out in testing / design.

PR #695, opened 2017-05-03
Ensure all udev rules are copied in Ubuntu

hglkrijger:
The storage rules which were added recently are not copied during setup in Ubuntu.


brendandixon:
@hglkrijger LGTM

Issue #701, opened 2017-05-04
Reduce polling time for extension processing


hglkrijger:
The current proposal is ~3sec

brendandixon:
Addressed by #702

Issue #631, opened 2017-03-28
Will "waagent -deprovision[+user]" wipe the cloud-init data?


yuxisun1217:
Hi there,

It seems that the cloud-init is going to be involved in the RHEL images. Do you think that the "waagent -deprovision[+user]" should wipe the data that generated by cloud-init during provisioning? Because if the cloud-init data is not cleaned, the cloud-init will not provision during the next boot.
Thanks!

Currently we found some cloud-init data like this:
1. /etc/sudoers.d/90-cloud-init-users
2. All the files in /var/lib/cloud/

hglkrijger:
@yuxisun1217, I confirmed that cloud-init will still provision, since it uses the VM unique id, which ends up in `/var/lib/cloud/instances/*`. We should still cleanup old data which may be sensitive though, so I will leave this issue open as a feature request.

yuxisun1217:
@hglkrijger, Sorry for the wrong understanding of this feature and thank you for your confirming.  :)

brendandixon:
@hglkrijger Are these safe to delete (e.g., `shutil.rmtree('/var/lib/cloud')`)?

hglkrijger:
@brendandixon I don't think we should remove the root directory, but selectively under it, such as:
```
/var/lib/cloud/instance
/var/lib/cloud/instances/*
/var/lib/cloud/data/*
```

brendandixon:
@paulmey Any comment? You know the `cloud-init` code better than the rest of us.

paulmey:
from `/usr/share/doc/cloud-init/var-lib-cloud.txt`:

> to clear out the current instance's data as if to force a "new run" on reboot
> do:
>   ( cd /var/lib/cloud/instance && sudo rm -Rf * )

but it's safe to delete all of `/var/lib/cloud`. It's not installed by the deb and `stages.py` recursively recreates the whole layout if it doesn't exist.

brendandixon:
Addressed by #698

cmelanson:
@brendandixon We are storing scripts in /var/lib/cloud/scripts/per-instance which are executed on the first boot, as we customize the image for our own purposes. This change breaks this functionality for us, and I believe this is a common use case. Would it be possible to not clear out /var/lib/cloud/scripts during this new cleanup procedure or make it configurable?

brendandixon:
@cmelanson Happy to adjust it. I'll alter the code to remove the minimum needed.

brendandixon:
[...]

PR #702, opened 2017-05-08
Added Provisioning telemetry duration

brendandixon:
Correctly handle cleaning directories that include symlinks
Reduced polling interval #701
Enable selective agent testing
Reduced cloud-init files removed on deprovision #631

### DIFF
--- a/azurelinuxagent/common/utils/fileutil.py
+++ b/azurelinuxagent/common/utils/fileutil.py
@@ -119,16 +119,20 @@ def rm_files(*args):
 
 def rm_dirs(*args):
     """
-    Remove all the contents under the directry
+    Remove the contents of each directry
     """
-    for dir_name in args:
-        if os.path.isdir(dir_name):
-            for item in os.listdir(dir_name):
-                path = os.path.join(dir_name, item)
-                if os.path.isfile(path):
-                    os.remove(path)
-                elif os.path.isdir(path):
-                    shutil.rmtree(path)
+    for p in args:
+        if not os.path.isdir(p):
+            continue
+
+        for pp in os.listdir(p):
+            path = os.path.join(p, pp)
+            if os.path.isfile(path):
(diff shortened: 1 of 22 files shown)


azurelinuxagent/common/utils/fileutil.py
```diff
@@ -119,16 +119,20 @@ def rm_files(*args):
...
+            continue
+
+        for pp in os.listdir(p):
+            path = os.path.join(p, pp)
+            if os.path.isfile(path):
+                os.remove(path)
+            elif os.path.islink(path):
+                os.unlink(path)
+            elif os.path.isdir(path):
+                shutil.rmtree(path)
```
hglkrijger: Won't this still fail if a sub-directory contains a symlink?

azurelinuxagent/ga/update.py
```diff
@@ -918,3 +958,61 @@ def __str__(self):
...
+        d = ','.join(platform.linux_distribution())
+        for v in self.versions:
+            if re.match(v, d):
+                return True
+        return False
+
+    @property
+    def in_slice(self):
+        n = int((60 * self.slice) / 100)
+        return (n - datetime.utcnow().second) > 0
```
hglkrijger: should this be `>=` ?

azurelinuxagent/common/utils/fileutil.py
brendandixon: @hglkrijger Nope. It only fails if the directory passed to `shutil.rmtree` is a symlink.

azurelinuxagent/ga/update.py
brendandixon: @hglkrijger Nope. Seconds range from 0-59 (that is, it is zero-based). Consider the 100% case: n will be 60...which will work for 0-59. :)

Issue #706, opened 2017-05-09
Provision Event is not Fired if Customer Forgets to Deprovision VM


boumenot:
Scenario:

 1. Customer deploys a gallery image.
 1. Agent sends a provision event, and touches /var/lib/waagent/provisioned.
 1. Customer turns off VM.
 1. Customer creates a new VM using the VHD from (1).
 1. Agent sees that /var/lib/waagent/provisioned exists, and skips further processing.

As a result of the agent skipping further processing, a provision event is never sent, and the customer sees a OS Provisioning Timeout.  It would be better if the agent took into account the unique ID of the VM, and observed that if this value changes it should send a provision event.  The provision should include a message to the customer indicating that they failed to deprovision the VM.

In the case of a cloud-init enabled VM, cloud-init will reprovision the VM.

brendandixon:
Addressed in #707

PR #707, opened 2017-05-10
Emit event if reprovisioning without a deprovision (#706)

brendandixon:
Emit even if reprovisioning without a deprovision (#706)
Correct enabling test agents
Blacklist agent due to any failure

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -30,13 +30,15 @@ import fcntl
 import base64
 import glob
 import datetime
+
 import azurelinuxagent.common.logger as logger
 import azurelinuxagent.common.conf as conf
-from azurelinuxagent.common.exception import OSUtilError
-from azurelinuxagent.common.future import ustr
 import azurelinuxagent.common.utils.fileutil as fileutil
 import azurelinuxagent.common.utils.shellutil as shellutil
 import azurelinuxagent.common.utils.textutil as textutil
+
+from azurelinuxagent.common.exception import OSUtilError
+from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.utils.cryptutil import CryptUtil
 
 __RULES_FILES__ = [ "/lib/udev/rules.d/75-persistent-net-generator.rules",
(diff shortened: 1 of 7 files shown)


azurelinuxagent/pa/provision/cloudinit.py
```diff
@@ -34,6 +36,10 @@
 from azurelinuxagent.pa.provision.default import ProvisionHandler
 
 
+PRODUCT_ID_FILE = '/sys/class/dmi/id/product_uuid'
+DMIDECODE_CMD = 'dmidecode | grep UUID'
```
hglkrijger: You might prefer `dmidecode --string system-uuid`, no dependency on grep

azurelinuxagent/common/osutil/default.py
```diff
@@ -58,6 +61,25 @@ def __init__(self):
...
+        Azure records a UUID as the instance ID
+        First check /sys/class/dmi/id/product_uuid.
+        If that is missing, then extracts from dmidecode
+        If nothing works (for old VMs), return the empty string
+        '''
+        if os.path.isfile(PRODUCT_ID_FILE):
+            return fileutil.read_file(PRODUCT_ID_FILE).strip()
+
+        rc, s = shellutil.run_get_output(DMIDECODE_CMD)
+        if rc != 0:
```
hglkrijger: Probably worth checking for 'Not Settable' as well, in case this is not supported; in that case the return code will still be 0.

brendandixon: @hglkrijger The following check, which splits the string, should then fail and return an empty string. Or am I missing something?

azurelinuxagent/pa/provision/cloudinit.py
brendandixon: @hglkrijger Thanks.

azurelinuxagent/common/osutil/default.py
brendandixon: @hglkrijger I adjusted the code to look for a UUID and reject it if the match fails.

PR #712, opened 2017-05-12
Add initial support for the OpenBSD operating system

reyk:
There are some differences between FreeBSD, OpenBSD, and Linux.

Notes:

- OpenBSD ships LibreSSL, but WALinuxAgent needs OpenSSL for CMS, so
use the openssl port and the "eopenssl" binary instead.

- Don't run the custom DHCP client but parse /var/db/dhclient.leases.hvn0.
OpenBSD's lease file uses a modified syntax.

- OpenBSD does not have /proc.  WALinuxAgent should never assume that
the /proc filesystem is available.

- OpenBSD does not have sudo, but its replacement doas.  The OpenBSD
class implements support for modifying the doas.conf file.

- Unlike FreeBSD, OpenBSD supports and mounts UDF DVDs just fine.

- Create a swap partition instead of a swap file on the resource disk.

- Many other minor changes for OpenBSD

TODO:

- The /proc checks needs to be replaced with pgrep/ps etc. checks for
OpenBSD.  For now it just checks if /proc is available or returns
without error.
Make sure to install waagent with --register-service as setuptools
will not set the permissions of /etc/rc.d/waagent correctly (the file
has to be executable to be used by OpenBSD's rc system).

### DIFF
--- a/azurelinuxagent/common/osutil/openbsd.py
+++ b/azurelinuxagent/common/osutil/openbsd.py
@@ -109,7 +109,7 @@ class OpenBSDOSUtil(DefaultOSUtil):
         if ret != 0:
             raise OSUtilError(("Failed to encrypt password for {0}: {1}"
                                "").format(username, output))
-        passwd_hash = output
+        passwd_hash = output.strip()
         cmd = "usermod -p '{0}' {1}".format(passwd_hash, username)
         ret, output = shellutil.run_get_output(cmd, log_cmd=False)
         if ret != 0:
--- a/azurelinuxagent/common/osutil/openbsd.py
+++ b/azurelinuxagent/common/osutil/openbsd.py
@@ -251,7 +251,6 @@ class OpenBSDOSUtil(DefaultOSUtil):
                 return
             if retry < max_retry - 1:
                 mountlist = shellutil.run_get_output("/sbin/mount")[1]
-                logger.info("mountlist: '{0}'", mountlist)
                 existing = self.get_mount_point(mountlist, dvd_device)
                 if existing is not None:
                     logger.info("{0} is mounted at {1}", dvd_device, existing)
--- a/azurelinuxagent/common/osutil/openbsd.py
+++ b/azurelinuxagent/common/osutil/openbsd.py
@@ -1,0 +1,342 @@
+# Microsoft Azure Linux Agent
+#
+# Copyright 2014 Microsoft Corporation
+# Copyright 2017 Reyk Floeter <reyk@openbsd.org>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Requires Python 2.4+ and OpenSSL 1.0+
+
+import os
+import re
+import time
+import glob
+import datetime
+
+import azurelinuxagent.common.utils.fileutil as fileutil
+import azurelinuxagent.common.utils.shellutil as shellutil
+import azurelinuxagent.common.utils.textutil as textutil
+import azurelinuxagent.common.logger as logger
+import azurelinuxagent.common.conf as conf
+
+from azurelinuxagent.common.exception import OSUtilError
+from azurelinuxagent.common.osutil.default import DefaultOSUtil
+from azurelinuxagent.common.future import ustr
+
+UUID_PATTERN = re.compile(
+    '^\s*[A-F0-9]{8}(?:\-[A-F0-9]{4}){3}\-[A-F0-9]{12}\s*$',
+    re.IGNORECASE)
+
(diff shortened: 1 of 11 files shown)


azurelinuxagent/common/osutil/openbsd.py
```diff
@@ -0,0 +1,342 @@
...
+
+        for retry in range(0, max_retry):
+            retcode = self.mount(dvd_device, mount_point, option="-o ro -t udf",
+                                 chk_err=chk_err)
+            if retcode == 0:
+                logger.info("Successfully mounted DVD")
+                return
+            if retry < max_retry - 1:
+                mountlist = shellutil.run_get_output("/sbin/mount")[1]
+                logger.info("mountlist: '{0}'", mountlist)
```
reyk: This is a debug message that should be removed.

lordgordon:
Hi,

thank you for your work, really appreciated. Do you think your changes are able to make WALinuxAgent working fine with LibreSSL on a FreeBSD?

Thanks

reyk:
@lordgordon: As LibreSSL comes from OpenBSD, I looked into using it directly. But we have to use the OpenSSL port instead because LibreSSL is not compatible with WALinuxAgent (pkg_add openssl; binary is installed as "eopenssl" on OpenBSD) .

Details: the Azure wire protocol uses CMS to encrypt the certificates payload that includes the SSH public keys. I don't know why public keys have to be encrypted this way, but the agent needs "openssl cms" to decrypt them. But CMS has been removed from LibreSSL because it was poor and complex code and not even enabled in OpenSSL for a long time (http://marc.info/?l=openbsd-cvs&m=147301176512716&w=2). 

@mbelop and me managed to port CMS as a stand-alone binary that links with LibreSSL, but it is a big and ugly port that needs to provide its own versions for many internal functions within libcrypto. I can try to release it separately, but the best solution for OpenBSD and maybe FreeBSD is to use upstream OpenSSL for WALinuxAgent. I know that kind of hurts when you try to move away from using it.

I also wonder why Azure doesn't just include the SSH public keys as Values in the ovf-env.xml file; this way we wouldn't need CMS.

lordgordon:
@reyk thank you very much for your detailed explanation. I agree, the real issue here is Azure's forcing to use CMS.

As I mainly need LibreSSL for nginx I think the easiest workaround is to install LibreSSL with nginx in a jail.

KylieLiang:
@hglkrijger Could you please review the patch from Reyk? They develop the patch to enable waagent for OpenBSD.

config/openbsd/waagent.conf
```diff
@@ -0,0 +1,105 @@
...
+# Extension.LogDir=/var/log/azure
+
+#
+# Home.Dir=/home
+
+# Enable RDMA management and set up, should only be used in HPC images
+# OS.EnableRDMA=y
+
+# Enable or disable goal state processing auto-update, default is enabled
+AutoUpdate.Enabled=n
```
hglkrijger: I recommend leaving this enabled

azurelinuxagent/pa/deprovision/default.py
```diff
@@ -95,6 +104,9 @@ def del_dhcp_lease(self, warnings, actions):
         actions.append(DeprovisionAction(fileutil.rm_files, ["/var/db/dhclient.leases.hn0",
                                                              "/var/lib/NetworkManager/dhclient-*.lease"]))
 
+        # For OpenBSD
+        actions.append(DeprovisionAction(fileutil.rm_files, ["/var/db/dhclient.leases.hvn0"]))
```
hglkrijger: Should this have a wildcard?

azurelinuxagent/common/version.py
```diff
@@ -141,7 +144,10 @@ def is_agent_path(path):
 
 def set_goal_state_agent():
     agent = None
-    pids = [pid for pid in os.listdir('/proc') if pid.isdigit()]
+    if os.path.isdir("/proc"):
+        pids = [pid for pid in os.listdir('/proc') if pid.isdigit()]
+    else:
+        pids = []
```
hglkrijger: What is the behavior of `waagent --version` here?

config/openbsd/waagent.conf
reyk: Sure, I can enable it again. I turned it off because it wouldn't work before the merge.

azurelinuxagent/pa/deprovision/default.py
reyk: I modeled it after the FreeBSD one above that only deletes hn0 (I don't know what NetworkManager is). But, yes, that should be OK to remove /var/db/dhclient.leases.hvn* in case that there was any additional configured NIC.

azurelinuxagent/common/version.py
reyk: WALinuxAgent-2.2.11 running on openbsd 6.1
    Python: 2.7.13
    Goal state agent: 2.2.11

Could you introduce a new osutil callback that allows OS without `/proc` to implement their own method to get a running process? On OpenBSD, we could implement a callback with the `pgrep` system utility. There are two places that could use it instead of accessing `/proc` directly: `set_goal_state_agent()` in `common/version.py` and `validate_cloud_init()` in `pa/provision/default.py`.

reyk: @hglkrijger I rebased and force-pushed an update based your recommendations, see the individual commits for the changes. Can we get it merged now?

hglkrijger: @reyk, sure, that sounds reasonable. Would you please open an issue to track?

PR #752, opened 2017-06-07
Handle 404 in AzureStack

### LINKED ISSUE #741: ProtocolError not caught
`HttpError` is rethrown as `ProtocolError` in `metadata.py#report_event`. This should be caught to prevent unhandled exceptions.

hglkrijger:
- do not raise a ProtocolError when reporting an event fails in metadata protocol
- fixes #741


brendandixon:
lgtm

PR #753, opened 2017-06-07
remove admin access requirement from AzureStack

### LINKED ISSUE #744: Remove ip tables rules
The root restricted iptables rules in metadata.py are not necessary. We should remove them to avoid confusion.

hglkrijger:
- fixes #744


brendandixon:
lgtm

PR #762, opened 2017-06-15
fix the issue of "dmidecode is not found" on FreeBSD

ccic:
On FreeBSD, the ''dmidecode --string system-uuid" cannot execute correctly, because FreeBSD init rc script has to set PATH by itself. Currently PATH only contains "/usr/local/bin", unfortunately, dmidecode locates in "/usr/local/sbin".

### DIFF
--- a/init/freebsd/waagent
+++ b/init/freebsd/waagent
@@ -6,7 +6,7 @@
 
 . /etc/rc.subr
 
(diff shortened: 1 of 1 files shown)


ccic:
Is there any concern on this fix?

brendandixon:
@ccic Nope. We just had not gotten around to looking at it.

PR #768, opened 2017-06-19
Report ready if provisioned by external source

### LINKED ISSUE #756: [2.2.12] VM cannot become running if Provisioning.Enabled=n and Provisioning.UseCloudInit=n
Hi,

When Provisioning.Enabled=n and Provisioning.UseCloudInit=n, the VM cannot become running status. (The same as #411)
I wonder if we support the scenario that customer use neither WALA nor cloud-init to provision there VM. If we support this, I think we'd better report ready status as before.

Best regards,
Yuxin Sun

jayden-at-arista:
fixes #756

### DIFF
--- a/azurelinuxagent/pa/provision/default.py
+++ b/azurelinuxagent/pa/provision/default.py
@@ -53,14 +53,20 @@ class ProvisionHandler(object):
         self.protocol_util = get_protocol_util()
 
     def run(self):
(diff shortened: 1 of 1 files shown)


jayden-at-arista:
Reporting ready was integrated into the external provisioning agent, WALinuxAgent change no longer required.

Issue #769, opened 2017-06-19
HostGAPlugin requests should never go through proxy


paulmey:
When downloading from storage fails, requests [are routed through the HostGAPlugin](https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/common/protocol/wire.py#L629), but they are also still going through the proxy. In the case that the proxy is misconfigured, this will lead to ultimate failure to download the artifact.
Additionally, there should probably be a warning when a proxy is configured and storage is still unreachable through the proxy.

brendandixon:
Addressed by #776

Issue #764, opened 2017-06-16
[2.2.12] waagent -configuration-path:/path -start not take effect


yuhuijiang:
hi there,
I've ran the command "waagent -configuration-path:/path -start",but the parameter "-configuration-path" not take effect.
This will affect WALA running on RHEL6.x
(Because on RHEL6.x, execute WALA service through "serivce waagent start",it will invoke /etc/init.d/waaget.And in this file,execute WALA through "waagent -start",so this will affect WALA)

below is code in agent.py
```
220 def start():
221     """
222     Start agent daemon in a background process and set stdout/stderr to
223     /dev/null
224     """
225     devnull = open(os.devnull, 'w')
226     subprocess.Popen([sys.argv[0], '-daemon'], stdout=devnull, stderr=devnull)
```

brendandixon:
Addressed by #776

Issue #761, opened 2017-06-15
MetadataProtocol has no client object


hglkrijger:
```
2017/06/14 22:46:43.331214 WARNING Agent WALinuxAgent-2.2.12 launched with command 'python3 -u /usr/sbin/waagent -run-exthandlers' failed with return code: 1
2017/06/14 22:46:43.339738 ERROR Event: name=WALinuxAgent, op=Enable, message=Agent WALinuxAgent-2.2.12 launched with command 'python3 -u /usr/sbin/waagent -run-exthandlers' failed with return code: 1
2017/06/14 22:46:43.348803 WARNING Agent WALinuxAgent-2.2.12 launched with command 'python3 -u /usr/sbin/waagent -run-exthandlers' returned code: 1
2017/06/14 22:46:43.356455 INFO Installed Agent WALinuxAgent-2.2.12 is the most current agent
2017/06/14 22:46:43.773154 INFO Agent WALinuxAgent-2.2.12 is running as the goal state agent
2017/06/14 22:46:43.856962 INFO Event: name=WALinuxAgent-2.2.12, op=HeartBeat, message=
2017/06/14 22:46:43.859580 INFO Start env monitor service.
2017/06/14 22:46:43.937592 INFO Configure routes
2017/06/14 22:46:43.965398 INFO Gateway:None
2017/06/14 22:46:43.976631 INFO Routes:None
2017/06/14 22:46:44.025814 INFO WALinuxAgent-2.2.12 running as process 5704
2017/06/14 22:46:44.033808 INFO WALinuxAgent-2.2.12 did not terminate cleanly
2017/06/14 22:46:44.040433 ERROR Event: name=WALinuxAgent, op=Restart, message=WALinuxAgent-2.2.12 did not terminate cleanly
2017/06/14 22:46:44.154968 ERROR [000008] 404 - POST: http://169.254.169.254/Microsoft.Compute/status/telemetry?api-version=2015-05-01-preview
2017/06/14 22:46:44.432870 WARNING Agent WALinuxAgent-2.2.12 failed with exception: 'MetadataProtocol' object has no attribute 'client'
2017/06/14 22:46:44.443499 WARNING Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/azurelinuxagent/ga/update.py", line 256, in run
    if self._upgrade_available():
  File "/usr/lib/python3/dist-packages/azurelinuxagent/ga/update.py", line 427, in _upgrade_available
[...]

Issue #757, opened 2017-06-09
[2.2.12]deprovision is executed even if input 'n'


yuxisun1217:
Hi,

When I run "waagent -deprovision" and input "n" at the confirmation, the deprovisioning is still executed.
It seems that it doesn't use the return value of "do_confirmation" function.(pa/deprovision/default.py Line#220)

brendandixon:
Addressed by #776

PR #776, opened 2017-06-21
Address #769, #764, #761,#757, #754, #746, #740

brendandixon:
[#769] -- HostGAPlugin requests should never go through proxy
[#764] -- waagent -configuration-path:/path -start not take effect
[#761] -- MetadataProtocol has no client object
[#757] -- deprovision is executed even if input 'n'
[#754] -- /etc/resolv.conf is removed if create a VM base on an specialized image
[#746] -- Agent WALinuxAgent-2.2.12 is blacklisted - skipping download
[#740] -- Duplicate logging on RHEL/CentOS

### DIFF
--- a/azurelinuxagent/common/logger.py
+++ b/azurelinuxagent/common/logger.py
@@ -20,7 +20,13 @@ Log utils
 import os
 import sys
 from azurelinuxagent.common.future import ustr
-from datetime import datetime
+from datetime import datetime, timedelta
+
+EVERY_DAY = timedelta(days=1)
+EVERY_HALF_DAY = timedelta(hours=12)
+EVERY_HOUR = timedelta(hours=1)
+EVERY_HALF_HOUR = timedelta(minutes=30)
+EVERY_FIFTEEN_MINUTES = timedelta(minutes=15)
 
 class Logger(object):
     """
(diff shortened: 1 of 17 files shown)


azurelinuxagent/common/logger.py
```diff
@@ -20,18 +20,37 @@
 import os
 import sys
 from azurelinuxagent.common.future import ustr
-from datetime import datetime
+from datetime import datetime, timedelta
+
+EVERY_DAY = timedelta(days=1)
```
hglkrijger: Can we avoid redefining these?

azurelinuxagent/common/protocol/metadata.py
```diff
@@ -236,14 +236,14 @@ def get_ext_handlers(self, last_etag=None):
...
-        logger.info("Get extension handler packages")
+        logger.verbose("Get extension handler packages")
         pkg_list = ExtHandlerPackageList()
 
         manifest = None
         for version_uri in ext_handler.versionUris:
             try:
                 manifest, etag = self._get_data(version_uri.uri)
-                logger.info("Successfully downloaded manifest")
+                logger.verbose("Successfully downloaded manifest")
```
hglkrijger: I think these may affect the BVTs, we'll see.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -597,9 +597,9 @@ def call_storage_service(http_req, *args, **kwargs):
         Call storage service, handle SERVICE_UNAVAILABLE(503)
         """
 
-        # force the chk_proxy arg to True, since all calls to storage should
-        #  use a configured proxy
-        kwargs['chk_proxy'] = True
+        # Default to use the configured HTTP proxy
+        if not 'chk_proxy' in kwargs or kwargs['chk_proxy'] is None:
```
hglkrijger: Thanks for fixing this.

azurelinuxagent/ga/update.py
```diff
@@ -287,6 +286,7 @@ def run(self):
                 time.sleep(GOAL_STATE_INTERVAL)
 
         except Exception as e:
+            print(e)
```
hglkrijger: Should this be removed?

brendandixon: Yikes! Yup.

PR #782, opened 2017-06-22
Add Extension.Enabled to enable or disable extensions.

reyk:
Hi,

while extension might provide a useful feature, it is also very desirable for many users to disable them entirely. This is especially important for OpenBSD where we don't support most of the Linux-specific extensions. And it matches the our users' expectation to not run remotely-installed code.

This patch keeps extensions enabled by default, but they can be disabled in `/etc/waagent.conf`:

    # Enable or disable extensions, default is enabled
    Extension.Enabled=n

Reyk

### DIFF
--- a/azurelinuxagent/common/conf.py
+++ b/azurelinuxagent/common/conf.py
@@ -112,6 +112,9 @@ def get_agent_pid_file_path(conf=__conf__):
     return conf.get("Pid.File", "/var/run/waagent.pid")
 
 
+def get_ext_enabled(conf=__conf__):
+    return conf.get_switch("Extension.Enabled", True)
+
 def get_ext_log_dir(conf=__conf__):
     return conf.get("Extension.LogDir", "/var/log/azure")
 
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -240,9 +240,10 @@ class UpdateHandler(object):
         from azurelinuxagent.ga.env import get_env_handler
         get_env_handler().run()
 
-        from azurelinuxagent.ga.exthandlers import get_exthandlers_handler, migrate_handler_state
-        exthandlers_handler = get_exthandlers_handler()
-        migrate_handler_state()
+        if conf.get_ext_enabled():
+            from azurelinuxagent.ga.exthandlers import get_exthandlers_handler, migrate_handler_state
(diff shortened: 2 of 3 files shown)


reyk:
Hi! Any feedback on this? :)

brendandixon:
@reyk Thank you for your contribution, but since managing extensions is largely what the agent does, having an all-up disable switch is too coarse. We are moving to a model that removes provisioning from the agent (to be handled by `cloud-init`, as happens on most Ubuntu images in Azure at present) which will then leave only extension handling in the agent. Then the switch would be even less appropriate.

Again, thank you, but the feature is at odds with the direction for the agent.

reyk:
So will it be possible to run (marketplace-compatible) VMs without the agent?

I thought that the agent is still mandatory because it signals ready state to the Azure fabric. Is this going away? And with cloud-init, my impression was that the agent is still needed to handle ready-state, the ovf-env.xml file and to get the SSH public keys from the wire server.

I would appreciate if this all could be done in a cloud-init style via
http://169.254.169.254/latest/meta-data/public-keys/0/openssh-key etc.

Most extensions seems to be Linux- or Windows-only. The Linux diagnostic extension doesn't even run and aborts on start. So we had to disable extensions to use the agent on OpenBSD.

https://github.com/openbsd/ports/tree/master/sysutils/waagent

brendandixon:
@reyk We would like to achieve that goal. A couple of challenges remain, as you note, around such issues as SSH keys. The goal is to move toward a more "industry standard" model.

Regarding OpenBSD: Azure does not install any extensions by default. If an extension is being installed, it must have been requested.

reyk:
OK, thanks for the clarification.

Azure tries to install the Linux Diagnostics extension as soon as you enable boot diagnostics for a VM. So the request for it is not obvious for users. The extension doesn't get to the point where it checks OS compatibility because it tries to import a non-existing file and aborts (see https://github.com/Azure/azure-linux-extensions/issues/411). I didn't find a clean way to handle this in waagent or our port, but maybe waagent should be more resilient against "broken" modules (ignore the module if it failed to load instead of returning an error to get the ready-state up to Running).

brendandixon:
@reyk It's been debated. We've, for now, decided to error since not installing a requested extension is a failure to achieve the state requested. Of course, the root cause of this error is the (poor) choice to import the old `waagent` code.

PR #786, opened 2017-06-23
Add Provisioning.SshHostKeyPairType=auto to support ssh-keygen -A

reyk:
The supported values are now `rsa`, `dsa`, `ecdsa`, `ed25519`, or **`auto`**.  **`auto`** generates all supported host key types in the correct directory by calling `ssh-keygen -A` (supported since ***Open*****SSH** 5.9 in 2011).  The code uses `rsa` for the thumbprint because only a single one is supported by the agent and the Azure protocol.


brendandixon:
@reyk LGTM.

Issue #792, opened 2017-07-04
Clearlinux distro_name does not match


chuyd:
Clearlinux distro_name does not match

brendandixon:
Address by #796

PR #793, opened 2017-07-04
Fix clearlinux distro_name (#792)

chuyd:
Updating distro_name condition to match /usr/lib/os-release for current 
clearlinux releases.

Signed-off-by: Ornelas Aguayo, Jesus <jesus.ornelas.aguayo@intel.com>

### DIFF
--- a/azurelinuxagent/common/osutil/factory.py
+++ b/azurelinuxagent/common/osutil/factory.py
@@ -41,7 +41,7 @@ def get_osutil(distro_name=DISTRO_NAME,
     if distro_name == "arch":
         return ArchUtil()
 
-    if distro_name == "clear linux software for intel architecture":
+    if distro_name == "clearlinux":
         return ClearLinuxUtil()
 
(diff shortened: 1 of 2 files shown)


albertomurillo:
What about
```python
if distro.name == 'clearlinux'
    return ClearLinuxUtil()
```

And make distro.name return 'clearlinux' rather than the full name?

brendandixon:
@chuyd We likely need to continue supporting older releases as well. What if you either checked a regex or a startswith for "clear linux "?

rnesius:
I opened a new pull request with a fix that supports both old and new name.

chuyd:
@brendandixon I updated the commit to support older releases

chuyd:
@rnesius I updated the commit  to add support for previous and current Clear Linux OS for
Intel Architecure releases by searching the "Clear Linux" string
in osinfo[0] and returning clearlinux in distro_name, what do you think?

brendandixon:
Addressed by #796

PR #796, opened 2017-07-07
Accommodate new Clear Linux distro string in os-release.

rnesius:
Support both new and old text for Clear Linux in os-release. Verified this works in a live test in Azure.


brendandixon:
@rnesius Looks good. Thanks.

PR #799, opened 2017-07-11
Os release fix

rnesius:
I did not have my usual vim environment when doing this work, so missed a trailing space. Also, I discovered I really should have tested from a clean build as I would have discovered this very same bug lurking in setup.py.


rnesius:
My name/github ID should be on the official list on the CLA between Intel & Microsoft.

rnesius:
Any luck validating my credentials on the CLA? I was told the CLA-BOT was supposed to auto-recognize this. 

That said I don't know if this patch really needs one...

PR #803, opened 2017-07-17
fix for broken setup.py

hglkrijger:
Fixes break due to #799 

/cc @rnesius fyi


rnesius:
Thx man.  Not sure how I got that right in my local patch and wrong in what I submitted. :-/

rnesius:
Oh I see now. Short-circuiting saved me. Doh! :-)

Issue #812, opened 2017-07-24
Agent fails and isn't recoverable if an extension's log directory is not present anymore


sriramsa:
If an extension's log directory is deleted by users, WALinuxAgent fails and exists with the following error. Some customers are cleaning up /var/log directory to reclaim space and are inadvertently deleting the /var/log/azure folder. Agent should handle this gracefully.

This is the error thrown.
```
2017/07/17 10:18:56.439857 INFO WALinuxAgent-2.2.10 did not terminate cleanly
2017/07/17 10:18:56.439962 ERROR Event: name=WALinuxAgent, op=Restart, message=WALinuxAgent-2.2.10 did not terminate cleanly
2017/07/17 10:18:56.510020 WARNING Agent WALinuxAgent-2.2.12 is blacklisted - skipping download
2017/07/17 10:18:56.510858 WARNING Agent WALinuxAgent-2.2.13 is blacklisted - skipping download
2017/07/17 10:18:56.520270 WARNING Agent WALinuxAgent-2.2.10 failed with exception: [Errno 20] Not a directory: '/var/log/azure/Microsoft.OSTCExtensions.VMAccessForLinux'
2017/07/17 10:18:56.524184 WARNING Traceback (most recent call last):
  File "build/bdist.linux-x86_64/egg/azurelinuxagent/ga/update.py", line 257, in run
    exthandlers_handler.run()
  File "build/bdist.linux-x86_64/egg/azurelinuxagent/ga/exthandlers.py", line 187, in run
    self.handle_ext_handlers(etag)
  File "build/bdist.linux-x86_64/egg/azurelinuxagent/ga/exthandlers.py", line 210, in handle_ext_handlers
    self.handle_ext_handler(ext_handler, etag)
 File "build/bdist.linux-x86_64/egg/azurelinuxagent/ga/exthandlers.py", line 213, in handle_ext_handler
    ext_handler_i = ExtHandlerInstance(ext_handler, self.protocol)
  File "build/bdist.linux-x86_64/egg/azurelinuxagent/ga/exthandlers.py", line 345, in __init__
    fileutil.mkdir(self.get_log_dir(), mode=0o755)
  File "build/bdist.linux-x86_64/egg/azurelinuxagent/common/utils/fileutil.py", line 94, in mkdir
    os.makedirs(dirpath)
  File "/usr/lib64/python2.7/os.py", line 150, in makedirs
    makedirs(head, mode)
[...]

Issue #811, opened 2017-07-23
Comments inline in /etc/waagent.conf cause config to not be read


mabicca:
Hi Guys,

Recently helped with an issue where swap was not being created/activated properly after upgrading the waagent.

It looks like we had these lines in /etc/waagent.conf where you can notice the comments inline (bold):

# Format if unformatted. If 'n', resource disk will not be mounted.
ResourceDisk.Format=y **#added to enable swap**

# File system on the resource disk
# Typically ext3 or ext4. FreeBSD images should use 'ufs2' here.
ResourceDisk.Filesystem=ext4

# Mount point for the resource disk
ResourceDisk.MountPoint=/mnt

# Create and use swapfile on resource disk.
ResourceDisk.EnableSwap=y **#added to enable swap**

# Size of the swapfile.
ResourceDisk.SwapSizeMB=2048 **#2gb swap size**

With these comments, the swap creation was not being done; however after removing the comments, it works as expected.

It was probably some recent change in the waagent that is causing this since older releases do not have this problem.

This was tested in Ubuntu 16.04 LTS and running waagent 2.2.14

brendandixon:
@mabicca The agent does not support end-of-line comments in its configuration file. Its parse code is quite simple (at present). We can better document this limitation.

mabicca:
Hi Brendan,

I guess that was changed recently, with previous versions that was working just fine. Yes, if we can at least document that, it would be great.

Thank you!

brendandixon:
@mabicca The agent has never accepted end-of-line comments. Odd that it "worked" previously.

Issue #801, opened 2017-07-13
Agent fails to clean-up PID files


brendandixon:
`update.py` writes the PID of the child "goal state" process. However, it never deletes those files. It should (likely within `update.py#_ensure_no_orphans`.

PR #813, opened 2017-07-24
[#801] -- Agent fails to clean-up PID files

brendandixon:
[#801] -- Agent fails to clean-up PID files
[#811] -- Comments inline in /etc/waagent.conf cause config to not be read
[#812] -- Agent fails and isn't recoverable if an extension's log directory is not present anymore

### DIFF
--- a/azurelinuxagent/agent.py
+++ b/azurelinuxagent/agent.py
@@ -64,7 +64,19 @@ class Agent(object):
         logger.add_logger_appender(logger.AppenderType.CONSOLE, level,
                                  path="/dev/console")
 
+        ext_log_dir = conf.get_ext_log_dir()
+        try:
+            if os.path.isfile(ext_log_dir):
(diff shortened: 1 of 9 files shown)


azurelinuxagent/agent.py
```diff
@@ -64,7 +64,19 @@ def __init__(self, verbose, conf_file_path=None):
         logger.add_logger_appender(logger.AppenderType.CONSOLE, level,
                                  path="/dev/console")
 
+        ext_log_dir = conf.get_ext_log_dir()
+        try:
+            if os.path.isfile(ext_log_dir):
+                raise Exception("{0} is a file".format(ext_log_dir))
+            if not os.path.isdir(ext_log_dir):
+                os.mkdir(ext_log_dir)
```
hglkrijger: might want to create the whole tree?

PR #819, opened 2017-07-27
Support RDMA driver configuration and update on Ubuntu

longlimsft:
Compared to CentOS and SUSE, Ubuntu uses a different method to package RDMA driver. It builds multiple versions of the driver in the same kernel. During boot process, WALA is responsible for locating the correct driver. Due to multiple versions of the same driver are present in the system, the hv_network_direct needs to be blacklisted in /etc/modprobe.d/vmbus-rdma.conf. WALA then probes the correct driver to load. If no suitable driver is found, WALA attempts to perform a kernel update.

### DIFF
--- a/azurelinuxagent/pa/rdma/ubuntu.py
+++ b/azurelinuxagent/pa/rdma/ubuntu.py
@@ -1,0 +1,112 @@
+# Microsoft Azure Linux Agent
+#
+# Copyright 2014 Microsoft Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Requires Python 2.4+ and Openssl 1.0+
+#
+
+import glob
+import os
+import re
+import time
+import azurelinuxagent.common.logger as logger
+import azurelinuxagent.common.utils.shellutil as shellutil
+from azurelinuxagent.common.rdma import RDMAHandler
+
+
+class UbuntuRDMAHandler(RDMAHandler):
+
+    def install_driver(self):
+        #Install the appropriate driver package for the RDMA firmware
+
+        nd_version = RDMAHandler.get_rdma_version()
+        if not nd_version:
+            logger.error("RDMA: Could not determine firmware version. No driver will be installed")
+            return
+        nd_version = re.sub('\..*', '', nd_version) #strip the trailing .*
+
+        #Check to see if we need to reconfigure driver
+        status,output = shellutil.run_get_output('modprobe -R hv_network_direct')
+        if status != 0:
+            logger.error("RDMA: modprobe -R hv_network_direct failed. Skip updating driver")
+            return
+        logger.info("RDMA: current RDMA driver %s nd_version %s" % (output, nd_version))
+        if output == 'hv_netowrk_direct_%s' % nd_version:
+            logger.info("RDMA: driver is installed and ND version matched. Skip reconfiguring driver")
+            return
+
+        #Reconfigure driver if one is available
+        status,output = shellutil.run_get_output('modinfo hv_network_direct_%s' % nd_version);
+        if status == 0:
+            logger.info("RDMA: driver with ND version is installed. Link to module name")
+            self.update_modprobed_conf(nd_version)
+            return
+
+	#Driver not found. We need to check to see if we need to update kernel
+        status,output = shellutil.run_get_output('uname -r')
+        if status != 0:
+            logger.error("RDMA: uname -r failed.")
+            return
+        kernel_version = re.sub('-azure$', '', output)
+
+        #Find the new kernel package version
+        status,output = shellutil.run_get_output('apt-get update')
+        if status != 0:
+            logger.error("RDMA: apt-get update failed.")
+            return
+        status,output = shellutil.run_get_output('apt show linux-azure')
+        if status != 0:
+            logger.error("RDMA: apt show linux-azure failed.")
+            return
+        r = re.search('Version: (\S+)', output)
+        if not r:
+            logger.error("RDMA: version not found in package linux-azure.")
+            return
+        package_version = r.groups()[0]
+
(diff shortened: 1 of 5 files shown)


azurelinuxagent/pa/rdma/ubuntu.py
```diff
@@ -0,0 +1,112 @@
...
+        if not nd_version:
+            logger.error("RDMA: Could not determine firmware version. No driver will be installed")
+            return
+        nd_version = re.sub('\..*', '', nd_version) #strip the trailing .*
+
+        #Check to see if we need to reconfigure driver
+        status,output = shellutil.run_get_output('modprobe -R hv_network_direct')
+        if status != 0:
+            logger.error("RDMA: modprobe -R hv_network_direct failed. Skip updating driver")
+            return
```
marcelocerri: If the alias is not found we probably need to try to create it instead of returning, right?

azurelinuxagent/pa/rdma/ubuntu.py
```diff
@@ -0,0 +1,112 @@
...
+            logger.error("RDMA: uname -r failed.")
+            return
+        kernel_version = re.sub('-azure$', '', output)
+
+        #Find the new kernel package version
+        status,output = shellutil.run_get_output('apt-get update')
+        if status != 0:
+            logger.error("RDMA: apt-get update failed.")
+            return
+        status,output = shellutil.run_get_output('apt show linux-azure')
```
marcelocerri: `apt` will print some error message when running from a script. Use `apt-cache show --no-all-versions linux-azure` instead.

azurelinuxagent/pa/rdma/ubuntu.py
```diff
@@ -0,0 +1,112 @@
...
+            logger.info("RDMA: driver with ND version is installed. Link to module name")
+            self.update_modprobed_conf(nd_version)
+            return
+
+	#Driver not found. We need to check to see if we need to update kernel
+        status,output = shellutil.run_get_output('uname -r')
+        if status != 0:
+            logger.error("RDMA: uname -r failed.")
+            return
+        kernel_version = re.sub('-azure$', '', output)
```
marcelocerri: If the user is not running the azure kernel this suffix will probably be `-generic`. I'm not sure how to proceed in that case, but maybe we should skip here if the suffix is not `-azure`.

azurelinuxagent/pa/rdma/ubuntu.py
```diff
@@ -0,0 +1,112 @@
...
+            logger.error("RDMA: apt show linux-azure failed.")
+            return
+        r = re.search('Version: (\S+)', output)
+        if not r:
+            logger.error("RDMA: version not found in package linux-azure.")
+            return
+        package_version = r.groups()[0]
+
+        logger.info('RDMA: kernel_version=%s package_version=%s' % (kernel_version, package_version))
+        if kernel_version < package_version:
```
marcelocerri: Those two versions follow slightly different formats:

- `uname -r` will return a version in the format `<base kernel version>-<ABI number>-<flavour>` (and you are properly removing the flavour).
- The `linux-azure` version will be in the format `<base kernel version>.<ABI number>.<upload number>` (please notice the dot instead of a dash and the additional upload number).

Besides that, since both variable are strings which might cause some incorrect results.

azurelinuxagent/pa/rdma/ubuntu.py
```diff
@@ -0,0 +1,112 @@
...
+                logger.error("RDMA: apt-get -y install linux-azure failed")
+                logger.error("RDMA: %s" % output)
+                logger.error("RDMA: kernel update failed")
+                return
+            self.reboot_system()
+        else:
+            logger.error("RDMA: no kernel update is avaiable for ND version %s" % nd_version)
+
+    def update_modprobed_conf(self, nd_version):
+        #Update /etc/modprobe.d/vmbus-rdma.conf to point to the correct driver
```
marcelocerri: Who would be responsible for creating the first version of the conf file?

azurelinuxagent/pa/rdma/ubuntu.py
longlimsft: Good points. In this case we should add an alias in /etc/modprobe.d/VMBus-rdma.conf to point to the correct module. I'll make the change.

azurelinuxagent/pa/rdma/ubuntu.py
longlimsft: I will make the change.

azurelinuxagent/pa/rdma/ubuntu.py
longlimsft: You are right. There is no reason to attempt a kernel update on a non azure kernel in HPC environment.

azurelinuxagent/pa/rdma/ubuntu.py
longlimsft: I'll fix the dots and dashes.

Does the <upload number> show up in the current kernel? What's the best way to find out if the kernel package version is newer than the one currently installed on the system?

Another approach is that when you release a new kernel package for new ND devices, the ABI is increased. A check for <base kernel version>.<ABI number> can decide if new kernel is available.

azurelinuxagent/pa/rdma/ubuntu.py
longlimsft: The conf file (along with some other things for HPC) needs to be there in the initial HPC image. This can be part of the image building process.

azurelinuxagent/pa/rdma/ubuntu.py
mhcerri: I think you just need to normalize the versions. Something like that should do the trick:

`version = [ int(x) for x in string_version.replace('-', '.').split('.') ][:4]`

(You will need to handle the ValueError exception just in case any of the tokens are not a number).

Some background: those formats differ because we actually have two different type of packages here.

The kernel is contained in the package `linux-image-<version>-<abi>-azure`. For example: the package named `linux-image-4.11.0-1003-azure` has the version `4.11.0-1003.3`. The kernel itself (and thus `uname -r`) doesn't carry the last number because that is just the upload number.

This package above is never installed directly. We usually install a meta package named `linux-azure` that has a dependency to the correct `linux-image` package. For the example above, the `linux-azure` version would be `4.11.0.1003.3`. The meta package doesn't use a dash because it's a native package.

azurelinuxagent/pa/rdma/ubuntu.py
mhcerri: In that case we would have to include this initial conf file inside the debian package.

azurelinuxagent/pa/rdma/ubuntu.py
mhcerri: Using an array of integers you can compare two versions without problem.

longlimsft:
Will make an updated pull request.

PR #820, opened 2017-07-27
Add RDMA support for Ubuntu

longlimsft:

### DIFF
--- a/azurelinuxagent/pa/rdma/ubuntu.py
+++ b/azurelinuxagent/pa/rdma/ubuntu.py
@@ -21,6 +21,7 @@ import glob
 import os
 import re
 import time
+import azurelinuxagent.common.conf as conf
 import azurelinuxagent.common.logger as logger
 import azurelinuxagent.common.utils.shellutil as shellutil
 from azurelinuxagent.common.rdma import RDMAHandler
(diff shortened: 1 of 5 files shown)


azurelinuxagent/pa/rdma/ubuntu.py
```diff
@@ -0,0 +1,116 @@
...
+        r = re.search('Version: (\S+)', output)
+        if not r:
+            logger.error("RDMA: version not found in package linux-azure.")
+            return
+        package_version = r.groups()[0]
+        #Remove the ending .<upload number> after <ABI number>
+        package_version = re.sub("\.\d+$", "", package_version)
+
+        logger.info('RDMA: kernel_version=%s package_version=%s' % (kernel_version, package_version))
+        if kernel_version < package_version:
```
mhcerri: I still think it's safer to convert the versions to arrays of integers before comparing them:

    >>> a = '4.4.0.1001'
    >>> b = '4.11.0.1004'
    >>> 
    >>> a > b
    True
    >>> [ int(x) for x in a.split('.') ] > [ int(x) for x in b.split('.') ]
    False

longlimsft: Good point. Will make the change.

longlimsft:
Will create an updated pull request.

PR #824, opened 2017-08-01
Support RDMA for Ubuntu

longlimsft:

### DIFF
--- a/azurelinuxagent/pa/rdma/ubuntu.py
+++ b/azurelinuxagent/pa/rdma/ubuntu.py
@@ -1,0 +1,118 @@
+# Microsoft Azure Linux Agent
+#
+# Copyright 2014 Microsoft Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Requires Python 2.4+ and Openssl 1.0+
+#
+
+import glob
+import os
+import re
+import time
+import azurelinuxagent.common.logger as logger
+import azurelinuxagent.common.utils.shellutil as shellutil
+from azurelinuxagent.common.rdma import RDMAHandler
+
+
+class UbuntuRDMAHandler(RDMAHandler):
+
+    def install_driver(self):
+        #Install the appropriate driver package for the RDMA firmware
+
+        nd_version = RDMAHandler.get_rdma_version()
+        if not nd_version:
+            logger.error("RDMA: Could not determine firmware version. No driver will be installed")
+            return
+        #strip the trailing .*, we are looking for number like 144
+        nd_version = re.sub('\..*', '', nd_version)
+
+        #Check to see if we need to reconfigure driver
+        status,module_name = shellutil.run_get_output('modprobe -R hv_network_direct', chk_err=False)
+        if status != 0:
+            logger.info("RDMA: modprobe -R hv_network_direct failed. Use module name hv_network_direct")
+            module_name = "hv_network_direct"
+        else:
+            module_name = module_name.strip()
+        logger.info("RDMA: current RDMA driver %s nd_version %s" % (module_name, nd_version))
+        if module_name == 'hv_network_direct_%s' % nd_version:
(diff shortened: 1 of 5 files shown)


config/ubuntu/waagent.conf
```diff
@@ -82,6 +82,9 @@ OS.SshDir=/etc/ssh
 # Enable RDMA management and set up, should only be used in HPC images
 # OS.EnableRDMA=y
 
+# Enable RDMA kernel update, should only be used with Ubuntu
```
brendandixon: @longlimsft Or, more correctly, the option is only used on Ubuntu images?

brendandixon:
@longlimsft Are there any tests you could add? What testing was done using live VMs?

config/ubuntu/waagent.conf
longlimsft: This option can be used for other distributions (although they are not checked in the code today). The comment there is for not confusing people running on other distributions.

longlimsft:
I don't have unit test cases to add. The tests are mostly done with Ubuntu linux-azure kernel in HPC clusters, running MPI and DAPL. There is no easy way to test with standalone agent. The tests also depend on package update repositories maintain by Canonical.

azurelinuxagent/pa/rdma/ubuntu.py
```diff
@@ -0,0 +1,118 @@
...
+            logger.error("RDMA: no kernel update is avaiable for ND version %s" % nd_version)
+
+    def update_modprobed_conf(self, nd_version):
+        #Update /etc/modprobe.d/vmbus-rdma.conf to point to the correct driver
+
+        modprobed_file = '/etc/modprobe.d/vmbus-rdma.conf'
+        if not os.path.isfile(modprobed_file):
+            logger.error("RDMA: % not found" % modprobed_file)
+            logger.error("RDMA: modprobed conf file is required for Ubuntu")
+            return
```
mhcerri: @longlimsft , can we create the conf file if it doesn't exist. It's not a good idea to have the .deb package shipping the initial version and then the agent changing it.

azurelinuxagent/pa/rdma/ubuntu.py
```diff
@@ -0,0 +1,118 @@
...
+
+        f = open(modprobed_file, 'r')
+        lines = f.read()
+        f.close()
+        r = re.search('alias hv_network_direct hv_network_direct_\S+', lines)
+        if r:
+            lines = re.sub('alias hv_network_direct hv_network_direct_\S+', 'alias hv_network_direct hv_network_direct_%s' % nd_version, lines)
+        else:
+            lines += '\nalias hv_network_direct hv_network_direct_%s\n' % nd_version
```
mhcerri: @longlimsft, Not sure if I'm missing something, but I couldn't find a call to `load_driver_module()` and we are not creating the hotplug alias either, then the driver will probably never get loaded.

azurelinuxagent/pa/rdma/ubuntu.py
longlimsft: This is a good point. I'll make the change.

azurelinuxagent/pa/rdma/ubuntu.py
longlimsft: The driver is loaded in RDMADeviceHandler:process at azurelinux/common/rdma.py.

config/ubuntu/waagent.conf
longlimsft: I'll change the wording. As of today, this option is only effective under Ubuntu.

azurelinuxagent/pa/rdma/ubuntu.py
longlimsft: On a second thought, there is a problem with this approach. How do we prevent two identical drivers from loading at the same time if there is no /etc/modprobe.d/vmbus-rdma.conf? It's required to have a conf file or something to not load hv_network_direct on boot.

mhcerri: Maybe we can have two files. I can add a file to `/lib/modprobe.d/` when the azure kernel is installed to prevent the drivers to be hotplugged. And then the agent can create the conf file in `/etc/modprobe.d` to override the conf file in `/lib/modprobe.d/`.

The agent can create aliases to `hv_network_direct` and to the hotplug alias.

I think the agent can create alias to

longlimsft: Maybe we can put "install vmbus:3daf2e8ca732094bab99bd1f1c86b501 true" to a file in /lib/modprobe.d? That will prevent hv_network_driver from auto loading.

Then the agent can check, modify or create /etc/modprobe.d/vmbus-rdma.conf, and add "alias alias hv_network_direct hv_network_direct_XXX".

longlimsft:
Close this pull request. Will create a new one to address comments.

Issue #823, opened 2017-07-31
Show config options in use


hglkrijger:
Request from support to add an option to show all configuration options in use, similar to `sshd -T`.

brendandixon:
Addressed by #837

PR #832, opened 2017-08-03
[#823] -- Show config options in use

brendandixon:
[#823] -- Show config options in use


brendandixon:
Resubmit later

PR #834, opened 2017-08-04
RFE: add support for monitoring configure file changes

jan-wang:
In Azure, several configure files have some special settings, such as, sshd configure file(KeepAlive etc), dhcp client configure file(send hostname). But if we do system upgrade(such like with one rpm file), these special settings will be overwritten. This will have some impactions for use, such like,
if sshd keepalive option is lost, user 's current terminal will hang after a while idle.
if dhcp send hostname option is lost, hostname cannot be resolved after changing hostname.

so this PR adds support to monitor these files changes and make sure these setting will be there even after system upgrade.

### DIFF
--- a/azurelinuxagent/common/conf.py
+++ b/azurelinuxagent/common/conf.py
@@ -200,6 +200,8 @@ def get_password_crypt_salt_len(conf=__conf__):
 def get_monitor_hostname(conf=__conf__):
     return conf.get_switch("Provisioning.MonitorHostName", False)
 
+def get_monitor_conf_file(conf=__conf__):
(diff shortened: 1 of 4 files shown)


jan-wang:
Hi, 
Can you review this?
appreciate any comments/suggestions.

brendandixon:
@jan-wang I believe we previously discussed this feature: It is not within the design of the agent to support monitoring such files. The agent exists only to enable VMs running in Azure (e.g., provision those VMs, manage Azure extensions). It is best if you write your own package to include on your VMs to monitor these files. Take care.

PR #835, opened 2017-08-04
Support RDMA for Ubuntu

longlimsft:


longlimsft:
There is a Ubuntu kernel configuration change to work on auto hotplug RDMA issue on boot. Will update this pull to work with the new configuration. Close this.

Issue #784, opened 2017-06-22
Adjust http retry and logging


hglkrijger:
The retry logic in `restutil#http_request` incorrectly adjusts `max_retry` which does not actually affect the retry loop, and has the side effect of disabling logging on subsequent http errors. 

We should decide on the appropriate behavior here and fix accordingly.

brendandixon:
Addressed by #837

PR #837, opened 2017-08-09
Fix issues

brendandixon:
[#823] -- Show config options in use
[#784] -- Adjust http retry and logging

### DIFF
--- a/azurelinuxagent/agent.py
+++ b/azurelinuxagent/agent.py
@@ -128,6 +128,11 @@ class Agent(object):
         update_handler = get_update_handler()
         update_handler.run()
 
+    def show_configuration(self):
+        configuration = conf.get_configuration()
+        for k in sorted(configuration.keys()):
+            print("{0} = {1}".format(k, configuration[k]))
+
 def main(args=[]):
     """
     Parse command line arguments, exit with usage() on error.
@@ -157,6 +162,8 @@ def main(args=[]):
                 agent.daemon()
             elif command == "run-exthandlers":
                 agent.run_exthandlers()
+            elif command == "show-configuration":
+                agent.show_configuration()
         except Exception:
             logger.error(u"Failed to run '{0}': {1}",
                          command,
@@ -198,6 +205,8 @@ def parse_args(sys_args):
             verbose = True
         elif re.match("^([-/]*)force", a):
             force = True
+        elif re.match("^([-/]*show-configuration", a):
+            cmd = "show-configuration"
         elif re.match("^([-/]*)(help|usage|\\?)", a):
             cmd = "help"
         else:
--- a/azurelinuxagent/common/conf.py
+++ b/azurelinuxagent/common/conf.py
@@ -85,6 +85,69 @@ def load_conf_from_file(conf_file_path, conf=__conf__):
         raise AgentConfigError(("Failed to load conf file:{0}, {1}"
                                 "").format(conf_file_path, err))
 
+__SWITCH_OPTIONS__ = {
+    "OS.EnableRDMA" : False,
+    "OS.UpdateRdmaDriver" : False,
+    "OS.CheckRdmaDriver" : False,
+    "Logs.Verbose" : False,
+    "OS.EnableFIPS" : False,
+    "Provisioning.Enabled" : True,
+    "Provisioning.UseCloudInit" : False,
+    "Provisioning.AllowResetSysUser" : False,
+    "Provisioning.RegenerateSshHostKeyPair" : False,
+    "Provisioning.DeleteRootPassword" : False,
+    "Provisioning.DecodeCustomData" : False,
+    "Provisioning.ExecuteCustomData" : False,
+    "Provisioning.MonitorHostName" : False,
+    "DetectScvmmEnv" : False,
+    "ResourceDisk.Format" : False,
+    "DetectScvmmEnv" : False,
+    "ResourceDisk.Format" : False,
+    "ResourceDisk.EnableSwap" : False,
+    "AutoUpdate.Enabled" : True,
+    "EnableOverProvisioning" : False,
+    "OS.AllowHTTP" : False
+}
+
+__STRING_OPTIONS__ = {
+    "Lib.Dir" : "/var/lib/waagent",
+    "DVD.MountPoint" : "/mnt/cdrom/secure",
+    "Pid.File" : "/var/run/waagent.pid",
+    "Extension.LogDir" : "/var/log/azure",
+    "OS.OpensslPath" : "/usr/bin/openssl",
+    "OS.SshDir" : "/etc/ssh",
+    "OS.HomeDir" : "/home",
+    "OS.PasswordPath" : "/etc/shadow",
+    "OS.SudoersDir" : "/etc/sudoers.d",
+    "OS.RootDeviceScsiTimeout" : None,
+    "Provisioning.SshHostKeyPairType" : "rsa",
+    "Provisioning.PasswordCryptId" : "6",
+    "HttpProxy.Host" : None,
+    "ResourceDisk.MountPoint" : "/mnt/resource",
+    "ResourceDisk.MountOptions" : None,
+    "ResourceDisk.Filesystem" : "ext3",
+    "AutoUpdate.GAFamily" : "Prod"
+}
+
+__INTEGER_OPTIONS__ = {
+    "Provisioning.PasswordCryptSaltLength" : 10,
+    "HttpProxy.Port" : None,
+    "ResourceDisk.SwapSizeMB" : 0,
+    "Autoupdate.Frequency" : 3600
+}
+
+def get_configuration(conf=__conf__):
+    options = {}
+    for option in __SWITCH_OPTIONS__:
+        options[option] = conf.get_switch(option, __SWITCH_OPTIONS__[option])
+
+    for option in __STRING_OPTIONS__:
+        options[option] = conf.get(option, __STRING_OPTIONS__[option])
+
+    for option in __INTEGER_OPTIONS__:
+        options[option] = conf.get_int(option, __INTEGER_OPTIONS__[option])
+
+    return options
 
 def enable_rdma(conf=__conf__):
     return conf.get_switch("OS.EnableRDMA", False) or \
@@ -252,3 +315,6 @@ def get_autoupdate_frequency(conf=__conf__):
 
 def get_enable_overprovisioning(conf=__conf__):
     return conf.get_switch("EnableOverProvisioning", False)
+
+def get_allow_http(conf=__conf__):
+    return conf.get_switch("OS.AllowHTTP", False)
--- a/azurelinuxagent/common/protocol/hostplugin.py
+++ b/azurelinuxagent/common/protocol/hostplugin.py
@@ -22,7 +22,8 @@ import json
 import traceback
 
 from azurelinuxagent.common import logger
-from azurelinuxagent.common.exception import ProtocolError, HttpError
+from azurelinuxagent.common.exception import BadRequestError, \
+                                            HttpError, ProtocolError
 from azurelinuxagent.common.future import ustr, httpclient
 from azurelinuxagent.common.utils import restutil
 from azurelinuxagent.common.utils import textutil
@@ -85,10 +86,10 @@ class HostPluginProtocol(object):
         try:
             headers = {HEADER_CONTAINER_ID: self.container_id}
             response = restutil.http_get(url, headers)
-            if response.status != httpclient.OK:
+            if restutil.request_failed(response):
                 logger.error(
                     "HostGAPlugin: Failed Get API versions: {0}".format(
-                        self.read_response_error(response)))
+                        restutil.read_response_error(response)))
             else:
                 return_val = ustr(remove_bom(response.read()), encoding='utf-8')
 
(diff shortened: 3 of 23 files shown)


azurelinuxagent/agent.py
```diff
@@ -157,6 +162,8 @@ def main(args=[]):
                 agent.daemon()
             elif command == "run-exthandlers":
                 agent.run_exthandlers()
+            elif command == "show-configuration":
```
boumenot: >show-configuration [](start = 29, length = 18)

Does this just dump the waagent.conf, or the agent's interpretation of it?  Something else?

brendandixon: @boumenot It dumps the configuration state as the agent "sees" it. So, for unspecified values, it will display their default settings. I believe that is the intent of the request.

azurelinuxagent/common/utils/restutil.py
```diff
@@ -98,43 +146,63 @@ def _http_request(method, host, rel_uri, port=None, data=None, secure=False,
...
-    """
-    Sending http request to server
-    On error, sleep 10 and retry max_retry times.
-    """
+def http_request(method,
+                url, data, headers=None,
+                chk_proxy=False,
+                max_retry=DEFAULT_RETRIES,
+                retry_codes=RETRY_CODES,
+                retry_delay=SHORT_DELAY_IN_SECONDS):
```
boumenot: >retry_delay=SHORT_DELAY_IN_SECONDS): [](start = 16, length = 36)

It is worth randomizing this value across a small range.

azurelinuxagent/common/utils/restutil.py
```diff
@@ -145,55 +213,123 @@ def http_request(method, url, data, headers=None, max_retry=3,
...
+            if resp.status == httpclient.BAD_REQUEST:
+                raise BadRequestError()
+
             return resp
+
         except httpclient.HTTPException as e:
-            retry_msg = 'HTTP exception: {0} {1}'.format(log_msg, e)
-            retry_interval = 5
+            msg = '[HTTP Failed] HTTP {0} Exception {1}'.format(method, e)
+            if len([x for x in RETRY_EXCEPTIONS if isinstance(e, x)]) > 0:
```
boumenot: >len([x for x in RETRY_EXCEPTIONS if isinstance(e, x)]) > 0: [](start = 15, length = 59)

I like your request_failed() and request_succeeded() methods, so why not add one for this too.  A name is better than a list comprehension.

azurelinuxagent/common/utils/restutil.py
```diff
@@ -145,55 +213,123 @@ def http_request(method, url, data, headers=None, max_retry=3,
...
+
+            if resp.status == httpclient.BAD_REQUEST:
+                raise BadRequestError()
+
             return resp
+
         except httpclient.HTTPException as e:
-            retry_msg = 'HTTP exception: {0} {1}'.format(log_msg, e)
-            retry_interval = 5
+            msg = '[HTTP Failed] HTTP {0} Exception {1}'.format(method, e)
```
boumenot: >Exception [](start = 42, length = 9)

HTTPException instead of Exception to match what you've done with IOError.

azurelinuxagent/common/utils/restutil.py
brendandixon: @boumenot Makes sense, but I am, for now, preserving behavior existing as I have concerns over how changing those values will affect the "real world."

tests/utils/test_rest_util.py
```diff
@@ -115,6 +119,165 @@ def test_http_request_with_retry(self, _http_request, sleep):
...
+
+        self.assertRaises(BadRequestError, restutil.http_get, "https://foo.bar")
+        self.assertEqual(1, _http_request.call_count)
+
+    @patch("time.sleep")
+    @patch("azurelinuxagent.common.utils.restutil._http_request")
+    def test_http_request_retries_exceptions(self, _http_request, _sleep):
+        # Note:
+        # -- It would be best to test each possible exception,
+        #    but their differing signatures makes that a fair bit of work
```
boumenot: Does it makes sense to assert all of the code you assert based on then?  If the list changes at least you'll break a test, so the breaker knows someone thought about this list previously.

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -117,42 +118,44 @@ def get_artifact_request(self, artifact_url, artifact_manifest_url=None):
...
+        # logger.periodic(
+        #     logger.EVERY_FIFTEEN_MINUTES,
+        #     "HostGAPlugin: Put VM log to [{0}]".format(url))
+        # try:
+        #     response = restutil.http_put(url, content, headers)
+        #     if response.status != httpclient.OK:
+        #         logger.error("HostGAPlugin: Put log failed: Code {0}".format(
+        #             response.status))
+        # except HttpError as e:
+        #     logger.error("HostGAPlugin: Put log exception: {0}".format(e))
```
boumenot: delete?

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -179,7 +183,15 @@ def put_vm_status(self, status_blob, sas_url, config_blob_type=None):
             if not HostPluginProtocol.is_default_channel():
                 logger.verbose("HostGAPlugin: Setting host plugin as default channel")
                 HostPluginProtocol.set_default_channel(True)
+
         except Exception as e:
+            # If the HostPlugin rejects the request,
+            # let the error continue, but set to use the HostPlugin
```
boumenot: I plead ignorance.  I don't understand this.  If HostGAPlugin rejects the request, why do you continue to use it?  Did your use of HostGAPlugin get reset in some way that I don't see?

boumenot:
:shipit:

azurelinuxagent/common/conf.py
```diff
@@ -85,6 +85,69 @@ def load_conf_from_file(conf_file_path, conf=__conf__):
         raise AgentConfigError(("Failed to load conf file:{0}, {1}"
                                 "").format(conf_file_path, err))
 
+__SWITCH_OPTIONS__ = {
+    "OS.EnableRDMA" : False,
```
hglkrijger: Maybe we should use some sort of structure to contain these keys, which is referenced by the get methods for the individual values, to prevent missing any in the future? Perhaps enums or just a dictionary. Not critical now, we could improve later.

azurelinuxagent/common/protocol/hostplugin.py
hglkrijger: +1

azurelinuxagent/common/protocol/wire.py
```diff
@@ -595,73 +590,78 @@ def save_cache(self, local_file, data):
...
-                    logger.info("Retry fetch in {0} seconds",
-                                SHORT_WAITING_INTERVAL)
-                    time.sleep(SHORT_WAITING_INTERVAL)
-                else:
-                    host.manifest_uri = version.uri
-                    logger.verbose("Manifest downloaded successfully from host plugin")
-                    if not HostPluginProtocol.is_default_channel():
-                        logger.info("Setting host plugin as default channel")
-                        HostPluginProtocol.set_default_channel(True)
+                    logger.verbose("Failed to download manifest, "
```
hglkrijger: I think there are some dungeon crawler tests which look for this string, might want to update those.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -595,73 +590,78 @@ def save_cache(self, local_file, data):
...
+                # let the error continue, but set to use the HostPlugin
+                except BadRequestError:
+                    HostPluginProtocol.set_default_channel(True)
+                    raise
+
+                # Note:
+                # -- The HostPlugin SHOULD NOT be used as a "cache" of the
+                #    manifest URI. The code SHOULD BE changed to properly
+                #    pass the manifest URI such that correct calls to the
+                #    HostPlugin may be formed.
```
hglkrijger: Is this a todo? Maybe we should rather have a GitHub issue instead of a comment.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -793,20 +793,55 @@ def get_ext_conf(self):
...
-        local_file = os.path.join(conf.get_lib_dir(), local_file)
-        xml_text = self.fetch_manifest(ext_handler.versionUris)
-        self.save_cache(local_file, xml_text)
-        return ExtensionManifest(xml_text)
+        update_goal_state = False
+
+        while True:
+            try:
+                if update_goal_state:
+                    update_goal_state = False
```
hglkrijger: should this be set after the update? otherwise a bad request there would just keep trying. 

the while true makes me twitchy, I'm not sure by reading the method what the exit criteria would be.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -793,20 +793,55 @@ def get_ext_conf(self):
...
     def get_gafamily_manifest(self, vmagent_manifest, goal_state):
-        local_file = MANIFEST_FILE_NAME.format(vmagent_manifest.family,
-                                               goal_state.incarnation)
-        local_file = os.path.join(conf.get_lib_dir(), local_file)
-        xml_text = self.fetch_manifest(vmagent_manifest.versionsManifestUris)
-        fileutil.write_file(local_file, xml_text)
-        return ExtensionManifest(xml_text)
+        update_goal_state = False
+
+        while True:
```
hglkrijger: same comment here. also, don't these differ by just a couple parameters? seems like it could be refactored.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -825,39 +860,56 @@ def check_wire_protocol_version(self):
             raise ProtocolNotFoundError(error)
 
     def upload_status_blob(self):
-        ext_conf = self.get_ext_conf()
+        update_goal_state = False
+
+        while True:
```
hglkrijger: I see the reason for changing this to a while loop, but having it open ended makes me nervous. does it make sense to put in a maximum retry limit?

azurelinuxagent/common/utils/restutil.py
```diff
@@ -98,43 +146,63 @@ def _http_request(method, host, rel_uri, port=None, data=None, secure=False,
...
     if chk_proxy:
-        proxy_host, proxy_port = get_http_proxy()
+        proxy_host, proxy_port = _get_http_proxy()
 
-    # If httplib module is not built with ssl support. Fallback to http
+    # If httplib module is not built with ssl support,
+    # fallback to HTTP if allowed
     if secure and not hasattr(httpclient, "HTTPSConnection"):
+        if not conf.get_allow_http():
+            raise HttpError("HTTPS is unavailable and required")
```
hglkrijger: Do not recall offhand how high this goes, but we might a ProtocolError or something else, to be handled differently?

azurelinuxagent/common/utils/textutil.py
```diff
@@ -259,6 +259,16 @@ def set_ini_config(config, name, val):
         config.insert(length - 1, text)
 
 
+def replace_non_ascii(incoming, replace_char=''):
+    outgoing = ''
+    for c in incoming:
```
hglkrijger: might want to check incoming for None

azurelinuxagent/common/utils/textutil.py
```diff
@@ -259,6 +259,16 @@ def set_ini_config(config, name, val):
...
+        if str_to_ord(c) > 128:
+            outgoing += replace_char
+        else:
+            outgoing += c
```
hglkrijger: maybe this should be delimited?

azurelinuxagent/common/conf.py
brendandixon: @hglkrijger Agreed. But later.

azurelinuxagent/common/protocol/hostplugin.py
brendandixon: Done.

azurelinuxagent/common/protocol/hostplugin.py
brendandixon: @boumenot As I shared in mail, the HostPlugin, as a channel is sound and should be used, it's only that the GoalState is stale.

azurelinuxagent/common/protocol/wire.py
brendandixon: @hglkrijger I'll remove the comment.

azurelinuxagent/common/protocol/wire.py
brendandixon: @hglkrijger You are correct. I was thinking more on this and will improve the logic.

azurelinuxagent/common/protocol/wire.py
brendandixon: @hglkrijger I'll review.

azurelinuxagent/common/protocol/wire.py
brendandixon: @hglkrijger Same comment as above. I'll rethink this a bit.

azurelinuxagent/common/utils/restutil.py
brendandixon: @hglkrijger I believe all callers support catching an HttpError and correctly handling it.

azurelinuxagent/common/utils/restutil.py
brendandixon: Done.

azurelinuxagent/common/utils/restutil.py
brendandixon: Ok.

azurelinuxagent/common/utils/textutil.py
brendandixon: Done.

azurelinuxagent/common/utils/textutil.py
brendandixon: Huh?

tests/utils/test_rest_util.py
brendandixon: Done.

azurelinuxagent/common/protocol/wire.py
hglkrijger: I thought you said you were going to revert this? Or did you decide to update tests?

azurelinuxagent/common/utils/restutil.py
hglkrijger: Ok, so we will just keep looping? I thought perhaps we might want to cleanly exit the process if there is no hope of continuing on.

brendandixon: `HttpError` is what `http_request` returns for failures. I've only introduced a new version of that already raised exception. All callers must already correctly catch / handle instances of `HttpError`.

azurelinuxagent/common/protocol/wire.py
brendandixon: I *believe* I reverted the message back to the original. The DungeonCrawler tests all passed against these changes...

PR #854, opened 2017-08-23
No need to set hostname again while hostname changes

jan-wang:
It is unnecessary to set hostname again while monitor detects hostname change.
Just publish it.


brendandixon:
@jan-wang This is an artifact of how the per-distribution utilities work. In the best of all possible worlds, you would be correct. But, at present, the behavior of the various utilities deviates enough that we need to preserve both steps even though they are, in some cases, redundant.

PR #877, opened 2017-09-08
- RDMA load fixes for SUSE

rjschwei:
+ Fix comparison issue that causes the installed RDMA package to not
    be detected. Tris triggers a re-install of the package which in turn
    triggers a reboot. Thus we end up in a endless reboot loop.

### DIFF
--- a/azurelinuxagent/pa/rdma/suse.py
+++ b/azurelinuxagent/pa/rdma/suse.py
@@ -18,7 +18,6 @@
 #
 
 import glob
-import os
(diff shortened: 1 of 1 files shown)


rjschwei:
@hglkrijger any chance we can get this moved forward?

boumenot:
@hglkrijger will be out of the office for a little while, but @brendandixon is available.

brendandixon:
@rjschwei I'll review this today.

Issue #875, opened 2017-09-08
The agent should include the incarnation number in ProcessGoalState events


brendandixon:
The agent emits a "ProcessGoalState" event when it completes processing a GoalState. It should include in the message string the associated incarnation / ETag number (e.g., "Incarnation 42").

brendandixon:
Addressed by #878

Issue #874, opened 2017-09-07
The interval allowed before forcibly terminating the child process is too long


brendandixon:
The interval defined at https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/ga/update.py#L74 should be, at most, 15 minutes (15 * 60 seconds). At present it is defined as 15 hours (15 * 60 * 60).

brendandixon:
Addressed by #878

PR #878, opened 2017-09-08
Added safe deployment support and more

brendandixon:
Added safe deployment support
[#875] -- The agent should include the incarnation number in ProcessGoalState events
[#874] -- The interval allowed before forcibly terminating the child process is too long

### DIFF
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -17,6 +17,7 @@
 # Requires Python 2.4+ and Openssl 1.0+
 #
 
+import fnmatch
 import glob
 import json
 import os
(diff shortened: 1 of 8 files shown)


azurelinuxagent/ga/update.py
```diff
@@ -410,9 +449,45 @@ def _evaluate_agent_health(self, latest_agent):
...
+            self._enable_agents(agents)
+
+            for blacklist in [a.deploy.blacklisted for a in agents]:
+                self._blacklist_agents(blacklist)
+
+            for agent in agents:
+                agent.mark_deployed()
+
+        except Exception as e:
+            print("YIKES! {0}".format(e))
```
hglkrijger: Did you want to leave this in?

azurelinuxagent/ga/update.py
```diff
@@ -410,9 +449,45 @@ def _evaluate_agent_health(self, latest_agent):
...
+            logger.warn(
+                "Exception evaluating agents for safe deployment: {0}".format(
```
hglkrijger: Should we also emit any telemetry here?

brendandixon: @hglkrijger Thanks! Fixed.

PR #880, opened 2017-09-11
Rolling Extension Upgrades

uiri:
These code changes add support for an `upgradeGuid` XML attribute in a Plugin and a `ExtensionRollingUpgrade` XML element in the `GuestAgentExtension` . Both of these are part of the extension configuration document attached to a GoalState.

Old upgrade GUIDs are stored in a dictionary inside the `ExtHandlersHandler`.


tests/ga/test_extension.py
```diff
@@ -371,6 +371,84 @@ def test_ext_handler_no_ext(self, *args):
...
+                                                        "seqNo=\"2\"")
+        exthandlers_handler.run()
+        self._assert_no_handler_status(protocol.report_vm_status)
+
+        #Test GUID change with new version available
+        test_data.goal_state = test_data.goal_state.replace("<Incarnation>4<",
+                                                            "<Incarnation>5<")
+        test_data.ext_conf = test_data.ext_conf.replace("FE0987654322", "FE0987654323")
+        exthandlers_handler.run()
+        self._assert_handler_status(protocol.report_vm_status, "Ready", 1, "1.1.0")
```
hglkrijger: You should also test the behavior of hotfix updates, as well as comparative tests for `ExtensionRollingUpgrade=false`

azurelinuxagent/ga/exthandlers.py
```diff
@@ -205,6 +206,15 @@ def run_status(self):
...
         return
 
+    def get_guid(self, name):
+        return self.last_guids.get(name, None)
+
+    def is_new_guid(self, ext_handler):
+        last_guid = self.get_guid(ext_handler.name)
+        if last_guid is None:
+            return True
+        return last_guid != ext_handler.properties.upgradeGuid
```
hglkrijger: How should this behave if the process crashes or vm restarts?

brendandixon: @uiri Can we ever have a case where the upgradeGuid disappears? That is, it was present on the previous manifest and now is no longer? If so, then `last_guid != ext_handler.properties.upgradeGuid` will compare a GUID against `None`, treating `None` as the GUID. Is that the desired effect?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -288,6 +298,18 @@ def handle_ext_handler(self, ext_handler, etag):
...
+            if state == u"enabled":
+                # there is an upgrade GUID and it is NOT new
+                if ext_handler.properties.upgradeGuid is not None and \
+                        not self.is_new_guid(ext_handler):
+                    logger.info("New GUID is the same as the old GUID. Exiting without upgrading.")
+                    return
+            elif state == u"disabled" or state == u"uninstall":
+                # Remove the GUID from the dictionary in this case so that
+                # it is upgraded upon re-enabling/re-install
+                self.last_guids.pop(ext_handler.name, None)
```
brendandixon: @uiri The `disable` or `uninstall` could fail. Perhaps the upgradeGuid should be removed after they've succeeded?

azurelinuxagent/ga/exthandlers.py
uiri: I believe there is one potential edge case (which may not even occur in production): where a customer opts out of our new feature after observing its behavior.

There is a `None` check on `ext_handler.properties.upgradeGuid` before the call to `is_new_guid` which ensures that we get the current (pre-change) behaviour in the cases where the upgradeGuid is missing. Even without the `None` check, returning `True` would produce the desired effect.

azurelinuxagent/ga/exthandlers.py
uiri: If the `disable` or `uninstall` fails, should a subsequent `enable` proceed even without a new upgrade GUID? My initial thinking was "yes" but perhaps enable after a botched uninstall (or botched disable) is not a good idea.

brendandixon: @uiri It seems consistent to remove the GUID only when disable / uninstall succeeds.

uiri: OK, that makes sense. I made the change in the most recent commit.

PR #907, opened 2017-10-12
Sleep 1 second between polls for cloud-init created SSH keys

dwj300:
We are not making any network calls, just simply querying for a file to exist. Personally I think this could be less than 1 second, maybe 0.1?

### DIFF
--- a/azurelinuxagent/pa/provision/cloudinit.py
+++ b/azurelinuxagent/pa/provision/cloudinit.py
@@ -103,7 +103,7 @@ class CloudInitProvisionHandler(ProvisionHandler):
                              "after {1}s".format(ovf_file_path,
                                                  max_retry * sleep_time))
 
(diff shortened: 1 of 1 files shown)


azurelinuxagent/pa/provision/cloudinit.py
```diff
@@ -103,7 +103,7 @@ def wait_for_ovfenv(self, max_retry=360, sleep_time=5):
                              "after {1}s".format(ovf_file_path,
                                                  max_retry * sleep_time))
 
-    def wait_for_ssh_host_key(self, max_retry=360, sleep_time=5):
+    def wait_for_ssh_host_key(self, max_retry=360, sleep_time=1):
```
paulmey: also do `max_retry` * 5 so that we poll the same amount of time...

Issue #908, opened 2017-10-13
Set Files to 0400 in /var/lib/waagent


boumenot:
The /var/lib/waagent directory has permission 0700 and is owned by root.  There are files within this directory with permission 0644, that should be 0400. Here's the glob pattern that should be updated

 * *.p7m
 * *.pem
 * *.crt
 * *.pem
 * ovf-env.xml

brendandixon:
Addressed by #910

Issue #906, opened 2017-10-12
Hardcoded value for sshd's ClientAliveInterval (180)


ricoli:
I have noticed that the agent is setting the value of ClientAliveInterval to 180, and that this is hardcoded in https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/common/osutil/default.py#L443. This can be problematic for hardened systems that disable TCPKeepAlive and/or set ClientAliveCountMax to 0, as any value of ClientAliveInterval set before the machine is brought up will be overridden to be 180 seconds, causing ssh to disconnect after those 180 seconds of inactivity, which will catch people by surprise.

brendandixon:
@szarkos Do you want to weigh in on why the agent sets this default value? It is rather old code.

szarkos:
Azure's LB has a default timeout of 4-mins, and of course all classic VMs at least are behind a load balancer.  The goal here was to ensure SSH sessions would not timeout, which proved annoying to customers.  Some SSH clients send keepalives, but not all do which lead to customer incidents.

However, the use case here is to actively close idle connections - set ClientAliveInterval to some value (i.e. 900), and ClientAliveCountMax to 0 or similar typically to conform to a CIS requirement.

I'm not sure what the right answer is.  Maybe make this value configurable with a default of 180 seconds.  Some CIS documents suggest a value of 900 and 0 for ClientAliveInterval and ClientAliveCountMax respectively, while newer CIS guidelines suggest that "ClientAliveInterval is 300 or less and ClientAliveCountMax is 3 or less".

In any case, if the VM is behind a load balancer and the customer hasn't modified its timeout, then setting this to anything >240 seconds should have no effect.  Perhaps in practice if your goal is to have a timeout closer to 900 seconds, then setting ClientAliveInterval and ClientAliveCountMax to something more like 220 and 3 would be close to what you want even if the VM is behind a load balancer.

ricoli:
It would be great if it's at least configurable, so that one can set it to match the security policy

brendandixon:
@ricoli @szarkos We'll add an integer configuration option -- OS.SshClientAliveInterval -- whose default will be 180 seconds.

ricoli:
Wonderful :)

Issue #899, opened 2017-09-27
Improve HeartBeat Event


boumenot:
It would be nice if the HeartBeat event had an IncarnationId (random GUID) that is assigned upon every (re)start of the agent. The HeartBeat event should have a monotonically increasing counter so we can judge if there were any gaps.  This counter should reset every time IncarnationId changes.

The Windows agent uses the following format.  Please use the same one.

```text
Message = "{current goal state incarnation};{Count};{IncarnationId}"
```

Issue #898, opened 2017-09-25
Send dummy status if extension fails to write a #.status file


boumenot:
There is divergent behavior between the Windows and Linux agent with regards to missing status file.  In Linux, if the status file is missing the agent will throw an exception.

```text
No such file or directory: '/var/lib/waagent/Microsoft.Azure.Extension.My-1.0/status/0.status'
```

Windows will create a dummy status file, and send it if the status file does not exist.

The underlying issue is the extension has not completed its work by the time enable has completed. The unfortunate issue is the two agents disagree.  It would be a breaking change to undo what the Windows' agent does, so I recommend we change the behavior of Linux to conform.

PR #909, opened 2017-10-16
Address issues

brendandixon:
[#908] -- Set Files to 0400 in /var/lib/waagent
[#906] -- Hardcoded value for sshd's ClientAliveInterval (180)
[#899] -- Improve HeartBeat Event
[#898] -- Send dummy status if extension fails to write a #.status file

### DIFF
--- a/azurelinuxagent/ga/monitor.py
+++ b/azurelinuxagent/ga/monitor.py
@@ -21,8 +21,10 @@ import os
 import platform
 import time
 import threading
+import uuid
 
 import azurelinuxagent.common.conf as conf
+import azurelinuxagent.common.utils.fileutil as fileutil
 import azurelinuxagent.common.logger as logger
 
 from azurelinuxagent.common.event import add_event, WALAEventOperation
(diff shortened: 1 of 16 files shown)


azurelinuxagent/ga/monitor.py
```diff
@@ -179,15 +181,24 @@ def collect_and_send_events(self):
...
         last_heartbeat = datetime.datetime.utcnow() - period
+
+        # Create a new identifier on each restart and reset the counter
+        heartbeat_id = str(uuid.uuid4()).upper()
+        counter = 0
         while True:
             if datetime.datetime.utcnow() >= (last_heartbeat + period):
                 last_heartbeat = datetime.datetime.utcnow()
+                incarnation = protocol.get_incarnation()
+                msg = "{0};{1};{2}".format(incarnation, counter, heartbeat_id)
```
boumenot: Thank you!

azurelinuxagent/ga/update.py
```diff
@@ -76,6 +77,13 @@
 
 AGENT_SENTINAL_FILE = "current_version"
 
+READONLY_FILE_GLOBS = [
+    "*.p7m",
+    "*.pem",
+    "*.crt",
```
boumenot: *.prv ?

brendandixon: @boumenot Ah, good catch. I'll add it.

Issue #822, opened 2017-07-31
Update docs


hglkrijger:
We do not have any `LBProbeResponder` support, but we have not removed the config and documentation accordingly.

brendandixon:
Addressed by #914

PR #914, opened 2017-10-18
Re-enable firewall / address #822

brendandixon:
Disable non-root access to wireserver
[#822] -- Update docs

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -57,9 +57,16 @@ IPTABLES_VERSION = "iptables --version"
 IPTABLES_LOCKING_VERSION = FlexibleVersion('1.4.21')
 
 FIREWALL_ACCEPT = "iptables {0} -t security -{1} OUTPUT -d {2} -p tcp -m owner --uid-owner {3} -j ACCEPT"
-FIREWALL_DROP = "iptables {0} -t security -{1} OUTPUT -d {2} -p tcp -j DROP"
-FIREWALL_LIST = "iptables {0} -t security -L"
+# Note:
+# -- Initially "flight" the change to ACCEPT packets and develop a metric baseline
+#    A subsequent release will convert the ACCEPT to DROP
+FIREWALL_DROP = "iptables {0} -t security -{1} OUTPUT -d {2} -p tcp -m conntrack --ctstate INVALID,NEW -j ACCEPT"
+# FIREWALL_DROP = "iptables {0} -t security -{1} OUTPUT -d {2} -p tcp -m conntrack --ctstate INVALID,NEW -j DROP"
+FIREWALL_LIST = "iptables {0} -t security -L -nxv"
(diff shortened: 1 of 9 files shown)


azurelinuxagent/common/osutil/default.py
```diff
@@ -57,10 +57,13 @@
 IPTABLES_LOCKING_VERSION = FlexibleVersion('1.4.21')
 
 FIREWALL_ACCEPT = "iptables {0} -t security -{1} OUTPUT -d {2} -p tcp -m owner --uid-owner {3} -j ACCEPT"
-FIREWALL_DROP = "iptables {0} -t security -{1} OUTPUT -d {2} -p tcp -j DROP"
-FIREWALL_LIST = "iptables {0} -t security -L"
+FIREWALL_DROP = "iptables {0} -t security -{1} OUTPUT -d {2} -p tcp -m conntrack --ctstate INVALID,NEW -j DROP"
+FIREWALL_LIST = "iptables {0} -t security -L -xv"
```
jasonzio: Do you want to add -n to suppress reverse-name lookups? If DNS is horked up, those could take a really long time. (Same question for the next line as well.)

brendandixon: Not a bad idea.

azurelinuxagent/common/osutil/default.py
```diff
@@ -76,6 +79,31 @@ def __init__(self):
...
+
+            pattern = re.compile(PACKET_PATTERN.format(dst_ip))
+            for line in output.split('\n'):
+                m = pattern.match(line)
+                if m is not None:
+                    return int(m.group(1))
+            
+            return 0
+
+        except Exception as e:
```
boumenot: There was an issue logged where the agent was spamming the logs because the VM did not have iptables installed.  (I vaguely remember CoreOS).  If we can determine iptables is not available I would not do this and thereby avoid the log spamming.

azurelinuxagent/common/osutil/default.py
```diff
@@ -76,6 +79,31 @@ def __init__(self):
         self.selinux = None
         self.disable_route_warning = False
 
+    def get_firewall_dropped_packets(self, dst_ip=None):
```
boumenot: Is it possible to just track drops without enabling any drop actions?  I would like to gauge the telemetry before we enable the actual rules if possible.

brendandixon: Not really. iptables attributes packets / bytes dropped / accepted to individual rules. The value the code reports is the number of dropped packets caused by *this* `DROP` rule. All prior values, since the rule did not exist, will be zero. And, since this is supposed to block inappropriate access, zero (or very nearly) seems a good baseline.

azurelinuxagent/common/osutil/default.py
brendandixon: It may already be Ok, but I'll review closely.

PR #919, opened 2017-10-26
Use cached extension version manifest if available

uiri:


brendandixon:
@uiri I don't believe we can accept this without a very thorough review. The agent learns of new extension versions *only* by re-fetching the manifest. The manifest, sadly, is not versioned nor is the agent given any indicator that a new manifest is available. As a result, the agent must fetch it fresh every time just in case new versions of the extension are available.

uiri:
You are right. I should have looked at this method more closely. The manifest cache is versioned on the GoalState incarnation number. That is probably not enough to prevent this method from accidentally picking up cached versions though.

Is it ever necessary to update the version manifest in a situation where the GoalState (and thus its incarnation number) has _not_ changed?

brendandixon:
@uiri The manifest is *stored* under the incarnation number, but is not versioned by it. For example, two successive incarnations *may* retrieve the *same* manifest, but both will be stored.

The general idea is this: When a GoalState arrives, the agent should retrieve all related files (the entire document tree), including extension manifests. It should cache those to disk and use only those until the next incarnation (GoalState change) arrives.

This code predates my involvement and I'd need to review it closely to see it aligns with this goal. We'll accept changes to make it align if it does not.

uiri:
Thank you for the clarification surrounding the intended behavior of this code.

It appears to me that:

- the very first thing the ext handlers handler does is update the goal state (via calling `get_ext_handlers` on the wire protocol)
- Unless extension handling is on hold, it will go through each extension handler. This then gets called indirectly by `decide_version`

It looks like this might update the goal state in the middle of processing one of the extensions. That seems broken to me. I think a better behavior would be passing in the GoalState incarnation from the extension handler handling code. If it does that, then the cache becomes reliable.

azurelinuxagent/ga/update.py
```diff
@@ -280,10 +281,15 @@ def run(self):
                             available_agent.name)
                     break
 
-                utc_start = datetime.utcnow()
 
-                last_etag = exthandlers_handler.last_etag
-                exthandlers_handler.run()
+                while True:
```
hglkrijger: Not a fan of this construct - we will restart the process on the next iteration anyway, not sure if there is a need for this.

tests/protocol/test_wire.py
```diff
@@ -83,15 +84,20 @@ def test_getters_with_stale_goal_state(self, *args):
         test_data = WireProtocolData(DATA_FILE)
         test_data.emulate_stale_goal_state = True
 
-        self._test_getters(test_data, *args)
+        try:
+            self._test_getters(test_data, *args)
+            # Expect this to throw a ProtocolError because of stale goal state
+            self.assertTrue(False)
```
hglkrijger: assertFail?

tests/protocol/test_wire.py
```diff
@@ -83,15 +84,20 @@ def test_getters_with_stale_goal_state(self, *args):
...
+        except ProtocolError:
+            pass
         # Ensure HostPlugin was invoked
-        self.assertEqual(1, test_data.call_counts["/versions"])
-        self.assertEqual(2, test_data.call_counts["extensionArtifact"])
+        #self.assertEqual(1, test_data.call_counts["/versions"])
```
hglkrijger: Should these be commented out?

azurelinuxagent/common/exception.py
```diff
@@ -113,6 +113,10 @@ def __init__(self, msg=None, inner=None):
         super(ProtocolNotFoundError, self).__init__(msg, inner)
 
 
+class RestartError(ProtocolError):
+    pass
```
hglkrijger: Comments

azurelinuxagent/common/protocol/wire.py
```diff
@@ -614,8 +613,7 @@ def fetch_manifest(self, version_uris):
 
                 try:
                     host = self.get_host_plugin()
-                    uri, headers = host.get_artifact_request(version.uri)
-                    response = self.fetch(uri, headers, use_proxy=False)
+                    response = self.fetch(*host.get_artifact_request(version.uri), use_proxy=False)
```
hglkrijger: I prefer not to stack the function calls here, it reduces the clarity of stack traces.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -843,25 +841,24 @@ def get_ext_conf(self):
...
-                local_file = os.path.join(conf.get_lib_dir(), local_file)
                 xml_text = self.fetch_manifest(ext_handler.versionUris)
-                self.save_cache(local_file, xml_text)
-                return ExtensionManifest(xml_text)
-
             except ResourceGoneError:
-                continue
+                raise RestartError("Failed to retrieve extension manifest")
+
+            self.save_cache(local_file, xml_text)
```
hglkrijger: I think this idea is fine, but I dislike the change in caching pattern for the extension config specifically.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -350,6 +352,8 @@ def handle_ext_handler(self, ext_handler, etag):
             else:
                 message = u"Unknown ext handler state:{0}".format(state)
                 raise ExtensionError(message)
+        except RestartError:
+            raise
```
hglkrijger: Logging + telemetry needed here

azurelinuxagent/ga/exthandlers.py
```diff
@@ -474,12 +478,11 @@ def __init__(self, ext_handler, protocol):
...
 
-    def decide_version(self, target_state=None):
+    def decide_version(self, etag, target_state=None):
         self.logger.verbose("Decide which version to use")
-        try:
-            pkg_list = self.protocol.get_ext_handler_pkgs(self.ext_handler)
-        except ProtocolError as e:
-            raise ExtensionError("Failed to get ext handler pkgs", e)
+        pkg_list = self.protocol.get_ext_handler_pkgs(self.ext_handler, etag)
+        # except ProtocolError as e:
```
hglkrijger: Remove comments

azurelinuxagent/common/exception.py
uiri: Added some.

azurelinuxagent/common/protocol/wire.py
uiri: Fixed.

azurelinuxagent/common/protocol/wire.py
uiri: Could you please elaborate on what you mean by this? What is the (old/new) caching pattern for the extension config?

azurelinuxagent/ga/exthandlers.py
uiri: Added log statement here.

Would be better to do the logging and telemetry where the error actually gets processed?

azurelinuxagent/ga/exthandlers.py
uiri: Whoops.

tests/protocol/test_wire.py
uiri: Fixed.

tests/protocol/test_wire.py
uiri: They no longer get triggered because the intended behavior that this tests for has changed. Removed.

azurelinuxagent/ga/update.py
uiri: Brendan had concerns about rerunning the first part of the loop upon the goal state going stale - specifically the self update logic. I put this construct in as a stopgap until I can investigate further.

azurelinuxagent/common/protocol/wire.py
hglkrijger: Old pattern: use cached manifest, if download fails, update goal state and retry
New pattern: use cached manifest, if download fails, throw, pass up the stack, then retry.

Can you clarify the benefit of this approach?

uiri: Let's say there are 5 extensions to process - A through E.
Let's say that the manifest fails to download the manifest for extension C.

Old pattern: Process extensions A & B with Goal State N. Process extensions C through E with Goal State N+1. If Goal State N+2 or later shows up when the loop comes back around to processing extensions, the extensions aren't processed according to Goal State N+1.

New pattern: Process extensions A& B with Goal State N. Exit before processing extension C because fetching the manifest failed. Process the extensions as listed in Goal State N+1.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -840,25 +839,36 @@ def get_ext_conf(self):
                 self.ext_conf = ExtensionsConfig(xml_text)
         return self.ext_conf
 
-    def get_ext_manifest(self, ext_handler, goal_state):
+    def get_ext_manifest(self, ext_handler, incarnation):
+
+        def get_local_file(name, incarnation):
```
hglkrijger: I think you are trying to use the cache file if it exists, see `fetch_cache()`.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -153,10 +153,9 @@ def get_ext_handlers(self):
         # In wire protocol, incarnation is equivalent to ETag
         return ext_conf.ext_handlers, goal_state.incarnation
 
-    def get_ext_handler_pkgs(self, ext_handler):
+    def get_ext_handler_pkgs(self, ext_handler, etag):
         logger.verbose("Get extension handler package")
-        goal_state = self.client.get_goal_state()
-        man = self.client.get_ext_manifest(ext_handler, goal_state)
+        man = self.client.get_ext_manifest(ext_handler, etag)
```
hglkrijger: Why remove the call to `get_goal_state()` here? It is simply loading from the disk cache.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -840,25 +839,36 @@ def get_ext_conf(self):
...
 
-    def get_ext_manifest(self, ext_handler, goal_state):
+    def get_ext_manifest(self, ext_handler, incarnation):
+
+        def get_local_file(name, incarnation):
+            local_file = MANIFEST_FILE_NAME.format(name, incarnation)
+            return os.path.join(conf.get_lib_dir(), local_file)
+
+        try:
+            local_file = get_local_file(ext_handler.name, incarnation)
```
hglkrijger: I think this adds complexity. If you feel the right thing to do is check the disk cache first, add that into the code below.

azurelinuxagent/common/protocol/wire.py
uiri: The only part of the goal state which is relevant is the incarnation number (the etag). The ExtHandlersHandler uses this number to keep track of whether the extensions are current or not anyways. It is more consistent to pass this in for each ExtHandler rather than load the entire goal state IMO.

azurelinuxagent/common/protocol/wire.py
uiri: If the ExtHandlersHandler is processing Goal State N, I think the right thing to do is to check the disk cache for Goal State N. I think that checking the disk cache for Goal State N+1 is the wrong thing to do. If the disk cache has an entry for a manifest labeled with N+1 (and implicitly, not for Goal State N), it is more likely than not that something weird happened.

That said, taking another look, I think I do see a way to simplify this a bit.

PR #921, opened 2017-10-26
Stability improvements

brendandixon:
Remove safe-deploy code
Adjust delay and retry values for failed HTTP requests
Ensure current agent is not removed from disk
Add operation to extension error events
Improve tracking of child agent exceptions


azurelinuxagent/common/utils/restutil.py
```diff
@@ -216,18 +216,31 @@ def http_request(method,
...
     msg = ''
     attempt = 0
-    delay = retry_delay
+    delay = 0
+    was_throttled = False
 
     while attempt < max_retry:
         if attempt > 0:
-            logger.info("[HTTP Retry] Attempt {0} of {1}: {2}",
+            # Compute the request delay
```
boumenot: This is the second change to this code relatively quickly.  What happened with the previous assumption?

azurelinuxagent/common/event.py
```diff
@@ -52,6 +52,7 @@ class WALAEventOperation:
     Disable = "Disable"
     Download = "Download"
     Enable = "Enable"
+    ExtensionProcessing = "ExtensionHandling"
```
boumenot: Why is the constant name different from the value?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -410,7 +419,11 @@ def report_ext_handlers_status(self):
                 logger.verbose("Completed vm agent status report")
         except ProtocolError as e:
             message = "Failed to report vm agent status: {0}".format(e)
-            add_event(AGENT_NAME, version=CURRENT_VERSION, is_success=False, message=message)
+            add_event(AGENT_NAME,
+                version=CURRENT_VERSION,
+                op=WALAEventOperation.ExtensionProcessing,
```
boumenot: This will be useful.

azurelinuxagent/common/event.py
brendandixon: I can change that.

azurelinuxagent/common/utils/restutil.py
brendandixon: Sadly. The first change reduced the delay when the server throttled the request. But, in reviewing the code, I became suspicious that it might not delay long enough in edge cases. These changes bring the *total* delay times (over all attempts) back in line with the previous code while leaving the throttle induced delay short.

azurelinuxagent/ga/exthandlers.py
brendandixon: Yup.

PR #950, opened 2017-11-28
Compute duration for download and other events

### LINKED ISSUE #946: Add duration metric to download
Also look at other telemetry for places we need duration

boumenot:
Closes #946

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -36,7 +36,7 @@ import azurelinuxagent.common.utils.restutil as restutil
 import azurelinuxagent.common.utils.shellutil as shellutil
 import azurelinuxagent.common.version as version
 
-from azurelinuxagent.common.event import add_event, WALAEventOperation
+from azurelinuxagent.common.event import add_event, WALAEventOperation, elapsed_milliseconds
 from azurelinuxagent.common.exception import ExtensionError, ProtocolError, HttpError
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.version import AGENT_VERSION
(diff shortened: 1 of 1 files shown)


bin/waagent2.0
```diff
@@ -54,6 +54,8 @@ import datetime
 import xml.sax.saxutils
 from distutils.version import LooseVersion
 
+from azurelinuxagent.common.event import elapsed_milliseconds
```
hglkrijger: You can revert all these changes, no need to update 2.0

azurelinuxagent/ga/exthandlers.py
```diff
@@ -647,12 +648,13 @@ def copy_status_files(self, old_ext_handler_i):
...
 
-    def report_event(self, message="", is_success=True):
+    def report_event(self, message="", is_success=True, duration=0):
         version = self.ext_handler.properties.version
         add_event(name=self.ext_handler.name, version=version, message=message, 
-                  op=self.operation, is_success=is_success)
+                  op=self.operation, is_success=is_success, duration=duration)
 
     def download(self):
+        begin_utc = datetime.datetime.utcnow()
```
hglkrijger: For the EG data, I am not sure if this is sufficient. We may need to have start and end events for frame analysis. Could you confirm?

boumenot: EG uses the timestamp and the duration.

A start and end event would be necessary if the event never completes or is interrupted.  If that did happen you would never get the event.  If you had a start and end event you would be able to see a start without a end.  I didn't want to double the number of events sent quite yet, which is why I started with this.

hglkrijger: Ok, sounds good

boumenot:
I reverted all of the changes on bin/waagen2.0.

PR #954, opened 2017-11-29
Emit a GuestState telemetry event after provision

### LINKED ISSUE #920: Telemetry Event to Report Detected CPU and Memory
There are cases where customers are paying for X CPUs and Y Memory, but they are not actually getting it. This is due to their kernel being improperly configured, or just being too old. The telemetry we have reports what Hyper-V gave the customer, which only tells half the story. I would like the agent to report the values from /proc/cpuinfo and /proc/meminfo (or something comparable), so we can detect these types of conditions.

boumenot:
Closes #920

### DIFF
--- a/azurelinuxagent/common/event.py
+++ b/azurelinuxagent/common/event.py
@@ -63,6 +63,7 @@ class WALAEventOperation:
     Partition = "Partition"
     ProcessGoalState = "ProcessGoalState"
     Provision = "Provision"
+    GuestState = "GuestState"
     ReportStatus = "ReportStatus"
     Restart = "Restart"
     UnhandledError = "UnhandledError"
--- a/azurelinuxagent/pa/provision/default.py
+++ b/azurelinuxagent/pa/provision/default.py
@@ -282,7 +282,8 @@ class ProvisionHandler(object):
             count = len([x for x in open('/proc/cpuinfo').readlines()
                          if x.startswith("processor")])
             return count
-        except:
+        except Exception as e:
+            logger.verbose(u"Failed to determine the CPU count: {0}.", ustr(e))
             pass
         return -1
 
@@ -292,7 +293,8 @@ class ProvisionHandler(object):
                 m = re.match('^MemTotal:\s*(\d+) kB$', line)
                 if m is not None:
                     return int(int(m.group(1)) / 1024)
-        except:
+        except Exception as e:
+            logger.verbose(u"Failed to determine the memory size: {0}..", ustr(e))
             pass
         return -1
 
--- a/azurelinuxagent/pa/provision/default.py
+++ b/azurelinuxagent/pa/provision/default.py
@@ -22,6 +22,7 @@ Provision handler
 import os
 import os.path
 import re
+import socket
 import time
 
 from datetime import datetime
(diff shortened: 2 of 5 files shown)


azurelinuxagent/common/event.py
```diff
@@ -63,6 +63,7 @@ class WALAEventOperation:
     Partition = "Partition"
     ProcessGoalState = "ProcessGoalState"
     Provision = "Provision"
+    GuestState = "GuestState"
```
boumenot: I am not crazy about the name, and I am open to suggestions.

tests/pa/test_provision.py
```diff
@@ -115,6 +119,85 @@ def test_is_provisioned_not_deprovisioned(self,
         ph.osutil.is_current_instance_id.assert_called_once()
         deprovision_handler.run_changed_unique_id.assert_called_once()
 
+    @distros()
+    @patch('azurelinuxagent.common.osutil.default.DefaultOSUtil.get_instance_id',
+        return_value='B9F3C233-9913-9F42-8EB3-BA656DF32502')
+    def test_provision_telemetry_success(self, mock_util, distro_name, distro_version,
```
boumenot: Let me know if there's a better way to do this.

azurelinuxagent/pa/provision/default.py
```diff
@@ -264,12 +269,51 @@ def deploy_ssh_keypairs(self, ovfenv):
...
-                    op=WALAEventOperation.Provision)
+                    op=operation)
+
+    def get_cpu_count(self):
+        try:
+            count = len([x for x in open('/proc/cpuinfo').readlines()
+                         if x.startswith("processor")])
+            return count
+        except:
+            pass
```
hglkrijger: In the exception case, should we add some logging?

azurelinuxagent/pa/provision/default.py
```diff
@@ -88,10 +89,14 @@ def run(self):
 
             self.write_provisioned()
 
-            self.report_event("Provision succeed",
+            self.report_event("Provision succeeded",
```
hglkrijger: Thanks

azurelinuxagent/pa/provision/default.py
```diff
@@ -264,12 +269,51 @@ def deploy_ssh_keypairs(self, ovfenv):
...
+                         if x.startswith("processor")])
+            return count
+        except:
+            pass
+        return -1
+
+    def get_mem_size_mb(self):
+        try:
+            for line in open('/proc/meminfo').readlines():
+                m = re.match('^MemTotal:\s*(\d+) kB$', line)
```
hglkrijger: Is it always kB?

azurelinuxagent/pa/provision/default.py
boumenot: How about logger.verbose?

azurelinuxagent/pa/provision/default.py
boumenot: The kernel docs imply it is always this way, see [filesystem/proc.txt](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/proc.txt?h=v4.15-rc3).

PR #969, opened 2017-12-13
No more ifup/ifdown on Ubuntu >=17.10

### LINKED ISSUE #962: Add support for Ubuntu 1710
1710 does not have ifup/ifdown commands. We need to use netplan to bounce the interface.

boumenot:
Closes #962

### DIFF
--- a/azurelinuxagent/common/osutil/ubuntu.py
+++ b/azurelinuxagent/common/osutil/ubuntu.py
@@ -16,8 +16,13 @@
 # Requires Python 2.4+ and Openssl 1.0+
 #
 
+import time
+
+import azurelinuxagent.common.logger as logger
 import azurelinuxagent.common.utils.shellutil as shellutil
+
 from azurelinuxagent.common.osutil.default import DefaultOSUtil
+
 
 class Ubuntu14OSUtil(DefaultOSUtil):
     def __init__(self):
@@ -41,6 +46,7 @@ class Ubuntu14OSUtil(DefaultOSUtil):
     def get_dhcp_lease_endpoint(self):
         return self.get_endpoint_from_leases_path('/var/lib/dhcp/dhclient.*.leases')
 
+
 class Ubuntu12OSUtil(Ubuntu14OSUtil):
     def __init__(self):
         super(Ubuntu12OSUtil, self).__init__()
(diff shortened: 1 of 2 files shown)


boumenot:
I am validating this change in DCR.  Based on my reading this is the correct way to bounce the interface, and force a renew of DHCP.

azurelinuxagent/common/osutil/ubuntu.py
```diff
@@ -60,6 +70,30 @@ def register_agent_service(self):
     def unregister_agent_service(self):
         return shellutil.run("systemctl mask walinuxagent", chk_err=False)
 
+
+class UbuntuOSUtil(Ubuntu16OSUtil):
+    def __init__(self):
+        super(UbuntuOSUtil, self).__init__()
+
+    def restart_if(self, ifname, retries=3, wait=5):
```
hglkrijger: I am hesitant to change this for the 16x images as well. Can we limit to 1710?

azurelinuxagent/common/osutil/ubuntu.py
```diff
@@ -60,6 +70,30 @@ def register_agent_service(self):
...
+        """
+        Restart an interface by bouncing the link. systemd-networkd observes
+        this event, and forces a renew of DHCP.
+        """
+        retry_limit=retries+1
+        for attempt in range(1, retry_limit):
+            return_code=shellutil.run("ip link set {0} down && ip link set {0} down".format(ifname))
```
hglkrijger: && link up

azurelinuxagent/common/osutil/ubuntu.py
```diff
@@ -60,6 +70,30 @@ def register_agent_service(self):
...
+            if return_code == 0:
+                return
+            logger.warn("failed to restart {0}: return code {1}".format(ifname, return_code))
```
hglkrijger: Please add dcr tests before merging

azurelinuxagent/common/osutil/ubuntu.py
```diff
@@ -50,7 +56,11 @@ def get_dhcp_pid(self):
         ret = shellutil.run_get_output("pidof dhclient3", chk_err=False)
         return ret[1] if ret[0] == 0 else None
 
-class UbuntuOSUtil(Ubuntu14OSUtil):
+
+class Ubuntu16OSUtil(Ubuntu14OSUtil):
```
hglkrijger: This will be either 17 or 18, I think. Convention has been to use LTS versions but I am ok either way.

boumenot: I took the approach that the latest version of Ubuntu is current, and if you need to make a change you use the version number.

 * UbuntuOSUtil is for >= 17.10
 * Ubuntu16OSUtil was created an now represents previous versions.  

This seems to be the style of the current code.

azurelinuxagent/common/osutil/ubuntu.py
boumenot: This change is only for 17.10 and above.

azurelinuxagent/common/osutil/ubuntu.py
boumenot: What is not already covered, that needs to be covered by this change?

PR #970, opened 2017-12-13
more debug for download failures

### LINKED ISSUE #945: Download operation failure
```Exception retrieving agent manifests: [ProtocolError] /var/lib/waagent/GoalState.1.xml is missing.```

boumenot:
Closes #945


boumenot:
I think this is the change we discussed.  Upon another look, and I not sure it is sufficient.

PR #971, opened 2017-12-13
Report prolonged report status failures

### LINKED ISSUE #956: Keep track of put status state
We should emit telemetry if a vm cannot report status successfully with a certain timeframe, this will help with alerting and health monitoring. Perhaps 15 minutes?

Suggest we add a timestamp in memory to `wire.py` and update/report it in or around this method: https://github.com/Azure/WALinuxAgent/blob/ab261b5f2ed0dfce4dd94adf924e68e0aaff5e7b/azurelinuxagent/common/protocol/wire.py#L189

boumenot:
The agent report transient failures, but this is not very indicative of
customer failures.  The transient failures generally resolve themselves, so the
transients are not very useful for a dashboard.  This change tracks failures
that repeat for 15 minutes, and then report them.  If an operation fails and
succeeds within 15 minutes, the timer resets.  It must be a continual stream of
errors for 15 minutes for this telemetry event to trigger.

Closes #956

### DIFF
--- a/azurelinuxagent/common/errorstate.py
+++ b/azurelinuxagent/common/errorstate.py
@@ -1,0 +1,30 @@
+from datetime import datetime, timedelta
+
+ERROR_STATE_DELTA = timedelta(minutes=15)
+
+class ErrorState(object):
+    def __init__(self, min_timedelta):
+        self.min_timedelta = min_timedelta
+
+        self.count = 0;
+        self.timestamp = None
+
+    def incr(self):
+        if self.count == 0:
+            self.timestamp = datetime.utcnow()
+
+        self.count += 1
+
+    def reset(self):
+        self.count = 0
+        self.timestamp = None
+
+    def is_triggered(self):
+        if self.timestamp is None:
+            return False
+
+        delta = datetime.utcnow() - self.timestamp
+        if delta >= self.min_timedelta:
+            return True
+
+        return False
--- a/azurelinuxagent/common/errorstate.py
+++ b/azurelinuxagent/common/errorstate.py
@@ -2,11 +2,12 @@ from datetime import datetime, timedelta
 
 ERROR_STATE_DELTA = timedelta(minutes=15)
 
+
 class ErrorState(object):
-    def __init__(self, min_timedelta):
+    def __init__(self, min_timedelta = ERROR_STATE_DELTA):
         self.min_timedelta = min_timedelta
 
-        self.count = 0;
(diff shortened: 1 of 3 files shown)


azurelinuxagent/common/errorstate.py
```diff
@@ -0,0 +1,30 @@
+from datetime import datetime, timedelta
+
+ERROR_STATE_DELTA = timedelta(minutes=15)
```
hglkrijger: Is this used?

azurelinuxagent/common/errorstate.py
```diff
@@ -0,0 +1,30 @@
...
+
+class ErrorState(object):
+    def __init__(self, min_timedelta):
```
hglkrijger: Might be cleaner to just manage the delta completely internally in this class, using the constant above.

azurelinuxagent/common/errorstate.py
hglkrijger: Should also be in constants, probably

azurelinuxagent/common/errorstate.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def reset(self):
+        self.count = 0
+        self.timestamp = None
+
+    def is_triggered(self):
+        if self.timestamp is None:
+            return False
+
+        delta = datetime.utcnow() - self.timestamp
```
hglkrijger: This seems to rely on the caller being well behaved. Is it better to have a model where the error state is dependent on the caller or should there be some sort of model where error state manages its own lifecycle? On the one hand I do not want to over-complicate things, and on the other we will miss certain classes of issues (deadlock, e.g.).

azurelinuxagent/common/errorstate.py
boumenot: Where?  There are quite a few constants, but I do not see one particular file dedicated to just constants.

azurelinuxagent/common/errorstate.py
boumenot: I like to be able to control this value for the purposes of testing. If I keep it in the class I cannot control it easily.

azurelinuxagent/common/errorstate.py
boumenot: Discussed this in person, and decided to go with it as it is now.  These are all valid points, but we should be able to learn with what we have. I can make improvements in the future based on our learnings.

PR #979, opened 2017-12-20
Bug fix for upgrade GUID change

uiri:
Reduce log spam from skipping an upgrade based on the GUID staying the same.

Ensure that the Agent reports the correct extension version and its state as enabled in cases where the GUID is the same.

Fix unit testing for the upgrade GUID feature

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -330,6 +330,8 @@ class ExtHandlersHandler(object):
                     not self.is_new_guid(ext_handler):
                 ext_handler_i.ext_handler.properties.version = ext_handler_i.get_installed_version()
                 ext_handler_i.set_logger()
+                if self.last_etag != etag:
+                    self.set_log_guid(ext_handler, True)
 
                 if self.get_log_guid(ext_handler):
                     ext_handler_i.logger.info("New GUID is the same as the old GUID. Exiting without upgrading.")
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -177,7 +177,7 @@ class ExtHandlersHandler(object):
         self.protocol = None
         self.ext_handlers = None
         self.last_etag = None
-        self.last_guids = {}
+        self.last_upgrade_guids = {}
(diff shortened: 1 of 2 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -223,7 +223,15 @@ def run_status(self):
         return
 
     def get_guid(self, name):
-        return self.last_guids.get(name, None)
+        return self.last_guids.get(name, (None, False))[0]
```
hglkrijger: should you check for None before reference? also the method below

azurelinuxagent/ga/exthandlers.py
```diff
@@ -223,7 +223,15 @@ def run_status(self):
...
+
+    def get_log_guid(self, ext_handler):
```
hglkrijger: consider renaming `log_guid`, I think it is a bit misleading

azurelinuxagent/ga/exthandlers.py
```diff
@@ -223,7 +223,15 @@ def run_status(self):
...
+        return self.last_guids.get(ext_handler.name, (None, False))[1]
+
+    def set_log_guid(self, ext_handler, log_val):
+        guid = self.get_guid(ext_handler.name)
+        if guid is not None:
+            self.last_guids[ext_handler.name] = (guid, log_val)
```
hglkrijger: Consider renaming `last_guids` as well, the context is not obvious to the reader

azurelinuxagent/ga/exthandlers.py
uiri: Fixed.

azurelinuxagent/ga/exthandlers.py
uiri: fixed.

azurelinuxagent/ga/exthandlers.py
uiri: `last_upgrade_guids` is initialized to an empty dictionary when the ExtHandlerHandler is initialized.
The second parameter to `get` specifies what is returned if `name` is not found in the dictionary.

PR #982, opened 2017-12-30
Update main.py

digeler:
Since we have a lot of issues "Unable to ssh to linux vm"
some of these cases happens because /etc/udev/rules.d/70-persistent-net.rules
is not having the correct interface entry.
i changed the agent.py to dump the rules:


hglkrijger:
These changes are not robust, nor well placed.

PR #989, opened 2018-01-04
Emit log message when work is done

hglkrijger:
Only emit log messages for `ProcessGoalState` when install, enable, uninstall or disable runs.


boumenot:
LGTM if the DCR passes.

PR #991, opened 2018-01-04
ship warning and errors logs to Geneva

boumenot:
Upload logs as telemetry events to Geneva.
Appease the PEP gods on a few things.

### DIFF
--- a/azurelinuxagent/agent.py
+++ b/azurelinuxagent/agent.py
@@ -63,6 +63,9 @@ class Agent(object):
                                  path="/var/log/waagent.log")
         logger.add_logger_appender(logger.AppenderType.CONSOLE, level,
                                  path="/dev/console")
+        logger.add_logger_appender(logger.AppenderType.TELEMETRY,
+                                   logger.LogLevel.WARNING,
+                                   path=event.add_log_event)
 
         ext_log_dir = conf.get_ext_log_dir()
         try:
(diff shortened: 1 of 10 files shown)


azurelinuxagent/agent.py
```diff
@@ -63,6 +63,9 @@ def __init__(self, verbose, conf_file_path=None):
                                  path="/var/log/waagent.log")
         logger.add_logger_appender(logger.AppenderType.CONSOLE, level,
                                  path="/dev/console")
+        logger.add_logger_appender(logger.AppenderType.TELEMETRY,
+                                   logger.LogLevel.WARNING,
+                                   path=event.add_log_event)
```
boumenot: This is a little bit of a hack.  I am open to ideas.

I overload the path parameter to pass a lambda that actually does the logging. I did this because this method calls 3 more methods all with the same parameters so the appropriate log appender is added.  I could have created appropriate overloads, but this *seemed* better.

I also use a lambda because I didn't want to create a situation where logger.py included event.py, which already includes logger.py.

azurelinuxagent/common/event.py
```diff
@@ -225,6 +227,20 @@ def add_event(self,
         except EventError as e:
             logger.error("{0}", e)
 
+    def add_log_event(self, level, message):
+        event = TelemetryEvent(7, "FFF0196F-EE4C-4EAF-9AA5-776F622DEB4F")
+        event.parameters.append(TelemetryEventParam('EventName', WALAEventOperation.Log))
+        event.parameters.append(TelemetryEventParam('CapabilityUsed', logger.LogLevel.STRINGS[level]))
+        event.parameters.append(TelemetryEventParam('Context1', ''))
```
boumenot: This should be the time in UTC, to ensure we properly record the value.  The timestamp of the event may differ from the timestamp of when it happened.

PR #1000, opened 2018-01-11
Remove HttpErrors from log

hglkrijger:
Switch http errors telemetry to INFO so they will not be logged.


boumenot:
LGTM.

I assume this is because these issues are transient, and we will simply retry.  If we exhaust retries, will an error be logged?

hglkrijger:
Yes, exactly.

PR #1005, opened 2018-01-17
support '=' is configuration values

boumenot:


azurelinuxagent/common/conf.py
```diff
@@ -87,7 +87,7 @@ def load_conf_from_file(conf_file_path, conf=__conf__):
 
 __SWITCH_OPTIONS__ = {
     "OS.AllowHTTP" : False,
-    "OS.EnableFirewall" : False,
+    "OS.EnableFirewall" : True,
```
hglkrijger: Did you mean to change this?

boumenot: No.  I fixed it to fix the tests, but that was the wrong fix.  Let me make the correct fix.

PR #1009, opened 2018-01-18
Check for minimum iptables version

### LINKED ISSUE #976: Clean up firewall rules
Found in logs. It is possible to have firewall rules in an unexpected state, which we do not handle:
```
Chain OUTPUT (policy ACCEPT 0 packets, 0 bytes)
    pkts      bytes target     prot opt in     out     source               destination         
      12     5688 ACCEPT     tcp  --  *      *       0.0.0.0/0            168.63.129.16        owner UID match 0
      18      936 DROP       tcp  --  *      *       0.0.0.0/0            168.63.129.16       
       0        0 ACCEPT     tcp  --  *      *       0.0.0.0/0            168.63.129.16        owner UID match 0
       0        0 ACCEPT     tcp  --  *      *       0.0.0.0/0            168.63.129.16        ctstate INVALID,NEW
```
We need to determine a safe way of cleaning up rules before enabling drops.

### LINKED ISSUE #977: Repeated firewall logs (CentOS 6.5 HPC image + WALA 2.2.19)
**Repro Steps:**
1. Deploy CentOS 6.5 HPC VM.
2. Upgrade WALA to latest using master branch.
3. Restart.

**Issue -**
1. waagent keeps dumping firewall related messages to  /var/log/waagent.log infinitely.
2. I've seen waagent.log file grow beyond 4GB.
3. If we allow this state, OS will run out of disk space.

boumenot:
Only enable the firewall if iptables is at least 1.4.11 or greater.

Closes #976
Closes #977

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -167,10 +167,14 @@ class DefaultOSUtil(object):
 
             # If the DROP rule exists, make no changes
             drop_rule = FIREWALL_DROP.format(wait, "C", dst_ip)
-
-            if shellutil.run(drop_rule, chk_err=False) == 0:
+            rc = shellutil.run(drop_rule, chk_err=False)
+            if rc == 0:
                 logger.verbose("Firewall appears established")
                 return True
+            elif rc == 2:
+                msg = "please upgrade iptables to a version that supports the -C option"
+                logger.warn(msg)
+                raise Exception(msg)
 
             # Otherwise, append both rules
             accept_rule = FIREWALL_ACCEPT.format(wait, "A", dst_ip, uid)
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -66,6 +66,10 @@ FIREWALL_LIST = "iptables {0} -t security -L -nxv"
 FIREWALL_PACKETS = "iptables {0} -t security -L OUTPUT --zero OUTPUT -nxv"
 FIREWALL_FLUSH = "iptables {0} -t security --flush"
 
+# Precisely delete the rules created by the agent.
+FIREWALL_DELETE_CONNTRACK = "iptables {0} -t security -D OUTPUT -d {1} -p tcp -m conntrack --ctstate INVALID,NEW -j ACCEPT"
+FIREWALL_DELETE_OWNER = "iptables {0} -t security -D OUTPUT -d {1} -p tcp -m owner --uid-owner {2} -j ACCEPT"
+
 PACKET_PATTERN = "^\s*(\d+)\s+(\d+)\s+DROP\s+.*{0}[^\d]*$"
 
 _enable_firewall = True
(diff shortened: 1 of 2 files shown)


boumenot:
This PR is not sufficient.  The iptables on CentOS 6.9 is 1.4.7, but it does support the -C option.  It appears that they backported this work, so a code change that relies solely on the version number is insufficient.

azurelinuxagent/common/exception.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, msg=None, inner=None):
 
 class AgentNetworkError(AgentError):
     """
-    When network is not avaiable.
+    When network is not available\.
```
hglkrijger: `\` at the end

azurelinuxagent/agent.py
```diff
@@ -63,6 +63,9 @@ def __init__(self, verbose, conf_file_path=None):
                                  path="/var/log/waagent.log")
         logger.add_logger_appender(logger.AppenderType.CONSOLE, level,
                                  path="/dev/console")
+        logger.add_logger_appender(logger.AppenderType.TELEMETRY,
```
hglkrijger: Do you need to merge/rebase?

azurelinuxagent/common/osutil/default.py
```diff
@@ -129,24 +133,40 @@ def get_firewall_will_wait(self):
                 else ""
         return wait
 
-    def remove_firewall(self):
+    def _delete_rule(self, rule):
+        """
+        Continually execute the delete operation until the return
+        code is non-zero.
+        """
+        while True:
```
hglkrijger: `while True` makes me nervous; can we put in another escape hatch?

azurelinuxagent/ga/update.py
```diff
@@ -284,10 +284,7 @@ def run(self):
                 utc_start = datetime.utcnow()
 
                 last_etag = exthandlers_handler.last_etag
-                try:
-                    exthandlers_handler.run()
-                except RestartError:
```
hglkrijger: Why remove this? Or do you need to sync?

azurelinuxagent/ga/update.py
```diff
@@ -299,7 +296,7 @@ def run(self):
                         duration=elapsed_milliseconds(utc_start),
                         message="Incarnation {0}".format(
                                             exthandlers_handler.last_etag),
-                        log_event=exthandlers_handler.log_process)
+                        log_event=True)
```
hglkrijger: Why remove this?

azurelinuxagent/ga/update.py
boumenot: I need to sync.

azurelinuxagent/ga/update.py
boumenot: I need to sync.

azurelinuxagent/common/osutil/default.py
boumenot: Give me some suggestions.  What do you typically do?

azurelinuxagent/common/exception.py
boumenot: I need to sync.

azurelinuxagent/agent.py
boumenot: Yes.

azurelinuxagent/common/osutil/default.py
hglkrijger: Count or time, so keep removing rules until done or we hit some large number, for example.

PR #1018, opened 2018-01-25
ignore transient iptables error

boumenot:


boumenot:
Let's discuss.  This may not be the correct or best fix

PR #1030, opened 2018-02-03
Fix brokenness caused by building an extension in Windows

uiri:
If the extension zip was generated on Windows, it may not have the
proper Unix permission bits set. It may have weird newlines which need
to be replaced.

Perform python equivalent of chmod a+x. Reading and writing the file to
a python string is sufficient to fix the weird newline characters.

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -945,9 +945,22 @@ class ExtHandlerInstance(object):
         begin_utc = datetime.datetime.utcnow()
         self.logger.verbose("Launch command: [{0}]", cmd)
         base_dir = self.get_base_dir()
+        base_dir_and_cmd = base_dir + "/" + cmd
+        try:
+            filename = base_dir_and_cmd.split(" ", 1)[0]
+            file_stat = os.stat(filename)
+            os.chmod(filename, file_stat.st_mode | stat.S_IEXEC)
+            with open(filename, "r") as f:
+                contents = f.read()
+            # Python automatically converts Windows newlines to Linux newlines
+            with open(filename, "w") as f:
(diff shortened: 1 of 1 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -945,9 +945,22 @@ def launch_command(self, cmd, timeout=300):
...
         base_dir = self.get_base_dir()
+        base_dir_and_cmd = base_dir + "/" + cmd
+        try:
+            filename = base_dir_and_cmd.split(" ", 1)[0]
+            file_stat = os.stat(filename)
+            os.chmod(filename, file_stat.st_mode | stat.S_IEXEC)
+            with open(filename, "r") as f:
+                contents = f.read()
+            # Python automatically converts Windows newlines to Linux newlines
+            with open(filename, "w") as f:
```
hglkrijger: Not sure if we want this behavior by default, seems like unwanted IO. Even if we do it only when required, should we be passing the cost on to the customer? Better to fail early so extension publishers see this in their testing. We can fail more gracefully but I believe the onus should be on publishers to give us valid packages.

uiri: Do you mean specifically just the line ending replacement?
Or do you mean the executable bit fix as well?

I am not sure what can be done about Unix file permissions when the zip was created on Windows. Of course, the line ending problem is easy enough to require be fixed on the extension side.

hglkrijger: More the line ending replacement, but in general I think the point stands for both. Is there a reason these extensions would need to built on Windows? I am not aware of any extension publishers which have asked for this.

uiri:
Closing as this patch is no longer necessary.

PR #469, opened 2016-10-20
Add Clear Linux distro support

bryteise:

PR #1034, opened 2018-02-07
Allow detection of Clear Linux OS in python3

bryteise:
In python3 the platform.linux_distribution() function does not display
content for Clear Linux. In addition the linux_distribution function
is due to be deprecated in python 3.7 (the next minor release, see
https://bugs.python.org/issue1322). In anticipation of this, switch to
sourcing the os-release file in the waagent.service and reading the
distribution information from the environment instead.

### DIFF
--- a/azurelinuxagent/common/version.py
+++ b/azurelinuxagent/common/version.py
@@ -86,8 +86,22 @@ def get_distro():
                                                   supported_dists=supported))
         full_name = platform.linux_distribution()[0].strip()
         osinfo.append(full_name)
-    else:
+    elif 'dist' in dir(platform):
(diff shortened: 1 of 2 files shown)


bryteise:
@hglkrijger So I'm not sure if somebody will see this without me tagging but the CLA bot isn't letting me pass but I'm signed off from another agreement. See https://github.com/Azure/WALinuxAgent/pull/469#issuecomment-255478762.

hglkrijger:
We have another PR in progress for this - #1036, will that meet your needs?

bryteise:
Ah yes that will work perfectly, I'll just need to change the service in that case. Thanks!

Issue #1015, opened 2018-01-24
distro detection


rjschwei:
Upstream Python is deprecating platform.linux_distribution() which is used during install of the agent, and the code will disappear in Python 3.7 [1].  There is a new external module [2]. Should the agent depend on the external module? 

[1] https://docs.python.org/3.6/library/platform.html#platform.linux_distribution
[2] https://github.com/nir0s/distro#distro---a-linux-os-platform-information-api

hglkrijger:
@rjschwei, apologies for the delay. Yes, it should - are you planning to switch the agent over to Python 3?

rjschwei:
@hglkrijger Yes, we are switching to Python 3. For openSUSE Leap 15 and SLE 15 the default Python interpreter will be Python 3. In SLE 15 Python 2 will only be available in the "Legacy Module" which means support end two years after initial release.

hglkrijger:
Good to know, thanks @rjschwei.

boumenot:
Sorry for the delay on this.  We have raised this issue with the other distros to ensure there are no surprises at packaging time, and to ensure it's suitable for them too.

boumenot:
The fix was merged.

PR #1036, opened 2018-02-08
- Obtain platform information in Python version dependent way

rjschwei:
+ The platform module has deprecated functionality in Python 3.5 and
    the functionality will be removed in 3.7. For Python versions 3.5
    and greater depend on the distro module to obtain linux distribution
    information

### DIFF
--- a/azurelinuxagent/common/future.py
+++ b/azurelinuxagent/common/future.py
@@ -2,7 +2,7 @@ import platform
 import sys
 
 # Note broken dependency handling to avoid potential backward
-# compantibility issues
+# compatibility issues on different distributions
 try:
     import distro
 except Exception:
@@ -46,15 +46,15 @@ def get_linux_distribution(get_full_name, supported_dists):
             )
         )
         if not osinfo or osinfo == ['','','']:
-            return get_linux_ditribution_from_distro(get_full_name)
-        full_name = platform_module.linux_distribution()[0].strip()
+            return get_linux_distribution_from_distro(get_full_name)
+        full_name = platform.linux_distribution()[0].strip()
         osinfo.append(full_name)
-    except Exception:
-        return get_linux_ditribution_from_distro(get_full_name)
+    except AttributeError:
+        return get_linux_distribution_from_distro(get_full_name)
 
     return osinfo
 
-def get_linux_ditribution_from_distro(get_full_name):
+def get_linux_distribution_from_distro(get_full_name):
     """Get the distribution information from the distro Python module."""
     # If we get here we have to have the distro module, thus we do
     # not wrap the call in a try-except block as it would mask the problem
--- a/azurelinuxagent/common/future.py
+++ b/azurelinuxagent/common/future.py
@@ -1,8 +1,12 @@
 import platform
 import sys
 
-if float(sys.version[:3]) >= 3.5:
+# Note broken dependency handling to avoid potential backward
+# compantibility issues
+try:
     import distro
+except Exception:
+    pass
     
 """
 Add alias for python2 and python3 libs and functions.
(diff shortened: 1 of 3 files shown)


rjschwei:
This addresses #1015 other implementation options exists. I think this is probably a reasonable option to start additional discussions.

rjschwei:
@hglkrijger something like this?

azurelinuxagent/common/future.py
```diff
@@ -24,3 +28,24 @@
 
 else:
     raise ImportError("Unknown python version: {0}".format(sys.version_info))
+
+
+def get_linux_distribution(get_full_name, supported_dists):
+    """Abstract platform.linux_distribution() call which is deprecated as of
+       Python 3.5"""
+    if float(sys.version[:3]) >= 3.5:
```
hglkrijger: This works. We also have the option of defining this method twice, one for `< 3.7` and one for `>= 3.7`. Depending on how the Python versions and deprecations change over time, one or the other may end up being more maintainable.

setup.py
```diff
@@ -198,6 +199,9 @@ def run(self):
             osutil.stop_agent_service()
             osutil.start_agent_service()
 
+requires = []
+if float(sys.version[:3]) >= 3.5:
```
hglkrijger: Perhaps we can substitute the version (or even the check and version) for a constant like `PLATFORM_REMOVED` or `PLATFORM_AVAILABLE`

azurelinuxagent/common/future.py
```diff
@@ -1,5 +1,9 @@
+import platform
 import sys
 
+if float(sys.version[:3]) >= 3.5:
```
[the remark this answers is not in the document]
rjschwei: Function is deprecated in 3.5, I think we should use that version. Anyway on SUSE in 3.6 the return frrom the platform module is already broken, i.e. it returns an empty strings as in ('','','') so if I set this to 3.7 that does not help me at all.

azurelinuxagent/common/future.py
rjschwei: As mentioned 3.7 does not work as check in SUSE, I am game to implement 2 methods or just on get_linux_distribution() and then have that be a redirector to _get_linux_distribution_py2() and _get_linux_distribution_py35p()

setup.py
rjschwei: Sorry I don't get this??

azurelinuxagent/common/future.py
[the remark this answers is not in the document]
hglkrijger: In that case we will have to have a distro specific abstraction, since this change would break Ubuntu.

azurelinuxagent/common/future.py
hglkrijger: I am good with either, your call.

setup.py
hglkrijger: Rather than having a specific version number with a comment, it might be more readable if we use a constant and/or helper method which indicates what the version check is for, something like:
```
if float() >= PLATFORM_DEPRECATED
```
or
```
if PLATFORM_SUPPORTED_VERSION
```
Just a thought for maintainability.

azurelinuxagent/common/future.py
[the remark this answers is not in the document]
rjschwei: Problem with "distro specific abstraction" is that we cannot detect the distribution it's a chicken and egg problem.

But I am not certain about why this would break other distros:
- If the user installs from source with pip then pip will pull in the distro package if python version > 3.5
- If other distros create a new package for the new version of WALinuxAgent then they need to also create a package for distro and release this together
- If the user installs from source using setup.py then the user gets an error if distro is not installed.

I do not see how making this retroactive to 3.5 breaks anything.

It is possible to not check the version at all and just rely on exception handling and value checking. However since the changes are ties to a specific Python version I think having the version checks makes things more explicit.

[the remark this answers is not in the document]
hglkrijger: Self-update is the other channel, which will break.

rjschwei:
@hglkrijger OK, now without Python version check. Behavior is purely based on implementation of the platform module and it's return values.

This avoids the chicken and egg problem at the cost of users/packages needing to have more knowledge of the dependencies.

azurelinuxagent/common/future.py
```diff
@@ -1,5 +1,13 @@
+import platform
 import sys
 
+# Note broken dependency handling to avoid potential backward
+# compantibility issues
```
hglkrijger: typo

azurelinuxagent/common/future.py
```diff
@@ -24,3 +32,43 @@
...
+                full_distribution_name=get_full_name,
+                supported_dists=supported
+            )
+        )
+        if not osinfo or osinfo == ['','','']:
+            return get_linux_ditribution_from_distro(get_full_name)
+        full_name = platform_module.linux_distribution()[0].strip()
+        osinfo.append(full_name)
+    except Exception:
+        return get_linux_ditribution_from_distro(get_full_name)
```
hglkrijger: Might want to check the exception details before forwarding the call, or something like this instead:
```
if 'linux_distribution' in dir(platform)
```

azurelinuxagent/common/future.py
```diff
@@ -24,3 +32,43 @@
...
+
+    return osinfo
+
+def get_linux_ditribution_from_distro(get_full_name):
+    """Get the distribution information from the distro Python module."""
+    # If we get here we have to have the distro module, thus we do
+    # not wrap the call in a try-except block as it would mask the problem
+    # and result in a broken agent installation
+    osinfo = list(
```
hglkrijger: Perhaps the import should be in here?

setup.py
```diff
@@ -198,6 +199,14 @@ def run(self):
             osutil.stop_agent_service()
             osutil.start_agent_service()
 
+# Note to packagers and users fromn source.
+# In version 3.5 of Python distribution information handling in the platform
+# module was deprecated. Depending on the Linux distribution the
+# implementation may be broken prior to Python 3.7 thus you may or may not
+# need to install the distro Python package. Due to concerns of breakage
+# the code cannot handle this automagically
+#if float(sys.version[:3]) >= 3.5:
```
hglkrijger: We could do this for 3.7, though?

azurelinuxagent/common/future.py
```diff
@@ -24,3 +32,43 @@
...
+        if not osinfo or osinfo == ['','','']:
+            return get_linux_ditribution_from_distro(get_full_name)
+        full_name = platform_module.linux_distribution()[0].strip()
+        osinfo.append(full_name)
+    except Exception:
+        return get_linux_ditribution_from_distro(get_full_name)
+
+    return osinfo
+
+def get_linux_ditribution_from_distro(get_full_name):
```
hglkrijger: typo

rjschwei:
@hglkrijger hopefully  starting to converge now

boumenot:
Sorry for the delay on this.  We have raised this issue with the other distros to ensure there are no surprises at packaging time, and to ensure it's suitable for them too.

PR #1040, opened 2018-02-12
add User-Agent to HTTP requests

### LINKED ISSUE #1027: Add User-Agent String to Storage Request
Please include the agent name and version in all package requests.

boumenot:
Closes #1027

### DIFF
--- a/azurelinuxagent/common/utils/restutil.py
+++ b/azurelinuxagent/common/utils/restutil.py
@@ -28,8 +28,7 @@ import azurelinuxagent.common.utils.textutil as textutil
 
 from azurelinuxagent.common.exception import HttpError, ResourceGoneError
 from azurelinuxagent.common.future import httpclient, urlparse, ustr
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/utils/restutil.py
```diff
@@ -78,6 +77,7 @@
 
 HTTP_PROXY_ENV = "http_proxy"
 HTTPS_PROXY_ENV = "https_proxy"
+HTTP_USER_AGENT = "{0}/{1}".format(AGENT_NAME, CURRENT_VERSION)
```
hglkrijger: I think you actually want the `GOAL_STATE_AGENT_VERSION`. There is some complexity because you are not guaranteed it is correct, however you might just include both for simplicity.

boumenot: Agreed.

I am more concerned with having WALinuxAgent than I am with the version at the moment.  Let's see if this change actually brings any benefit for PIR, and we can iterate if necessary.

Issue #1035, opened 2018-02-08
[2.2.21]"RuntimeError: maximum recursion depth exceeded" if events > 999


yuxisun1217:
Description of problem:
If events > 999, WALA keeps printing "WARNING Too many files under: /var/lib/waagent/events, removing oldest" and finally raise exception: "RuntimeError: maximum recursion depth exceeded"
WALA-2.2.20 doesn't have this issue

Version-Release number of selected component (if applicable):
WALinuxAgent-2.2.21

RHEL Version:
RHEL-7.4/6.9
How reproducible: 
100%

Steps to Reproduce: 
1. Prepare a RHEL7.4 VM in Azure. Install WALA-2.2.21
2. Create 1000 files under /var/lib/waagent/events:
 service waagent stop
 cd /var/lib/waagent/events
 touch test{0001..1000}
 service waagent start
3. Check /var/log/waagent

Actual results: 
```
2018/02/07 21:54:56.289956 INFO Start env monitor service.
2018/02/07 21:54:56.342659 WARNING Too many files under: /var/lib/waagent/events, removing oldest
2018/02/07 21:54:56.438029 WARNING Too many files under: /var/lib/waagent/events, removing oldest
...(hundreds of the same log)
2018/02/07 21:55:19.748939 WARNING Too many files under: /var/lib/waagent/events, removing oldest
2018/02/07 21:55:19.820634 ERROR Failed to run 'run-exthandlers': Traceback (most recent call last):
  File "/usr/lib/python2.6/site-packages/azurelinuxagent/agent.py", line 167, in main
    agent.run_exthandlers()
  File "/usr/lib/python2.6/site-packages/azurelinuxagent/agent.py", line 132, in run_exthandlers
    update_handler.run()
  File "/usr/lib/python2.6/site-packages/azurelinuxagent/ga/update.py", line 308, in run
    logger.warn(msg)
  File "/usr/lib/python2.6/site-packages/azurelinuxagent/common/logger.py", line 195, in warn
    DEFAULT_LOGGER.warn(msg_format, *args)
  File "/usr/lib/python2.6/site-packages/azurelinuxagent/common/logger.py", line 62, in warn
    self.log(LogLevel.WARNING, msg_format, *args)
  File "/usr/lib/python2.6/site-packages/azurelinuxagent/common/logger.py", line 87, in log
    appender.write(level, log_item)
[...]

PR #1042, opened 2018-02-12
disable telemetry logging for now

boumenot:
The chance of encountering #1035 is believed to be very low, but this type of issue means recovery is not possible.  We are hotfixing the agent, and disabling telemetry collection as a result.

### DIFF
--- a/azurelinuxagent/agent.py
+++ b/azurelinuxagent/agent.py
@@ -63,9 +63,9 @@ class Agent(object):
                                  path="/var/log/waagent.log")
         logger.add_logger_appender(logger.AppenderType.CONSOLE, level,
                                  path="/dev/console")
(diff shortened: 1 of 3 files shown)


azurelinuxagent/agent.py
```diff
@@ -63,9 +63,9 @@ def __init__(self, verbose, conf_file_path=None):
                                  path="/var/log/waagent.log")
         logger.add_logger_appender(logger.AppenderType.CONSOLE, level,
                                  path="/dev/console")
-        logger.add_logger_appender(logger.AppenderType.TELEMETRY,
-                                   logger.LogLevel.WARNING,
-                                   path=event.add_log_event)
+        # logger.add_logger_appender(logger.AppenderType.TELEMETRY,
```
hglkrijger: Maybe a comment or issue number for reference here

PR #1043, opened 2018-02-12
restart monitor and environment thread if they die

### LINKED ISSUE #1041: Ensure the Monitor Thead does not Die
The monitoring thread is not (ironically) not monitored.  There have been cases where VMs stop sending telemetry data, which is the responsibility of the monitoring thread.  The working theory is the thread died, and was not automatically restarted.

boumenot:
Closes #1041

### DIFF
--- a/azurelinuxagent/ga/monitor.py
+++ b/azurelinuxagent/ga/monitor.py
@@ -94,13 +94,19 @@ class MonitorHandler(object):
         self.osutil = get_osutil()
         self.protocol_util = get_protocol_util()
         self.sysinfo = []
+        self.event_thread = None
 
     def run(self):
         self.init_sysinfo()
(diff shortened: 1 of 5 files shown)


azurelinuxagent/ga/monitor.py
```diff
@@ -94,13 +94,19 @@ def __init__(self):
...
         self.init_sysinfo()
+        self.restart()
 
-        event_thread = threading.Thread(target=self.daemon)
-        event_thread.setDaemon(True)
-        event_thread.start()
+    def is_alive(self):
+        return self.event_thread.is_alive()
+
+    def restart(self):
```
hglkrijger: `restart` implies stop/start (to me anyway), perhaps just `start` or similar

PR #1044, opened 2018-02-12
reduce enable calls

### LINKED ISSUE #944: Reduce enable calls
On an already running VM with a running guest agent, we should change goal-state processing such that we *only* call Enable on extensions if there is an actual config change for that particular extension. 

For example, if a running VM has 3 extensions (A, B and C), if a goal-state update comes down where no extension config changes, then no Enable should be called.  If a goal-state update comes down where the config changes only for B, then Enable should only be called on B, but not on A and C.  

Upon a re-start of the guest agent, we should call Enable again on all extensions once.

boumenot:
Closes #944

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -91,6 +91,7 @@ def parse_ext_substatus(substatus):
     status.message = parse_formatted_message(formatted_message)
     return status
 
+
 def parse_ext_status(ext_status, data):
     if data is None or len(data) is None:
         return
@@ -117,6 +118,7 @@ def parse_ext_status(ext_status, data):
     for substatus in substatus_list:
         if substatus is not None:
             ext_status.substatusList.append(parse_ext_substatus(substatus))
+
 
 # This code migrates, if it exists, handler state and status from an
 # agent-owned directory into the handler-owned config directory
@@ -178,6 +180,7 @@ class ExtHandlersHandler(object):
         self.log_report = False
         self.log_etag = True
         self.log_process = False
+        self._global_enabled_forced = False # Extensions are always enabled on first start of the agent.
 
         self.report_status_error_state = ErrorState()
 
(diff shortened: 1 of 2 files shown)


boumenot:
This change is not ready for merging.  There is a lack of any unit tests to validate the change. I would still like feedback about the approach, which is fairly small change.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -178,6 +180,7 @@ def __init__(self):
         self.log_report = False
         self.log_etag = True
         self.log_process = False
+        self._global_force_enable = True # Extensions are always enabled on first start of the agent.
```
hglkrijger: I would probably flip this flag so the default value matches the default state, e.g. extension_enabled

azurelinuxagent/ga/exthandlers.py
```diff
@@ -784,6 +797,30 @@ def download(self):
...
+
+        The previous behavior was to call enable every time a goal state was received by the agent.  This
+        means the agent is calling more than is necessary, and while this has always been true it tends
+        to be problematic.  There have been bugs in extensions where re-execution caused bad problems for
+        customers.  The behavior is being changed, while still maintaining the existing extension contract,
+        in an attempt to mitigate some of these extensions bugs.  This also has the side effect of ensuring
+        logged enable commands actually correspond to a real enable.
+        """
+
+        should_enable_test = False
```
hglkrijger: Might be safer to default to match existing behavior, and flip the check

boumenot:
I am excited by your enthusiasm with my work. 😀 

>in general I think the approach is ok

boumenot:
I am putting this PR on hold for now.  We have reports of another extension that relies on this behavior, so our goal reducing enables may not be feasible.

boumenot:
Our idea of reducing enable calls does not sound viable.  Extensions rely on enable being called during extension upgrades, which is the big reason we wanted to make this change in the first place.  We can revisit this change later, but for now I am closing it.

PR #1052, opened 2018-02-23
Add and sort on extension dependency property

uiri:


azurelinuxagent/common/protocol/wire.py
```diff
@@ -1414,6 +1414,12 @@ def parse_plugin(self, plugin):
         ext_handler.properties.upgradeGuid = getattrib(plugin, "upgradeGuid")
         if not ext_handler.properties.upgradeGuid:
             ext_handler.properties.upgradeGuid = None
+
+        try:
+            ext_handler.properties.dependencyLevel = int(getattrib(plugin, "dependencyLevel"))
+        except ValueError:
+            ext_handler.properties.dependencyLevel = None
```
boumenot: Is there a need to propagte None?  If we just set it to a low value (0 or -1 or ...) we can lose the None checks in other parts of the code, right?

tests/ga/test_extension.py
```diff
@@ -381,6 +382,71 @@ def test_ext_handler_no_ext(self, *args):
...
+
+        #Test goal state changed
+        test_data.goal_state = test_data.goal_state.replace("<Incarnation>1<",
+                                                            "<Incarnation>2<")
+        test_data.ext_conf = test_data.ext_conf.replace("seqNo=\"0\"",
+                                                        "seqNo=\"1\"")
+        test_data.ext_conf = test_data.ext_conf.replace("dependencyLevel=\"2\"",
+                                                        "dependencyLevel=\"3\"")
+        test_data.ext_conf = test_data.ext_conf.replace("dependencyLevel=\"1\"",
+                                                        "dependencyLevel=\"4\"")
```
boumenot: Is there a test for dependencyLevel not being set, or are we covered by existing tests?  Is there a test case where this value is defined for some extensions and not others?  Is that even possible?

uiri: We are covered by existing tests for the case where it is not set on any extensions.

I do not think it is possible for the value to be defined for some but not others based on the way the attribute gets serialized to the agent.

azurelinuxagent/common/protocol/wire.py
uiri: None gets treated as 0 anyways, so I don't see why not.

azurelinuxagent/common/protocol/restapi.py
```diff
@@ -178,6 +179,14 @@ def __init__(self, name=None):
         self.properties = ExtHandlerProperties()
         self.versionUris = DataContractList(ExtHandlerVersionUri)
 
+    def sort_key(self):
+        level = self.properties.dependencyLevel
+        if level is None:
+            level = 0
+        if self.properties.state != u"enabled":
+            level = (0 - level) - 1
```
savyasachisamal: Would be good to have a comment here. I didn't understand the reason for (0 - level) -1.

PR #1057, opened 2018-03-03
Fix bug regarding downgrade

uiri:
Instead of raising an exception, log and report an error if updating an
exception would be a downgrade.

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -390,7 +390,12 @@ class ExtHandlersHandler(object):
         old_ext_handler_i = ext_handler_i.get_installed_ext_handler()
         if old_ext_handler_i is not None and \
            old_ext_handler_i.version_gt(ext_handler_i):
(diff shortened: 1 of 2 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -390,7 +390,12 @@ def handle_enable(self, ext_handler_i):
         old_ext_handler_i = ext_handler_i.get_installed_ext_handler()
         if old_ext_handler_i is not None and \
            old_ext_handler_i.version_gt(ext_handler_i):
-            raise ExtensionError(u"Downgrade not allowed")
```
boumenot: Can you implement a test case for this?  I would like to capture this behavior, and ensure people know it is defined.

boumenot:
I will issue a hotfix with this change.  This better aligns the Linux GA and the Windows GA.  The Windows GA treats downgrades as a no-op essentially. 

The fact that we got into this state is another concern that is being pursued internally.

PR #1061, opened 2018-03-09
Report extension status for disallowed downgrade

uiri:
Fix a bug in #1057

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -392,6 +392,8 @@ class ExtHandlersHandler(object):
            old_ext_handler_i.version_gt(ext_handler_i):
             msg = "Downgrade is not allowed. Skipping install and enable."
             ext_handler_i.logger.error(msg)
+            ext_handler_i.set_handler_state(ExtHandlerState.Enabled)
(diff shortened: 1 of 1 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -392,6 +392,8 @@ def handle_enable(self, ext_handler_i):
            old_ext_handler_i.version_gt(ext_handler_i):
             msg = "Downgrade is not allowed. Skipping install and enable."
```
boumenot: Is this message accurate based on your change?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -392,6 +392,8 @@ def handle_enable(self, ext_handler_i):
...
             ext_handler_i.logger.error(msg)
```
boumenot: This should be info or a verbose, but not an error.

azurelinuxagent/ga/exthandlers.py
uiri: Yes. This block hits a return statement at line 399. The install and enable logic is from lines 401 to 420 in this function.

boumenot:
LGTM.  Any more concerns, or are you ready to merge?

PR #1063, opened 2018-03-10
Hotfix 2.2.23

uiri:
Downgrade bug hotfix.


boumenot:
I will fix the test in the branch after this is merged.  This is known to have issues because it hard code the agent version, and makes static assumptions. I was going to fix this for 2.2.23, but we have another issue to hotfix.

PR #1068, opened 2018-03-15
Create scripts/tests.sh and add scripts/pre-commit symlink

uiri:
Travis uses this tests script now. I added a symlink for the pre-commit
hook but this could be changed to a script which does stuff other than
run the unit tests before creating the commit.

After cloning the repo, one must symlink .git/hooks/pre-commit to the
pre-commit script to activate it.


boumenot:
This seems like a personal change, as opposed to a repository level change.  If this is a personal preference, I don't think there is value is merging it.  If you think this is good for the repository, I am happy to discuss.

hglkrijger:
I agree this should not be merged. Closing.

PR #1076, opened 2018-03-26
Revert "Merge pull request #1057 from uiri/fix-downgrade-exception"

uiri:
This reverts commit 70cc1d2bd743ff8ab73a7c2f2cd5c60113b3b85b, reversing
changes made to c929ff0a21bf23e20497ba08ab85a63a34a25b46.

Given where we landed on the hotfix for the downgrade bug, I think that this revert is necessary for master.

### DIFF
--- a/azurelinuxagent/common/event.py
+++ b/azurelinuxagent/common/event.py
@@ -46,7 +46,6 @@ class WALAEventOperation:
     CustomData = "CustomData"
     Deploy = "Deploy"
     Disable = "Disable"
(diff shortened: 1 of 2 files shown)


boumenot:
I don't think is sufficient.  Everything that is in hotfix/2.2.25 needs to be merged.  Let me do that, and I'll have you review it.

uiri:
I was concerned that there might be merge conflicts without this but it looks like #1078 is clean.

PR #1077, opened 2018-03-26
Additional processing for dependencies

uiri:
This builds on #1052

### DIFF
--- a/azurelinuxagent/common/protocol/restapi.py
+++ b/azurelinuxagent/common/protocol/restapi.py
@@ -155,7 +155,7 @@ class Dependency(object):
 
     def __init__(self, handler=None, exts=None, timeout=None):
         self.handler = handler
-        self.exts = exts
+        self.exts = [] if exts is None else exts
         self.timeout = timeout
 
 
--- a/azurelinuxagent/common/protocol/restapi.py
+++ b/azurelinuxagent/common/protocol/restapi.py
@@ -144,6 +144,21 @@ class VMAgentManifestList(DataContract):
         self.vmAgentManifests = DataContractList(VMAgentManifest)
 
 
+class Dependency(object):
+    """
+    Represents an extension upon which another extension is dependent
+    handler - the ExtHandler of the dependency
+    exts - the Extension(s) of the dependency
+    timeout - the timeout for the extension to get to the Ready state
+              timeout is expressed in minutes.
+    """
+
+    def __init__(self, handler=None, exts=None, timeout=None):
+        self.handler = handler
+        self.exts = exts
+        self.timeout = timeout
+
+
 class Extension(DataContract):
     def __init__(self,
                  name=None,
@@ -160,6 +175,23 @@ class Extension(DataContract):
         self.certificateThumbprint = certificateThumbprint
         self.dependencyLevel = dependencyLevel
         self.dependencies = [] if dependencies is None else dependencies
+
+    def resolve_dependencies(self, ext_handlers):
+        resolved_dependencies = []
+        for (handler_name, ext_name) in self.dependencies:
+            # Gracefully handle garbage cases where handler name does
+            # not exist or is not unique. Same for the extension name.
+            for handler in ext_handlers:
+                if handler_name != handler.name:
+                    continue
+                resolved_dependencies.append(
+                    Dependency(
+                        handler = handler,
+                        exts = [ext for ext in handler.properties.extensions
+                                if ext.name == ext_name]
+                    )
+                )
+        self.dependencies = resolved_dependencies
 
 
 class ExtHandlerProperties(DataContract):
--- a/azurelinuxagent/common/protocol/restapi.py
+++ b/azurelinuxagent/common/protocol/restapi.py
@@ -150,12 +150,16 @@ class Extension(DataContract):
                  sequenceNumber=None,
                  publicSettings=None,
                  protectedSettings=None,
-                 certificateThumbprint=None):
+                 certificateThumbprint=None,
+                 dependencyLevel=0,
+                 dependencies=None):
         self.name = name
         self.sequenceNumber = sequenceNumber
         self.publicSettings = publicSettings
         self.protectedSettings = protectedSettings
         self.certificateThumbprint = certificateThumbprint
+        self.dependencyLevel = dependencyLevel
+        self.dependencies = [] if dependencies is None else dependencies
 
 
 class ExtHandlerProperties(DataContract):
@@ -163,7 +167,6 @@ class ExtHandlerProperties(DataContract):
         self.version = None
         self.upgradePolicy = None
         self.upgradeGuid = None
-        self.dependencyLevel = None
         self.state = None
         self.extensions = DataContractList(Extension)
 
@@ -180,9 +183,11 @@ class ExtHandler(DataContract):
         self.versionUris = DataContractList(ExtHandlerVersionUri)
 
     def sort_key(self):
-        level = self.properties.dependencyLevel
-        if level is None:
+        levels = [e.dependencyLevel for e in self.properties.extensions]
+        if len(levels) == 0:
             level = 0
+        else:
+            level = min(levels)
         # Process uninstall or disabled before enabled, in reverse order
         # remap 0 to -1, 1 to -2, 2 to -3, etc
         if self.properties.state != u"enabled":
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -1429,11 +1429,6 @@ class ExtensionsConfig(object):
         if not ext_handler.properties.upgradeGuid:
             ext_handler.properties.upgradeGuid = None
 
-        try:
-            ext_handler.properties.dependencyLevel = int(getattrib(plugin, "dependencyLevel"))
-        except ValueError:
-            ext_handler.properties.dependencyLevel = 0
-
         auto_upgrade = getattrib(plugin, "autoUpgrade")
         if auto_upgrade is not None and auto_upgrade.lower() == "true":
             ext_handler.properties.upgradePolicy = "auto"
@@ -1471,6 +1466,26 @@ class ExtensionsConfig(object):
             logger.error("Invalid extension settings")
             return
 
+        depends_on = None
+        depends_on_node = find(settings[0], "DependsOn")
+
+        depends_on_name = getattrib(depends_on_node, "name")
+        if depends_on_name == "":
+            depends_on_name = ext_handler.name
+
+        try:
+            depends_on_level = int(getattrib(depends_on_node, "dependencyLevel"))
+        except (ValueError, TypeError):
+            depends_on_level = 0
+
+        dependencies = []
+        for dependency in findall(depends_on_node, "DependsOnExtension"):
+            handler_name = gettext(dependency)
+            ext_name = getattrib(dependency, "name")
+            if ext_name == "":
+                ext_name = handler_name
+            dependencies.append((handler_name, ext_name))
+
         for plugin_settings_list in runtime_settings["runtimeSettings"]:
             handler_settings = plugin_settings_list["handlerSettings"]
             ext = Extension()
@@ -1480,6 +1495,9 @@ class ExtensionsConfig(object):
             ext.sequenceNumber = seqNo
             ext.publicSettings = handler_settings.get("publicSettings")
             ext.protectedSettings = handler_settings.get("protectedSettings")
+            if ext.name == depends_on_name:
+                ext.dependencies = dependencies
+                ext.dependencyLevel = depends_on_level
             thumbprint = handler_settings.get(
                 "protectedSettingsCertThumbprint")
             ext.certificateThumbprint = thumbprint
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -1484,8 +1484,8 @@ class ExtensionsConfig(object):
 
         dependencies = []
         for dependency in findall(depends_on_node, "DependsOnExtension"):
-            handler_name = gettext(dependency)
-            ext_name = getattrib(dependency, "name")
+            handler_name = getattrib(dependency, "handler")
+            ext_name = getattrib(dependency, "extension")
             if ext_name == "":
                 ext_name = handler_name
             dependencies.append((handler_name, ext_name))
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -1364,21 +1364,6 @@ class Certificates(object):
         return file_name
 
 
-class Dependency(object):
-    """
-    Represents an extension upon which another extension is dependent
-    handler - the ExtHandler of the dependency
-    exts - the Extension(s) of the dependency
-    timeout - the timeout for the extension to get to the Ready state
-              timeout is expressed in minutes.
-    """
-
-    def __init__(self, handler=None, exts=None, timeout=None):
-        self.handler = handler
-        self.exts = exts
-        self.timeout = timeout
-
-
 class ExtensionsConfig(object):
     """
     parse ExtensionsConfig, downloading and unpacking them to /var/lib/waagent.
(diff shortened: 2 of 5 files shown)


azurelinuxagent/common/protocol/wire.py
```diff
@@ -1396,6 +1411,21 @@ def parse(self, xml_text):
             self.ext_handlers.extHandlers.append(ext_handler)
             self.parse_plugin_settings(ext_handler, plugin_settings)
 
+        for ext_handler in self.ext_handlers.extHandlers:
```
boumenot: This is four or five levels of for loops. It is hard for me to track all of things you're trying to accomplish.   Please break this up into well named method or supply comments.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -319,9 +317,35 @@ def handle_ext_handlers(self, etag=None):
 
         self.ext_handlers.extHandlers.sort(key=operator.methodcaller('sort_key'))
         for ext_handler in self.ext_handlers.extHandlers:
-            # TODO: handle install in sequence, enable in parallel
+            self.wait_on_ext_handler_dependencies(ext_handler)
             self.handle_ext_handler(ext_handler, etag)
-    
+
+    def wait_on_ext_handler_dependencies(self, ext_handler):
```
boumenot: This method concerns me.  This may have an effect on TDP. The increase in time the agent will pay attention to an extension has gone up considerably.  I don't yet know what the impact of that will be because we don't have data to track that.

boumenot: Please add test cases for this.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1457,6 +1482,26 @@ def parse_plugin_settings(self, ext_handler, plugin_settings):
             logger.error("Invalid extension settings")
             return
 
+        depends_on = None
```
boumenot: Are there test cases for this code?  If not, please add some.  It looked like the test code changed minimally vs. the running code.

azurelinuxagent/common/protocol/wire.py
boumenot: Are there test cases for this?

tests/ga/test_extension.py
```diff
@@ -396,8 +396,8 @@ def test_ext_handler_sequencing(self, *args):
         self.assertTrue(exthandlers_handler.ext_handlers is not None)
         self.assertTrue(exthandlers_handler.ext_handlers.extHandlers is not None)
         self.assertEqual(len(exthandlers_handler.ext_handlers.extHandlers), 2)
-        self.assertEqual(exthandlers_handler.ext_handlers.extHandlers[0].properties.dependencyLevel, 1)
-        self.assertEqual(exthandlers_handler.ext_handlers.extHandlers[1].properties.dependencyLevel, 2)
+        self.assertEqual(exthandlers_handler.ext_handlers.extHandlers[0].properties.extensions[0].dependencyLevel, 1)
```
boumenot: I see, this is mostly schema movements/changes.

azurelinuxagent/common/protocol/restapi.py
```diff
@@ -180,9 +183,11 @@ def __init__(self, name=None):
         self.versionUris = DataContractList(ExtHandlerVersionUri)
 
     def sort_key(self):
-        level = self.properties.dependencyLevel
-        if level is None:
+        levels = [e.dependencyLevel for e in self.properties.extensions]
+        if len(levels) == 0:
             level = 0
+        else:
+            level = min(levels)
```
jasonzio: Min or max?
>  An extension's distance from the roots of the dependency graph is based on the maximum number of edges between that extension and any of the roots of the acyclic dependency graph.

uiri: This is for the handler which needs to be installed before any of its extensions may be enabled. Right now the agent only supports one extension per handler, but I think `min` makes sense for a handler with more extensions. Additional changes would be needed to support more than one extension per handler but that is out of scope for this PR.

jasonzio: That's not what your spec says, which I quoted. Assuming I understand it right.

azurelinuxagent/common/protocol/restapi.py
```diff
@@ -161,6 +176,24 @@ def __init__(self,
         self.dependencyLevel = dependencyLevel
         self.dependencies = [] if dependencies is None else dependencies
 
+    def resolve_dependencies(self, ext_handlers):
```
jasonzio: More "pythonic":
```
def resolve_dependencies(self, ext_handlers): 
    resolved_dependencies = [] 
    for (handler_name, ext_name) in self.dependencies: 
        # Gracefully handle garbage cases where handler name does 
        # not exist or is not unique. Same for the extension name. 
        for handler in ext_handlers:
            if handler_name == handler.name:
                resolved_dependencies.append(
                    Dependency(handler=handler,
                        exts = [ext in handler.properties.extensions where ext_name==ext.name]
                    ))
    self.dependencies = resolved_dependencies
```
I think this is more readable, but your mileage may vary.

jasonzio: I'm a little twitchy in that a side effect of this function is to change the contents of self.dependencies from a list of 2-tuples to a list of Dependency objects. I know Python doesn't care, but does this alteration of the shape of that member make it harder for humans to understand what's going on?

uiri: I would have put `Dependency` objects in the list from the get-go but I can't resolve the dependencies until it has already parsed all of the `Extension` objects because an extension's dependency might occur in the document after it has been parsed.

@boumenot thoughts on Jason's suggested list comprehension change?

azurelinuxagent/common/protocol/restapi.py
jasonzio: OK,  your in-office conversation with me convinced me. This is talking about the level at which the extension handler gets "handled", and it has to be done before the shallowest dependent extension gets "handled", and that means "min". Got it.

azurelinuxagent/common/protocol/restapi.py
```diff
@@ -144,26 +144,61 @@ def __init__(self):
...
 class Extension(DataContract):
     def __init__(self,
                  name=None,
                  sequenceNumber=None,
                  publicSettings=None,
                  protectedSettings=None,
-                 certificateThumbprint=None):
+                 certificateThumbprint=None,
+                 dependencyLevel=0,
+                 dependencies=None):
```
boumenot: Why don't you just default to []?  You remove the need for the if statement below, and null arrays are evil. 😝

azurelinuxagent/common/protocol/restapi.py
```diff
@@ -144,26 +144,61 @@ def __init__(self):
...
+    Represents an extension upon which another extension is dependent
+    handler - the ExtHandler of the dependency
+    exts - the Extension(s) of the dependency
+    timeout - the timeout for the extension to get to the Ready state
+              timeout is expressed in minutes.
+    """
+
+    def __init__(self, handler=None, exts=None, timeout=None):
+        self.handler = handler
+        self.exts = exts
```
boumenot: self.exts = [] ?

Same reason as comment below.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -362,9 +360,37 @@ def handle_ext_handlers(self, etag=None):
...
+        dependencies = sum([e.dependencies for e in ext_handler.properties.extensions], [])
+        for dependency in dependencies:
+            handler_i = ExtHandlerInstance(dependency.handler, self.protocol)
+            timeout = 90 if dependency.timeout is None else dependency.timeout
```
boumenot: I worry about the impact on TDP with this change, and possible side effects.  Does this code ever get run if a dependency does not exist?

azurelinuxagent/common/protocol/restapi.py
uiri: It is a Python gotcha. If I used `[]` instead of `None`, and simplified the logic below, then _every_ new instance of `Extension` gets a pointer to the same default list, so an append to the `dependencies` of one `Extension` is an append to the `dependencies` of every `Extension` that was instantiated with the default parameter.

I don't think I run into that gotcha here given the way `Extension`s are currently being built, but I'd rather proactively avoid WTF debugging sessions where I can.

azurelinuxagent/ga/exthandlers.py
uiri: What do you mean by "if a dependency does not exist" ?
If a handler's extensions have no dependencies then this `for` loop should not run.
A missing dependency should be rejected at the template level before the agent even sees it. Even if it does make its way to the agent, `resolve_dependencies` should end up generating a `Dependency` where `exts` is empty so the `for` loop below won't run.

hglkrijger: Where did 90 seconds come from? We give extensions 5 minutes, iirc? Either way, it should come from the spec.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -362,9 +360,37 @@ def handle_ext_handlers(self, etag=None):
...
+            timeout_delta = datetime.timedelta(seconds=(timeout * 60))
+            begin = datetime.datetime.utcnow()
+            for ext in dependency.exts:
+                status = None
+                while timeout_delta > (datetime.datetime.utcnow() - begin):
+                    status = handler_i.collect_ext_status(ext)
+                    if status is None:
+                        break
+                    all_success = status.status == "success"
```
hglkrijger: Should this be `&=`?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -362,9 +360,37 @@ def handle_ext_handlers(self, etag=None):
...
+                    for substatus in status.substatusList:
+                        if substatus.status != "success":
+                            all_success = False
+                            break
+                    if all_success:
+                        break
+                    time.sleep(10)
```
hglkrijger: Should this be `time.sleep(timeout_delta)`? Maybe I am confused about `timeout` and `timeout_delta` - could the naming be clearer?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -362,9 +360,37 @@ def handle_ext_handlers(self, etag=None):
...
+                while timeout_delta > (datetime.datetime.utcnow() - begin):
+                    status = handler_i.collect_ext_status(ext)
+                    if status is None:
+                        break
+                    all_success = status.status == "success"
+                    for substatus in status.substatusList:
+                        if substatus.status != "success":
+                            all_success = False
+                            break
+                    if all_success:
```
hglkrijger: If `all_success` is `False`, should that be logged or have some other impact?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -362,9 +360,37 @@ def handle_ext_handlers(self, etag=None):
...
+                        break
+                    time.sleep(10)
+                if (datetime.datetime.utcnow() - begin) > timeout_delta:
```
hglkrijger: I do not think it makes sense to log based on time, and control the loop based on a bool. Seems it should be the other way around.

tests/ga/test_extension.py
```diff
@@ -434,8 +441,8 @@ def test_ext_handler_sequencing(self, *args):
                                     1, "1.0.0",
                                     expected_handler_name="OSTCExtensions.OtherExampleHandlerLinux")
         self.assertEqual(len(exthandlers_handler.ext_handlers.extHandlers), 2)
-        self.assertEqual(exthandlers_handler.ext_handlers.extHandlers[0].properties.dependencyLevel, 4)
-        self.assertEqual(exthandlers_handler.ext_handlers.extHandlers[1].properties.dependencyLevel, 3)
+        self.assertEqual(exthandlers_handler.ext_handlers.extHandlers[0].properties.extensions[0].dependencyLevel, 4)
```
hglkrijger: These tests are not super intuitive, please add comments and/or examples for new readers.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -362,9 +360,37 @@ def handle_ext_handlers(self, etag=None):
 
         self.ext_handlers.extHandlers.sort(key=operator.methodcaller('sort_key'))
         for ext_handler in self.ext_handlers.extHandlers:
-            # TODO: handle install in sequence, enable in parallel
+            self.wait_on_ext_handler_dependencies(ext_handler)
```
hglkrijger: At least to me, this does not seem like an intuitive way of handling dependencies. In fact, they really are not dependencies, just a particular sequence. Why the need to resolve dependencies and build another chain, when the parent loop has all the info it needs? If the order of `self.ext_handlers.extHandlers` mapped to the dependency order, would that not do the same thing? 

It seems like the intent here is to have each link in the chain know something about their dependency (or really, their pre-cursor) and that seems to be just so we can query the status? We can do that from this loop more simply, imo. Let me know if I have misunderstood the intention here though.

azurelinuxagent/ga/exthandlers.py
uiri: 90 *minutes* is the existing timeout for an extension operation at the platform level. That includes the 15 minutes for install, and 5 minutes for enable. In the case of an update, that includes up to 45 minutes (disable + update + uninstall + install + enable).

Extensions have 5 minutes for the enable process to complete. If that process writes out the status blob then this will either return (almost) immediately in the case of a success, or it will timeout in the case of a failure and prevent the installation of further extensions which depend on the extension that failed.

If that process forks a child process which runs beyond the 5 minute timeout, then who knows when the final status blob will get written. Most likely not before the first iteration of this loop for an extension that depends on the extension whose enable command forked a child process.

azurelinuxagent/ga/exthandlers.py
uiri: `timeout` is an integer, `timeout_delta` is a timedelta... I've simplified so that `timeout` is an integer and then a timedelta.

10 seconds is some arbitrary number to avoid rereading the status blob too frequently. I am open to suggestions if you would prefer to make it more or less frequent (or to not sleep at all).

azurelinuxagent/ga/exthandlers.py
uiri: if `all_success` is `False`, then some `status` value in the status blob was reported as a value other than `success` to the platform which would take appropriate action.

azurelinuxagent/ga/exthandlers.py
uiri: Let's say we have 6 extension handlers (1 extension per handler): A, B, C, D, E, F.
D depends on A, E depends on B, F depends on C. The order is still: A, B, C, D, E, F
When A finishes, B can run right away. When B finishes, C can run right away. They do not depend on anything. D has to ensure that A was successful before it can run. If A and B were successful, but C was not, then D and E can run but F cannot. It is, of course, possible that A will fail and prevent all of D, E, F from running, but that is still a better outcome than A failing and B, C, D, E, F being prevented from running.

azurelinuxagent/ga/exthandlers.py
uiri: Fixed, I think. Thanks.

azurelinuxagent/ga/exthandlers.py
uiri: This is the "initial" assignment for `all_success` and subsequent assignments. If the status goes from "pending" (or similar) to "success", then it should be `True` rather than `False`.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -362,9 +360,29 @@ def handle_ext_handlers(self, etag=None):
...
             self.handle_ext_handler(ext_handler, etag)
-    
+
+    def wait_on_ext_handler_dependencies(self, ext_handler):
+        dependencies = sum([e.dependencies for e in ext_handler.properties.extensions], [])
+        for dependency in dependencies:
+            handler_i = ExtHandlerInstance(dependency.handler, self.protocol)
+            timeout = 90 if dependency.timeout is None else dependency.timeout
```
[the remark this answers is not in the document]
uiri: Customers may want some extensions (e.g. CustomScript or RunCommand) to have much shorter timeouts.

azurelinuxagent/ga/exthandlers.py
uiri: The platform timeout is adjusted to match the number of dependency levels. I have changed the code so that it more closely matches this requirement. It can't match exactly because of the time it takes for the platform to send a new GoalState and the agent to start processing extensions from that new GoalState.

hglkrijger:
@uiri are you still wanting to merge this? If so I can re-review next week, and merge when the release is done.

uiri:
+ @sirajudeens

Yes, this feature still needs to be merged. I think there is more testing that needs to be done before it is completely ready though.

hglkrijger:
Ok, if it is not ready I am going to close. Please reopen when it is updated and ready to go. We should be able to take it next week.

PR #1079, opened 2018-03-28
upgrade if incarnation resets

boumenot:


azurelinuxagent/common/protocol/wire.py
```diff
@@ -876,6 +880,25 @@ def get_gafamily_manifest(self, vmagent_manifest, goal_state):
...
+        file is not removed agent upgrade may be delayed.
+
+        :param family: GA family, e.g. Prod or Test
+        :param incarnation: incarnation of the current goal state
+        """
+        fn = AGENTS_MANIFEST_FILE_NAME.format(
+            family,
+            incarnation)
+
+        agent_manifest = os.path.join(conf.get_lib_dir(), fn)
```
uiri: Is there a reason to duplicate the filename generation logic rather than pass in the final `local_file` name?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -862,6 +862,10 @@ def get_gafamily_manifest(self, vmagent_manifest, goal_state):
                     self.update_goal_state(forced=True)
                     goal_state = self.get_goal_state()
 
+                self._remove_stale_agent_manifest(
+                    vmagent_manifest.family,
+                    goal_state.incarnation)
+
                 local_file = MANIFEST_FILE_NAME.format(
                                 vmagent_manifest.family,
                                 goal_state.incarnation)
```
uiri: It looks like it writes out the agent manifest file below but where does it prefer what it has on disk rather than what it just fetched over the network?

Should the `write_file` call below not just overwrite whatever was there before? How does doing an `rm` first change anything?

azurelinuxagent/common/protocol/wire.py
boumenot: I don't see the duplication.  There are two file names: MANIFEST_FILE_NAME and AGENTS_MANIFEST_FILE_NAME.

I am probably missing what you're seeing.

uiri: Nope, you're right. I misread the constants used.

azurelinuxagent/common/protocol/wire.py
boumenot: The agents manifest file is computed, and written to disk in a different function.  The different function also does a little dance depending upon the file existing or not, which is why I have to delete it here.  The manifest file is written whenever a new goal state arrives.

boumenot:
@uiri - any more concerns or questions?

azurelinuxagent/common/protocol/wire.py
uiri: It looks like we can just skip the call to `filter_package_list` if we're never going to use the disk cache anyways. Just have `get_vmagent_pkgs` return `ga_manifest.pkg_list` instead.

boumenot: The cache file is written when the new goal state arrives because the agent manifest can change over time.  The idea is to snapshot the current value when the goal state arrives to ensure you only update the agent within a goal state.  

It also provides a historical record when we have to debug issues.

PR #1080, opened 2018-03-29
improve test_get_first_if for other systems

### LINKED ISSUE #1046: unit test failure
The unit test "test_get_first_if()" in test_default makes an assumption about the network configuration of the system on which the unit test is being run:

self.assertTrue(ifname.startswith('eth'))

Thus this unit test necessarily fails on systems that use a bridge or systems that use predictable network interface names.

boumenot:
The test case makes an assumption about the user's network configuration that
is not always valid.  Rather than assume, the code has been changed to mock out
these dependencies.

Closes #1046

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -674,6 +674,7 @@ class DefaultOSUtil(object):
                              socket.IPPROTO_UDP)
         param = struct.pack('256s', (ifname[:15]+('\0'*241)).encode('latin-1'))
         info = fcntl.ioctl(sock.fileno(), IOCTL_SIOCGIFHWADDR, param)
+        sock.close()
         return ''.join(['%02X' % textutil.str_to_ord(char) for char in info[18:24]])
 
     @staticmethod
@@ -698,13 +699,16 @@ class DefaultOSUtil(object):
 
         buff = array.array('B', b'\0' * array_size)
         param = struct.pack('iL', array_size, buff.buffer_info()[0])
+
         sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, socket.IPPROTO_UDP)
         ret = fcntl.ioctl(sock.fileno(), IOCTL_SIOCGIFCONF, param)
         retsize = (struct.unpack('iL', ret)[0])
+        sock.close()
+
         if retsize == array_size:
             logger.warn(('SIOCGIFCONF returned more than {0} up '
                          'network interfaces.'), expected)
-        sock = None
+
         ifconf_buff = buff.tostring()
 
         ifaces = {}
@@ -820,6 +824,7 @@ class DefaultOSUtil(object):
         if not self.disable_route_warning:
             logger.info('interface [{0}] has flags [{1}], '
                         'is loopback [{2}]'.format(ifname, flags, isloopback))
+        s.close()
         return isloopback
 
     def get_dhcp_lease_endpoint(self):
--- a/tests/common/osutil/test_default.py
+++ b/tests/common/osutil/test_default.py
@@ -18,15 +18,18 @@
 import socket
 import glob
 import mock
+import traceback
 
 import azurelinuxagent.common.osutil.default as osutil
 import azurelinuxagent.common.utils.shellutil as shellutil
 from azurelinuxagent.common.exception import OSUtilError
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.osutil import get_osutil
-from azurelinuxagent.common.utils import fileutil
-from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
 from tests.tools import *
+
+
+def fake_is_loopback(_, iface):
+    return iface.startswith('lo')
 
 
 class TestOSUtil(AgentTestCase):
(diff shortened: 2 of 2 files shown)


azurelinuxagent/common/osutil/default.py
```diff
@@ -682,23 +682,13 @@ def get_first_if(self):
 
         # for 64bit the size is 40 bytes
         # for 32bit the size is 32 bytes
-        python_arc = platform.architecture()[0]
-        struct_size = 32 if python_arc == '32bit' else 40
+        python_arch = platform.architecture()[0]
+        struct_size = 32 if python_arch == '32bit' else 40
 
-        sock = socket.socket(socket.AF_INET,
```
boumenot: It was easier to move this block of code into it's own method, and mock out that method than it was to mock out every low level call in this method.

azurelinuxagent/common/osutil/default.py
```diff
@@ -713,6 +703,34 @@ def get_first_if(self):
 
         return iface.decode('latin-1'), socket.inet_ntoa(sock[i+20:i+24])
 
+    def _get_interface_list_as_string(self, num_devices, struct_size):
```
boumenot: This is an odd method.  I could have made it return a list of interfaces and IP addresses, but that's not really what the original code did.  I ended up keeping it as a "string" and left it to the caller to decode. I am open to suggestions.

azurelinuxagent/common/osutil/default.py
```diff
@@ -682,23 +682,13 @@ def get_first_if(self):
...
-            logger.warn(('SIOCGIFCONF returned more than {0} up '
-                         'network interfaces.'), expected)
-        sock = buff.tostring()
-        primary = bytearray(self.get_primary_interface(), encoding='utf-8')
+        sock = self._get_interface_list_as_string(expected, struct_size)
+
+        primary_intf = self.get_primary_interface()
+        primary = bytearray(primary_intf, encoding='utf-8')
         for i in range(0, struct_size * expected, struct_size):
             iface=sock[i:i+16].split(b'\0', 1)[0]
```
jasonzio: I would feel a little better if you defined an IFNAMSIZ constant instead of hardcoding that 16.

azurelinuxagent/common/osutil/default.py
jasonzio: I'd be happier if you returned a dictionary of interface names mapped the remainder of the string. The Linux struct ifreq is also a weird thing, since aside from the interface name, the rest is a union of possible representations for the interface information, and you have to know what kind of interface it is before you can correctly decode it. It's a sockaddr, and it's probably an IPv4 sockaddr, but that may change in the future, and there's no way to tell.

tests/common/osutil/test_default.py
```diff
@@ -88,14 +88,22 @@ def test_mount_dvd_failure(self, _):
                         self.assertTrue(msg in ustr(ose))
                         self.assertTrue(patch_run.call_count == 6)
 
-    def test_get_first_if(self):
+    @patch('azurelinuxagent.common.osutil.default.DefaultOSUtil.get_primary_interface', return_value='eth0')
+    @patch('azurelinuxagent.common.osutil.default.DefaultOSUtil._get_interface_list_as_string',
+           return_value=b'eth0\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x02\x00\x00\x00\xac\x11\xba9\x00\x00\x00\x00\x00')
```
jasonzio: I question the \xba9 value in this b'' string. Everything else is either an ordinary character or a two-digit hex value (8 bits); this one 16-bit hex value (without leading zero) makes it hard to check the string. (I know \xc1\x11 are the first two octets of the IPv4 address, but I can't _see_ the last two octets as individual octets.)

uiri: It looks like it is just a typo that should be `\xba\x39`

azurelinuxagent/common/osutil/default.py
boumenot: I inherited this code, which makes me a little leery of change.  I was going to go the dictionary route, but I have little confidence in the consequences of those decisions.  If you have confidence, I am happy to take yours.

  1. If I return a dictionary of interface to IP address is there a case where collision will happen?
  1. If I return a dictionary of interface to IP address is there a case where inet_ntoa will throw?

The code is a little odd because the for loop's variable is used outside the scope of the for loop.  Only when the loop is exhausted is the IP address captured.  It wasn't clear if there was intent behind this *oddity* or not.

tests/common/osutil/test_default.py
boumenot: I copied the value from the debugger (honest), and the unit test asserts the correct IP address.

jasonzio: Doh! @uiri is correct; the '9' is an absolute character, not part of the previous 8-bit escape. I'd like to see '9' replaced by \x39 to make it clear that it represents an octet in an IP address and not a printable character, but that is a stylistic nit.

azurelinuxagent/common/osutil/default.py
rjschwei: Since you split this out to fix a testing problem, no good deed shall go unpunished, I agree with @jasonzio that it would be nicer if the new method returned a dict. But yes this requires more fiddling in the client code, i.e. where the new method if called.

jasonzio: Speaking of no good deed going unpunished... :-) I will prepare a separate PR which does just this one thing. Once we approve/merge, you'll want to unwind this part.

jasonzio: Done, and merged to master.

azurelinuxagent/common/osutil/default.py
```diff
@@ -698,14 +698,17 @@ def _get_all_interfaces(self):
...
         retsize = (struct.unpack('iL', ret)[0])
+        sock.close()
+
         if retsize == array_size:
             logger.warn(('SIOCGIFCONF returned more than {0} up '
                          'network interfaces.'), expected)
-        sock = None
-        ifconf_buff = buff.tostring()
+
+        ifconf_buff = buff.tobytes()
```
boumenot: The tool complained that tostring() was deprecated, so I changed it to tobytes() per the recommendation.

azurelinuxagent/common/osutil/default.py
```diff
@@ -820,6 +823,7 @@ def is_loopback(self, ifname):
         if not self.disable_route_warning:
             logger.info('interface [{0}] has flags [{1}], '
                         'is loopback [{2}]'.format(ifname, flags, isloopback))
+        s.close()
```
boumenot: Unit test runner was complaining that the socket was not closed.

tests/common/osutil/test_default.py
```diff
@@ -88,14 +86,22 @@ def test_mount_dvd_failure(self, _):
                         self.assertTrue(msg in ustr(ose))
                         self.assertTrue(patch_run.call_count == 6)
 
-    def test_get_first_if(self):
+
+    @patch('azurelinuxagent.common.osutil.default.DefaultOSUtil.get_primary_interface', return_value='eth0')
+    @patch('azurelinuxagent.common.osutil.default.DefaultOSUtil._get_all_interfaces', return_value={'eth0':'10.0.0.1'})
+    @patch('azurelinuxagent.common.osutil.default.DefaultOSUtil.is_loopback', return_value=False)
```
jasonzio: There would be more utility to this mock if it actually looked at the interface name (which is an argument to the method) and returned True if and only if the name started with "lo". You would supply the same mock to test_get_first_if_nosuchprimary(); it would ensure that, no matter what Python does to its dictionary construction, the mocked interface dictionary `{'eth0':'192.0.2.100','lo':'127.0.0.1'}` would never pick 'lo' if the primary was not found.

jasonzio: Outside the test class:
```
def fake_is_loopback(iface):
    return iface.startswith('lo')
```

Then the patch becomes
```
@patch('azurelinuxagent.common.osutil.default.DefaultOSUtil.is_loopback', fake_is_loopback) 
```

Supply that patch to both test_get_first_if() and test_get_first_if_nosuchprimary().

azurelinuxagent/common/osutil/default.py
boumenot: Actually, I was wrong.  There appears to be another issue lurking here, so I am reverting it back.

PR #1085, opened 2018-03-30
ensure file handles are closed

boumenot:


boumenot:
The unit test runner was complaining that files were not cleaned up.

PR #1086, opened 2018-03-30
ignore contents of None for write_file

### LINKED ISSUE #1066: 'NoneType' object has no attribute 'encode'
We have noticed failures with CustomScriptForLinux with the following error message.

```text
'NoneType' object has no attribute 'encode'
```

This may affect other extensions in this repo, but the only report we have is for CustomScript.

boumenot:
Closes #1066

### DIFF
--- a/tests/ga/test_extension.py
+++ b/tests/ga/test_extension.py
@@ -233,6 +233,24 @@ class TestHandlerStateMigration(AgentTestCase):
         self.assertEquals(handler_status.message, message)
         return
 
+    def test_set_handler_status_ignores_none_content(self):
+        """
+        Validate that set_handler_status ignore cases where json.dumps
+        returns a value of None.
+        """
+        self._prepare_handler_state()
+        self._prepare_handler_config()
+
+        status = "Ready"
+        code = 0
+        message = "A message"
+
+        try:
+           with patch('json.dumps', return_value=None):
+                self.ext_handler_i.set_handler_status(status=status, code=code, message=message)
+        except Exception as e:
+            self.fail("set_handler_status threw an exception")
+
     @patch("shutil.move", side_effect=Exception)
     def test_migration_ignores_move_errors(self, shutil_mock):
         self._prepare_handler_state()
--- a/tests/utils/test_file_util.py
+++ b/tests/utils/test_file_util.py
@@ -38,7 +38,19 @@ class TestFileOperations(AgentTestCase):
         content_read = fileutil.read_file(test_file)
         self.assertEquals(content, content_read)
         os.remove(test_file)
-    
+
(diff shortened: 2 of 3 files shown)


azurelinuxagent/common/utils/fileutil.py
```diff
@@ -76,6 +76,9 @@ def write_file(filepath, contents, asbin=False, encoding='utf-8', append=False):
     """
     Write 'contents' to 'filepath'.
     """
+    if contents is None:
```
jasonzio: Are you _really_ sure this is what you want? If contents==None, then the file won't even be created. I would think that write_file(path, None) would create an empty file (file exists but is of zero length). That's the existing behavior, isn't it? I would expect this to break something.

tests/utils/test_file_util.py
```diff
@@ -38,7 +38,16 @@ def test_read_write_file(self):
...
         self.assertEquals(content, content_read)
         os.remove(test_file)
-    
+
+    def test_write_empty_file(self):
+        """
+        write_file should ignore contents == None, and do nothing
+        """
+        test_file=os.path.join(self.tmp_dir, self.test_file)
+        fileutil.write_file(test_file, None)
```
jasonzio: If the file exists before write_file(path, None) is called, then your new behavior results in the file being untouched. The previous behavior for results in the file being truncated to zero length when append is False. You're changing behavior. If you desire this new behavior, you should have explicit tests that address the permutations.

azurelinuxagent/common/utils/fileutil.py
jasonzio: From our office conversation: the post-condition for this method should be "In the absence of a raised exception, a file named 'filepath' will exist on disk". I don't care if "contents==None => Exception" or "contents==None  => create empty file"; either one is acceptable.

tests/utils/test_file_util.py
jasonzio: This test is badly named; you're not writing an empty file, you're writing None. `contents=""` is different from `contents=None`.

tests/ga/test_extension.py
```diff
@@ -233,6 +233,24 @@ def test_migration_does_not_overwrite(self):
...
+
+        status = "Ready"
+        code = 0
+        message = "A message"
+
+        try:
+           with patch('json.dumps', return_value=None):
+                self.ext_handler_i.set_handler_status(status=status, code=code, message=message)
+        except Exception as e:
+            self.fail("set_handler_status threw and exception")
```
jasonzio: typo (threw _an_ exception)

boumenot: Fixed.

PR #1099, opened 2018-04-03
Revert "Merge pull request #954 from Azure/pr-920-provision-telemetry"

boumenot:
This reverts commit 154a6ec208608223a83d749a58a0bbffa5798001, reversing
changes made to b943cbc7067fbb8500a6a2ca1c7588f803ff7c81.

### DIFF
--- a/azurelinuxagent/pa/provision/default.py
+++ b/azurelinuxagent/pa/provision/default.py
@@ -22,7 +22,6 @@ Provision handler
 import os
 import os.path
 import re
-import socket
 import time
 
 from datetime import datetime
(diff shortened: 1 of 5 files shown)


boumenot:
I made this change with the intent of tracking down a difference between the VM's report CPU/Memory and the CPU/Memory of the VM size the customer bought. Unfortunately, this change was unnecessary. The CPU/Memory as seen by the guest is already reported, and this event was superfluous.  I missed this fact, and so did my reviewer.

The code is unnecessary, which is why I am

azurelinuxagent/pa/provision/default.py
```diff
@@ -269,53 +264,12 @@ def deploy_ssh_keypairs(self, ovfenv):
             logger.info("Deploy ssh key pairs.")
             self.osutil.deploy_ssh_keypair(ovfenv.username, keypair)
 
-    def report_event(self, message, is_success=False, duration=0,
-                     operation=WALAEventOperation.Provision):
+    def report_event(self, message, is_success=False, duration=0):
```
jasonzio: I would recommend leaving this one change (the ability to specify the operation name) in place. You might not need it now, but you've already written and tested it, and there's zero cost to carrying it along should you need it in the future.

boumenot:
I liked your idea.  I preserved the parameter, and I kept the two tests.

PR #1102, opened 2018-04-05
capture stdout/stderr

### LINKED ISSUE #1083: Return Stdout / StdErr
Hi, during install and enable, please can you emit StdOut/StdErr.

boumenot:
Closes #1083

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -65,6 +65,9 @@ HANDLER_PKG_EXT = ".zip"
 HANDLER_PKG_PATTERN = re.compile(HANDLER_PATTERN+"\\"+HANDLER_PKG_EXT+"$",
                                 re.IGNORECASE)
 
+TELEMETRY_MESSAGE_MAX_LEN = 3200
+
+
 def validate_has_key(obj, key, fullname):
     if key not in obj:
         raise ExtensionError("Missing: {0}".format(fullname))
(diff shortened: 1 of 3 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -949,31 +993,37 @@ def launch_command(self, cmd, timeout=300):
...
+                                  env=os.environ)
+            stdout, stderr = process.communicate(timeout)
+        except subprocess.TimeoutExpired:
+            process.kill()
+            stdout, stderr = process.communicate()
+            msg = format_stdout_stderr(sanitize(stdout), sanitize(stderr))
+            raise ExtensionError("Timeout({0}): {1}\n{2}".format(timeout, cmd, msg))
         except Exception as e:
-            #TODO do not catch all exception
+            process.kill()
```
boumenot: The previous code called os.kill(process.pid, 9) whereas this code calls process.kill(). I made this change because this is the way Python does it.  I can be convinced either way.

tests/ga/test_extension.py
```diff
@@ -15,29 +15,13 @@
 # Requires Python 2.4+ and Openssl 1.0+
 #
 
-import glob
```
boumenot: The IDE said none of these imports were used, so I deleted them.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -949,31 +993,37 @@ def launch_command(self, cmd, timeout=300):
...
+
         try:
-            devnull = open(os.devnull, 'w')
-            child = subprocess.Popen(base_dir + "/" + cmd,
-                                     shell=True,
-                                     cwd=base_dir,
-                                     stdout=devnull,
-                                     env=os.environ)
+            # This should be .run(), but due to the wide variety
+            # of Python versions we must support we must use .communicate().
```
jasonzio: Which particular python versions do we support that require us to use `.communicate()`? Some day we will drop support for distros that use those versions and we'll be able to switch to `.run()`, and it would be nice to know when that day has come.

PR #1104, opened 2018-04-06
update copyright notice

boumenot:


boumenot:
It was driving me nuts to see 2014 and Python 2.4+ in every copyright notice. 😀

uiri:
@boumenot Did you clear this with legal?

Copyright should note the original year of authorship/publication, not the current year since it is used to calculate when the code will enter the public domain. 2014-2018 might be fine if each of 2014, 2015, 2016, 2017, and 2018 appear in `git blame`.

boumenot:
Legal's standpoint is to not have a PR. I'll prepare another PR.

PR #1105, opened 2018-04-06
add uptime to provision event

### LINKED ISSUE #1091: Track Boot Time in Provision Event
To better understand and break down the provision process please include the boot time in the provision event, or emit a boot event with an appropriate duration.

boumenot:
Closes #1091

### DIFF
--- a/tests/pa/test_provision.py
+++ b/tests/pa/test_provision.py
@@ -146,8 +146,12 @@ class TestProvision(AgentTestCase):
 
         ph.run()
 
-        call1 = call("Provisioning succeeded", duration=ANY, is_success=True)
-        ph.report_event.assert_has_calls([call1])
+        self.assertEqual(1, ph.report_event.call_count)
+        positional_args, kw_args = ph.report_event.call_args
+        # [call('Provisioning succeeded (146473.68)', duration=65, is_success=True)]
+        self.assertTrue(re.match(r'Provisioning succeeded \(\d+\.\d+\)', positional_args[0]) is not None)
+        self.assertTrue(isinstance(kw_args['duration'], int))
+        self.assertTrue(kw_args['is_success'])
 
     @distros()
     @patch(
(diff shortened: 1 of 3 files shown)


tests/pa/test_provision.py
```diff
@@ -146,8 +146,12 @@ def test_provision_telemetry_success(self, mock_util, distro_name, distro_versio
 
         ph.run()
 
-        call1 = call("Provisioning succeeded", duration=ANY, is_success=True)
-        ph.report_event.assert_has_calls([call1])
+        self.assertEqual(1, ph.report_event.call_count)
+        args = ph.report_event.call_args_list[0]
```
jasonzio: Since you've asserted that report_event() was called exactly once, you can simplify this to be `args = ph.report_event.call_args` which is the call object representing the last (and, as verified, only) call to the mocked method.

tests/pa/test_provision.py
```diff
@@ -146,8 +146,12 @@ def test_provision_telemetry_success(self, mock_util, distro_name, distro_versio
...
+        # [call('Provisioning succeeded (146473.68)', duration=65, is_success=True)]
+        self.assertTrue(re.match(r'Provisioning succeeded \(\d+\.\d+\)', args[0][0]) is not None)
+        self.assertTrue(isinstance(args[1]['duration'], int))
+        self.assertTrue(args[1]['is_success'])
```
jasonzio: This is a stylistic quibble, and you can feel free to ignore me.

While unittest.mock is amazing, some of the idioms for using it are opaque. I had to read the documentation to assure myself that what you're doing, as written, is precisely correct. I think it's kinder to humans to write the code with a bit more transparency, something like this:
```
positional_args, kw_args = ph.report_event.call_args
self.assertTrue(re.match(r'Provisioning succeeded \(\d+\.\d+\)', positional_args[0]) is not None)
self.assertTrue(isinstance(kw_args['duration'], int)) 
self.assertTrue(kw_args['is_success'])
```

boumenot: I was struggling to come up with a better names than `args`, and I failed.  I like your changed.  I have implemented them.

PR #1107, opened 2018-04-13
Use entry_points for waagent script

credbbl:
The use of waagent2.0 via loading waagent was removed from the Microsoft maintained extensions.


boumenot:
Unfortunately, there are still extensions that expect these files to exist.  It is **incorrect** and **bad** that they do depend on them, but if I delete these files it will break customers.  We have asked extension owners to correct this issue, and most have, but there are still versions installed on customer VMs that would break as a result of this change.

Let me discuss this with the team, and see what we can do.

PR #1109, opened 2018-04-16
compress and base64 encode stdout/stderr

boumenot:
The recent change to log stdout/stderr caused some other issues.  The output for most extenisons was never exposed before, and it is quite noisy.  There is some work that needs to be done on the part of extensions to cleanup this up.  The output contains (potentially) many newlines, which breaks the line oriented nature of waagent.log. These outputs are still useful for debugging, and we would like to preserve them.  The compromise is to keep the output, but encode it to keep the line oriented nature, and reduce the visual noise.  The data are compressed with zlib before encoding it to further reduce the base64 bloat.

### DIFF
--- a/azurelinuxagent/common/event.py
+++ b/azurelinuxagent/common/event.py
@@ -35,6 +35,7 @@ from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.protocol.restapi import TelemetryEventParam, \
     TelemetryEvent, \
     get_properties
+from azurelinuxagent.common.utils import textutil
 from azurelinuxagent.common.version import CURRENT_VERSION
 
 _EVENT_MSG = "Event: name={0}, op={1}, message={2}, duration={3}"
@@ -166,9 +167,9 @@ def _encode_message(op, message):
         return message
 
     try:
-        return base64.b64encode(zlib.compress(message))
+        return textutil.b64encode(textutil.compress(message))
     except Exception:
-        # If the message could not be encoded and dummy message is returned.
+        # If the message could not be encoded a dummy message ('<>') is returned.
         # The original message was still sent via telemetry, so all is not lost.
         return "<>"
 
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/event.py
```diff
@@ -128,9 +139,45 @@ def _save(self):
...
+    stdout/stderr from extension operations.  Extension operations tend to generate
+    a lot of noise, which makes it difficult to parse the line-oriented waagent.log.
+    The compromise is to encode the stdout/stderr so we preserve the data and do
+    not destroy the line oriented nature.
+
+    The data can be recovered using the following command:
+
+      $ echo '<encoded data>' | base64 -d | pigz -zd
+
+    You may need to install the pigz command.
```
jasonzio: Consider having the guest agent, at startup, log this brief summary of "how to read the stdout/stderr blob" to the agent log. That may help forestall a few support calls.

azurelinuxagent/common/event.py
```diff
@@ -128,9 +139,45 @@ def _save(self):
...
+    if len(message) == 0:
+        return message
+
+    if op not in SHOULD_ENCODE_MESSAGE_OP:
+        return message
+
+    try:
+        return textutil.b64encode(textutil.compress(message))
+    except Exception:
+        # If the message could not be encoded and dummy message is returned.
```
jasonzio: Typo

If the message could not be encoded, *a* dummy message is returned.

azurelinuxagent/common/event.py
boumenot: I don't think a startup message is the appropriate spot for this, but I am happy to add it to the wiki.

azurelinuxagent/common/event.py
boumenot: Fixed.

azurelinuxagent/common/event.py
jasonzio: That works. Discoverability is key.

PR #1110, opened 2018-04-16
purge on incarnation change; archive daily

boumenot:
Goal state information for an incarnation are now flushed to /var/lib/waagent/history whenever the incarnation number changes.  This should address the issue we have with agent upgrade and the incarnation number resetting.  The only cache files in /var/lib/waagent are for the current incarnation.

The *history* is archived and purged every 24 hours.


azurelinuxagent/common/utils/fileutil.py
```diff
@@ -52,6 +52,13 @@ def copy_file(from_path, to_path=None, to_dir=None):
     return to_path
 
 
+def move_file(from_path, to_path=None, to_dir=None):
+    if to_path is None:
+        to_path = os.path.join(to_dir, os.path.basename(from_path))
```
uiri: I think `shutil.move` already does this if the destination is a directory.

I think it looks weird to have both `to_path` and `to_dir` as parameters since `to_path` "wins" if you specify both.

azurelinuxagent/common/utils/archive.py
```diff
@@ -0,0 +1,213 @@
...
+        try:
+            fileutil.mkdir(self._target, mode=0o700)
+            return True
+        except:
+            return False
+
+
+# @total_ordering first appeared in Python 2.7 and 3.2
+# If there are more use cases for @total_ordering, I will
+# consider re-implementing it.
```
uiri: TODO: Use total_ordering once RHEL/CentOS 6 and SLES 11 are EOL.

azurelinuxagent/common/utils/archive.py
```diff
@@ -0,0 +1,213 @@
...
+
+class StateZip(State):
+    def __init__(self, path, timestamp):
+        super(StateZip,self).__init__(path, timestamp)
+
+    def delete(self):
+        os.remove(self._path)
+
+    def archive(self):
+        pass
```
uiri: Is this method necessary since it has the same definition in the parent `State` class?

azurelinuxagent/common/utils/archive.py
```diff
@@ -0,0 +1,213 @@
...
+
+    @property
+    def timestamp(self):
+        return self._timestamp
+
+    def delete(self):
+        pass
+
+    def archive(self):
+        pass
```
uiri: Should `archive` and `delete` raise `NotImplementedError` here ?

azurelinuxagent/common/utils/archive.py
jasonzio: The sooner we can drop support for 2.6, the happier we will all be. For example, any tests that expect exceptions need to be re-examined to see if self.assertRaises() should be used as a context manager.

azurelinuxagent/common/utils/archive.py
```diff
@@ -0,0 +1,213 @@
+# Microsoft Azure Linux Agent
+#
+# Copyright 2018 Microsoft Corporation
```
jasonzio: Drop the year from the copyright statement

azurelinuxagent/common/utils/archive.py
```diff
@@ -0,0 +1,213 @@
...
+        pass
+
+    def archive(self):
+        pass
+
+    def __eq__(self, other):
+        return self._timestamp == other.timestamp
+
+    def __ne__(self, other):
+        return self._timestamp != other.timestamp
```
jasonzio: Nit: Dual relops are often defined by stating one, then stating the other as the negation of the first. That is...
```
def __ne__(self, other):
    return not self == other
```
Similarly, __le__ is defined as not > and __ge__ is defined as not <.

azurelinuxagent/common/utils/archive.py
boumenot: I'll bake the cake.  It ain't gonna happen any time soon.

azurelinuxagent/common/utils/archive.py
```diff
@@ -0,0 +1,213 @@
...
+            for pattern in CACHE_PATTERNS:
+                m = pattern.match(f)
+                if m is not None:
+                    files.append(full_path)
+                    break
+
+        return files
+
+    def _archive(self, files):
+        for file in files:
```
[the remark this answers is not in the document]
boumenot: The pattern appears to `for f in files`, so that's what I changed it too.

azurelinuxagent/common/utils/fileutil.py
boumenot: You're right.  I deleted copy_file and move_file, and just used shutil.

azurelinuxagent/common/utils/archive.py
boumenot: No, it is not necessary.   It does make it explicit, which I like.  I removed it.

azurelinuxagent/common/utils/archive.py
boumenot: I switched to the approve style for these files.

azurelinuxagent/common/utils/archive.py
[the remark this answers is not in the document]
boumenot: This was also true for my use of zip.

azurelinuxagent/common/utils/archive.py
boumenot: Meh.  It's a pretty simple self-contained hierarchy.  I don't think that's a style I see very often in Python either.

azurelinuxagent/common/utils/archive.py
boumenot: Agreed, but I also think it reads clearer when all of the operators are inline.  Your eyes can just scroll down and see how ever comparison is made.

jasonzio: Which means the eyes glaze over and a flip of < for > or a missing (or extra) = is easy to miss. This may very well be a language-specific style; the C++ library style is very much as I described. I'm definitely not hardcore about this; having pushed back once, I'm done. :-)

PR #1112, opened 2018-04-16
Refactor subprocess startup to capture output on all Python versions

jasonzio:
Not yet used by exthandlers.py, still needs DCR tests, but the basic unit tests verify functionality on 2.7 and 3.5.


tests/utils/test_process_util.py
```diff
@@ -93,7 +102,74 @@ def test_format_stdout_stderr05(self):
...
+                                   preexec_fn=os.setsid)
+
+        if sys.version_info < (2, 7):
+            self.assertRaises(ExtensionError, capture_from_process_raw, process, cmd, 10)
+        else:
+            with self.assertRaises(ExtensionError) as ee:
+                capture_from_process_raw(process, cmd, 10)
+
+            body = str(ee.exception)
+            if sys.version_info >= (3, 2):
```
boumenot: I would just add this to test helpers so you can avoid the whole assertRegex vs assertRegexpMatches.  Save the reading of an if statement.

azurelinuxagent/common/utils/processutil.py
```diff
@@ -0,0 +1,226 @@
...
+    Captures stdout and stderr from an already-created process.
+
+    :param subprocess.Popen process: Created by subprocess.Popen()
+    :param str cmd: The command string to be included in any exceptions
+    :param int timeout: Number of seconds the process is permitted to run
+    :return: The stdout and stderr captured from the process
+    :rtype: (str, str)
+    :raises ExtensionError: if a timeout occurred or if anything was raised by Popen.communicate()
+    """
+    if not timeout:
```
boumenot: not timeout seems weird.  Does this mean timeout == 0 or timeout <= 0 or timeout < 0 or something else?

azurelinuxagent/common/utils/processutil.py
```diff
@@ -0,0 +1,226 @@
...
+        if e.errno != ESRCH:
+            raise
+        pass    # If the process is already gone, that's fine
+
+
+def capture_from_process_modern(process, cmd, timeout):
+    try:
+        stdout, stderr = process.communicate(timeout=timeout)
+    except subprocess.TimeoutExpired:
+        os.killpg(os.getpgid(process.pid), 9)
```
boumenot: You don't call _destroy_process because it closes the STDOUT/STDERR and you need to call .communicate() on more time?

azurelinuxagent/common/utils/processutil.py
```diff
@@ -0,0 +1,226 @@
...
+    watcher = Process(target=kill_on_timeout, args=(process.pid, timeout))
+    watcher.start()
+
+    try:
+        # Now, block "forever" waiting on process. If the timeout-limited Event wait in the watcher pops,
+        # it will kill the process and Popen.communicate() will return
+        stdout, stderr = process.communicate()
+    except OSError as e:
+        _destroy_process(process, signal.SIGKILL)
+        raise ExtensionError("Error while running '{0}': {1}".format(cmd, e.strerror))
```
boumenot: Why is this e.strerror and the others are just e?

azurelinuxagent/common/utils/processutil.py
```diff
@@ -0,0 +1,226 @@
...
+        """
+        Check for the continued existence of pid once per second. If pid no longer exists, exit with code 0.
+        If timeout (in seconds) elapses, kill pid and exit with code 1.
+        """
+        for iteration in range(watcher_timeout):
+            time.sleep(1)
+            try:
+                os.kill(pid, 0)
+            except OSError as ex:
+                if 3 == ex.errno:   # Process no longer exists
```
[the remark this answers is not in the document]
jasonzio: ESRCH. Fixed.

azurelinuxagent/common/utils/processutil.py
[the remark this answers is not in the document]
jasonzio: Yep.

jasonzio: Exactly correct. .communicate() will read stdout/stderr until EOF and then close them; .communicate() will then .wait() on the process, reaping the zombie. I need do nothing more than kill the process, then let .communicate() finish its job.

azurelinuxagent/common/utils/processutil.py
jasonzio: Everywhere I catch OSError, I explicitly display strerror, which is the string description of the underlying OS error (e.g. "File not found" or "No such process"). That string description is guaranteed to be present for OSError; it's actually the hallmark of that exception, which surfaces an errno from the underlying OS operations.

For all other exception types, the best I can do is convert the exception to a string and let the poor human find the pony under all the manure. For OSError, I can present a nice, clean pony.

azurelinuxagent/common/utils/processutil.py
```diff
@@ -0,0 +1,226 @@
...
+        stdout, stderr = process.communicate()
+    except OSError as e:
+        _destroy_process(process, signal.SIGKILL)
+        raise ExtensionError("Error while running '{0}': {1}".format(cmd, e.strerror))
+    except RuntimeError as e:
+        _destroy_process(process, signal.SIGKILL)
+        raise ExtensionError("RuntimeError while running '{0}': {1}".format(cmd, e))
+    except Exception as e:
+        _destroy_process(process, signal.SIGKILL)
+        raise ExtensionError("Exception while running '{0}': {1}".format(cmd, e))
```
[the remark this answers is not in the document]
boumenot: Makes sense.

azurelinuxagent/common/utils/processutil.py
boumenot: I would add a comment.

azurelinuxagent/common/utils/processutil.py
boumenot: I wish there was a way to resolve a comment.

azurelinuxagent/common/utils/processutil.py
boumenot: I think explicit is better than conversion.  YMMV.

tests/utils/test_process_util.py
```diff
@@ -93,7 +102,74 @@ def test_format_stdout_stderr05(self):
...
+        process = subprocess.Popen(cmd,
+                                   shell=True,
+                                   stdout=subprocess.PIPE,
+                                   stderr=subprocess.PIPE,
+                                   env=os.environ)
+
+        actual = capture_from_process(process, cmd)
+        self.assertEqual(expected, actual)
+
+    def test_process_timeout(self):
```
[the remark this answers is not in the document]
jasonzio: I'm going to leave this one as-is.

PR #1114, opened 2018-04-17
Fix PEP8 nits in exthandlers. Use processutil to capture subproc output

jasonzio:


azurelinuxagent/ga/exthandlers.py
```diff
@@ -925,6 +886,14 @@ def get_status_file_path(self):
         return seq_no, path
 
     def collect_ext_status(self, ext):
+        """
+        TODO This method takes a parameter but ignores it. Should it return status for just that ext, or all of them?
+        Current behavior is to return an object that reflects the entire status. Should the implementation reach
+        inside the ext_status object and strip it down to just the information about ext?
+
```
boumenot: I would delete these lines if you're not going to comment them.

PR #1120, opened 2018-04-19
do not call decode_config twice

### LINKED ISSUE #1119: Exception retrieving artifacts profile: TypeError: decoding str is not supported
I am tracking a annoying issue in the BVTs with the following signature.

```text
Exception retrieving artifacts profile: TypeError: decoding str is not supported
```

I increased debugging, and found this call stack.

```text
2018/04/18 06:03:51.919139 WARNING Exception retrieving artifacts profile: Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/azurelinuxagent/common/protocol/wire.py", line 1135, in get_artifacts_profile
    profile = self.decode_config(config)
  File "/usr/local/lib/python3.5/dist-packages/azurelinuxagent/common/protocol/wire.py", line 563, in decode_config
    xml_text = ustr(data, encoding='utf-8')
TypeError: decoding str is not supported
```

When I read the code for `get_artifacts_profile`, I noticed that decode_config is called twice.  The method `fetch` calls decode_config, and then `get_artifacts_profile` calls decode_config.  It appears that you cannot call decode_config twice with the same data.

The method decode_config is confusing because it has variables like xml_text, but the data passed in this case is actually JSON.  Azure does not do us any favors because some systems return XML and other returns JSON.  The agent could or should be smarter about this.

The bug was introduced with d0b583cc.

boumenot:
Closes #1119

### DIFF
--- a/tests/protocol/test_wire.py
+++ b/tests/protocol/test_wire.py
@@ -342,7 +342,7 @@ class TestWireProtocol(AgentTestCase):
         wire_protocol_client.ext_conf.artifacts_profile_blob = testurl
         goal_state = GoalState(WireProtocolData(DATA_FILE).goal_state)
         wire_protocol_client.get_goal_state = Mock(return_value=goal_state)
-        wire_protocol_client.fetch = Mock(side_effect=[None, '{"onHold": "true"}'.encode('utf-8')])
+        wire_protocol_client.fetch = Mock(side_effect=[None, '{"onHold": "true"}'])
         with patch.object(HostPluginProtocol,
                           "get_artifact_request",
                           return_value=['dummy_url', {}]) as artifact_request:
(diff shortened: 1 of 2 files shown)


tests/protocol/test_wire.py
```diff
@@ -342,7 +342,7 @@ def test_get_in_vm_artifacts_profile_host_ga_plugin(self, *args):
         wire_protocol_client.ext_conf.artifacts_profile_blob = testurl
         goal_state = GoalState(WireProtocolData(DATA_FILE).goal_state)
         wire_protocol_client.get_goal_state = Mock(return_value=goal_state)
-        wire_protocol_client.fetch = Mock(side_effect=[None, '{"onHold": "true"}'.encode('utf-8')])
```
boumenot: We were fooling ourselves with all of these mocks.  We improperly characterized the return type of this method.  It is not bytes, it is a string. 😭

PR #1122, opened 2018-04-19
overwrite if an archived file exists

boumenot:
There were two failures cases that need to be addressed.

 1. If purge() is called before archive() the history directory may not exist yet.  The first call should create the directory if it does not exist.
 1. If the same file is archived it should be overwritten.  The same file was being overwritten because the timestamp was defined at class instantiation time instead of at flush time.

### DIFF
--- a/azurelinuxagent/common/utils/archive.py
+++ b/azurelinuxagent/common/utils/archive.py
@@ -1,14 +1,13 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the Apache License.
-
 import os
 import re
 import shutil
 import zipfile
 
-from datetime import datetime
-
 from azurelinuxagent.common.utils import fileutil
+
+import azurelinuxagent.common.logger as logger
 
 
 """
@@ -34,6 +33,8 @@ is /var/lib/waagent.
 The timestamp is an ISO8601 formatted value.
 """
 
+ARCHIVE_DIRECTORY_NAME = 'history'
+
 MAX_ARCHIVED_STATES = 50
 
 CACHE_PATTERNS = [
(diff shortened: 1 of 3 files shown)


azurelinuxagent/common/utils/archive.py
```diff
@@ -76,19 +78,21 @@ def _get_files_to_archive(self):
...
-    def _mkdir(self):
+    def _mkdir(self, timestamp):
         try:
-            fileutil.mkdir(self._target, mode=0o700)
+            d = self.history_dir(timestamp)
+            fileutil.mkdir(d, mode=0o700)
             return True
-        except:
+        except IOError as e:
             return False
```
jasonzio: Is there a way to log e.strerror at this point? Otherwise we won't know why mkdir failed, and that might be useful.

azurelinuxagent/common/utils/archive.py
```diff
@@ -164,6 +168,12 @@ class StateArchiver(object):
     def __init__(self, lib_dir):
         self._source = os.path.join(lib_dir, 'history')
 
+        if not os.path.isdir(self._source):
+            try:
+                fileutil.mkdir(self._source, mode=0o700)
+            except IOError as e:
+                self.logger.error(u"Failed to create archive history dir: {0}", e)
```
jasonzio: Just log `e.strerror`, which is the description text for the actual error. I don't think you need the entire exception. Also, log the name of the directory (self._source) that couldn't be created.

tests/utils/test_archive.py
```diff
@@ -155,6 +155,57 @@ def test_archive02(self):
...
+
+        timestamp = datetime.utcnow()
+
+        _write_goal_state_files(temp_files)
+        test_subject = StateFlusher(self.tmp_dir)
+        test_subject.flush(timestamp)
+
+        # re-write all of the same files, and flush again. .flush()
+        # should overwrite the existing ones
+
```
jasonzio: I don't see the test actually making sure the contents of the file, after the second flush, are the correct contents. In other words, if file X held `123` as its contents, was flushed, then was changed to hold `456` and was flushed again, the test should confirm that X contains `456` in the archive directory.

azurelinuxagent/common/utils/archive.py
```diff
@@ -52,18 +52,20 @@ class StateFlusher(object):
     def __init__(self, lib_dir):
         self._timestamp = datetime.now()
```
hglkrijger: Should this be `utcnow()`?

azurelinuxagent/common/utils/archive.py
```diff
@@ -52,18 +52,20 @@ class StateFlusher(object):
...
 
-        if self._mkdir():
-            self._archive(files)
+        if self._mkdir(timestamp):
+            self._archive(files, timestamp)
         else:
             self._purge(files)
 
+    def history_dir(self, timestamp):
+        return os.path.join(self._source, 'history', timestamp.isoformat())
```
hglkrijger: Maybe `history` should be a constant.

tests/utils/test_archive.py
boumenot: Fixed.

PR #1127, opened 2018-04-20
translate bad status values to a correct one

### LINKED ISSUE #1002: Extension install failures timeout
The Windows GA reports a status which allows a fast failure, however the Linux GA just reports 'Not ready' which essentially waits for a CRP timeout. We should investigate if there is a substatus we are missing to allow a fast failure.

boumenot:
Closes #1002


boumenot:
I considered it, but I don't have a failing example nor do I know how to produce one.  In the cases that I do have, the sub-status is empty.

jasonzio:
LGTM

PR #1131, opened 2018-04-24
delete extension package on uninstall

### LINKED ISSUE #1054: /var/lib/waagent getting big after OMS agent upgrades
Hello
OMS downloaded zip files are not removed and historical data is accumulating over new releases, causing FS space consumption.
Since waagent is the installation handler, any ETA for including a purge process in order to only keep the n and n-1 artifacts on the local installation path ?

boumenot:
Delete the extension's package when an extension is uninstalled.  An extension may be uninstalled via an explicit action from the user, or due to an extension upgrade.

Closes #1054

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -408,6 +408,7 @@ class ExtHandlersHandler(object):
                 ext_handler_i.copy_status_files(old_ext_handler_i)
                 ext_handler_i.update()
                 old_ext_handler_i.uninstall()
+                old_ext_handler_i.rm_ext_pkg_zip()
                 old_ext_handler_i.rm_ext_handler_dir()
                 ext_handler_i.update_with_install()
         else:
@@ -432,6 +433,7 @@ class ExtHandlersHandler(object):
             if handler_state == ExtHandlerState.Enabled:
                 ext_handler_i.disable()
             ext_handler_i.uninstall()
+            ext_handler_i.rm_ext_pkg_zip()
         ext_handler_i.rm_ext_handler_dir()
 
     def report_ext_handlers_status(self):
@@ -640,6 +642,14 @@ class ExtHandlerInstance(object):
         self.set_logger()
         return
 
+    @property
+    def name(self):
+        return self.ext_handler.name
+
+    @property
+    def version(self):
+        return self.ext_handler.properties.version
+
     def set_logger(self):
         prefix = "[{0}]".format(self.get_full_name())
         self.logger = logger.Logger(logger.DEFAULT_LOGGER, prefix)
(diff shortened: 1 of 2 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -708,6 +718,70 @@ def report_event(self, message="", is_success=True, duration=0):
...
+        :rtype: str
+        """
+        if self.pkg_file is not None:
+            return self.pkg_file
+
+        def normalize_ext_name(f):
+            parts = f.split('__')
+            if len(parts) != 3:
+                return None
+            return "{0}.{1}.{2}".format(parts[0], parts[1], parts[2])
```
jasonzio: Consider `return ".".join(parts)` as a clearer expression of what you're doing (dot-connect these three elements). Both work, and I can live with what you wrote.

To be honest, the right answer is to refactor the implementation into a static method or simple function that takes an extension publisher, an extension type name, and a version, and returns the canonicalized string.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -708,6 +718,70 @@ def report_event(self, message="", is_success=True, duration=0):
...
+
+        try:
+            normalized_ext_name = "{0}.{1}.zip".format(
```
jasonzio: Similarly,
```
normalized_ext_name = ".".join(self.name, str(self.version), "zip")
```
(Don't forget to use the new properties you just added.)

azurelinuxagent/ga/exthandlers.py
```diff
@@ -830,7 +904,16 @@ def uninstall(self):
...
             self.report_event(message=ustr(e), is_success=False)
-    
+
+    def rm_ext_pkg_zip(self):
+        try:
+            fn = self.try_get_pkg_file()
+            if fn is not None and os.path.exists(fn):
+                self.logger.info("Remove extension package: {0}".format(fn))
+                os.remove(fn)
+        except IOError as e:
```
jasonzio: os.remove() is documented as raising OSError if anything goes wrong, in Python 2.7 and in 3.5. Later versions of Python alias IOError to OSError. as well. Upshot: this should be OSError.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -708,6 +718,70 @@ def report_event(self, message="", is_success=True, duration=0):
         add_event(name=self.ext_handler.name, version=ext_handler_version, message=message,
                   op=self.operation, is_success=is_success, duration=duration)
 
+    def try_get_pkg_file(self):
```
hglkrijger: We have a handle to this file when we unzip it. Why not either:

- remove the zip right after we successfully unzip
- persist that reference until the point you want to remove the zip

I vote for the first option.

jasonzio: Are there patterns of recovery from install failures where, today, we'd reuse the zip already on the disk? Or does the agent always download a fresh copy of the zip?

azurelinuxagent/ga/exthandlers.py
boumenot: I was trying to find a better way to express line 768, and `return ".".join(parts)` is what I want.

I would take it one step further in a future commit.  It deserves a proper type to encase these behaviors. The fact that PIR and CRP cannot agree on a consistent naming conventions stinks, but that's life.  This is not a string this is a type called ExtensionName (or some such).

azurelinuxagent/ga/exthandlers.py
boumenot: Gah! I thought I corrected those, thank you!

azurelinuxagent/ga/exthandlers.py
boumenot: I should change the code I copied from then, done.

azurelinuxagent/ga/exthandlers.py
boumenot: I have the same concerns that Jason does, which is why I opted for this path.  I **thought** there were situations were the agent would continually download something in the case of failure, and I wanted to avoid that unless there was an explicit uninstall.

boumenot:
After speaking with @hglkrijger, I have decided to go in a different direction. 

Instead of discovering the file name at run time, I will simply delete the .zip after it has been extracted.  My desire to not unnecessarily download packages again and again is not solved in this PR, so it would be better to simply delete the .zip after the extraction.  I will submit a new PR. The agent always re-downloads the .zip file, so there is no savings that can be realized.

PR #1132, opened 2018-04-25
issue warning if HWADDR is in ifcfg-

### LINKED ISSUE #1065: deprovision step should cleanup HWADDRR if present
After running the deprovisioning step, my VHD import still failed to complete because I forgot to cleanup a harcoded MAC. I would say is useful to add such a cleanup into the step itself.

```
Bringing up interface eth0:  Device eth0 has different MAC address than expected, ignoring.
[FAILED]
```

`cat /etc/sysconfig/network-scripts/ifcfg-eth0`

```
DEVICE=eth0
HWADDR=00:57:57:a5:76:78
ONBOOT=yes
BOOTPROTO=dhcp
DHCPCLASS=
```

boumenot:
Closes #1065

### DIFF
--- a/azurelinuxagent/pa/deprovision/default.py
+++ b/azurelinuxagent/pa/deprovision/default.py
@@ -24,17 +24,18 @@ import sys
 
 import azurelinuxagent.common.conf as conf
 import azurelinuxagent.common.utils.fileutil as fileutil
-import azurelinuxagent.common.utils.shellutil as shellutil
 
 from azurelinuxagent.common.exception import ProtocolError
 from azurelinuxagent.common.osutil import get_osutil
 from azurelinuxagent.common.protocol import get_protocol_util
+
 
 def read_input(message):
     if sys.version_info[0] >= 3:
         return input(message)
     else:
         return raw_input(message)
+
 
 class DeprovisionAction(object):
     def __init__(self, func, args=[], kwargs={}):
@@ -44,6 +45,7 @@ class DeprovisionAction(object):
 
     def invoke(self):
         self.func(*self.args, **self.kwargs)
+
 
 class DeprovisionHandler(object):
     def __init__(self):
@@ -72,7 +74,6 @@ class DeprovisionHandler(object):
                          "will be deleted.").format(username))
         actions.append(DeprovisionAction(self.osutil.del_account, 
                                          [username]))
-
 
     def regen_ssh_host_key(self, warnings, actions):
         warnings.append("WARNING! All SSH host key pairs will be deleted.")
@@ -118,7 +119,6 @@ class DeprovisionHandler(object):
         actions.append(DeprovisionAction(fileutil.rm_files,
                                          ["/var/lib/NetworkManager/dhclient-*.lease"]))
 
-
     def del_lib_dir_files(self, warnings, actions):
         known_files = [
             'HostingEnvironmentConfig.xml',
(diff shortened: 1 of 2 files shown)


azurelinuxagent/pa/deprovision/default.py
```diff
@@ -189,6 +189,11 @@ def reset_hostname(self, warnings, actions):
         actions.append(DeprovisionAction(self.osutil.set_dhcp_hostname, 
                                          localhost))
 
+    def check_ifcfg_hwaddr(self, warnings):
+        for f in glob.glob("/etc/sysconfig/network-scripts/ifcfg-*"):
+            if os.path.isfile(f) and fileutil.findre_in_file(f, '^\s*HWADDR='):
```
jasonzio: You should probably do this for MACADDR as well (see [the CentOS docs](https://www.centos.org/docs/5/html/Deployment_Guide-en-US/s1-networkscripts-interfaces.html)).

As I read that doc, the file syntax mandates there be no whitespace before the equal-sign, so this is good.
```
if os.path.isfile(f) and fileutil.findre_in_file(f, '^\s*(HW|MAC)ADDR='):
```

azurelinuxagent/pa/deprovision/default.py
```diff
@@ -189,6 +189,11 @@ def reset_hostname(self, warnings, actions):
...
+                warnings.append('WARNING! {0} has a HWADDR value, which may prevent successful re-deployment')
```
jasonzio: Adjust if you add MACADDR to the test.

tests/pa/test_deprovision.py
```diff
@@ -157,15 +156,63 @@ def test_del_lib_dir_files(self,
...
+
         deprovision_handler = get_deprovision_handler(distro_name,
                                                       distro_version,
                                                       distro_full_name)
-        warnings, actions = deprovision_handler.setup(deluser=False)
-        assert any("/etc/resolv.conf" in w for w in warnings)
+
+        warnings = []
+        deprovision_handler.check_ifcfg_hwaddr(warnings)
+        assert any("HWADDR" in w for w in warnings)
```
jasonzio: Why is this a naked assert instead of a self.assertSomething() method?

tests/pa/test_deprovision.py
```diff
@@ -157,15 +156,63 @@ def test_del_lib_dir_files(self,
...
+
+        mock_glob.return_value = [ifcfg_eth0_fn]
+
+        deprovision_handler = get_deprovision_handler(distro_name,
+                                                      distro_version,
+                                                      distro_full_name)
+
+        warnings = []
+        deprovision_handler.check_ifcfg_hwaddr(warnings)
+        assert not any("HWADDR" in w for w in warnings)
```
jasonzio: Why a naked assert instead of a unittest.assertSomething() method?

azurelinuxagent/pa/deprovision/default.py
```diff
@@ -189,6 +189,11 @@ def reset_hostname(self, warnings, actions):
         actions.append(DeprovisionAction(self.osutil.set_dhcp_hostname, 
                                          localhost))
 
+    def check_ifcfg_hwaddr(self, warnings):
+        for f in glob.glob("/etc/sysconfig/network-scripts/ifcfg-*"):
```
hglkrijger: Doesn't this need to be distro-specific?

azurelinuxagent/pa/deprovision/default.py
```diff
@@ -135,9 +135,9 @@ def del_lib_dir_files(self, warnings, actions):
         ]
 
         lib_dir = conf.get_lib_dir()
-        files = [f for f in \
-                    [os.path.join(lib_dir, kf) for kf in known_files] \
-                        if os.path.isfile(f)]
+        files = [f for f in
+                 [os.path.join(lib_dir, kf) for kf in known_files]
+                 if os.path.isfile(f)]
```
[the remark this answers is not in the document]
boumenot: I made these changes to address PEP warnings.  If you feel strongly about these changes I will move them into their own PR.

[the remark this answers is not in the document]
jasonzio: Not at all a strong feeling. The code is correct as-is.

boumenot:
After speaking with Hans, I have decided to not move forward with this change.

PR #1145, opened 2018-05-02
Split agent changes

richawil:
Please consider these proposed changes to the Azure/WALinuxAgent.

Allow me to provide some background information and the reason for the change request.

We run a virtual appliance in the Azure cloud.  In the Marketplace, it is called the Cisco CSR1000v.  It is a multi-protocol router.

After the initial boot up, control of all of the interfaces on the virtual machine are taken away from the kernel and given to the router.

In order to run the Linux agent code, we have to host the code in a container and provide network access to the container by attaching it to an internal port of the router.
Running the agent code in the container poses a few problems, the largest of which is the absence of a DHCP server on the internal network.

To work around this issue, we actually run the agent code twice.
The first time the agent code is run at boot time of the virtual machine, while the kernel still has control of the network interfaces.  During this time, DHCP is run, and provisioning is enabled.  Once the agent code reports it is ready to the fabric server, this instantiation of the agent is terminated.

Then, after the router has taken control of the interfaces, the agent process is again started, but this time within a container.  Using the waagent.conf file, we disable provisioning for this agent instance.

The proposed changes are to allow this split mode of agent operation.  
A “DHCP.Enabled” variable is added to the configuration file.
Places where the agent tries to run DHCP will now check the configuration variable.
We also derive the WireServer endpoint from the configuration file when DHCP is disabled.

One other change involves the product ID file.  In the CSR1000v, this file is used for instance identification for the purposes of software licensing.  The file is not accessible, even to the root user.  This prevents anyone from modifying the product ID in order to enable software licenses (granted to another router).
A try/except block has been added around the code that tries to read the product ID file.

### DIFF
--- a/azurelinuxagent/common/conf.py
+++ b/azurelinuxagent/common/conf.py
@@ -94,6 +94,7 @@ __SWITCH_OPTIONS__ = {
     "OS.CheckRdmaDriver" : False,
     "Logs.Verbose" : False,
     "Provisioning.Enabled" : True,
+    "DHCP.Enabled" : True,
     "Provisioning.UseCloudInit" : False,
     "Provisioning.AllowResetSysUser" : False,
     "Provisioning.RegenerateSshHostKeyPair" : False,
@@ -127,7 +128,8 @@ __STRING_OPTIONS__ = {
     "ResourceDisk.MountPoint" : "/mnt/resource",
     "ResourceDisk.MountOptions" : None,
     "ResourceDisk.Filesystem" : "ext3",
-    "AutoUpdate.GAFamily" : "Prod"
+    "AutoUpdate.GAFamily" : "Prod",
+    "Provisioning.Endpoint" : "168.63.129.16"
 }
 
 __INTEGER_OPTIONS__ = {
@@ -239,6 +241,12 @@ def get_ssh_host_keypair_mode(conf=__conf__):
 
 def get_provision_enabled(conf=__conf__):
     return conf.get_switch("Provisioning.Enabled", True)
+
+def get_dhcp_enabled(conf=__conf__):
+    return conf.get_switch("DHCP.Enabled", True)
+
+def get_provision_endpoint(conf=__conf__):
+    return conf.get("Provisioning.Endpoint", "168.63.129.16")
 
 def get_provision_cloudinit(conf=__conf__):
     return conf.get_switch("Provisioning.UseCloudInit", False)
--- a/azurelinuxagent/common/conf.py
+++ b/azurelinuxagent/common/conf.py
@@ -94,7 +94,6 @@ __SWITCH_OPTIONS__ = {
     "OS.CheckRdmaDriver" : False,
     "Logs.Verbose" : False,
     "Provisioning.Enabled" : True,
-    "DHCP.Enabled" : True,
     "Provisioning.UseCloudInit" : False,
     "Provisioning.AllowResetSysUser" : False,
     "Provisioning.RegenerateSshHostKeyPair" : False,
@@ -128,8 +127,7 @@ __STRING_OPTIONS__ = {
     "ResourceDisk.MountPoint" : "/mnt/resource",
     "ResourceDisk.MountOptions" : None,
     "ResourceDisk.Filesystem" : "ext3",
-    "AutoUpdate.GAFamily" : "Prod",
-    "Provisioning.Endpoint" : "168.63.129.16"
+    "AutoUpdate.GAFamily" : "Prod"
 }
 
 __INTEGER_OPTIONS__ = {
@@ -241,12 +239,6 @@ def get_ssh_host_keypair_mode(conf=__conf__):
 
 def get_provision_enabled(conf=__conf__):
     return conf.get_switch("Provisioning.Enabled", True)
-
-def get_dhcp_enabled(conf=__conf__):
-    return conf.get_switch("DHCP.Enabled", True)
-
-def get_provision_endpoint(conf=__conf__):
-    return conf.get("Provisioning.Endpoint", "168.63.129.16")
 
 def get_provision_cloudinit(conf=__conf__):
     return conf.get_switch("Provisioning.UseCloudInit", False)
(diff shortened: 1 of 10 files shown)


azurelinuxagent/common/conf.py
```diff
@@ -127,7 +128,8 @@ def load_conf_from_file(conf_file_path, conf=__conf__):
     "ResourceDisk.MountPoint" : "/mnt/resource",
     "ResourceDisk.MountOptions" : None,
     "ResourceDisk.Filesystem" : "ext3",
-    "AutoUpdate.GAFamily" : "Prod"
+    "AutoUpdate.GAFamily" : "Prod",
+    "Provisioning.Endpoint" : "168.63.129.16"
```
boumenot: Our contract is that we discover this via DHCP.  The value is always the same, but that's an implementation detail. Azure is free to change this value, and may. I believe this value is different in Azure Stack.

azurelinuxagent/common/conf.py
```diff
@@ -127,7 +128,8 @@ def load_conf_from_file(conf_file_path, conf=__conf__):
     "ResourceDisk.MountPoint" : "/mnt/resource",
     "ResourceDisk.MountOptions" : None,
     "ResourceDisk.Filesystem" : "ext3",
-    "AutoUpdate.GAFamily" : "Prod"
+    "AutoUpdate.GAFamily" : "Prod",
```
boumenot: Have you considered making a distinct class for your appliance?  I think that would fit into the current model better.

richawil: Can you elaborate, or point me at some documentation that describes the class model and the ability to define new ones / refine existing ones?

richawil: Please provide a reference to documentation on the OS factory and instructions on defining a custom distribution.

boumenot: At this point, the documentation is the code.  I recommend you start with the [osutil](https://github.com/Azure/WALinuxAgent/tree/master/azurelinuxagent/common/osutil), and see how it is handled for the other flavors. Let me know if this does not make sense.

richawil:
Thank you for your helpful comments. Based upon the feedback, I have made the following changes:
Defined a new custom Linux distribution named iosxe.
Created a custom waagent.conf file for the iosxe distribution.
Added a new osutil function to check if DHCP is available in a distribution.  The default answer is that DHCP is available.
Use the osutil function to bypass the use of DHCP when it is not available.

azurelinuxagent/common/osutil/default.py
```diff
@@ -287,12 +287,12 @@ def get_instance_id(self):
         '''
         if os.path.isfile(PRODUCT_ID_FILE):
             s = fileutil.read_file(PRODUCT_ID_FILE).strip()
-
+
```
boumenot: NIT: whitespace.

azurelinuxagent/common/osutil/default.py
```diff
@@ -287,12 +287,12 @@ def get_instance_id(self):
...
         if os.path.isfile(PRODUCT_ID_FILE):
             s = fileutil.read_file(PRODUCT_ID_FILE).strip()
-
+            
         else:
             rc, s = shellutil.run_get_output(DMIDECODE_CMD)
             if rc != 0 or UUID_PATTERN.match(s) is None:
                 return ""
-
+
```
boumenot: NIT: whitespace

richawil: Accidentally removed whitespace line 295.  Performed a commit to put it back.  This now matches the original code.

richawil:
Repaired syntax errors.

richawil:
Are there any other issues with the modified changes?

azurelinuxagent/common/osutil/iosxe.py
```diff
@@ -0,0 +1,84 @@
...
+        return shellutil.run("systemctl enable waagent", chk_err=False)
+
+    def unregister_agent_service(self):
+        return shellutil.run("systemctl disable waagent", chk_err=False)
+
+    def openssl_to_openssh(self, input_file, output_file):
+        DefaultOSUtil.openssl_to_openssh(self, input_file, output_file)
+
+    def is_dhcp_available(self):
+        return (False, '168.63.129.16')
```
boumenot: There is a constant for the value 168.63.129.16 called KNOWN_WIRESERVER_IP.  Please use this instead.  Sorry I didn't catch this the first time.

config/iosxe/waagent.conf
```diff
@@ -0,0 +1,119 @@
...
+# EnableOverProvisioning=y
+
+# Allow fallback to HTTP if HTTPS is unavailable
+# Note: Allowing HTTP (vs. HTTPS) may cause security risks
+# OS.AllowHTTP=n
+
+# Add firewall rules to protect access to Azure host node services
+# Note:
+# - The default is false to protect the state of existing VMs
+OS.EnableFirewall=y
```
boumenot: This is an appliance that has special networking.  Are you sure you want to enable this option?  This will cause the agent to call iptables, and I don't know if that is available on this appliance.

jasonzio:
LGTM

azurelinuxagent/common/osutil/iosxe.py
hglkrijger: +1 on that

azurelinuxagent/common/osutil/iosxe.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def get_instance_id(self):
```
hglkrijger: Isn't this the same as the base implementation?

azurelinuxagent/common/protocol/util.py
```diff
@@ -145,13 +145,26 @@ def _set_wireserver_endpoint(self, endpoint):
...
-            try:
-                self.dhcp_handler.run()
-            except DhcpError as e:
-                raise ProtocolError(ustr(e))
-            endpoint = self.dhcp_handler.endpoint
-        
+            '''
+            Check if DHCP can be used to get the wire protocol endpoint
+            '''
+            (dhcp_available, conf_endpoint) =  self.osutil.is_dhcp_available()
```
hglkrijger: Maybe a better name for `conf_endpoint`, perhaps `static_endpoint` or similar.

richawil:
What is the process for merging the proposed changes?  Do I initiate the merge once all comments have been resolved?  Or does the repo owner(s) initiate the merge?

hglkrijger:
Sorry for the delay on this, merging.

PR #1148, opened 2018-05-03
ignore empty URLs

### LINKED ISSUE #1147: ExtensionsConfig May Not Contain a failoverLocation Attribute
The agent expects ExtensionsConfig to have a location and failoverLocation for each plugin.  This has been proven to not be true for all regions.  I consider this to be a bug upstream, but the agent should be robust enough to handle this case.

boumenot:
Closes #1147


boumenot:
I opened #1161 to address the telemetry issue.  There is some sort of circular dependency issue that manifest on CI, but not locally.  I will debug it later, and add the necessary event.

Issue #1092, opened 2018-04-02
Memory Leak in "python -u /usr/sbin/waagent -run-exthandlers"


v-eslop:
I have a new openSUSE 42.3 VM in Azure with 2GB RAM, we noticed that the memory usage has been increasing with the days, after 9 days it is consuming 21.94% (400M).

PID USER      PR  NI    VIRT       RES     SHR S  %CPU  %MEM        TIME+ COMMAND
16296 root   20   0  658972 429484   7508 S   0.667     21.94   224:32.85 python

openSUSE1:/home/ # ps -aux | grep python
root      15392  0.0  0.8  99536 17052 ?        Ss   Mar23   0:00 /usr/bin/python -u /usr/sbin/waagent -daemon
root      16296  1.4 21.9 658972 429484 ?       Sl   Mar23 224:35 python -u /usr/sbin/waagent -run-exthandlers

openSUSE1:/home/ # waagent --version
WALinuxAgent-2.2.18 running on suse 42.3
Python: 2.7.13
Goal state agent: 2.2.18

It is running with python 2.7, the waagent.log* only show HeartBeats and the machine does not have any extension installed.
Any help on this will be appreciate.

v-eslop:
Thanks for your response Boumenot, I have another openSUSE VM where I installed the github waagent (2.2.21). It has been running almost 8 days and the memory usage is 636MB VIRT and 400MB RES and it is loading the version 2.2.25:

openSUSE2:/home/ # ps -aux | grep python | grep -v grep
root       1335  0.0  0.8  94044 15716 ?        Ss   Mar26   0:00 /usr/bin/python -u /usr/sbin/waagent -daemon
root       1477  1.5 20.5 636864 403140 ?       Sl   Mar26 169:50 python -u bin/WALinuxAgent-2.2.25-py2.7.egg -run-exthandlers

I have another SLES 12sp3 with 2.2.18 and similar config/resources, the memory usage is 100MB VIRT and 22MB RES.

Pistahh:
For the record, I ran into the same issue (on Debian) with 2.2.14 (not sure why the command line says 2.2.22):
```
USER        PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root     129607  0.9 82.8 1833120 1630728 ?     Sl   Feb22 634:08 python3 -u bin/WALinuxAgent-2.2.22-py2.7.egg -run-exthandlers
```

[...]

PR #1154, opened 2018-05-08
set the Connection close header

boumenot:
There is a slow-ish memory leak in the agent's HTTP request library.  The leak has been confirmed on OpenSuSE and Debian thus far. I have not been able to reproduce it on Ubuntu, RHEL, or CentOS.

I created a [branch](https://github.com/Azure/WALinuxAgent/tree/pr-tracemalloc) with tracemalloc instrumentation to help track down the issue. The leak appeared with the following file and line context.

```text
Top 50 lines
#1: python3.5/socket.py:647: 13609.4 KiB
    self._sock = None
```

This was just useful enough to identify that the issue had to do with sockets.  (tracemalloc supports grabbing a larger stacktrace around the memory allocation source, but I was not able to get that additional context.  I have not debugged why.)  The couple of bare socket calls that the agent makes are infrequent. The agent does make a lot of HTTP requests, so I started with the HTTP library.

The first thing I noticed is that the agent never explicitly cleans up an HTTP connection when it is done with it.  Note that the following method never calls `resp.connection.close()`.

```python
    def put_block_blob(self, url, data):
        logger.verbose("Put block blob")
        headers = self.get_block_blob_headers(len(data))
        resp = self.client.call_storage_service(restutil.http_put, url, data, headers)
        if resp.status != httpclient.CREATED:
            raise UploadError(
                "Failed to upload block blob: {0}".format(resp.status))
```

I thought this was fine. I would expect Python to clean up once all references were removed, but this does not appear to be true at least in the case of Debian/OpenSuSE.  I could fix the code to explicitly close the connection, but there are many places and branches where the agent makes HTTP requests.  The HTTP library returns the HTTPResponse and not the connection, which is generally fine because you can **usually** reach the connection via the response. Another fix, that appears to be working, is to explicitly set the HTTP Connection header to close.

In my travels, I found that the Python requests library also had this issue.  Their fix was to explicitly close the connection after use.  I think that we should make that fix too, but that is a bigger and more invasive change.  I would prefer to ship this fix as a hotfix, and leave the bigger fix for another release.

### DIFF
--- a/azurelinuxagent/common/utils/restutil.py
+++ b/azurelinuxagent/common/utils/restutil.py
@@ -170,6 +170,8 @@ def _http_request(method, host, rel_uri, port=None, data=None, secure=False,
                   headers=None, proxy_host=None, proxy_port=None):
 
     headers = {} if headers is None else headers
+    headers['Connection'] = 'close'
+
     use_proxy = proxy_host is not None and proxy_port is not None
 
     if port is None:
--- a/tests/utils/test_rest_util.py
+++ b/tests/utils/test_rest_util.py
@@ -195,7 +195,7 @@ class TestHttpOperations(AgentTestCase):
         ])
         HTTPSConnection.assert_not_called()
         mock_conn.request.assert_has_calls([
-            call(method="GET", url="/bar", body=None, headers={'User-Agent': HTTP_USER_AGENT})
+            call(method="GET", url="/bar", body=None, headers={'User-Agent': HTTP_USER_AGENT, 'Connection': 'close'})
         ])
         self.assertEqual(1, mock_conn.getresponse.call_count)
         self.assertNotEquals(None, resp)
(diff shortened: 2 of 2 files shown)


jasonzio:
I think there's a close() missing in restutil.http_request (restutil.py line 324 right before the final `raise HttpError(...)` at the end of the function). If a retryable HTTP request hits the maximum number of attempts, it looks to me like the connection leaks.

boumenot:
There are close() statements missing all over the code.  My leaks were not caused by HTTP errors, they were caused by normal operations.  I think we should go back and fix all of that code, but that's a bigger change than I want to make for a hotfix.

boumenot:
I edit my original comment and removed the `Closes` keyword.  This change appears to address the memory leak, but I am still observing a slow upward trend after 12 hours.  It is unclear if it is reaching a steady state, or memory has been lost and outside of tracemalloc's view.

The real fix is #1156, but that change is bigger than I want for a hotfix.
This PR addresses #1092 to an extent.

PR #1160, opened 2018-05-09
Only use exact version for extension as specified in ExtensionsConfig

uiri:
This change removes the need for the upgrade guid and the upgrade policy, so I have removed those attributes and their related code.

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -315,8 +315,22 @@ class ExtHandlersHandler(object):
 
         try:
             state = ext_handler.properties.state
+            max_retries = 5
+            retry_count = 0
             while ext_handler_i.decide_version(target_state=state) is None:
-                continue
+                if retry_count >= max_retries:
+                    err_msg = "Unable to find manifest for extension {0}".format(ext_handler_i.ext_handler.name)
+                    ext_handler_i.set_operation(WALAEventOperation.Download)
+                    ext_handler_i.set_handler_status(message=ustr(err_msg), code=-1)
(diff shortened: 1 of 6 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -312,8 +312,16 @@ def handle_ext_handler(self, ext_handler, etag):
...
             while ext_handler_i.decide_version(target_state=state) is None:
-                continue
+                if retry_count >= max_retries:
+                    err_msg = "Unable to find manifest for extension {0}".format(ext_handler_i.ext_handler.name)
+                    ext_handler_i.set_handler_status(message=ustr(err_msg), code=-1)
+                    ext_handler_i.report_event(message=ustr(err_msg), is_success=False)
+                    return
+                time.sleep(2**retry_count * 10)
+                retry_count += 1
             if not ext_handler_i.is_upgrade and self.last_etag == etag:
```
jasonzio: If some retries were required but we eventually succeeded, I think we should still emit a telemetry event that summarizes the number of retries and the total delay before success. I can see wanting to put an alert on this, since "having to retry" might be an early warning sign of an incipient failure we could head off. No need to set handler status; just emit the telemetry event.

uiri:
Hi @boumenot @hglkrijger ,

Given that 2.2.27 has shipped, is this good to merge into master now?

I think the most recent 2-3 commits still need to be reviewed.

-Will

azurelinuxagent/ga/exthandlers.py
dougclayton: I agree we should have telemetry around transient failures.

Issue #855, opened 2017-08-23
Waagent deprovision should not be messing with the cloudinit directories


dsutherland1492:
Deleting and changing the files under /var/lib/cloud upon deprovision fundamentally breaks how per-once and per-instance works.   You also lose the ability to track instances created from the original image.  At the minimum, there should be a flag to turn this functionality completely off.

szarkos:
Can you clarify your scenario a bit more?  Deprovisioning assumes that you want to create an "image" from the VM, which could then be used to provision any number of new VMs with their own new cloud-config.  For snapshots, etc. you generally wouldn't deprovision a specialized image.

dsutherland1492:
Correct, but let's say you make a base image from your current machine and want to use per-once on the first instance of a machine provisoned from that image.  An example situation would be where your image is going to be deployed to the Azure Marketplace.  Ultimately, you need to use waagent deprovision before doing that but this would inevitability destroy anything in the per-once directory.

szarkos:
I think the primary reasoning was just to clean /var/lib/cloud/instances/<foo> to remove obsolete user data (customdata).  Perhaps we can discuss with the cloud-init team what's safe to remove here.

dsutherland1492:
The instances folder keeps tabs of what machines were previously
instaniated by cloud-init.  Clearing that directory would impact the
per-once scripts by making they run on each new instance rather than just
on the first instance.

hglkrijger:
I agree, we should leave the cloud-init directories alone during deprovision.

szarkos:
Cleaning stale user data is still a valid thing to do at deprovisioning.  I spoke with Scott Moser during the Cloud-Init Summit a couple weeks ago.  He said he will provide us with some details about what could be cleaned here without impacting per-once/per-instance scripts.  I can post that here once we have those details.

He also mentioned a cloud-init clean utility that apparently is on their backlog, which could replace this at some point (and should have various parameters to determine what to clean, etc.).

brendandixon:
@szarkos Did you get the details of what is safe to remove?

szarkos:
Not yet.  Scott thought the issue is worthy of wider discussion and recommended we discuss the issue on the cloud-init mailing list, which I've started now.

hglkrijger:
@szarkos - any updates on the guidance here?

hglkrijger:
@szarkos, any updates?

hglkrijger:
[...]

PR #1164, opened 2018-05-09
Fixes name resolution in Ubuntu 1804

### LINKED ISSUE #1151: No name resolution on new Ubuntu 18 vm
Apologies if this is not the correct place for this issue. 

My team is building templates with Packer and deploying VMs with `az group create deployment ...`, based loosely on these docs: https://docs.microsoft.com/en-us/azure/virtual-machines/linux/build-image-with-packer

I noticed that after the provisioning done by Packer, waagent deletes resolv.conf. This is fine, and works as expected when deploying a 16 vm. 

However as we attempt to move onto Ubuntu 18, when the vm is created, `/etc/resolv.conf` is missing, meaning it is not recreated when the vm is created from the disk image. `nslookup microsoft.com` times out, and no names are resolved when using eg curl. `systemd-resolve` however does work. 

Is this something waagent is responsible for?

hglkrijger:
- fixes #1151 

In prior versions of Ubuntu we remove `/etc/resolv.conf` on deprovision if it has been linked to anything but `/run/resolvconf/resolv.conf`. We have the option of keeping that pattern of behavior with 1804, and removing it if it no longer points to `/run/systemd/resolve/stub-resolv.conf`, however I believe the right fix is to change this behavior and stop deleting files we do not own.


azurelinuxagent/pa/deprovision/ubuntu.py
```diff
@@ -40,3 +40,13 @@ def del_resolv(self, warnings, actions):
             files_to_del = ["/etc/resolvconf/resolv.conf.d/tail",
                             "/etc/resolvconf/resolv.conf.d/original"]
             actions.append(DeprovisionAction(fileutil.rm_files, files_to_del))
+
+
+class Ubuntu1804DeprovisionHandler(UbuntuDeprovisionHandler):
+    def __init__(self):
+        super(Ubuntu1804DeprovisionHandler, self).__init__()
+
+    def del_resolv(self, warnings, actions):
```
boumenot: Is there any reason to keep the prior behavior?

boumenot:
It looks like you have a lot of unnecessary commits in your history.

azurelinuxagent/pa/deprovision/ubuntu.py
hglkrijger: We still want the other actions implemented by the base handler.

boumenot: Sorry for being unclear.  We are warning about something we used to do that was bad.  If this is bad we shouldn't be doing it now or in previous versions.  I believe this is the intent of #855.  It seems odd to warn about something you previously did illegally especially considering it *breaks* the VM.

You could also argue that we will address #855 in a later version as a single PR.

hglkrijger: Changing established behavior in 1604 is highly risky, and I do not believe we should take the chance of breaking existing patterns. 1804 is a good opportunity to make that change though, and issuing a warning seems sensible if people have taken a dependency on that broken behavior, so they can change it. If you are arguing this should be an info message instead, sure.

#855 is specifically about cloud-init scripts, which is a sub-class of this problem. Aside from these two instances, I am not aware of any issues or asks in this area.

Rangi-th:
The fix mentioned in #1164 is not found; as expected in the WA Linux Agent version 2.2.31

The file has only the below content - `/usr/lib/python3/dist-packages/azurelinuxagent/pa/deprovision/ubuntu.py`

```
import os
from azurelinuxagent.pa.deprovision.default import DeprovisionHandler, \
    DeprovisionAction


class UbuntuDeprovisionHandler(DeprovisionHandler):
    def __init__(self):
        super(UbuntuDeprovisionHandler, self).__init__()

    def del_resolv(self, warnings, actions):
        if os.path.realpath(
                '/etc/resolv.conf') != '/run/resolvconf/resolv.conf':
            warnings.append("WARNING! /etc/resolv.conf will be deleted.")
            files_to_del = ["/etc/resolv.conf"]
            actions.append(DeprovisionAction(fileutil.rm_files, files_to_del))
        else:
            warnings.append("WARNING! /etc/resolvconf/resolv.conf.d/tail "
                            "and /etc/resolvconf/resolv.conf.d/original will "
                            "be deleted.")
            files_to_del = ["/etc/resolvconf/resolv.conf.d/tail",
                            "/etc/resolvconf/resolv.conf.d/original"]
            actions.append(DeprovisionAction(fileutil.rm_files, files_to_del))
```

PR #1174, opened 2018-05-15
ensure cleanup of HTTP responses

### LINKED ISSUE #1092: Memory Leak in "python -u /usr/sbin/waagent -run-exthandlers"

boumenot:
This is a more complete fix for the memory leak.  This PR ensures every HTTPResponse is properly closed.

Closes #1092 

>This is not ready for review.  It hasn't actually fixed the issue on the distros where we are experiencing leaks, and not having the fix does not cause the leak on most distros.

### DIFF
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -15,7 +15,7 @@
 # limitations under the License.
 #
 # Requires Python 2.6+ and Openssl 1.0+
-
+import traceback
 from datetime import datetime
 
 import json
@@ -36,6 +36,7 @@ from azurelinuxagent.common.protocol.hostplugin import HostPluginProtocol
 from azurelinuxagent.common.protocol.restapi import *
 from azurelinuxagent.common.utils.archive import StateFlusher
 from azurelinuxagent.common.utils.cryptutil import CryptUtil
+from azurelinuxagent.common.utils.restutil import HTTPResponseMaterialized
 from azurelinuxagent.common.utils.textutil import parse_doc, findall, find, \
     findtext, getattrib, gettext, remove_bom, get_bytes_from_pem, parse_json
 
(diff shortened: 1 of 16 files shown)


boumenot:
Do not merge, this fix is still being validated.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -537,24 +538,26 @@ def __init__(self, endpoint):
...
+                if restutil.request_failed(resp):
+                    msg = "[Wireserver Failed] URI {0} ".format(args[0])
+                    if resp is not None:
+                        msg += " [HTTP Failed] Status Code {0}".format(resp.status)
+                    raise ProtocolError(msg)
 
         # If the GoalState is stale, pass along the exception to the caller
-        except ResourceGoneError:
+        except ResourceGoneError as e:
+            print("WTF: {0}".format(traceback.format_exc()))
```
jasonzio: WTF == Wireserver Transfer Function?

azurelinuxagent/common/utils/restutil.py
```diff
@@ -150,6 +150,22 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         self._conn.close()
 
 
+class HTTPResponseMaterialized(object):
+    def __init__(self, resp):
+        self.status = resp.status
+        self.reason = resp.reason
+        self._content = resp.read()
```
jasonzio: Definitely not what I imagined. This context object actually results in the response object being destroyed immediately (after capturing a few, possibly expensive, attributes). Nothing "active" is done when __exit__ is called, because there's nothing to do.

My expectation was that this object would literally hold on to the response object for its entire lifetime, with __enter__() returning (a reference to) the wrapped request object which would be explicitly closed in the __exit__() method.

hglkrijger:
Is this ready for review?

azurelinuxagent/ga/monitor.py
```diff
@@ -201,6 +210,73 @@ def collect_and_send_events(self):
...
+
+        other = top_stats[limit:]
+        if other:
+            size = sum(stat.size for stat in other)
+            output.write("%s other: %.1f KiB\n" % (len(other), size / 1024))
+        total = sum(stat.size for stat in top_stats)
+        output.write("Total allocated size: %.1f KiB\n" % (total / 1024))
+
+        logger.info(output.getvalue())
+
```
jasonzio: 95% of these two methods is identical; please refactor to coalesce them. (The only difference I can see is the output.write inside the `for index, stat in...` loop.)

azurelinuxagent/common/protocol/wire.py
boumenot: 👍

boumenot:
This change is not needed.  We traced the issue to an interaction between Python (3.5.3) and OpenSSL (1.1) on the Debian 9 as opposed to something in the agent.

PR #1181, opened 2018-05-21
add ImageOrigin to telemetry

### LINKED ISSUE #1136: Telemetry Should Include an ImageOrigin Flag
This flag is an integer, and should report at least the following values.

 1. Custom
 1. Endorsed
 1. Marketplace

Ideally, these values should be sourced from the platform, but adding them to the agent is expedient and provides immediate value.

Please see https://docs.microsoft.com/en-us/azure/virtual-machines/linux/endorsed-distros for the definition of an endorsed image.

boumenot:
Closes #1136

### DIFF
--- a/azurelinuxagent/common/protocol/imds.py
+++ b/azurelinuxagent/common/protocol/imds.py
@@ -1,25 +1,169 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the Apache License, Version 2.0 (the "License");
 import json
+import re
 
 import azurelinuxagent.common.utils.restutil as restutil
-from azurelinuxagent.common.exception import HttpError, ProtocolError
+from azurelinuxagent.common.exception import HttpError
 from azurelinuxagent.common.future import ustr
+import azurelinuxagent.common.logger as logger
 from azurelinuxagent.common.protocol.restapi import DataContract, set_properties
+from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
 
 IMDS_ENDPOINT = '169.254.169.254'
 APIVERSION = '2017-12-01'
 BASE_URI = "http://{0}/metadata/instance/{1}?api-version={2}"
+
+IMDS_IMAGE_ORIGIN_UNKNOWN = 0
+IMDS_IMAGE_ORIGIN_CUSTOM = 1
+IMDS_IMAGE_ORIGIN_ENDORSED = 2
+IMDS_IMAGE_ORIGIN_PLATFORM = 3
 
 
 def get_imds_client():
     return ImdsClient()
 
 
+# A *slightly* future proof list of endorsed distros.
+#  -> e.g. I have predicted the future and said that 20.04-LTS will exist
+#     and is endored.
+#
+# See https://docs.microsoft.com/en-us/azure/virtual-machines/linux/endorsed-distros for
+# more details.
+#
+# This is not an exhaustive list. This is a best attempt to mark images as
+# endorsed or not.  Image publishers do not encode all of the requisite information
+# in their publisher, offer, sku, and version to definitively mark something as
+# endorsed or not.  This is not perfect, but it is approximately 98% perfect.
+ENDORSED_IMAGE_INFO_MATCHER_JSON = """{
+    "CANONICAL": {
+        "UBUNTUSERVER": {
+            "List": [
+                "14.04.0-LTS",
+                "14.04.1-LTS",
+                "14.04.2-LTS",
+                "14.04.3-LTS",
+                "14.04.4-LTS",
+                "14.04.5-LTS",
+                "14.04.6-LTS",
+                "14.04.7-LTS",
+                "14.04.8-LTS",
+                
+                "16.04-LTS",
+                "16.04.0-LTS",
+                "18.04-LTS",
+                "20.04-LTS",
+                "22.04-LTS"
+            ]
+        }
+    },
+    "COREOS": {
+        "COREOS": {
+            "STABLE": { "Minimum": "494.4.0" }
+        }
+    },
+    "CREDATIV": {
+        "DEBIAN": { "Minimum": "7" }
+    },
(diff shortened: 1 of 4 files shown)


azurelinuxagent/common/protocol/imds.py
```diff
@@ -1,22 +1,155 @@
...
+
+            if 'Match' in doci[key] and re.match(doci[key]['Match'], keys[0]):
+                return True
+
+            if 'Minimum' in doci[key]:
+                try:
+                    return FlexibleVersion(keys[0]) >= FlexibleVersion(doci[key]['Minimum'])
+                except ValueError:
+                    pass
+
```
jasonzio: Consider adding a "list" entry, whose value is a JSON list of strings; if keys[0] is in that list, return True. This truncates the search early in many cases, but lets you eliminate nearly all of your "Match": ".*" cases.

```
   "CANONICAL": {
        "UBUNTUSERVER": { 
            "List": [ "14.04.0-LTS", "14.04.1-LTS", "14.04.2-LTS", "14.04.3-LTS", etc ]
         }
    },
```

And...

```
    if 'List' in doci[key] and keys[0] in doci[key]['List']:
        return True
```

azurelinuxagent/common/protocol/imds.py
```diff
@@ -49,10 +182,34 @@ def __init__(self,
...
+            if self.publisher == "":
+                return IMDS_IMAGE_ORIGIN_CUSTOM
+
+            if ComputeInfo.__matcher.is_match(self.publisher, self.offer, self.sku, self.version):
+                return IMDS_IMAGE_ORIGIN_ENDORSED
+            else:
+                return IMDS_IMAGE_ORIGIN_PLATFORM
+
+        except Exception as e:
+            return IMDS_IMAGE_ORIGIN_UNKNOWN
```
jasonzio: If we catch an exception here, we should almost certainly log it, rather than swallowing it silently. Consider emitting telemetry.

tests/protocol/test_image_info_matcher.py
```diff
@@ -0,0 +1,105 @@
...
+        doc = '''{
+            "REDHAT": {
+                "RHEL": {
+                    "Minimum": "6.3"
+                }
+            }
+        }'''
+
+        test_subject = ImageInfoMatcher(doc)
+        self.assertFalse(test_subject.is_match("RedHat", "RHEL", "16.04-LTS", ""))
```
jasonzio: You also want to test a straightforward numerical comparison, i.e.
```
self.assertFalse(test_subject.is_match("RedHat", "RHEL", "6.04", ""))
```
Or is that supposed to succeed? Darned if I know.

tests/protocol/test_imds.py
```diff
@@ -3,7 +3,8 @@
 import json
 
 from azurelinuxagent.common.exception import HttpError
-from azurelinuxagent.common.protocol.imds import ComputeInfo, ImdsClient
+from azurelinuxagent.common.protocol.imds import ComputeInfo, ImdsClient, IMDS_IMAGE_ORIGIN_CUSTOM, \
+    IMDS_IMAGE_ORIGIN_ENDORSED, IMDS_IMAGE_ORIGIN_PLATFORM
```
jasonzio: Consider refactoring the azurelinuxagent.common.protocol.imds module so it can more simply be imported, i.e.
```
import azurelinuxagent.common.protocol.imds as imds
```

Then you would reference the objects of interest as `imds.ComputeInfo`, `imds.Client`, `imds.IMAGE_ORIGIN_CUSTOM` (or possible even `imds.ImageOrigin.custom` if you further refactored the image origin information).

tests/protocol/test_imds.py
```diff
@@ -98,3 +99,136 @@ def test_deserialize_ComputeInfo(self):
...
+
+        self.assertEqual(IMDS_IMAGE_ORIGIN_ENDORSED, self._setup_image_origin_assert("OpenLogic", "CentOS-HPC", "6.5", ""))
+        self.assertEqual(IMDS_IMAGE_ORIGIN_ENDORSED, self._setup_image_origin_assert("OpenLogic", "CentOS-HPC", "6.8", ""))
+        self.assertEqual(IMDS_IMAGE_ORIGIN_ENDORSED, self._setup_image_origin_assert("OpenLogic", "CentOS-HPC", "7.1", ""))
+        self.assertEqual(IMDS_IMAGE_ORIGIN_ENDORSED, self._setup_image_origin_assert("OpenLogic", "CentOS-HPC", "7.3", ""))
+        self.assertEqual(IMDS_IMAGE_ORIGIN_ENDORSED, self._setup_image_origin_assert("OpenLogic", "CentOS-HPC", "7.4", ""))
+
+        self.assertEqual(IMDS_IMAGE_ORIGIN_PLATFORM, self._setup_image_origin_assert("OpenLogic", "CentOS", "6.1", ""))
+        self.assertEqual(IMDS_IMAGE_ORIGIN_PLATFORM, self._setup_image_origin_assert("OpenLogic", "CentOS", "6.2", ""))
+
```
jasonzio: Even test cases can be table/data driven... :-)

azurelinuxagent/common/protocol/imds.py
boumenot: I cannot easily emit a telemetry event, but I can log a warning.  The plan is create a telemetry event for all warnings and error in vNext+1.

tests/protocol/test_image_info_matcher.py
boumenot: I added test cases to account for how FlexibleVersion works. We completely control the specification, so I am not concerned about this type of stuff creeping in.  If it does, there is test coverage to illustrate how it works.

tests/protocol/test_imds.py
boumenot: True.  I tend to prefer WET for test code, where WET means Write Explicit Tests.  It tends to be easier to read, and set a break point.

PR #1182, opened 2018-05-21
Telemetry updates

### LINKED ISSUE #1178: ExtensionProcessing emits Unnecessary Transient Events
The ExtensionProcessing telemetry event is emitted every time the agent fails to get goal state.  This is unnecessarily noisy, and does not actually indicate any issue with the agent.  Please use the `ErrorState` class just as was done for ReportStatus to make this a more meaningful value.

hglkrijger:
- fixes #1178 
- switches enduring report status failures to `ReportStatusExtended`

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -36,7 +36,7 @@ import azurelinuxagent.common.conf as conf
 import azurelinuxagent.common.logger as logger
 import azurelinuxagent.common.utils.fileutil as fileutil
 import azurelinuxagent.common.version as version
-from azurelinuxagent.common.errorstate import ErrorState, ERROR_STATE_DELTA
+from azurelinuxagent.common.errorstate import ErrorState, ERROR_STATE_DELTA_DEFAULT, ERROR_STATE_DELTA_INSTALL
 
 from azurelinuxagent.common.event import add_event, WALAEventOperation, elapsed_milliseconds, report_event
 from azurelinuxagent.common.exception import ExtensionError, ProtocolError
@@ -191,7 +191,7 @@ class ExtHandlersHandler(object):
         self.log_process = False
 
         self.report_status_error_state = ErrorState()
-        self.get_artifact_error_state = ErrorState(min_timedelta=datetime.timedelta(minutes=5))
+        self.get_artifact_error_state = ErrorState(min_timedelta=ERROR_STATE_DELTA_INSTALL)
 
     def run(self):
         self.ext_handlers, etag = None, None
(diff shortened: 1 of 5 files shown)


tests/ga/test_extension.py
```diff
@@ -541,16 +541,29 @@ def test_ext_handler_rollingupgrade(self, *args):
         self._assert_ext_status(protocol.report_ext_status, "success", 0)
 
     @patch('azurelinuxagent.ga.exthandlers.add_event')
-    def test_ext_handler_download_failure(self, mock_add_event, *args):
+    def test_ext_handler_download_failure_transient(self, mock_add_event, *args):
```
boumenot: Is there a corresponding test case for ReportStatusExtended?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -191,21 +191,28 @@ def __init__(self):
...
+            if self.get_artifact_error_state.is_triggered():
+                add_event(AGENT_NAME,
+                          version=CURRENT_VERSION,
+                          op=WALAEventOperation.GetArtifactExtended,
+                          is_success=False,
+                          message="Failed to get extension artifact for over "
+                                  "{0): {1}".format(self.get_artifact_error_state.min_timedelta, msg))
+                self.get_artifact_error_state.reset()
+            else:
+                logger.warn(msg)
```
boumenot: If we're going to retry, should this be verbose?  When retries have been exhausted we can log an error.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -191,21 +191,28 @@ def __init__(self):
         self.log_process = False
 
         self.report_status_error_state = ErrorState()
+        self.get_artifact_error_state = ErrorState(min_timedelta=datetime.timedelta(minutes=5))
```
boumenot: 5 should be a constant or the default for ErrorState.

tests/ga/test_extension.py
hglkrijger: Will add.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -191,21 +191,28 @@ def __init__(self):
...
 
     def run(self):
         self.ext_handlers, etag = None, None
         try:
             self.protocol = self.protocol_util.get_protocol()
             self.ext_handlers, etag = self.protocol.get_ext_handlers()
+            self.get_artifact_error_state.reset()
```
jasonzio: This belongs outside the try block.

azurelinuxagent/ga/exthandlers.py
hglkrijger: We do retry, but any failure is still abnormal. If we change to verbose, we would lose this detail, since we hardly ever get verbose logs. This exception handler is high up enough that I feel it's worth keeping the detail at the expense of potentially cluttering the logs.

azurelinuxagent/ga/exthandlers.py
hglkrijger: Sure, will update

PR #1184, opened 2018-05-21
do not raise an exception if get_first_if fails

### LINKED ISSUE #1130: DHCP regression
[This commit](https://github.com/Azure/WALinuxAgent/commit/9bb5d40ad80177b49e397d871df2faded2d6e48d) added an exception when there is no interface found, resulting in the following logs:

```
'2018/04/23 02:55:05.255749 ERROR Event: name=WALinuxAgent, op=UnhandledError, message=Traceback (most recent call last):\n', 
'2018/04/23 02:57:01.471314 ERROR Event: name=WALinuxAgent, op=UnhandledError, message=Traceback (most recent call last):\n', 
'2018/04/23 02:58:57.619919 ERROR Event: name=WALinuxAgent, op=UnhandledError, message=Traceback (most recent call last):\n', 
'2018/04/23 03:00:53.930831 ERROR Event: name=WALinuxAgent, op=UnhandledError, message=Traceback (most recent call last):\n', 
'2018/04/23 03:02:50.577667 ERROR Event: name=WALinuxAgent, op=UnhandledError, message=Traceback (most recent call last):\n', 
'2018/04/23 03:04:46.782791 ERROR Event: name=WALinuxAgent, op=UnhandledError, message=Traceback (most recent call last):\n', 
'2018/04/23 03:06:42.903693 ERROR Event: name=WALinuxAgent, op=UnhandledError, message=Traceback (most recent call last):\n', 
'2018/04/23 03:08:39.078856 ERROR Event: name=WALinuxAgent, op=UnhandledError, message=Traceback (most recent call last):\n', 
'2018/04/23 03:10:35.215515 ERROR Event: name=WALinuxAgent, op=UnhandledError, message=Traceback (most recent call last):\n', 
'2018/04/23 03:12:31.405739 ERROR Event: name=WALinuxAgent, op=UnhandledError, message=Traceback (most recent call last):\n'
```

As there is nothing to be done about this we should remove the exception.

hglkrijger:
- fixes #1130

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -662,7 +662,7 @@ class DefaultOSUtil(object):
 
     def get_mac_addr(self):
         """
-        Convienience function, returns mac addr bound to
+        Convenience function, returns mac addr bound to
         first non-loopback interface.
         """
         ifname = ''
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -665,9 +665,9 @@ class DefaultOSUtil(object):
         Convienience function, returns mac addr bound to
         first non-loopback interface.
         """
-        ifname=''
-        while len(ifname) < 2 :
-            ifname=self.get_first_if()[0]
+        ifname = ''
+        while len(ifname) < 2:
+            ifname = self.get_first_if()[0]
         addr = self.get_if_mac(ifname)
         return textutil.hexstr_to_bytearray(addr)
 
@@ -737,16 +737,15 @@ class DefaultOSUtil(object):
         if primary in ifaces:
             return primary, ifaces[primary]
 
-        logger.warn(('Primary interface {0} not found in ifconf list'), primary)
+        logger.warn('Primary interface {0} not found in ifconf list', primary)
         for iface_name in ifaces.keys():
             if not self.is_loopback(iface_name):
                 if not self.disable_route_warning:
                     logger.info("Choosing non-primary {0}".format(iface_name))
                 return iface_name, ifaces[iface_name]
 
-        msg = 'No non-loopback interface found in ifconf list'
-        logger.warn(msg)
-        raise Exception(msg)
+        logger.warn('No non-loopback interface found in ifconf list')
+        return '', ''
 
 
     def get_primary_interface(self):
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/osutil/default.py
```diff
@@ -665,9 +665,9 @@ def get_mac_addr(self):
         Convienience function, returns mac addr bound to
```
jasonzio: Might as well fix this typo, since you're here anyway. "Convienience" is misspelled.

azurelinuxagent/common/osutil/default.py
```diff
@@ -908,7 +907,10 @@ def is_missing_default_route(self):
         return True
 
     def get_if_name(self):
-        return self.get_first_if()[0]
+        ifname = ''
+        while len(ifname) < 2:
```
jasonzio: Why? Looks to me like all exit paths from `self.get_first_if()` return a 2-tuple.

azurelinuxagent/common/osutil/default.py
hglkrijger: Sure, fixed

azurelinuxagent/common/osutil/default.py
hglkrijger: Yes, but the callers for `get_if_name` do not all deal with `''` very well. Or did I misunderstand your concern?

jasonzio: I misread how ifname was being set, so my original comment is withdrawn.

I have a new comment, though: if there's no interface other than loopback, this code will run in a very tight loop. Clearly that scenario occurs, though, since there was strong complaint about throwing an exception in this case; were this a never-happens scenario, no one would have complained.

hglkrijger: This is true, and matches with the existing behavior in `get_mac_addr` which we hit occasionally too. The loop is somewhat controlled by reaching out to disk but if you prefer we could add a sleep. What would you suggest?

jasonzio: I'm just surprised this isn't a problem in practice. If you tell me it works out okay, I'm fine with that. Otherwise, I'd suggest exponential backoff and logging. (No point in trying to emit telemetry, since the essence of the problem is "I ain't got no network".) Hammering on the cpu in a tight loop might actually slow down whatever in-guest process will eventually provide a NIC.

hglkrijger: I am going to merge for the moment, and keep an eye on the bvts for anything unexpected. We can always add the backoff if necessary.

PR #1188, opened 2018-05-23
Clean up signal handling and sentinel file

hglkrijger:
- in a child process, when handling a `SIGTERM` from the default handler we should call `shutdown()`
- we should not emit a sentinel on normal startup because `SIGKILL` cannot be caught, and we consider this to be a clean shutdown; instead only emit a sentinel on an unhandled error
- fix naming of sentinel everywhere
- update tests

### DIFF
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -324,12 +324,7 @@ class UpdateHandler(object):
         sys.exit(0)
 
     def forward_signal(self, signum, frame):
-        # Note:
-        #  - At present, the handler is registered only for SIGTERM.
-        #    However, clean shutdown is both SIGTERM and SIGKILL.
-        #    A SIGKILL handler is not being registered at this time to
-        #    minimize perturbing the code.
-        if signum in (signal.SIGTERM, signal.SIGKILL):
+        if signum == signal.SIGTERM:
             self._shutdown()
 
         if self.child_process is None:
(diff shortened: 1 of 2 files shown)


azurelinuxagent/ga/update.py
```diff
@@ -382,7 +378,6 @@ def _emit_restart_event(self):
         except Exception:
             pass
 
-        self._set_sentinal(msg="Starting")
```
hglkrijger: Removing this is the primary fix here.

azurelinuxagent/ga/update.py
```diff
@@ -324,12 +324,7 @@ def run(self):
         sys.exit(0)
 
     def forward_signal(self, signum, frame):
-        # Note:
-        #  - At present, the handler is registered only for SIGTERM.
-        #    However, clean shutdown is both SIGTERM and SIGKILL.
-        #    A SIGKILL handler is not being registered at this time to
-        #    minimize perturbing the code.
-        if signum in (signal.SIGTERM, signal.SIGKILL):
+        if signum == signal.SIGTERM:
```
jasonzio: Dh  you want to catch SIGQUIT as well?

hglkrijger: I thought SIGQUIT was a keyboard signal?

hglkrijger:
All we really do in the signal handler is clean up the sentinel and child processes, but we also do a cleanup on restart. So I think the value is purely in knowing 'something other than SIGTERM killed me', which is not actionable or imo very useful. I am going to merge, but happy to discuss further.

PR #1189, opened 2018-05-23
fix unhandled errors

hglkrijger:
Took a look at the top unhandled errors which had a traceback, made some minor fixes.

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -1046,20 +1046,24 @@ class DefaultOSUtil(object):
         device = None
         path = "/sys/bus/vmbus/devices/"
         if os.path.exists(path):
-            for vmbus in os.listdir(path):
-                deviceid = fileutil.read_file(os.path.join(path, vmbus, "device_id"))
-                guid = deviceid.lstrip('{').split('-')
-                if guid[0] == g0 and guid[1] == "000" + ustr(port_id):
-                    for root, dirs, files in os.walk(path + vmbus):
-                        if root.endswith("/block"):
-                            device = dirs[0]
-                            break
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/osutil/default.py
```diff
@@ -1046,20 +1046,24 @@ def device_for_ide_port(self, port_id):
...
+                                device = dirs[0]
+                                break
+                            else:
+                                # older distros
+                                for d in dirs:
+                                    if ':' in d and "block" == d.split(':')[0]:
+                                        device = d.split(':')[1]
+                                        break
+                        break
+            except OSError:
```
boumenot: Any value in logging? It isn't clear to me what would happen that we are now ignoring.

hglkrijger: You're right - probably worth capturing this, thanks. Added.

azurelinuxagent/common/osutil/default.py
```diff
@@ -1046,20 +1046,24 @@ def device_for_ide_port(self, port_id):
...
-                                if ':' in d and "block" == d.split(':')[0]:
-                                    device = d.split(':')[1]
-                                    break
-                    break
+            try:
+                for vmbus in os.listdir(path):
+                    deviceid = fileutil.read_file(os.path.join(path, vmbus, "device_id"))
+                    guid = deviceid.lstrip('{').split('-')
+                    if guid[0] == g0 and guid[1] == "000" + ustr(port_id):
+                        for root, dirs, files in os.walk(path + vmbus):
```
jasonzio: If the deviceid is malformed, it's possible that guid may be a list with only one member, in which case referencing guid[1] will raise IndexError.

PR #1194, opened 2018-05-29
prefer HostGA for status updates

boumenot:


boumenot:
> CI is expected to fail.  I am checking the approach before fixing all of the unit test breakage.

@hglkrijger - does this match the approach we discussed?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -952,23 +952,35 @@ def upload_status_blob(self):
...
+                            return
+                    except HttpError:
+                        pass
 
+                    return
             except Exception as e:
-                # If the HostPlugin rejects the request,
-                # let the error continue, but set to use the HostPlugin
                 if isinstance(e, ResourceGoneError):
                     HostPluginProtocol.set_default_channel(True)
```
boumenot: Does it make sense to keep this? If so, should I incorporate its use into the above code.  This is a wonky.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -952,25 +952,36 @@ def upload_status_blob(self):
...
+                            self.report_status_event("Exception uploading status blob (HostPlugin): {0}", ustr(e))
 
-                    host = self.get_host_plugin()
-                    host.put_vm_status(self.status_blob,
-                                    ext_conf.status_upload_blob,
-                                    ext_conf.status_upload_blob_type)
-                    HostPluginProtocol.set_default_channel(True)
-                    return
+                    try:
+                        if self.status_blob.upload(blob_uri):
```
hglkrijger: I would like to know how often we have to fallback to direct - can we add telemetry here?

boumenot: Fixed.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -952,25 +955,43 @@ def upload_status_blob(self):
...
+                                log_event=False)
+                            return
+                    except HttpError:
+                        pass
 
+                    return
             except Exception as e:
-                # If the HostPlugin rejects the request,
-                # let the error continue, but set to use the HostPlugin
                 if isinstance(e, ResourceGoneError):
```
hglkrijger: Is this block still needed?

boumenot: We can still reach there is HostGA falls through, and direct throws this exception.  Correct?

tests/tools.py
```diff
@@ -28,7 +28,7 @@
 
 import time
 
-import azurelinuxagent.common.event as event
```
boumenot: This caused a bug.  We import `as event` and then pollute all of the test files that import this file, which is basically all of them.  We have a naked event import, which then collides with other files, and we see failures in TravisCI as a result. The fix is not to import `as event`, and only import the relevant files.

boumenot: I spoke too soon.  There's a circular dependency.  I am trying to see if I need to unwind it now, or if I can work around it.  It manifests when running particular test cases.

boumenot: I moved the import to exactly where it is needed so it isn't sucked in unnecessarily. This needs a better fix.  I'll open an issue.

PR #1195, opened 2018-05-31
Cgroups

jasonzio:
Needs to have the Memory metrics fleshed out, and needs to have the table which maps from extension name to dependent services filled in, but otherwise this has passed smoke tests and is ready to be reviewed.


azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+    def set_delete_empty_groups(setting):
+        CGroupsTelemetry.delete_empty_groups = setting
+
+    @staticmethod
+    def collect_all_tracked():
+        """
+        Return a dictionary mapping from the name of a tracked cgroup to the list of collected metrics for that cgroup.
+
+        :returns: Dictionary of collected metrics, by cgroup
+        :rtype: dict(str: [(str, str, float)])
```
boumenot: You have a great comment below about what this tuple means.  I was able to figure out from previous code, but it would be nice to see "(metric group name, metric name, metric value)".

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+                CGroupsTelemetry.stop_tracking(name)
+            else:
+                names.append("[{0}]".format(name))
+
+        now_tracking = " ".join(names)
+        if now_tracking != CGroupsTelemetry.was_tracked:
+            if len(now_tracking):
+                logger.info("After updating cgroup telemetry, tracking {0}".format(now_tracking))
+            else:
+                logger.warn("After updating cgroup telemetry, tracking no cgroups.")
```
boumenot: Is this worthy of a warning? If the customer saw it, should they be concerned?

jasonzio:
Yeah, looks like I didn't adjust my tests to match the changes I made for systemd. Should be quick fixes.

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+        :rtype: Bool
+        """
+        if CGroups._use_systemd is None:
+            hierarchy = HIERARCHIES[0]
+            path = CGroups.get_my_cgroup_folder(hierarchy)
+            CGroups._use_systemd = path.startswith(CGroups.construct_systemd_path_for_hierarchy(hierarchy))
+        return CGroups._use_systemd
+
+    @staticmethod
+    def _try_mkdir(path):
```
boumenot: There is a method to create a directory in fileutil.py. This code handles more cases.  I leave it to you to decide if they should be combined or not.

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+        if 'cpu' in self.cgroups:
+            total_units = float(self.get_parameter('cpu', 'cpu.cfs_period_us'))
+            limit_units = self._format_cpu_value(limit) * total_units
+            cpu_shares_file = self._get_cgroup_file('cpu', 'cpu.cfs_quota_us')
+            fileutil.write_file(cpu_shares_file, "{0}\n".format(limit_units))
+        else:
+            raise CGroupsException("CPU hierarchy not available in this cgroup")
+
+    @staticmethod
+    def get_num_cores():
```
boumenot: We calculate the number of CPUs in osutil.  Any reason not to use that?

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+        there's no point in returning the CGroups object itself; the goal is to move the child process into the
+        cgroup before the new code even starts running.
+
+        :param str name: Short name of extension, suitable for naming directories in the filesystem
+        :param int pid: Process id of extension to be added to the cgroup
+        """
+        if not CGroups.enabled():
+            return
+        if name == AGENT_NAME:
+            logger.warn('Extension cgroup name cannot match agent cgroup name({0})'.format(AGENT_NAME))
```
boumenot: If it cannot match should you return? The method still continues.

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+
+        try:
+            logger.info("Move process {0} into cgroups for extension {1}".format(pid, name))
```
boumenot: Some (most?) of these info statements feel verbose to me.  Have you reviewed a BVT run to see how verbose they are (or are not)?

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+import re
+import time
+
+from azurelinuxagent.common import logger
+from azurelinuxagent.common.utils import fileutil
+from azurelinuxagent.common.version import AGENT_NAME, CURRENT_VERSION
+# from azurelinuxagent.common.protocol.restapi import ExtHandler
+
+BASE_CGROUPS = '/sys/fs/cgroup'
+HIERARCHIES = ['cpu', 'memory']
```
boumenot: Reading the code it was not intuitive what HIERARCHIES meant.  I would suggest METRIC_HIERARCHIES, or something similar.

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+MEMORY_DEFAULT = -1
```
boumenot:
```python
MEMORY_DEFAULT_LIMIT = -1
CPU_DEFAULT_LIMIT = 0.25  # 25% of one core
```

Based on your code, I think CPU_DEFAULT should be 25, right?  This value is high IMO.  The value on Windows is 5% or 10%

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+        Compute the number of USER_HZ units of time have elapsed in all categories, across all cores, since boot.
+
+        :return: int
+        """
+        system_cpu = 0
+        proc_stat = Cpu.get_proc_stat()
+        if proc_stat is not None:
+            for line in proc_stat.splitlines():
+                if re_all_CPUs.match(line):
+                    system_cpu = sum(int(i) for i in line.split()[1:7])
```
boumenot: You use the output of /proc/stat a couple of times. Considering making this a real object so the caller does not have to parse it.  It will make it easier for readers to identify what you are actually sum()'ing.

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+        service_name = "{0}.service".format(name)
+        if not CGroupsTelemetry.is_tracked(service_name):
+            cgroup = CGroups.for_systemd_service(service_name)
+            tracker = CGroupsTelemetry(service_name, cgroup=cgroup)
+            CGroupsTelemetry._tracked[service_name] = tracker
+
+    @staticmethod
+    def track_extension(name, cgroup=None):
+        """
+        Create all required CGroups to track all metrics for an extension and its associated services.
```
boumenot: ...if the `name` is not already tracked.

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+
+        names = []
+        for name in CGroupsTelemetry._tracked.keys():
+            if CGroupsTelemetry.delete_empty_groups and name in not_enabled_extensions:
+                CGroupsTelemetry.stop_tracking(name)
+            else:
+                names.append("[{0}]".format(name))
+
+        now_tracking = " ".join(names)
+        if now_tracking != CGroupsTelemetry.was_tracked:
```
boumenot: I don't like this comparison.  You **appear** to be relying on a string to determine if things have changed or not.  I don't think that is a robust comparison.  Why isn't this a set compare instead?

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+        self.name = name
+        if cgroup is None:
+            cgroup = CGroups.for_extension(name)
+        self.cgroup = cgroup
+        self.cpu_count = CGroups.get_num_cores()
+        self.current_wall_time = time.time()
+        self.previous_wall_time = 0
+
+        self.data = {}
+        for hierarchy in CGroupsTelemetry.hierarchies():
```
boumenot: I would prefer metric_hierarchies vs. hierarchies.

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+                        raise CGroupsException(
+                            "Create directory for cgroup {0}: normal file already exists with that name".format(path)
+                        )
+                    else:
+                        pass    # There was a race to create the directory, but it's there now, and that's fine
+                elif e.errno == errno.EACCES:
+                    # This is unexpected, as the agent runs as root
+                    raise CGroupsException("Create directory for cgroup {0}: permission denied".format(path))
+                else:
+                    raise OSError(e)
```
boumenot: raise?

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+        """
+        if not self.enabled():
+            return
+
+        if self.is_wrapper_cgroup:
+            raise CGroupsException("Cannot add a process to the Agents+Extensions wrapper cgroup")
+
+        try:
+            # determine if pid exists by sending signal 0 to it
+            os.kill(pid, 0)
```
boumenot: This should just be a method in osutil.  This will avoid the need for a comment in the first place, and is useful in other parts of the code.  You can do it now, or open an issue to track.

```python
  def is_process_alive(pid):
      try:
          os.kill(pid, 0)
          return True
      except OSError:
          return False
```

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+                else:
+                    raise OSError(e)
+
+    @staticmethod
+    def construct_custom_path_for_hierarchy(hierarchy):
+        return os.path.join(BASE_CGROUPS, hierarchy, AGENT_NAME)
+
+    @staticmethod
+    def construct_systemd_path_for_hierarchy(hierarchy):
+        return os.path.join(BASE_CGROUPS, hierarchy, 'system.slice')
```
boumenot: Should system.slice be a value in the OS util class hierarchy?  Is is particular to a distro flavor, or is it always the same for systemd?

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+        if self.is_wrapper_cgroup:
+            raise CGroupsException("Cannot add a process to the Agents+Extensions wrapper cgroup")
+
+        try:
+            # determine if pid exists by sending signal 0 to it
+            os.kill(pid, 0)
+        except OSError:
+            raise CGroupsException('PID {0} does not exist'.format(pid))
+        for hierarchy, cgroup in self.cgroups.items():
+            tasks_file = self._get_cgroup_file(hierarchy, 'cgroup.procs')
```
boumenot: What is this file used for?  Debug?

azurelinuxagent/common/event.py
```diff
@@ -301,6 +306,33 @@ def add_log_event(self, level, message):
         except EventError:
             pass
 
+    def add_metric(self, category, counter, instance, value, log_it=False):
```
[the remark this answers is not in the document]
jasonzio: OK

azurelinuxagent/common/event.py
```diff
@@ -327,10 +359,31 @@ def report_event(op, is_success=True, message=''):
...
+    :param str category: The category of the metric (cpu, memory, etc)
+    :param str counter: The name of the metric ("%idle", etc)
+    :param str instance: For instanced metrics, the identifier of the instance. E.g. a disk drive name, a cpu core#
+    :param     value: The value of the metric
+    :param bool log_it: If True, log the metric in the agent log as well
+    :param EventLogger reporter: The EventLogger instance to which metric events should be sent
+    """
+    if reporter.event_dir is None:
+        from azurelinuxagent.common.version import AGENT_NAME
+        logger.warn("Cannot add metric event -- Event reporter is not initialized.")
```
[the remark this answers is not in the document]
jasonzio: OK

azurelinuxagent/common/osutil/ubuntu.py
```diff
@@ -46,6 +49,34 @@ def restore_rules_files(self, rules_files=""):
...
+                       option="-t tmpfs",
+                       chk_err=False)
+
+            fileutil.mkdir('/sys/fs/cgroup/cpu,cpuacct')
+            self.mount(device='cpu,cpuacct',
+                       mount_point='/sys/fs/cgroup/cpu,cpuacct/',
+                       option="-t cgroup -o cpu,cpuacct",
+                       chk_err=False)
+
+            if not os.path.exists('/sys/fs/cgroup/cpu'):
```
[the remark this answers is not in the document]
jasonzio: I'm not sure there's any guarantee in general that the cgroup hierarchy has been created; in theory, there may be a supported distro that doesn't have all of this setup beforehand. Might be worth adding a DCR test case that reports a failure if the hierarchy doesn't exist; if no supported distro fails, then all of this code can be removed.

azurelinuxagent/common/version.py
```diff
@@ -113,7 +113,7 @@ def get_distro():
 
 AGENT_NAME = "WALinuxAgent"
 AGENT_LONG_NAME = "Azure Linux Agent"
-AGENT_VERSION = '2.2.25.1'
+AGENT_VERSION = '2.2.26.1'
```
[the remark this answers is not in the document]
jasonzio: OK. This shouldn't have leaked into my check-in. (Besides, won't it actually be 2.2.27?)

azurelinuxagent/ga/monitor.py
```diff
@@ -130,32 +131,32 @@ def init_sysinfo(self):
...
+                                                    vm_info.roleInstanceName))
             self.sysinfo.append(TelemetryEventParam("ContainerId",
-                                                    vminfo.containerId))
+                                                    vm_info.containerId))
         except ProtocolError as e:
             logger.warn("Failed to get system info: {0}", e)
 
         try:
-            vminfo = self.imds_client.get_compute()
+            vm_info = self.imds_client.get_compute()
```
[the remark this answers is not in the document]
jasonzio: If I had to change a file for cgroup support, I also did my best to eliminate all PyCharm code inspection warnings while I was in there. I'll take your advice for the future.

azurelinuxagent/ga/monitor.py
```diff
@@ -204,54 +205,96 @@ def collect_and_send_events(self):
...
+                                                        protocol.endpoint)
+
+                    msg = "{0};{1};{2};{3}".format(
+                        incarnation, counter, heartbeat_id, dropped_packets)
+
                     add_event(
                         name=AGENT_NAME,
                         version=CURRENT_VERSION,
-                        op=WALAEventOperation.HttpErrors,
+                        op=WALAEventOperation.HeartBeat,
```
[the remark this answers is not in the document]
jasonzio: I thought I had a good reason for this, but apparently not. I'll revert this.

azurelinuxagent/ga/monitor.py
```diff
@@ -204,54 +205,96 @@ def collect_and_send_events(self):
...
+                # performance counters
+                if datetime.datetime.utcnow() >= (last_collection + collection_period):
+                    last_collection = datetime.datetime.utcnow()
+                    for cgroup_name, metrics in CGroupsTelemetry.collect_all_tracked().items():
+                        for metric_group, metric_name, value in metrics:
+                            if value > 0:
+                                report_metric(metric_group, metric_name, cgroup_name, value)
+            except Exception as e:
+                logger.warn("Failed to collect performance metrics: {0}", e)
+                logger.warn(traceback.format_exc())
```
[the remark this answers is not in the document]
jasonzio: Oversight. OK.

tests/common/test_cgroups.py
```diff
@@ -0,0 +1,94 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Requires Python 2.4+ and Openssl 1.0+
+#
+
+from __future__ import print_function
+
+import mock
```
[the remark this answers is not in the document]
jasonzio: OK

tests/common/test_cgroups.py
```diff
@@ -0,0 +1,94 @@
...
+class TestCGroups(AgentTestCase):
+    @patch("os.mkdir")
+    @patch("azurelinuxagent.common.cgroups.CGroups.create_user_cgroups")
+    def test_cgroups_telemetry_inplace(self, _, __):
+        """
+        Test raw measures and basic statistics for the cgroup in which this process is currently running.
+        """
+        cpu_hierarchy_id = CGroups.get_hierarchy_id('cpu')
+        self.assertTrue(cpu_hierarchy_id != '')
+        in_root_cgroup = True
```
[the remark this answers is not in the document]
jasonzio: systemd leaves nothing in the "root" cgroup; if you're not using systemd, any newly-created process is, by default, in the root cgroup.

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+
+import errno
+import os
+import re
+import time
+
+from azurelinuxagent.common import logger
+from azurelinuxagent.common.utils import fileutil
+from azurelinuxagent.common.version import AGENT_NAME, CURRENT_VERSION
+# from azurelinuxagent.common.protocol.restapi import ExtHandler
```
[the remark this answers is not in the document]
jasonzio: OK

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+class CGroupsException(Exception):
+
+    def __init__(self, msg):
+
+        self.msg = msg
+        if CGroups.enabled():
+            pid = os.getpid()
+            logger.verbose("[{1}] Disabling cgroup support: {0}"
+                           .format(msg, pid))
+            CGroups.disable()
```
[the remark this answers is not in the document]
jasonzio: Does not throw.

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+            results = fileutil.read_file('/proc/stat')
+        except (OSError, IOError) as ex:
+            logger.warn("Couldn't read /proc/stat: {0}".format(ex.strerror))
+
+        return results
+
+    @staticmethod
+    def get_current_system_cpu():
+        """
+        Compute the number of USER_HZ units of time have elapsed in all categories, across all cores, since boot.
```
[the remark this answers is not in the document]
jasonzio: OK

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+BASE_CGROUPS = '/sys/fs/cgroup'
+HIERARCHIES = ['cpu', 'memory']
+MEMORY_DEFAULT = -1
+CPU_DEFAULT = 0.25
+re_all_CPUs = re.compile('^cpu .*')
+re_user_system_times = re.compile('user (\d+)\nsystem (\d+)\n')
+
+related_services = {
+    "Microsoft.OSTCExtensions.LinuxDiagnostic":    ["omid", "omsagent-LAD", "mdsd-lde"],
+    "Microsoft.Azure.Diagnostics.LinuxDiagnostic": ["omid", "omsagent-LAD", "mdsd-lde"],
```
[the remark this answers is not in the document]
jasonzio: We need this info for all extensions; OMS is merely one of them. I am told many extensions do install services.

azurelinuxagent/common/cgroups.py
jasonzio: I would be concerned, since the "agent" cgroup should always be tracked.

azurelinuxagent/common/cgroups.py
jasonzio: Good catch.

azurelinuxagent/common/cgroups.py
jasonzio: There's no other historical record of what processes were associated with any particular cgroup. Any extension developer would need to see this information if investigating "Why did some of my resource consumption not get tracked".

azurelinuxagent/common/cgroups.py
jasonzio: OK.

azurelinuxagent/common/cgroups.py
jasonzio: Constant for systemd.

azurelinuxagent/ga/monitor.py
```diff
@@ -204,54 +205,96 @@ def collect_and_send_events(self):
...
+        protocol = self.protocol_util.get_protocol()
+
+        # heartbeat
         period = datetime.timedelta(minutes=30)
-        protocol = self.protocol_util.get_protocol()        
         last_heartbeat = datetime.datetime.utcnow() - period
 
+        # performance counters
+
+        # Track metrics for the roll-up cgroup and for the agent cgroup
```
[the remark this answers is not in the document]
jasonzio: I'll consider it.

azurelinuxagent/ga/monitor.py
```diff
@@ -204,54 +205,96 @@ def collect_and_send_events(self):
...
+                            is_success=True,
+                            message=msg,
+                            log_event=False)
+            except Exception as e:
+                logger.warn("Failed to send heartbeat: {0} [{1}]".format(e, traceback.format_exc()))
+
+            try:
+                # performance counters
+                if datetime.datetime.utcnow() >= (last_collection + collection_period):
+                    last_collection = datetime.datetime.utcnow()
```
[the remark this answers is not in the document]
jasonzio: Open an issue, please. I have a good idea how to do this, but don't want to tackle this now.

tests/common/test_event.py
```diff
@@ -217,3 +217,11 @@ def test_save_event_cleanup(self):
...
         utc_start = datetime.utcnow() + timedelta(days=1)
         self.assertEqual(0, elapsed_milliseconds(utc_start))
+
+    @patch('azurelinuxagent.common.event.EventLogger.save_event')
+    def test_report_metric(self, mock_event):
+        event.report_metric("cpu", "%idle", "_total", 10.0)
+        self.assertEqual(1, mock_event.call_count)
+        event_json = mock_event.call_args[0][0]
+        self.assertIn("69B669B9-4AF8-4C50-BDC4-6006FA76E975", event_json)
+        self.assertIn("%idle", event_json)
```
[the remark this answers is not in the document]
jasonzio: OK. Round-trip back into dictionary form via json.loads(), then check for specific contents.

azurelinuxagent/common/cgroups.py
jasonzio: OK

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,698 @@
...
+    def collect_all_tracked():
+        """
+        Return a dictionary mapping from the name of a tracked cgroup to the list of collected metrics for that cgroup.
+
+        :returns: Dictionary of collected metrics, by cgroup
+        :rtype: dict(str: [(str, str, float)])
+        """
+        results = {}
+        for cgroup_name, collector in CGroupsTelemetry._tracked.items():
+            cgroup_name = cgroup_name if cgroup_name else "Agent+Extensions"
```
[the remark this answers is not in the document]
jasonzio: OK

azurelinuxagent/common/cgroups.py
jasonzio: They're distinct. The version here changes exception types to make them more focused to my purpose (making cgroup directories).

azurelinuxagent/common/cgroups.py
jasonzio: When adding an extension, this generates two events for the new extension (once for invoking with -install and once for -enable) plus one for each existing extension (the -enable that every active extension gets upon moving to a new goal state). This is roughly a 10%-20% increase in the amount of log spam every time there's a new goal state. In the steady state (when no goal state changes occur), no net new log entries are created.

azurelinuxagent/common/cgroups.py
jasonzio: The cgroup documentation calls these "hierarchies", but I take your intent.

azurelinuxagent/common/cgroups.py
jasonzio: I added sample content to the definition of the regex. I'd like to defer refactoring use of /proc/stat, though; there's a lot data in that file, and creating an object reflecting all of it is not cheap.

azurelinuxagent/common/cgroups.py
jasonzio: OK

azurelinuxagent/common/cgroups.py
jasonzio: Significantly refactored, switched to using a set.

azurelinuxagent/common/cgroups.py
jasonzio: OK.

azurelinuxagent/common/cgroups.py
jasonzio: Correct.

azurelinuxagent/common/cgroups.py
jasonzio: There already is check_pid_alive(), but it's a worse implementation (calls os.path.isdir('/proc/<pid>')). I'll fix that and use it.

jasonzio:
Cancelling this PR and resubmitting after rebasing.

PR #1196, opened 2018-05-31
Health store integration

hglkrijger:
The intent of this work is to add health signals for the host channel, which we can use to monitor PF rollouts. The following signals are in place:

#### GuestAgentPluginHeartbeat
- monitors the host plugin process is alive by calling `/health`
- check every 1m, send every 1m, failure > 5m unhealthy
```
{
  "Version": "1.0",
  "Observations": [
    {
      "IsHealthy": true,
      "Description": "",
      "Value": "",
      "ObservationName": "GuestAgentPluginHeartbeat"
    }
  ],
  "Api": "reporttargethealth",
  "Source": "WALinuxAgent"
}
```
#### GuestAgentPluginVersions
- monitors the host plugin channel is available by calling `/versions`
- determine health on usage (process start), send immediately, no failure rollup
```
{
  "Version": "1.0",
  "Observations": [
    {
      "Description": "",
      "Value": "",
      "IsHealthy": true,
      "ObservationName": "GuestAgentPluginVersions"
    }
  ],
  "Api": "reporttargethealth",
  "Source": "WALinuxAgent"
}
```
#### GuestAgentPluginArtifact
- monitors the `/extensionArtifact` api
- failure is determined by status codes `5xx`, except `502`, which is an upstream failure
- determine health on usage (extension/manifest/agent download), send every 1m, failure > 5m unhealthy
- we fetch for the purpose of extension handling (WireClient) and agent updates (GuestAgent), and may want to distinguish between these signals, so the source is listed in the `Description` field
- WireClient:
```
{
  "Version": "1.0",
  "Observations": [
    {
      "IsHealthy": true,
      "Description": "WireClient",
      "Value": "",
      "ObservationName": "GuestAgentPluginArtifact"
    }
  ],
  "Api": "reporttargethealth",
  "Source": "WALinuxAgent"
}
```
- GuestAgent:
```
{
  "Version": "1.0",
  "Observations": [
    {
      "IsHealthy": true,
      "Description": "GuestAgent",
      "Value": "",
      "ObservationName": "GuestAgentPluginArtifact"
    }
  ],
  "Api": "reporttargethealth",
  "Source": "WALinuxAgent"
}
```
#### GuestAgentPluginStatus
- monitors the `/status` api
- failure is determined by status codes `5xx`, except `502`, which is an upstream failure
- check every goal_state_interval (3s), send every 1m, failure > 5m unhealthy
```
{
  "Api": "reporttargethealth",
  "Observations": [
    {
      "Value": "",
      "ObservationName": "GuestAgentPluginStatus",
      "Description": "",
      "IsHealthy": true
    }
  ],
  "Version": "1.0",
  "Source": "WALinuxAgent"
}
```

Observations are reported to `http://{wireserver_ip}:80/HealthService`

Additional changes
- removed 400s from the `ResourceGoneError`; this is legacy host plugin behavior which has since been fixed
- switched extension package downloads from the base handler to client `fetch`, to simplify number of changes
- updated the logging prefix to `ExtHandler`, for brevity
- removed enable events from the logs, for clarity
- updated tests accordingly

### DIFF
--- a/azurelinuxagent/common/protocol/healthservice.py
+++ b/azurelinuxagent/common/protocol/healthservice.py
@@ -48,7 +48,7 @@ class HealthService(object):
     API = 'reporttargethealth'
     VERSION = "1.0"
     OBSERVER_NAME = 'WALinuxAgent'
-    HOST_PLUGIN_HEARTBEAT_OBSERVATION_NAME = 'HostPluginHeartbeat'
+    HOST_PLUGIN_HEARTBEAT_OBSERVATION_NAME = 'GuestAgentPluginHeartbeat'
 
     def __init__(self, endpoint):
         self.endpoint = HealthService.ENDPOINT.format(endpoint)
--- a/azurelinuxagent/common/protocol/healthservice.py
+++ b/azurelinuxagent/common/protocol/healthservice.py
@@ -64,6 +64,7 @@ class HealthService(object):
     HOST_PLUGIN_STATUS_OBSERVATION_NAME = 'GuestAgentPluginStatus'
     HOST_PLUGIN_VERSIONS_OBSERVATION_NAME = 'GuestAgentPluginVersions'
     HOST_PLUGIN_ARTIFACT_OBSERVATION_NAME = 'GuestAgentPluginArtifact'
+    MAX_OBSERVATIONS = 10
 
     def __init__(self, endpoint):
         self.endpoint = HealthService.ENDPOINT.format(endpoint)
@@ -87,8 +88,8 @@ class HealthService(object):
         Reports a signal for /health
         :param is_healthy: whether the call succeeded
         """
-        self.observations.append(Observation(name=HealthService.HOST_PLUGIN_HEARTBEAT_OBSERVATION_NAME,
-                                             is_healthy=is_healthy))
+        self._observe(name=HealthService.HOST_PLUGIN_HEARTBEAT_OBSERVATION_NAME,
+                      is_healthy=is_healthy)
         self._report()
 
     def report_host_plugin_versions(self, is_healthy, response):
@@ -97,9 +98,9 @@ class HealthService(object):
         :param is_healthy: whether the api call succeeded
         :param response: debugging information for failures
         """
-        self.observations.append(Observation(name=HealthService.HOST_PLUGIN_VERSIONS_OBSERVATION_NAME,
-                                             is_healthy=is_healthy,
-                                             value=response))
+        self._observe(name=HealthService.HOST_PLUGIN_VERSIONS_OBSERVATION_NAME,
+                      is_healthy=is_healthy,
+                      value=response)
         self._report()
 
     def report_host_plugin_extension_artifact(self, is_healthy, source, response):
@@ -110,10 +111,10 @@ class HealthService(object):
         :param response: debugging information for failures
         :return:
         """
-        self.observations.append(Observation(name=HealthService.HOST_PLUGIN_ARTIFACT_OBSERVATION_NAME,
-                                             is_healthy=is_healthy,
-                                             description=source,
-                                             value=response))
+        self._observe(name=HealthService.HOST_PLUGIN_ARTIFACT_OBSERVATION_NAME,
+                      is_healthy=is_healthy,
+                      description=source,
+                      value=response)
         self._report()
 
     def report_host_plugin_status(self, is_healthy, response):
@@ -123,10 +124,19 @@ class HealthService(object):
         :param response: debugging information for failures
         :return:
         """
-        self.observations.append(Observation(name=HealthService.HOST_PLUGIN_STATUS_OBSERVATION_NAME,
-                                             is_healthy=is_healthy,
-                                             value=response))
+        self._observe(name=HealthService.HOST_PLUGIN_STATUS_OBSERVATION_NAME,
+                      is_healthy=is_healthy,
+                      value=response)
         self._report()
+
+    def _observe(self, name, is_healthy, value='', description=''):
+        # ensure we keep the list size within bounds
+        if len(self.observations) >= HealthService.MAX_OBSERVATIONS:
+            del self.observations[:HealthService.MAX_OBSERVATIONS-1]
+        self.observations.append(Observation(name=name,
+                                             is_healthy=is_healthy,
+                                             value=value,
+                                             description=description))
 
     def _report(self):
         logger.verbose('HealthService: report observations')
@@ -136,6 +146,5 @@ class HealthService(object):
         except HttpError as e:
             logger.warn("HealthService: could not report observations: {0}", ustr(e))
         finally:
-            # TODO: add safety boundaries
             # these signals are not timestamped, so there is no value in persisting data
             del self.observations[:]
--- a/azurelinuxagent/common/protocol/healthservice.py
+++ b/azurelinuxagent/common/protocol/healthservice.py
@@ -20,6 +20,7 @@
 import json
 
 from azurelinuxagent.common import logger
+from azurelinuxagent.common.dhcp import KNOWN_WIRESERVER_IP
 from azurelinuxagent.common.exception import HttpError
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.utils import restutil
@@ -44,6 +45,7 @@ class Observation(object):
 
 class HealthService(object):
 
+    VNET_ENDPOINT = '169.254.169.254'
     ENDPOINT = 'http://{0}:80/HealthService'
     API = 'reporttargethealth'
     OBSERVER_NAME = 'WALinuxAgent'
@@ -51,7 +53,8 @@ class HealthService(object):
     HOST_PLUGIN_HEARTBEAT_OBSERVATION_NAME = 'HostPluginHeartbeat'
 
     def __init__(self, endpoint):
-        self.endpoint = HealthService.ENDPOINT.format(endpoint)
+        self.endpoint = HealthService.ENDPOINT.format(endpoint if endpoint != KNOWN_WIRESERVER_IP
+                                                      else HealthService.VNET_ENDPOINT)
         self.api = HealthService.API
         self.version = HealthService.VERSION
         self.source = HealthService.OBSERVER_NAME
--- a/azurelinuxagent/common/protocol/healthservice.py
+++ b/azurelinuxagent/common/protocol/healthservice.py
@@ -79,7 +79,7 @@ class HealthService(object):
             # TODO: remove
             logger.info('Report observation to {0}: {1}', self.endpoint, self.as_json)
 
-            restutil.http_post(self.endpoint, self.as_json)
+            restutil.http_post(self.endpoint, self.as_json, headers={'Content-Type': 'application/json'})
             del self.observations[:]
         except HttpError as e:
             logger.warn("HealthService could not report observations: {0}", ustr(e))
--- a/azurelinuxagent/common/protocol/healthservice.py
+++ b/azurelinuxagent/common/protocol/healthservice.py
@@ -27,6 +27,18 @@ from azurelinuxagent.common.utils import restutil
 
 class Observation(object):
     def __init__(self, name, is_healthy, description='', value=''):
+        if name is None:
+            raise ValueError("Observation name must be provided")
+
+        if is_healthy is None:
+            raise ValueError("Observation health must be provided")
+
+        if value is None:
+            value = ''
+
+        if description is None:
+            description = ''
+
         self.name = name
         self.is_healthy = is_healthy
         self.description = description
@@ -35,10 +47,10 @@ class Observation(object):
     @property
     def as_obj(self):
         return {
-            "ObservationName": self.name,
+            "ObservationName": self.name[:64],
             "IsHealthy": self.is_healthy,
-            "Description": self.description,
-            "Value": self.value
+            "Description": self.description[:128],
+            "Value": self.value[:128]
         }
 
 
--- a/azurelinuxagent/common/protocol/healthservice.py
+++ b/azurelinuxagent/common/protocol/healthservice.py
@@ -89,7 +89,7 @@ class HealthService(object):
         """
         self.observations.append(Observation(name=HealthService.HOST_PLUGIN_HEARTBEAT_OBSERVATION_NAME,
                                              is_healthy=is_healthy))
-        self.report()
+        self._report()
 
     def report_host_plugin_versions(self, is_healthy, response):
         """
@@ -100,7 +100,7 @@ class HealthService(object):
         self.observations.append(Observation(name=HealthService.HOST_PLUGIN_VERSIONS_OBSERVATION_NAME,
                                              is_healthy=is_healthy,
                                              value=response))
-        self.report()
+        self._report()
 
     def report_host_plugin_extension_artifact(self, is_healthy, source, response):
         """
@@ -114,7 +114,7 @@ class HealthService(object):
                                              is_healthy=is_healthy,
                                              description=source,
                                              value=response))
-        self.report()
+        self._report()
 
     def report_host_plugin_status(self, is_healthy, response):
         """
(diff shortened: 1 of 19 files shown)


azurelinuxagent/common/protocol/healthservice.py
```diff
@@ -0,0 +1,140 @@
...
+            "Value": self.value[:128]
+        }
+
+
+class HealthService(object):
+
+    ENDPOINT = 'http://{0}:80/HealthService'
+    API = 'reporttargethealth'
+    VERSION = "1.0"
+    OBSERVER_NAME = 'WALinuxAgent'
```
boumenot: >OBSERVER_NAME = 'WALinuxAgent [](start = 4, length = 29)

There should be an existing constant we can key off for this, no need for a new one.

azurelinuxagent/common/protocol/healthservice.py
```diff
@@ -0,0 +1,140 @@
...
+    HOST_PLUGIN_HEARTBEAT_OBSERVATION_NAME = 'GuestAgentPluginHeartbeat'
```
boumenot: >GuestAgentPluginHeartbeat [](start = 46, length = 25)

In conversation I call this HostGA, in the code we call it HostPlugin, and here we are calling it GuestAgentPlugin.  Is this the same things or a different thing?  If it is the same, I would prefer HostPlugin or HostGA.

azurelinuxagent/common/protocol/healthservice.py
```diff
@@ -0,0 +1,140 @@
...
+        }
+        return json.dumps(data)
+
+    def report_host_plugin_heartbeat(self, is_healthy):
+        """
+        Reports a signal for /health
+        :param is_healthy: whether the call succeeded
+        """
+        self.observations.append(Observation(name=HealthService.HOST_PLUGIN_HEARTBEAT_OBSERVATION_NAME,
+                                             is_healthy=is_healthy))
```
boumenot: Is it possible for these to grow unbounded?  Should you only keep the last 5? 10? observations?

azurelinuxagent/common/protocol/healthservice.py
```diff
@@ -0,0 +1,140 @@
...
+            "ObservationName": self.name[:64],
+            "IsHealthy": self.is_healthy,
+            "Description": self.description[:128],
+            "Value": self.value[:128]
+        }
+
+
+class HealthService(object):
+
+    ENDPOINT = 'http://{0}:80/HealthService'
```
boumenot: >:80 [](start = 26, length = 3)

Is this always 80?

azurelinuxagent/common/protocol/healthservice.py
```diff
@@ -0,0 +1,140 @@
...
+            "Api": self.api,
+            "Version": self.version,
+            "Source": self.source,
+            "Observations": [o.as_obj for o in self.observations]
+        }
+        return json.dumps(data)
+
+    def report_host_plugin_heartbeat(self, is_healthy):
+        """
+        Reports a signal for /health
```
boumenot: > Reports a signal for /health [](start = 7, length = 29)

Can you add a comment or point me to the documentation that defines what is_healthy means?

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -117,6 +147,44 @@ def get_artifact_request(self, artifact_url, artifact_manifest_url=None):
...
+
+    def report_status_health(self, is_healthy, response=''):
+        if self.should_report(is_healthy,
+                              self.status_error_state,
+                              self.status_last_timestamp,
+                              HostPluginProtocol.STATUS_REPORTING_PERIOD):
+            self.status_last_timestamp = datetime.datetime.utcnow()
+            health_signal = self.status_error_state.is_triggered() is False
+            self.health_service.report_host_plugin_status(is_healthy=health_signal,
+                                                          response=response)
```
boumenot: I think this method and the one above are really distinct classes.

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -117,6 +147,44 @@ def get_artifact_request(self, artifact_url, artifact_manifest_url=None):
...
+
+        if is_healthy:
+            error_state.reset()
+        else:
+            error_state.incr()
+
+        if last_timestamp is None:
+            last_timestamp = datetime.datetime.utcnow() - period
+
+        return datetime.datetime.utcnow() >= (last_timestamp + period)
```
boumenot: >  return datetime.datetime.utcnow() >= (last_timestamp + period) [](start = 6, length = 64)

If you report, should you reset and start over again for the next period?  Is that handled elsewhere?

azurelinuxagent/common/utils/restutil.py
```diff
@@ -62,6 +61,10 @@
     httpclient.ACCEPTED
 ]
 
+UPSTREAM_FAILURE_CODES = [
```
boumenot: HOSTPLUGIN_UPSTREAM_FAILURE_CODES ?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1117,7 +1113,7 @@ def set_handler_status(self, status="NotReady", message="", code=0):
         except (IOError, ValueError, ProtocolError) as e:
             fileutil.clean_ioerror(e,
                 paths=[status_file])
-            self.logger.error("Failed to save handler status: {0}", traceback.format_exc())
+            self.logger.error("Failed to save handler status: {0}", ustr(e))
```
boumenot: > ustr(e) [](start = 67, length = 8)

The traceback was valuable in locating the cause.  I would not remove it.

azurelinuxagent/common/protocol/healthservice.py
hglkrijger: There is not, this is new.

azurelinuxagent/common/protocol/healthservice.py
hglkrijger: Decided on this after consulting the health store folks - this is a name which needs to be identifiable  from their perspective - they are host focused.

azurelinuxagent/common/protocol/healthservice.py
hglkrijger: No, they are cleared after every report.

azurelinuxagent/common/protocol/healthservice.py
hglkrijger: According to health store folks, yes.

azurelinuxagent/common/protocol/healthservice.py
hglkrijger: Sent by email.

azurelinuxagent/common/protocol/hostplugin.py
hglkrijger: No, we need to keep reporting failure until we see success. This is handled by line 179. This is a little different to how the existing error states are used.

azurelinuxagent/common/utils/restutil.py
hglkrijger: Updated

azurelinuxagent/ga/exthandlers.py
hglkrijger: I removed this because I found that it was not populated, due to the way the threads work. Are you sure this specific one was useful? If so I can add it back.

azurelinuxagent/ga/monitor.py
```diff
@@ -267,3 +256,84 @@ def add_sysinfo(self, event):
...
+        Send a health signal every HOST_PLUGIN_HEARTBEAT_PERIOD. The signal is 'Healthy' when we have been able to
+        communicate with HostGAPlugin at least once in the last HOST_PLUGIN_HEALTH_PERIOD.
+        """
+
+        if last_host_plugin_heartbeat is None:
+            last_host_plugin_heartbeat = datetime.datetime.utcnow() - MonitorHandler.HOST_PLUGIN_HEARTBEAT_PERIOD
+
+        if datetime.datetime.utcnow() < (last_host_plugin_heartbeat + MonitorHandler.HOST_PLUGIN_HEARTBEAT_PERIOD):
+            return last_host_plugin_heartbeat
+
```
boumenot: This construct is used over and over again.  This should be refactored into something for better re-use.  Please open an issue if you cannot address it for this commit.

azurelinuxagent/ga/monitor.py
```diff
@@ -172,90 +180,71 @@ def collect_event(self, evt_file_name):
...
-                hostplugin_errors = io_errors.get("hostplugin")
-                protocol_errors = io_errors.get("protocol")
-                other_errors = io_errors.get("other")
-
-                if hostplugin_errors > 0 \
-                        or protocol_errors > 0 \
-                        or other_errors > 0:
-
-                    msg = "hostplugin:{0};protocol:{1};other:{2}"\
-                        .format(hostplugin_errors,
```
boumenot: Did you just re-format the file?  It's hard to see real changes when the formatting gets adjusted this much.  Keep the format changes because I want them, but consider doing them in a distinct phase to make code review easier.

azurelinuxagent/ga/update.py
```diff
@@ -540,7 +542,7 @@ def _purge_agents(self):
 
         known_versions = [agent.version for agent in self.agents]
         if CURRENT_VERSION not in known_versions:
-            logger.info(
+            logger.verbose(
```
boumenot: +1

tests/protocol/test_hostplugin.py
```diff
@@ -231,12 +245,18 @@ def test_validate_http_request(self):
...
-                self.assertTrue(patch_http.call_count == 1)
+                self.assertTrue(patch_http.call_count == 2)
+
+                # first call is to host plugin
                 self._validate_hostplugin_args(
                     patch_http.call_args_list[0],
                     test_goal_state,
                     exp_method, exp_url, exp_data)
 
+                # second call is to health service
```
boumenot: ># second call is to health service [](start = 16, length = 34)

Thank you, this makes it so much easier to read.

tests/protocol/test_hostplugin.py
```diff
@@ -365,15 +399,159 @@ def test_validate_get_extension_artifacts(self):
...
+        self.assertEqual(1, len(obj['Observations']))
+        self.assertTrue(obj['Observations'][0]['IsHealthy'])
+        self.assertEqual('GuestAgentPluginVersions', obj['Observations'][0]['ObservationName'])
+
+        # signal for /status
+        self.assertEqual(health_service_url, patch_http_post.call_args_list[1][0][0])
+        jstr = patch_http_post.call_args_list[1][0][1]
+        obj = json.loads(jstr)
+        self.assertEqual(1, len(obj['Observations']))
+        self.assertFalse(obj['Observations'][0]['IsHealthy'])
```
boumenot: > self.assertFalse(obj['Observations'][0]['IsHealthy']) [](start = 7, length = 54)

I think this is really what you want the test to assert, but it takes a lot of code to actually get there.  I have found this to be true for other parts of the agent. :cry:

azurelinuxagent/ga/monitor.py
hglkrijger: I have made some changes here, which I will push. Let me know if you think it still requires further refactoring.

tests/protocol/test_hostplugin.py
```diff
@@ -365,15 +399,159 @@ def test_validate_get_extension_artifacts(self):
             for k in expected_headers:
                 self.assertTrue(k in actual_headers)
                 self.assertEqual(expected_headers[k], actual_headers[k])
-    
+
+    @patch("azurelinuxagent.common.utils.restutil.http_get")
+    def test_health(self, patch_http_get):
```
boumenot: >test_health [](start = 8, length = 11)

I think there should be a test case for HTTP 502.  I think there should be a test case of /extensionArtifact and /vmAgentLog.  What else am I missing?

azurelinuxagent/ga/monitor.py
hglkrijger: I usually keep reformats/whitespace to separate commits, but in a review this size I understand it becomes too difficult to look at them separately.

tests/protocol/test_hostplugin.py
hglkrijger: There are quite a few tests still coming.

azurelinuxagent/common/protocol/healthservice.py
boumenot: Have a constant called AGENT_NAME that has the value `WALinuxAgent`.  Why is this not the correct one to use?

azurelinuxagent/common/protocol/healthservice.py
boumenot: I dislike the proliferation of more names. ☹️

azurelinuxagent/common/protocol/healthservice.py
boumenot: Is it possible for reports to be delayed?

tests/protocol/test_hostplugin.py
```diff
@@ -543,6 +545,86 @@ def test_put_status_unhealthy_signal_permanent(self, patch_http_put, patch_http_
...
+    @patch("azurelinuxagent.common.protocol.healthservice.HealthService.report_host_plugin_status")
+    def test_report_status_health(self, patch_report_status, patch_should_report):
+        host_plugin = self._init_host()
+        self.assertEqual(None, host_plugin.status_last_timestamp)
+        host_plugin.report_status_health(is_healthy=True)
+        self.assertNotEqual(None, host_plugin.status_last_timestamp)
+        self.assertEqual(1, patch_should_report.call_count)
+        self.assertEqual(1, patch_report_status.call_count)
+
+    def test_should_report(self):
```
boumenot: This is test clarifies a lot of things for me - nice!

tests/ga/test_monitor.py
```diff
@@ -90,14 +90,16 @@ def test_heartbeats(self, patch_hostplugin_heartbeat, patch_send_events, patch_t
...
         monitor_handler.start()
         time.sleep(1)
         self.assertTrue(monitor_handler.is_alive())
+
         self.assertNotEqual(0, patch_hostplugin_heartbeat.call_count)
         self.assertNotEqual(0, patch_send_events.call_count)
         self.assertNotEqual(0, patch_telemetry_heartbeat.call_count)
 
-        monitor_handler.should_run = False
+        monitor_handler.stop()
```
boumenot: I like this change!

azurelinuxagent/common/protocol/healthservice.py
hglkrijger: Semantically, these are distinct. We should be able to change one without changing the other.

azurelinuxagent/common/protocol/healthservice.py
hglkrijger: Fair point, let me consider how to ensure additional safety bounds.

azurelinuxagent/common/protocol/hostplugin.py
hglkrijger: Sure.

azurelinuxagent/ga/exthandlers.py
hglkrijger: There is no harm in having both - I will add it back along with the error message, in case I missed something.

tests/ga/test_monitor.py
```diff
@@ -76,3 +80,94 @@ def test_add_sysinfo(self, _, __):
...
+        MonitorHandler.TELEMETRY_HEARTBEAT_PERIOD = timedelta(seconds=1)
+        MonitorHandler.EVENT_COLLECTION_PERIOD = timedelta(seconds=1)
+        MonitorHandler.HOST_PLUGIN_HEARTBEAT_PERIOD = timedelta(seconds=1)
+
+        self.assertEqual(None, monitor_handler.last_host_plugin_heartbeat)
+        self.assertEqual(None, monitor_handler.last_event_collection)
+        self.assertEqual(None, monitor_handler.last_telemetry_heartbeat)
+
+        monitor_handler.start()
+        time.sleep(1)
```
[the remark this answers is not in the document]
hglkrijger: I made these shorter. Since I want to test how the thread sleep functions, I felt patching time would negate the value of the test.

tests/protocol/test_healthservice.py
```diff
@@ -0,0 +1,199 @@
...
+        self.assertEqual('', o.value)
+        self.assertEqual('', o.description)
+
+        long_str = 's' * 200
+        o = Observation(name=long_str, is_healthy=True, value=long_str, description=long_str)
+        self.assertEqual(200, len(o.name))
+        self.assertEqual(200, len(o.value))
+        self.assertEqual(200, len(o.description))
+
+        self.assertEqual(64, len(o.as_obj['ObservationName']))
```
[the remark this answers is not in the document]
hglkrijger: This test is to assert the maximum string length for these fields; longer values cause 400s. The other tests do assert value. Specifically this ensures that even though we set a 200 character string, the json representation is 64/128.

tests/protocol/test_hostplugin.py
hglkrijger: Awesome.

azurelinuxagent/common/protocol/healthservice.py
hglkrijger: I have made an update for this, enforcing a maximum list size of 10

hglkrijger:
Merging, since this is approved, but if there is any more feedback I will be happy to address it in the next PR for integrating IMDS health signals.

azurelinuxagent/common/protocol/healthservice.py
jasonzio: Which strongly argues for us to stop calling this HostGA or HostPlugin and start calling it the GuestAgentPlugin, or GAP.

PR #1199, opened 2018-06-06
Cgroup support for monitoring agent and extensions

jasonzio:
Passes hand smoke-tests on Ubuntu 14.04 (without systemd) and 16.04 (with systemd). Unit tests are disabled to keep Travis from trying to run them; mounting the cgroup device doesn't seem to work terribly well within containers, at least as Travis has them configured.

### DIFF
--- a/azurelinuxagent/common/cgroups.py
+++ b/azurelinuxagent/common/cgroups.py
@@ -30,10 +30,6 @@ WRAPPER_CGROUP_NAME = "Agent+Extensions"
 METRIC_HIERARCHIES = ['cpu', 'memory']
 MEMORY_DEFAULT = -1
 
-# cpu  813599 3940 909253 154538746 874851 0 6589 0 0 0
-# cpu0 401094 1516 453006 77276738 452939 0 3312 0 0 0
-# cpu1 412505 2423 456246 77262007 421912 0 3276 0 0 0
-re_all_CPUs = re.compile('^cpu .*')
 re_user_system_times = re.compile('user (\d+)\nsystem (\d+)\n')
 
 related_services = {
@@ -45,12 +41,7 @@ related_services = {
 class CGroupsException(Exception):
 
     def __init__(self, msg):
-
         self.msg = msg
-        if CGroups.enabled():
-            pid = os.getpid()
-            # logger.verbose("[{0}] Disabling cgroup support: {1}".format(pid, msg))
-            # CGroups.disable()
 
     def __str__(self):
         return repr(self.msg)
@@ -66,9 +57,10 @@ class Cpu(object):
         :return:
         """
         self.cgt = cgt
+        self.osutil = get_osutil()
         self.current_cpu_total = self.get_current_cpu_total()
         self.previous_cpu_total = 0
-        self.current_system_cpu = Cpu.get_current_system_cpu()
+        self.current_system_cpu = self.osutil.get_total_cpu_ticks_since_boot()
         self.previous_system_cpu = 0
 
     def __str__(self):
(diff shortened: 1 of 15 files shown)


azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,689 @@
...
+        Determine if systemd is managing system services. Many extensions are structured as a set of services,
+        including the agent itself; systemd expects those services to remain in the cgroups in which it placed them.
+        If this process (presumed to be the agent) is in a cgroup that looks like one created by systemd, we can
+        assume systemd is in use.
+
+        :return: True if systemd is managing system services
+        :rtype: Bool
+        """
+        if CGroups._use_systemd is None:
+            hierarchy = METRIC_HIERARCHIES[0]
```
boumenot: I know that [0] is cpu, but it isn't a very constant.

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,689 @@
...
+            path = CGroups.get_my_cgroup_folder(hierarchy)
+            CGroups._use_systemd = path.startswith(CGroups._construct_systemd_path_for_hierarchy(hierarchy, ""))
+        return CGroups._use_systemd
+
+    @staticmethod
+    def _try_mkdir(path):
```
boumenot: Lost comments.  Seems useful on its own.  Consider moving to osutil for a future PR.

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,689 @@
...
+    def get_my_cgroup_path(hierarchy_id):
+        """
+        Get the cgroup path "suffix" for this process for the given hierarchy ID. The leading "/" is always stripped,
+        so the suffix is suitable for passing to os.path.join(). (If the process is in the root cgroup, an empty
+        string is returned, and os.path.join() will still do the right thing.)
+
+        :param hierarchy_id: str
+        :return: str
+        """
+        cgroup_paths = fileutil.read_file("/proc/self/cgroup")
```
boumenot: Lost comments?

I believe this is an example of what you're parsing, correct?  If so, please add in a comment.

```text
3:cpu,cpuacct:/user.slice
```

Please add an example.  Please consider encapsulating this parsing to return an object to make code read easier for a future PR.

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,689 @@
...
+    def get_hierarchy_id(hierarchy):
+        """
+        Get the cgroups hierarchy ID for a given hierarchy name
+
+        :param hierarchy:
+        :return: str
+        """
+        cgroup_states = fileutil.read_file("/proc/cgroups")
+        for entry in cgroup_states.splitlines():
+            fields = entry.split('\t')
```
boumenot: Thank you CGroups for using '\t' and ':' for separation.

boumenot: Please consider adding a comment on what you're parsing.

```text
#subsys_name    hierarchy       num_cgroups     enabled
cpuset  9       1       1
cpu     3       74      1
cpuacct 3       74      1
blkio   10      74      1
memory  12      111     1
devices 6       74      1
freezer 11      1       1
net_cls 2       1       1
perf_event      8       1       1
net_prio        2       1       1
hugetlb 5       1       1
pids    4       75      1
rdma    7       1       1
```

Please consider returning an object with each field called out to make it easier for readers in a future PR.

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,689 @@
...
+            raise CGroupsException('CPU Limit must be convertible to a float')
+
+        if limit <= float(0) or limit > float(CGroups.get_num_cores() * 100):
+            raise CGroupsException('CPU Limit must be between 0 and 100 * numCores')
+
+        return limit / 100.0
+
+    def get_file(self, hierarchy, file_name):
+        """
+        Retrieve the value of a parameter from a hierarchy.
```
boumenot: The method is called get_file, but that doesn't match the comment or code.  Is read_file or get_file_contents better? Is this method only intended to be call by get_parameter?

azurelinuxagent/common/version.py
```diff
@@ -113,7 +113,7 @@ def get_distro():
 
 AGENT_NAME = "WALinuxAgent"
 AGENT_LONG_NAME = "Azure Linux Agent"
-AGENT_VERSION = '2.2.25.1'
+AGENT_VERSION = '2.2.26'
```
boumenot: Lost comment? New version is 2.2.27.

azurelinuxagent/common/cgroups.py
```diff
@@ -0,0 +1,689 @@
...
+        not_enabled_extensions = set()
+        for extension in ext_handlers:
+            if extension.properties.state == u"enabled":
+                CGroupsTelemetry.track_extension(extension.name)
+            else:
+                not_enabled_extensions.add(extension.name)
+
+        names_now_tracked = set(CGroupsTelemetry._tracked.keys())
+        if CGroupsTelemetry.tracked_names != names_now_tracked:
+            now_tracking = " ".join("[{0}]".format(name) for name in names_now_tracked)
```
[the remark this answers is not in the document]
jasonzio: OK

azurelinuxagent/common/cgroups.py
jasonzio: I don't care which hierarchy it is; it just has to be one of the hierarchies from which we're selecting metrics. It could be "memory" for all I care, and the code will work.

azurelinuxagent/common/cgroups.py
jasonzio: I'll consider it... later. :-)

azurelinuxagent/common/cgroups.py
jasonzio: CGroups.get_file_contents() it is. Any holder of a CGroups object can legitimately call this; cgroups is so wildly varied that it's hard to know. I am pretty sure at least one DCR test for cgroups relies on this method being public.

azurelinuxagent/common/version.py
jasonzio: At the time, your comment said "make it 2.2.26", so I did. Gosh, you're so picky. :-)

PR #1208, opened 2018-06-12
Fix host plugin status reporting

hglkrijger:
Couple issues here:

- when we receive 410, the goal state is never force updated, due to a simple logic bug
- in this and other cases, we never raise a protocol error high enough to trigger the error state, effectively hiding extended report status failures

This addresses both, and updates unit tests.

### DIFF
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -969,16 +969,17 @@ class WireClient(object):
                                    ext_conf.status_upload_blob,
                                    ext_conf.status_upload_blob_type)
                 return
+            except ResourceGoneError:
+                # do not attempt direct, force goal state update and wait to try again
+                self.update_goal_state(forced=True)
+                return
             except Exception as e:
-                if isinstance(e, ResourceGoneError):
-                    # do not attempt direct, force goal state update and wait to try again
-                    self.update_goal_state(forced=True)
-                    return
+            	# for all other errors, fall back to direct
+		pass
 
-            # for all other errors, fall back to direct
             self.report_status_event("direct")
             if self.status_blob.upload(blob_uri):
-                return
+		return
 
         except Exception as e:
             self.report_status_event("Exception uploading status blob: {0}", ustr(e))
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -974,12 +974,12 @@ class WireClient(object):
                 self.update_goal_state(forced=True)
                 return
             except Exception as e:
-            	# for all other errors, fall back to direct
-		pass
+                # for all other errors, fall back to direct
+                pass
 
             self.report_status_event("direct")
             if self.status_blob.upload(blob_uri):
-		return
+                return
 
         except Exception as e:
             self.report_status_event("Exception uploading status blob: {0}", ustr(e))
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -935,73 +935,55 @@ class WireClient(object):
             raise ProtocolNotFoundError(error)
 
     def upload_status_blob(self):
-        for update_goal_state in [False, True]:
+        try:
+            self.update_goal_state()
+            ext_conf = self.get_ext_conf()
+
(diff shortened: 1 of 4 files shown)


tests/protocol/test_hostplugin.py
```diff
@@ -129,65 +129,169 @@ def _validate_hostplugin_args(self, args, goal_state, exp_method, exp_url, exp_d
...
+        wire_protocol_client.upload_status_blob()
+
+        # assert direct route is not called
+        self.assertEqual(0, patch_upload.call_count, "Direct channel was used")
+
+        # assert host plugin route is called
+        self.assertEqual(1, patch_put.call_count, "Host plugin was not used")
+
+        # assert update goal state is only called once, non-forced
+        self.assertEqual(1, patch_update.call_count, "Unexpected call count")
```
boumenot: Assert that forced was set to False.

tests/protocol/test_hostplugin.py
```diff
@@ -129,65 +129,169 @@ def _validate_hostplugin_args(self, args, goal_state, exp_method, exp_url, exp_d
...
+        wire_protocol_client.get_goal_state = Mock(return_value=test_goal_state)
+        wire_protocol_client.ext_conf = wire.ExtensionsConfig(None)
+        wire_protocol_client.ext_conf.status_upload_blob = sas_url
+        wire_protocol_client.ext_conf.status_upload_blob_type = page_blob_type
+        wire_protocol_client.status_blob.set_vm_status(status)
+
+        # act
+        wire_protocol_client.upload_status_blob()
+
+        # assert direct route is not called
```
boumenot: The comment and the assert message do not appear to agree.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -935,74 +935,56 @@ def check_wire_protocol_version(self):
...
+            # to see how this behaves at scale, and have a fallback should things go
+            # wrong.  This is why we try HostPlugin then direct.
+            try:
+                host = self.get_host_plugin()
+                host.put_vm_status(self.status_blob,
+                                   ext_conf.status_upload_blob,
+                                   ext_conf.status_upload_blob_type)
+                return
             except Exception as e:
                 if isinstance(e, ResourceGoneError):
```
jasonzio: If you're going to treat one exception type differently, then catch it via a separate `except` statement.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -935,74 +935,56 @@ def check_wire_protocol_version(self):
...
+                    # do not attempt direct, force goal state update and wait to try again
+                    self.update_goal_state(forced=True)
+                    return
 
-                self.report_status_event(
-                    "Exception uploading status blob: {0}", ustr(e))
+            # for all other errors, fall back to direct
+            self.report_status_event("direct")
+            if self.status_blob.upload(blob_uri):
                 return
```
jasonzio:
```
try:
    host = self.get_host_plugin()
    host.put_vm_status(self.status_blob,
        ext_conf.status_upload_blob,
        ext_conf.status_upload_blob_type) 
    return
except ResourceGoneError as e:
    # do not attempt direct, force goal state update and wait to try again
    self.update_goal_state(forced=True)
except Exception as e:
    # for all other errors, fall back to direct 
    pass
```
etc.

tests/protocol/test_hostplugin.py
hglkrijger: That is what the check in L163 does - validate that no parameters were passed to the method, which defaults to False.

tests/protocol/test_hostplugin.py
hglkrijger: Thanks, fixed.

azurelinuxagent/common/protocol/wire.py
hglkrijger: Should have cleaned this up, thanks.

azurelinuxagent/common/protocol/wire.py
hglkrijger: Updated.

hglkrijger:
Feedback is addressed; I am going to merge since I would like to get this in the nightly test runs. If I missed anything, please let me know.

PR #1212, opened 2018-06-14
Better handling for missing certs

hglkrijger:
- slightly more graceful handling of this error case
- logging cleanup
- minor pep8 fixes

This mainly raises an IOError for the case when cert files are not found. For the moment there is no behavior change, just slightly different logging, but now that the errors are handled separately we can decide whether to hide those that are transient.

### DIFF
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -15,16 +15,17 @@
 # limitations under the License.
 #
 # Requires Python 2.6+ and Openssl 1.0+
-
-from datetime import datetime
-
+import datetime
 import json
 import os
 import random
 import re
 import sys
 import time
+import traceback
 import xml.sax.saxutils as saxutils
+
+from datetime import datetime
 
 import azurelinuxagent.common.conf as conf
 import azurelinuxagent.common.utils.fileutil as fileutil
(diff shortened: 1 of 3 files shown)


boumenot:
LGTM.

I went back and forth on this in my head, but unit tests may or may not be valuable.  They will assert this behavior, and they will assert that the right file throws an exception if it does not exist.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -779,17 +779,17 @@ def update_goal_state(self, forced=False, max_retry=3):
 
                 return
 
+            except IOError as e:
+                logger.warn("IOError processing goal state, retrying [{0}]", ustr(e))
```
jasonzio: IOError exception object has a strerror member which is probably more appropriate for error messages, instead of converting the exception object to a string.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -779,17 +779,17 @@ def update_goal_state(self, forced=False, max_retry=3):
...
                 goal_state = None
 
             except Exception as e:
-                log_method = logger.info \
-                                if type(e) is ProtocolError \
-                                else logger.warn
-                log_method(
-                    "Exception processing GoalState-related files: {0}".format(
-                        ustr(e)))
+                log_method = logger.verbose if type(e) is ProtocolError else logger.error
```
jasonzio: I'm not a big fan of this idiom (catch Exception, then change behavior based on checking type()). I'd rather see two separate "except" statements which are explicit about what they do.
```
except ProtocolError as e:
    logger.verbose(exception_msg.format(ustr(e), traceback.format_exc()))
except Exception as e:
    logger.error(exception_msg.format(ustr(e), traceback.format_exc()))
```

PR #1217, opened 2018-06-19
BVT cleanup

hglkrijger:
- catch json parsing exceptions for artifacts profile blob and send telemetry
- explicitly handle control flow when cert file does not exist
- remove vscode settings and minor cleanup

### DIFF
--- a/azurelinuxagent/common/utils/cryptutil.py
+++ b/azurelinuxagent/common/utils/cryptutil.py
@@ -51,46 +51,45 @@ class CryptUtil(object):
     def get_pubkey_from_prv(self, file_name):
         if not os.path.exists(file_name):
             raise IOError("File not found: {0}".format(file_name))
-
-        cmd = "{0} rsa -in {1} -pubout 2>/dev/null".format(self.openssl_cmd, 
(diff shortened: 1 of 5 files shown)


azurelinuxagent/common/utils/cryptutil.py
```diff
@@ -50,47 +50,46 @@ def gen_transport_cert(self, prv_file, crt_file):
 
     def get_pubkey_from_prv(self, file_name):
         if not os.path.exists(file_name):
-            raise IOError("File not found: {0}", file_name)
-
-        cmd = "{0} rsa -in {1} -pubout 2>/dev/null".format(self.openssl_cmd, 
-                                                           file_name)
-        pub = shellutil.run_get_output(cmd)[1]
-        return pub
+            raise IOError("File not found: {0}".format(file_name))
```
jasonzio: Not the right way to raise IOError. While this sets the message, it doesn't set the errno member of the exception instance, it doesn't set the filename, and it puts too much information in strerror.

If you have to do this, this is the right way to do it:
```
```

PR #1219, opened 2018-06-20
Skip cgroups tests under WSL

hglkrijger:
- adds a skip predicate for WSL

### DIFF
--- a/azurelinuxagent/common/osutil/ubuntu.py
+++ b/azurelinuxagent/common/osutil/ubuntu.py
@@ -17,6 +17,7 @@
 #
 
 import os
+import platform
 import time
 
 import azurelinuxagent.common.logger as logger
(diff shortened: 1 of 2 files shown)


tests/common/test_cgroups.py
```diff
@@ -62,7 +63,10 @@ def path_maker(hierarchy, _):
 def i_am_root():
     return os.geteuid() == 0
 
+def is_wsl():
+    return '-Microsoft-' in platform.platform()
 
+@skip_if_predicate_true(is_wsl, "CGroups not supported under WSL")
```
jasonzio: All this does is suppress the test; it doesn't keep the agent code itself from trying to do the CGroup thing. The better fix would be to arrange for CGroups.enabled to always be False if it's running under WSL.

Is there an osutil subclass for WSL? If we created one, that would be the place to insert the necessary code to cause CGroups functionality to be completely disabled (including the tests).

jasonzio: In osutil/ubuntu.py:
```
    def is_cgroups_supported(self):
        if ('TRAVIS' in os.environ and os.environ['TRAVIS'] == 'true') \
                or ('-Microsoft-' in platform.platform()):
            return False
        return True
```

hglkrijger: Ok, updated

azurelinuxagent/common/osutil/ubuntu.py
```diff
@@ -56,9 +57,9 @@ def get_dhcp_lease_endpoint(self):
         return self.get_endpoint_from_leases_path('/var/lib/dhcp/dhclient.*.leases')
 
     def is_cgroups_supported(self):
-        if 'TRAVIS' in os.environ and os.environ['TRAVIS'] == 'true':
-            return False
-        return True
+        is_wsl = '-Microsoft-' in platform.platform()
```
boumenot: Nice.

PR #1222, opened 2018-06-21
Log network routes if changed

jasonzio:
Check the sha1 hash of the route table (/proc/net/route); if it changes, emit the route table as a JSON array of route structures, to the log.

### DIFF
--- a/azurelinuxagent/ga/monitor.py
+++ b/azurelinuxagent/ga/monitor.py
@@ -40,8 +40,9 @@ from azurelinuxagent.common.protocol.restapi import TelemetryEventParam, \
                                                     TelemetryEventList, \
                                                     TelemetryEvent, \
                                                     set_properties
+import azurelinuxagent.common.utils.networkutil as networkutil
 from azurelinuxagent.common.utils.restutil import IOErrorCounter
-from azurelinuxagent.common.utils.textutil import parse_doc, findall, find, getattrib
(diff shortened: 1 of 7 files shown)


azurelinuxagent/ga/monitor.py
```diff
@@ -416,3 +419,11 @@ def send_cgroup_telemetry(self):
                 logger.warn("Monitor: updating tracked extensions raised {0}: {1}", e, traceback.format_exc())
 
             self.last_cgroup_telemetry = datetime.datetime.utcnow()
+
+    def log_route_table(self):
+        raw_route_list = self.osutil.read_route_table()
+        digest = hash_strings(raw_route_list)
```
boumenot: Is it possible for the route list to be the same, but be in a different order.  The hash will change, but the meaning really hasn't.  I wouldn't think so...

```text
{"a": 1, "b": 2} == {"b": 2, "a": 1} but the hashes are different.
```

jasonzio: It's possible, but since /proc/net/route is a direct pipeline to the kernel data structure, there'd be no reason for the order of entries to change if nothing perturbed it. My testing shows the table (as exposed through /proc) doesn't change over a period of hours if there's no actual changes to routes; that's good enough for me.

boumenot: Me too.

PR #1225, opened 2018-06-21
disable wrong private key tests

matkin-msft:


tests/utils/test_crypt_util.py
```diff
@@ -49,19 +50,20 @@ def test_decrypt_encrypted_text_missing_private_key(self):
         encrypted_string = load_data("wire/encrypted.enc")
         prv_key = os.path.join(self.tmp_dir, "TransportPrivate.pem")
         crypto = CryptUtil(conf.get_openssl_cmd())
-        self.assertRaises(CalledProcessError, crypto.decrypt_secret, encrypted_string, "abc" + prv_key)
-    
+        self.assertRaises(CryptError, crypto.decrypt_secret, encrypted_string, "abc" + prv_key)
+
+    @unittest.skip("Disabled due to OpenSSLBug")
```
hglkrijger: Could we rather add a predicate for Python 2.6? Or just simply remove the test if it should be disabled everywhere.

jasonzio:
```
import sys

def is_python_version_26():
    return sys.version_info.major == 2 and sys.version_info.minor == 6
```

And then, later on...

```
    @skip_if_predicate_true(is_python_version_26, "Disabled on Python 2.6")
    def test_my_bad_test(self):
        # etc etc
```

matkin-msft: it fails in any version, it just seems 2.6 is far more often.... I've made the change and we can always remove it completely if it causes issues in later versions.

---
In reply to: [197294949](https://github.com/Azure/WALinuxAgent/pull/1225#discussion_r197294949) [](ancestors = 197294949)

matkin-msft: added per Jason's comment below

---
In reply to: [197290246](https://github.com/Azure/WALinuxAgent/pull/1225#discussion_r197290246) [](ancestors = 197290246)

tests/utils/test_crypt_util.py
```diff
@@ -34,6 +36,11 @@
 from tests.tools import *
 from subprocess import CalledProcessError
 
+
+def is_python_version_26():
+    return sys.version_info.major == 2 and sys.version_info.minor == 6
```
jasonzio: Sigh. Python 2.6 *sucks*. The .major and .minor attributes appeared in 2.7. This ought to work:

```
return sys.version_info[0] == 2 and sys.version_info[1] == 6
```

jasonzio:
LGTM

PR #1228, opened 2018-06-22
Filter out empty artifact profile blobs

hglkrijger:
```
2018/06/22 07:16:03.677467 ERROR ExtHandler Event: name=WALinuxAgent, op=ArtifactsProfileBlob, message=Content: [<0x00><0x00><0x00><0x00>
```
- add a check for null only content when the blob has not been written yet
- only log telemetry for failures when `log_event` is true
- add unit test for empty strings

### DIFF
--- a/azurelinuxagent/common/event.py
+++ b/azurelinuxagent/common/event.py
@@ -256,7 +256,7 @@ class EventLogger(object):
                   is_internal=False,
                   log_event=True):
 
-        if not is_success or log_event:
+        if not is_success and log_event:
             _log_event(name, op, message, duration, is_success=is_success)
 
         self._add_event(duration, evt_type, is_internal, is_success, message, name, op, version, eventId=1)
@@ -347,13 +347,14 @@ def elapsed_milliseconds(utc_start):
                     (d.microseconds / 1000.0))
 
 
-def report_event(op, is_success=True, message=''):
+def report_event(op, is_success=True, message='', log_event=True):
     from azurelinuxagent.common.version import AGENT_NAME, CURRENT_VERSION
     add_event(AGENT_NAME,
               version=CURRENT_VERSION,
               is_success=is_success,
               message=message,
-              op=op)
+              op=op,
+              log_event=log_event)
 
 
 def report_periodic(delta, op, is_success=True, message=''):
--- a/azurelinuxagent/common/event.py
+++ b/azurelinuxagent/common/event.py
@@ -256,7 +256,7 @@ class EventLogger(object):
                   is_internal=False,
                   log_event=True):
 
-        if not is_success and log_event:
+        if (not is_success) and log_event:
             _log_event(name, op, message, duration, is_success=is_success)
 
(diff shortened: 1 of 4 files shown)


azurelinuxagent/common/event.py
```diff
@@ -256,7 +256,7 @@ def add_event(self,
                   is_internal=False,
                   log_event=True):
 
-        if not is_success or log_event:
+        if not is_success and log_event:
```
jasonzio: Consider adding parentheses to make precedence clear without requiring the reader to look up whether `not` or `and` binds more tightly.

tests/utils/test_text_util.py
```diff
@@ -159,6 +159,33 @@ def test_hash_list(self):
...
+
+        self.assertFalse(textutil.is_str_empty(u' \x01 '))
+        self.assertFalse(textutil.is_str_empty(u'foo'))
+        self.assertFalse(textutil.is_str_empty('bar'))
+
+        self.assertFalse(textutil.is_str_none_or_whitespace(u'\x00'))
+        self.assertFalse(textutil.is_str_none_or_whitespace(u' \x00 '))
+
+        self.assertTrue(textutil.is_str_empty(u'\x00'))
+        self.assertTrue(textutil.is_str_empty(u' \x00 '))
```
jasonzio: Consider pulling these strings into named variables (which can live within this method) so you can be certain you're invoking the two methods with identical strings.

tests/utils/test_text_util.py
```diff
@@ -159,6 +159,33 @@ def test_hash_list(self):
         self.assertEqual(result_from_list, hash_from_string.digest())
         self.assertEqual(hash_from_string.hexdigest(), '6367c48dd193d56ea7b0baad25b19455e529f5ee')
 
+    def test_empty_strings(self):
+        self.assertTrue(textutil.is_str_none_or_whitespace(' '))
```
boumenot: textutil.is_str_none_or_whitespace(None) ??

tests/utils/test_text_util.py
jasonzio: These last two cases are critical: they identify scenarios where the two methods are _required_ to return different values. Consider reframing those test cases to assert that the distinction exists. That is...
```
hex_null = u'\x00'
hex_null_with_blanks = u' \x00 '
self.assertTrue(textutil.is_str_empty(hex_null) and not textutil.is_str_none_or_whitespace(hex_null))
self.assertTrue(textutil.is_str_empty(hex_null_with_blanks) and not textutil.is_str_none_or_whitespace(hex_null_with_blanks))
```

tests/utils/test_text_util.py
hglkrijger: Updated

azurelinuxagent/common/event.py
hglkrijger: Updated

tests/utils/test_text_util.py
hglkrijger: Updated

hglkrijger: Updated

PR #1229, opened 2018-06-22
Log network interface settings at startup and when they change.

jasonzio:


azurelinuxagent/common/utils/networkutil.py
```diff
@@ -61,3 +61,40 @@ def __str__(self):
...
+    def __eq__(self, other):
+        return self.link == other.link and \
+               self.ipv4 == other.ipv4 and \
+               self.ipv6 == other.ipv6
+
+    @staticmethod
+    def _json_string_or_array(items):
+        quoted_values = ['"{0}"'.format(x) for x in sorted(items)]
+        contents = ",".join(quoted_values)
+        if len(items) == 1:
```
boumenot: Why isn't this always an array?  Why force the caller to have to check?

azurelinuxagent/common/osutil/default.py
```diff
@@ -1238,3 +1238,68 @@ def get_total_cpu_ticks_since_boot():
...
+
+        self._alter_nic_state(state, "ip -6 -a -d -o address", NetworkInterfaceCard.add_ipv6, "an IPv6 address")
+        """
+        1: lo    inet6 ::1/128 scope host \       valid_lft forever preferred_lft forever
+        2: eth0    inet6 fe80::20d:3aff:fe30:c35a/64 scope link \       valid_lft forever preferred_lft forever
+        """
+
+        return state
+
+    def _alter_nic_state(self, state, ip_command, handler, description):
```
boumenot: alter -> update ?

azurelinuxagent/common/osutil/default.py
```diff
@@ -1248,7 +1248,7 @@ def get_NIC_state(self):
         """
         state = {}
 
-        status, output = shellutil.run_get_output("ip -a -d -o link")
+        status, output = shellutil.run_get_output("/sbin/ip -a -d -o link")
```
boumenot: I don't think we use full path.  I would be leery of hard coding a path because it prevents customers from overriding 

I would either disable this test in Travis, or modify Travis' environment to include ip.  You may have to do the former.

PR #1230, opened 2018-06-25
BVT fixes

hglkrijger:
- when ip is not present or not all options are supported, ignore failures
- add info logging when process goal state is completed
- do not add periodic logs when exceptions are thrown
- pep8 fixes and cleanup


jasonzio:
Learn something new everyday.

Sent from my Windows 10 phone

PR #1235, opened 2018-06-27
Fix for extension handler cache

hglkrijger:
- correct the extension handler reference cache from monitor.py


hglkrijger:
Merging for BVTs, let me know if there is any feedback.

boumenot:
LGTM.

PR #1246, opened 2018-07-10
Use fileutil.makedir consistently to avoid "No such file or directory" errors

grafi-tt:
If `/var/lib/waagent` weren't created manually, `waagent` continues to fail with "No such file or directory" error, because it can't create `/var/lib/waagent/events`. I suppose officially supported images alrady have the directory. But anyway I'veencounterd this error on Gentoo Linux installation.

This PR is to avoid such errors by using fileutil.makedir, which creates directories recursively. Its use is consistent with other parts of the codes.

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -390,12 +390,12 @@ class DefaultOSUtil(object):
         if not remove:
             # for older distros create sudoers.d
             if not os.path.isdir(sudoers_dir):
-                sudoers_file = os.path.join(sudoers_dir, '../sudoers')
                 # create the sudoers.d directory
-                os.mkdir(sudoers_dir)
+                fileutil.mkdir(sudoers_dir)
                 # add the include of sudoers.d to the /etc/sudoers
-                sudoers = '\n#includedir ' + sudoers_dir + '\n'
-                fileutil.append_file(sudoers_file, sudoers)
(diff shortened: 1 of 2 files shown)


hglkrijger:
Thanks for the submission. This should actually be addressed by #1261. Could you please verify?

hglkrijger:
Thanks for the submission. I would like to take this as soon as the next release is stabilized.

azurelinuxagent/common/osutil/default.py
```diff
@@ -390,12 +390,12 @@ def conf_sudoer(self, username, nopasswd=False, remove=False):
         if not remove:
             # for older distros create sudoers.d
             if not os.path.isdir(sudoers_dir):
-                sudoers_file = os.path.join(sudoers_dir, '../sudoers')
                 # create the sudoers.d directory
-                os.mkdir(sudoers_dir)
+                fileutil.mkdir(sudoers_dir)
```
jasonzio: The sudoers needs to be created with specific protections. Please specify `mode=0o550`. Just as /etc/sudoers itself should be mode 0440 (not writeable), the /etc/sudoers.d directory shouldn't be writeable. (Root can write anything, of course.)

hglkrijger: True, but the missing permissions are not introduced with this PR. Merging as it still fixes an issue with the code.

PR #1271, opened 2018-07-24
Update the timeout mechanism for child processes

hglkrijger:
In the past we have used `poll()` to determine whether a child is active. Unfortunately that usage is flawed, and for any forked process this will always return a value. In other words our timeout enforcement has never really worked. 

The last set of changes to use `communicate(timeout)` and `os.kill(pid, 0)` do actually enforce the timeout correctly, but this is a breaking change in behavior, even though it is in the extension contract. As such we have to revert it.

In order to try and capture the process output wherever possible, I have made the following changes:

- for 3.3 and later: block for the initial timeout, then check whether the process is forked. For well-behaved short-lived and non-forking long-lived processes, this will capture output.
- earlier than 3.3: revert to the previous behavior for forked and short lived processes. For long-lived non-forked processes capture the output and terminate the process.

### DIFF
--- a/azurelinuxagent/common/utils/processutil.py
+++ b/azurelinuxagent/common/utils/processutil.py
@@ -103,13 +103,24 @@ def _destroy_process(process, signal_to_send=signal.SIGKILL):
 def capture_from_process_modern(process, cmd, timeout):
     try:
         stdout, stderr = process.communicate(timeout=timeout)
+        # process completed
     except subprocess.TimeoutExpired:
-        # Just kill the process. The .communicate method will gather stdout/stderr, close those pipes, and reap
-        # the zombie process. That is, .communicate() does all the other stuff that _destroy_process does.
-        os.killpg(os.getpgid(process.pid), signal.SIGKILL)
-        stdout, stderr = process.communicate()
-        msg = format_stdout_stderr(sanitize(stdout), sanitize(stderr))
-        raise ExtensionError("Timeout({0}): {1}\n{2}".format(timeout, cmd, msg))
+        return_code = process.poll()
+        if return_code is None:
+            # process did not fork and is still running after timeout
+            # we need to kill and clean up
+            os.killpg(os.getpgid(process.pid), signal.SIGKILL)
+            stdout, stderr = process.communicate()
+            msg = format_stdout_stderr(sanitize(stdout), sanitize(stderr))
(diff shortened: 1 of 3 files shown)


jasonzio:
For clarity: When Hans mentions "3.3 and later" and "earlier than 3.3", that refers to the version of Python which is executing the agent code. Python 3.3 added a timeout mechanism to Subprocess.Communicate() which makes the code much simpler and more robust.

jasonzio:
Would it be simpler to look at the identity of the extension being run and, if it's any flavor of Custom Script, bump the timeout up to 21 minutes? That is, narrow down the "exception from the contract" to just the case where the agent team had authorized a variance from the contract?

boumenot:
CustomScript is not the only extension that does this, and I am not convinced we know all of the ones that are.

azurelinuxagent/common/utils/processutil.py
```diff
@@ -103,13 +103,24 @@ def _destroy_process(process, signal_to_send=signal.SIGKILL):
...
         stdout, stderr = process.communicate(timeout=timeout)
+        # process completed
     except subprocess.TimeoutExpired:
-        # Just kill the process. The .communicate method will gather stdout/stderr, close those pipes, and reap
-        # the zombie process. That is, .communicate() does all the other stuff that _destroy_process does.
-        os.killpg(os.getpgid(process.pid), signal.SIGKILL)
-        stdout, stderr = process.communicate()
-        msg = format_stdout_stderr(sanitize(stdout), sanitize(stderr))
-        raise ExtensionError("Timeout({0}): {1}\n{2}".format(timeout, cmd, msg))
+        return_code = process.poll()
```
boumenot: I am surprised that this did not caused any tests to change.  Why?

hglkrijger: Because `process.poll()` will return `None` here, essentially matching the previous expected behavior. The one test we have does not fork, and tests that case. If we agree on this approach, I can add tests for the negative (forked) cases.

hglkrijger: I went ahead and added a new test to illustrate.

escogom:
Hello - what ever happened after this?  This is still a problem for my team (we are using CustomScript extension (https://github.com/Azure/custom-script-extension-linux/issues/106)

PR #1272, opened 2018-07-24
Make provisioning faster

andyliuliming:
1. make some tasks in the agent run in parallel to speed up the provisioning.

2. adjust the interval of the checking period to speed up the provisioning.

### DIFF
--- a/azurelinuxagent/common/logger.py
+++ b/azurelinuxagent/common/logger.py
@@ -17,8 +17,9 @@
 """
 Log utils
 """
+import os
 import sys
-
+import threading
 from azurelinuxagent.common.future import ustr
 from datetime import datetime, timedelta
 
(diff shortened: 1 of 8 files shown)


azurelinuxagent/common/protocol/util.py
```diff
@@ -144,50 +160,75 @@ def _set_wireserver_endpoint(self, endpoint):
...
-                self.dhcp_handler.run()
-            except DhcpError as e:
-                raise ProtocolError(ustr(e))
+        """
+        detect the wire protocol.
+        """
+        self.wire_lock.acquire()
+        try:
+            if self.wire_protocol != None:
+                return self.wire_protocol
```
jasonzio: Use `if x is not None`; do not use `if x != None`. PEP-8 compliance, proper use of python, etc.

andyliuliming:
@jasonzio tested only on the ubuntu xenial and our stemcell(CF) for now. I have dungen crawler environment, could you please tell me which scenario to run for the full test?

azurelinuxagent/common/logger.py
```diff
@@ -77,11 +78,12 @@ def log(self, level, msg_format, *args):
...
         time = datetime.now().strftime(u'%Y/%m/%d %H:%M:%S.%f')
         level_str = LogLevel.STRINGS[level]
+        thread_name = threading.currentThread().name
         if self.prefix is not None:
-            log_item = u"{0} {1} {2} {3}\n".format(time, level_str, self.prefix,
+            log_item = u"{0} {1} {2} {3} {4}\n".format(time, thread_name, level_str, self.prefix,
                                                    msg)
         else:
-            log_item = u"{0} {1} {2}\n".format(time, level_str, msg)
+            log_item = u"{0} {1} {2} {3}\n".format(time, thread_name, level_str, msg)
```
hglkrijger: Fyi, these changes will break some of the dungeon crawler tests, which expect a specific log format

andyliuliming:
@hglkrijger Thanks for the reviewing.
I will seperate the clean up with the "parallel" change, and send out the clean up pr first :)
and try to run the full bvt, and see the time metrics change.

hglkrijger:
Closing this, since you opened a new PR

PR #1273, opened 2018-07-25
Update child process handling and capture

### LINKED ISSUE #1270: Agent Prematurely Terminates Long Running Extensions
A user opened an issue against [Azure/custom-script-extension-linux #139](https://github.com/Azure/custom-script-extension-linux/issues/139) about the extension prematurely timing out.  I have verified this issue against 2.2.29.

```sh
az vm extension set -g my-resource-group --vm-name vm1 --publisher Microsoft.Azure.Extensions --name CustomScript --version 2.0 --settings "{ \"commandToExecute\": \"sleep 360; ls\" }"
```

The agent will kill the process after five minutes, which is different from previous behavior.  The logs for the agent claim that it moved the extension PID into a cGroup, but the PID it referenced in the logs is the PID *agent ext-handler process* not the extension.

hglkrijger:
Previously, we used `poll()` to determine whether a child is active. Unfortunately that usage was flawed, and for any forked process this will always return a value. In other words our timeout enforcement has never really worked.

The last set of changes to use `communicate(timeout)` and `os.kill(pid, 0)` do actually enforce the timeout correctly, but this is a breaking change in behavior, even though it is in the extension contract. As such we have to revert it.

For processes which are non-forking, we capture the process output. For processes which do fork, it is possible to capture the output in an asynchronous manner, but I will leave that for a future PR.

- fixes #1270

### DIFF
--- a/azurelinuxagent/common/utils/processutil.py
+++ b/azurelinuxagent/common/utils/processutil.py
@@ -16,7 +16,7 @@
 #
 # Requires Python 2.6+ and Openssl 1.0+
 #
-
+import multiprocessing
 import subprocess
 import sys
 import os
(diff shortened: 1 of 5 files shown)


boumenot:
Sync'ed offline.  LGTM.

PR #1277, opened 2018-07-26
Redacting SAS tokens while logging

### LINKED ISSUE #1213: Waagent Logs SAS Keys
SAS keys can be found in waagent.log, for example when an HTTP error occurs the full request including SAS key is logged. These should be removed and not logged.

vrdmr:
Added the fix to check the URL format, and if it has SAS tokens, then replace the signature with `<SAS SIGNATURE>` while logging. Resolves #1213.

### DIFF
--- a/azurelinuxagent/common/utils/restutil.py
+++ b/azurelinuxagent/common/utils/restutil.py
@@ -18,6 +18,7 @@
 #
 
 import os
+import re
 import threading
 import time
 import traceback
@@ -37,6 +38,9 @@ DELAY_IN_SECONDS = 1
 
 THROTTLE_RETRIES = 25
 THROTTLE_DELAY_IN_SECONDS = 1
+
+REDACTED_TEXT = "<SAS_SIGNATURE>"
+SAS_TOKEN_RETRIEVAL_REGEX = re.compile(r'^(https?://[a-zA-Z0-9.].*sig=)([a-zA-Z0-9%-]*)(.*)$')
 
 RETRY_CODES = [
     httpclient.RESET_CONTENT,
@@ -170,6 +174,10 @@ def _get_http_proxy(secure=False):
     return host, port
 
 
+def redact_sas_tokens_in_urls(url):
+    return SAS_TOKEN_RETRIEVAL_REGEX.sub(r"\1" + REDACTED_TEXT + r"\3", url)
+
+
 def _http_request(method, host, rel_uri, port=None, data=None, secure=False,
                   headers=None, proxy_host=None, proxy_port=None):
 
@@ -205,7 +213,7 @@ def _http_request(method, host, rel_uri, port=None, data=None, secure=False,
 
     logger.verbose("HTTP connection [{0}] [{1}] [{2}] [{3}]",
                    method,
-                   url,
+                   redact_sas_tokens_in_urls(url),
                    data,
                    headers)
 
@@ -316,7 +324,7 @@ def http_request(method,
 
         except httpclient.HTTPException as e:
             msg = '[HTTP Failed] {0} {1} -- HttpException {2}'.format(
-                method, url, e)
+                method, redact_sas_tokens_in_urls(url), e)
             if _is_retry_exception(e):
                 continue
             break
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/utils/restutil.py
```diff
@@ -170,6 +171,14 @@ def _get_http_proxy(secure=False):
     return host, port
 
 
+def cleanup_sas_tokens_from_urls(url):
+    redacted_text = "<SAS SIGNATURE>"
```
hglkrijger: Might be worth putting this in a constant somewhere, since it is referenced in the tests

azurelinuxagent/common/utils/restutil.py
```diff
@@ -170,6 +171,14 @@ def _get_http_proxy(secure=False):
...
+    regexp = re.compile(r'(^http[s]:/+[/a-zA-Z0-9.]*?.*&*sig=)([a-zA-Z0-9%]*)(.*&*$)')
```
hglkrijger: You probably want to move this to the class level so it is compiled once

azurelinuxagent/common/utils/restutil.py
```diff
@@ -205,7 +214,7 @@ def _http_request(method, host, rel_uri, port=None, data=None, secure=False,
 
     logger.verbose("HTTP connection [{0}] [{1}] [{2}] [{3}]",
                    method,
-                   url,
+                   cleanup_sas_tokens_from_urls(url),
```
hglkrijger: Is this the only place we log url?

tests/utils/test_rest_util.py
```diff
@@ -115,6 +115,78 @@ def test_parse_url(self):
...
+        filtered_url = restutil.cleanup_sas_tokens_from_urls(test_url)
+        self.assertEquals(redacted_test_url, filtered_url)
+
+        test_url = "https://vamerudevlinuxdiag152.blob.core.windows.net/?sv=2017-11-09&ss=b&srt=o&sp=r&se=2018-07" \
+                   "-26T02:20:42Z&st=2018-07-25T18:20:44Z&spr=https&sig=cmluQEHnOGsVK9NDm83ruuPdPWNQcerfjOAbkspNZXU%3D"
+        redacted_test_url = "https://vamerudevlinuxdiag152.blob.core.windows.net/?sv=2017-11-09&ss=b&srt=o&sp=r&se" \
+                            "=2018-07-26T02:20:42Z&st=2018-07-25T18:20:44Z&spr=https&sig" \
+                            "="+ redacted_text
+        filtered_url = restutil.cleanup_sas_tokens_from_urls(test_url)
+        self.assertEquals(redacted_test_url, filtered_url)
```
hglkrijger: Good tests, bit hard to read though

azurelinuxagent/common/utils/restutil.py
vrdmr: This is where we explicitly log. Added the cleanup action in a couple of more places where exceptions are raised.

vrdmr: Added in https://github.com/Azure/WALinuxAgent/pull/1277/files#diff-aa566f84668a2b9beb5c97dbcefda221R329 and https://github.com/Azure/WALinuxAgent/pull/1277/files#diff-aa566f84668a2b9beb5c97dbcefda221R337 as well.

hglkrijger: Thanks

azurelinuxagent/common/utils/restutil.py
```diff
@@ -170,6 +174,12 @@ def _get_http_proxy(secure=False):
     return host, port
 
 
+def cleanup_sas_tokens_from_urls(url):
+    if SAS_TOKEN_RETRIEVAL_REGEX.search(url):
+        return SAS_TOKEN_RETRIEVAL_REGEX.sub(r"\1" + REDACTED_TEXT + r"\3", url)
```
jasonzio: There's no need to call re.search() and re.sub(). From the re.sub() documentation:

> Return the string obtained by replacing the leftmost non-overlapping occurrences of pattern in string by the replacement repl. If the pattern isn’t found, string is returned unchanged.

Delete the `if` line; just return what SAS_TOKEN_RETRIEVAL_REGEX.sub() returns.

azurelinuxagent/common/utils/restutil.py
```diff
@@ -170,6 +174,12 @@ def _get_http_proxy(secure=False):
     return host, port
 
 
+def cleanup_sas_tokens_from_urls(url):
```
jasonzio: Preferred name: `redact_sas_tokens_in_urls`. The word "cleanup" could mean many things, but the intent is to do exactly this one thing.

vrdmr:
@jasonzio Thanks for the review. Made the changes.

PR #1266, opened 2018-07-19
Updating the osinfo to list

vrdmr:
The issue seen when running on a Mac is

```
======================================================================
ERROR: Failure: TypeError ('tuple' object does not support item assignment)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/varadmeru/work/microsoft/github/WALinuxAgent/venv2.7/lib/python2.7/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/Users/varadmeru/work/microsoft/github/WALinuxAgent/venv2.7/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/Users/varadmeru/work/microsoft/github/WALinuxAgent/venv2.7/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/Users/varadmeru/work/microsoft/github/WALinuxAgent/tests/test_import.py", line 1, in <module>
    from tests.tools import *
  File "/Users/varadmeru/work/microsoft/github/WALinuxAgent/tests/tools.py", line 31, in <module>
    import azurelinuxagent.common.event as event
  File "/Users/varadmeru/work/microsoft/github/WALinuxAgent/azurelinuxagent/common/event.py", line 33, in <module>
    from azurelinuxagent.common.protocol.restapi import TelemetryEventParam, \
  File "/Users/varadmeru/work/microsoft/github/WALinuxAgent/azurelinuxagent/common/protocol/__init__.py", line 18, in <module>
    from azurelinuxagent.common.protocol.util import get_protocol_util, \
  File "/Users/varadmeru/work/microsoft/github/WALinuxAgent/azurelinuxagent/common/protocol/util.py", line 34, in <module>
    from azurelinuxagent.common.osutil import get_osutil
  File "/Users/varadmeru/work/microsoft/github/WALinuxAgent/azurelinuxagent/common/osutil/__init__.py", line 18, in <module>
    from azurelinuxagent.common.osutil.factory import get_osutil
  File "/Users/varadmeru/work/microsoft/github/WALinuxAgent/azurelinuxagent/common/osutil/factory.py", line 20, in <module>
    from azurelinuxagent.common.version import *
  File "/Users/varadmeru/work/microsoft/github/WALinuxAgent/azurelinuxagent/common/version.py", line 135, in <module>
    __distro__ = get_distro()
[...]

PR #1281, opened 2018-07-31
Delete JIT accounts

matkin-msft:
Delete flow. The intention here was to also send status back up through the wire service as something exposed to the fabric controller, but after talking with the Host Agent team, we'll need to make some changes their first so that will be implemented at a later date.

### DIFF
--- a/azurelinuxagent/ga/remoteaccess.py
+++ b/azurelinuxagent/ga/remoteaccess.py
@@ -46,7 +46,7 @@ from pwd import getpwall
 from azurelinuxagent.common.errorstate import ErrorState
 
 from azurelinuxagent.common.event import add_event, WALAEventOperation, elapsed_milliseconds
-from azurelinuxagent.common.exception import ExtensionError, ProtocolError
+from azurelinuxagent.common.exception import ExtensionError, ProtocolError, RemoteAccessError
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.protocol.restapi import ExtHandlerStatus, \
                                                     ExtensionStatus, \
@@ -80,6 +80,7 @@ class RemoteAccessHandler(object):
         self.cryptUtil = CryptUtil(conf.get_openssl_cmd())
         self.remote_access = None
         self.incarnation = 0
+        self.error_message = ""
 
     def run(self):
         try:
@@ -90,8 +91,7 @@ class RemoteAccessHandler(object):
                     # something changed. Handle remote access if any.
                     self.incarnation = current_incarnation
                     self.remote_access = self.protocol.client.get_remote_access()
-                    if self.remote_access is not None:
-                        self.handle_remote_access()
+                    self.handle_remote_access()
         except Exception as e:
             msg = u"Exception processing remote access handler: {0} {1}".format(ustr(e), traceback.format_exc())
             logger.error(msg)
@@ -102,19 +102,43 @@ class RemoteAccessHandler(object):
                       message=msg)
 
     def handle_remote_access(self):
+        # Get JIT user accounts.
+        all_users = self.os_util.get_users()
+        existing_jit_users = [u[0] for u in all_users if self.validate_jit_user(u[4])]
+        self.err_message = ""
         if self.remote_access is not None:
-            # Get JIT user accounts.
-            all_users = self.os_util.get_users()
-            jit_users = set()
-            for usr in all_users:
-                if self.validate_jit_user(usr[4]):
-                    jit_users.add(usr[0])
             for acc in self.remote_access.user_list.users:
-                raw_expiration = acc.expiration
-                account_expiration = datetime.strptime(raw_expiration, REMOTE_USR_EXPIRATION_FORMAT)
-                now = datetime.utcnow()
-                if acc.name not in jit_users and now < account_expiration:
-                    self.add_user(acc.name, acc.encrypted_password, account_expiration)
+                try:
+                    raw_expiration = acc.expiration
+                    account_expiration = datetime.strptime(raw_expiration, REMOTE_USR_EXPIRATION_FORMAT)
+                    now = datetime.utcnow()
+                    if acc.name not in existing_jit_users and now < account_expiration:
+                        self.add_user(acc.name, acc.encrypted_password, account_expiration)
+                    elif acc.name in existing_jit_users and now > account_expiration:
+                        # user account expired, delete it.
+                        logger.info("user {0} expired from remote_access".format(acc.name))
+                        self.remove_user(acc.name)
+                except RemoteAccessError as rae:
+                    self.err_message = self.err_message + "Error processing user {0}. Exception: {1}"\
+                        .format(acc.name, ustr(rae))
+            for user in existing_jit_users:
+                try:
+                    if user not in [u.name for u in self.remote_access.user_list.users]:
+                        # user explicitly removed
+                        logger.info("User {0} removed from remote_access".format(user))
+                        self.remove_user(user)
+                except RemoteAccessError as rae:
+                    self.err_message = self.err_message + "Error removing user {0}. Exception: {1}"\
+                        .format(user, ustr(rae))
+        else:
+            # All users removed, remove any remaining JIT accounts.
+            for user in existing_jit_users:
+                try:
+                    logger.info("User {0} removed from remote_access. remote_access empty".format(user))
+                    self.remove_user(user)
+                except RemoteAccessError as rae:
+                    self.err_message = self.err_message + "Error removing user {0}. Exception: {1}"\
+                        .format(user, ustr(rae))
 
     def validate_jit_user(self, comment):
         return comment == REMOTE_ACCESS_ACCOUNT_COMMENT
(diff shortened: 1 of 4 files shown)


azurelinuxagent/ga/remoteaccess.py
```diff
@@ -102,60 +102,78 @@ def run(self):
...
-            return
-        except OSError as oe:
-            self.handle_failed_create(username, oe.strerror)
+            logger.info("User '{0}' added successfully with expiration in {1}".format(username, expiration_date))
         except Exception as e:
-            self.handle_failed_create(username, ustr(e))
+            try:
+                self.handle_failed_create(username)
+            except RemoteAccessError as rae:
+                raise RemoteAccessError("Error adding user {0} and error cleaning up {1}".format(username, ustr(e)), rae)
```
hglkrijger: I might suggest simplifying this to a single raise, and setting flags based on what was successful

tests/ga/test_remoteaccess_handler.py
```diff
@@ -98,14 +121,21 @@ def test_add_user_bad_password_data(self, _1, _2, _3):
         tstuser = "foobar"
         expiration_date = datetime.utcnow() + timedelta(days=1)
         pwd = tstpassword
-        rah.add_user(tstuser, pwd, expiration_date)
+        error = "Error adding user {0} cleanup successful\nInner error: test exception for bad password".format(tstuser)
+        if is_python_version_36():
```
hglkrijger: Not a huge fan of this type of check. We have started adding version agnostic helpers to `tests/tools.py`, for example see #1266 which adds `assertListEqual` if missing. Would you mind doing the same for `assertRaisesRegex`?

matkin-msft: I actually already did this, but being that with Python 2.6 doesn't support the assert regex so that part is skipped, I thought this was better since it clearly indicates to anyone lookin in the future what's going on.

---
In reply to: [207063999](https://github.com/Azure/WALinuxAgent/pull/1281#discussion_r207063999) [](ancestors = 207063999)

matkin-msft: ... never mind, I can just do the regex myself.  Done.

---
In reply to: [207063999](https://github.com/Azure/WALinuxAgent/pull/1281#discussion_r207063999) [](ancestors = 207063999)

azurelinuxagent/ga/remoteaccess.py
```diff
@@ -102,60 +102,78 @@ def run(self):
                       message=msg)
 
     def handle_remote_access(self):
+        # Get JIT user accounts.
+        all_users = self.os_util.get_users()
+        existing_jit_users = [u[0] for u in all_users if self.validate_jit_user(u[4])]
```
jasonzio: This should be a set instead of a list. You check to see if names are in `existing_jit_users` in multiple places, and the one time you iterate the collection, the order doesn't appear to be relevant.

azurelinuxagent/ga/remoteaccess.py
```diff
@@ -102,60 +102,78 @@ def run(self):
...
+                    elif acc.name in existing_jit_users and now > account_expiration:
+                        # user account expired, delete it.
+                        logger.info("user {0} expired from remote_access".format(acc.name))
+                        self.remove_user(acc.name)
+                except RemoteAccessError as rae:
+                    self.err_message = self.err_message + "Error processing user {0}. Exception: {1}"\
+                        .format(acc.name, ustr(rae))
+            for user in existing_jit_users:
+                try:
+                    if user not in [u.name for u in self.remote_access.user_list.users]:
```
jasonzio: Can you hoist the construction of this list (`[u.name for u in self.remote_access.user_list.users]`) outside the loop? If `self.remove_user(user)` alters `self.remote_access.user_list.users`, then recompute the object when you remove a user.

tests/tools.py
```diff
@@ -163,6 +166,24 @@ def emulate_assertIsNotNone(self, x, msg=None):
             msg = msg if msg is not None else '{0} is None'.format(_safe_repr(x))
             self.fail(msg)
 
+    def emulate_raises_regex(self, exception_type, regex, function, *args, **kwargs):
+        if PY_VERSION_MAJOR == 3 and PY_VERSION_MINOR == 6:
+            with self.assertRaisesRegex(exception_type, regex):
+                function(*args)
```
jasonzio: If you take **kwargs, then you should pass it to the invoked function.

tests/tools.py
```diff
@@ -163,6 +166,24 @@ def emulate_assertIsNotNone(self, x, msg=None):
...
+        elif PY_VERSION_MAJOR == 3 and PY_VERSION_MINOR == 7:
+            with self.assertRaisesRegexp(exception_type, regex):
+                function(*args)
+        else:
+            try:
+                function(*args)
```
jasonzio: If you take **kwargs, then you should pass it to the invoked function.

tests/tools.py
```diff
@@ -163,6 +166,24 @@ def emulate_assertIsNotNone(self, x, msg=None):
             msg = msg if msg is not None else '{0} is None'.format(_safe_repr(x))
             self.fail(msg)
 
+    def emulate_raises_regex(self, exception_type, regex, function, *args, **kwargs):
+        if PY_VERSION_MAJOR == 3 and PY_VERSION_MINOR == 6:
```
jasonzio: But I thought Python 3.6 and higher already implemented assertRaisesRegex? If that's the case, then emulate_raises_regex will never be called (because it won't be patched into cls), so there's no need for this code at all.

I am assuming that the signature for assertRaisesRegex is unchanged in 3.6 and 3.7, and that emulate_raises_regex implements exactly that signature.

tests/tools.py
```diff
@@ -163,6 +166,24 @@ def emulate_assertIsNotNone(self, x, msg=None):
...
+            with self.assertRaisesRegex(exception_type, regex):
+                function(*args)
+        elif PY_VERSION_MAJOR == 3 and PY_VERSION_MINOR == 7:
+            with self.assertRaisesRegexp(exception_type, regex):
+                function(*args)
```
jasonzio: If you take **kwargs, then you should pass it to the invoked function.

tests/tools.py
matkin-msft: Sorry, I mucked with the version numbers while testing and forgot to change it back.  It's 3.6 vs 2.7.  I've removed the code for 3.6 as it's not needed as you said.  That said, the versions still aren't correct and I'm going to fix that though I don't have every version of python to test, so I'm going to go off the documentation.

---
In reply to: [207088817](https://github.com/Azure/WALinuxAgent/pull/1281#discussion_r207088817) [](ancestors = 207088817)

matkin-msft: Ok, fixed the versions based on what I've read.  assertRaisesRegexp was introduced in 2.7 and was replaced by assertRaisesRegex in 3.2.  I'm not going to guess at python 4, but technically this method would still work.

---
In reply to: [207311754](https://github.com/Azure/WALinuxAgent/pull/1281#discussion_r207311754) [](ancestors = 207311754,207088817)

tests/tools.py
```diff
@@ -163,6 +166,21 @@ def emulate_assertIsNotNone(self, x, msg=None):
             msg = msg if msg is not None else '{0} is None'.format(_safe_repr(x))
             self.fail(msg)
 
+    def emulate_raises_regex(self, exception_type, regex, function, *args, **kwargs):
+        if (PY_VERSION_MAJOR == 2 and PY_VERSION_MINOR >= 7) or (PY_VERSION_MAJOR == 3 and PY_VERSION_MINOR >= 2):
```
jasonzio: Don't test for the version number here; test for the availability of `assertRaisesRegexp`, the thing you'd like to use. That is...
```
if hasattr(cls, "assertRaisesRegexp"):
    with self.assertRaisesRegexp(exception_type, regex):
        function(*args, **kwargs)
else:
    try:
        function(*args, **kwargs)
    except
etc
```

matkin-msft: Ahh, fixed it a little different above, but same thing.

---
In reply to: [207340686](https://github.com/Azure/WALinuxAgent/pull/1281#discussion_r207340686) [](ancestors = 207340686)

jasonzio:
I thought about your solution. It works, and I'm not going to request a change, but it's not the best choice. Any newly-written tests should avoid assertRaisesRegexp, since that was a dead-end tool that was replaced by assertRaiseRegex. Adding an implementation of the deprecated method might be seen as encouraging its use; more accurately, deliberately leaving the deprecated method *unimplemented* would be seen as discouraging its use in favor of the newer approach.

As I said, what you've done is acceptable, and there's no need to change it.

matkin-msft:
Not sure I follow.  It shouldn't encourage usage of assertRaisesRegexp in the tests... at least no more than before this change was made.  If you use assertRaisesRegexp it will still fail when runing python 3.2 or later.  This change just says if assertRaisesRegexp exists (then we know you're on 2.7-3.1) so create a new method assertRaisesRegex that just happens to use assertRaisesRegexp

---
In reply to: [410066544](https://github.com/Azure/WALinuxAgent/pull/1281#issuecomment-410066544) [](ancestors = 410066544)

PR #1283, opened 2018-08-01
Fix for test_isloopback

narrieta:
test_isloopback assumes there is a network interface named 'eth0'; this is not true on my system.

Instead, I am enumerating all the interfaces and running the test on all of them.

### DIFF
--- a/tests/common/osutil/test_default.py
+++ b/tests/common/osutil/test_default.py
@@ -190,8 +190,11 @@ class TestOSUtil(AgentTestCase):
                 self.assertEqual(('', ''), osutil.DefaultOSUtil().get_first_if())
 
     def test_isloopback(self):
-        self.assertTrue(osutil.DefaultOSUtil().is_loopback('lo'))
-        self.assertFalse(osutil.DefaultOSUtil().is_loopback('eth0'))
+        for iface in osutil.DefaultOSUtil()._get_all_interfaces():
+            if iface == 'lo':
+                self.assertTrue(osutil.DefaultOSUtil().is_loopback(iface))
+            else:
+                self.assertFalse(osutil.DefaultOSUtil().is_loopback(iface))
 
     def test_isprimary(self):
         routing_table = "\
--- a/tests/common/osutil/test_default.py
+++ b/tests/common/osutil/test_default.py
@@ -199,7 +199,7 @@ class TestOSUtil(AgentTestCase):
             else:
                 non_loopback_count += 1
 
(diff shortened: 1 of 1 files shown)


tests/common/osutil/test_default.py
```diff
@@ -190,8 +190,11 @@ def test_get_first_if_all_loopback(self):
                 self.assertEqual(('', ''), osutil.DefaultOSUtil().get_first_if())
 
     def test_isloopback(self):
-        self.assertTrue(osutil.DefaultOSUtil().is_loopback('lo'))
-        self.assertFalse(osutil.DefaultOSUtil().is_loopback('eth0'))
+        for iface in osutil.DefaultOSUtil()._get_all_interfaces():
+            if iface == 'lo':
+                self.assertTrue(osutil.DefaultOSUtil().is_loopback(iface))
+            else:
+                self.assertFalse(osutil.DefaultOSUtil().is_loopback(iface))
```
jasonzio: Test code should count the number of assertTrue and assertFalse that are actually executed at runtime. If the assertTrue is never fired, the test code should force a failure, because that indicates the testing environment itself is flawed. (If the system-under-test lacks a loopback interface, then the test_isloopback test didn't actually test anything, and the developer needs to know that.)

When unit tests are run under DCR, this is the best way to find a flawed VM added to the DCR fleet.

narrieta: Yes, thanks. That sounds like a good test for _get_all_interfaces. I can add one asserting that it returns exactly 1 loopback and 1 or more non-loopback interfaces. Sounds good?

jasonzio: SGTM, thanks.

narrieta: OK, it's added now.

tests/common/osutil/test_default.py
```diff
@@ -189,6 +189,19 @@ def test_get_first_if_all_loopback(self):
...
+        loopback_count = 0
+        non_loopback_count = 0
+
+        for iface in osutil.DefaultOSUtil()._get_all_interfaces():
+            if iface == 'lo':
+                loopback_count += 1
+            else:
+                non_loopback_count += 1
+
+        self.assertEqual(loopback_count, 1, 'Only 1 loopback network interface should exist')
```
jasonzio: "Exactly", not "Only".

narrieta: Yup; fixed.

Issue #838, opened 2017-08-10
Initial extensionless support


hglkrijger:
Before cloud-init changes are available in each distro, we should be able to respond to the `ProvisionGuestAgent` flag, which is passed in via `ovf-env.xml`:
```
<MediaContent ... >
	<LinuxProvisioningConfigurationSet>
	...  
	</LinuxProvisioningConfigurationSet>
	<PlatformSettings>
		...
		<ProvisionGuestAgent>false</ProvisionGuestAgent>
		...
	</PlatformSettings>
</MediaContent>
```
When this is false, the agent should disable extension support after initial provisioning is completed, and emit informational messages in the log.

hglkrijger:
Fix merged

PR #1289, opened 2018-08-07
Agentless

hglkrijger:
## Description

Addresses #838

This introduces two new modes of operation for the agent:
- extensions disabled
- agent disabled

### Extensions disabled

This allows an end user to reduce the agent overhead when extensions are not required. This serves customers who have provisioned VMs already and cannot re-provision without the agent. To enable this mode, a new config option `Extensions.Enabled=y|n` is provided, which the user may set. 

When `y` (default), the agent runs without any changes.

When `n`, the goal state interval is reduced to 5 minutes, and extension configuration is not processed. This should help customers who do not want to run the agent, but still need the VM to report status, handle JIT requests, and so on.

### Agent disabled

As of Azure.Compute 2018-06-01, a new option is provided in the ARM API to disable the Linux guest agent completely. Documentation on that is forthcoming. This API will end up setting the `ProvisionGuestAgent` flag in `ovf-env.xml`, which the agent will honor with this change. 

If that flag is set to `false`, the agent will write `/var/lib/waagent/disable_agent`, and if that file exists, sleep forever. No status reports or goal state handling of any kind are processed. 

Once cloud-init supports this flag, the agent service will be disabled and/or removed on provision.

---

### PR information


azurelinuxagent/daemon/main.py
```diff
@@ -99,6 +99,17 @@ def check_pid(self):
 
         fileutil.write_file(pid_file, ustr(os.getpid()))
 
+    def sleep_if_disabled(self):
+        agent_disabled_file_path = conf.get_disable_agent_file_path()
+        if os.path.exists(agent_disabled_file_path):
+            import threading
+            logger.warn("Disabling the guest agent by sleeping forever; "
```
boumenot: There's no API to disable the agent, there's just a file on disk?

I assumed we would just stop the service, and no code would be running.

tests/daemon/test_daemon.py
```diff
@@ -75,13 +78,58 @@ def test_set_openssl_fips(self, mock_conf, mock_daemon):
...
+
+        self.assertEqual(1, patch_run_provision.call_count)
+        self.assertEqual(1, patch_run_latest.call_count)
+
+    @patch('azurelinuxagent.ga.update.UpdateHandler.run_latest', side_effect=AgentTestCase.fail)
+    @patch('azurelinuxagent.pa.provision.default.ProvisionHandler.run', side_effect=ProvisionHandler.write_agent_disabled)
+    @patch('azurelinuxagent.pa.provision.get_provision_handler', return_value=ProvisionHandler())
+    def test_daemon_agent_disabled(self, _, __, patch_run_latest):
+        """
+        Agent should provision, then sleep forever when disable_agent is found
```
boumenot: Why do we start the agent at all?

tests/pa/test_provision.py
```diff
@@ -188,7 +269,48 @@ def test_provision_telemetry_fail(self, mock_util, distro_name,
...
+        self.assertEqual(3, patch_write_agent_disabled.call_count)
+
+        ph.handle_provision_guest_agent(provision_guest_agent='true')
+        self.assertEqual(3, patch_write_agent_disabled.call_count)
+
+        ph.handle_provision_guest_agent(provision_guest_agent='True')
+        self.assertEqual(3, patch_write_agent_disabled.call_count)
+
+        ph.handle_provision_guest_agent(provision_guest_agent='TRUE')
+        self.assertEqual(3, patch_write_agent_disabled.call_count)
```
boumenot: Thorough! 💯

README.md
```diff
@@ -213,6 +214,18 @@ may be used for some string type configuration entries as detailed below.
 
 ### Configuration File Options
 
+#### __Extensions.Enabled__
+
+_Type: Boolean_
```
boumenot: Please add this to the various waagent.conf versions in the tree.  I think people are more likely to read that file than they are the web site.

hglkrijger: Good call, updated

azurelinuxagent/daemon/main.py
hglkrijger: The API is in CRP, exposed through ARM/SDK/CLI. Cloud-init will honor this flag by stopping/disabling the service, but for distros without cloud-init support, I believe this is the best we can do.

tests/daemon/test_daemon.py
hglkrijger: We do not start it, the system does. If we just exit the service will be restarted. I definitely do not want to get into the business of trying to update the service from a running agent.

PR #1298, opened 2018-08-13
Implementing extension sequencing in azure linux agent

sirajudeens:
## Adding support for extension sequencing in the azure linux agent

Issue # 
VMSS extensions can be installed and enabled in a sequence by specifying dependsOn property values in the ARM template. CRP processes the dependency data, builds dependency graph and generates dependency level for each extension. The dependency level information is passed on to the VM agent. The VM agent needs to parse the dependency data and install the extensions based on the dependency level.

---

### PR information

### DIFF
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -1591,11 +1591,6 @@ class ExtensionsConfig(object):
         ext_handler.properties.version = getattrib(plugin, "version")
         ext_handler.properties.state = getattrib(plugin, "state")
 
-        try:
-            ext_handler.properties.dependencyLevel = int(getattrib(plugin, "dependencyLevel"))
-        except ValueError:
-            ext_handler.properties.dependencyLevel = 0
-
         location = getattrib(plugin, "location")
         failover_location = getattrib(plugin, "failoverlocation")
         for uri in [location, failover_location]:
@@ -1627,6 +1622,26 @@ class ExtensionsConfig(object):
             logger.error("Invalid extension settings")
             return
 
+        depends_on = None
+        depends_on_node = find(settings[0], "DependsOn")
+
+        depends_on_name = getattrib(depends_on_node, "name")
+        if depends_on_name == "":
+            depends_on_name = ext_handler.name
+
+        try:
+            depends_on_level = int(getattrib(depends_on_node, "dependencyLevel"))
+        except (ValueError, TypeError):
+            depends_on_level = 0
+
+        dependencies = []
+        for dependency in findall(depends_on_node, "DependsOnExtension"):
+            handler_name = gettext(dependency)
+            ext_name = getattrib(dependency, "name")
+            if ext_name == "":
+                ext_name = handler_name
+            dependencies.append((handler_name, ext_name))
+
         for plugin_settings_list in runtime_settings["runtimeSettings"]:
             handler_settings = plugin_settings_list["handlerSettings"]
             ext = Extension()
@@ -1636,6 +1651,9 @@ class ExtensionsConfig(object):
             ext.sequenceNumber = seqNo
             ext.publicSettings = handler_settings.get("publicSettings")
             ext.protectedSettings = handler_settings.get("protectedSettings")
+            if ext.name == depends_on_name:
+                ext.dependencies = dependencies
+                ext.dependencyLevel = depends_on_level
             thumbprint = handler_settings.get(
                 "protectedSettingsCertThumbprint")
             ext.certificateThumbprint = thumbprint
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -1530,21 +1530,6 @@ class Certificates(object):
         return file_name
 
 
-class Dependency(object):
-    """
-    Represents an extension upon which another extension is dependent
-    handler - the ExtHandler of the dependency
-    exts - the Extension(s) of the dependency
-    timeout - the timeout for the extension to get to the Ready state
-              timeout is expressed in minutes.
-    """
-
-    def __init__(self, handler=None, exts=None, timeout=None):
-        self.handler = handler
-        self.exts = exts
-        self.timeout = timeout
-
-
 class ExtensionsConfig(object):
     """
     parse ExtensionsConfig, downloading and unpacking them to /var/lib/waagent.
@@ -1593,18 +1578,7 @@ class ExtensionsConfig(object):
 
         for ext_handler in self.ext_handlers.extHandlers:
             for extension in ext_handler.properties.extensions:
-                resolved_dependencies = []
-                for (handler_name, ext_name) in extension.dependencies:
-                    # Gracefully handle garbage cases where handler name does
-                    # not exist or is not unique. Same for the extension name.
-                    handlers = [handler for handler in self.ext_handlers.extHandlers
-                                if handler_name == handler.name]
-                    for handler in handlers:
-                        exts = [e for e in handler.properties.extensions
-                                if ext_name == e.name]
-                        resolved_dependencies.append(
-                            Dependency(handler=handler, exts=exts))
-                extension.dependencies = resolved_dependencies
+                extension.resolve_dependencies(self.ext_handlers.extHandlers)
 
         self.status_upload_blob = findtext(xml_doc, "StatusUploadBlob")
         self.artifacts_profile_blob = findtext(xml_doc, "InVMArtifactsProfileBlob")
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -1622,12 +1622,14 @@ class ExtensionsConfig(object):
             logger.error("Invalid extension settings")
             return
 
+        depends_on_level = 0
         depends_on_node = find(settings[0], "DependsOn")
-
-        try:
-            depends_on_level = int(getattrib(depends_on_node, "dependencyLevel"))
-        except (ValueError, TypeError):
-            depends_on_level = 0
+        if depends_on_node != None:
+            try:
+                depends_on_level = int(getattrib(depends_on_node, "dependencyLevel"))
+            except (ValueError, TypeError):
+                logger.warn("Could not parse dependencyLevel for handler {0}. Setting it to 0".format(name))
+                depends_on_level = 0
 
         for plugin_settings_list in runtime_settings["runtimeSettings"]:
             handler_settings = plugin_settings_list["handlerSettings"]
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -1576,10 +1576,6 @@ class ExtensionsConfig(object):
             self.ext_handlers.extHandlers.append(ext_handler)
             self.parse_plugin_settings(ext_handler, plugin_settings)
 
-        for ext_handler in self.ext_handlers.extHandlers:
-            for extension in ext_handler.properties.extensions:
-                extension.resolve_dependencies(self.ext_handlers.extHandlers)
-
         self.status_upload_blob = findtext(xml_doc, "StatusUploadBlob")
         self.artifacts_profile_blob = findtext(xml_doc, "InVMArtifactsProfileBlob")
 
@@ -1626,25 +1622,12 @@ class ExtensionsConfig(object):
             logger.error("Invalid extension settings")
             return
 
-        depends_on = None
         depends_on_node = find(settings[0], "DependsOn")
-
-        depends_on_name = getattrib(depends_on_node, "name")
-        if depends_on_name == "":
-            depends_on_name = ext_handler.name
 
         try:
             depends_on_level = int(getattrib(depends_on_node, "dependencyLevel"))
         except (ValueError, TypeError):
             depends_on_level = 0
-
-        dependencies = []
-        for dependency in findall(depends_on_node, "DependsOnExtension"):
-            handler_name = getattrib(dependency, "handler")
-            ext_name = getattrib(dependency, "extension")
-            if ext_name == "":
-                ext_name = handler_name
-            dependencies.append((handler_name, ext_name))
 
         for plugin_settings_list in runtime_settings["runtimeSettings"]:
             handler_settings = plugin_settings_list["handlerSettings"]
@@ -1655,9 +1638,7 @@ class ExtensionsConfig(object):
             ext.sequenceNumber = seqNo
             ext.publicSettings = handler_settings.get("publicSettings")
             ext.protectedSettings = handler_settings.get("protectedSettings")
-            if ext.name == depends_on_name:
-                ext.dependencies = dependencies
-                ext.dependencyLevel = depends_on_level
+            ext.dependencyLevel = depends_on_level
             thumbprint = handler_settings.get(
                 "protectedSettingsCertThumbprint")
             ext.certificateThumbprint = thumbprint
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -1640,8 +1640,8 @@ class ExtensionsConfig(object):
 
         dependencies = []
         for dependency in findall(depends_on_node, "DependsOnExtension"):
-            handler_name = gettext(dependency)
-            ext_name = getattrib(dependency, "name")
+            handler_name = getattrib(dependency, "handler")
+            ext_name = getattrib(dependency, "extension")
             if ext_name == "":
                 ext_name = handler_name
             dependencies.append((handler_name, ext_name))
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -1530,6 +1530,21 @@ class Certificates(object):
         return file_name
 
 
+class Dependency(object):
+    """
+    Represents an extension upon which another extension is dependent
+    handler - the ExtHandler of the dependency
+    exts - the Extension(s) of the dependency
+    timeout - the timeout for the extension to get to the Ready state
+              timeout is expressed in minutes.
+    """
+
+    def __init__(self, handler=None, exts=None, timeout=None):
+        self.handler = handler
+        self.exts = exts
+        self.timeout = timeout
+
+
 class ExtensionsConfig(object):
     """
     parse ExtensionsConfig, downloading and unpacking them to /var/lib/waagent.
@@ -1575,6 +1590,21 @@ class ExtensionsConfig(object):
             ext_handler = self.parse_plugin(plugin)
             self.ext_handlers.extHandlers.append(ext_handler)
             self.parse_plugin_settings(ext_handler, plugin_settings)
+
+        for ext_handler in self.ext_handlers.extHandlers:
+            for extension in ext_handler.properties.extensions:
+                resolved_dependencies = []
+                for (handler_name, ext_name) in extension.dependencies:
+                    # Gracefully handle garbage cases where handler name does
+                    # not exist or is not unique. Same for the extension name.
+                    handlers = [handler for handler in self.ext_handlers.extHandlers
+                                if handler_name == handler.name]
+                    for handler in handlers:
+                        exts = [e for e in handler.properties.extensions
+                                if ext_name == e.name]
+                        resolved_dependencies.append(
+                            Dependency(handler=handler, exts=exts))
+                extension.dependencies = resolved_dependencies
 
         self.status_upload_blob = findtext(xml_doc, "StatusUploadBlob")
         self.artifacts_profile_blob = findtext(xml_doc, "InVMArtifactsProfileBlob")
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -335,9 +335,10 @@ class ExtHandlersHandler(object):
                     time.sleep(10)
                 if (datetime.datetime.utcnow() - begin) > timeout_delta:
                     raise ExtensionError("Timeout waiting for success status "
-                                         "from dependency {}/{} for {}",
-                                         dependency.handler, ext,
-                                         ext_handler.name)
+                                         "from dependency {}/{} for {}"
+                                         "status was: {}".format(
+                                             dependency.handler.name, ext.name,
+                                             ext_handler.name, status))
 
     def handle_ext_handler(self, ext_handler, etag):
         ext_handler_i = ExtHandlerInstance(ext_handler, self.protocol)
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -333,12 +333,12 @@ class ExtHandlersHandler(object):
 
             self.handle_ext_handler(ext_handler, etag)
 
-    '''
-    Check the status of the previous extension installed.
-    Wait until it becomes success or times out.
-    Return True if it is installed successfully. False if timed out.
-    '''
     def wait_for_prev_handler_installation(self, prev_handler, cur_handler, wait_until):
+        '''
+        Check the status of the previous extension installed.
+        Wait until it becomes success or times out.
+        Return True if it is installed successfully. False if timed out.
+        '''
         # No need to wait on anything for the very first extension
         if prev_handler == None:
             return True
@@ -356,11 +356,11 @@ class ExtHandlersHandler(object):
             # on this one can be skipped processing
             if not success_status:
                 msg = "Timeout waiting for success status " \
-                      "from dependency {0}/{1} for {2}" \
+                      "from dependency {0}/{1} for {2} " \
                       "status was: {3}".format(prev_handler.name,
                                                ext.name,
                                                cur_handler.name,
-                                               status)
+                                               status.status if status != None else None)
                 logger.info(msg)
                 add_event(AGENT_NAME,
                           version=CURRENT_VERSION,
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -935,11 +935,25 @@ class ExtHandlerInstance(object):
 
         return ext_status
     
+    def get_ext_handling_status(self, ext):
+        seq_no, ext_status_file = self.get_status_file_path(ext)
+        if seq_no < 0 or ext_status_file is None:
+            return None
+
+        # Missing status file is considered a non-terminal state here
+        # so that extension sequencing can wait until it becomes existing
+        if not os.path.exists(ext_status_file):
+            ext_status = ExtensionStatus(seq_no=seq_no)
+            ext_status.message = u"Failed to get status file for {0}".format(ext.name)
+            ext_status.code = -1
+            ext_status.status = "warning"
+        else:
+            ext_status = self.collect_ext_status(ext)
+
+        return ext_status
+
     def is_ext_handling_complete(self, ext):
-        try:
-            status = self.collect_ext_status(ext)
-        except FileNotFoundError as e:
-            return (False, None)
+        status = self.get_ext_handling_status(ext)
 
         # when there is no status, the handling is complete and return None status
         if status is None:
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -312,52 +312,30 @@ class ExtHandlersHandler(object):
                 return
 
         wait_until = datetime.datetime.utcnow() + datetime.timedelta(minutes=DEFAULT_EXT_TIMEOUT_MINUTES)
-        prev_handler = None
-        ext_seq_required = self.is_ext_seq_required()
+        max_dep_level = max([handler.sort_key() for handler in self.ext_handlers.extHandlers])
 
         self.ext_handlers.extHandlers.sort(key=operator.methodcaller('sort_key'))
         for ext_handler in self.ext_handlers.extHandlers:
-            if ext_seq_required:
-                # Make sure that the previous extension is handled.
-                # If handled successfully, proceed with the current handler.
-                # Otherwise, skip the rest of them.
-                dep_level = ext_handler.sort_key()
-                if dep_level >= 0:
-                    if self.wait_for_prev_handler_completion(prev_handler, wait_until):
-                        prev_handler = ext_handler
-                    else:
-                        break
-
             self.handle_ext_handler(ext_handler, etag)
 
-    def is_ext_seq_required(self):
-        '''
-        If all the extensions have same dependency level
-        It is considered as extension sequencing is not required
-        '''
-        dep_level = None
-        for ext_handler in self.ext_handlers.extHandlers:
-            if dep_level is None:
-                dep_level = ext_handler.sort_key()
-                continue
+            # Wait for the extension installation until it is handled.
+            # This is done for the install and enable. Not for the uninstallation.
+            # If handled successfully, proceed with the current handler.
+            # Otherwise, skip the rest of the extension installation.
+            dep_level = ext_handler.sort_key()
+            if dep_level >= 0 and dep_level < max_dep_level:
+                if not self.wait_for_handler_successful_completion(ext_handler, wait_until):
+                    logger.warn("An extension failed or timed out, will skip processing the rest of the extensions")
+                    break
 
-            if dep_level != ext_handler.sort_key():
-                return True
-
-        return False
-
-    def wait_for_prev_handler_completion(self, prev_handler, wait_until):
+    def wait_for_handler_successful_completion(self, ext_handler, wait_until):
         '''
-        Check the status of the previous extension handled.
+        Check the status of the extension being handled.
         Wait until it has a terminal state or times out.
         Return True if it is handled successfully. False if not.
         '''
-        # No need to wait on anything for the very first extension
-        if prev_handler == None:
-            return True
-
-        handler_i = ExtHandlerInstance(prev_handler, self.protocol)
-        for ext in prev_handler.properties.extensions:
+        handler_i = ExtHandlerInstance(ext_handler, self.protocol)
+        for ext in ext_handler.properties.extensions:
             ext_completed, status = handler_i.is_ext_handling_complete(ext)
 
             # Keep polling for the extension status until it becomes success or times out
@@ -367,9 +345,19 @@ class ExtHandlersHandler(object):
 
             # In case of timeout or terminal error state, we log it and return false
             # so that the extensions waiting on this one can be skipped processing
-            if not ext_completed or status != EXTENSION_STATUS_SUCCESS:
-                msg = "Extension {0} was timedout or status was: {1}".format(ext.name, status)
-                logger.info(msg)
+            if datetime.datetime.utcnow() > wait_until:
+                msg = "Extension {0} did not reach a terminal state within the allowed timeout. Last status was {1}".format(ext.name, status)
+                logger.warn(msg)
+                add_event(AGENT_NAME,
+                          version=CURRENT_VERSION,
+                          op=WALAEventOperation.ExtensionProcessing,
+                          is_success=False,
+                          message=msg)
+                return False
+
+            if status != EXTENSION_STATUS_SUCCESS:
+                msg = "Extension {0} did not succeed. Status was {1}".format(ext.name, status)
+                logger.warn(msg)
                 add_event(AGENT_NAME,
                           version=CURRENT_VERSION,
                           op=WALAEventOperation.ExtensionProcessing,
@@ -943,28 +931,26 @@ class ExtHandlerInstance(object):
         # Missing status file is considered a non-terminal state here
         # so that extension sequencing can wait until it becomes existing
         if not os.path.exists(ext_status_file):
-            ext_status = ExtensionStatus(seq_no=seq_no)
-            ext_status.message = u"Failed to get status file for {0}".format(ext.name)
-            ext_status.code = -1
-            ext_status.status = "warning"
+            status = "warning"
         else:
             ext_status = self.collect_ext_status(ext)
+            status = ext_status.status if ext_status is not None else None
 
-        return ext_status
+        return status
 
     def is_ext_handling_complete(self, ext):
         status = self.get_ext_handling_status(ext)
 
-        # when there is no status, the handling is complete and return None status
+        # when seq < 0 (i.e. no new user settings), the handling is complete and return None status
         if status is None:
             return (True, None)
 
         # If not in terminal state, it is incomplete
-        if status.status not in EXTENSION_TERMINAL_STATUSES:
-            return (False, status.status)
+        if status not in EXTENSION_TERMINAL_STATUSES:
+            return (False, status)
 
         # Extension completed, return its status
-        return (True, status.status)
+        return (True, status)
 
     def report_ext_status(self):
         active_exts = []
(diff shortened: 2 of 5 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -67,6 +67,7 @@
 
 TELEMETRY_MESSAGE_MAX_LEN = 3200
 
+DEFAULT_EXT_TIMEOUT_MINUTES = 90
```
hglkrijger: It looks like this is a per-extension timeout, won't CRP timeout well before this for multiple extensions?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -360,11 +359,38 @@ def handle_ext_handlers(self, etag=None):
...
+        deps_wait_until = None
+        handlers_wait_until = datetime.datetime.utcnow()
+
         self.ext_handlers.extHandlers.sort(key=operator.methodcaller('sort_key'))
         for ext_handler in self.ext_handlers.extHandlers:
-            # TODO: handle install in sequence, enable in parallel
+            if dep_level != ext_handler.sort_key():
+                dep_level = ext_handler.sort_key()
+                deps_wait_until = handlers_wait_until
+                handlers_wait_until += datetime.timedelta(seconds=(DEFAULT_EXT_TIMEOUT_MINUTES * 60))
```
hglkrijger: Nit - doesn't timedelta have a min property?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -360,11 +359,38 @@ def handle_ext_handlers(self, etag=None):
...
+            self.wait_on_ext_handler_dependencies(ext_handler, deps_wait_until)
             self.handle_ext_handler(ext_handler, etag)
-    
+
+    def wait_on_ext_handler_dependencies(self, ext_handler, wait_until):
+        dependencies = sum([e.dependencies for e in ext_handler.properties.extensions], [])
+        for dependency in dependencies:
+            handler_i = ExtHandlerInstance(dependency.handler, self.protocol)
+            if dependency.timeout is not None and dependency.timeout < DEFAULT_EXT_TIMEOUT_MINUTES:
```
hglkrijger: Let's not keep converting between seconds and minutes, it's an easy thing to miss.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -360,11 +359,38 @@ def handle_ext_handlers(self, etag=None):
...
+                wait_until -= datetime.timedelta(seconds=( (DEFAULT_EXT_TIMEOUT_MINUTES - dependency.timeout) * 60))
+            for ext in dependency.exts:
+                success_status, status = handler_i.is_ext_status_success(ext)
+                while not success_status:
```
hglkrijger: This is a strange construct - a `while not` around an `if` - please clean up.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -947,6 +973,17 @@ def collect_ext_status(self, ext):
 
         return ext_status
     
+    def is_ext_status_success(self, ext):
+        status = self.collect_ext_status(ext)
+        if status is None:
+            return (True, status)
+        if status.status != "success":
```
hglkrijger: Can you really rely on this value, and this casing?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -947,6 +973,17 @@ def collect_ext_status(self, ext):
 
         return ext_status
     
+    def is_ext_status_success(self, ext):
+        status = self.collect_ext_status(ext)
+        if status is None:
+            return (True, status)
```
hglkrijger: Seems odd, why is this `True`?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -947,6 +973,17 @@ def collect_ext_status(self, ext):
...
+        if status.status != "success":
+            return (False, status)
+        for substatus in status.substatusList:
+            if substatus.status != "success":
+                return (False, status)
+        return (True, status)
```
hglkrijger: If this is your default, you are not really testing for success, but failure.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -360,11 +359,38 @@ def handle_ext_handlers(self, etag=None):
...
+                success_status, status = handler_i.is_ext_status_success(ext)
+                while not success_status:
+                    if datetime.datetime.utcnow() > wait_until:
+                        raise ExtensionError("Timeout waiting for success status "
+                                             "from dependency {}/{} for {}"
+                                             "status was: {}".format(
+                                                 dependency.handler.name, ext.name,
+                                                 ext_handler.name, status))
+                    else:
+                        time.sleep(10)
```
hglkrijger: Seems like a long wait.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -360,11 +359,38 @@ def handle_ext_handlers(self, etag=None):
...
+        handlers_wait_until = datetime.datetime.utcnow()
+
         self.ext_handlers.extHandlers.sort(key=operator.methodcaller('sort_key'))
         for ext_handler in self.ext_handlers.extHandlers:
-            # TODO: handle install in sequence, enable in parallel
+            if dep_level != ext_handler.sort_key():
+                dep_level = ext_handler.sort_key()
+                deps_wait_until = handlers_wait_until
+                handlers_wait_until += datetime.timedelta(seconds=(DEFAULT_EXT_TIMEOUT_MINUTES * 60))
+            self.wait_on_ext_handler_dependencies(ext_handler, deps_wait_until)
```
hglkrijger: Still not sure I agree with the way this is done, which is the feedback I gave @uiri previously. If the dependency chain is simple, this should be just an ordering problem. The way this is built allows for a more complex scenario (A depends on B and C), which was not my expectation. Please clarify.

hglkrijger:
@sirajudeens please address the conflicts as well.

@vrdmr, @narrieta please review.

azurelinuxagent/ga/exthandlers.py
sirajudeens: This is the timeout value per dependency level.
CRP has an independent time out value. I will start an email thread including the CRP-side owner, on when/how to update the CRP timeout value.

azurelinuxagent/ga/exthandlers.py
sirajudeens: Agent throws exception when the status data is not present or the value is invalid. So I think we can rely on this value and casing.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -365,31 +365,39 @@ def handle_ext_handlers(self, etag=None):
...
-                    if datetime.datetime.utcnow() > wait_until:
-                        raise ExtensionError("Timeout waiting for success status "
-                                             "from dependency {}/{} for {}"
-                                             "status was: {}".format(
-                                                 dependency.handler.name, ext.name,
-                                                 ext_handler.name, status))
-                    else:
-                        time.sleep(10)
-                    success_status, status = handler_i.is_ext_status_success(ext)
+            success_status, status = handler_i.is_ext_status_success(dependency.ext)
```
uiri: Missing `None` check for `dependency.ext`.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -365,31 +365,39 @@ def handle_ext_handlers(self, etag=None):
...
+
+            # Keep polling for the extension status until it becomes success or times out
+            while not success_status and datetime.datetime.utcnow() <= wait_until:
+                time.sleep(5)
+                success_status, status = handler_i.is_ext_status_success(dependency.ext)
+
+            # In case of timeout, we log it and continue with the rest of the extensions
```
uiri: This is not the correct behaviour.

The extension should *not* be run if its dependency timed out. The extension runs once this function returns, unless this function throws an exception.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -365,31 +365,39 @@ def handle_ext_handlers(self, etag=None):
...
     def wait_on_ext_handler_dependencies(self, ext_handler, wait_until):
+        # Collect all the dependencies in the handler
+        # Check the status of the extension in each dependency
+        # Wait until it becomes success or times out
+
         dependencies = sum([e.dependencies for e in ext_handler.properties.extensions], [])
         for dependency in dependencies:
             handler_i = ExtHandlerInstance(dependency.handler, self.protocol)
-            if dependency.timeout is not None and dependency.timeout < DEFAULT_EXT_TIMEOUT_MINUTES:
-                wait_until -= datetime.timedelta(seconds=( (DEFAULT_EXT_TIMEOUT_MINUTES - dependency.timeout) * 60))
```
uiri: Why has this logic been removed entirely?

Further divergence from the Windows agent is not desirable.

azurelinuxagent/ga/exthandlers.py
sirajudeens: Some extensions do not report any status. We should not wait for such extensions to report status. That's why we return True here.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -310,11 +309,68 @@ def handle_ext_handlers(self, etag=None):
...
+        dep_level = None
+        deps_wait_until = None
+        handlers_wait_until = datetime.datetime.utcnow()
+        prev_handler = None
+
         self.ext_handlers.extHandlers.sort(key=operator.methodcaller('sort_key'))
         for ext_handler in self.ext_handlers.extHandlers:
-            # TODO: handle install in sequence, enable in parallel
+            # Use updated time limit for each dependency level
+            if dep_level != ext_handler.sort_key():
```
hglkrijger: Why not simply move through the extHandlers sorted by dependency level, and remove all the additional wait code?

sirajudeens: We not only make sure the extensions are installed sequentially, we need to make sure that the previous extension in the sequence is installed successfully before proceeding with the next extension in the sequence. Hence this wait code with different timeout value for each level.

hglkrijger: Consider something like this:

```
  for ext_handler in extHandlers:
    handle_ext_handler(ext_handler)
    wait_if_necessary(ext_handler)
```

Seems a little bit simpler than tracking full dependencies per se. We really just care if the last handler is done, if there is a stated dependency.

Regarding the timeouts, I thought we had decided on a total 90?

azurelinuxagent/ga/exthandlers.py
narrieta: But by returning True here we are saying that a extension that has not reported status yet has succeeded... correct?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -310,11 +309,68 @@ def handle_ext_handlers(self, etag=None):
...
+            while not success_status and datetime.datetime.utcnow() <= wait_until:
+                time.sleep(5)
+                success_status, status = handler_i.is_ext_status_success(ext)
+
+            # In case of timeout, we log it and return false so that the extensions waiting
+            # on this one can be skipped processing
+            if not success_status:
+                msg = "Timeout waiting for success status " \
+                      "from dependency {0}/{1} for {2}" \
+                      "status was: {3}".format(prev_handler.name,
```
narrieta: status is an instance of ExtensionStatus, so format will produce something like "<...ExtensionStatus object at 0x000001BF0B6B7DA0>"

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1591,11 +1591,6 @@ def parse_plugin(self, plugin):
         ext_handler.properties.version = getattrib(plugin, "version")
         ext_handler.properties.state = getattrib(plugin, "state")
 
-        try:
-            ext_handler.properties.dependencyLevel = int(getattrib(plugin, "dependencyLevel"))
-        except ValueError:
-            ext_handler.properties.dependencyLevel = 0
-
```
narrieta: I think we also need to remove the initialization of dependencyLevel in ExtHandlerProperties.__init__

azurelinuxagent/ga/exthandlers.py
```diff
@@ -871,6 +927,17 @@ def collect_ext_status(self, ext):
...
         return ext_status
     
+    def is_ext_status_success(self, ext):
+        status = self.collect_ext_status(ext)
+        if status is None:
+            return (True, status)
+        if status.status != "success":
+            return (False, status)
+        for substatus in status.substatusList:
+            if substatus.status != "success":
```
narrieta: @hglkrijger - what is the interpretation of status/substatus/status... are extensions allowed to report "success" in status and, say, "error" in one of the status/substatus/status? is this considered an extension error?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1627,6 +1622,13 @@ def parse_plugin_settings(self, ext_handler, plugin_settings):
             logger.error("Invalid extension settings")
             return
 
+        depends_on_node = find(settings[0], "DependsOn")
+
+        try:
+            depends_on_level = int(getattrib(depends_on_node, "dependencyLevel"))
+        except (ValueError, TypeError):
+            depends_on_level = 0
```
narrieta: do we need to catch these errors? this bumps the priority to 0.

azurelinuxagent/ga/exthandlers.py
narrieta: Agree with @hglkrijger that the for loop can be simplified to not make reference to the dependencies, but just ext_handler. Are the dependencies used only to increase the timeout 90 min per dependency level?

sirajudeens: Even now, the for loop does not make any reference to the dependency. It just references what was the previous extension installed in order to wait on that.

I think the one @hglkrijger suggests is that 'install an extension, wait for it'.This has a side effect of waiting for the highest level extensions as well, which is not required to be waited for.

Another way based on his suggestion is 'install an extension, wait for it if it is not the highest dependency level'. This requires calculating the highest dependency level in advance.
Do you think we need to change the logic to match this?

azurelinuxagent/ga/exthandlers.py
sirajudeens: Yes. I think we can return False in this case. @uiri do you have any comment on this?

azurelinuxagent/common/protocol/wire.py
sirajudeens: We expect string value corresponding to an integer value, which is parsed to get the integer value.
In order to guard against any corrupted value, we catch this exception. The dependency level is set to default in in that case. We assume that that kind of extension does not depend on anything so it gets its priority bumped up. Do you have any better suggestion in this case?

azurelinuxagent/ga/exthandlers.py
uiri: `collect_ext_status` returns None if the extension does not report any status. Not all extensions report status.

If you return `False` in this case, the agent will hang until it times out on such extensions.

sirajudeens: Base on offline discussion with @uiri , an extension can install and enable but may not create any status file. In this case, we don't want to wait for it until it is timed out. So we return True in this case.

azurelinuxagent/common/protocol/wire.py
sirajudeens: Removed it.

azurelinuxagent/common/protocol/wire.py
narrieta: OK, I double-checked the previous code and it was doing the same (setting the priority to 0 on error). I think we should log a warning in case bumping it up to 0 affects any dependencies and causes an extension failure.

azurelinuxagent/ga/exthandlers.py
narrieta: Per offline discussion @sirajudeens will double-check the validation of substatus in CRP to be sure this matches.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -310,11 +309,68 @@ def handle_ext_handlers(self, etag=None):
...
         self.ext_handlers.extHandlers.sort(key=operator.methodcaller('sort_key'))
         for ext_handler in self.ext_handlers.extHandlers:
-            # TODO: handle install in sequence, enable in parallel
+            # Use updated time limit for each dependency level
+            if dep_level != ext_handler.sort_key():
+                dep_level = ext_handler.sort_key()
+                deps_wait_until = handlers_wait_until
+                handlers_wait_until += datetime.timedelta(minutes=DEFAULT_EXT_TIMEOUT_MINUTES)
+
+            # Make sure that the previous extension is installed.
```
narrieta: minor: update the comment to mention install/enable instead of only install

azurelinuxagent/ga/exthandlers.py
```diff
@@ -310,11 +309,68 @@ def handle_ext_handlers(self, etag=None):
...
+            # Keep polling for the extension status until it becomes success or times out
+            while not success_status and datetime.datetime.utcnow() <= wait_until:
+                time.sleep(5)
+                success_status, status = handler_i.is_ext_status_success(ext)
+
+            # In case of timeout, we log it and return false so that the extensions waiting
+            # on this one can be skipped processing
+            if not success_status:
+                msg = "Timeout waiting for success status " \
+                      "from dependency {0}/{1} for {2} " \
```
narrieta: I'd remove the "dependency" part from this message, since there may not be any dependency between cur_handler and prev_handler

azurelinuxagent/ga/exthandlers.py
narrieta: per offline discussion we will got for a flat 90 min

azurelinuxagent/ga/exthandlers.py
```diff
@@ -310,11 +309,68 @@ def handle_ext_handlers(self, etag=None):
...
+        Wait until it becomes success or times out.
+        Return True if it is installed successfully. False if timed out.
+        '''
+        # No need to wait on anything for the very first extension
+        if prev_handler == None:
+            return True
+
+        handler_i = ExtHandlerInstance(prev_handler, self.protocol)
+        for ext in prev_handler.properties.extensions:
+            success_status, status = handler_i.is_ext_status_success(ext)
```
narrieta: TBD: Should we wait for success *or* error or just for success?

azurelinuxagent/ga/exthandlers.py
narrieta: TBD: What should we return if there is no status file?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -871,6 +917,34 @@ def collect_ext_status(self, ext):
...
+        # when there is no status file, it is incomplete
+        if status is None:
+            return (False, status)
+
+        # If not in terminal state, it is incomplete
+        if status.status not in EXTENSION_TERMINAL_STATUSES:
+            return (False, status.status)
+
+        # Main status is terminal.
+        # However, if there is any substatus in terminal state, it it incomplete
```
narrieta: the comment does not match the condition at line 934

azurelinuxagent/ga/exthandlers.py
```diff
@@ -871,6 +917,34 @@ def collect_ext_status(self, ext):
...
+        for substatus in status.substatusList:
+            if substatus.status not in EXTENSION_TERMINAL_STATUSES:
```
narrieta: This check looks incorrect... an extension may report a warning in a substatus and success in the status. This check would incorrectly return that the extension has not completed.

azurelinuxagent/common/protocol/wire.py
uiri: I think this warning is going to create a lot of noise in the logs if the platform is only sending down dependencyLevel for the cases where dependencies exist.

@sirajudeens are we sending down dependencyLevel all the time regardless or only when the customer specifies dependencies? What about for single VMs?

azurelinuxagent/ga/exthandlers.py
uiri: I think the behaviour should match whatever is done on Windows.

azurelinuxagent/ga/exthandlers.py
uiri: Install and enable were already run for the extension. This comment should say "Make sure that the previous extension reported status successfully" or something like that.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -871,6 +917,34 @@ def collect_ext_status(self, ext):
 
         return ext_status
     
+    def is_ext_handling_complete(self, ext):
```
uiri: I think it would be more useful for debugging and logging to define a `__str__` method on ExtensionStatus and return the full status.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -935,11 +935,25 @@ def collect_ext_status(self, ext):
 
         return ext_status
     
+    def get_ext_handling_status(self, ext):
+        seq_no, ext_status_file = self.get_status_file_path(ext)
+        if seq_no < 0 or ext_status_file is None:
+            return None
+
+        # Missing status file is considered a non-terminal state here
+        # so that extension sequencing can wait until it becomes existing
```
uiri: This will break existing extensions.

sirajudeens: can you give me an example of breaking case?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -310,11 +311,74 @@ def handle_ext_handlers(self, etag=None):
...
+        '''
+        dep_level = None
+        for ext_handler in self.ext_handlers.extHandlers:
+            if dep_level is None:
+                dep_level = ext_handler.sort_key()
+                continue
+
+            if dep_level != ext_handler.sort_key():
+                return True
+
```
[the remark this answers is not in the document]
uiri: The whole method seems unnecessary. The agent should not wait for the leaf nodes. In the case where there are no dependencies, every node is a leaf node.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -310,11 +311,74 @@ def handle_ext_handlers(self, etag=None):
...
                 return
 
+        wait_until = datetime.datetime.utcnow() + datetime.timedelta(minutes=DEFAULT_EXT_TIMEOUT_MINUTES)
+        prev_handler = None
+        ext_seq_required = self.is_ext_seq_required()
+
         self.ext_handlers.extHandlers.sort(key=operator.methodcaller('sort_key'))
         for ext_handler in self.ext_handlers.extHandlers:
-            # TODO: handle install in sequence, enable in parallel
+            if ext_seq_required:
```
narrieta: I would suggest switching the order of the calls to handle_ext_handler and wait_for_prev_handler_completion (call handle first and wait second) and just add a check to avoid waiting for the last extension. I think this express the intention of the code a little more clearly.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -310,11 +311,74 @@ def handle_ext_handlers(self, etag=None):
...
+                # Make sure that the previous extension is handled.
+                # If handled successfully, proceed with the current handler.
+                # Otherwise, skip the rest of them.
+                dep_level = ext_handler.sort_key()
+                if dep_level >= 0:
```
narrieta: Let's add a comment to explain the >=0 (we do this only for install/enable and not for uninstall/disable)

azurelinuxagent/ga/exthandlers.py
```diff
@@ -310,11 +311,74 @@ def handle_ext_handlers(self, etag=None):
...
+            if dep_level is None:
+                dep_level = ext_handler.sort_key()
+                continue
+
+            if dep_level != ext_handler.sort_key():
+                return True
+
+        return False
+
+    def wait_for_prev_handler_completion(self, prev_handler, wait_until):
```
narrieta: I would remove 'prev_' from the name of this function and its parameters. The function just waits for a handler, by itself it doesn't have a concept of "previous handler"

azurelinuxagent/ga/exthandlers.py
```diff
@@ -310,11 +311,74 @@ def handle_ext_handlers(self, etag=None):
...
+
+            # Keep polling for the extension status until it becomes success or times out
+            while not ext_completed and datetime.datetime.utcnow() <= wait_until:
+                time.sleep(5)
+                ext_completed, status = handler_i.is_ext_handling_complete(ext)
+
+            # In case of timeout or terminal error state, we log it and return false
+            # so that the extensions waiting on this one can be skipped processing
+            if not ext_completed or status != EXTENSION_STATUS_SUCCESS:
+                msg = "Extension {0} was timedout or status was: {1}".format(ext.name, status)
```
narrieta: Suggest rephrasing this message to something similar to "Extension {0} did not reach a terminal state within the allowed timeout. Last status was {1}"

azurelinuxagent/ga/exthandlers.py
```diff
@@ -310,11 +311,74 @@ def handle_ext_handlers(self, etag=None):
...
+            if ext_seq_required:
+                # Make sure that the previous extension is handled.
+                # If handled successfully, proceed with the current handler.
+                # Otherwise, skip the rest of them.
+                dep_level = ext_handler.sort_key()
+                if dep_level >= 0:
+                    if self.wait_for_prev_handler_completion(prev_handler, wait_until):
+                        prev_handler = ext_handler
+                    else:
+                        break
```
narrieta: Let's add a message saying we stopped processing other extensions

azurelinuxagent/ga/exthandlers.py
uiri: Please don't use `for i in range(L)` in a Python 2 or Python 2/3 codebase such as this one. It creates a list of integers unnecessarily.

For this particular case, keeping track of `i` is unnecessary. The second to last `ext_handler` may also be a leaf node so the test needs to be against the maximum dependency level, not against its position in the list. `if dep_level >= 0 and dep_level < max_dep_level:` ought to be a sufficient test.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -871,6 +935,37 @@ def collect_ext_status(self, ext):
...
     
+    def get_ext_handling_status(self, ext):
+        seq_no, ext_status_file = self.get_status_file_path(ext)
+        if seq_no < 0 or ext_status_file is None:
+            return None
+
+        # Missing status file is considered a non-terminal state here
+        # so that extension sequencing can wait until it becomes existing
+        if not os.path.exists(ext_status_file):
+            ext_status = ExtensionStatus(seq_no=seq_no)
```
narrieta: I think there no need to return an ExtensionStatus object. I think only the 'status' field is used by the code

azurelinuxagent/ga/exthandlers.py
```diff
@@ -871,6 +935,37 @@ def collect_ext_status(self, ext):
...
+            ext_status.message = u"Failed to get status file for {0}".format(ext.name)
+            ext_status.code = -1
+            ext_status.status = "warning"
```
narrieta: "no status file" or similar may be more appropriate

azurelinuxagent/ga/exthandlers.py
```diff
@@ -871,6 +935,37 @@ def collect_ext_status(self, ext):
...
+        else:
+            ext_status = self.collect_ext_status(ext)
+
+        return ext_status
+
+    def is_ext_handling_complete(self, ext):
+        status = self.get_ext_handling_status(ext)
+
+        # when there is no status, the handling is complete and return None status
```
narrieta: We need to update the comment... this condition would be reached when seq < 0 (i.e. no new user settings) rather than when there is no status file

azurelinuxagent/ga/exthandlers.py
narrieta: agree

azurelinuxagent/ga/exthandlers.py
jasonzio: Disagree. If the extension completes within the allotted time but returns an error, you still wind up emitting this message. Your suggestion makes that error sound like a timeout. The right thing to do is to cleanly separate the "returned an error" condition from the "did not complete within the allotted time limit" condition, emitting a clear (and distinct) message for each of those two cases.

sirajudeens: @jasonzio , is your concern about the error message or the functionality itself when there is a failure?
As per the last discussion with the VM agent folks, if an extension timedout/failed, we should stop processing the rest of the extensions, when the extension sequencing is required.

narrieta: @jasonzio agree that 2 separate messages would be better

azurelinuxagent/ga/exthandlers.py
uiri: > I think only the 'status' field is used by the code

It will be easier to debug if we have the entire status object and not just the fields used by the code. Also, it is impossible to tell whether it is a status or a substatus otherwise.

sirajudeens: No more substatus per @narrieta 's suggestion.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -310,11 +311,63 @@ def handle_ext_handlers(self, etag=None):
                 logger.info("Extension handling is on hold")
                 return
 
+        wait_until = datetime.datetime.utcnow() + datetime.timedelta(minutes=DEFAULT_EXT_TIMEOUT_MINUTES)
+        prev_handler = None
```
narrieta: this variable is not referenced anymore

azurelinuxagent/ga/exthandlers.py
```diff
@@ -310,11 +311,63 @@ def handle_ext_handlers(self, etag=None):
...
-    
+
+            # Wait for the extension installation until it is handled.
+            # We wait only for the installation. Not for the uninstallation.
+            # If handled successfully, proceed with the current handler.
+            # Otherwise, skip the rest of the extension installation.
+            dep_level = ext_handler.sort_key()
+            if dep_level >= 0 and dep_level < max_dep_level:
+                if not self.wait_for_handler_successful_completion(ext_handler, wait_until):
+                    logger.warn("Skipping processing extensions")
```
narrieta: Let's add the extension name and rephrase the message to read something similar to "Extension {0} failed or timed out, will skip processing the rest of the extensions"

azurelinuxagent/ga/exthandlers.py
```diff
@@ -310,11 +311,63 @@ def handle_ext_handlers(self, etag=None):
...
+                logger.warn(msg)
+                add_event(AGENT_NAME,
+                          version=CURRENT_VERSION,
+                          op=WALAEventOperation.ExtensionProcessing,
+                          is_success=False,
+                          message=msg)
+                return False
+
+            if status != EXTENSION_STATUS_SUCCESS:
+                msg = "Extension {0} did not succeed. Last status was {1}".format(ext.name, status)
```
narrieta: Minor: "Status was" instead of "Last status was"

azurelinuxagent/ga/exthandlers.py
sirajudeens: This will be too much logging. We already log the extension name in the wait function. This is to just log that the rest of the extensions will be skipped.

narrieta: Sure, we can leave the extension name out if you prefer so. Let's be explicit about why we are skipping the rest of the extensions, though. Maybe something like "An extension failed or timed out, will skip processing the rest of the extensions". Hopefully this will be more helpful for CSS or a customer reading the log.

tests/ga/test_extension.py
```diff
@@ -666,6 +700,126 @@ def test_ext_handler_no_reporting_status(self, *args):
...
+        ExtHandlerInstance.get_ext_handling_status = MagicMock(return_value=None)
+        self.assertTrue(exthandlers_handler.wait_for_handler_successful_completion(handler, datetime.datetime.utcnow()))
+
+    def _test_wait_for_handler_successful_completion(self, exthandlers_handler):
+        handler_name = "Handler"
+        exthandler = ExtHandler(name=handler_name)
+        extension = Extension(name=handler_name)
+        exthandler.properties.extensions.append(extension)
+
+        # Override the timout value to minimize the test duration
```
narrieta: typo: "timout"

tests/ga/test_extension.py
```diff
@@ -666,6 +700,126 @@ def test_ext_handler_no_reporting_status(self, *args):
...
+    def test_wait_for_handler_successful_completion_empty_exts(self, *args):
+        test_data = WireProtocolData(DATA_FILE)
+        exthandlers_handler, protocol = self._create_mock(test_data, *args)
+
+        handler = ExtHandler(name="handler")
+
+        ExtHandlerInstance.get_ext_handling_status = MagicMock(return_value=None)
+        self.assertTrue(exthandlers_handler.wait_for_handler_successful_completion(handler, datetime.datetime.utcnow()))
+
+    def _test_wait_for_handler_successful_completion(self, exthandlers_handler):
```
narrieta: let's remove "test" from the name of this helper function. this threw me off... gave me the impression that it was a test

azurelinuxagent/ga/exthandlers.py
```diff
@@ -871,6 +923,35 @@ def collect_ext_status(self, ext):
...
     
+    def get_ext_handling_status(self, ext):
+        seq_no, ext_status_file = self.get_status_file_path(ext)
+        if seq_no < 0 or ext_status_file is None:
+            return None
+
+        # Missing status file is considered a non-terminal state here
+        # so that extension sequencing can wait until it becomes existing
+        if not os.path.exists(ext_status_file):
+            status = "warning"
```
narrieta: we should change this status... this gives the impression that the extension is reporting a warning when in actuality there is no status file. I would suggest return something like "status file not found" or similar

tests/ga/test_extension.py
```diff
@@ -666,6 +700,126 @@ def test_ext_handler_no_reporting_status(self, *args):
...
+        exthandler = ExtHandler(name=handler_name)
+        extension = Extension(name=handler_name)
+        exthandler.properties.extensions.append(extension)
+
+        # Override the timout value to minimize the test duration
+        wait_until = datetime.datetime.utcnow() + datetime.timedelta(seconds=5)
+        return exthandlers_handler.wait_for_handler_successful_completion(exthandler, wait_until)
+
+    def test_wait_for_handler_successful_completion_none(self, *args):
+        test_data = WireProtocolData(DATA_FILE)
```
narrieta: let's add a comment about what each test is verifying or, better, a more descriptive test name... for this one maybe "test_wait_for_handler_successful_completion_returns_false_when_there_is_no_status_file"

tests/ga/test_extension.py
```diff
@@ -666,6 +700,126 @@ def test_ext_handler_no_reporting_status(self, *args):
...
+        exthandlers_handler, protocol = self._create_mock(test_data, *args)
+
+        ExtHandlerInstance.get_ext_handling_status = MagicMock(return_value=None)
```
narrieta: what is this test verifying? that wait_for_handler_successful_completion returns false when there is no status file? If that is the case then I would instead mock get_status_file_path to return None

tests/ga/test_extension.py
```diff
@@ -666,6 +700,126 @@ def test_ext_handler_no_reporting_status(self, *args):
...
+    def test_wait_for_handler_successful_completion_error_status(self, *args):
+        test_data = WireProtocolData(DATA_FILE)
+        exthandlers_handler, protocol = self._create_mock(test_data, *args)
+
+        status = "error"
+
+        ExtHandlerInstance.get_ext_handling_status = MagicMock(return_value=status)
+        self.assertFalse(self._test_wait_for_handler_successful_completion(exthandlers_handler))
+
+    def test_wait_for_handler_successful_completion_timeout(self, *args):
```
narrieta: I think this test needs to be re-written... the way it reads now it looks like it is testing that wait_for_handler_successful_completion returns false when the status is "warning" (the test name seems to imply that this should be testing when the extension times out before reporting a terminal status)

tests/ga/test_extension.py
```diff
@@ -666,6 +700,126 @@ def test_ext_handler_no_reporting_status(self, *args):
...
+        test_data = WireProtocolData(DATA_FILE)
+        exthandlers_handler, protocol = self._create_mock(test_data, *args)
+
+        handler_name = "Handler"
+        exthandler = ExtHandler(name=handler_name)
+        extension = Extension(name=handler_name)
+        exthandler.properties.extensions.append(extension)
+
+        # list of [seq_no, ext_status_file, is_status_file_exist, expected_result]
+        test_cases = [
```
narrieta: can we add comments about what each test case is verifying? e.g. the first 2 seems to be checking the result when there are no new test settings, the 3rd one seems to be verifying the case when the extension doesn't create a status file, etc.

azurelinuxagent/ga/exthandlers.py
sirajudeens: This value is verified against a set of valid values. I think we should adhere to those values. Also it is clearly commented above that condition in which it can occur.

tests/ga/test_extension.py
sirajudeens: This is testing when there is no extension. The case you are referring is tested in test_get_ext_handling_status()

tests/ga/test_extension.py
sirajudeens: "warning" is a non-terminal state which will eventually timeout the wait function and return false. This is what is tested here. We have set the timeout value to be 5 seconds to minimize the duration.
Do you mean to rename this function? If it needs to be rewritten, can you please explain what is missing here?

tests/ga/test_extension.py
sirajudeens: It is already commented in line# 765.

tests/ga/test_extension.py
sirajudeens: When get_ext_handling_status() is mocked, it is unnecessary to mock the get_status_file_path(). It will be a no-op in that case.
The case which you are referring is covered in test_get_ext_handling_status()

azurelinuxagent/ga/exthandlers.py
```diff
@@ -871,6 +923,35 @@ def collect_ext_status(self, ext):
 
         return ext_status
     
+    def get_ext_handling_status(self, ext):
```
uiri: I'm still not clear as to the purpose of this method.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -310,11 +311,62 @@ def handle_ext_handlers(self, etag=None):
...
+        max_dep_level = max([handler.sort_key() for handler in self.ext_handlers.extHandlers])
+
         self.ext_handlers.extHandlers.sort(key=operator.methodcaller('sort_key'))
         for ext_handler in self.ext_handlers.extHandlers:
-            # TODO: handle install in sequence, enable in parallel
             self.handle_ext_handler(ext_handler, etag)
-    
+
+            # Wait for the extension installation until it is handled.
+            # This is done for the install and enable. Not for the uninstallation.
```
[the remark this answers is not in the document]
narrieta: The initial ask for a comment similar to this was to clarify the intention of the "dep_level >= 0" check... we wait only when the extension was installed or enabled. As it is currently phrased it seems like it is not helping so feel free to remove it or change it accordingly.

azurelinuxagent/ga/exthandlers.py
narrieta: My concern is that if reach this path, then the status in this message:

"Extension {0} did not reach a terminal state within the allowed timeout. Last status was {1}".

will be "warning". When looking at the log, it'll seemed like the extension actually reported a warning, when in actuality  it didn't report any status. That may be confusing.

uiri: We need to pass up the entire status, not just the status field, to give a more complete log message and add a `__str__` to `ExtStatus` so that it logs something sensible

narrieta:
BVTs OK

Issue #1294, opened 2018-08-09
InitializeHostPlugin should be called after incarnation resets


vrdmr:
After a live migration, when the agent resets, it is not updating the host plugin information (on the GA). After the new incarnation comes in, there are no new `InitializeHostPlugin` events, which points to that the agent is continuing to use the container id/role name in the memory.

The [Waagent.log](https://gist.githubusercontent.com/vrdmr/b999902765c1405a37fe628bbf88b8c2/raw/a4d5d920db8c0e94507ff757b7e8e3a232258835/waagent.log) has a large gap during this live migration as well, and does not have any restart logs.

```
2018/08/03 10:54:34.454022 INFO ExtHandler Setting host plugin as default channel
2018/08/04 03:10:41.088150 WARNING ExtHandler Fetch failed from [http://168.63.129.16:32526/extensionArtifact]: [HTTP Failed] [400: Bad Request] b'{  "errorCode": "InvalidContainerConfiguration",  "message": "Invalid request.",  "details": "[0x8000100a] ValidateIpAddress"}'
```

Based on the investigation @hglkrijger did, the potential issue might be [here](https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/ga/update.py#L681), where the `ResourceGoneError` does not account for any 400s.

hglkrijger:
```
2018/08/01 06:40:27.767168 ERROR ExtHandler Event: name=WALinuxAgent, op=ExtensionProcessing, message=Failed to report vm agent status: [ProtocolError] Failed to upload status blob via either channel, duration=0
2018/08/01 06:40:30.780706 ERROR ExtHandler HostGAPlugin: Failed Get API versions: [HTTP Failed] [400: Bad Request] b'{  "errorCode": "InvalidContainerConfiguration",  "message": "Invalid request.",  "details": "[0x8000100a] ValidateIpAddress"}'
2018/08/01 06:40:30.786391 ERROR ExtHandler Event: name=WALinuxAgent, op=InitializeHostPlugin, message=, duration=0
2018/08/01 06:40:30.788608 WARNING ExtHandler Exception retrieving agent manifests: Traceback (most recent call last):
  File "bin/WALinuxAgent-2.2.29-py2.7.egg/azurelinuxagent/ga/update.py", line 653, in _upgrade_available
    pkg_list = protocol.get_vmagent_pkgs(manifests[0])
[...]

PR #1299, opened 2018-08-16
Invalid container configuration

hglkrijger:
## Description

Issue #1294 

Making requests to HostGAPlugin uses a `container-id` header, which is part of goal state. When a VM live-migrates, or goal state is outdated for another reason, and we make a request with this outdated container-id, we receive a `400` with `InvalidContainerConfiguration`. This should really be a `410`, but in order to address the issue now, the fix is to map these specific `400`s to `410`s. This will trigger a forced goal state update when `_upgrade_available` is called. 

Before this call happens we may try heartbeat against HostGAPlugin, which would fail for the same reason. In that case we should not report failure; this addresses a mismatch in behavior compared to the other APIs we report against. 

---

### PR information


hglkrijger:
Reviewed offline

PR #1303, opened 2018-08-17
Take wireserver address from option 245 in DHCP lease

### LINKED ISSUE #1176: don't use the dns server from dhcp as the api endpoint
https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/common/osutil/default.py#L846

get_endpoint_from_leases_path incorrectly uses the HEADER_DNS as the API endpoint when instead the endpoint should be pulled from dhcp option 245 using get_dhcp_lease_endpoint (which isn't implemented yet) 

This breaks an Azure deployment if an end user modifies a vnet to use a custom DNS server because waagent no longer uses the correct KNOWN_WIRESERVER_IP or option 245 from the DHCP lease file.

narrieta:
## Description

Fixes #1176

On default vnet configurations, the wireserver address shows up both on domain-name-servers and option 245. Currently we take the address from domain-name-servers. If the vnet has a custom DNS server, domain-name-servers will have the address of that server, not of wireserver. Changed the code to take the address from option 245 instead.

---

### PR information

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -937,28 +937,23 @@ class DefaultOSUtil(object):
         endpoint = None
 
         HEADER_LEASE = "lease"
-        HEADER_OPTION = "option unknown-245"
-        HEADER_DNS = "option domain-name-servers"
+        HEADER_OPTION_245 = "option unknown-245"
         HEADER_EXPIRE = "expire"
         FOOTER_LEASE = "}"
         FORMAT_DATETIME = "%Y/%m/%d %H:%M:%S"
+        option_245_re = re.compile(r'\s*option\s+unknown-245\s+([0-9a-fA-F]+):([0-9a-fA-F]+):([0-9a-fA-F]+):([0-9a-fA-F]+);')
 
         logger.info("looking for leases in path [{0}]".format(pathglob))
         for lease_file in glob.glob(pathglob):
             leases = open(lease_file).read()
-            if HEADER_OPTION in leases:
+            if HEADER_OPTION_245 in leases:
                 cached_endpoint = None
-                has_option_245 = False
+                option_245_match = None
                 expired = True  # assume expired
                 for line in leases.splitlines():
                     if line.startswith(HEADER_LEASE):
                         cached_endpoint = None
-                        has_option_245 = False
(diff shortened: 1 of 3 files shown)


tests/data/dhcp.leases.custom.dns
```diff
@@ -0,0 +1,56 @@
+lease {
```
boumenot: Is this a typical example?  

The same lease is repeated three times. Is this is expected. When you are searching for `option unknown-245` should you look for the first or last? Does it matter?  Should you take the interface into account, e.g. interface "eth0" when parsing?

azurelinuxagent/common/osutil/default.py
```diff
@@ -937,28 +937,23 @@ def get_endpoint_from_leases_path(pathglob):
         endpoint = None
 
         HEADER_LEASE = "lease"
-        HEADER_OPTION = "option unknown-245"
-        HEADER_DNS = "option domain-name-servers"
+        HEADER_OPTION_245 = "option unknown-245"
```
boumenot: OpenBSD referecnes HEADER_OPTION too.  Do any changes need to be made there?

narrieta: Yup, good catch. Will update there as well.

tests/data/dhcp.leases.custom.dns
narrieta: This is the same file as used for the other unit tests (except for the last entry, which has a custom DNS server: 8.8.8.8). I would say it is typical, looks very similar to the one on my VM. According to man, the file acts as a log and the last entry should always be used.

Actually, I need to update the address in option 245 to a dummy address, to ensure the last one is being picked (with the previous code this was ensured by using 'invalid' and 'expired' in domain-name-servers of the first 2 entries.

narrieta: I just updated the test file -- added dummy wireserver addresses to the first 2 entries to ensure the test is pickling up the last one (third one).

azurelinuxagent/common/osutil/default.py
narrieta: After further discussion I will leave OpenBSD as is for now. Thanks for pointing this out.

PR #1305, opened 2018-08-20
Extension state cleanup

### LINKED ISSUE #948: Investigate whether old extension state should be removed
When a VM or disk is migrated, old extension state may be reported. We need to investigate whether this is the case, and possibly clear old extension status before processing goal state.

### LINKED ISSUE #1116: exthandlers.ExtHandlerInstance.collect_ext_status doesn't use 'ext' parameter
This method takes a parameter but ignores it. Should it return status for just that ext, or all of them? Current behavior is to return an object that reflects the entire status. Should the implementation reach inside the ext_status object and strip it down to just the information about ext?

hglkrijger:
## Description

1. On vm or disk migration, clear out status and settings files; fixes #948

Example:
```
/var/lib/waagent/Microsoft.Azure.Extensions.CustomScript-2.0.6/status/*.status
/var/lib/waagent/Microsoft.Azure.Extensions.CustomScript-2.0.6/config/*.settings
/var/lib/waagent/Microsoft.Azure.Extensions.CustomScript-2.0.6/config/HandlerStatus
```

2. When reporting extension status, use the goal-state (current) sequence number, not the largest on disk; fixes #1116

Example:
```
{
  "handlerName": "Microsoft.Azure.Extensions.CustomScript",
  "handlerVersion": "2.0.6",
  "status": "Ready",
  "code": 0,
  "formattedMessage": {
    "lang": "en-US",
    "message": "Plugin enabled"
  },
  "runtimeSettingsStatus": {
**  "sequenceNumber": "366",
    "settingsStatus": {
      "timestampUTC": "2018-06-29T15:36:04Z",
      "status": {
        "name": "Microsoft.Azure.Extensions.CustomScript",
        "operation": "Enable",
        "status": "success",
        "formattedMessage": {
          "lang": "en-US",
          "message": "Enable succeeded: ..."
        }
      }
    }
  }
}
```

---

### PR information

### DIFF
--- a/azurelinuxagent/pa/deprovision/default.py
+++ b/azurelinuxagent/pa/deprovision/default.py
@@ -19,16 +19,20 @@
 
 import glob
 import os.path
+import re
 import signal
 import sys
 
 import azurelinuxagent.common.conf as conf
 import azurelinuxagent.common.utils.fileutil as fileutil
 import azurelinuxagent.common.utils.shellutil as shellutil
+from azurelinuxagent.common import version
 
 from azurelinuxagent.common.exception import ProtocolError
 from azurelinuxagent.common.osutil import get_osutil
 from azurelinuxagent.common.protocol import get_protocol_util
+from azurelinuxagent.ga.exthandlers import HANDLER_NAME_PATTERN
+
 
 def read_input(message):
     if sys.version_info[0] >= 3:
@@ -118,6 +122,20 @@ class DeprovisionHandler(object):
         actions.append(DeprovisionAction(fileutil.rm_files,
                                          ["/var/lib/NetworkManager/dhclient-*.lease"]))
 
+    def del_ext_handler_files(self, warnings, actions):
+        ext_dirs = [d for d in os.listdir(conf.get_lib_dir())
+                    if os.path.isdir(os.path.join(conf.get_lib_dir(), d))
+                    and re.match(HANDLER_NAME_PATTERN, d) is not None
+                    and not version.is_agent_path(d)]
+
+        for ext_dir in ext_dirs:
+            ext_base = os.path.join(conf.get_lib_dir(), ext_dir)
+            files = glob.glob(os.path.join(ext_base, 'status', '*.status'))
+            files += glob.glob(os.path.join(ext_base, 'config', '*.settings'))
+            files += glob.glob(os.path.join(ext_base, 'config', 'HandlerStatus'))
+
+            if len(files) > 0:
+                actions.append(DeprovisionAction(fileutil.rm_files, files))
 
     def del_lib_dir_files(self, warnings, actions):
         known_files = [
@@ -221,6 +239,7 @@ class DeprovisionHandler(object):
                     include_once=False, deluser=False)
         self.del_dhcp_lease(warnings, actions)
         self.del_lib_dir_files(warnings, actions)
+        self.del_ext_handler_files(warnings, actions)
 
         return warnings, actions
 
--- a/tests/ga/test_exthandlers.py
+++ b/tests/ga/test_exthandlers.py
@@ -2,8 +2,8 @@
 # Licensed under the Apache License.
 import json
 
-from azurelinuxagent.common.protocol.restapi import ExtensionStatus
-from azurelinuxagent.ga.exthandlers import parse_ext_status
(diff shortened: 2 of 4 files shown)


tests/pa/test_deprovision.py
```diff
@@ -133,11 +138,18 @@ def test_del_lib_dir_files(self,
             'GoalState.1.xml',
             'Extensions.2.xml',
             'ExtensionsConfig.2.xml',
-            'GoalState.2.xml'
+            'GoalState.2.xml',
+            'Microsoft.Azure.Extensions.CustomScript-2.0.6/config/42.settings',
+            'Microsoft.Azure.Extensions.CustomScript-2.0.6/config/HandlerStatus',
+            'Microsoft.Azure.Extensions.CustomScript-2.0.6/config/HandlerState',
+            'Microsoft.Azure.Extensions.CustomScript-2.0.6/status/12.notstatus',
+            'WALinuxAgent-2.2.26/config/0.settings'
```
boumenot: 'Microsoft.Azure.Extensions.CustomScript-2.0.6/mrseq'

azurelinuxagent/pa/deprovision/default.py
```diff
@@ -118,6 +122,20 @@ def del_dhcp_lease(self, warnings, actions):
...
+        ext_dirs = [d for d in os.listdir(conf.get_lib_dir())
+                    if os.path.isdir(os.path.join(conf.get_lib_dir(), d))
+                    and re.match(HANDLER_NAME_PATTERN, d) is not None
+                    and not version.is_agent_path(d)]
+
+        for ext_dir in ext_dirs:
+            ext_base = os.path.join(conf.get_lib_dir(), ext_dir)
+            files = glob.glob(os.path.join(ext_base, 'status', '*.status'))
+            files += glob.glob(os.path.join(ext_base, 'config', '*.settings'))
+            files += glob.glob(os.path.join(ext_base, 'config', 'HandlerStatus'))
```
boumenot: files += glob.glob(os.path.join(ext_base, 'mrseq')

tests/ga/test_exthandlers.py
```diff
@@ -72,3 +72,44 @@ def test_parse_extension_status01(self):
...
+
+        self.assertEqual(expected_sequence_number, seq)
+        if seq > -1:
+            self.assertTrue(path.endswith('/foo-1.2.3/status/{0}.status'.format(expected_sequence_number)))
+        else:
+            self.assertIsNone(path)
+
+    def test_extension_sequence_number(self):
+        self.assert_extension_sequence_number(goal_state_sequence_number="12",
+                                              disk_sequence_number=366,
```
boumenot: I understand the value in asserting what the agent will do in this case, but I don't believe the answer is *correct*. Extensions do not know what settings to execute when enabled. Extensions only know to execute the latest, and the latest is arbitrary. Extensions execute a directory listing, get the number of the file, sort, and take the last one in ascending order. The agent never explicitly says which settings to execute, which is a mistake, but that ship has sailed.

I would keep the test as is because I believe the only way to get into this situation is if CRP's sequence for an extension install stops being monotonically increasing.  If CRP is wrong then we have other issues. It is worth adding a telemetry event for the case where get_largest_seq_no() > extension.sequenceNumber as a signal that things have gone pear shaped.

azurelinuxagent/pa/deprovision/default.py
hglkrijger: Added, thanks

tests/ga/test_exthandlers.py
hglkrijger: Added telemetry

tests/pa/test_deprovision.py
hglkrijger: Updated

tests/tools.py
```diff
@@ -122,7 +122,7 @@ def setUpClass(cls):
         if not hasattr(cls, "assertIsNone"):
             cls.assertIsNone = cls.emulate_assertIsNone
         if not hasattr(cls, "assertIsNotNone"):
-            cls.assertIsNone = cls.emulate_assertIsNotNone
+            cls.assertIsNotNone = cls.emulate_assertIsNotNone
```
boumenot: Oops.

PR #1310, opened 2018-08-21
Verify extension deletion scenario

### LINKED ISSUE #959: Detection and mitigation of deleted extension versions could be improved
There are references to extension versions `1.4.*.*` in `/var/lib/waagent/Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux.104.manifest.xml` and older, but not in `/var/lib/waagent/Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux.105.manifest.xml` and newer. It seems the extension publisher has deleted these versions, and basically we do not support that scenario in the agent (we always ask publishers to never remove the latest version).

In this particular case, the issue manifested when a VM repeatedly gave "Failed to load manifest file" errors in `/var/log/waagent.log` approximately every 8 seconds.

The agent should be able to deal with this if the customer stops the agent, removes any reference of this version from disk (if the files are complete, the agent is already doing this) and then restarts the agent. However, perhaps this could be more hardened and automatic in the agent.

More details can be found in the incident 52697262; Hans and lagalbra have context.

hglkrijger:
- since #959 was opened we have made some changes in how we call decide_version, this avoids the issue listed
- adds a unit test to verify behavior
- code cleanup to remove retry in decide_version, as this wait is not useful
- unit test cleanup so extension tests can be run with cgroups disabled
- closes #959
---

### PR information

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -319,22 +319,13 @@ class ExtHandlersHandler(object):
 
         try:
             state = ext_handler.properties.state
-            max_retries = 5
-            retry_count = 0
-            while ext_handler_i.decide_version(target_state=state) is None:
-                if retry_count >= max_retries:
-                    err_msg = "Unable to find manifest for extension {0}".format(ext_handler_i.ext_handler.name)
-                    ext_handler_i.set_operation(WALAEventOperation.Download)
-                    ext_handler_i.set_handler_status(message=ustr(err_msg), code=-1)
-                    ext_handler_i.report_event(message=ustr(err_msg), is_success=False)
-                    return
-                time.sleep(2**retry_count * 10)
-                retry_count += 1
-            if retry_count != 0:
+            if ext_handler_i.decide_version(target_state=state) is None:
+                err_msg = "Unable to find manifest for extension {0}".format(ext_handler_i.ext_handler.name)
                 ext_handler_i.set_operation(WALAEventOperation.Download)
-                err_msg = "Able to find manifest for extension {0} after {1} failed attempts.".format(
-                    ext_handler_i.ext_handler.name, retry_count)
-                ext_handler_i.report_event(message=ustr(err_msg))
+                ext_handler_i.set_handler_status(message=ustr(err_msg), code=-1)
+                ext_handler_i.report_event(message=ustr(err_msg), is_success=False)
+                return
+
             self.get_artifact_error_state.reset()
             if not ext_handler_i.is_upgrade and self.last_etag == etag:
                 if self.log_etag:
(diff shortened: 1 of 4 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -319,22 +319,13 @@ def handle_ext_handler(self, ext_handler, etag):
...
-                    err_msg = "Unable to find manifest for extension {0}".format(ext_handler_i.ext_handler.name)
-                    ext_handler_i.set_operation(WALAEventOperation.Download)
-                    ext_handler_i.set_handler_status(message=ustr(err_msg), code=-1)
-                    ext_handler_i.report_event(message=ustr(err_msg), is_success=False)
-                    return
-                time.sleep(2**retry_count * 10)
-                retry_count += 1
-            if retry_count != 0:
+            if ext_handler_i.decide_version(target_state=state) is None:
+                err_msg = "Unable to find manifest for extension {0}".format(ext_handler_i.ext_handler.name)
```
uiri: Re-reading the error message that I wrote for this, I realize that it is inaccurate. If the agent can't find the manifest, it will throw an exception.

Maybe something like this?
```python
"Unable to find version {0} in manifest for extension {1}".format(ext_handler_i.ext_handler.properties.version, ext_handler_i.ext_handler.name)
```

azurelinuxagent/ga/exthandlers.py
```diff
@@ -319,22 +319,13 @@ def handle_ext_handler(self, ext_handler, etag):
...
-            max_retries = 5
-            retry_count = 0
-            while ext_handler_i.decide_version(target_state=state) is None:
-                if retry_count >= max_retries:
-                    err_msg = "Unable to find manifest for extension {0}".format(ext_handler_i.ext_handler.name)
-                    ext_handler_i.set_operation(WALAEventOperation.Download)
-                    ext_handler_i.set_handler_status(message=ustr(err_msg), code=-1)
-                    ext_handler_i.report_event(message=ustr(err_msg), is_success=False)
-                    return
-                time.sleep(2**retry_count * 10)
```
uiri: Could we perhaps keep 1 (or 2) retries without the sleep?
I am not sure if the agent has any caching for HTTP(S) responses but perhaps clear the cache and retry if the first try fails?

azurelinuxagent/ga/exthandlers.py
hglkrijger: Sure, will update

azurelinuxagent/ga/exthandlers.py
hglkrijger: The goal state by definition has been updated at this point, so fetching again will not change anything.

uiri: The goal state contains a URL that points to the extensions config which itself contains a URL for each extension which points to each extension's manifest. I am referring to refetching that extension manifest (e.g. `Publisher.Extension.3.xml` or whatever filename)

boumenot: There was a change in 2.2.27+ where the cache is always cleared when a new goal state is fetched. I believe this addresses your concerns.

PR #1319, opened 2018-08-24
Stream large downloads to disk

### LINKED ISSUE #1138: Agent Should not Buffer an Extension Download in Memory
The agent reads an entire extension package into memory, and **then** writes it to disk.  This was never a good idea, but the size of extensions has grown considerably since the code was written.

```python
    def download_ext_handler_pkg(self, uri, headers=None, use_proxy=True):
        try:
            resp = restutil.http_get(uri, headers=headers, use_proxy=use_proxy)
            if restutil.request_succeeded(resp):
                return resp.read()
        except Exception as e:
            logger.warn("Failed to download from: {0}".format(uri), e)
```

```python
    try:
            fileutil.write_file(self.pkg_file, bytearray(package), asbin=True)
            zipfile.ZipFile(self.pkg_file).extractall(self.get_base_dir())
        except IOError as e:
            fileutil.clean_ioerror(e,
                paths=[self.get_base_dir(), self.pkg_file])
            raise ExtensionError(u"Failed to write and unzip plugin", e)
```

hglkrijger:
## Description

- when fetching extension packages we should not buffer the entire response content in memory
- manifests, agent downloads, and all other downloads are unchanged
- move the default download implementation out of `Protocol` and into `MetadataProtocol` for clarity
- minor cleanup and layout fixes in restutil
- adds a unit test for large downloads which is disabled in Travis
- closes #1138 

---

### PR information

### DIFF
--- a/tests/protocol/test_wire.py
+++ b/tests/protocol/test_wire.py
@@ -16,9 +16,13 @@
 #
 
 import glob
+import stat
+import zipfile
 
 from azurelinuxagent.common import event
 from azurelinuxagent.common.protocol.wire import *
+from azurelinuxagent.common.utils.shellutil import run_get_output
+from tests.common.osutil.test_default import running_under_travis
 from tests.protocol.mockwiredata import *
 
 data_with_bom = b'\xef\xbb\xbfhehe'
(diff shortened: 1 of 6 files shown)


tests/protocol/test_wire.py
```diff
@@ -162,16 +166,46 @@ def test_get_host_ga_plugin(self, *args):
...
         self.assertEqual(patch_http.call_args_list[0][0][1], ext_uri)
         self.assertEqual(patch_http.call_args_list[1][0][1], host_uri)
 
+    @skip_if_predicate_true(running_under_travis, "Travis unit tests should not have external dependencies")
+    def test_download_ext_handler_pkg_stream(self, *args):
+        ext_uri = 'https://dcrdata.blob.core.windows.net/files/packer.zip'
+        tmp = tempfile.mkdtemp()
+        destination = os.path.join(tmp, 'test_download_ext_handler_pkg_stream.zip')
+
+        success = WireProtocol(wireserver_url).download_ext_handler_pkg(ext_uri, destination)
```
vrdmr: qq - Do we need a separate test for protocol.metadata.download_ext_handler_pkg()? Underneath both the versions use restutil.http_get() so most likely no.

hglkrijger: We are not great about having Azure Stack tests in general. In this case I would say there is no change in behavior, so it is not worth addressing as part of this PR.

vrdmr: Got it.

Thanks for doing this fix! :)

PR #1322, opened 2018-08-27
restart ssh service only after ssh host key is regenerated .

andyliuliming:
we are trying to speed up the provisioning of our bosh stemcell (special ubuntu image).
and thinking about to set Provisioning.RegenerateSshHostKeyPair=n in it.
and when testing it, found that the ssh service would be restarted even when the host key not been generated.
I think this is not necessary, could you please help review and help merge?
Thanks,
Andy

### DIFF
--- a/azurelinuxagent/pa/provision/default.py
+++ b/azurelinuxagent/pa/provision/default.py
@@ -83,9 +83,10 @@ class ProvisionHandler(object):
 
             self.provision(ovf_env)
 
-            thumbprint = self.reg_ssh_host_key()
-            self.osutil.restart_ssh_service()
-
+            if conf.get_regenerate_ssh_host_key():
+                self.reg_ssh_host_key()
+                self.osutil.restart_ssh_service()
+            thumbprint = self.get_ssh_host_key_thumbprint()
             self.write_provisioned()
 
             self.report_event("Provisioning succeeded ({0}s)".format(self._get_uptime_seconds()),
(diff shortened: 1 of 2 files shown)


andyliuliming:
https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/66

hglkrijger:
@vrdmr @narrieta please review

tests/pa/test_provision.py
```diff
@@ -204,6 +205,7 @@ def _provision_test(self,
         ph.protocol_util.osutil = mock_osutil
         ph.protocol_util.get_protocol = MagicMock()
 
+        conf.get_regenerate_ssh_host_key = Mock(return_value=True)
```
narrieta: just a question: why do we make this True here and at line 267? Thanks

boumenot:
The agent updates sshd configuration for non-cloud-init cases. SSHD should be restarted if there is a change in configuration, which is not handled in this change.  If you want to support this scenario fully I think you will need more code.

For example, if the VM is configured as public/private key authentication only and the user deploys the VM with a password protected account SSHD's configuration must be changed and the SSHD service must be restarted.  I don't see this case handled by this change.

tests/pa/test_provision.py
andyliuliming: I set that twice because they're in different test case function.

narrieta: Sorry, I meant to ask why those 2 tests need this set to true. Thanks

andyliuliming:
@boumenot sorry for missing that case, then I think we should judge whether the config file is updated, and then restart it or not, will modify the code to that.

azurelinuxagent/pa/provision/default.py
```diff
@@ -83,9 +83,10 @@ def run(self):
 
             self.provision(ovf_env)
 
-            thumbprint = self.reg_ssh_host_key()
-            self.osutil.restart_ssh_service()
-
+            if conf.get_regenerate_ssh_host_key():
+                self.reg_ssh_host_key()
+                self.osutil.restart_ssh_service()
+            thumbprint = self.get_ssh_host_key_thumbprint()
```
jasonzio: This change probably makes the mainline path slower, since "thumbprint" can't be retrieved from the ssh service until that service is fully up and running. It would be smarter to hold on to the thumbprint and use it in the mainline code path. That is...

```python
thumbprint = ""

if conf.get_regenerate_ssh_host_key():
    thumbprint = self.reg_ssh_host_key()
    self.osutil.restart_ssh_service()
else
    thumbprint = self.get_ssh_host_key_thumbprint()
```

tests/pa/test_provision.py
andyliuliming: that's because in the original code. the test mocked "reg_ssh_host_key" method.
in this PR, it's seperated into two steps, and for reach the second step. we need to mock the get_regenerate_ssh_host_key to true.

andyliuliming:
@boumenot @narrieta found this [issue](https://github.com/Azure/WALinuxAgent/issues/1334) when handling the sshd configuration case.

andyliuliming:
saw the ssh service is actually not restart: https://github.com/Azure/WALinuxAgent/blob/88ce24de47e77efb01395bd76e9819d8b43eddd3/azurelinuxagent/common/osutil/default.py#L1093

so close this PR.

Issue #1316, opened 2018-08-23
CGroups


hglkrijger:
Today cgroups are only enabled for Ubuntu. We need to enable them by default, and enforce limits.

PR #1325, opened 2018-08-29
Enable cgroups by default

hglkrijger:
## Description

Issue #1316 

This is the first commit for #1316, which enables cgroups by default for all distros. It maintains the ability to override distro-specific functionality as part of the OS factory mechanism. 

- move cgroups implementation to the default osutil class
- cleanup the cgroups setup and logging
- update the agent version to 2.2.31.2

---

### PR information

### DIFF
--- a/azurelinuxagent/common/cgroups.py
+++ b/azurelinuxagent/common/cgroups.py
@@ -22,11 +22,11 @@ import time
 from azurelinuxagent.common import logger
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.osutil import get_osutil
+from azurelinuxagent.common.osutil.default import BASE_CGROUPS
 from azurelinuxagent.common.utils import fileutil
 from azurelinuxagent.common.version import AGENT_NAME, CURRENT_VERSION
 
 
-BASE_CGROUPS = '/sys/fs/cgroup'
 WRAPPER_CGROUP_NAME = "Agent+Extensions"
 METRIC_HIERARCHIES = ['cpu', 'memory']
 MEMORY_DEFAULT = -1
@@ -505,7 +505,6 @@ class CGroups(object):
             Add this process to the "agent" cgroup, if required
         Actual collection of metrics from cgroups happens in the -run-exthandlers instance
         """
-        cgroups_enabled = True
         if CGroups.enabled():
             try:
                 CGroups._osutil.mount_cgroups()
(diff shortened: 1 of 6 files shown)


azurelinuxagent/common/cgroups.py
```diff
@@ -517,21 +516,26 @@ def setup(suppress_process_add=False):
                         cg.add(pid)
                         logger.info("Add daemon process pid {0} to {1} cgroup".format(pid, cg.name))
                     else:
-                        logger.info("Daemon process pid {0} cgroup managed by systemd".format(pid))
-                status = "OK"
+                        logger.info("Add daemon process pid {0} to systemd cgroup".format(pid))
+                status = "ok"
```
boumenot: Is the case use anywhere that it is important?

azurelinuxagent/common/osutil/default.py
```diff
@@ -95,6 +95,9 @@
 
 IP_COMMAND_OUTPUT = re.compile('^\d+:\s+(\w+):\s+(.*)$')
 
+BASE_CGROUPS = '/sys/fs/cgroup'
+
+
 class DefaultOSUtil(object):
```
boumenot: If you bring this into DefaultOSUtil I would have expected more edits where you would not implement it, such as FreeBSD, etc.

azurelinuxagent/common/cgroups.py
hglkrijger: DCR

azurelinuxagent/common/osutil/default.py
hglkrijger: This will fail gracefully on BSD or anywhere which does not have the base path or where cgroups cannot be mounted.

boumenot: SGTM.

azurelinuxagent/common/osutil/default.py
```diff
@@ -290,11 +293,54 @@ def is_current_instance_id(self, id_that):
...
+                               chk_err=False)
+
+            for metric_hierarchy in ['cpu', 'cpuacct']:
+                target_path = self._cgroup_path(metric_hierarchy)
+                if not os.path.exists(target_path):
+                    os.symlink(self._cgroup_path('cpu,cpuacct'), target_path)
+
+        except OSError as oe:
+            # log a warning for read-only file systems
+            logger.warn("Could not mount cgroups: {0}", ustr(oe))
```
narrieta: do we need to be more specific when checking the exception? the comment mentions read-only file systems

hglkrijger: I do not believe so, the end result is the same. If there is any kind of exception we will disable cgroups, though some could be expected and therefore are considered warnings. The comment is just an example.

PR #1387, opened 2018-11-02
Do not use a subprocess to capture stdout/stderr

narrieta:
## Description
Do not use a subprocess to capture stdout/stderr; it caused the agent to hang. Instead, read the pipes using select.select() and os.read().

---

### PR information

### DIFF
--- a/azurelinuxagent/common/utils/processutil.py
+++ b/azurelinuxagent/common/utils/processutil.py
@@ -16,17 +16,13 @@
 #
 # Requires Python 2.6+ and Openssl 1.0+
 #
-import multiprocessing
 import select
 import subprocess
-import sys
 import os
 import time
 import signal
 from errno import ESRCH
-from multiprocessing import Process
 
-import azurelinuxagent.common.logger as logger
 from azurelinuxagent.common.exception import ExtensionError
 from azurelinuxagent.common.future import ustr
 
@@ -129,10 +125,10 @@ def capture_from_process_poll(process, cmd, timeout, code):
     stdout = bytearray()
     stderr = bytearray()
 
-    def read_pipe(pipe_instance, buffer):
+    def read_pipe(pipe_instance, read_buffer):
         while True:
             block = os.read(pipe_instance.fileno(), 512)
-            buffer.extend(block)
+            read_buffer.extend(block)
             if len(block) < 512:
                 break
 
--- a/azurelinuxagent/common/utils/processutil.py
+++ b/azurelinuxagent/common/utils/processutil.py
@@ -17,6 +17,7 @@
 # Requires Python 2.6+ and Openssl 1.0+
 #
 import multiprocessing
+import select
 import subprocess
 import sys
 import os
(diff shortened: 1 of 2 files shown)


tests/utils/test_process_util.py
```diff
@@ -122,21 +163,43 @@ def test_process_stdout_stderr(self):
         actual = capture_from_process(process, cmd)
         self.assertEqual(expected, actual)
 
-    def test_process_timeout_non_forked(self):
+    def test_it_should_throw_if_process_is_not_leader_and_timeout_is_given(self):
```
narrieta: this is not a new test... it's an existing test (was test_process_bad_pgid) that I moved here so that it is adjacent to the related 'test_it_should_capture_the_output_if_process_is_not_leader_and_timeout_is_not_given'

azurelinuxagent/common/utils/processutil.py
```diff
@@ -123,36 +120,34 @@ def capture_from_process_poll(process, cmd, timeout, code):
...
-        if len(ret_dict) == 2:
-            stdout = ret_dict[0]
-            stderr = ret_dict[1]
-
-    except Exception:
-        pass
-
-    return stdout, stderr
+    # at this point the parent process has completed and we need to pickup its output; we use select()
+    # with a short timeout to capture the beginning of the output of any child processes.
```
jasonzio: If you wait until child process exit to consume stdout or stderr, you WILL deadlock if the child produces more output than will fit in the socket buffer. There _is_ a limit. When the child process does a write to a full socket, the write() syscall will block; the child process will never exit; the agent will never see process death and will never drain the socket.

There were good reasons for the previous design. This is a bad change.

azurelinuxagent/common/utils/processutil.py
```diff
@@ -16,16 +16,13 @@
 #
 # Requires Python 2.6+ and Openssl 1.0+
 #
-import multiprocessing
+import select
```
vrdmr: We might need to confirm if this is out of box. (I am certain `select` should be)

azurelinuxagent/common/utils/processutil.py
vrdmr: @jasonzio When you say _child process exit_, do you mean the child process of the `ext_handler` (which is the extension code) or the child of the extension itself (fork from an extension, a la custom script approach).
- If it is the extension itself, the code at line#119 is already a blocking call.
- And for the forked process of the extension, the earlier approach was not taking care of cases when the daemon was detaching the parent process (an extension which starts a daemon, but did not successfully detach itself). Thus when it did the `_process.communicate()` in the `proc_comm(_process, _return)`, it would not exit as well ([communicate](https://docs.python.org/2/library/subprocess.html#subprocess.Popen.communicate) being a blocking call).

tests/utils/test_process_util.py
```diff
@@ -15,6 +15,7 @@
 # Requires Python 2.6+ and Openssl 1.0+
 #
 import datetime
+import stat
```
vrdmr: same as above - adding dependency caution.

azurelinuxagent/common/utils/processutil.py
narrieta: DCR would have caught that (if it required installing an extra dependency), I think.

azurelinuxagent/common/utils/processutil.py
narrieta: I still need to verify the extension itself doesn't block... if this turns out to be problematic the other option we are considering is stop capturing output altogether. The current approach of creating a subprocess that calls communicate() is also producing hangs in the agent. I will update this thread.

narrieta:
Closing this PR -- will use a different approach for this.

PR #1391, opened 2018-11-09
Ignore overprovisioning in metadata protocol

adepue:
## Description

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -316,10 +316,13 @@ class ExtHandlersHandler(object):
             return
 
         if conf.get_enable_overprovisioning():
-            artifacts_profile = self.protocol.get_artifacts_profile()
-            if artifacts_profile and artifacts_profile.is_on_hold():
-                logger.info("Extension handling is on hold")
-                return
+            if not self.protocol.supports_overprovisioning():
+                logger.verbose("Overprovisioning is enabled but protocol does not support it.")
+            else:
+                artifacts_profile = self.protocol.get_artifacts_profile()
+                if artifacts_profile and artifacts_profile.is_on_hold():
+                    logger.info("Extension handling is on hold")
+                    return
 
         wait_until = datetime.datetime.utcnow() + datetime.timedelta(minutes=DEFAULT_EXT_TIMEOUT_MINUTES)
         max_dep_level = max([handler.sort_key() for handler in self.ext_handlers.extHandlers])


azurelinuxagent/ga/exthandlers.py
```diff
@@ -315,7 +316,7 @@ def handle_ext_handlers(self, etag=None):
             logger.verbose("No extension handler config found")
             return
 
-        if conf.get_enable_overprovisioning():
+        if conf.get_enable_overprovisioning() and not isinstance(self.protocol, MetadataProtocol):
```
narrieta: maybe instead add a supports_overprovisioning() method to the protocol class? and a warning if the setting is true but the protocol doesn't support it? thanks

adepue: @narrieta Just updated with two commits to cover this.  Let me know what you think.
I debated having the base class raise NotImplemented, but didn't want to have to go to every subclass and override it.  Instead opted for the "everyone supports unless explicitly setting not".  Happy to change it if you prefer

narrieta: thanks @adepue, I think your approach is fine. can add a warning if the setting is true, but the protocol doesn't support it?

After that i can run our automation on your PR. On your side, there is a makepkg.py script on the root of the repo that will create a package for the agent. You can test that package to see if there are any other issues (you need to sudo to execute the script).

adepue: @narrieta Updated to log a warning.  Will test the package on this side as well.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -316,10 +316,13 @@ def handle_ext_handlers(self, etag=None):
             return
 
         if conf.get_enable_overprovisioning():
-            artifacts_profile = self.protocol.get_artifacts_profile()
-            if artifacts_profile and artifacts_profile.is_on_hold():
-                logger.info("Extension handling is on hold")
-                return
+            if not self.protocol.supports_overprovisioning():
+                logger.warn("Overprovisioning is enabled but protocol does not support it.")
```
narrieta: Sorry, my bad! I just realized this is in a loop that executes too frequently so the logs on Azure Stack would be too verbose. Can we change warn() to verbose()?

narrieta: Thank you, and sorry for the extra change!

narrieta:
code looks good; running automation now

Issue #1363, opened 2018-10-12
CGroup limits should be specified outside of the code


vrdmr:
The precursor to #1329.

We have to get out of specifying the Threshold limits in [code](https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/common/cgroups.py#L36-L41) and get it into a conf file. This then can be extended to conf per extension for #1329.

PR #1394, opened 2018-11-13
Fix the threshold telemetry issue

vrdmr:
## Description
Earlier the thresholds were being assigned during the set_cgroups_limits (and being maintained in the Cgroup class, and member of _tracked in CGroupTelemetry). The monitor thread wasn't picking up the data and we were seeing `2018/11/07 21:18:42.654547 WARNING ExtHandler Monitor: failed to collect Cgroups performance metrics: 'NoneType' object has no attribute '__getitem__'`.

This change now explicitly updates the `limits[cgroup_name]` dictionary, which is returned by the `collect_all_tracked()`.

Issue #1395, also does half work for #1363 

---

### PR information

### DIFF
--- a/azurelinuxagent/common/cgroups.py
+++ b/azurelinuxagent/common/cgroups.py
@@ -186,8 +186,6 @@ class CGroupsTelemetry(object):
 
     tracked_names = set()
 
-    _osutil = get_osutil()
-
     @staticmethod
     def metrics_hierarchies():
         return CGroupsTelemetry._hierarchies
@@ -270,24 +268,6 @@ class CGroupsTelemetry(object):
             del (CGroupsTelemetry._tracked[name])
 
     @staticmethod
-    def get_cpu_limits(name):
-        # default values
-        cpu_limit = DEFAULT_CPU_LIMIT_AGENT if AGENT_NAME.lower() in name.lower() else DEFAULT_CPU_LIMIT_EXT
-
-        return cpu_limit
-
-    @staticmethod
-    def get_memory_limits(name):
-        # default values
-        mem_limit = max(DEFAULT_MEM_LIMIT_MIN_MB, round(_osutil.get_total_mem() * DEFAULT_MEM_LIMIT_PCT / 100, 0))
-
-        # agent values
-        if AGENT_NAME.lower() in name.lower():
-            mem_limit = min(DEFAULT_MEM_LIMIT_MAX_MB, mem_limit)
-
-        return mem_limit
-
-    @staticmethod
     def collect_all_tracked():
         """
         Return a dictionary mapping from the name of a tracked cgroup to the list of collected metrics for that cgroup.
@@ -304,8 +284,7 @@ class CGroupsTelemetry(object):
         for cgroup_name, collector in CGroupsTelemetry._tracked.copy().items():
             cgroup_name = cgroup_name if cgroup_name else WRAPPER_CGROUP_NAME
             results[cgroup_name] = collector.collect()
-            limits[cgroup_name] = {"cpu": CGroupsTelemetry.get_cpu_limits(cgroup_name),
-                                   "memory": CGroupsTelemetry.get_memory_limits(cgroup_name)}
+            limits[cgroup_name] = collector.cgroup.threshold
 
         return results, limits
 
@@ -392,12 +371,12 @@ class CGroups(object):
         return os.path.join(BASE_CGROUPS, hierarchy, 'system.slice', cgroup_name).rstrip(os.path.sep)
 
     @staticmethod
-    def for_extension(name):
-        return CGroups(name, CGroups._construct_custom_path_for_hierarchy)
+    def for_extension(name, limits=None):
+        return CGroups(name, CGroups._construct_custom_path_for_hierarchy, limits)
 
     @staticmethod
-    def for_systemd_service(name):
-        return CGroups(name.lower(), CGroups._construct_systemd_path_for_hierarchy)
+    def for_systemd_service(name, limits=None):
+        return CGroups(name.lower(), CGroups._construct_systemd_path_for_hierarchy, limits)
 
     @staticmethod
     def enabled():
@@ -411,14 +390,14 @@ class CGroups(object):
     def enable():
         CGroups._enabled = True
 
-    def __init__(self, name, path_maker):
+    def __init__(self, name, path_maker, limits=None):
         """
         Construct CGroups object. Create appropriately-named directory for each hierarchy of interest.
 
         :param str name: Name for the cgroup (usually the full name of the extension)
         :param path_maker: Function which constructs the root path for a given hierarchy where this cgroup lives
         """
-        if name == "":
+        if not name or name == "":
             self.name = "Agents+Extensions"
             self.is_wrapper_cgroup = True
         else:
@@ -426,7 +405,8 @@ class CGroups(object):
             self.is_wrapper_cgroup = False
 
         self.cgroups = {}
-        self.threshold = None
+
+        self.threshold = CGroupsLimits(self.name, limits)
 
         if not self.enabled():
             return
@@ -503,22 +483,6 @@ class CGroups(object):
             tasks_file = self._get_cgroup_file(hierarchy, 'cgroup.procs')
             fileutil.append_file(tasks_file, "{0}\n".format(pid))
 
-    def get_cpu_limits(self):
-        # default values
-        cpu_limit = DEFAULT_CPU_LIMIT_AGENT if AGENT_NAME.lower() in self.name.lower() else DEFAULT_CPU_LIMIT_EXT
-
-        return cpu_limit
-
-    def get_memory_limits(self):
-        # default values
-        mem_limit = max(DEFAULT_MEM_LIMIT_MIN_MB, round(self._osutil.get_total_mem() * DEFAULT_MEM_LIMIT_PCT / 100, 0))
-
-        # agent values
-        if AGENT_NAME.lower() in self.name.lower():
-            mem_limit = min(DEFAULT_MEM_LIMIT_MAX_MB, mem_limit)
-
-        return mem_limit
-
     def set_limits(self):
         """
         Set per-hierarchy limits based on the cgroup name (agent or particular extension)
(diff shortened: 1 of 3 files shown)


tests/common/test_cgroups.py
```diff
@@ -144,6 +144,13 @@ def exercise_telemetry_instantiation(self, test_cgroup):
             else:
                 self.fail("Unknown metric {0}/{1} value {2}".format(metric_family, metric_name, metric_value))
 
+        my_limits = limits[test_extension_name]
+        for heirarchy, limit in my_limits.items():
```
jasonzio: typo: hierarchy

azurelinuxagent/common/cgroups.py
```diff
@@ -267,6 +267,26 @@ def stop_tracking(name):
...
+
+        # default values
+        mem_limit = max(DEFAULT_MEM_LIMIT_MIN_MB, round(osutil.get_total_mem() * DEFAULT_MEM_LIMIT_PCT / 100, 0))
+
+        # agent values
+        if AGENT_NAME.lower() in name.lower():
+            mem_limit = min(DEFAULT_MEM_LIMIT_MAX_MB, mem_limit)
+
+        return mem_limit
+
```
jasonzio: DRY code smell. You now compute these limits _twice_; once when you collect results, and once when you actually supply the limits to the OS cgroups implementation.

This is not the right way to do it.

azurelinuxagent/common/cgroups.py
```diff
@@ -538,9 +560,6 @@ def set_limits(self):
                 message=msg,
                 log_event=False)
```
jasonzio: Just move the creation of self.threshold into the CGroup constructor and it will always be present, even if that particular instance of CGroup object was never used to set the thresholds.

vrdmr: Do you mean:
1. assign the value in the constructor itself (compute the memory/CPU limits in the constructor and assign, which would be defaults)?
2. I could change this in the constructor of CGroup class - changing `__init__(self, name, path_maker)` to `__init__(self, name, path_maker, limits=None)`, and use it to be populated by the `for_extension(name, limits=None)` call, and if the limits is None, then use approach 1.
3. just to have an empty `{}` 

The previous implementation set the thresholds in the [set_limits()](https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/common/cgroups.py#L235) call, and the [cgroup is tracked](https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/common/cgroups.py#L236) (added to the common data structure _tracked). This was then referred by the monitor thread, but still, it did not work and was giving the `'NoneType' object has no attribute '__getitem__'` error - could be signaling that the cgroup in the [`_tracked`](https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/common/cgroups.py#L180) dictionary might not be populated in the first place with the right Cgroups objects (which has the thresholds assigned.

azurelinuxagent/common/cgroups.py
vrdmr: Yup - I didn't want to do it. Please go over my comment [below](https://github.com/Azure/WALinuxAgent/pull/1394#discussion_r232926231) and I can make this better.

azurelinuxagent/common/cgroups.py
jasonzio: The original line 429:
```
        self.threshold = None
```
Replace with:
```    
        cpu_limit = self.get_cpu_limits()
        mem_limit = self.get_memory_limits()
        self.threshold = {"cpu": cpu_limit, "memory": mem_limit}
```
These are default values, but you have no mechanism for defining non-default values anyway.

Another alternative: remove CGroups.threshold as a value and replace it with CGroups.threshold() as a method. When this function is called, the CGroups object computes thresholds right then and there. Your initial implementation would still return only defaults. A subsequent implementation (one which, for example, did file I/O to discover a per-extension setting) would only do that work when (and if) the threshold() method was called.

vrdmr: Cleaned the implementation - Moved the limits into its [own class](https://github.com/Azure/WALinuxAgent/pull/1394/files#diff-0d947136d13fb1ee11231cfa47a36890R784), which can be extended in future if other new limits need to be added (IO, etc.) and enabled the plumbing of the limits to be set per CGroup.

azurelinuxagent/common/cgroups.py
```diff
@@ -404,7 +405,9 @@ def __init__(self, name, path_maker):
             self.is_wrapper_cgroup = False
 
         self.cgroups = {}
-        self.threshold = None
+
+        self.threshold = CGroupsLimits(self.name, cpu_threshold=limits["cpu"],
+                                       memory_threshold=limits["memory"]) if limits else CGroupsLimits(self.name)
```
jasonzio: This will throw an exception if limits["cpu"] or limits["memory"] isn't defined. The problem is the design of the CGroupsLimits constructor; see my comment there.

azurelinuxagent/common/cgroups.py
```diff
@@ -796,3 +779,27 @@ def set_memory_limit(self, limit=None, unit='megabytes'):
             fileutil.write_file(memory_limit_file, '{0}\n'.format(value))
         else:
             raise CGroupsException("Memory hierarchy not available in this cgroup")
+
+
+class CGroupsLimits(object):
+    def __init__(self, cgroup_name, cpu_threshold=None, memory_threshold=None):
+        self.cpu_limit = cpu_threshold if cpu_threshold else self.get_default_cpu_limits(cgroup_name)
+        self.memory_limit = memory_threshold if memory_threshold else self.get_default_cpu_limits(cgroup_name)
```
jasonzio: The signature of the constructor changes if you add a new limit. Combine that with my earlier comment in the CGroups constructor, and I have to conclude the design of this constructor isn't right.

Suggestion:
```
def __init__(self, cgroup_name, thresholds=none):
    self.cpu_limit = thresholds["cpu"] if thresholds and "cpu" in thresholds else self.get_default_cpu_limits(cgroup_name)
    self.memory_limit = thresholds["memory"] if thresholds and "memory" in thresholds else self.get_default_memory_limits(cgroup_name)
```

vrdmr: Yes. Corrected this. The new constructor of `CGroupsLimits` now takes the dictionary and thus keeps it cleaner.

azurelinuxagent/common/cgroups.py
```diff
@@ -796,3 +778,27 @@ def set_memory_limit(self, limit=None, unit='megabytes'):
             fileutil.write_file(memory_limit_file, '{0}\n'.format(value))
         else:
             raise CGroupsException("Memory hierarchy not available in this cgroup")
+
+
+class CGroupsLimits(object):
+    def __init__(self, cgroup_name, threshold=None):
+        self.cpu_limit = threshold["cpu"] if threshold else self.get_default_cpu_limits(cgroup_name)
+        self.memory_limit = threshold["memory"] if threshold else self.get_default_cpu_limits(cgroup_name)
```
jasonzio: Nope - you need `if threshold and "cpu" in threshold`. It should be legit to pass a hash that has a setting for only one value, thus asking for the default for the other value. If/when you add a third threshold, your first implementation task would be to add a new get_default call, but that's *all* you'd have to do; every existing use of a CGroupsLimits object would pass a hash that had no value for the new limit, and the `if threshold and "newlimit" in threshold` test would do the right thing.

vrdmr: Good catch. Thanks.

azurelinuxagent/ga/monitor.py
```diff
@@ -427,9 +427,11 @@ def send_cgroup_telemetry(self):
                             report_metric(metric_group, metric_name, cgroup_name, value)
 
                         if metric_group == "Memory":
-                            if value >= thresholds["memory"]:
-                                msg = "CGroup {0}: Crossed the Memory Threshold. Current Value:{1}, Threshold:{2}.".format(
-                                    cgroup_name, value, thresholds["memory"])
+                            # Memory is collected in bytes, and limit is set in megabytes.
+                            if value >= CGroups._format_memory_value('megabytes', thresholds.memory_limit):
```
vrdmr: Added this conversion (MBs to Bytes) to keep the units same.

tests/common/test_cgroups.py
```diff
@@ -17,8 +17,8 @@
 
 from __future__ import print_function
 
-from azurelinuxagent.common.cgroups import CGroupsTelemetry, CGroups, CGroupsException, BASE_CGROUPS, Cpu, Memory, \
-    DEFAULT_MEM_LIMIT_MIN_MB
+from azurelinuxagent.common.cgroups import CGroupsTelemetry, CGroups, CGroupsException, CGroupsLimits, BASE_CGROUPS, \
+    Cpu, Memory, DEFAULT_MEM_LIMIT_MIN_MB
```
jasonzio: Add unit tests for CGroupsLimits that exercise

- missing threshold map
- partial threshold map
- fully specified threshold map
- empty threshold map

vrdmr: Done. Thanks for reminding me to do this.

azurelinuxagent/common/cgroups.py
```diff
@@ -389,22 +390,23 @@ def disable():
...
-    def __init__(self, name, path_maker):
+    def __init__(self, name, path_maker, limits=None):
         """
         Construct CGroups object. Create appropriately-named directory for each hierarchy of interest.
 
         :param str name: Name for the cgroup (usually the full name of the extension)
         :param path_maker: Function which constructs the root path for a given hierarchy where this cgroup lives
         """
-        if name == "":
+        if name or name == "":
```
jasonzio: Are you sure about that? Did you mean
```
if not name or name == "":
```

azurelinuxagent/common/cgroups.py
```diff
@@ -796,3 +778,32 @@ def set_memory_limit(self, limit=None, unit='megabytes'):
...
+
+
+class CGroupsLimits(object):
+    def __init__(self, cgroup_name, threshold=None):
+        if not cgroup_name or cgroup_name == "":
+            cgroup_name = "Agents+Extensions"
+
+        self.cpu_limit = threshold["cpu"] if threshold and "cpu" in threshold else self.get_default_cpu_limits(
+            cgroup_name)
+        self.memory_limit = threshold["memory"] if threshold and "memory" in threshold else self.get_default_memory_limits(
```
jasonzio: Suggestion (feel free to ignore me): I'm a big fan of DRY. These two lines are mostly repeated; if you added a third limit, it would be mostly a repeat of these. Consider refactoring to extract the common code:

```
    def _get_value_or_default(threshold, limit, compute_default):
        return threshold[limit] if threshold and limit in threshold else compute_default(self, self.cgroup_name)
```

That makes these two lines look like:
```
    self.cpu_limit = self._get_value_or_default(threshold, "cpu", CGroupsLimits.get_default_cpu_limits)
    self.memory_limit = self._get_value_or_default(threshold, "memory", CGroupsLimits.get_default_memory_limits)
```

As I said, this is just a suggestion.

Issue #1396, opened 2018-11-14
Not reporting failure events for extension install/enable/disable


vrdmr:
The extension commands are run in `launch_command` call in [`enable(self)`](https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/ga/exthandlers.py#L828)/[`disable(self)`](https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/ga/exthandlers.py#L837)/[`install(self)`](https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/ga/exthandlers.py#L847). 

If it throws an `ExtensionError` (Which is when return code is non-0 or some other issue), the error is propogated up - eg. `launch_command` ->`enable` ->`handle_enable` -> `handle_ext_handler()`, where the [`except ExtensionError as e`](https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/ga/exthandlers.py#L419) catches it, and gives it to the [`handle_handle_ext_handler_error`](https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/ga/exthandlers.py#L424) method

Today [`handle_handle_ext_handler_error`](https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/ga/exthandlers.py#L424) does not report `report_event` for all the failures - only when `get_artifact_error_state.is_triggered()`. This is not the right behavior - It should be sneding events for all the exceptions and not only when the `get_artifact_error_state.is_triggered()`.

jasonzio:
The intention behind the "is_triggered" thing is to avoid flooding the logs when things go bad. The idea was for the event to start out "triggered" and be "reset" the first time it's seen; periodically, all events were reset (in some other thread) to "triggered". If the initial state is erroneously set (i.e. to the "reset" state) or if the periodic reset (a) never happens or (b) actually sets the trigger state incorrectly, then that's the bug.

You'll want to be very careful of the event volume you see after you've made this change. You may need to revert it in a hurry.

vrdmr:
+@hglkrijger, @boumenot - If I am missing some context, please let me know.

[...]

PR #1397, opened 2018-11-15
Send events when extensions fail to complete operation

vrdmr:
## Description
- Fixed the error reporting when the extension fails to do an operation
- Changed the logging event table

Issue #1396 

---

### PR information

### DIFF
--- a/azurelinuxagent/common/event.py
+++ b/azurelinuxagent/common/event.py
@@ -262,6 +262,7 @@ class EventLogger(object):
         if (not is_success) and log_event:
             _log_event(name, op, message, duration, is_success=is_success)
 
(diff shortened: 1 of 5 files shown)


jasonzio:
The intention behind the "is_triggered" thing is to avoid flooding the logs when things go bad. The idea was for the event to start out "triggered" and be "reset" the first time it's seen; periodically, all events were reset (in some other thread) to "triggered". If the initial state is erroneously set (i.e. to the "reset" state) or if the periodic reset (a) never happens or (b) actually sets the trigger state incorrectly, then that's the bug.
You'll want to be very careful of the event volume you see after you've made this change. You may need to revert it in a hurry.

PR #1408, opened 2018-11-29
[DoNotMerge] Honoring the 'no_proxy' flag

vrdmr:
## Description
Honoring the No_Proxy environment variable now. Need tests. Will update the PR with the tests.

Issue #1382  

---

### PR information
- [ ] The title of the PR is clear and informative.
- [ ] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).
- [ ] Except for special cases involving multiple contributors, the PR is started from a fork of the main repository, not a branch.
- [ ] If applicable, the PR references the bug/issue that it fixes in the description.
- [ ] New Unit tests were added for the changes made and Travis.CI is passing.

### Quality of Code and Contribution Guidelines
- [ ] I have read the [contribution guidelines](https://github.com/Azure/WALinuxAgent/blob/master/.github/CONTRIBUTING.md).

Issue #1382, opened 2018-10-29
Selective proxy usage?


kaosmonk:
Hi,

So now the waagent respects the `http_proxy` and `https_proxy` env vars in order to use a proxy or not. And if the `HttpProxy.Host` and `HttpProxy.Port` configuration variables are provided they will override whatever value is set by the aforementioned env vars.

But for some use cases I'd prefer for waagent to respect the `no_proxy` global env var as well so that certain requests made by waagent that match whatever domain set with that env var are not pushed via the proxy.

Can this be easily implemented and/or achieved in any other way?

Thank you in advance!

vrdmr:
Hi @kaosmonk 
We have added this to our backlog. If you can send in a PR, we can take a look at it as well. Please let me know if you need more information.

Thanks.

PR #1415, opened 2018-12-06
Work for no_proxy not set cases as well

vrdmr:
## Description
This case for missed in #1408. Adding the fix for the No_proxy not available flag.

Issue #1382 

---

### PR information

### DIFF
--- a/tests/utils/test_rest_util.py
+++ b/tests/utils/test_rest_util.py
@@ -454,6 +454,54 @@ class TestHttpOperations(AgentTestCase):
             host = "10.0.1.1"
             self.assertEquals(should_use_proxy, use_proxy and not restutil.bypass_proxy(host))
 
+        # When No_proxy is empty
+        with patch.dict(os.environ, {
+            'no_proxy': ""
+        }):
+            host = "10.0.0.1"
+            self.assertTrue(use_proxy and not restutil.bypass_proxy(host))
+
+            host = "foo.com"
+            self.assertTrue(use_proxy and not restutil.bypass_proxy(host))
+
+            host = "www.google.com"
+            self.assertTrue(use_proxy and not restutil.bypass_proxy(host))
+
+            host = "168.63.129.16"
+            self.assertTrue(use_proxy and not restutil.bypass_proxy(host))
+
+            host = "www.bar.com"
+            self.assertTrue(use_proxy and not restutil.bypass_proxy(host))
+
+            host = "10.0.0.1"
+            self.assertTrue(use_proxy and not restutil.bypass_proxy(host))
+
+            host = "10.0.1.1"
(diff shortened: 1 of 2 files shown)


tests/utils/test_rest_util.py
```diff
@@ -454,6 +454,34 @@ def test_proxy_conditions_with_no_proxy(self):
             host = "10.0.1.1"
             self.assertEquals(should_use_proxy, use_proxy and not restutil.bypass_proxy(host))
 
+    def test_proxy_conditions_with_no_proxy_not_set(self):
+        should_use_proxy = True
+        use_proxy = True
+
+        host = "10.0.0.1"
```
narrieta: I think the test should explicitly express that the env variable is not defined (as is, it is actually relying on the current process env)

tests/utils/test_rest_util.py
```diff
@@ -454,6 +454,34 @@ def test_proxy_conditions_with_no_proxy(self):
...
+        self.assertEquals(should_use_proxy, use_proxy and not restutil.bypass_proxy(host))
+
+        host = "www.bar.com"
+        self.assertEquals(should_use_proxy, use_proxy and not restutil.bypass_proxy(host))
+
+        host = "10.0.0.1"
+        self.assertEquals(should_use_proxy, use_proxy and not restutil.bypass_proxy(host))
+
+        host = "10.0.1.1"
+        self.assertEquals(should_use_proxy, use_proxy and not restutil.bypass_proxy(host))
```
narrieta: I know this is just a copy of the test above, but the assert is kinda hard to read.. why not simply AssertTrue (or AssertFalse)

tests/utils/test_rest_util.py
```diff
@@ -454,6 +454,34 @@ def test_proxy_conditions_with_no_proxy(self):
...
+        host = "www.google.com"
+        self.assertEquals(should_use_proxy, use_proxy and not restutil.bypass_proxy(host))
+
+        host = "168.63.129.16"
+        self.assertEquals(should_use_proxy, use_proxy and not restutil.bypass_proxy(host))
+
+        host = "www.bar.com"
+        self.assertEquals(should_use_proxy, use_proxy and not restutil.bypass_proxy(host))
+
+        host = "www.bar.com"
```
pgombar: Why is this test exactly the same as the test above? Was the flag supposed to change in the meantime?

tests/utils/test_rest_util.py
vrdmr: I just wanted to make it explicit that proxy should be used in these cases (when no_proxy is not set). I added a comment and refactored to make it cleaner.

tests/utils/test_rest_util.py
vrdmr: Cleaned the tests. I didn't need a new test. Just added new asserts to the previous test.

Issue #1344, opened 2018-09-21
utf-8 encoding required for Linux Agent


larohra:
Error being thrown:
 ERROR ExtHandler [ProtocolError] [Wireserver Exception] 'latin-1' codec can't encode characters in position 1207-1212: Body ('\u691c') is not valid Latin-1. Use body.encode('utf-8') if you want to send it encoded in UTF-8.

narrieta:
We need to understand impact.

rebekahHess:
When a Linux VM with Japanese characters in the resource group or vm name connects to serial console, the following Error message is printed repeatedly in the terminal:
2018/09/21 18:18:06.310566 ERROR ExtHandler [ProtocolError] [Wireserver Exception] 'latin-1' codec can't encode characters in position 1207-1212: Body ('\u691c\u7d22\u30b7\u30b9\u30c6\u30e0') is not valid Latin-1. Use body.encode('utf-8') if you want to send it encoded in UTF-8.
This interferes with serial console usage as the user may be typing commands in the terminal, and these error messages are suddenly injected.

larohra:
+Azure Linux Agent

PR #1416, opened 2018-12-06
Fix utf-8 encoding for telemetry payload

pgombar:
## Description

Added utf-8 encoding when sending telemetry payload, compatible with Python 2 and 3. Added unit test.

Issue #1344  

---

### PR information

### DIFF
--- a/tests/protocol/test_wire.py
+++ b/tests/protocol/test_wire.py
@@ -457,7 +457,7 @@ class TestWireProtocol(AgentTestCase):
         headers = kwargs['headers']
 
         self.assertTrue("utf-8" in headers['Content-Type'])
-        self.assertIn(event_str.encode("utf-8"), body_received)
+        self.assertIn(str(event_str.encode("utf-8")), body_received)
 
 
 class MockResponse:
--- a/tests/protocol/test_wire.py
+++ b/tests/protocol/test_wire.py
@@ -1,3 +1,4 @@
+# -*- encoding: utf-8 -*-
 # Copyright 2018 Microsoft Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


tests/protocol/test_wire.py
```diff
@@ -442,6 +443,29 @@ def test_report_vm_status(self, *args):
...
+        args, kwargs = first_call
+        method, url, data_received = args
+        headers = kwargs['headers']
+
+        data_format = ('<?xml version="1.0"?>'
+                       '<TelemetryData version="1.0">'
+                       '<Provider id="{0}">{1}'
+                       '</Provider>'
+                       '</TelemetryData>')
+        data_encoded = data_format.format("foo", event_str.encode("utf-8"))
```
narrieta: I think the test has too much knowledge of the internal implementation of send_event (it needs to know the exact format of the xml), which makes it fragile in the sense that if the implementation changes the test may break. Maybe using a longer body (a longer test string) and contains() instead of strict equality is a better option?

tests/protocol/test_wire.py
```diff
@@ -442,6 +443,29 @@ def test_report_vm_status(self, *args):
...
+    def test_send_event(self, mock_http_request, *args):
+        mock_http_request.return_value = MockResponse("", 200)
+
+        event_str = u'ćevapčići'
+        client = WireProtocol(wireserver_url).client
+        client.send_event("foo", event_str)
+
+        first_call = mock_http_request.call_args_list[0]
+        args, kwargs = first_call
+        method, url, data_received = args
```
narrieta: minor: i would rename data_received to body_received (request_body?) or add a comment to make it clearer that we are checking the body of the request

tests/protocol/test_wire.py
pgombar: Agreed, changed test to be more robust.

tests/protocol/test_wire.py
```diff
@@ -1,3 +1,4 @@
+# -*- encoding: utf-8 -*-
```
vrdmr: I didn't get this line. Copyright should be the first line.

pgombar: This line is needed in the beginning of the file to set the encoding of the source file itself. Without this, I couldn't have the non-utf characters in the test, Python complains about them.

PR #1425, opened 2018-12-13
Fix is_current_instance_id to work with newer kernels

### LINKED ISSUE #1409: Bug: is_current_instance_id() should be performing a case insensitive check
Newer Linux kernels, such as 4.19, present the VM unique ID as lowercase.  The commit and explanation can be found here: https://github.com/torvalds/linux/commit/712ff25450bd01366301eef81c33e865d901e7b7#diff-f2bd14bc67b5e2da67116bca971bbd0b

The current is_current_instance_id() does a direct comparison on this value from a previously known value, so this change will trip up the Azure agent if a kernel update comes down the includes this patch.

jeking3:
## Description

UUID string comparison should be case-insensitive.

This fixes #1409 

---

### PR information
- [X] Except for special cases involving multiple contributors, the PR is started from a fork of the main repository, not a branch.
- [X] If applicable, the PR references the bug/issue that it fixes in the description.
- [X] New Unit tests were added for the changes made and Travis.CI is passing.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1425)

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -291,8 +291,10 @@ class DefaultOSUtil(object):
         may have been persisted using the incorrect byte ordering.
         '''
         id_this = self.get_instance_id()
-        return id_that == id_this or \
-            id_that == self._correct_instance_id(id_this)
+        logger.verbose("current instance id: {0}".format(id_this))
+        logger.verbose(" former instance id: {0}".format(id_that))
+        return id_this.lower() == id_that.lower() or \
+            id_this.lower() == self._correct_instance_id(id_that).lower()
 
     @staticmethod
     def is_cgroups_supported():
--- a/tests/common/osutil/test_default.py
+++ b/tests/common/osutil/test_default.py
@@ -506,6 +506,9 @@ Match host 192.168.1.2\n\
         self.assertEqual(
             "D0DF4C54-4ECB-4A4B-9954-5BDF3ED5C3B8",
             util._correct_instance_id("544CDFD0-CB4E-4B4A-9954-5BDF3ED5C3B8"))
+        self.assertEqual(
+            "d0df4c54-4ecb-4a4b-9954-5bdf3ed5c3b8",
+            util._correct_instance_id("544cdfd0-cb4e-4b4a-9954-5bdf3ed5c3b8"))
 
     @patch('os.path.isfile', return_value=True)
     @patch('azurelinuxagent.common.utils.fileutil.read_file',
@@ -562,11 +565,23 @@ Match host 192.168.1.2\n\
     def test_is_current_instance_id_from_file(self, mock_read, mock_isfile):
         util = osutil.DefaultOSUtil()
 
+        mock_read.return_value = "11111111-2222-3333-4444-556677889900"
+        self.assertFalse(util.is_current_instance_id(
+            "B9F3C233-9913-9F42-8EB3-BA656DF32502"))
+
         mock_read.return_value = "B9F3C233-9913-9F42-8EB3-BA656DF32502"
         self.assertTrue(util.is_current_instance_id(
             "B9F3C233-9913-9F42-8EB3-BA656DF32502"))
 
         mock_read.return_value = "33C2F3B9-1399-429F-8EB3-BA656DF32502"
+        self.assertTrue(util.is_current_instance_id(
+            "B9F3C233-9913-9F42-8EB3-BA656DF32502"))
+
+        mock_read.return_value = "b9f3c233-9913-9f42-8eb3-ba656df32502"
+        self.assertTrue(util.is_current_instance_id(
+            "B9F3C233-9913-9F42-8EB3-BA656DF32502"))
+
+        mock_read.return_value = "33c2f3b9-1399-429f-8eb3-ba656df32502"
         self.assertTrue(util.is_current_instance_id(
             "B9F3C233-9913-9F42-8EB3-BA656DF32502"))
 


asofron:
Hi @jeking3 ,

I tested the pr with Azure image Canonical UbuntuServer  16.04-DAILY-LTS latest and Kernel linux-next-4.20.0-rc6-14cf8c1d5b90_10122018 and the issue still is there . 

Reproduction steps : 

- Provision ubuntu azure vm 
- Install walinuxagent from PR
- Install new kernel 
- reboot ( here the UUID is changed uuid)
before kernel install: B016A6CF-B90D-B743-8379-902BBF5E0037
after kernel install: b016a6cf-b90d-b743-8379-902bbf5e0037
reboot 
reboot again
cloud-init disables the username password authentication

Also the vm is provisioned with  /etc/waagent.conf :
Provisioning.Enabled=n
Provisioning.UseCloudInit=y
Provisioning.DeleteRootPassword=y
Provisioning.RegenerateSshHostKeyPair=n
Provisioning.SshHostKeyPairType=rsa
Provisioning.MonitorHostName=n
Provisioning.DecodeCustomData=n
Provisioning.ExecuteCustomData=n
#Provisioning.PasswordCryptId=6
#Provisioning.PasswordCryptSaltLength=10
Provisioning.AllowResetSysUser=n
# EnableOverProvisioning=y

jeking3:
Did you check the log for:
```
                logger.warn("VM is provisioned, "
                            "but the VM unique identifier has changed -- "
                            "clearing cached state")
```
So that we know beyond any doubt it was this code path it went down?

jeking3:
I pushed another update which also includes verbose level logging inside is_current_instance_id.  Would you be able to re-run this test with the latest code here?

vrdmr:
@jeking3 - Can you add the smaller fix for 2.6 done. It should be a small one.
is_current_instance_id > logger.verbose("id_this: {}".format(id_this))

jeking3:
Ah, okay, I see now. :)

azurelinuxagent/common/osutil/default.py
```diff
@@ -85,7 +85,7 @@
 DMIDECODE_CMD = 'dmidecode --string system-uuid'
 PRODUCT_ID_FILE = '/sys/class/dmi/id/product_uuid'
 UUID_PATTERN = re.compile(
-    r'^\s*[A-F0-9]{8}(?:\-[A-F0-9]{4}){3}\-[A-F0-9]{12}\s*$',
+    r'^\s*[a-fA-F0-9]{8}(?:\-[a-fA-F0-9]{4}){3}\-[a-fA-F0-9]{12}\s*$',
```
vrdmr: Is there a need to do `a-f` explicitly, when the `re.IGNORECASE` flag is also passed?

tests/common/osutil/test_default.py
```diff
@@ -570,6 +577,14 @@ def test_is_current_instance_id_from_file(self, mock_read, mock_isfile):
         self.assertTrue(util.is_current_instance_id(
             "B9F3C233-9913-9F42-8EB3-BA656DF32502"))
 
+        mock_read.return_value = "b9f3c233-9913-9f42-8eb3-ba656df32502"
+        self.assertTrue(util.is_current_instance_id(
+            "B9F3C233-9913-9F42-8EB3-BA656DF32502"))
+
+        mock_read.return_value = "33c2f3b9-1399-429f-8eb3-ba656df32502"
+        self.assertTrue(util.is_current_instance_id(
```
vrdmr: This would fail - It should be `assertFalse`

jeking3: Why is that?  It's just the lowercase variant of the test on new line 576?

vrdmr: I assume you are trying to match Guids, but they don't match - `33c2f3b9-1399-429f-8eb3-ba656df32502` wont match `B9F3C233-9913-9F42-8EB3-BA656DF32502`. Please let me know if I am missing something.

jeking3: is_current_instance_id checks for an exact match or for an endian conversion match on the first three components of the GUID to deal with ambiguity in interpreting the specification... look at the other comparisons that exist in the file.  Did I enter the endian swapped data wrong?

azurelinuxagent/common/osutil/default.py
```diff
@@ -289,10 +289,13 @@ def is_current_instance_id(self, id_that):
         '''
         Compare two instance IDs for equality, but allow that some IDs
         may have been persisted using the incorrect byte ordering.
+        Ensure conformity with RFC-4122 allowing for case insensitivity.
         '''
         id_this = self.get_instance_id()
-        return id_that == id_this or \
-            id_that == self._correct_instance_id(id_this)
+        logger.verbose("id_this: {0}".format(id_this))
```
narrieta: Are these 2 log statements leftovers from debugging? If they are meant to actually be logged, could we improve the message. Thanks

narrieta:
Running automation

azurelinuxagent/common/osutil/default.py
jeking3: I added those following the initial rejection claiming the fix wasn't working.  Yes, we could make the message better.  Let me know the results of the automation before I open this up again.

narrieta: OK, thanks

narrieta:
Automation is OK

jeking3:
Does the automation suite test the kernel upgrade scenario which is the mechanism by which the original issue was caught?

narrieta:
@jeking3 No, the automation does not cover that scenario.

azurelinuxagent/common/osutil/default.py
narrieta: Sorry, I missed this conversation before... we should remove this change, since it is already done by re.IGNORECASE

jeking3: I agree, however without an automation test that exposes the original issue and proves this fixes it, I don't know how to proceed here.  Clearly someone ran a test, supposedly against code with the fix, and it didn't resolve it.  I have no way of knowing the new code was actually used in the test however.  The general steps to reproduce were written, but not exact.

narrieta: Did you have an scenario that was failing due to this issue when you submitted your PR? Maybe you can verify that way?

@asofron is an external contributor. Andrei, maybe you can repeat your test with the logging info that James added? Thanks

asofron: Will check and tell results  @narrieta

jeking3:
I resolved the two issues discussed in the review.

narrieta:
@jeking3 Thanks for the updates, running automation

azurelinuxagent/common/osutil/default.py
asofron: With lastest linux-next kernel issue is gone

narrieta: @asofron - Thanks for checking!

narrieta:
With the new changes automation is failing: the agent fails to start.... I am in the middle of an agent release, will investigate further after that.

narrieta:
Automation is OK

Issue #1426, opened 2018-12-14
Error while testing SLES 15SP1


ricardobranco777:
![2932](https://user-images.githubusercontent.com/10237751/50007810-66dfba80-ffb2-11e8-8212-1ab96727cd4e.png)

The naive version check in `int(distro_version) > 11` is causing trouble.

ricardobranco777:
Fix:

https://github.com/Azure/WALinuxAgent/pull/1427

pgombar:
Hi Ricardo, I just approved your PR (#1427), thanks for the fix. The fix will be published in the new release in January.

PR #1427, opened 2018-12-14
Use Version() for version check

ricardobranco777:
## Description

Issue #1426

This PR uses Version() as used elsewhere in the code to check for the SLES version.


pgombar:
Our end-to-end tests are passing. Thanks for your contribution!

Issue #1357, opened 2018-10-09
Review stderr/stdout capture


narrieta:
We added code to capture the stderr/stdout of extensions on v2.2.32. While investigating an install failure of the VMAccess extension (couldn't import waagent) I noticed that the agent's log did not include stderr/sdtout for that operation.

narrieta:
If extension returns non-zero, we throw without the adding the captured stderr. Will be addressed as part of the stdout/stderr cleanup.

PR #1434, opened 2018-12-31
Use files instead of pipes to capture stdout/stderr

narrieta:
* Use files instead of pipes to capture stdout/stderr
* Test fixes for ExtensionTestCase (test_extension.py)
* Added a few assert methods to AgentTestCase (tools.py)

Addresses #1357 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1434)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -1068,55 +1070,79 @@ def is_responsive(heartbeat_file):
...
+                return msg
+
+    @staticmethod
+    def _capture_process_output(process, stdout_file, stderr_file, cmd, timeout, code=-1):
+        retry = timeout
+        while retry > 0 and process.poll() is None:
+            time.sleep(1)
+            retry -= 1
+
+        def read_output():
```
pgombar: As discussed, it would be a good idea to use try/catch for this method in the future.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1068,55 +1070,79 @@ def is_responsive(heartbeat_file):
     def launch_command(self, cmd, timeout=300, extension_error_code=1000, env=None):
         begin_utc = datetime.datetime.utcnow()
         self.logger.verbose("Launch command: [{0}]", cmd)
+
         base_dir = self.get_base_dir()
 
-        if env is None:
-            env = {}
-        env.update(os.environ)
+        with tempfile.TemporaryFile(dir=base_dir, mode="w+b") as stdout:
```
vrdmr: Something to discuss - I would recommend redirecting the stdout/stderr in the `/var/log/azure/(EXTN)/` directory - If in case the file becomes too large, the user should be able to delete the file forcefully, and need not be root.

azurelinuxagent/common/utils/processutil.py
```diff
@@ -78,7 +68,7 @@ def to_s(captured_stdout, stdout_offset, captured_stderr, stderr_offset):
         return to_s(stdout, -1*max_len_each, stderr, -1*max_len_each)
 
 
-def _destroy_process(process, signal_to_send=signal.SIGKILL):
+def destroy_process(process, signal_to_send=signal.SIGKILL):
```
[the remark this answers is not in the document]
narrieta: Thanks for pointing this out, @vrdmr. I reviewed the code and confirmed the functionality on this function is no longer needed so I removed it.

azurelinuxagent/ga/exthandlers.py
narrieta: Discussed offline - the log directory is also restricted to root (for writes), and also deleting the file has no effect since the extension still holds a handle to it. When the extension completes, the file will be deleted by the OS.

azurelinuxagent/ga/exthandlers.py
narrieta: Thanks, @pgombar. Added error handling and also limited the amount of data we load into memory.

narrieta:
@vrdmr @pgombar - I addressed your comments in a new commit, would you take a look at the diffs? Thanks

PR #1463, opened 2019-02-13
Improve resilience of extension downloads

narrieta:
## Description

- Expanding the extension zip should be done inside the loop that tries the different uris
- Added 5 tries at 1 min interval
- Unzipping a bad file throws Exception, not IoError
- Fixed cleanup logic when a failure occurs

---

### PR information

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -746,7 +746,7 @@ class ExtHandlerInstance(object):
                   op=self.operation, is_success=is_success, duration=duration, log_event=log_event)
 
     def _download_extension_package(self, source_uri, target_file):
-        self.logger.info("Downloading extension package:", source_uri)
+        self.logger.info("Downloading extension package: {0}", source_uri)
         try:
             if not self.protocol.download_ext_handler_pkg(source_uri, target_file):
                 raise Exception("Failed to download extension package - no error information is available")
@@ -771,7 +771,6 @@ class ExtHandlerInstance(object):
 
     def download(self):
         begin_utc = datetime.datetime.utcnow()
-        self.logger.info("Downloading extension package")
         self.set_operation(WALAEventOperation.Download)
 
         if self.pkg is None or self.pkg.uris is None or len(self.pkg.uris) == 0:
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -75,6 +75,7 @@ DEFAULT_EXT_TIMEOUT_MINUTES = 90
 
 AGENT_STATUS_FILE = "waagent_status.json"
 
+NUMBER_OF_DOWNLOAD_RETRIES = 5
 
 def get_traceback(e):
     if sys.version_info[0] == 3:
@@ -450,6 +451,7 @@ class ExtHandlersHandler(object):
         if handler_state == ExtHandlerState.NotInstalled:
             ext_handler_i.set_handler_state(ExtHandlerState.NotInstalled)
             ext_handler_i.download()
+            ext_handler_i.initialize()
             ext_handler_i.update_settings()
             if old_ext_handler_i is None:
                 ext_handler_i.install()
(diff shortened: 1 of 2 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -743,54 +745,84 @@ def report_event(self, message="", is_success=True, duration=0, log_event=True):
...
 
-                if os.path.exists(destination):
-                    file_downloaded = True
-                    self.pkg_file = destination
-                    break
-                else:
-                    file_downloaded = self.protocol.download_ext_handler_pkg(uri.uri, destination)
+        if not package_exists:
+            downloaded = False
+            i = 0
```
pgombar: nit: a better name for this variable might be "retry_count"?

tests/ga/test_exthandlers_download_extension.py
```diff
@@ -0,0 +1,203 @@
...
+
+        mock_download_ext_handler_pkg.assert_called_once()
+
+        self._assert_download_and_expand_succeeded()
+
+    def test_it_should_use_alternate_uris_when_download_fails(self):
+        self.download_failures = 0
+
+        def download_ext_handler_pkg(_uri, destination):
+            if self.download_failures < 3:
```
pgombar: Can we use NUMBER_OF_DOWNLOAD_RETRIES - 2 instead of 3? It seems like an arbitrary number otherwise.

tests/ga/test_exthandlers_download_extension.py
```diff
@@ -0,0 +1,203 @@
...
+
+        self.assertEquals(mock_download_ext_handler_pkg.call_count, self.download_failures + 1)
+
+        self._assert_download_and_expand_succeeded()
+
+    def test_it_should_use_alternate_uris_when_download_raises_an_exception(self):
+        self.download_failures = 0
+
+        def download_ext_handler_pkg(_uri, destination):
+            if self.download_failures < 3:
```
pgombar: Same as above.

tests/ga/test_exthandlers_download_extension.py
```diff
@@ -0,0 +1,203 @@
...
+
+        self.assertEquals(mock_download_ext_handler_pkg.call_count, self.download_failures + 1)
+
+        self._assert_download_and_expand_succeeded()
+
+    def test_it_should_use_alternate_uris_when_it_downloads_an_invalid_package(self):
+        self.download_failures = 0
+
+        def download_ext_handler_pkg(_uri, destination):
+            if self.download_failures < 3:
```
pgombar: Same as above.

tests/ga/test_exthandlers_download_extension.py
pgombar: Or is this related to the number of uris you initialize in the test case set up method? In that case, I would also refer to len(self.pkg.uris).

narrieta:
@pgombar 

> 2\. we download a corrupt zip every time (the publisher messed up) and we fail to expand.
    Currently, we are not explicitly addressing 2). Would it be worth it to add another type of exception for this case?

We are addressing both cases (with the same exception). The log will contain the specific errors. There is a different work item to add telemetry for download failures; over there it may be useful to differentiate both cases.

narrieta:
@pgombar 

> Also, should we be testing if we've properly cleaned up invalid downloads/expansions?

test_it_should_ignore_existing_extension_package_when_it_is_invalid verifies that, if for any reason we don't cleanup an invalid zip, we re-download.

I'll add a couple of asserts to test_it_should_raise_an_exception_when_all_downloads_fail to check that the zip & extension dir are gone

azurelinuxagent/ga/exthandlers.py
narrieta: Thanks. The loop is short and very simple; I think in this case the longer variable name wouldn't add much.

tests/ga/test_exthandlers_download_extension.py
narrieta: It is an arbitrary number :) - Read it as "fail a few times, then succeed".

tests/ga/test_exthandlers_download_extension.py
narrieta: idem

tests/ga/test_exthandlers_download_extension.py
narrieta: idem

tests/ga/test_exthandlers_download_extension.py
narrieta: added comment to code

pgombar:
Thanks, Norberto. LGTM.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -743,54 +745,84 @@ def report_event(self, message="", is_success=True, duration=0, log_event=True):
...
+            if not self.protocol.download_ext_handler_pkg(source_uri, target_file):
+                raise Exception("Failed to download extension package - no error information is available")
+        except Exception as exception:
+            self.logger.info("Error downloading extension package: {0}", ustr(exception))
+            if os.path.exists(target_file):
+                os.remove(target_file)
+            return False
+        return True
+
+    def _unzip_extension_package(self, source_file, target_directory):
```
vrdmr: Just a clarification - Would the issue of corrupt zip arise for Agent as well? and we need similar logic in `azurelinuxagent.ga.update.GuestAgent#_unpack`?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -743,54 +745,84 @@ def report_event(self, message="", is_success=True, duration=0, log_event=True):
...
 
+    def _download_extension_package(self, source_uri, target_file):
+        self.logger.info("Downloading extension package: {0}", source_uri)
+        try:
+            if not self.protocol.download_ext_handler_pkg(source_uri, target_file):
+                raise Exception("Failed to download extension package - no error information is available")
+        except Exception as exception:
+            self.logger.info("Error downloading extension package: {0}", ustr(exception))
+            if os.path.exists(target_file):
+                os.remove(target_file)
```
vrdmr: The `fileutil.clean_ioerror` already does this partially. We should change (overload) this as util and use that in both the place (756, 766 and 768).

azurelinuxagent/ga/exthandlers.py
```diff
@@ -743,54 +745,84 @@ def report_event(self, message="", is_success=True, duration=0, log_event=True):
...
 
-            except Exception as e:
-                logger.warn("Error while downloading extension: {0}", ustr(e))
+                if downloaded:
+                    break
 
-        if not file_downloaded:
-            raise ExtensionDownloadError("Failed to download extension", code=1001)
+                self.logger.info("Failed to download the extension package from all uris, will retry after a minute")
+                time.sleep(60)
```
vrdmr: Good call! 👍

azurelinuxagent/ga/exthandlers.py
```diff
@@ -743,54 +745,84 @@ def report_event(self, message="", is_success=True, duration=0, log_event=True):
...
+                i += 1
 
-        self.logger.verbose("Unzip extension package")
-        try:
-            zipfile.ZipFile(self.pkg_file).extractall(self.get_base_dir())
-        except IOError as e:
-            fileutil.clean_ioerror(e, paths=[self.get_base_dir(), self.pkg_file])
-            raise ExtensionError(u"Failed to unzip extension package", e, code=1001)
+            if not downloaded:
+                raise ExtensionDownloadError("Failed to download extension", code=1001)
```
vrdmr: I'll open a task. We need to document these magical numbers. (such as 1001).

tests/ga/test_exthandlers_download_extension.py
```diff
@@ -0,0 +1,210 @@
...
+
+        ext_handler_properties = ExtHandlerProperties()
+        ext_handler_properties.version = "1.0.0"
+        ext_handler = ExtHandler(name='Microsoft.CPlat.Core.RunCommandLinux')
+        ext_handler.properties = ext_handler_properties
+
+        protocol = WireProtocol("http://Microsoft.CPlat.Core.RunCommandLinux/foo-bar")
+
+        self.pkg = ExtHandlerPackage()
+        self.pkg.uris = [ ExtHandlerVersionUri(), ExtHandlerVersionUri(), ExtHandlerVersionUri(), ExtHandlerVersionUri(), ExtHandlerVersionUri() ]
```
vrdmr: If you change `ExtHandlerVersionUri` to the following, It would make both the class as well as the test a little cleaner:
```
class ExtHandlerVersionUri(DataContract):
    def __init__(self, uri=None):
        self.uri = uri
```

```
self.pkg.uris = [ 
ExtHandlerVersionUri('https://zrdfepirv2cy4prdstr00a.blob.core.windows.net/f72653efd9e349ed9842c8b99e4c1712-foobar/Microsoft.CPlat.Core__RunCommandLinux__1.0.0'), 
ExtHandlerVersionUri('https://zrdfepirv2cy4prdstr01a.blob.core.windows.net/f72653efd9e349ed9842c8b99e4c1712-foobar/Microsoft.CPlat.Core__RunCommandLinux__1.0.0'), 
ExtHandlerVersionUri('https://zrdfepirv2cy4prdstr02a.blob.core.windows.net/f72653efd9e349ed9842c8b99e4c1712-foobar/Microsoft.CPlat.Core__RunCommandLinux__1.0.0'), 
ExtHandlerVersionUri('https://zrdfepirv2cy4prdstr02a.blob.core.windows.net/f72653efd9e349ed9842c8b99e4c1712-foobar/Microsoft.CPlat.Core__RunCommandLinux__1.0.0'), 
ExtHandlerVersionUri('https://zrdfepirv2cy4prdstr02a.blob.core.windows.net/f72653efd9e349ed9842c8b99e4c1712-foobar/Microsoft.CPlat.Core__RunCommandLinux__1.0.0')
]
```

azurelinuxagent/ga/exthandlers.py
narrieta: Yup, and I intend to fix that as well. However, this is a hotfix release, so I do not want to add any other changes to it. Thanks for pointing it out.

azurelinuxagent/ga/exthandlers.py
narrieta: That function seems to be meant to handle very specific errors and cannot handle this one in particular - it has a check for the exception type. Since that function is used in many places in the code, I decided to leave it alone for the moment.

azurelinuxagent/ga/exthandlers.py
narrieta: ok

tests/ga/test_exthandlers_download_extension.py
narrieta: ExtHandlerVersionUri is meant to be used to deserialize from xml, etc. The class has other members other than uri. I think it would not be worth it to add all of those parameters to __init__  just for the sake of this test.

Issue #1331, opened 2018-09-04
waagent.conf should allow the user to override CGroup resource limits for individual extensions


jasonzio:
#1327 applies a default set of resource limit to any extension. The agent config file should allow the user to override that default for any desired extension. (This is distinct from #1330 which simply says "exclude from limits at all".)

Issue #1329, opened 2018-09-04
CGroup limits should be specified on a per-extension basis


jasonzio:
The metadata which describes an extension to the Azure Compute RP and to the agent should include default resource limits set by the extension developer. If none are specified in the metadata, the agent should apply the "global" default. Extension metadata should also allow an extension developer to declare "there are no useful limits which can be applied to this extension", e.g. for CustomScript. This eliminates the need for the two special-cases wired into #1327.

PR #1327, opened 2018-09-04
Enforce cgroups limits

hglkrijger:
## Description

- closes #1316 
- add a global configuration flag for enforcing cgroups limits
- enforce limits by default on non-systemd-managed cgroups
- exclude CustomScript and RunCommand extensions
- add telemetry for enforced limits
- adds unit tests

---

### PR information
- [x] The title of the PR is clear and informative.
- [x] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).
- [x] Except for special cases involving multiple contributors, the PR is started from a fork of the main repository, not a branch.
- [x] If applicable, the PR references the bug/issue that it fixes in the description.
- [x] New Unit tests were added for the changes made and Travis.CI is passing.

### Quality of Code and Contribution Guidelines
- [x] I have read the [contribution guidelines](https://github.com/Azure/WALinuxAgent/blob/master/.github/CONTRIBUTING.md).

PR #1482, opened 2019-03-06
Added description for Extension Error Codes

vrdmr:
## Description
Second take for #1469.

---

### PR information
- [x] The title of the PR is clear and informative.
- [x] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).
- [x] Except for special cases involving multiple contributors, the PR is started from a fork of the main repository, not a branch.
- [x] If applicable, the PR references the bug/issue that it fixes in the description.
- [x] New Unit tests were added for the changes made and Travis.CI is passing.

### Quality of Code and Contribution Guidelines
- [x] I have read the [contribution guidelines](https://github.com/Azure/WALinuxAgent/blob/master/.github/CONTRIBUTING.md).

PR #1496, opened 2019-03-27
Reading extensions handler configuration file provided by publishers.

vrdmr:
## Description
This is the first of 2 (or possibly three PR). This one plumbs the HandlerConfiguration file to be read by the GA. This PR does not consume the HandlerConfiguration yet and will be followed up with another PR to read the limits, and the third one with the notifications.

Issue #1331, #1327, #1329 and #1363 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1496)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -1396,3 +1437,107 @@ def is_update_with_install(self):
...
+        else:
+            return None
+
+
+class CpuLimits(object):
+    def __init__(self, cpu_node):
+        self.cpu_limits = []
+        self.cores = []
+
+        for i in cpu_node:
```
pgombar: naming: something more descriptive than `i`, like `property` or `parameter`?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1396,3 +1437,107 @@ def is_update_with_install(self):
...
+            raise ExtensionError("Default CPU limit not set."
+                                 " Core configuration for {0} not present".format(DEFAULT_CORES_COUNT))
+
+        self.cpu_limits = sorted(self.cpu_limits)
+
+    def __str__(self):
+        return self.cpu_limits.__str__()
+
+
+class CpuLimit(object):
```
pgombar: naming; This will be difficult to differentiate from the other class, `CpuLimits`. maybe something like `CpuLimitInstance`?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1396,3 +1437,107 @@ def is_update_with_install(self):
...
+    def __str__(self):
+        return {"cores": self.cores, "limit_percentage": self.limit_percentage}.__str__()
+
+    def __repr__(self):
+        return {"cores": self.cores, "limit_percentage": self.limit_percentage}.__str__()
+
+
+class MemoryLimit(object):
+    def __init__(self, memory_node):
+        set_for_memory_oom_kill = ["enabled", "disabled"]
```
pgombar: naming; this set is not a set :) maybe `memory_oom_kill_options`?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1396,3 +1437,107 @@ def is_update_with_install(self):
...
+
+        self.max_limit_percentage = memory_node[
```
pgombar: In the next 4 lines, you can use: `get(key[, default])`. For example, `memory_node.get("max_limit_percentage", None)`

tests/data/ext/sample_ext_with_limits-1.4.0/HandlerConfiguration.json
```diff
@@ -0,0 +1,31 @@
+{
```
pgombar: I don't see the difference between this configuration and the malformed one?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -873,42 +875,75 @@ def initialize(self):
...
+        try:
+            self.set_operation(WALAEventOperation.Update)
+            man = self.load_manifest()
+            handler_configuration = self.load_handler_configuration()
+
+            update_cmd = man.get_update_command()
+            self.logger.info("Update extension [{0}]".format(update_cmd))
+            self.launch_command(update_cmd,
+                                timeout=900,
+                                extension_error_code=1008,
```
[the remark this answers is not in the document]
vrdmr: I don't know how this missed in #1482. Added that error code.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1396,3 +1437,107 @@ def is_update_with_install(self):
...
+            cpu_limits_object = CpuLimits(resource_config["cpu"])
+            return cpu_limits_object
+        else:
+            return None
+
+    def get_memory_limits_for_extension(self):
+        resource_config = self.get_resource_configurations()
+
+        if resource_config and "memory" in resource_config:
+            return MemoryLimit(resource_config["memory"])
```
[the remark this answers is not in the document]
vrdmr: Done. Added individual entries as "instance"

tests/data/ext/sample_ext_with_limits-1.4.0/HandlerConfiguration.json
vrdmr: Do you mean the one in the test `test_handler_configuration_incorrect_memory_configuration` - I added `"memory_oom_kill": "enabledd"` <- the op is incorrect.

And for the test is `test_handler_configuration_incorrect_configuration` - It is trying to replicate different types of specs for different cases - Empty resource, no configs, etc.

azurelinuxagent/ga/exthandlers.py
vrdmr: Missed in the next commit - Will resolve in the subsequent commits.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1191,6 +1210,24 @@ def read_output():
 
         return read_output()
 
+    def load_handler_configuration(self):
+        handler_config = None
+        data = None
+        config_file = self.get_handler_configuration_file()
+        try:
+            data = json.loads(fileutil.read_file(config_file))
+        except (IOError, OSError) as e:
```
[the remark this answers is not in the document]
vrdmr: The `test_load_handler_configuration` test checks for all the cases - 

```
def test_load_handler_configuration(self, *args):
    ...
    with patch(
            "azurelinuxagent.ga.exthandlers.ExtHandlerInstance.get_handler_configuration_file") as \
            patch_get_handler_configuration_file:
        patch_get_handler_configuration_file.return_value = load_data_path("ext/SampleHandlerConfiguration.json")
        handler_config = ext_handler_i.load_handler_configuration()
        self.assertNotEqual(handler_config, None)

    with patch(
            "azurelinuxagent.ga.exthandlers.ExtHandlerInstance.get_handler_configuration_file") as \
            patch_get_handler_configuration_file:
        patch_get_handler_configuration_file.return_value = load_data_path("ext/NotPresent.json")
        handler_config = ext_handler_i.load_handler_configuration()
        self.assertEquals(handler_config, None)

    with patch(
            "azurelinuxagent.ga.exthandlers.ExtHandlerInstance.get_handler_configuration_file") as \
            patch_get_handler_configuration_file:
        patch_get_handler_configuration_file.return_value = load_data_path("ext/SampleMalformedHandlerConfiguration.json")
        handler_config = ext_handler_i.load_handler_configuration()
        self.assertEquals(handler_config, None)
```

tests/data/ext/sample_ext_with_limits-1.4.0/HandlerConfiguration.json
pgombar: That's the difference, I didn't notice the typo, thanks.

tests/ga/test_extension.py
```diff
@@ -407,6 +406,24 @@ def test_ext_handler_no_public_settings(self, *args):
...
+    def test_ext_handler_with_resource_limits_passed_in_hanlder_configuration(self, *args):
+        test_data = WireProtocolData(DATA_FILE_CGROUP_LIMITS_SET)
+        exthandlers_handler, protocol = self._create_mock(test_data, *args)
+
+        # Test enable scenario.
+        exthandlers_handler.run()
+        self._assert_handler_status(protocol.report_vm_status, "Ready", 1, "1.0.0")
+        self._assert_ext_status(protocol.report_ext_status, "success", 0)
+
+    def test_ext_handler_with_invalid_resource_limits_passed_in_hanlder_configuration(self, *args):
```
pgombar: typo in test name: `hanlder` -> `handler`

tests/ga/test_extension.py
```diff
@@ -567,6 +584,318 @@ def test_ext_handler_reporting_status_file(self, *args):
...
+  }
+}
+'''
+        handlerConfig = HandlerConfiguration(json.loads(data))
+
+        self.assertEqual(handlerConfig.get_name(), "ExampleHandlerLinux")
+        self.assertEqual(handlerConfig.get_version(), 1.0)
+
+        cpu_limits = handlerConfig.get_cpu_limits_for_extension().cpu_limits
+        self.assertEquals(3, len(cpu_limits))  # 3 configurations sent
```
pgombar: Can you search and replace `assertEquals` with `assertEqual` everywhere? Primarily because of consistency and because `assertEquals` has been deprecated since Python 2.7.

tests/tools.py
```diff
@@ -374,6 +374,11 @@ def load_data(name):
         return data_file.read()
 
 
+def load_data_path(name):
```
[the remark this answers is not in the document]
vrdmr: The below method reads the contents of the files and returns contents. I wanted the action to read the file to be performed by the `load_handler_configuration` method, and thus only wanted file path.

```
patch_get_handler_configuration_file.return_value = load_data_path("ext/SampleHandlerConfiguration.json")
handler_config = ext_handler_i.load_handler_configuration()
```

The method `load_bin_data` is used when the tests want the contents - `self.ext = load_bin_data(data_files.get("test_ext"))`

tests/ga/test_extension.py
vrdmr: The `assertEquals` to `assertEqual` change is a big one (more than 100 matches). I think that could be a different PR as It'll cause a lot of changes.

azurelinuxagent/common/cgroups/cgroups.py
```diff
@@ -748,7 +617,7 @@ def set_cpu_limit(self, limit=None):
         if not CGroups.enabled():
             return
 
-        if limit is None:
+        if not limit:
```
narrieta: why the change to "not"? thanks
btw given this test we should probably remove the default value for limit (None)

tests/tools.py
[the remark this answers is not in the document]
pgombar: Oops, GitHub was only showing me the first two lines of the function below. It makes sense when I expand it :)

tests/ga/test_extension.py
pgombar: Sure, but please at least be consistent within your changes.

vrdmr: I've only made changes in one file - test_extHandlers for the timebeing. I'll slowly make the changes in other places as well.

azurelinuxagent/common/cgroups/cgroups.py
vrdmr: > why the change to "not"? thanks
Making it pythonic.

>btw given this test we should probably remove the default value for limit (None)
`None` could still be passed from the caller function. I'll check the code, and try to see if theres a test which covers it.

narrieta: not that it matters, but I think the best practice is the comparison against None (i.e. the way the code was written) -- no need to change, i was just curious.

for the default value -- the design is kinda weird.... set_cpu_limit with no arguments is a no-op... why allow that at all?

azurelinuxagent/common/cgroups/cgutils.py
```diff
@@ -0,0 +1,220 @@
...
+        Initialize data collection for the Cpu hierarchy. User must call update() before attempting to get
+        any useful metrics.
+
+        :param cgt: CGroupsTelemetry
+        :return:
+        """
+        self.cgt = cgt
+        self.osutil = get_osutil()
+        self.current_cpu_total = self.get_current_cpu_total()
+        self.previous_cpu_total = 0
```
narrieta: seems like previous is used only internally, let's mark it as private

azurelinuxagent/common/cgroups/cgutils.py
```diff
@@ -0,0 +1,220 @@
...
+        cpu_delta = self.current_cpu_total - self.previous_cpu_total
+        system_delta = max(1, self.current_system_cpu - self.previous_system_cpu)
+
+        return round(float(cpu_delta * self.cgt.cpu_count * 100) / float(system_delta), 3)
+
+    def collect(self):
+        """
+        Collect and return a list of all cpu metrics. If no metrics are collected, return an empty list.
+
+        :rtype: [(str, str, float)]
```
narrieta: can we document what these str, str, float are?

azurelinuxagent/common/cgroups/cgutils.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def get_cpu_limits(self, name, handler_configuration, compute_default):
+        limit_requested = None
+
+        cpu_limits_requested_by_extn = handler_configuration.get_cpu_limits_for_extension() if handler_configuration \
+            else None
+
+        if cpu_limits_requested_by_extn:
+            cores_count = self.osutil.get_processor_cores()
+            # sorted by cores. -1 is the default entry - and the first entry.
```
narrieta: who guarantees they are sorted? thanks

azurelinuxagent/common/exception.py
```diff
@@ -79,6 +79,15 @@ def __init__(self, msg=None, inner=None, code=-1):
         super(ExtensionOperationError, self).__init__(msg, inner, code)
 
 
+class ExtensionConfigurationError(ExtensionError):
+    """
+    When failed to configure an extension
+    """
```
narrieta: What do we mean by "failing to configure an extension"?
The current usage of the extension is for reporting bad configuration files... are you thinking of other uses for this exception?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1396,3 +1437,158 @@ def is_update_with_install(self):
...
+
+
+class CpuLimits(object):
+    def __init__(self, cpu_node):
+        self.cpu_limits = []
+        self.cores = []
+
+        for property in cpu_node:
+            # integers and > 0.
+            if "cores" not in property or "limit_percentage" not in property:
```
narrieta: we should do all the validation of the configuration file when we load it

azurelinuxagent/common/cgroups/cgutils.py
vrdmr: It is used in the `__str__` method, and cannot be made private to the `__init__` method.

azurelinuxagent/ga/exthandlers.py
vrdmr: I do 4-level validation at the time of loading up itself, but doing a very nested level validation would be very time-consuming.

Eg: In the following, I validate the following keys for absence and structure. 
level | key
1 - `name`, `version`, `handlerConfiguration`, 
2 - `handlerConfiguration.linux`, 
3 - `handlerConfiguration.linux.resources`, 
4 - `handlerConfiguration.linux.resources.{cpu|memory}`
```
{
      "name": "ExampleHandlerLinux",
      "version": 1.0,
      "handlerConfiguration": {
        "linux": {
          "resources": {
            "cpu": [
              {
                "cores": 2,
                "limit_percentage": 25
              },
              {
                "cores": 8,
                "limit_percentage": 20
              },
              {
                "cores": -1,
                "limit_percentage": 15
              }
            ],
            "memory": {
              "max_limit_percentage": 20,
              "max_limit_MBs": 1000,
              "memory_pressure_warning": "low",
              "memory_oom_kill": "enabled"
            }
          }
        },
        "windows": {}
      }
    }
```

We can discuss the viability of doing a rigid check of the whole file, but I am erroring on the side of publishers knowing the structure. Let me know your thoughts.

azurelinuxagent/common/cgroups/cgroups.py
vrdmr: >for the default value -- the design is kinda weird.... set_cpu_limit with no arguments is a no-op... why allow that at all?
makes sense. Let me take a look at this piece of code, and how it handles the set-limit in the new world order.

azurelinuxagent/common/cgroups/cgroups.py
```diff
@@ -221,7 +87,7 @@ def track_systemd_service(name):
             CGroupsTelemetry._tracked[service_name] = tracker
 
     @staticmethod
-    def track_extension(name, cgroup=None):
+    def track_extension(name, cgroup=None, handler_configuration=None):
```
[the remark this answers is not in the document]
narrieta: great, thank you!

azurelinuxagent/common/cgroups/cgutils.py
narrieta: I mean add "_" to the beginning of the name - naming convention for "private" members

azurelinuxagent/ga/exthandlers.py
narrieta: i would add more validation, including typos in the values (e.g. "cores" should be an integer)

the idea is that extension publishers should be able to find any errors/typos in their initial testing, and not after the extension has been deployed.

can you look if DataContract facilitates validation? (I have not read the code).

i would not expect validation to take more than tenths or seconds at worst, do you think validation time is a concern? thanks

azurelinuxagent/common/osutil/default.py
```diff
@@ -299,12 +300,17 @@ def is_current_instance_id(self, id_that):
...
     def is_cgroups_supported():
         """
-        Enabled by default; disabled in WSL/Travis
+        Enabled by default; disabled in WSL/Travis (Disabled on Trusty only)
         """
         is_wsl = '-Microsoft-' in platform.platform()
         is_travis = 'TRAVIS' in os.environ and os.environ['TRAVIS'] == 'true'
+
+        # Travis with python 2.6 fails (on Trusty based systems)
+        if is_travis and PY_VERSION_MAJOR == 2 and PY_VERSION_MINOR == 6:
```
narrieta: this code looks funny... "if it is travis and we are running 2.6 then it is not travis"?

vrdmr: yup. Fixed with better structure

```
supported_in_travis = True # By default supported in Travis
# Travis with python 2.6 fails (on Trusty based systems)
if is_travis and PY_VERSION_MAJOR == 2 and PY_VERSION_MINOR == 6:
supported_in_travis = False
```

azurelinuxagent/ga/exthandlers.py
vrdmr: > the idea is that extension publishers should be able to find any errors/typos in their initial testing, and not after the extension has been deployed.
Just for my clarification - the extension will still be deployed, as we don't want to deny users using the extension only because cpu/memory values were not set correctly? This might not be true in the case of capabilities, as we don't want to setup something with more privilege (which can then be actively done in `ExtensionCapabilitiesConfiguration`-like class).

and the extension, while being deployed is going to hit the validation points, where the telemetry would capture the validation issues, if any. This would be exercised by the `CGroupsTelemetry.track_extension(self.ext_handler.name, cg, handler_configuration.get_resource_configurations())` call after the launch of the extension is done.

> can you look if DataContract facilitates validation? (I have not read the code).
I don't think so. I only a simple fetch of data done by the `get_properties(obj)` method using the DataContract. Its a verly loosely specified.
```
def get_properties(obj):
    if isinstance(obj, DataContract):
        data = {}
        props = vars(obj)
        for prob_name, prob in list(props.items()):
            data[prob_name] = get_properties(prob)
        return data
    elif isinstance(obj, DataContractList):
        data = []
        for item in obj:
            item_data = get_properties(item)
            data.append(item_data)
        return data
    else:
        return obj

class DataContract(object):
    pass
```

> i would not expect validation to take more than tenths or seconds at worst, do you think validation time is a concern? thanks
It isn't not much, but I think `handlerConfig` not be something that all extensions would be signing up for, and would rely more on the default limits, and all the work we are doing on validation is mostly not needed.

I am okay to do the change - let me check what all things are required for the changes.

azurelinuxagent/common/cgroups/cgroups.py
vrdmr: @narrieta - This is now done. The HandlerConfig is validated at load time. After going through the usecases for tests, it made sense to load and report the issues immediately. Thanks.

azurelinuxagent/ga/exthandlers.py
vrdmr: @narrieta - This is now done. The HandlerConfig is validated at load time. After going through the usecases for tests, it made sense to load and report the issues immediately. Thanks.

azurelinuxagent/common/cgroups/cgroups.py
```diff
@@ -474,6 +340,30 @@ def _try_mkdir(path):
                 else:
                     raise
 
+    @staticmethod
+    def _try_rmdir(path):
+        """
+        Try to create a directory, recursively. If it already exists as such, do nothing. Raise the appropriate
```
narrieta: comment copy and pasted from somewhere else?

azurelinuxagent/common/cgroups/cgroups.py
```diff
@@ -474,6 +340,30 @@ def _try_mkdir(path):
                 else:
                     raise
 
+    @staticmethod
+    def _try_rmdir(path):
```
narrieta: this functions seems to be used only from remove_cgroup(), which does not seem to be used

azurelinuxagent/common/cgroups/cgroups.py
narrieta: any updates?

azurelinuxagent/common/exception.py
narrieta: I was trying to ask what the intent of this exception is... that'll help people consuming it. 

"When failed... configuring the extension requirements for capabilities or cgroups" ... you can omit capabilities for the moment. As for cgroups, I see many places we throw CGroupsException when setting up cgroups. When should ExtensionConfigurationError be used instead?

azurelinuxagent/common/cgroups/cgroups.py
vrdmr: I'll fix the comment.

tests/common/test_cgroups.py
```diff
@@ -290,44 +297,928 @@ def test_limits(self):
...
+            with patch("azurelinuxagent.common.osutil.default.DefaultOSUtil.get_total_mem") as patch_get_total_mem:
+                # Configuration
+                # RAM - 512 MB
+                # # of cores - 8
+                total_ram = 256
+                patch_get_processor_cores.return_value = 8
+                patch_get_total_mem.return_value = total_ram
+
+                expected_cpu_limit = CGroupsLimits.get_default_cpu_limits(cgroup_name)
+                expected_memory_limit = total_ram * 1  # min of %age of totalram vs max gives total ram edge here.
```
pgombar: Why do we allow setting the memory limit for extensions to something higher than the maximal allowed limit for extensions?

azurelinuxagent/common/cgroups/cgroups.py
vrdmr: It's used in the tests `TestExtensionWithCGroupsEnabled`. This is to cleanup the cgroup for tests.

```
class TestExtensionWithCGroupsEnabled(AgentTestCase):
    @staticmethod
    def _cleanup_cgroups(cgroup):
        try:
            cgroup.remove_cgroup()
        except CGroupsException:
            pass

        CGroupsTelemetry.stop_tracking(cgroup.name)
```

tests/ga/test_extension.py
```diff
@@ -1593,6 +1833,534 @@ def test_it_should_parse_boolean_values_encoded_as_strings(self):
...
+      }
+    }
+    '''
+        handler_config = HandlerConfiguration(json.loads(data))
+        resource_config = handler_config.get_resource_configurations()
+
+        self.assertEqual(resource_config.get_memory_limits_for_extension(), None)
+
+        cpu_limits = resource_config.get_cpu_limits_for_extension().cpu_limits
+        self.assertEqual(1, len(cpu_limits))  # 3 configurations sent
```
pgombar: Update comment

azurelinuxagent/common/cgroups/cgroups.py
narrieta: should be moved to the test then?

tests/ga/test_extension.py
```diff
@@ -1593,6 +1833,534 @@ def test_it_should_parse_boolean_values_encoded_as_strings(self):
...
+          }
+        }
+      }
+    }
+    '''
+        handler_config = HandlerConfiguration(json.loads(data))
+        resource_config = handler_config.get_resource_configurations()
+
+        self.assertEqual(resource_config.get_memory_limits_for_extension(), None)
+        self.assertEqual(resource_config.get_cpu_limits_for_extension(), None)
```
pgombar: I think I'm missing why the cpu limits wouldn't be set in this case, can you elaborate?

tests/ga/test_extension.py
```diff
@@ -1593,6 +1833,534 @@ def test_it_should_parse_boolean_values_encoded_as_strings(self):
...
+        ext_handler.properties.version = handler_version
+        ext_handler_i = ExtHandlerInstance(ext_handler, "dummy protocol")
+
+        with patch(
+                "azurelinuxagent.ga.exthandlers.ExtHandlerInstance.get_handler_configuration_file") as \
+                patch_get_handler_configuration_file:
+            patch_get_handler_configuration_file.return_value = load_data_path("ext/SampleHandlerConfiguration.json")
+            handler_config = ext_handler_i.load_handler_configuration()
+            self.assertNotEqual(handler_config, None)
+            self.assertEqual(patch_send_handler_configuration_event.call_count, 0)
```
pgombar: I thought `SampleHandlerConfiguration.json` was valid, why do we expect this isn't called?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1396,3 +1445,189 @@ def is_update_with_install(self):
         if update_mode is None:
             return True
         return update_mode.lower() == "updatewithinstall"
+
+
+class HandlerConfiguration(object):
```
narrieta: let's move this class to its own file

tests/ga/test_exthandlers.py
```diff
@@ -9,7 +9,7 @@
 from azurelinuxagent.common.exception import ProtocolError, ExtensionError, ExtensionErrorCodes
 from azurelinuxagent.common.event import WALAEventOperation
 from azurelinuxagent.common.utils.processutil import TELEMETRY_MESSAGE_MAX_LEN, format_stdout_stderr
-from azurelinuxagent.common.cgroups import CGroups
```
narrieta: there are many changes in exthandlers.py that don't seem to have a corresponding unit test here

tests/data/ext/SampleInvalidButCorrectJsonHandlerConfiguration.json
```diff
@@ -0,0 +1,31 @@
...
+            "cores": 8,
+            "limit_percentage": 20
+          },
+          {
+            "cores": 0,
+            "limit_percentage": 15
+          }
+        ],
+        "memory": {
+          "max_limit_percentage": 20,
```
anubhav929: Add a min threshold limit

azurelinuxagent/common/exception.py
vrdmr: Made this exception more specific now `ExtensionHandlerConfigurationError`.

azurelinuxagent/common/cgroups/cgroups.py
vrdmr: Its more useful for removing cgroups in general, and we should add that logic into CGroups. This is similar to the `stop_tracking` call done in `CGroupsTelemetry` class. It is specific to the cgroup.

azurelinuxagent/common/cgroups/cgroups.py
```diff
@@ -794,36 +687,30 @@ def set_memory_limit(self, limit=None, unit='megabytes'):
...
+        else:
+            raise CGroupsException("Memory hierarchy not available in this cgroup")
 
-        # agent values
-        if AGENT_CGROUP_NAME.lower() in cgroup_name.lower():
-            mem_limit = min(DEFAULT_MEM_LIMIT_MAX_MB, mem_limit)
-        return mem_limit
+    def remove_cgroup(self):
+        for cgroup_hierarchy in self.cgroups.items():
+            self._try_rmdir(cgroup_hierarchy[1])
```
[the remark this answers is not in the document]
vrdmr: I am using the `.items()` call, to directly get the path, and if in future this can support any other hierarchies we want to support.

azurelinuxagent/ga/exthandlers.py
vrdmr: Done. Added `exthandlers_util` file.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1142,12 +1161,16 @@ def pre_exec_function():
...
                 try:
-                    cg = CGroups.for_extension(self.ext_handler.name)
-                    CGroupsTelemetry.track_extension(self.ext_handler.name, cg)
+                    resource_config = None
+                    if handler_configuration:
+                        resource_config = handler_configuration.get_resource_configurations()
+                    cg = CGroups.for_extension(self.ext_handler.name, resource_config)
+                    CGroupsTelemetry.track_extension(self.ext_handler.name, cg, resource_config)
                 except Exception as e:
                     self.logger.warn("Unable to setup cgroup {0}: {1}".format(self.ext_handler.name, e))
```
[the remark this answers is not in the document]
vrdmr: it does not throw an exception now; it is either null or populated.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1396,3 +1445,189 @@ def is_update_with_install(self):
...
+    def __init__(self, cpu_node):
+        self.cpu_limits = []
+        self.cores = []
+
+        for property in cpu_node:
+            # integers and > 0.
+            if "cores" not in property or "limit_percentage" not in property:
+                raise ExtensionConfigurationError("Malformed CPU limit node in HandlerConfiguration")
+            self.cpu_limits.append(CpuLimitInstance(property["cores"], property["limit_percentage"]))
+            self.cores.append(property["cores"])
```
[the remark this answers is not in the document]
vrdmr: The casting is checked inside `CpuLimitInstance`, and raises an exception if it has any issues.

tests/common/test_cgroups.py
```diff
@@ -290,44 +297,928 @@ def test_limits(self):
...
-        self.assertEqual(limits.cpu_limit, CGroupsLimits.get_default_cpu_limits(cgroup_name ))
-        self.assertEqual(limits.memory_limit, CGroupsLimits.get_default_memory_limits(cgroup_name ))
+        self.assertEqual(limits.cpu_limit, CGroupsLimits.get_default_cpu_limits(cgroup_name))
+        self.assertEqual(limits.memory_limit, CGroupsLimits.get_default_memory_limits(cgroup_name))
 
         limits = CGroupsLimits(None)
         self.assertEqual(limits.cpu_limit, CGroupsLimits.get_default_cpu_limits(cgroup_name))
         self.assertEqual(limits.memory_limit, CGroupsLimits.get_default_memory_limits(cgroup_name))
 
     @patch("azurelinuxagent.common.osutil.default.DefaultOSUtil.get_total_mem", return_value=1024)
```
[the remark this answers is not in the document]
vrdmr: Fixed this.

tests/common/test_cgroups.py
```diff
@@ -290,44 +297,928 @@ def test_limits(self):
...
+                               resource_configuration=handler_configuration.get_resource_configurations())
+
+        with patch(
+                "azurelinuxagent.common.osutil.default.DefaultOSUtil.get_processor_cores") as patch_get_processor_cores:
+            patch_get_processor_cores.return_value = 10
+            self.assert_limits(ext_name="normal_extension", expected_cpu_limit=float(15), exception_raised=True,
+                               resource_configuration=handler_configuration.get_resource_configurations(),
+                               set_memory_oom_flag_called=False)
+
+    def test_limits_with_cpu_resource_configuration(self):
```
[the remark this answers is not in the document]
vrdmr: `test_limits_with_cpu_resource_configuration` is checking only if the CPU node is passed, and combinations within it.

tests/ga/test_extension.py
vrdmr: I am checking This is the telemetry event check. To make sure if telemetry is being sent for invalid cases.

tests/ga/test_extension.py
vrdmr: We can sync tomorrow to describe the case. tl;dr- If the json is incorrect or invalid, resource config won't be populated.

tests/common/test_cgroups.py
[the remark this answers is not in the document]
vrdmr: Yes. This checks the cpu-only case.

azurelinuxagent/common/cgroups/cgroups.py
vrdmr: This is fixed in the new code -where the defaults also come from top. limits should now always be populated.

tests/common/test_cgroups.py
vrdmr: Can you explain this comment?

azurelinuxagent/common/cgroups/cgroups.py
narrieta: I'd remove if from the agent if it is not currently needed, we can always add it when needed. 

If you want to add it, then it needs improvements and unit tests (improvements - for example: both the name and documentation imply the function can be used to remove a directory, but the implementation includes some sort of error handling that seems specific to cgroups.

tests/ga/test_extension.py
```diff
@@ -1419,18 +1419,35 @@ class TestExtensionWithCGroupsEnabled(AgentTestCase):
...
-            pass
+        except Exception as e:
+            logger.info(ustr(e))
 
         CGroupsTelemetry.stop_tracking(cgroup.name)
 
     @staticmethod
-    def _cgroup_translate_into_oom_flags(memory_flags_dict):
-        if memory_flags_dict["memory_oom_kill"] == "enabled":
+    def _cgroup_translate_into_oom_flags(memory_flags):
```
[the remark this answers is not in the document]
vrdmr: There can be multiple flags passed. Currently, 2 flags can be passed.

```
default_memory_flags = {"memory_pressure_warning": None, "memory_oom_kill": "disabled"}
```

azurelinuxagent/ga/utils/exthandler_utils.py
```diff
@@ -0,0 +1,343 @@
...
+                                                  'No data present')
+            if "handlerConfiguration" not in handler_config_json:
+                raise ExtensionHandlerConfigurationError('Malformed handler configuration file. '
+                                                  '"HandlerConfiguration" key not present')
+            if "linux" not in handler_config_json['handlerConfiguration']:
+                raise ExtensionHandlerConfigurationError('No linux configurations present in '
+                                                  'HandlerConfiguration')
+
+            try:
+                self.resource_config = ExtensionResourcesConfiguration(self.get_linux_configurations())
```
narrieta: at this point there has been no validation on _data yet, so get_linux_configurations can throw KeyError if handlerConfiguration is missing (or has a typo)

azurelinuxagent/ga/utils/exthandler_utils.py
```diff
@@ -0,0 +1,343 @@
...
+                                                                  operation=WALAEventOperation.HandlerConfiguration)
+
+    def get_name(self):
+        return self._data.get("name", None)
+
+    def get_version(self):
+        return self._data.get("version", None)
+
+    def get_linux_configurations(self):
+        return self._data['handlerConfiguration']['linux']
```
narrieta: will throw if 'linux' is missing or has typo

narrieta: similar comment in other places (e.g. ExtensionResourcesConfiguration); we use the [] operator without previous validation that the key exists

azurelinuxagent/ga/utils/exthandler_utils.py
```diff
@@ -0,0 +1,343 @@
...
+
+
+class ExtensionResourcesConfiguration(object):
+    def __init__(self, resource_configuration):
+        self._data = resource_configuration["resources"]
+        self.cpu_limits = None
+        self.memory_limits = None
+
+        try:
+            if "cpu" not in self._data and "memory" not in self._data:
```
narrieta: i think the validation needs to be stricter... this only checks for the "cpu" or "memory" substring to be anywhere in the json text

azurelinuxagent/ga/utils/exthandler_utils.py
```diff
@@ -0,0 +1,343 @@
...
+                raise ExtensionHandlerConfigurationError('No "cpu" node present in '
```
narrieta: check is for memory and cpu - exception mentions cpu only

azurelinuxagent/ga/utils/exthandler_utils.py
narrieta: never mind - it's a dict

azurelinuxagent/ga/utils/exthandler_utils.py
vrdmr: The validation above check for that - I can change the logic to see directly into `self._data`, but it does the same thing.

```
            if "linux" not in handler_config_json['handlerConfiguration']:
                raise ExtensionHandlerConfigurationError('No linux configurations present in '
                                                         'HandlerConfiguration')
```

azurelinuxagent/ga/utils/exthandler_utils.py
```diff
@@ -0,0 +1,343 @@
...
+            if "linux" not in handler_config_json['handlerConfiguration']:
+                raise ExtensionHandlerConfigurationError('No linux configurations present in '
+                                                  'HandlerConfiguration')
+
+            try:
+                self.resource_config = ExtensionResourcesConfiguration(self.get_linux_configurations())
+            except ExtensionHandlerConfigurationError as e:
+                raise e
+
+            if self.resource_config or CGROUPS_ENFORCE_DEFAULT_LIMITS:
```
narrieta: shouldn't this be 'and' (instead or 'or')

can we add a unit test for this condition?

azurelinuxagent/ga/utils/exthandler_utils.py
vrdmr: It was already been checked in the constructor. I'll change to look in `_data` rather than `handler_config_json`.

```
if "linux" not in self._data['handlerConfiguration']:
    raise ExtensionHandlerConfigurationError('No linux configurations present in '
                                             'HandlerConfiguration')
```

azurelinuxagent/ga/utils/exthandler_utils.py
```diff
@@ -0,0 +1,343 @@
...
+
+
+class CpuLimitInstance(object):
+    def __init__(self, cores, limit_percentage):
+        if type(cores) is not int:
+            raise ExtensionHandlerConfigurationError(
+                "Incorrect types for CPU values | field - {0}/ value - {1}/ type - {2}".format("cores", cores,
+                                                                                               type(cores)))
+
+        if type(limit_percentage) is not float and type(limit_percentage) is not int:
```
narrieta: do you want to check strictly against float? (e.g. force "10.0" instead of "10")

azurelinuxagent/ga/utils/exthandler_utils.py
```diff
@@ -0,0 +1,343 @@
...
+    def __str__(self):
+        return {"cores": self.cores, "limit_percentage": self.limit_percentage}.__str__()
+
+    def __repr__(self):
+        return {"cores": self.cores, "limit_percentage": self.limit_percentage}.__str__()
+
+
+class MemoryLimits(object):
+    def __init__(self, memory_node):
+        if "max_limit_MBs" in memory_node and "max_limit_percentage" in memory_node:
```
narrieta: we need to be very clear wrt capitalization in our documentation for publishers (or just make comparison case insensitive)

tests/ga/utils/test_exthandler_utils.py
```diff
@@ -0,0 +1,1252 @@
...
+        resource_config = handler_config.get_resource_configurations()
+
+        with patch(
+                "azurelinuxagent.common.osutil.default.DefaultOSUtil.get_processor_cores") as patch_get_processor_cores:
+            total_ram = 1024
+            patch_get_processor_cores.return_value = 8
+            patch_get_total_mem.return_value = total_ram
+
+            expected_cpu_limit = 20
+            expected_memory_limit = 256  # .2 of total ram is lower than default, reset to default.
```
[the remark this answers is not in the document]
vrdmr: Yes. This "DEFAULT" value is kept in there with the variety of VMs in mind. This would be unnecessary with the `min_limit_MBs` flag in the future, where the onus comes on the extensions to get the minimum value set and thus would make DEFAULTs unnecessary in those cases.

azurelinuxagent/ga/utils/exthandler_utils.py
```diff
@@ -0,0 +1,343 @@
...
+            raise ExtensionHandlerConfigurationError(
+                "Default max memory limit not set. max_limit_MBs: {0}, max_limit_percentage: {1}".format(
+                    memory_node.get("max_limit_MBs", None), memory_node.get("max_limit_percentage", None)))
+
+        self.memory_pressure_warning = memory_node.get("memory_pressure_warning", None)
+        self.memory_oom_kill = None  # default will be set in compute_default for memory flags.
+
+        if "memory_oom_kill" in memory_node:
+            if memory_node["memory_oom_kill"].lower() in MEMORY_OOM_KILL_OPTIONS:
+                self.memory_oom_kill = memory_node["memory_oom_kill"].lower()
```
narrieta: oh, so sometimes we are case-insensitive 

we should be consistent

vrdmr: We aren't being case-insensitive about the key, but the value, in this case. Keys are strict.

narrieta: why the difference?

azurelinuxagent/ga/utils/exthandler_utils.py
vrdmr: In this case, I felt it was okay to give `float` as well as `int` as a percentage value. I am just making sure it is a number that they pass. I don't want to be too strict with floats. People will forget the `.0` suffix.

narrieta: yup; agree - you should then try to parse the value (or something like that) instead of strict type check

>>> type(1)
<class 'int'>
>>> type(1.0)
<class 'float'>
>>>

narrieta: nevermind, i see your point now

azurelinuxagent/ga/utils/exthandler_utils.py
vrdmr: Added. `test_ext_handler_with_cgroup_invalid_limits_passed_defaults_set` and `test_ext_handler_with_no_limits_passed_defaults_set`.

azurelinuxagent/common/cgroups/cgroups.py
vrdmr: Got it. Yes, refactoring/cleaning up the whole Cgroups is another big task. I'll move this method.

azurelinuxagent/ga/utils/exthandler_utils.py
vrdmr: the keys need to be strict - as we would be documenting them and don't want to add ambiguity in the key-names. The values can be made strict (enumify) and well documented. I can add that part of validation later.

vrdmr:
Closing this PR in favor of the new implementation of the HandlerConfiguration and adding limits.

Issue #1512, opened 2019-04-24
Add logging when sending ready signal to WireServer


rifrankl:
After the call to ProvisionHandler.write_provisioned(), there is no more logging until the "Provision complete" message. It would be nice if there was some additional logging within the ProvisionHandler.report_ready() call.

ZhidongPeng:
Provisioning part.

Issue #1511, opened 2019-04-24
[BUG] Failure to send provision complete signal in case of restart after provisioning completes


rifrankl:
Currently, if WAAgent is unable to report to WireServer that provisioning completed and the VM restarts, WAAgent will never again send the provisioning complete signal. The flow is something like:
1. Do provisioning work (ProvisionHandler.run())
2. Write provisioned file to filesystem (ProvisionHandler.write_provisioned())
3. Notify WireServer (ProvisionHandler.report_ready())
4. Provisioning completed.

However, in step 3, the issue arises if WireServer is unavailable (ie, WireServer is down, or networking is faulty, or whatever) and something causes WAAgent to restart. In this case, WAAgent will come up and see the provisioned file already exists, but will not report to WireServer. (ProvisionHandler.run() just returns without calling report_ready()).

pgombar:
@AOhassan is this still active? If not, we will close it.

Issue #1456, opened 2019-02-04
[BUG] "WARNING mounting DVD failed" when provisioning agent crashes and is restarted


jasonzio:
**Describe the bug**
When the provisioning agent starts, its first action is to capture the ovf.xml file. The specific sequence of operations:

- mount the dvd
- read the ovf.xml file from the dvd
- write the (redacted) ovf.xml file to disk
- umount the dvd
- eject the dvd

This is a non-idempotent sequence of operations. If the provisioning agent crashes before completing provisioning, it gets restarted, but the attempt to mount the dvd fails because it has been ejected.

The "eject the dvd" operation should be deferred until after provisioning is complete.

**Distro and WALinuxAgent details (please complete the following information):**
 - CemtOS 7.5 and earlier
 - WALinuxAgent version is unknown but older than 2.2.34

**Additional context**
Unfortunately, the "cannot mount dvd" handler cannot just use the copy of ovf.xml written to disk by the first (failed) instance of the provisioning agent because the agent password has been redacted from the copy written to disk.

PR #1528, opened 2019-05-16
Signalbug and Ejectbug

AOhassan:
## Description

Issue # #1511,#1512 and #1456

Added signaling to is_provisioned and run methods.
created signaling file and functions to read/write signaling file.
implemented more logging for is_provisioned and run methods.
fixed regex syntax that caused telemetry tests to fail.
changed report_ready to return boolean.
deferred eject_dvd to be called after provisioning is complete.
implemented tests for eject_dvd.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1528)


azurelinuxagent/pa/provision/default.py
```diff
@@ -57,45 +58,71 @@ def run(self):
...
 
-            self.handle_provision_guest_agent(ovf_env.provision_guest_agent)
 
-            self.report_ready(thumbprint)
-            logger.info("Provisioning complete")
+                self.report_event("Provisioning succeeded ({0}s)".format(self._get_uptime_seconds()),
+                    is_success=True,
+                    duration=elapsed_milliseconds(utc_start))
+
+                self.handle_provision_guest_agent(ovf_env.provision_guest_agent)
```
jasonzio: The very last thing that happens during provisioning should be the call to write_provisioned(); a crash immediately after that call would mean that, upon restart, provision_state == "provisioned_not_signaled" and the VM would immediately send the "done provisioning" signal to the control plane. However, as this is written, lines 110-117 are executed after write_provisioned(). Unless handle_provision_guest_agent() doesn't do any provisioning work (which I'm pretty sure is not a true statement), this code violates the expectation that provisioning is DONE when write_provisioned() is called.

azurelinuxagent/pa/provision/default.py
```diff
@@ -166,11 +193,20 @@ def get_ssh_host_key_thumbprint(self, chk_err=True):
...
+        return os.path.isfile(self.provisioned_file_path())
+
+    def isfile_signaled(self):
+        return os.path.isfile(self.signaled_file_path())
+
     def is_provisioned(self):
         '''
-        A VM is considered provisionend *anytime* the provisioning
+        A VM is considered provisioned *anytime* the provisioning
         sentinel file exists and not provisioned *anytime* the file
```
jasonzio: See my previous comment about handle_provision_guest_agent(). I don't think the code, as changed, lives up to this correct comment.

azurelinuxagent/pa/provision/default.py
```diff
@@ -57,45 +58,68 @@ def run(self):
         if not conf.get_provision_enabled():
             logger.info("Provisioning is disabled, skipping.")
             self.write_provisioned()
-            self.report_ready()
+            logger.info("signaling...")
+            if self.report_ready():
+                self.write_signaled()
+                logger.info("signaling complete")
+            else:
+                logger.info("signaling incomplete")
```
Jdzions: This block of code appears three times, and in all three cases subsequent code does not care which branch of the `if` was taken. You can hoist this up into a function local to the run() method.

azurelinuxagent/pa/provision/default.py
Jdzions: Turns out that handle_provision_guest_agent() does NOT do any provisioning work, just writes some log stuff. Weird name for a method that just logs.

I withdraw this objection.

---
In reply to: [284935536](https://github.com/Azure/WALinuxAgent/pull/1528#discussion_r284935536) [](ancestors = 284935536)

narrieta:
@AOhassan - Hi, just checking if this PR is still active. Thanks!

narrieta:
Please re-open if still active. Thanks!

PR #1538, opened 2019-05-28
Making cgroup telemetry work end to end, with new spec for metrics

vrdmr:
## Description
Enabling Telemetry E2E for Cgroup Perf counters for extensions.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1538)

### DIFF
--- a/azurelinuxagent/common/cgroup.py
+++ b/azurelinuxagent/common/cgroup.py
@@ -13,13 +13,16 @@
 # limitations under the License.
 #
 # Requires Python 2.6+ and Openssl 1.0+
-
+import os
 import re
 
+from azurelinuxagent.common import logger
 from azurelinuxagent.common.exception import CGroupsException
+from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.osutil import get_osutil
+from azurelinuxagent.common.utils import fileutil
 
-re_user_system_times = re.compile('user (\d+)\nsystem (\d+)\n')
+re_user_system_times = re.compile(r'user (\d+)\nsystem (\d+)\n')
 
 # The metric classes (Cpu, Memory, etc) can all assume that CGroups is enabled, as the CGroupTelemetry
 # class is very careful not to call them if CGroups isn't enabled. Any tests should be disabled if the osutil
@@ -27,81 +30,114 @@ re_user_system_times = re.compile('user (\d+)\nsystem (\d+)\n')
 
 
 class CGroup(object):
-    def __init__(self):
-        pass
+    def __init__(self, extension_name, cgroup_path, controller_type):
+        self.name = extension_name
+        self.path = cgroup_path
+        self.controller = controller_type
 
-    def _get_cgroup_file(self, hierarchy, file_name):
-        return os.path.join(self.cgroups[hierarchy], file_name)
+    def _get_cgroup_file(self, file_name):
+        return os.path.join(self.path, file_name)
 
-
-    def get_file_contents(self, hierarchy, file_name):
+    def get_file_contents(self, controller, file_name):
         """
-        Retrieve the value of a parameter from a hierarchy.
+        Retrieve the value of a parameter from a controller.
 
-        :param str hierarchy: Name of cgroup metric hierarchy
-        :param str file_name: Name of file within that metric hierarchy
+        :param str controller: Name of cgroup metric controller
+        :param str file_name: Name of file within that metric controller
         :return: Entire contents of the file
         :rtype: str
         """
-        if hierarchy in self.cgroups:
-            parameter_file = self._get_cgroup_file(hierarchy, file_name)
+        if controller == self.controller:
+            parameter_file = self._get_cgroup_file(file_name)
 
             try:
                 return fileutil.read_file(parameter_file)
             except Exception:
-                raise CGroupsException("Could not retrieve cgroup file {0}/{1}".format(hierarchy, file_name))
+                raise CGroupsException("Could not retrieve cgroup file {0}/{1}".format(self.path, file_name))
         else:
-            raise CGroupsException("{0} subsystem not available in cgroup {1}. cgroup paths: {2}".format(
-                hierarchy, self.name, self.cgroups))
+            raise CGroupsException("Incorrect controller: {0} passed for cgroup: {1}. Cgroup path is: {2}".format(
+                controller, self.name, self.path))
 
-    def get_parameter(self, hierarchy, parameter_name):
+    def get_parameter(self, controller, parameter_name):
         """
         Retrieve the value of a parameter from a hierarchy.
         Assumes the parameter is the sole line of the file.
 
-        :param str hierarchy: Name of cgroup metric hierarchy
+        :param str controller: Name of cgroup metric hierarchy
         :param str parameter_name: Name of file within that metric hierarchy
         :return: The first line of the file, without line terminator
         :rtype: str
         """
         result = ""
         try:
-            values = self.get_file_contents(hierarchy, parameter_name).splitlines()
+            values = self.get_file_contents(controller, parameter_name).splitlines()
             result = values[0]
         except IndexError:
-            parameter_filename = self._get_cgroup_file(hierarchy, parameter_name)
+            parameter_filename = self._get_cgroup_file(parameter_name)
             logger.error("File {0} is empty but should not be".format(parameter_filename))
-        except CGroupsException as e:
+        except CGroupsException:
             # ignore if the file does not exist yet
             pass
         except Exception as e:
-            parameter_filename = self._get_cgroup_file(hierarchy, parameter_name)
+            parameter_filename = self._get_cgroup_file(parameter_name)
             logger.error("Exception while attempting to read {0}: {1}".format(parameter_filename, ustr(e)))
         return result
 
+    def get_parameters(self, controller, parameter_name):
+        """
+        Retrieve the value of a parameter from a hierarchy.
+        Returns a list of values in the file.
+
+        :param str controller: Name of cgroup metric hierarchy
+        :param str parameter_name: Name of file within that metric hierarchy
+        :return: The first line of the file, without line terminator
+        :rtype: [str]
+        """
+        result = []
+        try:
+            result = self.get_file_contents(controller, parameter_name).splitlines()
+        except IndexError:
+            parameter_filename = self._get_cgroup_file(parameter_name)
+            logger.error("File {0} is empty but should not be".format(parameter_filename))
+        except CGroupsException:
+            # ignore if the file does not exist yet
+            pass
+        except Exception as e:
+            parameter_filename = self._get_cgroup_file(parameter_name)
+            logger.error("Exception while attempting to read {0}: {1}".format(parameter_filename, ustr(e)))
+        return result
+
+    def collect(self):
+        raise NotImplementedError()
+
+    def is_active(self):
+        tasks = self.get_parameters(self.controller, "tasks")
+        return len(tasks) != 0
+
 
 class CpuCgroup(CGroup):
-    def __init__(self, cgt):
+    def __init__(self, extension_name, cgroup_path, controller="cpu"):
         """
         Initialize data collection for the Cpu hierarchy. User must call update() before attempting to get
         any useful metrics.
 
-        :param cgt: CGroupsTelemetry
+        :param path: Filepath
         :return:
         """
-        self.cgt = cgt
-        self.osutil = get_osutil()
-        self.current_cpu_total = self.get_current_cpu_total()
-        self.previous_cpu_total = 0
-        self.current_system_cpu = self.osutil.get_total_cpu_ticks_since_boot()
-        self.previous_system_cpu = 0
+        super(CpuCgroup, self).__init__(extension_name, cgroup_path, controller)
+
+        self._osutil = get_osutil()
+        self._current_cpu_total = self._get_current_cpu_total()
+        self._previous_cpu_total = 0
+        self._current_system_cpu = self._osutil.get_total_cpu_ticks_since_boot()
+        self._previous_system_cpu = 0
 
     def __str__(self):
         return "Cgroup: Current {0}, previous {1}; System: Current {2}, previous {3}".format(
-            self.current_cpu_total, self.previous_cpu_total, self.current_system_cpu, self.previous_system_cpu
+            self._current_cpu_total, self._previous_cpu_total, self._current_system_cpu, self._previous_system_cpu
         )
 
-    def get_current_cpu_total(self):
+    def _get_current_cpu_total(self):
         """
         Compute the number of USER_HZ of CPU time (user and system) consumed by this cgroup since boot.
 
@@ -109,8 +145,7 @@ class CpuCgroup(CGroup):
         """
         cpu_total = 0
         try:
-            cpu_stat = self.cgt.cgroup.\
-                get_file_contents('cpu', 'cpuacct.stat')
+            cpu_stat = self.get_file_contents('cpu', 'cpuacct.stat')
             if cpu_stat is not None:
                 m = re_user_system_times.match(cpu_stat)
                 if m:
@@ -123,17 +158,17 @@ class CpuCgroup(CGroup):
             pass
         return cpu_total
 
-    def update(self):
+    def _update_cpu_data(self):
         """
         Update all raw data required to compute metrics of interest. The intent is to call update() once, then
         call the various get_*() methods which use this data, which we've collected exactly once.
         """
-        self.previous_cpu_total = self.current_cpu_total
-        self.previous_system_cpu = self.current_system_cpu
-        self.current_cpu_total = self.get_current_cpu_total()
-        self.current_system_cpu = self.osutil.get_total_cpu_ticks_since_boot()
+        self._previous_cpu_total = self._current_cpu_total
+        self._previous_system_cpu = self._current_system_cpu
+        self._current_cpu_total = self._get_current_cpu_total()
+        self._current_system_cpu = self._osutil.get_total_cpu_ticks_since_boot()
 
-    def get_cpu_percent(self):
+    def _get_cpu_percent(self):
         """
         Compute the percent CPU time used by this cgroup over the elapsed time since the last time this instance was
         update()ed.  If the cgroup fully consumed 2 cores on a 4 core system, return 200.
@@ -141,8 +176,8 @@ class CpuCgroup(CGroup):
         :return: CPU usage in percent of a single core
         :rtype: float
         """
-        cpu_delta = self.current_cpu_total - self.previous_cpu_total
-        system_delta = max(1, self.current_system_cpu - self.previous_system_cpu)
+        cpu_delta = self._current_cpu_total - self._previous_cpu_total
+        system_delta = max(1, self._current_system_cpu - self._previous_system_cpu)
 
         return round(float(cpu_delta * self.cgt.cpu_count * 100) / float(system_delta), 3)
 
@@ -152,29 +187,41 @@ class CpuCgroup(CGroup):
 
         :rtype: [(str, str, float)]
         """
-        self.update()
-        usage = self.get_cpu_percent()
-        return [("Process", "% Processor Time", usage)]
+        self._update_cpu_data()
+        usage = self._get_cpu_percent()
+        return [CollectedMetrics("cpu", "% Processor Time", usage)]
 
 
-class MemoryCGroup(CGroup):
-    def __init__(self, cgt):
+class MemoryCgroup(CGroup):
+    def __init__(self, extension_name, cgroup_path, controller="memory"):
         """
         Initialize data collection for the Memory hierarchy
 
         :param CGroupsTelemetry cgt: The telemetry object for which memory metrics should be collected
         :return:
         """
-        self.cgt = cgt
+        super(MemoryCgroup, self).__init__(extension_name, cgroup_path, controller)
 
-    def get_memory_usage(self):
+    def _get_memory_usage(self):
         """
         Collect memory.usage_in_bytes from the cgroup.
 
         :return: Memory usage in bytes
         :rtype: int
         """
-        usage = self.cgt.cgroup.get_parameter('memory', 'memory.usage_in_bytes')
+        usage = self.get_parameter('memory', 'memory.usage_in_bytes')
+        if not usage:
+            usage = "0"
+        return int(usage)
+
+    def _get_memory_max_usage(self):
+        """
+        Collect memory.usage_in_bytes from the cgroup.
+
+        :return: Memory usage in bytes
+        :rtype: int
+        """
+        usage = self.get_parameter(self.controller, 'memory.max_usage_in_bytes')
         if not usage:
             usage = "0"
         return int(usage)
@@ -185,8 +232,16 @@ class MemoryCGroup(CGroup):
 
         :rtype: [(str, str, float)]
         """
-        usage = self.get_memory_usage()
-        return [("Memory", "Total Memory Usage", usage)]
+        usage = self._get_memory_usage()
+        max_usage = self._get_memory_max_usage()
+        return [CollectedMetrics("memory", "Total Memory Usage", usage), CollectedMetrics("memory", "Max Memory Usage", max_usage)]
+
+
+class CollectedMetrics(object):
+    def __init__(self, controller, metric_name, value):
+        self.controller = controller
+        self.metric_name = metric_name
+        self.value = value
 
 #
 # TODO: Do we need this code?
--- a/azurelinuxagent/common/cgroup.py
+++ b/azurelinuxagent/common/cgroup.py
@@ -24,78 +24,47 @@ from azurelinuxagent.common.utils import fileutil
 
 re_user_system_times = re.compile(r'user (\d+)\nsystem (\d+)\n')
 
-# The metric classes (Cpu, Memory, etc) can all assume that CGroups is enabled, as the CGroupTelemetry
-# class is very careful not to call them if CGroups isn't enabled. Any tests should be disabled if the osutil
-# is_cgroups_support() method returns false.
-
 
 class CGroup(object):
-    def __init__(self, extension_name, cgroup_path, controller_type):
-        self.name = extension_name
+    def __init__(self, name, cgroup_path, controller_type):
+        self.name = name
         self.path = cgroup_path
         self.controller = controller_type
 
     def _get_cgroup_file(self, file_name):
         return os.path.join(self.path, file_name)
 
-    def get_file_contents(self, controller, file_name):
+    def get_file_contents(self, file_name):
         """
         Retrieve the value of a parameter from a controller.
 
-        :param str controller: Name of cgroup metric controller
         :param str file_name: Name of file within that metric controller
         :return: Entire contents of the file
         :rtype: str
         """
-        if controller == self.controller:
-            parameter_file = self._get_cgroup_file(file_name)
 
-            try:
-                return fileutil.read_file(parameter_file)
-            except Exception:
-                raise CGroupsException("Could not retrieve cgroup file {0}/{1}".format(self.path, file_name))
-        else:
-            raise CGroupsException("Incorrect controller: {0} passed for cgroup: {1}. Cgroup path is: {2}".format(
-                controller, self.name, self.path))
+        parameter_file = self._get_cgroup_file(file_name)
 
-    def get_parameter(self, controller, parameter_name):
-        """
-        Retrieve the value of a parameter from a hierarchy.
-        Assumes the parameter is the sole line of the file.
-
-        :param str controller: Name of cgroup metric hierarchy
-        :param str parameter_name: Name of file within that metric hierarchy
-        :return: The first line of the file, without line terminator
-        :rtype: str
-        """
-        result = ""
         try:
-            values = self.get_file_contents(controller, parameter_name).splitlines()
-            result = values[0]
-        except IndexError:
-            parameter_filename = self._get_cgroup_file(parameter_name)
-            logger.error("File {0} is empty but should not be".format(parameter_filename))
-        except CGroupsException:
-            # ignore if the file does not exist yet
-            pass
-        except Exception as e:
-            parameter_filename = self._get_cgroup_file(parameter_name)
-            logger.error("Exception while attempting to read {0}: {1}".format(parameter_filename, ustr(e)))
-        return result
+            return fileutil.read_file(parameter_file)
+        except Exception:
+            raise CGroupsException("Could not retrieve cgroup file {0}/{1}".
+                                   format(self.path, file_name))
 
-    def get_parameters(self, controller, parameter_name):
+    def get_parameters(self, parameter_name, first_line_only=False):
         """
         Retrieve the value of a parameter from a hierarchy.
         Returns a list of values in the file.
 
-        :param str controller: Name of cgroup metric hierarchy
+        :param first_line_only: return only the first line.
         :param str parameter_name: Name of file within that metric hierarchy
         :return: The first line of the file, without line terminator
         :rtype: [str]
         """
         result = []
         try:
-            result = self.get_file_contents(controller, parameter_name).splitlines()
+            values = self.get_file_contents(parameter_name).splitlines()
+            result = values[0] if first_line_only else values
         except IndexError:
             parameter_filename = self._get_cgroup_file(parameter_name)
             logger.error("File {0} is empty but should not be".format(parameter_filename))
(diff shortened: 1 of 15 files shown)


narrieta:
azurelinuxagent/common/cgroups/__init__.py is empty (not needed)

azurelinuxagent/common/cgroup.py
```diff
@@ -13,104 +13,139 @@
...
-        :param str hierarchy: Name of cgroup metric hierarchy
-        :param str file_name: Name of file within that metric hierarchy
+        :param str controller: Name of cgroup metric controller
+        :param str file_name: Name of file within that metric controller
         :return: Entire contents of the file
         :rtype: str
         """
-        if hierarchy in self.cgroups:
-            parameter_file = self._get_cgroup_file(hierarchy, file_name)
+        if controller == self.controller:
```
pgombar: Why do we need this check and why do we need to pass the `controller` variable instead of just using `self.controller`? I might be missing something here.

azurelinuxagent/common/cgroup.py
```diff
@@ -152,29 +187,41 @@ def collect(self):
...
+        super(MemoryCgroup, self).__init__(extension_name, cgroup_path, controller)
+
+    def _get_memory_usage(self):
+        """
+        Collect memory.usage_in_bytes from the cgroup.
+
+        :return: Memory usage in bytes
+        :rtype: int
+        """
+        usage = self.get_parameter('memory', 'memory.usage_in_bytes')
```
pgombar: Is there a reason you are not using `self.controller` instead of `'memory'` like in L224?

azurelinuxagent/common/cgroupapi.py
```diff
@@ -167,6 +168,15 @@ def _get_extension_cgroup_path(self, controller, extension_name):
 
         return os.path.join(extensions_root, cgroup_name)
 
+    def _get_extension_cgroup(self, controller, extension_name):
```
pgombar: Maybe `create` instead of `get` is a more descriptive function name.

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -13,47 +13,184 @@
...
+        return [metric.average(), metric.min(), metric.max(), metric.median(), metric.count(), metric.first_polltime(),
+                metric.last_polltime()]
+
+    @staticmethod
+    def _process_cgroup_metric(cgroup_metrics):
+        current_memory_usage, max_memory_levels, current_cpu_usage = cgroup_metrics.get_metrics()
+        processed_extension = {
+            "memory": {
+                "cur_mem": CGroupsTelemetry._get_metrics_list(current_memory_usage),
+                "max_mem": CGroupsTelemetry._get_metrics_list(max_memory_levels)
```
pgombar: What's the need for average, min, max, median, count... for max memory?

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -13,47 +13,184 @@
...
+        CGroupsTelemetry._rlock.acquire()
+        for cgroup in CGroupsTelemetry._tracked:
+            if not cgroup.is_active():
+                CGroupsTelemetry._tracked.remove(cgroup)
+
+        CGroupsTelemetry._rlock.release()
+
+    @staticmethod
+    def cleanup():
+        CGroupsTelemetry._tracked *= 0  # emptying the list
```
pgombar: This is a pretty neat way of doing this. I hope it's compatible with all our Python versions!

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -13,47 +13,184 @@
...
+        if controller is "cpu":
+            self._add_cpu_usage(metric)
+        elif controller is "memory":
+            self._add_memory_usage(metric)
+
+    def get_metrics(self):
+        return self.current_memory_usage, self.max_memory_levels, self.current_cpu_usage
+
+    def clear(self):
+        self.current_memory_usage.clear()
```
pgombar: `clear()` isn't supported for lists in Python <3.3

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -13,47 +13,184 @@
...
+
+    def append(self, metric):
+        self.data.append(metric)
+        self.last_poll_time = dt.utcnow()
+
+    def clear(self):
+        self.data *= 0
+
+    def average(self):
+        return sum(self.data) / len(self.data)
```
pgombar: We need an explicit cast to float here if we don't want to end up with an int value and use integer division.

tests/common/test_cgroups.py
```diff
@@ -304,6 +321,129 @@ def test_limits(self):
...
+            name = "dummy_extension_{0}".format(i)
+
+            self.assertIn(name, collected_metrics)
+            self.assertIn("memory", collected_metrics[name])
+            self.assertIn("cur_mem", collected_metrics[name]["memory"])
+            self.assertIn("max_mem", collected_metrics[name]["memory"])
+            self.assertEqual(num_summarization_values, len(collected_metrics[name]["memory"]["cur_mem"]))
+            self.assertEqual(num_summarization_values, len(collected_metrics[name]["memory"]["max_mem"]))
+
+            self.assertListEqual(generate_metric_list(memory_usage_values),
```
pgombar: Can you add another test where you actually assert concrete values that got computed? This way we can check the validity of the computations as well. I'm mostly concerned about division.

azurelinuxagent/common/cgroupstelemetry.py
vrdmr: This clear calls the clear within Metric() (which internally does *= 0)

azurelinuxagent/common/cgroupstelemetry.py
vrdmr: Verified between Py2 and 3. :)

azurelinuxagent/common/cgroup.py
```diff
@@ -13,104 +13,139 @@
...
+from azurelinuxagent.common import logger
 from azurelinuxagent.common.exception import CGroupsException
+from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.osutil import get_osutil
+from azurelinuxagent.common.utils import fileutil
 
-re_user_system_times = re.compile('user (\d+)\nsystem (\d+)\n')
+re_user_system_times = re.compile(r'user (\d+)\nsystem (\d+)\n')
 
 # The metric classes (Cpu, Memory, etc) can all assume that CGroups is enabled, as the CGroupTelemetry
```
narrieta: let's remove this comment, it is no longer applicable

azurelinuxagent/common/cgroup.py
```diff
@@ -13,104 +13,139 @@
...
 
 
 class CGroup(object):
-    def __init__(self):
-        pass
-
-    def _get_cgroup_file(self, hierarchy, file_name):
-        return os.path.join(self.cgroups[hierarchy], file_name)
+    def __init__(self, extension_name, cgroup_path, controller_type):
+        self.name = extension_name
```
narrieta: what is the name used for? i think i saw it only on an error message that also includes the path... the path should be enough to identify the cgroup.

If we do need the name it should not be "extension_name", since we will also have cgroups for the agent, services, etc

azurelinuxagent/common/cgroup.py
narrieta: +1 --- Also, in the next round of refactoring let's eliminate the controller_type, as we discussed.

azurelinuxagent/common/cgroup.py
```diff
@@ -13,104 +13,139 @@
...
         try:
-            values = self.get_file_contents(hierarchy, parameter_name).splitlines()
+            values = self.get_file_contents(controller, parameter_name).splitlines()
             result = values[0]
         except IndexError:
-            parameter_filename = self._get_cgroup_file(hierarchy, parameter_name)
+            parameter_filename = self._get_cgroup_file(parameter_name)
+            logger.error("File {0} is empty but should not be".format(parameter_filename))
+        except CGroupsException:
+            # ignore if the file does not exist yet
```
narrieta: this comment is misleading, or we are catching the wrong exception... seems to me we can get CGroupsException under conditions other that the file not existing

azurelinuxagent/common/cgroup.py
```diff
@@ -13,104 +13,139 @@
...
         Retrieve the value of a parameter from a hierarchy.
         Assumes the parameter is the sole line of the file.
 
-        :param str hierarchy: Name of cgroup metric hierarchy
+        :param str controller: Name of cgroup metric hierarchy
         :param str parameter_name: Name of file within that metric hierarchy
         :return: The first line of the file, without line terminator
         :rtype: str
         """
         result = ""
```
narrieta: None may be a better choice that "" for the return value

azurelinuxagent/common/cgroup.py
```diff
@@ -13,104 +13,139 @@
...
         else:
-            raise CGroupsException("{0} subsystem not available in cgroup {1}. cgroup paths: {2}".format(
-                hierarchy, self.name, self.cgroups))
+            raise CGroupsException("Incorrect controller: {0} passed for cgroup: {1}. Cgroup path is: {2}".format(
+                controller, self.name, self.path))
 
-    def get_parameter(self, hierarchy, parameter_name):
+    def get_parameter(self, controller, parameter_name):
         """
         Retrieve the value of a parameter from a hierarchy.
```
narrieta: "hierarchy" -> "controller"

azurelinuxagent/common/cgroup.py
```diff
@@ -13,104 +13,139 @@
...
+        self.controller = controller_type
 
+    def _get_cgroup_file(self, file_name):
+        return os.path.join(self.path, file_name)
 
-    def get_file_contents(self, hierarchy, file_name):
+    def get_file_contents(self, controller, file_name):
         """
-        Retrieve the value of a parameter from a hierarchy.
+        Retrieve the value of a parameter from a controller.
```
narrieta: >  Retrieve the value of a parameter from a controller

that is what get_parameter does... this comment should probably state that the function returns the contents of a file

azurelinuxagent/common/cgroup.py
```diff
@@ -13,104 +13,139 @@
...
-                raise CGroupsException("Could not retrieve cgroup file {0}/{1}".format(hierarchy, file_name))
+                raise CGroupsException("Could not retrieve cgroup file {0}/{1}".format(self.path, file_name))
         else:
-            raise CGroupsException("{0} subsystem not available in cgroup {1}. cgroup paths: {2}".format(
-                hierarchy, self.name, self.cgroups))
+            raise CGroupsException("Incorrect controller: {0} passed for cgroup: {1}. Cgroup path is: {2}".format(
+                controller, self.name, self.path))
 
-    def get_parameter(self, hierarchy, parameter_name):
+    def get_parameter(self, controller, parameter_name):
```
narrieta: it is too since to get confused between get_parameter and get_parameters.

probably this assumption should be verified by the code
> Assumes the parameter is the sole line of the file.

or just ged rid of get_parameter and always use get_parameters

azurelinuxagent/common/cgroup.py
```diff
@@ -13,104 +13,139 @@
...
+    def __init__(self, extension_name, cgroup_path, controller_type):
+        self.name = extension_name
+        self.path = cgroup_path
+        self.controller = controller_type
 
+    def _get_cgroup_file(self, file_name):
+        return os.path.join(self.path, file_name)
 
-    def get_file_contents(self, hierarchy, file_name):
+    def get_file_contents(self, controller, file_name):
```
narrieta: i think this should be private

azurelinuxagent/common/cgroup.py
narrieta: should be private

azurelinuxagent/common/cgroup.py
```diff
@@ -185,8 +232,16 @@ def collect(self):
...
+
+
+class CollectedMetrics(object):
+    def __init__(self, controller, metric_name, value):
+        self.controller = controller
+        self.metric_name = metric_name
+        self.value = value
 
 #
 # TODO: Do we need this code?
```
narrieta: I copy-pasted this code here in case we needed it. If we don't, let's remove it

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -13,47 +13,184 @@
...
 
     @staticmethod
     def track_cgroup(cgroup):
         """
-        Adds the given item to the list of tracked cgroups
+        Adds the given item to the dictionary of tracked cgroups
         """
-        pass  # TODO
+        # Making track thread-safe.
+        CGroupsTelemetry._rlock.acquire()
```
narrieta: let's acquire/release the lock using try/finally

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -13,47 +13,184 @@
...
+        CGroupsTelemetry._rlock.release()
 
     @staticmethod
-    def is_tracked(name):
+    def is_tracked(name, controller):
         """
         Returns true if the given item is in the list of tracked items
         """
-        pass  # TODO
+        return name in CGroupsTelemetry._tracked and controller in CGroupsTelemetry._tracked[name]
```
narrieta: should the access to _tracked be protected by the lock?

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -13,47 +13,184 @@
...
 
     @staticmethod
     def stop_tracking(name):
         """
         Stop tracking the cgroups for the given name
         """
-        if CGroupsTelemetry.is_tracked(name):
-            del (CGroupsTelemetry._tracked[name])  # TODO - thread-safeness; review implementation
+        # Pop is atomic and thread-safe.
```
narrieta: > Pop is atomic and thread-safe.

What about append?

If one of them is not thread safe then probably we should protect all of them regardless of thread-safety

azurelinuxagent/common/cgroup.py
vrdmr: Removed the controller itself. No need to handle this hot-potato. :)

azurelinuxagent/common/cgroupstelemetry.py
vrdmr: First - Implementation-wise it was easier to manage all the metrics as of type Metric. I am thinking of simplifying it with the next round of change, but this makes it simple.

```
class CgroupMetrics(object):
    def __init__(self):
        self.current_memory_usage = Metric()
        self.max_memory_levels = Metric()
        self.current_cpu_usage = Metric()
```

Also, the Max of the Max list can still give the absolute Max, and its a value which can change as well.

tests/common/test_cgroups.py
vrdmr: Yup. I am adding that in the test_extension test, which will do the entire extension_handler.run() and then do a poll_tracked_list. That unittest will go in another PR.

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -13,47 +13,184 @@
...
+            "cpu": {
+                "cur_cpu": CGroupsTelemetry._get_metrics_list(current_cpu_usage)
+            }
+        }
+        return processed_extension
 
     @staticmethod
     def metrics_hierarchies():
-        return CGroupsTelemetry._hierarchies
+        pass  # return CGroupsTelemetry._hierarchies
```
narrieta: left-over?

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -13,47 +13,184 @@
...
+                CGroupsTelemetry._tracked.remove(cgroup)
+
+        CGroupsTelemetry._rlock.release()
+
+    @staticmethod
+    def prune_all_tracked():
+        CGroupsTelemetry._rlock.acquire()
+        for cgroup in CGroupsTelemetry._tracked:
+            if not cgroup.is_active():
+                CGroupsTelemetry._tracked.remove(cgroup)
```
narrieta: can we log the cgroups we are pruning?

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -13,47 +13,184 @@
...
     @staticmethod
     def track_cgroup(cgroup):
         """
-        Adds the given item to the list of tracked cgroups
+        Adds the given item to the dictionary of tracked cgroups
         """
-        pass  # TODO
+        # Making track thread-safe.
+        CGroupsTelemetry._rlock.acquire()
+        CGroupsTelemetry._tracked.append(cgroup)
```
narrieta: can we log the cgroup we are beginning to track?

tests/common/test_cgroups.py
```diff
@@ -69,6 +67,25 @@ def i_am_root():
...
+
+def generate_metric_list(lst):
+    return [sum(lst) / len(lst),
+            min(lst),
+            max(lst),
+            median(lst),
+            len(lst)]
+
+
 @skip_if_predicate_false(lambda: False, "TODO: Need unit tests")
```
narrieta: I disabled the old tests since we were rewriting this code... could you check whether we need any of those tests and remove them if not?

azurelinuxagent/ga/monitor.py
```diff
@@ -253,13 +262,14 @@ def collect_and_send_events(self):
...
-                        MonitorHandler.CGROUP_TELEMETRY_PERIOD,
+                        MonitorHandler.CGROUP_TELEMETRY_POLLING_PERIOD,
+                        MonitorHandler.CGROUP_TELEMETRY_REPORTING_PERIOD,
                         MonitorHandler.EVENT_COLLECTION_PERIOD,
                         MonitorHandler.HOST_PLUGIN_HEARTBEAT_PERIOD,
                         MonitorHandler.IMDS_HEARTBEAT_PERIOD).seconds
         while self.should_run:
             self.send_telemetry_heartbeat()
-            self.send_cgroup_telemetry()
+            self.send_extension_metrics_telemetry()
```
narrieta: minor: name needs to be a little more generic, since this is for any cgroups (extensions, agent, services)

azurelinuxagent/ga/monitor.py
```diff
@@ -399,54 +409,38 @@ def send_telemetry_heartbeat(self):
...
-                                          is_success=True,
-                                          message=msg,
-                                          log_event=True)
-
-            except Exception as e:
-                logger.warn("Monitor: failed to collect cgroups performance metrics: {0}", ustr(e))
-                logger.verbose(traceback.format_exc())
-
-            self.last_cgroup_telemetry = datetime.datetime.utcnow()
+    def send_extension_metrics_telemetry(self):
```
narrieta: minor: can we separate this into two functions (poll and report) and call both from the main loop (makes it a little clearer that they are separate operations with different periods)

azurelinuxagent/common/cgroup.py
vrdmr: We need the name in the Cgroup metrics. Yes, I'll change the parameter of the `__init__` to `name` rather than `extension_name`

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -166,9 +167,12 @@ def start_extension_command(self, extension_name, command, cwd, env, stdout, std
                 def track_cgroups():
                     cgroups = self._cgroups_api.get_extension_cgroups(extension_name)
 
-                    # TODO: Add to tracking list
+                    for cgroup in cgroups:
```
narrieta: we also need to track the agent in create_agent_cgroups

narrieta: there is a TODO in remove_extension_cgroups that probably it is not needed anymore given the way you are pruning the tracking list (I like that)... could you remove the TODO?

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -13,47 +13,184 @@
...
+    def clear(self):
+        self.current_memory_usage.clear()
+        self.max_memory_levels.clear()
+        self.current_cpu_usage.clear()
+
+
+class Metric(object):
+    def __init__(self):
+        self.data = []
+        self.first_poll_time = dt.utcnow()
```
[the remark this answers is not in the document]
vrdmr: Refactored all the three (data/polltimes) to be private.

azurelinuxagent/ga/monitor.py
vrdmr: The `extension` in `send_extension_metrics_telemetry` was targetted more on the extension as a package (which could depend on other things on the box as well like services), rather than only focusing on extensions runtime. It could in the future send not only cgroups but extension status, python package information, etc.

azurelinuxagent/common/cgroupconfigurator.py
vrdmr: Good catch. I will add it here.

azurelinuxagent/ga/monitor.py
vrdmr: When you say - 

> call both from the main loop

With the main loop, do you mean to add another method call only for polling, in monitor.py#270? I wanted the method to be self-contained with the task of handling all the extension_metrics related operations.
```
while self.should_run:
    ...
    self.send_extension_metrics_telemetry()
    ...
```

I have tried to simplify the `send_extension_metrics_telemetry` method further. Let me know if that works. We can discuss this as well.

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -97,8 +98,8 @@ def __impl():
                 cgroups = self._cgroups_api.create_agent_cgroups()
 
                 if track_cgroups:
-                    # TODO: Add to tracking list
-                    pass
+                    for cgroup in cgroups:
+                        CGroupsTelemetry.track_cgroup(cgroup)
```
narrieta: CGroupApi.create_agent_cgroups needs to be updated to return CGroups instead of directory paths

azurelinuxagent/ga/monitor.py
narrieta: if we add more data, it is likely the reporting period will be different. I would make this specific to resource metrics

azurelinuxagent/ga/monitor.py
narrieta: Since it is 2 different operations (collect and send) with different periods I think having 2 calls from the main loop makes the intention clearer

tests/common/test_cgroups.py
```diff
@@ -304,6 +321,133 @@ def test_limits(self):
         self.assert_limits(ext_name="normal_extension", expected_cpu_limit=40, exception_raised=True)
 
 
+class TestCGroupsTelemetry(AgentTestCase):
```
narrieta: let's move this to test_cgroupstelemetry.py

azurelinuxagent/common/cgroup.py
```diff
@@ -13,104 +13,108 @@
...
-# class is very careful not to call them if CGroups isn't enabled. Any tests should be disabled if the osutil
-# is_cgroups_support() method returns false.
+re_user_system_times = re.compile(r'user (\d+)\nsystem (\d+)\n')
 
 
 class CGroup(object):
-    def __init__(self):
-        pass
+    def __init__(self, name, cgroup_path, controller_type):
+        self.name = name
```
narrieta: why do cgroups need a name? the path is already unique

tests/common/test_cgroups.py
```diff
@@ -304,6 +321,133 @@ def test_limits(self):
...
+                        patch_get_cpu_percent.return_value = 30
+                        patch_get_memory_usage.return_value = 209715200  # example 200 MB
+                        patch_get_memory_max_usage.return_value = 471859200  # example 450 MB
+
+                        CGroupsTelemetry.poll_all_tracked()
+
+                        self.assertEqual(CGroupsTelemetry._cgroup_metrics.__len__(), num_extensions)
+                        for cgroup_name, cgroup_metric in CGroupsTelemetry._cgroup_metrics.items():
+                            current_memory_usage, max_memory_levels, current_cpu_usage = cgroup_metric.get_metrics()
+                            self.assertEqual(len(current_memory_usage._data), 1)
```
narrieta: should we check the values too? (not only the length)

azurelinuxagent/ga/monitor.py
```diff
@@ -399,54 +409,39 @@ def send_telemetry_heartbeat(self):
...
+
+        if not self.last_cgroup_polling_telemetry:
+            self.last_cgroup_polling_telemetry = time_now
+
+        if not self.last_cgroup_report_telemetry:
+            self.last_cgroup_report_telemetry = time_now
+
+        if time_now >= (self.last_cgroup_polling_telemetry +
+                        MonitorHandler.CGROUP_TELEMETRY_POLLING_PERIOD):
+            CGroupsTelemetry.poll_all_tracked()
```
narrieta: The names for poll_all_tracked and collect_all_tracked are too similar... it is hard to see at a glance what they are doing. 

I think CGROUP_TELEMETRY_POLLING_PERIOD and CGROUP_TELEMETRY_REPORTING_PERIOD express very clearly their intent. Can we rename to include "reporting" somewhere?

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -13,47 +13,188 @@
...
-        pass  # TODO
+        collected_metrics = {}
+        for name, cgroup_metrics in CGroupsTelemetry._cgroup_metrics.items():
+            collected_metrics[name] = CGroupsTelemetry._process_cgroup_metric(cgroup_metrics)
+            cgroup_metrics.clear()
+
+            if cgroup_metrics.marked_for_delete:
+                # Clear the _cgroup_metrics to make sure we don't send stale entries,
+                # only after sending its content first
+                CGroupsTelemetry._cgroup_metrics.pop(name, None)
```
narrieta: > only after sending its content first

Yes, that's what it should do... but we are clearing before sending

tests/common/test_cgroups.py
narrieta: The current tests cover poll and collect... we need tests for the rest of the methods

azurelinuxagent/common/cgroupstelemetry.py
vrdmr: Changed this to use `with` as suggested in [here](https://stackoverflow.com/a/25704099). This takes care of releasing the lock if we hit an issue.

Thanks for catching this.

azurelinuxagent/common/cgroupstelemetry.py
vrdmr: Its metrics are captured in line 81 - `collected_metrics[name] = CGroupsTelemetry._process_cgroup_metric(cgroup_metrics)`, and we return `collected_metrics` from `collect_all_tracked()`

azurelinuxagent/common/cgroup.py
vrdmr: The Name helps me to create a new cgroup_metric of the same name, which is used in the json metric as well.

Example: in `azurelinuxagent/common/cgroupstelemetry.py:96`
`CGroupsTelemetry._cgroup_metrics[cgroup.name] = CgroupMetrics()`

azurelinuxagent/common/cgroup.py
```diff
@@ -152,29 +156,41 @@ def collect(self):
...
-    def get_memory_usage(self):
+    def _get_memory_usage(self):
         """
         Collect memory.usage_in_bytes from the cgroup.
 
         :return: Memory usage in bytes
         :rtype: int
         """
-        usage = self.cgt.cgroup.get_parameter('memory', 'memory.usage_in_bytes')
+        usage = self.get_parameters('memory.usage_in_bytes', first_line_only=True)
```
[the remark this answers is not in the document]
vrdmr: This particular file contains only one value and the `first_line_only` in `get_parameters()` is only to get that value. I could do [0] by the calling function, but that brings some ambiguity.

azurelinuxagent/common/cgroup.py
vrdmr: Returning None here and handling it upstream.

azurelinuxagent/common/cgroup.py
```diff
@@ -13,104 +13,108 @@
...
+re_user_system_times = re.compile(r'user (\d+)\nsystem (\d+)\n')
 
 
 class CGroup(object):
-    def __init__(self):
-        pass
+    def __init__(self, name, cgroup_path, controller_type):
+        self.name = name
+        self.path = cgroup_path
+        self.controller = controller_type
```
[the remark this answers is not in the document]
vrdmr: It is used inside the CGroupMetrics - The structure of the CGroupMetrics will be changed with the next PR, and controller will be removed then.

```
for cgroup in CGroupsTelemetry._tracked:
    if cgroup.name not in CGroupsTelemetry._cgroup_metrics:
        CGroupsTelemetry._cgroup_metrics[cgroup.name] = CgroupMetrics()
    CGroupsTelemetry._cgroup_metrics[cgroup.name].add_new_data(cgroup.controller, cgroup.collect())
```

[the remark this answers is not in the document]
narrieta: cool

azurelinuxagent/common/cgroup.py
narrieta: still, you could use the path as key instead of the name... the path is guaranteed to be unique... as we add new cgroups to track services, cron jobs and what not we will need to come up with a way to name these and to ensure the name remains unique... path does not have that issue.

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -13,47 +13,180 @@
...
 
     @staticmethod
-    def is_tracked(name):
+    def is_tracked(name, controller):
         """
         Returns true if the given item is in the list of tracked items
+        O(n) operation. But limited to few cgroup objects we have.
         """
-        pass  # TODO
+        for cgroup in CGroupsTelemetry._tracked:
```
narrieta: You need to hold the lock while iterating thru _tracked

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -13,47 +13,180 @@
...
 
     @staticmethod
-    def stop_tracking(name):
+    def stop_tracking(cgroup):
         """
         Stop tracking the cgroups for the given name
         """
-        if CGroupsTelemetry.is_tracked(name):
-            del (CGroupsTelemetry._tracked[name])  # TODO - thread-safeness; review implementation
+        CGroupsTelemetry._tracked.remove(cgroup)
```
narrieta: please check that all uses of _tracked are protected by the lock

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -13,47 +13,180 @@
...
+        with CGroupsTelemetry._rlock:
+            for cgroup in CGroupsTelemetry._tracked[:]:
+                if not cgroup.is_active():
+                    CGroupsTelemetry.stop_tracking(cgroup)
+
+    @staticmethod
+    def cleanup():
+        CGroupsTelemetry._tracked *= 0  # emptying the list
+
+    def __init__(self):
```
narrieta: leftover?

tests/common/test_cgroupstelemetry.py
```diff
@@ -162,3 +160,81 @@ def test_telemetry_calculations(self, *args):
...
+        num_extensions = 5
+        num_controllers = 2
+        for i in range(num_extensions):
+            dummy_cpu_cgroup = CGroup.create("dummy_path", "cpu", "dummy_extension_{0}".format(i))
+            CGroupsTelemetry.track_cgroup(dummy_cpu_cgroup)
+
+            dummy_memory_cgroup = CGroup.create("dummy_path", "memory", "dummy_extension_{0}".format(i))
+            CGroupsTelemetry.track_cgroup(dummy_memory_cgroup)
+
+        self.assertEqual(num_extensions * num_controllers, len(CGroupsTelemetry._tracked))
```
narrieta: a better assertion could be that is_tracked returns true for all the items you added (that also has the advantage of testing via the public interface, rather than reaching into _tracked)

tests/common/test_cgroupstelemetry.py
```diff
@@ -162,3 +160,81 @@ def test_telemetry_calculations(self, *args):
...
+
+        CGroupsTelemetry.prune_all_tracked()
+        self.assertEqual(0, len(CGroupsTelemetry._tracked))
+
+        for i in CGroupsTelemetry._tracked:
```
narrieta: why is this needed?

azurelinuxagent/common/cgroupstelemetry.py
vrdmr: This would cause locking issues in the `poll_all_tracked`. The `poll_all_tracked` has to lock to make sure it is polling data on the latest stable list. The `CGroupsTelemetry.stop_tracking(cgroup)` would hit the lock acquisition failure.

Current implementation:
```
def poll_all_tracked():
    with CGroupsTelemetry._rlock:
        for cgroup in CGroupsTelemetry._tracked[:]:
            if cgroup.name not in CGroupsTelemetry._cgroup_metrics:
                CGroupsTelemetry._cgroup_metrics[cgroup.name] = CgroupMetrics()
            CGroupsTelemetry._cgroup_metrics[cgroup.name].add_new_data(cgroup.controller, cgroup.collect())

            if not cgroup.is_active():
                CGroupsTelemetry.stop_tracking(cgroup)
                CGroupsTelemetry._cgroup_metrics[cgroup.name].marked_for_delete = True
```

If you want the locking to only happen in primitive operations such as `stop_tracking` and `track_cgroup`, It will still work.

tests/common/test_cgroupstelemetry.py
vrdmr: leftover. Cleaned

PR #1541, opened 2019-06-03
Adding more unittests to enable cgroup telemetry tests

vrdmr:
## Description
Adding newer unit tests to test the E2E scenario for exercising the cgroups telemetry with mocked and actual cgroup.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1541)

### DIFF
--- a/azurelinuxagent/common/cgroupapi.py
+++ b/azurelinuxagent/common/cgroupapi.py
@@ -209,6 +209,8 @@ class FileSystemCgroupsApi(CGroupsApi):
         """
         Creates the directory within the cgroups file system that will contain the cgroups for the extensions.
         """
+        cgroups = []
+
         def create_cgroup(controller):
             path = self._get_extension_cgroups_root_path(controller)
 
@@ -216,7 +218,11 @@ class FileSystemCgroupsApi(CGroupsApi):
                 FileSystemCgroupsApi._try_mkdir(path)
                 logger.info("Created {0}".format(path))
 
+            cgroups.append(path)
+
         self._foreach_controller(create_cgroup, 'Will not create a root cgroup for extensions')
+
+        return cgroups
 
     def create_extension_cgroups(self, extension_name):
         """
--- a/azurelinuxagent/common/cgroupapi.py
+++ b/azurelinuxagent/common/cgroupapi.py
@@ -209,8 +209,6 @@ class FileSystemCgroupsApi(CGroupsApi):
         """
         Creates the directory within the cgroups file system that will contain the cgroups for the extensions.
         """
-        cgroups = []
-
         def create_cgroup(controller):
             path = self._get_extension_cgroups_root_path(controller)
 
@@ -218,11 +216,7 @@ class FileSystemCgroupsApi(CGroupsApi):
                 FileSystemCgroupsApi._try_mkdir(path)
                 logger.info("Created {0}".format(path))
 
-            cgroups.append(path)
-
         self._foreach_controller(create_cgroup, 'Will not create a root cgroup for extensions')
-
-        return cgroups
 
     def create_extension_cgroups(self, extension_name):
         """
--- a/azurelinuxagent/common/cgroupconfigurator.py
+++ b/azurelinuxagent/common/cgroupconfigurator.py
@@ -121,7 +121,7 @@ class CGroupConfigurator(object):
             def __impl():
                 return self._cgroups_api.create_extension_cgroups(name)
 
-            self._invoke_cgroup_operation(__impl, "Failed to create a cgroup for extension '{0}'; resource usage will not be tracked".format(name))
+            return self._invoke_cgroup_operation(__impl, "Failed to create a cgroup for extension '{0}'; resource usage will not be tracked".format(name))
 
         def remove_extension_cgroups(self, name):
             """
--- a/azurelinuxagent/common/cgroupconfigurator.py
+++ b/azurelinuxagent/common/cgroupconfigurator.py
@@ -185,7 +185,3 @@ class CGroupConfigurator(object):
         if CGroupConfigurator._instance is None:
             CGroupConfigurator._instance = CGroupConfigurator.__impl()
         return CGroupConfigurator._instance
-
-    @staticmethod
-    def is_cgroups_supported():
-        return CGroupConfigurator.get_instance().enabled()
--- a/azurelinuxagent/common/cgroupconfigurator.py
+++ b/azurelinuxagent/common/cgroupconfigurator.py
@@ -185,3 +185,7 @@ class CGroupConfigurator(object):
         if CGroupConfigurator.__instance is None:
             CGroupConfigurator.__instance = CGroupConfigurator.__impl()
         return CGroupConfigurator.__instance
+
+    @staticmethod
+    def is_cgroups_supported():
+        return CGroupConfigurator.get_instance().enabled()
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -16,7 +16,6 @@
 # Requires Python 2.6+ and Openssl 1.0+
 #
 
-import array
 import base64
 import datetime
 import errno
@@ -32,22 +31,21 @@ import socket
 import struct
 import sys
 import time
+from pwd import getpwall
 
-import azurelinuxagent.common.logger as logger
+import array
+
 import azurelinuxagent.common.conf as conf
+import azurelinuxagent.common.logger as logger
 import azurelinuxagent.common.utils.fileutil as fileutil
 import azurelinuxagent.common.utils.shellutil as shellutil
 import azurelinuxagent.common.utils.textutil as textutil
-
 from azurelinuxagent.common.exception import OSUtilError
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.utils.cryptutil import CryptUtil
 from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
 from azurelinuxagent.common.utils.networkutil import RouteEntry, NetworkInterfaceCard
-
-from pwd import getpwall
-
-from azurelinuxagent.common.version import PY_VERSION_MAJOR, PY_VERSION_MINOR
+from azurelinuxagent.common.version import DISTRO_CODE_NAME
 
 __RULES_FILES__ = [ "/lib/udev/rules.d/75-persistent-net-generator.rules",
                     "/etc/udev/rules.d/70-persistent-net.rules" ]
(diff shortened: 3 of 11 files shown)


azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -121,7 +121,7 @@ def create_extension_cgroups(self, name):
             def __impl():
                 return self._cgroups_api.create_extension_cgroups(name)
 
-            self._invoke_cgroup_operation(__impl, "Failed to create a cgroup for extension '{0}'; resource usage will not be tracked".format(name))
+            return self._invoke_cgroup_operation(__impl, "Failed to create a cgroup for extension '{0}'; resource usage will not be tracked".format(name))
```
vrdmr: This was a bug in the [previous impl](https://github.com/Azure/WALinuxAgent/blob/cgroups/azurelinuxagent/ga/exthandlers.py#L871) which was not returning cgroups when created.

```
self.cgroups = CGroupConfigurator.get_instance().
create_extension_cgroups(self.get_full_name())
```

azurelinuxagent/common/osutil/default.py
```diff
@@ -299,12 +301,19 @@ def is_current_instance_id(self, id_that):
...
-        Enabled by default; disabled in WSL/Travis
+        Enabled by default; disabled in WSL/Travis (Disabled on Trusty only)
         """
         is_wsl = '-Microsoft-' in platform.platform()
         is_travis = 'TRAVIS' in os.environ and os.environ['TRAVIS'] == 'true'
+
+        supported_in_travis = True  # By default supported in Travis
+
+        # Travis with python 2.6 fails (on Trusty based systems)
+        if is_travis and PY_VERSION_MAJOR == 2 and PY_VERSION_MINOR == 6:
```
vrdmr: This would be removed once validated.

narrieta: The check should not be based on the Python version, but on the platform (trusty). Also, if all the other platforms Travis runs on support cgroups, then we should remove the check on Travis altogether.

tests/common/test_cgroupstelemetry.py
```diff
@@ -44,6 +47,49 @@ def generate_metric_list(lst):
...
+    waste = []
+    for x in range(1, 3):
+        waste.append([random.random()] * 10000)
+        time.sleep(0.1)
+        waste *= 0
+    return waste
+
+
+def make_root_cgroup():
+    cpu_controller_id = CGroupsApi._get_controller_id('cpu')
```
narrieta: Try to write the tests using the public interface instead of private methods when possible. You could use the public interface to create a couple of test cgroups. Reaching into the private methods makes the tests more fragile (they may start failing or doing incorrect stuff if the internal implementation changes)

tests/common/test_cgroupstelemetry.py
```diff
@@ -392,6 +438,51 @@ def test_cgroup_is_tracked(self):
         self.assertFalse(CGroupsTelemetry.is_tracked("not_present_dummy_extension_{0}".format(i), 'cpu'))
         self.assertFalse(CGroupsTelemetry.is_tracked("not_present_dummy_extension_{0}".format(i), 'memory'))
 
+    @skip_if_predicate_false(CGroupConfigurator.get_instance().enabled, "CGroups not supported in this environment")
```
narrieta: Let's expose CGroupConfigurator._cgroups_supported as a public method and do this kind of checks based on that method

tests/common/test_cgroupstelemetry.py
```diff
@@ -392,6 +438,51 @@ def test_cgroup_is_tracked(self):
...
+        name = "test-cgroup"
+
+        cgs = make_self_cgroup(name)
+
+        self.assertEqual(len(cgs), 2)
+
+        for cgroup in cgs:
+            CGroupsTelemetry.track_cgroup(cgroup)
+
+        for i in range(num_polls):
```
narrieta: I don't quite get the intention of the loop...
It'd be nice if you can consume memory/cpu in parallel this loop and check that each iteration reports an increasing amount of resources. That would also allow you to make an assertion stronger than >= 0

tests/tools.py
```diff
@@ -359,6 +361,12 @@ def emulate_assertListEqual(self, seq1, seq2, msg=None, seq_type=None):
         msg = self._formatMessage(msg, standardMsg)
         self.fail(msg)
 
+    def emulate_assertIsInstance(self, obj, object_type, msg=None):
+        if not type(obj) == object_type:
```
narrieta: why not use isintance?

tests/common/test_cgroupstelemetry.py
vrdmr: This loop is consuming `cpu` and `memory` after every poll. The issue with tracking the self cgroup is that it cannot be a long-running extension, and path of self is only `/sys/fs/cgroup/cpu` & `/sys/fs/cgroup/memory`. And for more exact calculation, it'll require a different test in the E2E side. I can add that task.

```
for i in range(num_polls):
    CGroupsTelemetry.poll_all_tracked()
    consume_cpu_time()  # Eat some CPU
    consume_memory()  # Eat some Memory
```

azurelinuxagent/common/cgroupapi.py
```diff
@@ -209,15 +209,21 @@ def create_extension_cgroups_root(self):
         """
         Creates the directory within the cgroups file system that will contain the cgroups for the extensions.
         """
+        cgroups = []
```
narrieta: Do we ever need these cgroups? (for the root directory of all extensions?)

tests/common/test_cgroupstelemetry.py
```diff
@@ -63,34 +71,57 @@ def consume_memory():
...
+        self.log_dir = os.path.join(self.temp_dir, "log")
+        self.mock_get_log_dir = patch("azurelinuxagent.ga.exthandlers.ExtHandlerInstance.get_log_dir",
+                                      lambda *_: self.log_dir)
+        self.mock_get_log_dir.start()
+
+    def tearDown(self):
+        if self.temp_dir is not None:
+            shutil.rmtree(self.temp_dir)
+
+    def _create_script(self, file_name, contents):
```
narrieta: we should not duplicate this code... if needed across multiple tests let's move it to a common module

tests/common/test_cgroupstelemetry.py
```diff
@@ -438,47 +469,59 @@ def test_cgroup_is_tracked(self):
         self.assertFalse(CGroupsTelemetry.is_tracked("not_present_dummy_extension_{0}".format(i), 'cpu'))
         self.assertFalse(CGroupsTelemetry.is_tracked("not_present_dummy_extension_{0}".format(i), 'memory'))
 
+    @skip_if_predicate_false(i_am_root, "This test will only run as root")
     @skip_if_predicate_false(CGroupConfigurator.is_cgroups_supported, "CGroups not supported in this environment")
     def test_telemetry_with_tracked_cgroup(self):
-        num_polls = 5
-        name = "test-cgroup"
+        CGroupsTelemetry.cleanup()
```
narrieta: since CGroupsTelemetry is a singleton shared by all these tests, should you do the cleanup in the test setUp method

tests/common/test_cgroupstelemetry.py
```diff
@@ -438,47 +469,59 @@ def test_cgroup_is_tracked(self):
...
-        self.assertIsInstance(collected_metrics[name]["cpu"]["cur_cpu"][5], str)
-        self.assertIsInstance(collected_metrics[name]["cpu"]["cur_cpu"][6], str)
+        self.assertIsInstance(collected_metrics[extn_name]["cpu"]["cur_cpu"][5], str)
+        self.assertIsInstance(collected_metrics[extn_name]["cpu"]["cur_cpu"][6], str)
 
         for i in range(5):
-            self.assertGreater(collected_metrics[name]["memory"]["cur_mem"][i], 0)
-            self.assertGreater(collected_metrics[name]["memory"]["max_mem"][i], 0)
-            self.assertGreaterEqual(collected_metrics[name]["cpu"]["cur_cpu"][i], 0)
+            self.assertGreater(collected_metrics[extn_name]["memory"]["cur_mem"][i], 0)
```
narrieta: can we have an assertion stronger than > 0

azurelinuxagent/common/osutil/default.py
```diff
@@ -299,12 +301,17 @@ def is_current_instance_id(self, id_that):
...
         """
-        Enabled by default; disabled in WSL/Travis
+        Enabled by default; disabled in WSL/Travis (Disabled on Trusty only)
         """
         is_wsl = '-Microsoft-' in platform.platform()
-        is_travis = 'TRAVIS' in os.environ and os.environ['TRAVIS'] == 'true'
+        supported = True  # By default supported in Travis
+
+        # Travis with python 2.6 fails (on Trusty based systems)
+        if PY_VERSION_MAJOR == 2 and PY_VERSION_MINOR == 6:
```
narrieta: this test should be based on the platform (trusty) rather than on the python version

azurelinuxagent/common/cgroupapi.py
vrdmr: No. Removed this part. I needed it for some testing.

tests/common/test_cgroupstelemetry.py
vrdmr: Currently, I cannot estimate the exact or approximate value of the memory or CPU consumed. But it should be greater than 0 if cgroups did start tracking the cgroup correctly.

azurelinuxagent/common/cgroupconfigurator.py
narrieta: create_agent_cgroups has the same issue, could you add a return there as well?

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -185,3 +185,7 @@ def get_instance():
         if CGroupConfigurator._instance is None:
             CGroupConfigurator._instance = CGroupConfigurator.__impl()
         return CGroupConfigurator._instance
+
+    @staticmethod
+    def is_cgroups_supported():
+        return CGroupConfigurator.get_instance().enabled()
```
narrieta: this should return self._cgroups_supported

azurelinuxagent/common/osutil/default.py
```diff
@@ -299,12 +299,17 @@ def is_current_instance_id(self, id_that):
     @staticmethod
     def is_cgroups_supported():
         """
-        Enabled by default; disabled in WSL/Travis
+        Enabled by default; disabled in WSL/Travis (Trusty only)
```
narrieta: We need to remove the reference to Travis in this comment

azurelinuxagent/common/osutil/default.py
```diff
@@ -299,12 +299,17 @@ def is_current_instance_id(self, id_that):
...
         """
         is_wsl = '-Microsoft-' in platform.platform()
-        is_travis = 'TRAVIS' in os.environ and os.environ['TRAVIS'] == 'true'
+        supported = True
         base_fs_exists = os.path.exists(BASE_CGROUPS)
-        return not is_wsl and not is_travis and base_fs_exists
+
+        # Travis with python 2.6 fails on Trusty based systems
```
narrieta: We need to remove the reference to Travis in this comment

tests/common/test_cgroupstelemetry.py
narrieta: ok

azurelinuxagent/common/osutil/default.py
```diff
@@ -299,13 +299,13 @@ def is_current_instance_id(self, id_that):
     @staticmethod
     def is_cgroups_supported():
         """
-        Enabled by default; disabled in WSL/Travis (Trusty only)
+        Enabled by default; disabled in WSL (Trusty only)
```
narrieta: nitpick: "disabled in WSL and Trusty"

Issue #1550, opened 2019-06-11
walinuxagent fails to start when host is using EC keys [BUG]


johncrim:
We're seeing the following error messages every 10 seconds on new Ubuntu hosts. Needless to say, none of the VM extensions are running, so we're unable to create new hosts.

```
2019/06/11 04:48:55.591593 INFO Daemon WireServer endpoint is not found. Rerun dhcp handler
2019/06/11 04:48:55.592240 INFO Daemon Test for route to 168.63.129.16
2019/06/11 04:48:55.594100 INFO Daemon Route to 168.63.129.16 exists
2019/06/11 04:48:55.595355 INFO Daemon Wire server endpoint:168.63.129.16
2019/06/11 04:48:55.609510 INFO Daemon Fabric preferred wire protocol version:2015-04-05
2019/06/11 04:48:55.610492 INFO Daemon Wire protocol version:2012-11-30
2019/06/11 04:48:55.612152 INFO Daemon Server preferred version:2015-04-05
2019/06/11 04:48:55.949508 ERROR Daemon Command: [/usr/bin/openssl rsa -in /var/lib/waagent/1.prv -pubout 2>/dev/null], return code: [1], result: []
2019/06/11 04:48:55.962315 ERROR Daemon Command: [/usr/bin/openssl rsa -in /var/lib/waagent/2.prv -pubout 2>/dev/null], return code: [1], result: []
2019/06/11 04:48:56.040907 ERROR Daemon Command: [/usr/bin/openssl rsa -in /var/lib/waagent/1.prv -pubout 2>/dev/null], return code: [1], result: []
2019/06/11 04:48:56.053502 ERROR Daemon Command: [/usr/bin/openssl rsa -in /var/lib/waagent/2.prv -pubout 2>/dev/null], return code: [1], result: []
2019/06/11 04:48:56.132301 ERROR Daemon Command: [/usr/bin/openssl rsa -in /var/lib/waagent/1.prv -pubout 2>/dev/null], return code: [1], result: []
2019/06/11 04:48:56.140927 ERROR Daemon Command: [/usr/bin/openssl rsa -in /var/lib/waagent/2.prv -pubout 2>/dev/null], return code: [1], result: []
2019/06/11 04:48:56.169281 ERROR Daemon Exception processing goal state, giving up: ['']
2019/06/11 04:48:56.169792 INFO Daemon WireServer is not responding. Reset endpoint
[...]

PR #1551, opened 2019-06-11
Support both ECC and RSA keys during initialization

johncrim:
* `openssl pkey` supports both RSA and ECC private keys

Fixes: #1550

Pretty minor fix - the commands work locally (I've tested with both RSA and ECC .pkey files), but I'm not able to update the `waagent20` python file on my VM - when I make this change to `/usr/sbin/waagent2.0`, it doesn't seem to get picked up, for some reason.

### Quality of Code and Contribution Guidelines
- [X ] I have read the [contribution guidelines](https://github.com/Azure/WALinuxAgent/blob/master/.github/CONTRIBUTING.md).

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1551)

PR #1552, opened 2019-06-11
Support both ECC and RSA keys during initialization

johncrim:
* `openssl pkey` supports both RSA and ECC private keys

Fixes: #1550

## Description

Issue #1550 

---

### PR information
- [X] The title of the PR is clear and informative.
- [X] Except for special cases involving multiple contributors, the PR is started from a fork of the main repository, not a branch.
- [X] If applicable, the PR references the bug/issue that it fixes in the description.
- [X] New Unit tests were added for the changes made and Travis.CI is passing.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1552)

### DIFF
--- a/bin/waagent2.0
+++ b/bin/waagent2.0
@@ -3292,7 +3292,7 @@ class Certificates(object):
         index = 1
         filename = str(index) + ".prv"
         while os.path.isfile(filename):
-            pubkey = RunGetOutput(Openssl + " rsa -in " + filename + " -pubout 2> /dev/null ")[1]
+            pubkey = RunGetOutput(Openssl + " pkey -in " + filename + " -pubout 2> /dev/null ")[1]
             os.rename(filename, keys[pubkey] + ".prv")
             os.chmod(keys[pubkey] + ".prv", 0600)
             MyDistro.setSelinuxContext( keys[pubkey] + '.prv','unconfined_u:object_r:ssh_home_t:s0')
--- a/azurelinuxagent/common/utils/cryptutil.py
+++ b/azurelinuxagent/common/utils/cryptutil.py
@@ -33,6 +33,7 @@ import azurelinuxagent.common.utils.textutil as textutil
 
 DECRYPT_SECRET_CMD = "{0} cms -decrypt -inform DER -inkey {1} -in /dev/stdin"
 
+
 class CryptUtil(object):
     def __init__(self, openssl_cmd):
         self.openssl_cmd = openssl_cmd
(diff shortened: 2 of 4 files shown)


johncrim:
@vrdmr - this is the new PR, replacing https://github.com/Azure/WALinuxAgent/pull/1551

vrdmr:
Running the change through our end-to-end testing workflow.

bin/waagent2.0
```diff
@@ -3292,7 +3292,7 @@ class Certificates(object):
         index = 1
         filename = str(index) + ".prv"
         while os.path.isfile(filename):
-            pubkey = RunGetOutput(Openssl + " rsa -in " + filename + " -pubout 2> /dev/null ")[1]
+            pubkey = RunGetOutput(Openssl + " pkey -in " + filename + " -pubout 2> /dev/null ")[1]
```
vrdmr: I think this is the wrong place to make the change in. We don't use waagent2.0 anymore.

You'll need to make a change in `get_pubkey_from_prv` method of the `CryptUtil` class located in the `azurelinuxagent/common/utils/cryptutil.py` file.

```
        else:
            cmd = "{0} rsa -in {1} -pubout 2>/dev/null".format(self.openssl_cmd,
                                                               file_name)
            pub = shellutil.run_get_output(cmd)[1]
            return pub
```

Also, when you make the change in the new location, could you please add a corresponding change in `tests/utils/test_crypt_util.py` as well? 

Thanks

johncrim: ok - I had no idea about walinuxagent not using waagent2.0 anymore. I'm also not a python developer, so it might be better for someone with more python expertise to fix this.

Can you tell me how I can test this on an Azure VM, to make sure it works? [See my comment here, perhaps my workaround for testing this was wrong](https://github.com/Azure/WALinuxAgent/issues/1550#issuecomment-501027678).

johncrim: @vrdmr - I've made the additional change you requested to `cryptutil.py`.

I didn't find any tests in `test_crypt_util.py` referencing `get_pubkey_from_prv()`, so I didn't add a test case for a non RSA cert either. I understand the desirability of adding one, but I'm afraid my attempts at adding a test case would not be a good use of time for either of us. Otoh, this would mostly be testing openssl.

I'd be happy to help test this in Azure, if you can give me guidance on the best way to test this on a VM/vmset.

vrdmr: @johncrim - I added couple of unittests. Please let me know if it looks good. Thanks.

PR #1553, opened 2019-06-11
Preventing incorrect metrics to be sent on telemetry

vrdmr:
## Description

Issue # 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1553)

### DIFF
--- a/azurelinuxagent/common/cgroup.py
+++ b/azurelinuxagent/common/cgroup.py
@@ -104,9 +104,9 @@ class CGroup(object):
             if tasks:
                 return len(tasks) != 0
         except CGroupsException as e:
-            logger.periodic(logger.EVERY_HALF_HOUR, 'Could not get list of tasks from "tasks" file in the cgroup: {0}.'
-                                                    ' Internal error: {1}'.format(self.path, ustr(e)),
-                            logger.LogLevel.WARNING)
+            logger.periodic_warn(logger.EVERY_HALF_HOUR,
+                                 'Could not get list of tasks from "tasks" file in the cgroup: {0}.'
+                                 ' Internal error: {1}'.format(self.path, ustr(e)))
             return False
 
         return False
@@ -127,9 +127,9 @@ class CpuCgroup(CGroup):
         try:
             self._current_cpu_total = self._get_current_cpu_total()
         except CGroupsException as e:
-            logger.periodic(logger.EVERY_HALF_HOUR,
-                            'Could not get current CPU total usage from cgroup: {0}. Internal error :{1}'.format(
-                                self.path, ustr(e)), logger.LogLevel.WARNING)
+            logger.periodic_warn(logger.EVERY_HALF_HOUR,
+                                 'Could not get current CPU total usage from cgroup: {0}. '
+                                 'Internal error :{1}'.format(self.path, ustr(e)))
 
         self._previous_cpu_total = 0
         self._current_system_cpu = self._osutil.get_total_cpu_ticks_since_boot()
--- a/azurelinuxagent/common/cgroup.py
+++ b/azurelinuxagent/common/cgroup.py
@@ -87,8 +87,8 @@ class CGroup(object):
             parameter_filename = self._get_cgroup_file(parameter_name)
             logger.error("File {0} is empty but should not be".format(parameter_filename))
             raise CGroupsException("File {0} is empty but should not be".format(parameter_filename))
-        except CGroupsException as e:
-            raise e
+        except CGroupsException:
+            raise
         except Exception as e:
             parameter_filename = self._get_cgroup_file(parameter_name)
             logger.error("Exception while attempting to read {0}: {1}".format(parameter_filename, ustr(e)))
@@ -185,12 +185,9 @@ class CpuCgroup(CGroup):
 
         :rtype: [(str, str, float)]
         """
-        try:
-            self._update_cpu_data()
-            usage = self._get_cpu_percent()
-            return [CollectedMetrics("cpu", "% Processor Time", usage)]
-        except CGroupsException as e:
-            raise e
+        self._update_cpu_data()
+        usage = self._get_cpu_percent()
+        return [CollectedMetrics("cpu", "% Processor Time", usage)]
 
 
 class MemoryCgroup(CGroup):
@@ -214,15 +211,11 @@ class MemoryCgroup(CGroup):
         :return: Memory usage in bytes
         :rtype: int
         """
-        try:
-            usage = self._get_parameters('memory.usage_in_bytes', first_line_only=True)
+        usage = self._get_parameters('memory.usage_in_bytes', first_line_only=True)
 
-            if not usage:
-                usage = "0"
-            return int(usage)
-
-        except CGroupsException as e:
-            raise e
+        if not usage:
+            usage = "0"
+        return int(usage)
 
     def _get_memory_max_usage(self):
         """
@@ -231,13 +224,10 @@ class MemoryCgroup(CGroup):
         :return: Memory usage in bytes
         :rtype: int
         """
-        try:
-            usage = self._get_parameters('memory.max_usage_in_bytes', first_line_only=True)
-            if not usage:
-                usage = "0"
-            return int(usage)
-        except CGroupsException as e:
-            raise e
+        usage = self._get_parameters('memory.max_usage_in_bytes', first_line_only=True)
+        if not usage:
+            usage = "0"
+        return int(usage)
 
     def collect(self):
         """
@@ -245,13 +235,10 @@ class MemoryCgroup(CGroup):
 
         :rtype: [(str, str, float)]
         """
-        try:
-            usage = self._get_memory_usage()
-            max_usage = self._get_memory_max_usage()
-            return [CollectedMetrics("memory", "Total Memory Usage", usage),
-                    CollectedMetrics("memory", "Max Memory Usage", max_usage)]
-        except CGroupsException as e:
-            raise e
+        usage = self._get_memory_usage()
+        max_usage = self._get_memory_max_usage()
+        return [CollectedMetrics("memory", "Total Memory Usage", usage),
+                CollectedMetrics("memory", "Max Memory Usage", max_usage)]
 
 
 class CollectedMetrics(object):
--- a/azurelinuxagent/common/cgroup.py
+++ b/azurelinuxagent/common/cgroup.py
@@ -86,22 +86,27 @@ class CGroup(object):
         except IndexError:
             parameter_filename = self._get_cgroup_file(parameter_name)
             logger.error("File {0} is empty but should not be".format(parameter_filename))
-        except CGroupsException:
-            return None
+            raise CGroupsException("File {0} is empty but should not be".format(parameter_filename))
+        except CGroupsException as e:
+            raise e
         except Exception as e:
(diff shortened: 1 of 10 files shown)


azurelinuxagent/common/cgroup.py
```diff
@@ -175,9 +185,12 @@ def collect(self):
...
         :rtype: [(str, str, float)]
         """
-        self._update_cpu_data()
-        usage = self._get_cpu_percent()
-        return [CollectedMetrics("cpu", "% Processor Time", usage)]
+        try:
+            self._update_cpu_data()
+            usage = self._get_cpu_percent()
+            return [CollectedMetrics("cpu", "% Processor Time", usage)]
+        except CGroupsException as e:
```
narrieta: not needed

narrieta: (since you are re-raising the exception without any other action)

azurelinuxagent/common/cgroup.py
```diff
@@ -201,10 +214,15 @@ def _get_memory_usage(self):
...
-            usage = "0"
-        return int(usage)
+        try:
+            usage = self._get_parameters('memory.usage_in_bytes', first_line_only=True)
+
+            if not usage:
+                usage = "0"
+            return int(usage)
+
+        except CGroupsException as e:
```
narrieta: not needed

azurelinuxagent/common/cgroup.py
```diff
@@ -86,23 +86,28 @@ def _get_parameters(self, parameter_name, first_line_only=False):
         except IndexError:
             parameter_filename = self._get_cgroup_file(parameter_name)
             logger.error("File {0} is empty but should not be".format(parameter_filename))
-        except CGroupsException:
-            return None
+            raise CGroupsException("File {0} is empty but should not be".format(parameter_filename))
+        except CGroupsException as e:
+            raise e
```
narrieta: does this change the traceback in the exception?

maybe this instead?
>
       except CGroupsException:
            raise

azurelinuxagent/common/cgroup.py
```diff
@@ -143,7 +153,7 @@ def _get_current_cpu_total(self):
             # has not yet started (or has stopped) an associated service on a VM using systemd, the cgroup for
             # the service will not exist ('cause systemd will tear it down). This might be a transient or a
             # long-lived state, so there's no point in logging it, much less emitting telemetry.
-            pass
+            raise
```
narrieta: what is the intention of this change? the caller is still setting the value to 0

azurelinuxagent/common/cgroup.py
```diff
@@ -213,21 +231,27 @@ def _get_memory_max_usage(self):
...
-        if not usage:
-            usage = "0"
-        return int(usage)
+        try:
+            usage = self._get_parameters('memory.max_usage_in_bytes', first_line_only=True)
+            if not usage:
+                usage = "0"
+            return int(usage)
+        except CGroupsException as e:
+            raise e
```
narrieta: not needed

azurelinuxagent/common/cgroup.py
```diff
@@ -213,21 +231,27 @@ def _get_memory_max_usage(self):
...
-        max_usage = self._get_memory_max_usage()
-        return [CollectedMetrics("memory", "Total Memory Usage", usage),
-                CollectedMetrics("memory", "Max Memory Usage", max_usage)]
+        try:
+            usage = self._get_memory_usage()
+            max_usage = self._get_memory_max_usage()
+            return [CollectedMetrics("memory", "Total Memory Usage", usage),
+                    CollectedMetrics("memory", "Max Memory Usage", max_usage)]
+        except CGroupsException as e:
+            raise e
```
narrieta: not needed

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -34,15 +35,24 @@ def _get_metrics_list(metric):
...
-                "max_mem": CGroupsTelemetry._get_metrics_list(max_memory_levels)
-            },
-            "cpu": {
-                "cur_cpu": CGroupsTelemetry._get_metrics_list(current_cpu_usage)
-            }
-        }
+
+        processed_extension = {}
+
+        if current_cpu_usage.count() > 0:
```
narrieta: the consumer of this object will have to deal with the fact that sometimes some of these values will not be there... is that ok?

azurelinuxagent/common/cgroup.py
vrdmr: The method `_update_cpu_data` uses this function `_get_current_cpu_total` and thus the raise is needed. The `0` is set initially, but subsequent checks would need to have this check.

azurelinuxagent/common/cgroupstelemetry.py
vrdmr: The `cgroup_metrics.get_metrics()` call returns 3 objects of type `Metric`. The list:data inside the Metric would capture all the details of not being able to get data sometimes. If we were not able to capture any data, the length in that case would be 0. Thus the check.

azurelinuxagent/common/cgroup.py
```diff
@@ -86,23 +86,28 @@ def _get_parameters(self, parameter_name, first_line_only=False):
...
             parameter_filename = self._get_cgroup_file(parameter_name)
             logger.error("File {0} is empty but should not be".format(parameter_filename))
+            raise CGroupsException("File {0} is empty but should not be".format(parameter_filename))
         except CGroupsException:
-            return None
+            raise
         except Exception as e:
             parameter_filename = self._get_cgroup_file(parameter_name)
             logger.error("Exception while attempting to read {0}: {1}".format(parameter_filename, ustr(e)))
+            raise CGroupsException("Exception while attempting to read {0}: {1}".format(parameter_filename, ustr(e)))
```
narrieta: this loses the traceback of the original exception

azurelinuxagent/common/cgroup.py
```diff
@@ -86,23 +86,28 @@ def _get_parameters(self, parameter_name, first_line_only=False):
...
     def is_active(self):
-        tasks = self._get_parameters("tasks")
-        if tasks:
-            return len(tasks) != 0
-        else:
+        try:
+            tasks = self._get_parameters("tasks")
+            if tasks:
+                return len(tasks) != 0
+        except CGroupsException:
```
narrieta: this will mask exceptions; should probably log the error

azurelinuxagent/common/cgroup.py
```diff
@@ -115,7 +120,12 @@ def __init__(self, name, cgroup_path):
         super(CpuCgroup, self).__init__(name, cgroup_path, "cpu")
 
         self._osutil = get_osutil()
-        self._current_cpu_total = self._get_current_cpu_total()
+        self._current_cpu_total = 0
+        try:
+            self._current_cpu_total = self._get_current_cpu_total()
+        except:
+            pass
```
narrieta: we lose the exception... log?

azurelinuxagent/common/cgroup.py
narrieta: OK - you don't need the except/raise here

azurelinuxagent/common/cgroupstelemetry.py
narrieta: let's sync offline, please ping me

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -96,9 +105,21 @@ def report_all_tracked():
...
                 if cgroup.name not in CGroupsTelemetry._cgroup_metrics:
                     CGroupsTelemetry._cgroup_metrics[cgroup.name] = CgroupMetrics()
-                CGroupsTelemetry._cgroup_metrics[cgroup.name].add_new_data(cgroup.controller, cgroup.collect())
+
+                metric = None
+                # noinspection PyBroadException
+                try:
+                    metric = cgroup.collect()
+                except Exception:
+                    # Wanted to have other smaller granular exceptions but the recurring nature of this call could
```
narrieta: > could  cause overfill the logs

but makes harder to debug issues when we are not collecting data

tests/common/test_cgroupconfigurator.py
```diff
@@ -183,6 +183,9 @@ def test_start_extension_command_should_start_tracking_the_extension_cgroups(sel
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE)
 
+        for i in CGroupsTelemetry._tracked:
```
narrieta: leftover code?

azurelinuxagent/common/cgroupstelemetry.py
pgombar: Agreed, don't we still want to log the failure to collect metrics? It doesn't seem these failures would be transient, so we'd like to know if/why we fail.

azurelinuxagent/common/osutil/default.py
```diff
@@ -1270,6 +1270,7 @@ def _get_proc_stat():
             results = fileutil.read_file('/proc/stat')
         except (OSError, IOError) as ex:
             logger.warn("Couldn't read /proc/stat: {0}".format(ex.strerror))
+            raise ex
```
pgombar: Since you're raising here, this bubbles up to `get_total_cpu_ticks_since_boot`, which doesn't catch, which bubbles up to `__init__` of CpuCgroup (not caught), and to `_update_cpu_data` (not caught). Is this intended?

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -90,15 +100,29 @@ def report_all_tracked():
...
                 if cgroup.name not in CGroupsTelemetry._cgroup_metrics:
                     CGroupsTelemetry._cgroup_metrics[cgroup.name] = CgroupMetrics()
-                CGroupsTelemetry._cgroup_metrics[cgroup.name].add_new_data(cgroup.controller, cgroup.collect())
+
+                metric = None
+                # noinspection PyBroadException
+                try:
+                    metric = cgroup.collect()
+                except Exception as e:
+                    logger.periodic(logger.EVERY_HALF_HOUR,
```
narrieta: this should use the convenience functions (periodic_warn)

azurelinuxagent/common/logger.py
```diff
@@ -49,12 +49,30 @@ def is_period_elapsed(self, delta, h):
...
 
-    def periodic(self, delta, msg_format, *args):
+    def periodic_info(self, delta, msg_format, *args):
         h = hash(msg_format)
         if self.is_period_elapsed(delta, h):
             self.info(msg_format, *args)
             self.logger.periodic_messages[h] = datetime.now()
 
+    def periodic_verbose(self, delta, msg_format, *args):
+        h = hash(msg_format)
```
narrieta: can we refactor this code so that we do not have 4 copies of it?

azurelinuxagent/common/osutil/default.py
narrieta: "raise" instead of "raise ex" (to preserve the original traceback)

azurelinuxagent/common/cgroupstelemetry.py
vrdmr: The periodic_warn is a method of the Logger class, but we use the logger functions specified outside. This is why I pass the log level explicitly.

PR #1567, opened 2019-06-27
Reducing the erroneous download events to send only periodically

vrdmr:
## Description
Also, activating the skipped unittests.

Issue # 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1567)

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -39,7 +39,7 @@ import azurelinuxagent.common.utils.fileutil as fileutil
 import azurelinuxagent.common.version as version
 from azurelinuxagent.common.cgroupconfigurator import CGroupConfigurator
 from azurelinuxagent.common.errorstate import ErrorState, ERROR_STATE_DELTA_INSTALL
-from azurelinuxagent.common.event import add_event, WALAEventOperation, elapsed_milliseconds
+from azurelinuxagent.common.event import add_event, WALAEventOperation, elapsed_milliseconds, report_event
 from azurelinuxagent.common.exception import ExtensionError, ProtocolError, ProtocolNotFoundError, \
     ExtensionDownloadError, ExtensionOperationError, ExtensionErrorCodes
 from azurelinuxagent.common.future import ustr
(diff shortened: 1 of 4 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -441,7 +441,17 @@ def handle_ext_handler(self, ext_handler, etag):
     def handle_handle_ext_handler_error(self, ext_handler_i, e, code=-1):
         msg = ustr(e)
         ext_handler_i.set_handler_status(message=msg, code=code)
-        ext_handler_i.report_event(message=msg, is_success=False, log_event=True)
+
+        if isinstance(e, ExtensionDownloadError):
+            self.get_artifact_error_state.incr()
+            if self.get_artifact_error_state.is_triggered():
```
narrieta: ok to send the event periodically, but the error should still be logged, right?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -623,7 +633,7 @@ def decide_version(self, target_state=None):
         try:
             pkg_list = self.protocol.get_ext_handler_pkgs(self.ext_handler)
         except ProtocolError as e:
-            raise ExtensionError("Failed to get ext handler pkgs", e)
+            raise ExtensionDownloadError("Failed to get ext handler pkgs", e)
```
narrieta: get_ext_handler_pkgs() may raise ProtocolError for reasons other than download errors. I think you want to make the conversion to DownloadError deeper in the code, probably somewhere in the call to or implementation of get_ext_manifest

azurelinuxagent/ga/exthandlers.py
vrdmr: For `report_event` function, the default value for param `log_event` is True. Which in this case (as `is_success` is False) would log the line as well. I'll make it explicit.

azurelinuxagent/ga/exthandlers.py
vrdmr: Thanks for the tip. Went through the call tree. Will take another look to narrow down the exception causing areas.

PR #1580, opened 2019-07-12
Parametrize agent service name and fix distro resolution on Travis

pgombar:
## Description

This PR brings in two changes: 1) parametrizing the agent service name, instead of hard-coding it since it depends on the distro and 2) fixing distro resolution in Travis.

For 2), function `platform.get_linux_distribution()` fails when running on Travis. We only specify xenial and trusty as Travis configurations, but they resolve to `debian` with version being `jessie/sid` or `stretch/sid` which is incorrect and breaks the `LooseVersion` library we call to resolve the distro version. This change introduces treating Travis environment differently to avoid this case.

Added unit tests for the changes.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1580)

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -104,6 +104,7 @@ class DefaultOSUtil(object):
         self.selinux = None
         self.disable_route_warning = False
         self.jit_enabled = False
+        self.service_name = self.get_service_name()
 
     @staticmethod
     def get_service_name():
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -105,6 +105,10 @@ class DefaultOSUtil(object):
         self.disable_route_warning = False
         self.jit_enabled = False
 
+    @staticmethod
+    def get_service_name():
+        return "waagent"
+
     def get_firewall_dropped_packets(self, dst_ip=None):
         # If a previous attempt failed, do not retry
         global _enable_firewall
--- a/azurelinuxagent/common/osutil/openbsd.py
+++ b/azurelinuxagent/common/osutil/openbsd.py
@@ -35,12 +35,14 @@ UUID_PATTERN = re.compile(
     r'^\s*[A-F0-9]{8}(?:\-[A-F0-9]{4}){3}\-[A-F0-9]{12}\s*$',
     re.IGNORECASE)
 
+
 class OpenBSDOSUtil(DefaultOSUtil):
 
     def __init__(self):
         super(OpenBSDOSUtil, self).__init__()
         self.jit_enabled = True
         self._scsi_disks_timeout_set = False
+        self.service_name = self.get_service_name()
 
     def get_instance_id(self):
         ret, output = shellutil.run_get_output("sysctl -n hw.uuid")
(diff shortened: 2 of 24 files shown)


azurelinuxagent/common/osutil/openbsd.py
```diff
@@ -35,12 +35,14 @@
...
     re.IGNORECASE)
 
+
 class OpenBSDOSUtil(DefaultOSUtil):
 
     def __init__(self):
         super(OpenBSDOSUtil, self).__init__()
         self.jit_enabled = True
         self._scsi_disks_timeout_set = False
+        self.service_name = self.get_service_name()
```
vrdmr: You don't need to define self.service_name here once it is defined in DefaultOSUtil.

azurelinuxagent/common/osutil/default.py
```diff
@@ -105,6 +105,10 @@ def __init__(self):
         self.disable_route_warning = False
         self.jit_enabled = False
```
vrdmr: It'll be better to defile service_name here, rather than defining in all the child classes.
```suggestion
        self.service_name = self.get_service_name()
```

azurelinuxagent/common/osutil/openwrt.py
```diff
@@ -33,6 +33,7 @@ def __init__(self):
         self.dhclient_name = 'udhcpc'
         self.ip_command_output = re.compile('^\d+:\s+(\w+):\s+(.*)$')
         self.jit_enabled = True
+        self.service_name = self.get_service_name()
```
vrdmr: same as above.

azurelinuxagent/common/osutil/ubuntu.py
```diff
@@ -29,15 +29,20 @@ class Ubuntu14OSUtil(DefaultOSUtil):
     def __init__(self):
         super(Ubuntu14OSUtil, self).__init__()
         self.jit_enabled = True
+        self.service_name = self.get_service_name()
+
+    @staticmethod
+    def get_service_name():
+        return "walinuxagent"
```
vrdmr: Just a food for thought - Can we define the Service Names (`waagent`, `walinuxagent`) in a place as Constants (at the defaultOSUtil) so that these loose strings are not present?

Similar constants can be used in the test as well - for cleanliness.

vrdmr: Similar for the service commands, but thats for later.

tests/common/osutil/test_factory.py
```diff
@@ -0,0 +1,242 @@
...
+    @patch("azurelinuxagent.common.logger.warn")
+    def test_get_osutil_it_should_return_default(self, patch_logger):
+        ret = get_osutil(distro_name="",
+                         distro_code_name="",
+                         distro_version="",
+                         distro_full_name="")
+        self.assertTrue(type(ret) == DefaultOSUtil)
+        self.assertEquals(patch_logger.call_count, 1)
+
+    def test_get_osutil_it_should_return_ubuntu(self):
```
vrdmr: Can you move the Distro specific tests to its own classes? Keeping them in `DefaultOsUtilTestCase` is not the correct place.

pgombar: Misnomer, updated the name to reflect the scope of the tests.

azurelinuxagent/common/osutil/openbsd.py
pgombar: Good point, thanks.

pgombar: Actually, I take that back, the `self.service_name` doesn't resolve if it's not defined in the current class constructor.

azurelinuxagent/common/osutil/ubuntu.py
pgombar: I agree, but that cleanup should also involve taking the same constants out of `setup.py`, so I'd rather do it separately.

pgombar:
I've updated the PR by extracting the logic for Travis resolution from production code to test code.

PR #1592, opened 2019-07-26
enable sudo tests in Travis

pgombar:
## Description
Enable running tests in Travis that require sudo access (mostly cgroups-related tests).
The tests using the FileSystemCgroupsAPI will be run on Trusty and skipped on Xenial, and vice-versa for the SystemdCgroupsAPI.

Since we ran the tests that require sudo access for the first time in Travis, two tests were uncovered as failing:
1) `test_ext_handler_and_monitor_handler_with_cgroup_enabled` was using `str`, changed to `ustr` to support both Python 2 and 3.
2) `test_collect_and_send_events_cannot_read_events` assumed non-root execution and an "Access denied" exception thrown, mocked this exception instead.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1592)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ os: linux
 dist: xenial
 language: python
 env:
-  - NOSEOPTS="" SETUPOPTS=""
+  - NOSEOPTS="--verbose" SETUPOPTS=""
   # Add SETUPOPTS="check flake8" to enable flake8 checks
 
 matrix:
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -17,26 +17,26 @@ matrix:
...
-  - pip install -r requirements.txt
-  - pip install -r test-requirements.txt
+  - sudo pip install -r requirements.txt
+  - sudo pip install -r test-requirements.txt
 
 script:
   # future: - pylint setup.py makepkg.py azurelinuxagent/
-  - nosetests $NOSEOPTS tests
-  - if [ ! -z "$SETUPOPTS" ]; then /usr/bin/env python setup.py $SETUPOPTS; fi
+  - sudo env "PATH=$PATH" nosetests $NOSEOPTS tests
```
narrieta: does this run *all* unit tests as root?

.travis.yml
```diff
@@ -17,26 +17,26 @@ matrix:
...
+  - if [ ! -z "$SETUPOPTS" ]; then sudo /usr/bin/env python setup.py $SETUPOPTS; fi
```
narrieta: question? why is sudo needed here? was this broken before?

tests/common/test_cgroupapi.py
```diff
@@ -167,6 +163,7 @@ def controller_operation(controller):
             self.assertIn('Error in cgroup controller "cpu": A test exception.', message)
 
 
+@skip_if_predicate_true(is_systemd_present, "FileSystem cgroups API doesn't manage cgroups on systems using systemd.")
```
narrieta: why skip those tests?

tests/common/test_cgroupstelemetry.py
```diff
@@ -619,9 +620,8 @@ def test_extension_temetry_not_sent_for_empty_perf_metrics(self, *args):
 
     @skip_if_predicate_false(i_am_root, "This test will only run as root")
     @skip_if_predicate_false(are_cgroups_enabled, "Does not run when Cgroups are not enabled")
+    @skip_if_predicate_true(is_trusty_in_travis, "Does not run on Trusty in Travis")
```
narrieta: why this test does not run on Trusty in Travis?

.travis.yml
pgombar: Yes. We currently don't have tests that need to be run as a normal user in order to succeed.

.travis.yml
pgombar: `sudo` starts a new session which is outside of the virtual environment that was already set up by Travis. By using `env "PATH=$PATH"` we ensure we are referencing the same packages as if we were not sudo.

tests/common/test_cgroupapi.py
pgombar: They could run and succeed on a systemd machine, but since the systemd tests don't run on a filesystem machine, I wanted to make a clear cut between the two environments and their cgroups tests.

tests/common/test_cgroupstelemetry.py
pgombar: I added a comment on that function explaining why: https://github.com/Azure/WALinuxAgent/pull/1592/files/6b2db46027037dad5e2095aac8cc6a6716d19448?file-filters%5B%5D=.py#diff-ed85166e54c2abaa1e8a2fcc867ad749R144

PR #1594, opened 2019-07-29
Enable selective sudo in Travis

pgombar:
## Description

Remove global sudo for all tests introduced in #1592 and run it selectively on tests that have a particular attribute.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1594)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,7 @@ install:
 script:
   # future: - pylint setup.py makepkg.py azurelinuxagent/
   - nosetests $NOSEOPTS tests
+  - sudo env "PATH=$PATH" nosetests $NOSEOPTS tests.common.test_cgroupapi.SystemdCgroupsApiTestCase
   - if [ ! -z "$SETUPOPTS" ]; then /usr/bin/env python setup.py $SETUPOPTS; fi
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,13 +30,13 @@ matrix:
           SETUPOPTS=""
 
 install:
-  - sudo pip install -r requirements.txt
-  - sudo pip install -r test-requirements.txt
+  - pip install -r requirements.txt
+  - pip install -r test-requirements.txt
 
 script:
   # future: - pylint setup.py makepkg.py azurelinuxagent/
-  - sudo env "PATH=$PATH" nosetests $NOSEOPTS tests
-  - if [ ! -z "$SETUPOPTS" ]; then sudo /usr/bin/env python setup.py $SETUPOPTS; fi
+  - nosetests $NOSEOPTS tests
+  - if [ ! -z "$SETUPOPTS" ]; then /usr/bin/env python setup.py $SETUPOPTS; fi
 
 after_success:
   - if [[ $TRAVIS_PYTHON_VERSION == 3.7 ]]; then codecov; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,8 +35,8 @@ install:
 
 script:
   # future: - pylint setup.py makepkg.py azurelinuxagent/
-  - nosetests $NOSEOPTS tests
-  - sudo env "PATH=$PATH" nosetests $NOSEOPTS --verbose -a permissions=root tests
+  - nosetests $NOSEOPTS --attr '!requires_sudo' tests
+  - sudo env "PATH=$PATH" nosetests $NOSEOPTS --verbose --attr 'requires_sudo' tests
   - if [ ! -z "$SETUPOPTS" ]; then /usr/bin/env python setup.py $SETUPOPTS; fi
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ install:
 script:
   # future: - pylint setup.py makepkg.py azurelinuxagent/
   - nosetests $NOSEOPTS tests
-  - sudo env "PATH=$PATH" nosetests $NOSEOPTS --testmatch="(?:\b|_)[Tt]est_requires_root"
+  - sudo env "PATH=$PATH" nosetests $NOSEOPTS --verbose -a permissions=root tests
   - if [ ! -z "$SETUPOPTS" ]; then /usr/bin/env python setup.py $SETUPOPTS; fi
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ install:
 script:
   # future: - pylint setup.py makepkg.py azurelinuxagent/
   - nosetests $NOSEOPTS tests
-  - sudo env "PATH=$PATH" nosetests $NOSEOPTS tests.common.test_cgroupapi.SystemdCgroupsApiTestCase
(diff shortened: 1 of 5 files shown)


tests/common/test_cgroupapi.py
```diff
@@ -303,7 +303,8 @@ def test_it_should_return_extension_slice_name(self):
         self.assertEqual(extension_slice_name, "system-walinuxagent.extensions-Microsoft.Azure.DummyExtension_1.0.slice")
 
     @skip_if_predicate_false(i_am_root, "Test does not run when normal user")
-    def test_if_extensions_root_slice_is_created(self):
+    @attr(permissions='root')
```
narrieta: suggestion: I'd made this a bool instead (e.g. @attr(requires_root_privileges)); that makes test selection a little easier

narrieta: Let's remove the skip on i_am_root, or convert it to an assertion within the test. This way people running the tests on their dev machines are notified that the test needs to run as root (instead of simply skipping it, which may go unnoticed)

.travis.yml
```diff
@@ -30,13 +30,14 @@ matrix:
...
-  - sudo pip install -r test-requirements.txt
+  - pip install -r requirements.txt
+  - pip install -r test-requirements.txt
 
 script:
   # future: - pylint setup.py makepkg.py azurelinuxagent/
-  - sudo env "PATH=$PATH" nosetests $NOSEOPTS tests
-  - if [ ! -z "$SETUPOPTS" ]; then sudo /usr/bin/env python setup.py $SETUPOPTS; fi
+  - nosetests $NOSEOPTS tests
+  - sudo env "PATH=$PATH" nosetests $NOSEOPTS --verbose -a permissions=root tests
```
narrieta: How does Travis detect test failures? If any of the commands in the "script:" fails will it report a failure?

pgombar: Yes, it outputs the exit code of each line in the 'script' section, e.g.:
`The command "sudo env "PATH=$PATH" nosetests $NOSEOPTS --verbose -a permissions=root tests" exited with 0.`

tests/common/test_cgroupapi.py
pgombar: When running the test in PyCharm or from command line, if the test was skipped, it will be reported as skipped.

If we remove the decorator (and add the assertion within the test), those tests will fail when invoked as a normal user. So, if we don't want to see failures in Travis, we'd have to modify the nosetest call to exclude the tests with the root attribute. The same applies when running the tests locally, which seems like an unnecessary step each time we run tests.

I'm in favour of keeping the decorator because it's very clear which tests are skipped instead of having those tests just fail and potentially add confusion as to why they are failing.

pgombar: Sure, sounds good.

.travis.yml
narrieta: ok

larohra:
___
*[.travis.yml, line 39 at r1](https://reviewable.io/reviews/azure/walinuxagent/1594#-Ll3FnH41ZfCsdRlcyly:-Ll3FnH41ZfCsdRlcylz:bbs5z0h) ([raw file](https://github.com/azure/walinuxagent/blob/70c2083ddf9ea541faa37665a104c1f347a225b0/.travis.yml#L39)):*
> ```
> --verbose -a permissions=root
> ```
So would this always run the script as root? Or would it check permissions? How does this work?

larohra:
___
*[tests/common/test_cgroupapi.py, line 306 at r1](https://reviewable.io/reviews/azure/walinuxagent/1594#-Lkzb-qMKeBafg0kCohH-r1-306:-Ll3GCxt2HwEFupB---S:bo58cp7) ([raw file](https://github.com/azure/walinuxagent/blob/70c2083ddf9ea541faa37665a104c1f347a225b0/tests/common/test_cgroupapi.py#L306)):*
<details><summary><i>Previously, pgombar (Paula Gombar) wrote…</i></summary><blockquote>

Sure, sounds good.
</blockquote></details>

Agree to Norberto's suggestion on making it a bool, else everything LGTM

.travis.yml
pgombar: Discussed offline, the 'script' lines are executed sequentially and are not dependent on each other.

tests/common/test_cgroupapi.py
pgombar: Discussed offline, in favour of making failures explicit instead of skipped, adding the assertion if root within the tests and changing the invocation for non-root and root tests in the .yml file.

pgombar:
___
*[.travis.yml, line 39 at r1](https://reviewable.io/reviews/azure/walinuxagent/1594#-Ll3FnH41ZfCsdRlcyly:-Ll3NLBG-om-rN_ZKWKm:by6hsdp) ([raw file](https://github.com/azure/walinuxagent/blob/70c2083ddf9ea541faa37665a104c1f347a225b0/.travis.yml#L39)):*
<details><summary><i>Previously, larohra wrote…</i></summary><blockquote>

> ```
> --verbose -a permissions=root
> ```
So would this always run the script as root? Or would it check permissions? How does this work?
</blockquote></details>

`-a permissions=root` discovers tests that have the attribute "permission=root" set and only executes those tests with sudo privileges.

larohra:
___
*[.travis.yml, line 39 at r1](https://reviewable.io/reviews/azure/walinuxagent/1594#-Ll3FnH41ZfCsdRlcyly:-Ll3haoO9YTX6XfZ6F-H:b6k139u) ([raw file](https://github.com/azure/walinuxagent/blob/70c2083ddf9ea541faa37665a104c1f347a225b0/.travis.yml#L39)):*
<details><summary><i>Previously, pgombar (Paula Gombar) wrote…</i></summary><blockquote>

`-a permissions=root` discovers tests that have the attribute "permission=root" set and only executes those tests with sudo privileges.
</blockquote></details>

awesome thanks for the info!

tests/common/test_cgroupapi.py
```diff
@@ -410,8 +416,10 @@ def mock_popen(*args, **kwargs):
 
                     self.assert_cgroups_created(extension_cgroups)
 
-    @skip_if_predicate_false(i_am_root, "Test does not run when normal user")
+    @attr('requires_sudo')
```
narrieta: this test actually does no require root privileges

tests/common/test_cgroupapi.py
```diff
@@ -460,8 +468,10 @@ def mock_popen(*args, **kwargs):
                         # No cgroups should have been created
                         self.assertEquals(extension_cgroups, [])
 
-    @skip_if_predicate_false(i_am_root, "Test does not run when normal user")
+    @attr('requires_sudo')
```
narrieta: nor this one

tests/common/test_cgroupapi.py
pgombar: You're right, updated.

tests/common/test_cgroupapi.py
pgombar: Thanks, updated.

PR #1595, opened 2019-07-29
Clean up old daemon cgroups

pgombar:
## Description

Daemons from version 2.2.31 to 2.2.40 added their PID to a now-obsolete cgroup `WALinuxAgent/WALinuxAgent`. Add logic to move that PID to the new cgroup, `walinuxagent.service` for each controller to ensure we track both the daemon and extension handler process.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1595)

### DIFF
--- a/azurelinuxagent/common/cgroupapi.py
+++ b/azurelinuxagent/common/cgroupapi.py
@@ -16,6 +16,7 @@
 
 import errno
 import os
+import shutil
 import subprocess
 import time
 import uuid
@@ -34,6 +35,7 @@ CGROUP_CONTROLLERS = ["cpu", "memory"]
 VM_AGENT_CGROUP_NAME = "walinuxagent.service"
 EXTENSIONS_ROOT_CGROUP_NAME = "walinuxagent.extensions"
 UNIT_FILES_FILE_SYSTEM_PATH = "/etc/systemd/system"
+DAEMON_PID_FILE = "/var/run/waagent.pid"
 
 
 class CGroupsApi(object):
@@ -56,6 +58,9 @@ class CGroupsApi(object):
         raise NotImplementedError()
 
     def start_extension_command(self, extension_name, command, shell, cwd, env, stdout, stderr):
+        raise NotImplementedError()
+
+    def cleanup_old_cgroups(self):
         raise NotImplementedError()
 
     @staticmethod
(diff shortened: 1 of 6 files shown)


azurelinuxagent/common/cgroupapi.py
```diff
@@ -34,6 +35,7 @@
 VM_AGENT_CGROUP_NAME = "walinuxagent.service"
 EXTENSIONS_ROOT_CGROUP_NAME = "walinuxagent.extensions"
 UNIT_FILES_FILE_SYSTEM_PATH = "/etc/systemd/system"
+DAEMON_PID_FILE = "/var/run/waagent.pid"
```
narrieta: the file can be changed in waagent.conf - let's use get_agent_pid_file_path instead of hardcoding the path

azurelinuxagent/daemon/main.py
```diff
@@ -69,6 +69,7 @@ def run(self, child_args=None):
         self.initialize_environment()
 
         CGroupConfigurator.get_instance().create_agent_cgroups(track_cgroups=False)
+        CGroupConfigurator.get_instance().cleanup_old_cgroups()
```
narrieta: since the cleanup is just temporary we should move the logic to create_agent_cgroups instead of calling a separate function

azurelinuxagent/common/cgroupapi.py
```diff
@@ -183,6 +188,27 @@ def _add_process_to_cgroup(pid, cgroup_path):
...
+        def cleanup_old_controller(controller):
+            old_path = os.path.join(CGROUPS_FILE_SYSTEM_ROOT, controller, "WALinuxAgent", "WALinuxAgent")
+            new_path = os.path.join(CGROUPS_FILE_SYSTEM_ROOT, controller, VM_AGENT_CGROUP_NAME)
+
+            if not os.path.isdir(old_path):
+                return
+
+            contents = fileutil.read_file(os.path.join(old_path, "cgroup.procs"))
+
+            if daemon_pid in contents:
```
narrieta: do we need this check?

azurelinuxagent/common/cgroupapi.py
```diff
@@ -183,6 +188,27 @@ def _add_process_to_cgroup(pid, cgroup_path):
...
+                fileutil.append_file(os.path.join(new_path, "cgroup.procs"), daemon_pid)
+
```
narrieta: need to remove the old cgroup directory

azurelinuxagent/common/cgroupapi.py
pgombar: Missed that function, thanks.

azurelinuxagent/common/cgroupapi.py
pgombar: Thanks, updated now.

azurelinuxagent/daemon/main.py
pgombar: I'm in favour of keeping it separate to maintain one responsibility per function and to make the testing easier. The removal logic stays the same wherever the call happens.

azurelinuxagent/common/cgroupapi.py
pgombar: I would keep it to make sure that we're only writing the daemon PID if it's not already there (by explicitly checking it's still in the old cgroup).

PR #1602, opened 2019-08-02
Poll for artifact blob

larohra:
## Description
This PR is to fix the bug where the agent doesn't poll for change in the artifact_blob.on_hold property and waits for a new goalstate to start extension processing.

Note: DCR testing passed (except for the known cgroup issue already fixed in develop)

Issue # 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1602)

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -246,8 +246,10 @@ class ExtHandlersHandler(object):
             logger.verbose(msg)
             # Log status report success on new config
             self.log_report = True
-            self.handle_ext_handlers(etag)
-            self.last_etag = etag
+
+            if self.extension_processing_allowed():
+                self.handle_ext_handlers(etag)
+                self.last_etag = etag
 
             self.report_ext_handlers_status()
             self.cleanup_outdated_handlers()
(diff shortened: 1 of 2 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -246,8 +246,10 @@ def run(self):
             logger.verbose(msg)
             # Log status report success on new config
             self.log_report = True
-            self.handle_ext_handlers(etag)
-            self.last_etag = etag
+
+            if self.extension_processing_allowed():
+                self.handle_ext_handlers(etag)
+                self.last_etag = etag
```
narrieta: previously we were always updating etag, now it is done only when extension processing is allowed. Is this intentional?

larohra: This is what was breaking the previous case, we were waiting for a new GS when we technically shouldn't have processed the previous one if the extension processing was on hold.

larohra: We should only process a GS (and update the etag) only when the extension processing is allowed

narrieta: got it, i'd rename it to "last_processed_etag" just to make it a little clearer (I was thinking this meant the last etag we received in a goal state). thanks!

Issue #1514, opened 2019-04-29
ERROR ExtHandler Exception processing goal state, giving up: ['-----BEGIN PUBLIC KEY-----<key>-----END PUBLIC KEY-----\n']


bobsun0917:
**ISSUE**
Image: Ubuntu 16.04 LTS
Waagent version: 2.2.32 | 2.2.40

waagent.log is filled with following messages,

INFO ExtHandler WireServer is not responding. Reset endpoint
INFO ExtHandler Protocol endpoint not found: WireProtocol, [ProtocolError] Exceeded max retry updating goal state
INFO ExtHandler Protocol endpoint not found: MetadataProtocol, [ProtocolError] 404 - GET: http://169.254.169.254/Microsoft.Compute/identity?api-version=2015-05-01-preview
INFO ExtHandler Retry detect protocols: retry=45
INFO ExtHandler WireServer endpoint is not found. Rerun dhcp handler
INFO ExtHandler Test for route to 168.63.129.16
INFO ExtHandler Route to 168.63.129.16 exists
INFO ExtHandler Wire server endpoint:168.63.129.16
INFO ExtHandler Fabric preferred wire protocol version:2015-04-05
INFO ExtHandler Wire protocol version:2012-11-30
INFO ExtHandler Server preferred version:2015-04-05
ERROR ExtHandler Exception processing goal state, giving up: ['-----BEGIN PUBLIC KEY-----<key>-----END PUBLIC KEY-----\n']

**TROUBLESHOOTING**
1) The issue persists in version 2.2.32 and 2.2.40
2) Removing walinuxagent and install again but error remains.
3) paping port 80 on 168.63.129.16 and 169.254.169.254 can succeed

johncrim:
I'm experiencing this exact same issue. This is a blocker for me/my company - we're unable to add new servers to our VM Scaleset. Any ideas for troubleshooting?

johncrim:
Upon further review, I'm getting a slightly different error message - logged every 10 seconds. Again, this is new, didn't happen in deploys a month ago:

```
2019/06/11 04:29:13.084036 INFO Daemon WireServer endpoint is not found. Rerun dhcp handler
2019/06/11 04:29:13.090200 INFO Daemon Test for route to 168.63.129.16
2019/06/11 04:29:13.095114 INFO Daemon Route to 168.63.129.16 exists
2019/06/11 04:29:13.099579 INFO Daemon Wire server endpoint:168.63.129.16
[...]

PR #1606, opened 2019-08-07
Ignore stderr from openssl commands when saving to a variable

pgombar:
## Description

When calling openssl commands, we aren't ignoring stderr and saving the output directly to a variable, which then causes dictionary key exceptions to be thrown, like in issue #1514.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1606)

### DIFF
--- a/azurelinuxagent/common/utils/cryptutil.py
+++ b/azurelinuxagent/common/utils/cryptutil.py
@@ -47,15 +47,13 @@ class CryptUtil(object):
                "-out {2}").format(self.openssl_cmd, prv_file, crt_file)
         rc = shellutil.run(cmd)
         if rc != 0:
-            logger.error("Failed to create {0} and {1} certificates".format(
-                prv_file, crt_file))
+            logger.error("Failed to create {0} and {1} certificates".format(prv_file, crt_file))
 
     def get_pubkey_from_prv(self, file_name):
         if not os.path.exists(file_name):
             raise IOError(errno.ENOENT, "File not found", file_name)
         else:
-            cmd = "{0} pkey -in {1} -pubout 2>/dev/null".format(self.openssl_cmd,
-                                                                file_name)
+            cmd = "{0} pkey -in {1} -pubout 2>/dev/null".format(self.openssl_cmd, file_name)
             pub = shellutil.run_get_output(cmd)[1]
             return pub
 
@@ -63,8 +61,7 @@ class CryptUtil(object):
         if not os.path.exists(file_name):
             raise IOError(errno.ENOENT, "File not found", file_name)
         else:
-            cmd = "{0} x509 -in {1} -pubkey -noout".format(self.openssl_cmd,
-                                                           file_name)
+            cmd = "{0} x509 -in {1} -pubkey -noout 2>/dev/null".format(self.openssl_cmd, file_name)
             pub = shellutil.run_get_output(cmd)[1]
             return pub
 
@@ -72,8 +69,7 @@ class CryptUtil(object):
         if not os.path.exists(file_name):
             raise IOError(errno.ENOENT, "File not found", file_name)
         else:
-            cmd = "{0} x509 -in {1} -fingerprint -noout".format(self.openssl_cmd,
-                                                                file_name)
+            cmd = "{0} x509 -in {1} -fingerprint -noout 2>/dev/null".format(self.openssl_cmd, file_name)
             thumbprint = shellutil.run_get_output(cmd)[1]
             thumbprint = thumbprint.rstrip().split('=')[1].replace(':', '').upper()
             return thumbprint
--- a/azurelinuxagent/common/utils/cryptutil.py
+++ b/azurelinuxagent/common/utils/cryptutil.py
@@ -53,24 +53,24 @@ class CryptUtil(object):
         if not os.path.exists(file_name):
             raise IOError(errno.ENOENT, "File not found", file_name)
         else:
-            cmd = "{0} pkey -in {1} -pubout 2>/dev/null".format(self.openssl_cmd, file_name)
-            pub = shellutil.run_get_output(cmd)[1]
+            cmd = "{0} pkey -in {1} -pubout".format(self.openssl_cmd, file_name)
+            pub = shellutil.run_command(cmd.split(" "))
             return pub
 
     def get_pubkey_from_crt(self, file_name):
         if not os.path.exists(file_name):
             raise IOError(errno.ENOENT, "File not found", file_name)
         else:
-            cmd = "{0} x509 -in {1} -pubkey -noout 2>/dev/null".format(self.openssl_cmd, file_name)
-            pub = shellutil.run_get_output(cmd)[1]
+            cmd = "{0} x509 -in {1} -pubkey -noout".format(self.openssl_cmd, file_name)
+            pub = shellutil.run_command(cmd.split(" "))
             return pub
 
     def get_thumbprint_from_crt(self, file_name):
         if not os.path.exists(file_name):
             raise IOError(errno.ENOENT, "File not found", file_name)
         else:
-            cmd = "{0} x509 -in {1} -fingerprint -noout 2>/dev/null".format(self.openssl_cmd, file_name)
-            thumbprint = shellutil.run_get_output(cmd)[1]
+            cmd = "{0} x509 -in {1} -fingerprint -noout".format(self.openssl_cmd, file_name)
+            thumbprint = shellutil.run_command(cmd.split(" "))
             thumbprint = thumbprint.rstrip().split('=')[1].replace(':', '').upper()
             return thumbprint
 
--- a/azurelinuxagent/common/utils/cryptutil.py
+++ b/azurelinuxagent/common/utils/cryptutil.py
@@ -20,7 +20,6 @@
 import base64
 import errno
 import struct
-import sys
 import os.path
 import subprocess
 
@@ -29,7 +28,7 @@ from azurelinuxagent.common.exception import CryptError
 
 import azurelinuxagent.common.logger as logger
 import azurelinuxagent.common.utils.shellutil as shellutil
-import azurelinuxagent.common.utils.textutil as textutil
+
 
 DECRYPT_SECRET_CMD = "{0} cms -decrypt -inform DER -inkey {1} -in /dev/stdin"
 
@@ -53,24 +52,24 @@ class CryptUtil(object):
         if not os.path.exists(file_name):
             raise IOError(errno.ENOENT, "File not found", file_name)
         else:
-            cmd = "{0} pkey -in {1} -pubout".format(self.openssl_cmd, file_name)
-            pub = shellutil.run_command(cmd.split(" "))
+            cmd = [self.openssl_cmd, "pkey", "-in", file_name, "-pubout"]
+            pub = shellutil.run_command(cmd)
             return pub
 
     def get_pubkey_from_crt(self, file_name):
         if not os.path.exists(file_name):
             raise IOError(errno.ENOENT, "File not found", file_name)
         else:
-            cmd = "{0} x509 -in {1} -pubkey -noout".format(self.openssl_cmd, file_name)
-            pub = shellutil.run_command(cmd.split(" "))
+            cmd = [self.openssl_cmd, "x509", "-in", file_name, "-pubkey", "-noout"]
+            pub = shellutil.run_command(cmd)
             return pub
 
     def get_thumbprint_from_crt(self, file_name):
         if not os.path.exists(file_name):
             raise IOError(errno.ENOENT, "File not found", file_name)
         else:
-            cmd = "{0} x509 -in {1} -fingerprint -noout".format(self.openssl_cmd, file_name)
-            thumbprint = shellutil.run_command(cmd.split(" "))
+            cmd = [self.openssl_cmd, "x509", "-in", file_name, "-fingerprint", "-noout"]
+            thumbprint = shellutil.run_command(cmd)
             thumbprint = thumbprint.rstrip().split('=')[1].replace(':', '').upper()
             return thumbprint
 
--- a/azurelinuxagent/common/utils/shellutil.py
+++ b/azurelinuxagent/common/utils/shellutil.py
@@ -79,6 +79,9 @@ def run_get_output(cmd, chk_err=True, log_cmd=True, expected_errors=[]):
     Execute 'cmd'.  Returns return code and STDOUT, trapping expected
     exceptions.
     Reports exceptions to Error if chk_err parameter is True
+
+    For new callers, consider using run_command instead as it separates stdout from stderr,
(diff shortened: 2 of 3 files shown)


azurelinuxagent/common/utils/cryptutil.py
```diff
@@ -47,33 +47,29 @@ def gen_transport_cert(self, prv_file, crt_file):
...
             pub = shellutil.run_get_output(cmd)[1]
             return pub
 
     def get_thumbprint_from_crt(self, file_name):
         if not os.path.exists(file_name):
             raise IOError(errno.ENOENT, "File not found", file_name)
         else:
-            cmd = "{0} x509 -in {1} -fingerprint -noout".format(self.openssl_cmd,
-                                                                file_name)
+            cmd = "{0} x509 -in {1} -fingerprint -noout 2>/dev/null".format(self.openssl_cmd, file_name)
```
narrieta: but when something fails, how will we know the error?

narrieta: we should add unit tests

pgombar: Before this PR, the three methods that issued openssl commands and stored the return value were `get_pubkey_from_prv`, `get_pubkey_from_crt` and `get_thumbprint_from_crt`. Only one of them, `get_pubkey_from_prv` was redirecting the stderr to /dev/null, the other two weren't, which caused two incidents so far. My change adds the redirection of stderr to the other two commands.

None of the three functions check the return code of the commands, so we would never know if openssl failed and why. However, in the two incidents, the command wasn't failing, we were just capturing unnecessary data instead of the final return value. For example, all of these warnings would get saved to the `thumbprint` variable:
```
/usr/bin/openssl x509 -in /var/lib/waagent/{cerificate_name}.crt -fingerprint -noout
/usr/bin/openssl: /usr/local/openssl/lib/libssl.so.1.0.0: no version information available (required by /usr/bin/openssl)
/usr/bin/openssl: /usr/local/openssl/lib/libssl.so.1.0.0: no version information available (required by /usr/bin/openssl)
/usr/bin/openssl: /usr/local/openssl/lib/libssl.so.1.0.0: no version information available (required by /usr/bin/openssl)
/usr/bin/openssl: /usr/local/openssl/lib/libcrypto.so.1.0.0: no version information available (required by /usr/bin/openssl)
/usr/bin/openssl: /usr/local/openssl/lib/libcrypto.so.1.0.0: no version information available (required by /usr/bin/openssl)
/usr/bin/openssl: /usr/local/openssl/lib/libcrypto.so.1.0.0: no version information available (required by /usr/bin/openssl)
/usr/bin/openssl: /usr/local/openssl/lib/libcrypto.so.1.0.0: no version information available (required by /usr/bin/openssl)
SHA1 Fingerprint={thumbprint_value}
```
By redirecting the stderr, we ensure we only capture the output of the command itself (in this case the last line).

As for unit tests, there are unit tests already for all three methods that check the correctness of the return values, not the openssl commands itself.

azurelinuxagent/common/utils/shellutil.py
```diff
@@ -110,6 +110,34 @@ def run_get_output(cmd, chk_err=True, log_cmd=True, expected_errors=[]):
...
+
+        if retcode:
+            logger.error(u"Command: [{0}], return code: [{1}], "
+                         u"stdout: [{2}] stderr: [{3}]".format(command,
+                                                               retcode,
+                                                               _encode_command_output(stdout),
+                                                               _encode_command_output(stderr)))
+            raise subprocess.CalledProcessError(retcode, command)
+    except Exception as e:
+        logger.error(u"Command [{0}] raised unexpected exception: [{1}]".format(command, ustr(e)))
```
narrieta: you are logging twice when the command returns != 0

azurelinuxagent/common/utils/cryptutil.py
```diff
@@ -47,34 +47,30 @@ def gen_transport_cert(self, prv_file, crt_file):
...
 
     def get_pubkey_from_prv(self, file_name):
         if not os.path.exists(file_name):
             raise IOError(errno.ENOENT, "File not found", file_name)
         else:
-            cmd = "{0} pkey -in {1} -pubout 2>/dev/null".format(self.openssl_cmd,
-                                                                file_name)
-            pub = shellutil.run_get_output(cmd)[1]
+            cmd = "{0} pkey -in {1} -pubout".format(self.openssl_cmd, file_name)
+            pub = shellutil.run_command(cmd.split(" "))
```
larohra: Any particular reason for not using `shlex.split()` instead of `cmd.split(" ")`?

azurelinuxagent/common/utils/shellutil.py
```diff
@@ -110,6 +110,34 @@ def run_get_output(cmd, chk_err=True, log_cmd=True, expected_errors=[]):
...
+        stdout, stderr = process.communicate()
+        retcode = process.returncode
+
+        if retcode:
+            logger.error(u"Command: [{0}], return code: [{1}], "
+                         u"stdout: [{2}] stderr: [{3}]".format(command,
+                                                               retcode,
+                                                               _encode_command_output(stdout),
+                                                               _encode_command_output(stderr)))
+            raise subprocess.CalledProcessError(retcode, command)
```
narrieta: Not sure we should raise this as a CalledProcessError... the documentation is specific about this exception being raised by cehck_call/check_output (vs. Popen:
> a exception subprocess.CalledProcessError
Exception raised when a process run by check_call() or check_output() returns a non-zero exit status.

Using this exception might not be a big issue is that the output and stderr properties would be populated, I think that is missing from your code.

Consider whether the caller actually needs the return code/stdout/stderr of the command and, if not, you might be OK just raising an Exception

azurelinuxagent/common/utils/cryptutil.py
pgombar: Just wasn't aware of it. Thanks, will use it!

azurelinuxagent/common/utils/shellutil.py
```diff
@@ -110,6 +110,34 @@ def run_get_output(cmd, chk_err=True, log_cmd=True, expected_errors=[]):
     return 0, output
 
 
+def _encode_command_output(output):
+    return ustr(output, encoding='utf-8', errors="backslashreplace")
+
+
+def run_command(command):
```
narrieta: let's add a comment on run_get_output to explain when/why to use this new function instead

azurelinuxagent/common/utils/cryptutil.py
```diff
@@ -47,34 +47,30 @@ def gen_transport_cert(self, prv_file, crt_file):
...
+            logger.error("Failed to create {0} and {1} certificates".format(prv_file, crt_file))
 
     def get_pubkey_from_prv(self, file_name):
         if not os.path.exists(file_name):
             raise IOError(errno.ENOENT, "File not found", file_name)
         else:
-            cmd = "{0} pkey -in {1} -pubout 2>/dev/null".format(self.openssl_cmd,
-                                                                file_name)
-            pub = shellutil.run_get_output(cmd)[1]
+            cmd = "{0} pkey -in {1} -pubout".format(self.openssl_cmd, file_name)
```
narrieta: let's double-check the caller does not depend on stderr being there

azurelinuxagent/common/utils/cryptutil.py
```diff
@@ -47,34 +47,30 @@ def gen_transport_cert(self, prv_file, crt_file):
...
 
     def get_pubkey_from_crt(self, file_name):
         if not os.path.exists(file_name):
             raise IOError(errno.ENOENT, "File not found", file_name)
         else:
-            cmd = "{0} x509 -in {1} -pubkey -noout".format(self.openssl_cmd,
-                                                           file_name)
-            pub = shellutil.run_get_output(cmd)[1]
+            cmd = "{0} x509 -in {1} -pubkey -noout".format(self.openssl_cmd, file_name)
+            pub = shellutil.run_command(cmd.split(" "))
```
narrieta: the split would break if, for example, file_name contains spaces. I would just use an array explicitly:

 [self.openssl_cmd, "x509", "-in", file_name, "-pubkey", "-noout"]

larohra: that's where shlex.split() helps, I think @pgombar already added that

azurelinuxagent/common/utils/shellutil.py
larohra: If we're redesigning run_get_output we should return all 3 - exit_code, out, err and let it be for the caller to decide what to use

azurelinuxagent/common/utils/cryptutil.py
pgombar: As discussed offline, changing the underlying method for invoking commands to capture stderr in case of failures. As this is done by separating stdout and stderr streams, it effectively means the return value will only consists of stdout (without stderr), which is what was broken before.

azurelinuxagent/common/utils/cryptutil.py
pgombar: It doesn't.

azurelinuxagent/common/utils/shellutil.py
pgombar: Done.

azurelinuxagent/common/utils/shellutil.py
pgombar: Removed the raise of CalledProcessError since it's not needed as long as we log on error and raise an exception on failure. Plus, `stderr` was only introduced to the CalledProcessError constructor in Python 3 so trying to avoid supporting both.

@larohra I'm deciding to keep it as returning stdout only for this iteration since we don't consume exit_code and stderr. We will change the design as needed when we make the final switch from using `run_get_output` to `run_command`.

azurelinuxagent/common/utils/shellutil.py
pgombar: Removed the raise from L133, not the case anymore.

larohra:
___
*[azurelinuxagent/common/utils/cryptutil.py, line 65 at r2](https://reviewable.io/reviews/azure/walinuxagent/1606#-LlhRk5qnPyE8F5QNHA3-r2-65:-LlnUc0S5EXjWl_dOKCs:b-56e7rj) ([raw file](https://github.com/azure/walinuxagent/blob/9ebc2e7f36deb9f52869e0fea993e2bde0c444ef/azurelinuxagent/common/utils/cryptutil.py#L65)):*
> ```
> {0} x509 -in {1} -pubkey -noout
> ```
Ahh makes sense. 
What if you escape the filename with quotes instead? Like - [{0} x509 -in '{1}' -pubkey -noout]
This would be handled by shlex

larohra:
___
*[azurelinuxagent/common/utils/cryptutil.py, line 65 at r2](https://reviewable.io/reviews/azure/walinuxagent/1606#-LlhRk5qnPyE8F5QNHA3-r2-65:-LlnWXYm7ugryzfzt-7D:b-569fc6) ([raw file](https://github.com/azure/walinuxagent/blob/9ebc2e7f36deb9f52869e0fea993e2bde0c444ef/azurelinuxagent/common/utils/cryptutil.py#L65)):*
<details><summary><i>Previously, narrieta (Norberto Arrieta) wrote…</i></summary><blockquote>

sure, but also quote '{0}' and that would work for now... next person changing the code: please think hard if you need to escape anything :)

or just write the argument array and forget about shell parsing
</blockquote></details>

Haha agreed, escaping can always be a pain in the butt!

azurelinuxagent/common/utils/shellutil.py
```diff
@@ -110,6 +113,37 @@ def run_get_output(cmd, chk_err=True, log_cmd=True, expected_errors=[]):
...
+    """
+    Wrapper for subprocess.Popen. Executes the given command and returns its stdout.
+    The command parameter is a list of strings, e.g. ["ps" "aux"].
```
narrieta: I'd remove this: 

> The command parameter is a list of strings

It could also be a string (e.g. a command with no arguments)

azurelinuxagent/common/utils/shellutil.py
```diff
@@ -110,6 +113,37 @@ def run_get_output(cmd, chk_err=True, log_cmd=True, expected_errors=[]):
...
+    Logs any errors executing the command and raises an exception.
+    """
+    retcode = 0
+
+    try:
+        process = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=False)
+        stdout, stderr = process.communicate()
+        retcode = process.returncode
+    except Exception as e:
+        logger.error(u"Command [{0}] raised unexpected exception: [{1}]".format(command, ustr(e)))
```
narrieta: It is not the command who raised the exception (the command was not executed). Maybe something like "Cannot execute {0}. Error: {1}"?

tests/utils/test_shell_util.py
```diff
@@ -142,5 +139,43 @@ def test_it_should_log_unexpected_errors_as_errors(self):
...
+        command = ["ls", "nonexistent_file"]
+        expected_returncode = 2
+
+        with patch("azurelinuxagent.common.utils.shellutil.logger", autospec=True) as mock_logger:
+            with self.assertRaises(Exception):
+                shellutil.run_command(command)
+
+            self.assertEquals(mock_logger.error.call_count, 1)
+
+            logged_error_message = u"Command: [{0}], return code: [{1}]".format(command, expected_returncode)
```
narrieta: You should also check the message in the exception

tests/utils/test_shell_util.py
```diff
@@ -142,5 +139,43 @@ def test_it_should_log_unexpected_errors_as_errors(self):
...
+            self.assertIn(logged_error_message, mock_logger.error.call_args_list[0][0][0])
+
+    def test_run_command_it_should_log_and_raise_an_exception_from_subprocess_popen(self):
+        command = ["ls", "nonexistent_file"]
+
+        with patch("azurelinuxagent.common.utils.shellutil.logger", autospec=True) as mock_logger:
+            with patch("azurelinuxagent.common.utils.shellutil.subprocess.Popen",
```
narrieta: suggestion: instead of mocking Popen I'd just pass it a command that does not exist

azurelinuxagent/common/utils/shellutil.py
pgombar: Thanks, removed.

azurelinuxagent/common/utils/shellutil.py
pgombar: Updated.

tests/utils/test_shell_util.py
pgombar: Added.

tests/utils/test_shell_util.py
pgombar: Thanks, that's seems like a cleaner way to do it.

PR #1609, opened 2019-08-15
Expose disable error on update

larohra:
## Description
This PR is for handling the case where failure of disable.cmd in update scenario would now log from the old extension version and not the new one

Issue # 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1609)


tests/ga/test_extension.py
```diff
@@ -1327,6 +1327,44 @@ def test__extension_upgrade_failure_when_prev_version_disable_fails(self, patch_
...
+        test_data.ext_conf = test_data.ext_conf.replace('version="%s"' % old_version, 'version="%s"' % new_version)
+        test_data.manifest = test_data.manifest.replace(old_version, new_version)
+
+        with patch.object(ExtHandlerInstance, "report_event", autospec=True) as patch_report_event:
+            # Disable of the old extn fails
+            patch_get_disable_command.return_value = "exit 1"
+            exthandlers_handler.run()  # Download the new update the first time, and then we patch the download method.
+            self.assertEqual(1, patch_get_disable_command.call_count)
+
+            new_version_args, new_version_kwargs = patch_report_event.call_args
```
pgombar: Wouldn't `report_event` get called from the old version first, from the failed disable command?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -471,20 +472,44 @@ def handle_enable(self, ext_handler_i):
...
+            try:
+                if op == WALAEventOperation.Disable:
+                    ext_handler_i.disable()
+                elif op == WALAEventOperation.UnInstall:
+                    # Currently uninstall failures swallows the exception so we wont catch anything here
+                    ext_handler_i.uninstall()
+            except ExtensionError as e:
+                # Reporting the event with the old handler and marking log_event as false to avoid redundant reporting
+                ext_handler_i.report_event(message=ustr(e), is_success=False)
+                e.log_event = False
```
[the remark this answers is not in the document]
larohra: `ext_handler_i.report_event(message=ustr(e), is_success=False)` 
This line is now logging the event as part of the old handler. Earlier this same error was being logged with the new verison and as a Download error. 

`e.log_event = False`
This is so that we dont log the same error again with the new version and as a Download error

tests/ga/test_extension.py
larohra: yes you're correct, the report event gets called 3 times in this test scenario run. 
`patch_report_event.call_args` This property saves the last call only (`patch_report_event.call_args_list` saves all the times this method was called). In our case the last call to report_event gets called from the new version since the error is propagated back to the calling method. 

`new_version_args, new_version_kwargs = patch_report_event.call_args` 
This is getting the data for the last call. I'm using it to check that `log_event` is false for that one as we've already logged it earlier in the 2nd call to report_event

azurelinuxagent/common/exception.py
```diff
@@ -65,6 +65,8 @@ class ExtensionError(AgentError):
     def __init__(self, msg=None, inner=None, code=-1):
         super(ExtensionError, self).__init__(msg, inner)
         self.code = code
+        # Setting log_event to always log by default
+        self.log_event = True
```
narrieta: since now the exception always has the log_event attribute, this check on handle_ext_handler_error is not needed:
>     log_event = e.log_event if hasattr(e, "log_event") else True

azurelinuxagent/ga/exthandlers.py
```diff
@@ -443,7 +443,8 @@ def handle_ext_handler(self, ext_handler, etag):
     def handle_ext_handler_error(self, ext_handler_i, e, code=-1):
         msg = ustr(e)
         ext_handler_i.set_handler_status(message=msg, code=code)
-        ext_handler_i.report_event(message=msg, is_success=False, log_event=True)
+        log_event = e.log_event if hasattr(e, "log_event") else True
```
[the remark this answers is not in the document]
narrieta: but the check for "hasattr(e, "log_event")" is not needed... the attribute will always be there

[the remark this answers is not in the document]
larohra: `handle_ext_handler_error` this function is used by multiple Extension classes (ExtensionError, Exception, etc). I added the `log_event` only to ExtensionError class but if its a generic error then it wont have the log_event attribute in it

tests/ga/test_extension.py
```diff
@@ -1351,18 +1351,12 @@ def test__old_handler_reports_failure_on_disable_fail_on_update(self, patch_get_
...
 
-            new_version_args, new_version_kwargs = patch_report_event.call_args
-            self.assertFalse(new_version_kwargs['log_event'],
-                             "The final call to report_events should not log event")
-            self.assertEqual(new_version_args[0].ext_handler.properties.version, new_version,
-                             "The final report event call should be from the new version ext-handler "
-                             "as we want to report failure for the new version in the status blob")
-
-            old_version_args, old_version_kwargs = patch_report_event.call_args_list[1]
+            old_version_args, old_version_kwargs = patch_report_event.call_args
```
pgombar: Can we also assert that the `report_event` was only called once?

larohra: Its actually called twice - 
- Once to report that download of the new version was successful
- Second time to report the disable failure

I think asserting that the last call to report_event was done to report the disable failure is good enough as there should be no other calls after this in this run.

larohra: `patch_report_event.call_args` 
The `call_args` property of a patched object always picks up the last call parameters, so this is ensuring that we're only checking the last call

pgombar: In that case, can we explicitly assert the call args for both calls :)
You can use `call_args_list` to get all calls and their args.

larohra: Good point, I'll send out an update. Thanks :)

PR #1616, opened 2019-08-23
Logging a disable error instead of Download error on extension update…

larohra:
… failures due to disable cmd

## Description
This PR is for handling the case where failure of disable.cmd in update scenario would now log from the old extension version and not the new one

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -41,7 +41,7 @@ from azurelinuxagent.common.cgroupconfigurator import CGroupConfigurator
 from azurelinuxagent.common.errorstate import ErrorState, ERROR_STATE_DELTA_INSTALL
 from azurelinuxagent.common.event import add_event, WALAEventOperation, elapsed_milliseconds, report_event
 from azurelinuxagent.common.exception import ExtensionError, ProtocolError, ProtocolNotFoundError, \
-    ExtensionDownloadError, ExtensionOperationError, ExtensionErrorCodes
+    ExtensionDownloadError, ExtensionOperationError, ExtensionErrorCodes, ExtensionUpdateError
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.protocol import get_protocol_util
 from azurelinuxagent.common.protocol.restapi import ExtHandlerStatus, \
@@ -436,6 +436,9 @@ class ExtHandlersHandler(object):
             else:
                 message = u"Unknown ext handler state:{0}".format(state)
                 raise ExtensionError(message)
+        except ExtensionUpdateError as e:
+            # Only setting the handler status here as the error has already been reported from the old version
+            ext_handler_i.set_handler_status(message=ustr(e), code=e.code)
         except ExtensionOperationError as e:
             self.handle_ext_handler_error(ext_handler_i, e, e.code)
         except ExtensionDownloadError as e:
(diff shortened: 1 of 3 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -436,6 +436,9 @@ def handle_ext_handler(self, ext_handler, etag):
             else:
                 message = u"Unknown ext handler state:{0}".format(state)
                 raise ExtensionError(message)
+        except ExtensionUpdateError as e:
+            # Only setting the handler status here as the error has already been reported from the old version
+            ext_handler_i.set_handler_status(message=ustr(e), code=e.code)
```
narrieta: this works fine now, but if we ever add/change logic in handle_ext_handler_error (which currently is the central place to handle errors) we may miss this one... consider adding a flag to handle_ext_handler_error to do/not do the logging and call it from here as well.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -476,20 +479,45 @@ def handle_enable(self, ext_handler_i):
...
-                old_ext_handler_i.disable()
-                ext_handler_i.copy_status_files(old_ext_handler_i)
-                if ext_handler_i.version_gt(old_ext_handler_i):
-                    ext_handler_i.update()
-                else:
-                    old_ext_handler_i.update(version=ext_handler_i.ext_handler.properties.version)
-                old_ext_handler_i.uninstall()
-                old_ext_handler_i.remove_ext_handler()
-                ext_handler_i.update_with_install()
+                ExtHandlersHandler._update_extension_handler(old_ext_handler_i, ext_handler_i)
```
narrieta: Thanks for this small refactoring, makes the intention of the code clearer... just a question: why a static method?

azurelinuxagent/ga/exthandlers.py
larohra: Makes sense, I'll modify the code

azurelinuxagent/ga/exthandlers.py
larohra: Since I'm passing both the old_handler_i and new_handler_i there was no point making it a member function. The options were either make it a global function or make it a static member function of ExtHandlersHandler class. 

I'm open to any other design suggestion if you have an idea.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -476,20 +479,45 @@ def handle_enable(self, ext_handler_i):
...
+                msg = ustr(e)
+                ext_handler_i.report_event(message=msg, is_success=False)
+                raise ExtensionUpdateError(msg)
+
+        execute_old_handler_command(old_ext_handler_i, WALAEventOperation.Disable)
+        ext_handler_i.copy_status_files(old_ext_handler_i)
+        if ext_handler_i.version_gt(old_ext_handler_i):
+            ext_handler_i.update()
+        else:
+            old_ext_handler_i.update(version=ext_handler_i.ext_handler.properties.version)
```
narrieta: wow! we support extension downgrade?

larohra: Haha yes, this is at par with Windows. Not sure how much its been tested though.

larohra:
Successful DCR run for this PR - [Successful DCR Run](https://tuxgold.corp.microsoft.com/job/AzLinux/job/DungeonCrawler/job/DungeonCrawler.larohra2/170/)

azurelinuxagent/ga/exthandlers.py
```diff
@@ -476,20 +479,45 @@ def handle_enable(self, ext_handler_i):
...
+
+        def execute_old_handler_command(ext_handler_i, op):
+            """
+            Created a common function to execute all commands that need to be executed from the old handler
+            so that it can have a common exception handling mechanism
+            :param ext_handler_i: The extension handler instance to run the operation on
+            :param op: Which operation to run
+            """
+            try:
+                if op == WALAEventOperation.Disable:
```
narrieta: suggestion: if you pass a lambda then you don't need the switch on operation code

azurelinuxagent/ga/exthandlers.py
narrieta: no, this makes sense, thanks

azurelinuxagent/ga/exthandlers.py
narrieta: would you open a work item to test it and double-check whether the extension publishers doc talks about this? i'm sure some extensions publishers will be surprised (let's also check whether the VERSION env variable is documented). Thanks!

tests/ga/test_extension.py
```diff
@@ -1460,6 +1460,48 @@ def test__extension_upgrade_failure_when_prev_version_disable_fails_incorrect_zi
         # On the next iteration, update should not be retried
         self._assert_handler_status(protocol.report_vm_status, "NotReady", expected_ext_count=0, version="1.0.1")
 
+    @patch('azurelinuxagent.ga.exthandlers.HandlerManifest.get_disable_command')
```
narrieta: Thanks for the test, made the review a lot easier, too!

azurelinuxagent/ga/exthandlers.py
larohra: Added a [task](https://msazure.visualstudio.com/One/_workitems/edit/5209494). Will follow up later about this

azurelinuxagent/ga/exthandlers.py
larohra: Nice! Didnt know about it, will implement it!

PR #1622, opened 2019-08-27
udev: mark the SRIO-V functions as unamanged by NetworkManager

lkundrak:
## Description
From https://bugzilla.redhat.com/show_bug.cgi?id=1720157:

Accelerated Networking on Azure exposes a new SRIOV interface to the VM.
This interface is transparently bonded to the synthetic interface,
so NetworkManager should just ignore any SRIOV interfaces.

The rule is originally from Haiyang Zhang <haiyangz@microsoft.com>.

---

### PR information


larohra:
DCR Tests (End to end tests) passed for this change

narrieta:
@trstringer could you take a look? thanks!

lkundrak:
ping ping ping

lkundrak:
lala lala

trstringer:
@lkundrak what's the behavior of a VM _without_ this udev rule applied?

lkundrak:
Thanks for the response.

> @lkundrak deploying a VM with the `redhat:rhel:8:latest` image I see this rule in the image already.

Yes, but that's essentially a hack.

> You think this should apply to all distros?

Yes. Essentially all distros these days provide an option to use NetworkManager, even if they happen not to install it by default.

> What about those that don't use NetworkManager? This udev rule seems very NetworkManager-specific.

Most importantly, the rule is Azure-specific. Without NetworkManager, the rule is somewhat unnecessary, yet it can't do any harm. It would do harm though, if it was present outside Azure, which is why it can't be shipped with NetworkManager.

trstringer:
You bring up a good point. Ok let me do some testing and see what this looks like in other distros with other network services, and also we need to make sure this udev rule doesn't break any other networking configurations inadvertently. I'll test a few different scenarios and let you know what I find.

Thanks for bringing this to our attention!

trstringer:
So after much testing and investigation, I'm confident that this approach is a good one and I think we should move forward into getting this udev rule pushed into upstream. For documentation purposes, my findings include looking at how other distros and network services handle this.

For instance, with Debian Stretch (with backports) the reason that this isn't a problem is because it doesn't utilize NetworkManager, but uses `networking.service`, which behind the scenes does `ifup` and `ifdown` execs. This relies on `/etc/network/interfaces`, which includes similar configuration:

```
$ cat /etc/network/interfaces
# This file describes the network interfaces available on your system
# and how to activate them. For more information, see interfaces(5).

# The loopback network interface
auto lo
iface lo inet loopback

# The normal eth0
auto eth0
iface eth0 inet dhcp
 up sleep 5; dhclient -1 -6 -nw -cf /etc/dhcp/dhclient6.conf -lf /var/lib/dhcp/dhclient6.eth0.leases -v eth0 || true # ignore failure

# Maybe the VM has 2 NICs?
allow-hotplug eth1
iface eth1 inet dhcp

# Maybe the VM has 3 NICs?
allow-hotplug eth2
iface eth2 inet dhcp
```

So the reason why other non-NetworkManager implementations don't need to ignore/unmanage the SRIOV interface is because they are more explicit in which interfaces they will manage.

Finally, what I like about this improvement is that if a user decides to utilize NetworkManager in the future for a VM or image, this udev rule will already exist, providing a seamless transition.

@lkundrak can you rebase upstream develop branch on yours? Once this is done, I'll do a few things and then we can move forward.

lkundrak:
> So the reason why other non-NetworkManager implementations don't need to ignore/unmanage the SRIOV interface is because they are more explicit in which interfaces they will manage.

Yes. Note that NetworkManager can be configured to behave in the same way (not attempt to configure interfaces unless asked to do so), by installing the NetworkManager-config-server package. That is in general not recommended, because it needs more manual tinkering.

> @lkundrak can you rebase upstream develop branch on yours? Once this is done, I'll do a few things and then we can move forward.

I don't see why would I need to? The branch merges cleanly, and I'm not able to test the changes because my Azure account expired and I don't really have a time to spend on this any more (I'm now on a parental leave).

vrdmr:
>> @lkundrak can you rebase upstream develop branch on yours? Once this is done, I'll do a few things and then we can move forward.

>I don't see why would I need to? The branch merges cleanly, and I'm not able to test the changes because my Azure account expired and I don't really have a time to spend on this any more (I'm now on a parental leave).

Hi @lkundrak - Updated branch would help us test your changes against our E2E testing automation - so that the no old changes are tested (and throw us off if there are any issues).

If you want, I can do the rebase if you don't have time? Please let me know.

vittyvk:
> If you want, I can do the rebase if you don't have time? Please let me know.

Lubomir is currently on a long term leave, I can pick this up from Red Hat side. Please let me know if anything else from us is needed to make this merged. Thanks!

lkundrak:
> @lkundrak can you rebase upstream develop branch on yours?

I assume you meant rebase on develop. Done. Note that I haven't tested the results, but there were no conflicts.

vrdmr:
@trstringer Can you take a look? I'll also run couple of runs from our Test Automation.

trstringer:
@vrdmr Looks good to me.

config/68-azure-sriov-nm-unmanaged.rules
```diff
@@ -0,0 +1,4 @@
+# Accelerated Networking on Azure exposes a new SRIOV interface to the VM.
+# This interface is transparently bonded to the synthetic interface,
+# so NetworkManager should just ignore any SRIOV interfaces.
+SUBSYSTEM=="net", DRIVERS=="hv_pci", ACTION=="add", ENV{NM_UNMANAGED}="1"
```
dustymabe, line 4:
```suggestion
SUBSYSTEM=="net", DRIVERS=="hv_pci", ACTION=="add|change|move", ENV{NM_UNMANAGED}="1"
```

I was testing out a version of this over in https://github.com/coreos/fedora-coreos-config/pull/2176 and wasn't observing `NM_UNMANAGED=1` getting set in the initramfs until I added the `change|move`, which is what is done over in https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/38d3834e2c464bd4820392800a3a1fe3f286fc62/data/85-nm-unmanaged.rules#L5

dustymabe: @lkundrak WDYT?

cjp256: I would NOT recommend this change without a driver check (mlx4, mlx5).  There is potential for hv_pci devices to be managed.  We're working on a better story wrt MANA: https://learn.microsoft.com/en-us/azure/virtual-network/accelerated-networking-mana-linux

dustymabe: @cjp256 do you have an example modification to this rule to achieve that goal?

dustymabe:
Can we bring this PR back from the dead?

narrieta:
@anhvoms - could you take a look at this PR? It seems like trstringer did the initial review.

vittyvk:
Note that there's now a competing proposal in https://github.com/Azure/azure-vm-utils/pull/41/  where the rule is
```
SUBSYSTEM=="net", ACTION!="remove", DRIVERS=="mana|mlx4_core|mlx5_core", ATTR{flags}=="0x?[89ABCDEF]??", ENV{AZURE_UNMANAGED_SRIOV}="1", ENV{ID_NET_MANAGED_BY}="unmanaged", ENV{NM_UNMANAGED}="1"
```
and we already carry this rule downstream in RHEL[10,9].

dustymabe:
> Note that there's now a competing proposal in [Azure/azure-vm-utils#41](https://github.com/Azure/azure-vm-utils/pull/41) where the rule is

Carrying the rule in `azure-vm-utils` makes us happy. We have a [PR](https://github.com/coreos/fedora-coreos-config/pull/3378) to include the package in Fedora CoreOS and a request to include the package in RHEL so we can pick it up in RHCOS.

dustymabe:
We should close this since https://github.com/Azure/azure-vm-utils/pull/41 merged and that's the recommended place for these rules to be maintained.

PR #1623, opened 2019-08-27
Refactor start_extension_command to handle subprocess completion or timeout

pgombar:
## Description

Before, it was the responsibility of `ExtHandlerInstance`'s `launch_command` to wait for the subprocess to terminate or timeout, parse its output and raise exceptions if needed.

Since the actual invocation of the subprocess is done in the CGroup API layer, if `systemd-run` times out, we wouldn't know until the `ExtHandlerInstance` layer, at which point we wouldn't retry invoking the subprocess using `Popen` and we would erroneously raise an `ExtensionError` which would be counted against the extension's SLA.

This PR refactors the code so that these responsibilities are done in the `start_extension_command` in the CGroup API layer instead of in`launch_command`. Helper functions for waiting for process completion are added to `process_utils.py` and corresponding unit tests are added or updated.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1623)

### DIFF
--- a/azurelinuxagent/common/cgroupapi.py
+++ b/azurelinuxagent/common/cgroupapi.py
@@ -28,7 +28,7 @@ from azurelinuxagent.common.event import add_event, WALAEventOperation
 from azurelinuxagent.common.exception import CGroupsException, ExtensionError, ExtensionErrorCodes
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.utils import fileutil, shellutil
-from azurelinuxagent.common.utils.processutil import read_output, start_subprocess_and_wait_for_completion, \
+from azurelinuxagent.common.utils.extensionprocessutil import read_output, handle_process_completion, \
                                                      wait_for_process_completion_or_timeout
 from azurelinuxagent.common.version import AGENT_NAME, CURRENT_VERSION
 
@@ -348,16 +348,21 @@ class FileSystemCgroupsApi(CGroupsApi):
                 logger.warn("Failed to add extension {0} to its cgroup. Resource usage will not be tracked. "
                             "Error: {1}".format(extension_name, ustr(e)))
 
+        process = subprocess.Popen(command,
+                                   shell=shell,
+                                   cwd=cwd,
+                                   env=env,
+                                   stdout=stdout,
+                                   stderr=stderr,
+                                   preexec_fn=pre_exec_function)
+
         self.track_cgroups(extension_cgroups)
-        process_output = start_subprocess_and_wait_for_completion(command=command,
-                                                                  timeout=timeout,
-                                                                  shell=shell,
-                                                                  cwd=cwd,
-                                                                  env=env,
-                                                                  stdout=stdout,
-                                                                  stderr=stderr,
-                                                                  preexec_fn=pre_exec_function,
-                                                                  error_code=error_code)
+        process_output = handle_process_completion(process=process,
+                                                   command=command,
+                                                   timeout=timeout,
+                                                   stdout=stdout,
+                                                   stderr=stderr,
+                                                   error_code=error_code)
 
         return extension_cgroups, process_output
 
@@ -463,7 +468,7 @@ After=system-{1}.slice""".format(extension_name, EXTENSIONS_ROOT_CGROUP_NAME)
         return cgroups
 
     @staticmethod
-    def is_systemd_failure(scope_name, process_output):
+    def _is_systemd_failure(scope_name, process_output):
         unit_not_found = "Unit {0} not found.".format(scope_name)
         return unit_not_found in process_output or scope_name not in process_output
 
@@ -501,7 +506,7 @@ After=system-{1}.slice""".format(extension_name, EXTENSIONS_ROOT_CGROUP_NAME)
         else:
             # The extension didn't terminate successfully. Determine whether it was due to systemd errors or
             # extension errors.
-            systemd_failure = self.is_systemd_failure(scope_name, process_output)
+            systemd_failure = self._is_systemd_failure(scope_name, process_output)
 
             if not systemd_failure:
                 # There was an extension error; it either timed out or returned a non-zero exit code.
@@ -528,15 +533,20 @@ After=system-{1}.slice""".format(extension_name, EXTENSIONS_ROOT_CGROUP_NAME)
                 stderr.truncate(0)
 
                 # Try invoking the process again, this time without systemd-run
-                process_output = start_subprocess_and_wait_for_completion(command=command,
-                                                                          timeout=timeout,
-                                                                          shell=shell,
-                                                                          cwd=cwd,
-                                                                          env=env,
-                                                                          stdout=stdout,
-                                                                          stderr=stderr,
-                                                                          preexec_fn=os.setsid,
-                                                                          error_code=error_code)
+                process = subprocess.Popen(command,
+                                           shell=shell,
+                                           cwd=cwd,
+                                           env=env,
+                                           stdout=stdout,
+                                           stderr=stderr,
+                                           preexec_fn=os.setsid)
+
+                process_output = handle_process_completion(process=process,
+                                                           command=command,
+                                                           timeout=timeout,
+                                                           stdout=stdout,
+                                                           stderr=stderr,
+                                                           error_code=error_code)
 
                 return [], process_output
 
--- a/tests/common/test_cgroupapi.py
+++ b/tests/common/test_cgroupapi.py
@@ -313,16 +313,24 @@ class FileSystemCgroupsApiTestCase(AgentTestCase):
                     stdout=stdout,
                     stderr=stderr)
 
-        expected_output = "[stdout]\n{0}\n\n\n[stderr]\n"
+        # The expected format of the process output is [stdout]\n{PID}\n\n\n[stderr]\n"
+        pattern = re.compile(r"\[stdout\]\n(\d+)\n\n\n\[stderr\]\n")
+        m = pattern.match(process_output)
+
+        try:
+            pid_from_output = int(m.group(1))
+        except Exception as e:
+            self.fail("No PID could be extracted from the process output! Error: {0}".format(ustr(e)))
 
         for cgroup in extension_cgroups:
             cgroups_procs_path = os.path.join(cgroup.path, "cgroup.procs")
             with open(cgroups_procs_path, "r") as f:
                 contents = f.read()
-            pid = int(contents)
+            pid_from_cgroup = int(contents)
 
-            self.assertEquals(process_output, expected_output.format(pid),
-                              "The PID of the command was not added to {0}. Expected:\n{1}\ngot:\n{2}".format(cgroups_procs_path, expected_output.format(pid), process_output))
+            self.assertEquals(pid_from_output, pid_from_cgroup,
+                              "The PID from the process output ({0}) does not match the PID found in the"
+                              "process cgroup {1} ({2})".format(pid_from_output, cgroups_procs_path, pid_from_cgroup))
 
 
 @skip_if_predicate_false(is_systemd_present, "Systemd cgroups API doesn't manage cgroups on systems not using systemd.")
@@ -517,7 +525,7 @@ class SystemdCgroupsApiTestCase(AgentTestCase):
                 with patch("azurelinuxagent.common.cgroupapi.subprocess.Popen", side_effect=mock_popen):
                     with patch("azurelinuxagent.common.cgroupapi.wait_for_process_completion_or_timeout",
                                return_value=[True, None]):
-                        with patch("azurelinuxagent.common.cgroupapi.SystemdCgroupsApi.is_systemd_failure",
+                        with patch("azurelinuxagent.common.cgroupapi.SystemdCgroupsApi._is_systemd_failure",
                                    return_value=True):
                             extension_cgroups, process_output = SystemdCgroupsApi().start_extension_command(
                                 extension_name="Microsoft.Compute.TestExtension-1.2.3",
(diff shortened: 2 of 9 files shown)


azurelinuxagent/ga/env.py
```diff
@@ -80,7 +80,7 @@ def run(self):
         logger.info("Start env monitor service.")
         self.dhcp_handler.conf_routes()
         self.hostname = self.osutil.get_hostname_record()
-        self.dhcp_id_list = get_dhcp_client_pid()
+        self.dhcp_id_list = self.get_dhcp_client_pid()
```
larohra: Rebase develop to avoid unnecessary merge conflicts

azurelinuxagent/common/cgroupapi.py
```diff
@@ -455,52 +480,57 @@ def start_extension_command(self, extension_name, command, shell, cwd, env, stdo
...
-
-        return process, cgroups
+            extension_cgroups.append(CGroup.create(cgroup_path, controller, extension_name))
+
+        self._foreach_controller(create_cgroup, 'Cannot create cgroup for extension {0}; '
+                                                'resource usage will not be tracked.'.format(extension_name))
+        self.track_cgroups(extension_cgroups)
+
+        # Wait for process completion or timeout
+        timed_out, return_code = wait_for_process_completion_or_timeout(process, timeout)
```
narrieta: can you check the case when the extension completes with failure (i.e. timed_out == False, return_code != 0)

also, the series of if statements below are a little hard to follow? maybe we can re-arrange them somehow to make very obvious that all conditions are covered?

Thanks

pgombar: Any extension-related failures (timing out or returning != 0) are handled by the `start_subprocess_and_wait_for_completion` method in `processutil`. This is method handles a "normal" extension invocation (without systemd) and is used in several places: in the FileSystemCGroupAPI, in `exthandlers.py` when cgroups are disabled, and in the fallback option when systemd doesn't work.

pgombar: I stand corrected! The case when the extension fails when invoked by systemd was not covered. Sending an update now. Thanks for catching it.

azurelinuxagent/common/utils/processutil.py
```diff
@@ -17,11 +17,58 @@
...
 from azurelinuxagent.common.future import ustr
+import os
+import signal
+import subprocess
+import time
 
 TELEMETRY_MESSAGE_MAX_LEN = 3200
 
 
+def wait_for_process_completion_or_timeout(process, timeout):
```
narrieta: Since now this has become a utility function we should document its behavior and return value. For example, as opposed to the standard Python wait(timeout), this function will terminate the process when the timeout elapses.

azurelinuxagent/common/utils/processutil.py
```diff
@@ -17,11 +17,58 @@
...
+    while timeout > 0 and process.poll() is None:
+        time.sleep(1)
+        timeout -= 1
+
+    return_code = None
+
+    if timeout == 0:
+        os.killpg(os.getpgid(process.pid), signal.SIGKILL)
+    else:
+        # process completed or forked; sleep 1 sec to give the child process (if any) a chance to start
```
narrieta: expand this comment to explain why we sleep (previously this was in a private function used in a very specific context so it was easier to understand why)

tests/utils/test_process_util.py
```diff
@@ -32,6 +36,101 @@ def tearDown(self):
...
+            "sleep 4s",
+            shell=True,
+            cwd=self.tmp_dir,
+            env={},
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            preexec_fn=os.setsid)
+
+        with patch('azurelinuxagent.common.utils.processutil.os.killpg') as mock_kill:
+            timed_out, ret = wait_for_process_completion_or_timeout(process, 2)
```
narrieta: 2 seconds is too long for a unit test.

if you add some synchronization with the child process you may avoid the sleep altogether. For example, the child process creates a file and then sleeps for a minute or so.

the test waits for the file to be created and then it just uses an arbitrary timeout. You don't mock kill, but mock sleep. Then you assert that sleep was called the expected number of times (i.e. we waited for the timeout period before killing the process) and that the process was terminated (i.e. the process was killed after the timeout elapsed).

tests/utils/test_process_util.py
```diff
@@ -32,6 +36,101 @@ def tearDown(self):
...
+
+            self.assertEquals(mock_kill.call_count, 1)
+            self.assertEquals(timed_out, True)
+            self.assertEquals(ret, None)
+
+    def test_wait_for_process_completion_or_timeout_should_return_nonzero_return_code(self):
```
narrieta: "should_return_nonzero_when_the_process_fails"

tests/utils/test_process_util.py
```diff
@@ -32,6 +36,101 @@ def tearDown(self):
...
+                process_output = start_subprocess_and_wait_for_completion(command=command,
+                                                                          timeout=5,
+                                                                          shell=True,
+                                                                          cwd=self.tmp_dir,
+                                                                          env={},
+                                                                          stdout=stdout,
+                                                                          stderr=stderr,
+                                                                          preexec_fn=os.setsid,
+                                                                          error_code=42)
+        expected_output = "[stdout]\ndummy output\n\n\n[stderr]\n"
```
narrieta: let's use a test command that produces some stderr

tests/utils/test_process_util.py
```diff
@@ -32,6 +36,101 @@ def tearDown(self):
...
+        self.assertEquals(process_output, expected_output)
+
+    def test_start_subprocess_and_wait_for_completion_should_raise_on_timeout(self):
+        command = "sleep 4s"
+        timeout = 2
```
narrieta: avoid sleeping for 2 sec

azurelinuxagent/common/utils/processutil.py
```diff
@@ -17,11 +17,58 @@
...
+        time.sleep(1)
+        return_code = process.wait()
+
+    return timeout == 0, return_code
+
+
+def start_subprocess_and_wait_for_completion(command, timeout, shell, cwd, env, stdout, stderr, preexec_fn, error_code):
```
narrieta: document that stdout stderr must be files (we seek on them)

azurelinuxagent/common/utils/processutil.py
```diff
@@ -17,11 +17,58 @@
...
+
+    if timed_out:
+        raise ExtensionError("Timeout({0}): {1}\n{2}".format(timeout, command, process_output),
+                             code=ExtensionErrorCodes.PluginHandlerScriptTimedout)
+
+    if return_code != 0:
+        raise ExtensionError("Non-zero exit code: {0}, {1}\n{2}".format(return_code, command, process_output),
+                             code=error_code)
+
+    return process_output
```
narrieta: given the above comments you may want to reconsider whether these should be generic utility functions or stay as private methods where they are used

azurelinuxagent/common/cgroupapi.py
```diff
@@ -327,20 +340,26 @@ def pre_exec_function():
...
 
-        process = subprocess.Popen(
-            command,
-            shell=shell,
-            cwd=cwd,
-            stdout=stdout,
-            stderr=stderr,
-            env=env,
-            preexec_fn=pre_exec_function)
+        self.track_cgroups(extension_cgroups)
```
narrieta: Can you double-check the code in the telemetry thread? I believe it'll drop the cgroup if it does not have any processes. If that is the case, then we have a raise condition here, because we are starting the process after requesting the telemetry thread to track them

tests/common/test_cgroupapi.py
```diff
@@ -505,6 +497,88 @@ def mock_popen(*args, **kwargs):
...
+                return
+            else:
+                return original_popen(*args, **kwargs)
+
+        expected_output = "[stdout]\n{0}\n\n\n[stderr]\n"
+
+        with tempfile.TemporaryFile(dir=self.tmp_dir, mode="w+b") as stdout:
+            with tempfile.TemporaryFile(dir=self.tmp_dir, mode="w+b") as stderr:
+                with patch("azurelinuxagent.common.cgroupapi.subprocess.Popen", side_effect=mock_popen):
+                    with patch("azurelinuxagent.common.cgroupapi.wait_for_process_completion_or_timeout",
```
narrieta: given how critical this code is, I would also add tests that do not mock the timeout, but use a real process timing out (one of my previous comments has a suggestion on how to avoid the sleep)

you could that test or replace this test with that one

narrieta: similar comment for mocking systemd-run

narrieta: I know you are isolating components and writing unit tests for those components, but we need to complement (or replace them) with more "end-to-end" (integration) tests

tests/common/test_cgroupapi.py
```diff
@@ -505,6 +497,88 @@ def mock_popen(*args, **kwargs):
...
+                        command="ls folder_does_not_exist",
+                        timeout=300,
+                        shell=True,
+                        cwd=self.tmp_dir,
+                        env={},
+                        stdout=stdout,
+                        stderr=stderr)
+
+                    self.assertEquals(context_manager.exception.code, ExtensionErrorCodes.PluginUnknownFailure)
+                    self.assertIn("Non-zero exit code", ustr(context_manager.exception))
```
narrieta: how do we verify the fallback was not used?

azurelinuxagent/common/utils/processutil.py
pgombar: Thanks, added

pgombar: Updated

azurelinuxagent/common/utils/processutil.py
```diff
@@ -17,11 +17,58 @@
...
+        stdout=stdout,
+        stderr=stderr,
+        preexec_fn=preexec_fn)
+
+    # Wait for process completion or timeout
+    timed_out, return_code = wait_for_process_completion_or_timeout(process, timeout)
+    process_output = read_output(stdout, stderr)
+
+    if timed_out:
+        raise ExtensionError("Timeout({0}): {1}\n{2}".format(timeout, command, process_output),
```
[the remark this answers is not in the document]
pgombar: As discussed, the entire `processutil` module contains functions that are extension-related, so the solution is to rename the module to `extensionprocessutil` to make it clearer what it's meant for.

azurelinuxagent/common/utils/processutil.py
pgombar: Done

azurelinuxagent/common/utils/processutil.py
pgombar: Keeping them in this module, but renaming the module to be more specific to extension processes.

azurelinuxagent/common/utils/processutil.py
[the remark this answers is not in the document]
narrieta: and add comments explaining what it is meant for :)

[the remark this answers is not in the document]
pgombar: But of course :)

azurelinuxagent/common/cgroupapi.py
pgombar: We do a cleanup of inactive cgroups every time we poll them to report their telemetry, which is every 5 minutes. It is a race condition, I'll refactor the code to start tracking immediately after invoking the subprocess.

azurelinuxagent/common/cgroupapi.py
```diff
@@ -443,7 +462,13 @@ def create_cgroup(controller):
 
         return cgroups
 
-    def start_extension_command(self, extension_name, command, shell, cwd, env, stdout, stderr):
+    @staticmethod
+    def is_systemd_failure(scope_name, process_output):
```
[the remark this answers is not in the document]
pgombar: Updated

tests/utils/test_process_util.py
pgombar: Thanks, updated

tests/utils/test_process_util.py
pgombar: Updated test to output to both streams

tests/common/test_cgroupapi.py
pgombar: As discussed, this particular test relies on systemd timing out. For that to happen, I need to mock systemd-run. Other tests that test the extension command timing out or failing do not mock systemd-run.

tests/common/test_cgroupapi.py
```diff
@@ -300,24 +301,28 @@ def test_start_extension_command_should_add_the_child_process_to_the_extension_c
...
+        expected_output = "[stdout]\n{0}\n\n\n[stderr]\n"
 
         for cgroup in extension_cgroups:
             cgroups_procs_path = os.path.join(cgroup.path, "cgroup.procs")
             with open(cgroups_procs_path, "r") as f:
                 contents = f.read()
             pid = int(contents)
 
-            self.assertEquals(pid, process.pid, "The PID of the command was not added to {0}. Expected: {1}, got: {2}".format(cgroups_procs_path, process.pid, pid))
+            self.assertEquals(process_output, expected_output.format(pid),
```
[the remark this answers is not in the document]
pgombar: Reversed the assertion, thanks

tests/common/test_cgroupapi.py
pgombar: Added asserts for that

tests/utils/test_process_util.py
pgombar: As discussed, this test doesn't even need to synchronize with the child process since all it needs is the process to have already started. I'm mocking sleep and asserting we're "waiting" before killing the process. I've removed the mock for the kill to make the test more end-to-end.

tests/utils/test_process_util.py
pgombar: Mocked sleep in all the tests

narrieta: yay!

azurelinuxagent/common/utils/processutil.py
pgombar: Discussed offline, the comment was already here from before and doesn't change w.r.t refactoring this method out of the `exthandlers` module.

PR #1626, opened 2019-09-09
Extension update on failure

larohra:
## Description
Currently the update sequence of an extension consists of the following operations - 
Disable (OV) -> Update (NV) -> Uninstall (OV) -> Install if updateMode == UpdateWithInstall (NV) -> Enable (NV).

This PR addresses the issue where Extension Updates are blocked if there's any failure in the old version of the Extension. We've introduced a new flag 'continueOnUpdateFailure' which can be set in the new HandlerManifer of the extension to unblock the extension update.

Issue # 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1626)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -486,33 +491,44 @@ def handle_enable(self, ext_handler_i):
...
             """
+            continue_on_update_failure = False
             try:
+                continue_on_update_failure = ext_handler_i.load_manifest().is_continue_on_update_failure()
                 func()
             except ExtensionError as e:
                 # Reporting the event with the old handler and raising a new ExtensionUpdateError to set the
                 # handler status on the new version
-                msg = ustr(e)
+                msg = "%s; ContinueOnUpdate: %s" % (ustr(e), continue_on_update_failure)
```
larohra: Logging the status of ContinueOnUpdate flag incase of errors

azurelinuxagent/ga/exthandlers.py
```diff
@@ -652,6 +674,8 @@ def __init__(self, ext_handler, protocol):
         self.is_upgrade = False
         self.logger = None
         self.set_logger()
+        self.handler_env = {EXTENSION_PATH: self.get_base_dir(),
```
larohra: Added a new property in ExtHandlerInstance to maintain a common set of Environment Variables that would be passed to every launch_command call

azurelinuxagent/ga/exthandlers.py
```diff
@@ -949,14 +980,11 @@ def install(self):
         self.set_handler_state(ExtHandlerState.Installed)
 
     def uninstall(self):
-        try:
-            self.set_operation(WALAEventOperation.UnInstall)
-            man = self.load_manifest()
-            uninstall_cmd = man.get_uninstall_command()
-            self.logger.info("Uninstall extension [{0}]".format(uninstall_cmd))
-            self.launch_command(uninstall_cmd)
-        except ExtensionError as e:
```
larohra: Now we'd only be swallowing exception if the user is trying to delete an extension but would not swallow it if the extension update fails.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1436,3 +1466,6 @@ def is_update_with_install(self):
         if update_mode is None:
             return True
         return update_mode.lower() == "updatewithinstall"
+
+    def is_continue_on_update_failure(self):
+        return self.data['handlerManifest'].get('continueOnUpdateFailure', False)
```
larohra: Added a new flag `continueOnUpdateFailure` to specify weather to continue of update failure or not

tests/ga/test_extension.py
```diff
@@ -294,6 +300,7 @@ def tearDownClass(cls):
             CGroupConfigurator.get_instance().disable()
 
 
+@patch('time.sleep', side_effect=lambda _: mock_sleep(0.001))
```
larohra: Added a mock sleep to this class to reduce the time it took for running test cases

azurelinuxagent/ga/exthandlers.py
pgombar: We should also log which operation failed (`op`) in the message

azurelinuxagent/ga/exthandlers.py
```diff
@@ -486,33 +491,44 @@ def handle_enable(self, ext_handler_i):
...
+        def execute_old_handler_command(op, func):
             """
             Created a common wrapper to execute all commands that need to be executed from the old handler
             so that it can have a common exception handling mechanism
+            :param op: The current running operation
             :param func: The command to be executed on the old handler
             """
+            continue_on_update_failure = False
             try:
+                continue_on_update_failure = ext_handler_i.load_manifest().is_continue_on_update_failure()
```
pgombar: I just realized that, on each extension command, we call `load_manifest` which in turn always reads the file from disk and parses it. I don't see why we would have to parse the same file over and over again. Can we have a method that will, when called the first time, parse the file, and when called subsequently just return the parsed contents?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -76,6 +76,11 @@
 
 NUMBER_OF_DOWNLOAD_RETRIES = 5
 
+DISABLE_FAILED = "DISABLE_FAILED"
```
pgombar: These strings seem generic and can potentially collide with existing environment variables. Can we add something like a `WALinuxAgent` prefix to them so the chance of interfering with customer env variables is lower?

azurelinuxagent/ga/exthandlers.py
pgombar: What's the purpose of these environment variables? Maybe I'm missing something, but I don't see them being used anywhere?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -532,7 +548,13 @@ def handle_uninstall(self, ext_handler_i):
         if handler_state != ExtHandlerState.NotInstalled:
             if handler_state == ExtHandlerState.Enabled:
                 ext_handler_i.disable()
-            ext_handler_i.uninstall()
+
+            # Try uninstalling the extension and swallow any exceptions incase of failures after logging them
```
pgombar: nit: incase -> in case

azurelinuxagent/ga/exthandlers.py
pgombar: Do you mean the reverse? After your changes, a simple extension uninstall would result in an exception, but if it's part of an update sequence, we would swallow it and not block the update.

tests/ga/test_extension.py
pgombar: This is awesome! Our unit test suite is now executing in ~1 minute instead of 7!

tests/ga/test_extension.py
```diff
@@ -336,6 +343,34 @@ def _create_mock(self, test_data, mock_http_get, MockCryptUtil):
         handler.protocol_util.get_protocol = Mock(return_value=protocol)
         return handler, protocol
 
+    def _setup_update_test_and_update_gs(self, patch_command, *args):
+        """
+        This helper function sets up the Update test by setting up the protocol and ext_handler and asserts the
+        ext_handler runs fine the first time before patching a failure command for testing.
+        :param patch_command: The patch_command to setup for failure
```
pgombar: nit: setup (noun) -> set up (verb)

tests/ga/test_extension.py
```diff
@@ -1422,23 +1411,9 @@ def test__extension_upgrade_failure_when_prev_version_disable_fails_and_recovers
...
-        self.assertEqual(0, patch_get_disable_command.call_count)
-
-        self._assert_handler_status(protocol.report_vm_status, "Ready", expected_ext_count=1, version="1.0.0")
-        self._assert_ext_status(protocol.report_ext_status, "success", 0)
-
-        # Next incarnation, update version
-        test_data.goal_state = test_data.goal_state.replace("<Incarnation>1<", "<Incarnation>2<")
-        test_data.ext_conf = test_data.ext_conf.replace('version="1.0.0"', 'version="1.0.1"')
-        test_data.manifest = test_data.manifest.replace('1.0.0', '1.0.1')
+        test_data, exthandlers_handler, protocol = self._setup_update_test_and_update_gs(patch_get_disable_command,
```
pgombar: The tests look much cleaner with this being refactored in its own method. Nice!

tests/ga/test_extension.py
```diff
@@ -1513,26 +1475,129 @@ def test_upgrade_failure_with_exception_handling(self, patch_get_update_command,
...
+                             "This should be called twice, for both disable and uninstall")
 
-        # Update command should fail
-        patch_get_update_command.return_value = "exit 1"
-        exthandlers_handler.run()
-        self.assertEqual(1, patch_get_update_command.call_count)
-        self.assertEqual(1, patch_handle_ext_handler_error.call_count)
+        self._assert_handler_status(protocol.report_vm_status, "Ready", expected_ext_count=1, version="1.0.1")
+
+    @patch('azurelinuxagent.ga.exthandlers.HandlerManifest.is_continue_on_update_failure', return_value=True)
```
pgombar: You already have this patch in the test.

tests/ga/test_extension.py
```diff
@@ -1513,26 +1475,129 @@ def test_upgrade_failure_with_exception_handling(self, patch_get_update_command,
...
+    @patch('azurelinuxagent.ga.exthandlers.HandlerManifest.get_uninstall_command')
+    def test_extension_upgrade_should_pass_when_continue_on_update_failue_is_true_and_prev_version_uninstall_fails(
```
pgombar: nit: failue -> failure

pgombar: Actually, there are several occurrences of this typo. Search and replace :)

tests/ga/test_extension.py
```diff
@@ -1513,26 +1475,129 @@ def test_upgrade_failure_with_exception_handling(self, patch_get_update_command,
...
+            self.assertTrue(DISABLE_FAILED == disable_args[1])
+            self.assertTrue(UNINSTALL_FAILED == uninstall_args[1])
+
+        self._assert_handler_status(protocol.report_vm_status, "Ready", expected_ext_count=1, version="1.0.1")
+
+    def test_ext_path_and_version_env_variables_set_for_ever_operation(self, *args):
+        test_data = WireProtocolData(DATA_FILE_EXT_SINGLE)
+        exthandlers_handler, protocol = self._create_mock(test_data, *args)
+
+        with patch('azurelinuxagent.ga.exthandlers.ExtHandlerInstance._capture_process_output',
```
pgombar: Depending on which PR goes in first, this logic changes with #1623

tests/ga/test_extension.py
```diff
@@ -1961,5 +2026,174 @@ def test_ext_handler_with_systemd_cgroup_enabled(self, *args):
         self._assert_no_handler_status(protocol.report_vm_status)
 
 
+@patch('azurelinuxagent.ga.exthandlers.ExtHandlerInstance._capture_process_output', side_effect="Process Successful")
```
pgombar: Same comment, logic will change

tests/ga/test_extension.py
```diff
@@ -1961,5 +2026,174 @@ def test_ext_handler_with_systemd_cgroup_enabled(self, *args):
...
+            "continueOnUpdateFailure": continue_on_update_failure
+        }
+
+        ext_handler_properties = ExtHandlerProperties()
+        ext_handler_properties.version = version
+        ext_handler = ExtHandler(name=name)
+        ext_handler.properties = ext_handler_properties
+        ext_handler_i = ExtHandlerInstance(ext_handler=ext_handler, protocol=None)
+        ext_handler_i.load_manifest = MagicMock(return_value=HandlerManifest({'handlerManifest': handler_json}))
+        fileutil.mkdir(ext_handler_i.get_base_dir())
```
pgombar: This should be moved/mocked to a temporary folder which gets deleted on test cleanup.

azurelinuxagent/ga/exthandlers.py
larohra: Good point but we're only setting the environment variables for that particular process and not the systemwide environment variables. That being said we do pass `os.environ` to the process too which might have the same variables set by the customer. The chance is low of a conflict but definitely can happen. 

How about GUEST_AGENT_DISABLE_FAILED as we want the variable names to be generic for both our agents?

azurelinuxagent/ga/exthandlers.py
larohra: Good point while doing this code I felt the same way. We can definitely make it part of the ExtHandlerInstance or ExtHandler object to avoid the extra I/O. 
That would be another PR though as this issue is outside the scope of this problem :)

azurelinuxagent/ga/exthandlers.py
larohra: The ExtensionError exception message `e` would already have the specifics of weather disable or uninstall failed. I feel logging the `op` again would just be redundant

azurelinuxagent/ga/exthandlers.py
larohra: This was an ask by the publishers to provide them with Extension_path and Extension_Version on every call to the extension so that it would save them the time to parse it.

azurelinuxagent/ga/exthandlers.py
larohra: No No this function `def uninstall(self):` is called twice - Once from `def handle_uninstall(self, ext_handler_i):` and other time from the `def _update_extension_handler(old_ext_handler_i, ext_handler_i):`. 
Initially we were swallowing exceptions inside this `def uninstall(self):` method which used to always eat the exception. I moved the swallow logic to this `def handle_uninstall(self, ext_handler_i):` method which is only called when the customer initiates an extension removal from the portal. Now the exception would not be swallowed during the update scenario where it will be caught and propagated as a `ExtensionUpdateError` if the `continueOnUpdateFailure` flag is not set.

tests/ga/test_extension.py
larohra: haha thanks :)

tests/ga/test_extension.py
larohra: Thanks 💃

tests/ga/test_extension.py
larohra: Ohh yeah thanks good catch!

tests/ga/test_extension.py
larohra: Haha thanks done!

tests/ga/test_extension.py
larohra: Ohh yeah. Your changes would actually make that patch obsolete and better :)
The fix would be super simple for this test case whenever we need to change it

tests/ga/test_extension.py
larohra: This patch was just to avoid waiting for the output. With your change it would be even simpler as I can just remove this patch or add a patch to start_extension_command instead

tests/ga/test_extension.py
larohra: The base director `ext_handler_i.get_base_dir()` is already set to a /tmp/ folder so no additional checks are needed.
Eg - `'/tmp/TestExtensionUpdateOnFailure_fk9fsa6r/foo-1.0.0'`

tests/ga/test_extension.py

azurelinuxagent/ga/exthandlers.py

azurelinuxagent/ga/exthandlers.py
narrieta: Maybe add "AZURE" somewhere there?

larohra: sounds good, I'll add AZURE_GUEST_AGENT prefix to all the variables (DISABLE_FAILED, UNINSTALL_FAILED, EXTENSION_PATH and EXTENSION_VERSION

azurelinuxagent/ga/exthandlers.py
```diff
@@ -76,6 +76,11 @@
 
 NUMBER_OF_DOWNLOAD_RETRIES = 5
 
+DISABLE_FAILED = "DISABLE_FAILED"
+UNINSTALL_FAILED = "UNINSTALL_FAILED"
+EXTENSION_PATH = "EXTENSION_PATH"
+EXTENSION_VERSION = "EXTENSION_VERSION"
```
narrieta: There is already a VERSION env variable that gets set in ExtHandlerInstance.update

larohra: Yes but that's only there for when the update command is called. This would be available for every command. I didnt wanna remove VERSION variable because customers might have some dependencies on it

azurelinuxagent/ga/exthandlers.py
narrieta: why not explicitly define those variables in each of the commands (update, enable, etc)? By adding this member now we need to be sure its state is kept update as the workflow progresses, plus is a little hard to follow which variables are set on which operations

azurelinuxagent/ga/exthandlers.py
```diff
@@ -486,33 +491,44 @@ def handle_enable(self, ext_handler_i):
             ext_handler_i.update_settings()
 
         ext_handler_i.enable()
+        # Remove the UNINSTALL_FAILED from env variables list after calling Install and Enable
+        ext_handler_i.remove_key_from_handler_env(UNINSTALL_FAILED)
```
narrieta: if enable throws the env doesn't get updated

azurelinuxagent/ga/exthandlers.py
pgombar: But the call to `report_event` doesn't have the operation, which means we couldn't tell from our telemetry if disable or uninstall failed.

narrieta: Why do we initialize continue_on_update_failure within the try except? thanks

larohra: @pgombar we set the operation on every call to disable/enable/install/etc (like `self.set_operation(WALAEventOperation.Disable)`). If there's an error thrown then the report event just uses that same operation to report the event `op=self.operation,`. Thanks for pointing this out though, I'll reuse the same `op=self.operation` in my function instead of passing another op parameter.

larohra: @narrieta I added it to the try catch because the load_manifest might throw an ExtensionError which we should catch and report as an ExtensionUpdateError (since it is being caught during the update sequence)

```python
continue_on_update_failure = ext_handler_i.load_manifest().is_continue_on_update_failure()


    def load_manifest(self):
        man_file = self.get_manifest_file()
        try:
            data = json.loads(fileutil.read_file(man_file))
        except (IOError, OSError) as e:
            raise ExtensionError('Failed to load manifest file ({0}): {1}'.format(man_file, e.strerror),
                                 code=ExtensionErrorCodes.PluginHandlerManifestNotFound)
        except ValueError:
            raise ExtensionError('Malformed manifest file ({0}).'.format(man_file),
                                 code=ExtensionErrorCodes.PluginHandlerManifestDeserializationError)

        return HandlerManifest(data[0])
```

azurelinuxagent/ga/exthandlers.py
narrieta: yes, we should not remove it. It is odd. though, that VERSION and EXTENSION_VERSION are handled differently in the code (they have the same value and semantics, right?

azurelinuxagent/ga/exthandlers.py
narrieta: self.handler_env  should be private

azurelinuxagent/ga/exthandlers.py
narrieta: >  Now the exception would not be swallowed during the update scenario where it will be caught and propagated as a ExtensionUpdateError if the continueOnUpdateFailure flag is not set.

The current implementation doesn't distinguish between the new flag not being set, and it being set to false. We should change the behavior only when the flag is actually set (either to True or False)

narrieta: never mind, it's the same behavior

tests/ga/test_extension.py
narrieta: 1 min!!!! Dude, I need to try this myself -- if True, I'll buy you a beer.

pgombar: Just check out his Travis run times: https://travis-ci.org/Azure/WALinuxAgent/builds/583298862?utm_source=github_status&utm_medium=notification
Laveesh deserves all the beers 😄

azurelinuxagent/ga/exthandlers.py
larohra: I'm setting the handler_env in the `ExtHandlerInstance` object. If enable throws then that particular object is deleted and no future states are affected.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -486,33 +491,44 @@ def handle_enable(self, ext_handler_i):
...
+
+                logger.info("Continue on Update failure flag is set, proceeding with update")
+                error_env_name = DISABLE_FAILED if op == WALAEventOperation.Disable else UNINSTALL_FAILED
+                ext_handler_i.set_handler_env_variable({error_env_name: 'True'})
 
-        execute_old_handler_command(lambda: old_ext_handler_i.disable())
+        execute_old_handler_command(op=WALAEventOperation.Disable, func=lambda: old_ext_handler_i.disable())
         ext_handler_i.copy_status_files(old_ext_handler_i)
         if ext_handler_i.version_gt(old_ext_handler_i):
             ext_handler_i.update()
```
larohra: Similarly for this, if update throws then the Extension processing is stopped and the object is deleted, not causing any state mismatches at all

azurelinuxagent/ga/exthandlers.py
```diff
@@ -486,33 +491,44 @@ def handle_enable(self, ext_handler_i):
...
+            # Remove the DISABLE_FAILED from env variables list after calling Update
+            ext_handler_i.remove_key_from_handler_env(DISABLE_FAILED)
```
larohra: I need this line still because we documented to the extension publishers that the DISABLE_FAILED environment variable would only be available for update command. If the update command was successful, we need to remove it from the handler_env to ensure that we dont send it to any other commands (install/enable/uninstall)

tests/ga/test_extension.py
```diff
@@ -1513,26 +1475,128 @@ def test_upgrade_failure_with_exception_handling(self, patch_get_update_command,
...
+    def test_extension_upgrade_should_pass_when_continue_on_update_failue_is_true_and_prev_version_uninstall_fails(
+            self, patch_get_uninstall_command, *args):
+        test_data, exthandlers_handler, protocol = self._set_up_update_test_and_update_gs(patch_get_uninstall_command,
+                                                                                          *args)
+
+        with patch('azurelinuxagent.ga.exthandlers.HandlerManifest.is_continue_on_update_failure', return_value=True) \
+                as mock_continue_on_update_failure:
+            exthandlers_handler.run()
+            self.assertEqual(1, patch_get_uninstall_command.call_count)
+            self.assertEqual(2, mock_continue_on_update_failure.call_count,
```
narrieta: asserting on the call count of is_continue_on_update_failure() makes this test too fragile (too dependent on implementation). For example, if we implemented Paula's suggestion of reading the manifest only once and caching it, most likely this test would break). Do we have a better way to assert this? Can we assert that disable and uninstalled were called?

I know existing tests have similar issues by heavily relying on the call count of mocks of the get_*_command functions, but hopefully we'll improve those over time.

azurelinuxagent/ga/exthandlers.py
larohra: A new object of ExtHandlerInstance is created everytime the agent starts extension processing so the state maintenance is only for that particular goalstate call, which is easier to maintain

larohra: Good catch, I'll make this change

tests/ga/test_extension.py
larohra: haha always down for a beer 🍻

tests/ga/test_extension.py
```diff
@@ -1961,5 +2026,174 @@ def test_ext_handler_with_systemd_cgroup_enabled(self, *args):
...
+        handler_i.remove_key_from_handler_env(name_to_delete)
+        self.assertTrue(name_to_delete not in handler_i.handler_env, "The variable not removed from handler_env list")
+
+        # Retrying the removal operation to ensure that no exception is thrown if key not found
+        try:
+            handler_i.remove_key_from_handler_env(name_to_delete)
+        except:
+            self.fail("Removing a key that's not present shouldn't throw an error")
+
+
```
narrieta: we should also have tests that check the env variables in the actual extension commands

azurelinuxagent/ga/exthandlers.py
larohra: So version is passed as a parameter to the update() command IF we're downgrading the extension.
`old_ext_handler_i.update(version=ext_handler_i.ext_handler.properties.version)`

 Otherwise it uses the same semantics - 
```python
    def update(self, version=None):
        if version is None:
```

tests/ga/test_extension.py
```diff
@@ -1961,5 +2054,240 @@ def test_ext_handler_with_systemd_cgroup_enabled(self, *args):
...
+        # Script prints env variables passed to this process and prints all starting with AZURE_
+        test_file = """
+            printenv | grep AZURE_
+            """
+
+        self.create_script(file_name=test_file_name, contents=test_file,
+                           file_path=os.path.join(new_handler_i.get_base_dir(), test_file_name))
+
+        with patch.object(new_handler_i.logger, 'verbose', autospec=True) as mock_verbose:
+            # Since we're not mocking the ExtHandlerInstance._capture_process_output, both disable.cmd and uninstall.cmd
```
pgombar: Outdated comment?

larohra: Ohh yeah good catch! I'll send an update

tests/ga/test_extension.py

azurelinuxagent/ga/exthandlers.py
narrieta: You need to protect this against exceptions in the code above this; for example, if update() raises you wouldn't remove the variable

tests/tools.py
```diff
@@ -438,12 +438,13 @@ def _create_files(tmp_dir, prefix, suffix, count, with_sleep=0):
...
             time.sleep(with_sleep)
 
-    def _create_script(self, file_name, contents):
+    def create_script(self, file_name, contents, file_path=None):
         """
         Creates an executable script with the given contents.
         If file_name ends with ".py", it creates a Python3 script, otherwise it creates a bash script
         """
-        file_path = os.path.join(self.tmp_dir, file_name)
+        if not file_path:
```
narrieta: let's document the new parameter

tests/ga/test_extension.py
```diff
@@ -1961,5 +2054,241 @@ def test_ext_handler_with_systemd_cgroup_enabled(self, *args):
...
+
+        self.create_script(file_name=test_file_name, contents=test_file,
+                           file_path=os.path.join(new_handler_i.get_base_dir(), test_file_name))
+
+        with patch.object(new_handler_i.logger, 'verbose', autospec=True) as mock_verbose:
+            # Since we're not mocking the azurelinuxagent.common.cgroupconfigurator..handle_process_completion,
+            # both disable.cmd and uninstall.cmd would raise ExtensionError exceptions and set the DISABLE_FAILED and
+            # UNINSTALL_FAILED env variables.
+            # For update and install we're running the script above to print all the env variables starting with AZURE_
+            # and verify accordingly if the corresponding env variables are set properly or not
```
narrieta: let's add a small comment explaining why the variables show in the verbose log

azurelinuxagent/ga/exthandlers.py

tests/ga/test_extension.py

tests/tools.py
larohra: Done

PR #1628, opened 2019-09-10
Retrieve latest container id from cache when sending telemetry events

pgombar:
## Description

Today, we populate a group of parameters called `vminfo` once at the start of the monitoring thread. One of these parameters is `ContainerId` which can change without redeploying the resource and restarting the agent (e.g. during live migration).

Adding the functionality to fetch the `ContainerId` parameter during the construction of the telemetry event from the process environment variables. The value is set during a goal state update.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1628)

### DIFF
--- a/azurelinuxagent/common/event.py
+++ b/azurelinuxagent/common/event.py
@@ -28,6 +28,7 @@ import azurelinuxagent.common.conf as conf
 import azurelinuxagent.common.logger as logger
 from azurelinuxagent.common.exception import EventError
 from azurelinuxagent.common.future import ustr
+from azurelinuxagent.common.protocol.wire import CONTAINER_ID
 from azurelinuxagent.common.protocol.restapi import TelemetryEventParam, \
     TelemetryEvent, \
     get_properties
@@ -93,6 +94,7 @@ SHOULD_ENCODE_MESSAGE_OP = [
     WALAEventOperation.Install,
     WALAEventOperation.UnInstall,
 ]
+
 
 class EventStatus(object):
     EVENT_STATUS_FILE = "event_status.json"
@@ -277,6 +279,7 @@ class EventLogger(object):
         event.parameters.append(TelemetryEventParam('Message', message))
         event.parameters.append(TelemetryEventParam('Duration', duration))
         event.parameters.append(TelemetryEventParam('ExtensionType', evt_type))
+        event.parameters.append(TelemetryEventParam('ContainerId', os.environ.get(CONTAINER_ID, None)))
 
         data = get_properties(event)
         try:
--- a/azurelinuxagent/common/event.py
+++ b/azurelinuxagent/common/event.py
@@ -28,7 +28,7 @@ import azurelinuxagent.common.conf as conf
 import azurelinuxagent.common.logger as logger
 from azurelinuxagent.common.exception import EventError
 from azurelinuxagent.common.future import ustr
-from azurelinuxagent.common.protocol.wire import CONTAINER_ID
+from azurelinuxagent.common.protocol.wire import CONTAINER_ID_ENV_VARIABLE
 from azurelinuxagent.common.protocol.restapi import TelemetryEventParam, \
     TelemetryEvent, \
     get_properties
@@ -279,7 +279,8 @@ class EventLogger(object):
         event.parameters.append(TelemetryEventParam('Message', message))
         event.parameters.append(TelemetryEventParam('Duration', duration))
         event.parameters.append(TelemetryEventParam('ExtensionType', evt_type))
-        event.parameters.append(TelemetryEventParam('ContainerId', os.environ.get(CONTAINER_ID, None)))
+        event.parameters.append(TelemetryEventParam('ContainerId',
+                                                    os.environ.get(CONTAINER_ID_ENV_VARIABLE, "UNINITIALIZED")))
 
         data = get_properties(event)
         try:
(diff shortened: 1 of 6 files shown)


azurelinuxagent/ga/monitor.py
```diff
@@ -287,6 +288,19 @@ def add_sysinfo(self, event):
                 event.parameters.remove(param)
         event.parameters.extend(self.sysinfo)
 
+    def update_container_id(self, event):
+        last_known_container_id = self.protocol.client.get_container_id_from_goal_state()
```
narrieta: can self.protocol be an instance of Metadata? (Azure Stack)

narrieta: how does this instance of protocol gets its client's goal state updated? this instance is not the same as the one in the extension handler thread.

pgombar: You are correct. I see two possible solutions for this: 1) make `HostPluginProtocol` a static class, or 2) always read the container id from the goal state on disk.

pgombar: It can, in which case any reference to `self.protocol.client` would break.
Except this instance, we already have another one checked in :( Here it is: https://github.com/Azure/WALinuxAgent/blame/develop/azurelinuxagent/ga/monitor.py#L336

pgombar: Went for the environment variable approach to share state between threads.

pgombar: Removed.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -56,6 +56,7 @@
 AGENTS_MANIFEST_FILE_NAME = "{0}.{1}.agentsManifest"
 TRANSPORT_CERT_FILE_NAME = "TransportCert.pem"
 TRANSPORT_PRV_FILE_NAME = "TransportPrivate.pem"
+CONTAINER_ID = "AZURE_GUEST_AGENT_CONTAINER_ID"
```
narrieta: the name of the variable gives me the impression that its value is the container id, try CONTAINER_ID_ENV_VARIABLE or similar

azurelinuxagent/common/event.py
```diff
@@ -277,6 +279,7 @@ def _add_event(self, duration, evt_type, is_internal, is_success, message, name,
         event.parameters.append(TelemetryEventParam('Message', message))
         event.parameters.append(TelemetryEventParam('Duration', duration))
         event.parameters.append(TelemetryEventParam('ExtensionType', evt_type))
+        event.parameters.append(TelemetryEventParam('ContainerId', os.environ.get(CONTAINER_ID, None)))
```
narrieta: a better default would be "UNINITIALIZED", or something like that. we don't want to be confused by values in Kusto

azurelinuxagent/common/protocol/wire.py
narrieta: let's also document what the variable is used for (basically passing a value across threads)

tests/common/test_event.py
```diff
@@ -239,7 +270,7 @@ def test_save_event_message_with_non_ascii_characters(self):
             event_str = MonitorHandler.collect_event(os.path.join(self.tmp_dir, tld_file))
             event_json = json.loads(event_str)
 
-            self.assertEqual(len(event_json["parameters"]), 8)
+            self.assertEqual(len(event_json["parameters"]), 9)
```
narrieta: why did the count increment?

tests/ga/test_extension.py
```diff
@@ -27,6 +27,12 @@
 from azurelinuxagent.ga.exthandlers import *
 from azurelinuxagent.common.protocol.wire import WireProtocol, InVMArtifactsProfile
 
+SLEEP = time.sleep
```
narrieta: add comments

tests/common/test_event.py
```diff
@@ -34,6 +35,36 @@
...
+            xml_text = load_data("wire/goal_state.xml")
+            goal_state = GoalState(xml_text)
+
+            container_id = goal_state.container_id
+            event.add_event(name='dummy_name')
+            data = fileutil.read_file(tmp_file)
+            self.assertIn('{{"name": "ContainerId", "value": "{0}"}}'.format(container_id), data)
+
+        os.environ.pop(event.CONTAINER_ID)
+
```
larohra: Awesome tests! 1 suggestion - You can add 1 more case where you ensure that change in container ID in GoalState is actually modifying containerId in add_event

azurelinuxagent/common/event.py
pgombar: Good point. Will update.

azurelinuxagent/common/protocol/wire.py
pgombar: Thanks, updated.

tests/common/test_event.py
pgombar: This tests counts the existing telemetry parameters before `sys_info` is added. Since I moved the `ContainerId` parameter from the `sys_info` parameter group to be populated at event creation time, it reflects as an additional parameter in the first stage.

tests/ga/test_extension.py
pgombar: Ah sorry, I've been adding this to all my PRs ever since @larohra discovered this magical mock :) forgot to add a comment to this instance.

tests/common/test_event.py
pgombar: Agreed, useful test case. Thanks, added!

tests/ga/test_extension.py
vrdmr: Minor - Can you rename the var to MOCK_SLEEP? Only SLEEP felt like a boolean to me.

tests/ga/test_extension.py
```diff
@@ -27,11 +27,12 @@
 from azurelinuxagent.ga.exthandlers import *
 from azurelinuxagent.common.protocol.wire import WireProtocol, InVMArtifactsProfile
 
-SLEEP = time.sleep
+# Mock sleep to reduce test execution time
+MOCK_SLEEP = time.sleep
```
narrieta: this is not the mock sleep; this is the original sleep

PR #1632, opened 2019-09-13
Add telemetry event for HostGAPlugin retry logic

pgombar:
## Description

We want to emit telemetry whenever we invoke the HostGAPlugin retry logic and log the success or failure of the calls (first or second try). For that, `WireProtocol` needed to use the `add_event` method, which introduced circular dependencies.

Imports are fixed with the following changes:
- Extract the `DataContract` object(s) and related methods to a common module outside of `restapi`.
- Extract the `TelemetryEvent` object(s) to a common module outside of `restapi`.
- Update all imports referencing these objects and methods.

Also, added sleep mocking in `test_extension` module to reduce unit test execution time from 7+ minutes to ~2.5 minutes.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1632)

### DIFF
--- a/azurelinuxagent/common/datacontract.py
+++ b/azurelinuxagent/common/datacontract.py
@@ -1,0 +1,83 @@
+# Microsoft Azure Linux Agent
+#
+# Copyright 2019 Microsoft Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Requires Python 2.6+ and Openssl 1.0+
+#
+
+from azurelinuxagent.common.exception import ProtocolError
+import azurelinuxagent.common.logger as logger
+
+"""
+Data contract between guest and host
+"""
+
+
+class DataContract(object):
+    pass
+
+
+class DataContractList(list):
+    def __init__(self, item_cls):
+        self.item_cls = item_cls
+
+
+def validate_param(name, val, expected_type):
+    if val is None:
+        raise ProtocolError("{0} is None".format(name))
+    if not isinstance(val, expected_type):
+        raise ProtocolError(("{0} type should be {1} not {2}"
+                             "").format(name, expected_type, type(val)))
+
+
+def set_properties(name, obj, data):
+    if isinstance(obj, DataContract):
+        validate_param("Property '{0}'".format(name), data, dict)
+        for prob_name, prob_val in data.items():
+            prob_full_name = "{0}.{1}".format(name, prob_name)
+            try:
+                prob = getattr(obj, prob_name)
+            except AttributeError:
+                logger.warn("Unknown property: {0}", prob_full_name)
+                continue
+            prob = set_properties(prob_full_name, prob, prob_val)
+            setattr(obj, prob_name, prob)
+        return obj
+    elif isinstance(obj, DataContractList):
+        validate_param("List '{0}'".format(name), data, list)
+        for item_data in data:
+            item = obj.item_cls()
+            item = set_properties(name, item, item_data)
+            obj.append(item)
+        return obj
+    else:
+        return data
+
+
+def get_properties(obj):
+    if isinstance(obj, DataContract):
+        data = {}
+        props = vars(obj)
+        for prob_name, prob in list(props.items()):
+            data[prob_name] = get_properties(prob)
+        return data
+    elif isinstance(obj, DataContractList):
+        data = []
+        for item in obj:
+            item_data = get_properties(item)
+            data.append(item_data)
+        return data
+    else:
+        return obj
--- a/azurelinuxagent/common/datacontract.py
+++ b/azurelinuxagent/common/datacontract.py
@@ -21,7 +21,7 @@ from azurelinuxagent.common.exception import ProtocolError
 import azurelinuxagent.common.logger as logger
 
 """
-Data contract between guest and host
+Base class for data contracts between guest and host and utilities to manipulate the properties in those contracts
 """
 
 
--- a/azurelinuxagent/common/protocol/metadata.py
+++ b/azurelinuxagent/common/protocol/metadata.py
@@ -25,9 +25,14 @@ import sys
 import traceback
 
 import azurelinuxagent.common.conf as conf
+from azurelinuxagent.common.datacontract import get_properties, set_properties, validate_param
+from azurelinuxagent.common.exception import HttpError, ProtocolError
+import azurelinuxagent.common.logger as logger
+from azurelinuxagent.common.utils import restutil
 import azurelinuxagent.common.utils.fileutil as fileutil
 import azurelinuxagent.common.utils.shellutil as shellutil
 import azurelinuxagent.common.utils.textutil as textutil
+from azurelinuxagent.common.telemetryevent import TelemetryEventList
 
 from azurelinuxagent.common.future import httpclient
 from azurelinuxagent.common.protocol.restapi import *
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -1061,14 +1061,14 @@ class WireClient(object):
                               message=msg,
                               log_event=False)
 
-            except Exception as e:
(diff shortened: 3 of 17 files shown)


azurelinuxagent/common/datacontract.py
```diff
@@ -0,0 +1,83 @@
...
+# limitations under the License.
+#
+# Requires Python 2.6+ and Openssl 1.0+
+#
+
+from azurelinuxagent.common.exception import ProtocolError
+import azurelinuxagent.common.logger as logger
+
+"""
+Data contract between guest and host
```
narrieta: "Base class for data contracts between guest and host and utilities to manipulate the properties in those contracts"

azurelinuxagent/common/datacontract.py
```diff
@@ -0,0 +1,83 @@
...
+"""
```
narrieta: There some imports referencing the old module, please do a global search of usages of these classes/methods.

E.g.
> from azurelinuxagent.common.protocol.restapi import DataContract, set_properties

in imds.py

azurelinuxagent/common/protocol/metadata.py
```diff
@@ -25,9 +25,14 @@
 import traceback
 
 import azurelinuxagent.common.conf as conf
+from azurelinuxagent.common.datacontract import get_properties, set_properties, validate_param
+from azurelinuxagent.common.exception import HttpError, ProtocolError
```
narrieta: what was happening here? I see metadata.py does reference HttpError, etc... but there was no import before???

narrieta: was working because of the import * from restapi at line 38?

tests/ga/test_extension.py
```diff
@@ -27,6 +27,12 @@
 from azurelinuxagent.ga.exthandlers import *
 from azurelinuxagent.common.protocol.wire import WireProtocol, InVMArtifactsProfile
 
+SLEEP = time.sleep
```
narrieta: Let's add some comments here

azurelinuxagent/common/protocol/metadata.py
larohra: restapi was importing this `from azurelinuxagent.common.exception import ProtocolError, HttpError` before and the `import *` was pulling it in

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1024,18 +1031,45 @@ def send_request_using_appropriate_channel(self, direct_func, host_func):
...
+                              log_event=False)
+
+            except Exception as e:
+                add_event(name=AGENT_NAME,
+                          version=CURRENT_VERSION,
+                          op=WALAEventOperation.HostPlugin,
+                          is_success=False,
+                          message=ustr(e),
+                          log_event=False)
+                raise e
```
narrieta: "raise" instead of "raise e" to re-raise the original exception

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1024,18 +1031,45 @@ def send_request_using_appropriate_channel(self, direct_func, host_func):
...
+
```
narrieta: Why is this needed?

```
         except Exception:
            raise
```

azurelinuxagent/common/protocol/metadata.py
narrieta: yup --- hidden/unintended dependencies --- please stay away from this kind of imports in new code (and let's try to fix existing code when possible)

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1024,18 +1031,45 @@ def send_request_using_appropriate_channel(self, direct_func, host_func):
...
+                                                                             old_role_config_name, new_role_config_name)
+                    add_event(name=AGENT_NAME,
+                              version=CURRENT_VERSION,
+                              op=WALAEventOperation.HostPlugin,
+                              is_success=True,
+                              message=msg,
+                              log_event=False)
+
+            except Exception as e:
+                add_event(name=AGENT_NAME,
```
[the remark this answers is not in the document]
larohra: why restrict the error to only ResourceGone and InvalidContainer? Shouldn't we log that we failed irrespective of the error type?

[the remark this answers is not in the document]
narrieta: the intention is to measure whether the refresh helped to solve ResourceGone/InvalidContainer
if retry fails with ResourceGone/InvalidContainer then it didn't help; if it fails with some other error then it's something else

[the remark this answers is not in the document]
narrieta: > if it fails with some other error then it's something else
you may want to measure whether the retry introduced another issue, but i wouldn't go there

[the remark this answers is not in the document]
larohra: I agree to the measurement but we should filter out ResourceGone/InvalidContainer in the query rather than out here in the source. Logging all error might help us investigate unknown issues later on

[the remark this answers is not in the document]
narrieta: if you want to measure failures communicating with the hostgaplugin then we need more instrumentation (e.g. failures in the first call, not only on the retry)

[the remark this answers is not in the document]
narrieta: btw, I am thinking these measurements should be only temporary --- we are measuring the effectiveness of specific fixes, and we may end up accummulating tons of those over time
we should have an offline talk about this

azurelinuxagent/common/protocol/metadata.py
pgombar: Exactly, my reasoning was to eliminate all the `*` imports in the files in was changing. It's scary.

azurelinuxagent/common/protocol/wire.py
pgombar: Thanks

azurelinuxagent/common/protocol/wire.py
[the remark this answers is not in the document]
pgombar: Agree with Norberto. Since the purpose of this event is to merely gather data about the fix for `ResourceGoneError` and `InvalidContainerError`, it makes sense to only emit events for those exceptions.

azurelinuxagent/common/protocol/wire.py
pgombar: I needed a try/catch to emit a telemetry event, but originally we weren't swallowing the exceptions from the retry call.

azurelinuxagent/common/datacontract.py
pgombar: Thanks

tests/ga/test_extension.py
pgombar: Added

azurelinuxagent/common/protocol/wire.py
narrieta: But it is not needed. We aren't doing anything in the exception handler, just re-raising the exception.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1024,22 +1031,58 @@ def send_request_using_appropriate_channel(self, direct_func, host_func):
...
+            try:
+                ret = host_func()
+                if ret:
+                    msg = "Request succeeded using the host plugin channel after goal state refresh." \
+                          "ContainerId changed from {0} to {1}, " \
+                          "role config file changed from {2} to {3}.".format(old_container_id, new_container_id,
+                                                                             old_role_config_name, new_role_config_name)
+                    add_event(name=AGENT_NAME,
+                              version=CURRENT_VERSION,
+                              op=WALAEventOperation.HostPlugin,
```
larohra: Are we gonna report the event always? Or is it only for monitoring this release?

pgombar: I'm going for always reporting. We were missing pieces of info before related to HostGAPlugin, with these events we'd have a more complete picture when investigating issues.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1024,22 +1031,58 @@ def send_request_using_appropriate_channel(self, direct_func, host_func):
...
+                              is_success=True,
+                              message=msg,
+                              log_event=False)
+
+            except (ResourceGoneError, InvalidContainerError) as e:
+                add_event(name=AGENT_NAME,
+                          version=CURRENT_VERSION,
+                          op=WALAEventOperation.HostPlugin,
+                          is_success=False,
+                          message=ustr(e),
```
narrieta: The message should be more explicit, something along the lines of the message at line 1059.
Also, probably we should not include the entire exception, but only its type (in case there are any secrets in the request)

pgombar: Good catch, thanks. Updated.

PR #1633, opened 2019-09-13
Add cloud-init auto-detect to prevent multiple provisioning mechanisms from relying on configuration for coordination

trstringer:
## Description

Currently waagent relies on explicit configuration (`Provisioning.UseCloudInit` to be set, and `Provisioning.Enabled` to be falsy). This can pose an interesting problem with the introduction of cloud-init. When waagent and cloud-init live side-by-side, they will contend with each other to be the provisioning agent. Image maintainers need to explicitly set configuration (above) so that waagent allows cloud-init to be the provisioning agent. With maintainers possibly not knowing about this behavior, they could unknowing cause provisioning issues.

This PR changes that behavior by deprecating `Provisioning.UseCloudInit` and `Provisioning.Enabled` configuration options and introduces a new configuration setting: `Provisioning.Agent`. The following values change the behavior of which provisioning agent to use:

* `auto` (default) - Allow waagent to autodetect to see if cloud-init is enabled. If so, use cloud-init as the provisioning agent. If not, use waagent as the provisioning agent.
* `waagent` - Force waagent as the provisioning agent.
* `cloud-init` - Force cloud-init as the provisioning agent.
* `disabled` - Disable all provisioning.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1633)


README.md
```diff
@@ -237,6 +237,14 @@ without the agent. In order to do that, the `provisionVMAgent` flag must be set
 provisioning time, via whichever API is being used. We will provide more details on
 this on our wiki when it is generally available. 
 
+#### __Provisioning.Force__
```
Moustafa-Moustafa: I think it's a bit confusing that waagent can be forced to provision but at the same time it's not enabled. In which case provisioning through cloud-init and waagent will be skipped.
One possible solution is to make the "Force=true" option overrides the "Enabled" option for waagent.

azurelinuxagent/pa/provision/cloudinit.py
```diff
@@ -132,3 +126,43 @@ def wait_for_ssh_host_key(self, max_retry=1800, sleep_time=1):
...
+    except dbus.DBusException as exc:
+        # If the unit does not exist, then take that as a falsy
+        # condition. We don't want to throw an exception in this
+        # case because waagent should be able to handle the situation
+        # where there is no cloud-init installed.
+        if exc.get_dbus_name() == 'org.freedesktop.DBus.Error.FileNotFound':
+            unit_is_enabled = False
+        else:
+            # Any other dbus exception should be bubbled up though.
+            logger.error('Unexpected dbus error: {}'.format(exc))
```
Moustafa-Moustafa: Why do we need to raise the error here and fail the provisioning. Can't we just consider this as if the cloud-init is not enabled and fall back to waagent ?

trstringer: That's exactly one of the changes I'm making now. I don't think any code path in this should fail provisioning at all. I think any exception is just a falsy return with no exception. Agreed.

README.md
trstringer: The more I think about this, the more I think that we are trying to force a bool where it doesn't belong. I think I'm more inclined to completely remove `Provisioning.Force` and deprecate `Provisioning.Enabled` in favor of a string type `Provisioning.Agent`, defaulting to `auto` but allowing you to specify `waagent` or `cloud-init`.

`auto` behavior could be auto detect, but otherwise it gives the user total control of which provisioning agent they could use which would act as an override. I think this is much more intuitive then trying to combine a few bools to get the desired effect.

What do you think?

Moustafa-Moustafa: I like the idea of "Provisioning.Agent". It's more intuitive and easier to grasp. Perhaps it can be added to the conf file with a comment of all the possible values.

trstringer:
Ready for review. // @jasonzio

azurelinuxagent/pa/provision/cloudinit.py
```diff
@@ -132,3 +126,29 @@ def wait_for_ssh_host_key(self, max_retry=1800, sleep_time=1):
...
+    Returns:
+        bool - True if cloud-init is enabled, False if otherwise.
+    """
+
+    try:
+        systemctl_output = subprocess.check_output([
+            'systemctl',
+            'is-enabled',
+            'cloud-init-local.service'
+        ], stderr=subprocess.STDOUT).decode('utf-8').replace('\n', '')
```
waldiTM: This will produce an exception if the service is missing.

I'm also not sure using `systemctl` is a particular good idea.

README.md
```diff
@@ -237,19 +236,32 @@ without the agent. In order to do that, the `provisionVMAgent` flag must be set
...
 agent. Valid values are "y" or "n". If provisioning is disabled, SSH host and
 user keys in the image are preserved and any configuration specified in the
 Azure provisioning API is ignored.
 
-#### __Provisioning.UseCloudInit__
-  
-_Type: Boolean_  
+_Note_: This configuration option has been deprecated and has no effect. waagent
+now auto-detects cloud-init as a provisioning agent (with an option to override
+with `Provisioning.Agent`).
```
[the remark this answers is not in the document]
trstringer: You're right, "deprecated" is probably not the best wording. I'll change that to "removed". As for old configs, in previous versions the old config options would still take effect. If they persist in newer versions, it won't break anything. They just would have no effect.

azurelinuxagent/pa/provision/cloudinit.py
trstringer: An exception is ok. This is wrapped in a try/except so that any exception case (such as a missing service) will just result in a falsy return. As for using `systemctl`, it is the best option forward from what I've seen. I originally tried using dbus but there is no guarantee that the system dbus is up and running at this point, whereas systemd has it's own internal dbus for communication (which systemctl uses) that we can rely on at this boot stage. Let me know if you have any other hesitations for using systemctl otherwise.

anhvoms:
LGTM

trstringer:
Thanks for the reviews, @jasonzio and @anhvoms. @narrieta, please hold off on merging this in just yet, I want to do a few more manual tests before we merge.

trstringer:
Validated with `canonical:ubuntuserver:18.04-lts:latest` and `credativ:debian:9:latest`.

@narrieta, let me know if there are any other things needed for this PR to get merged. Thanks!

README.md
```diff
@@ -237,19 +236,32 @@ without the agent. In order to do that, the `provisionVMAgent` flag must be set
...
 
-_Type: Boolean_  
+_Type: String_
+_Default: auto_
+
+Choose which provisioning agent to use (or allow waagent to figure it out by
+specifying "auto"). Possible options are "auto" (default), "waagent", "cloud-init",
+or "disabled".
+
+#### __Provisioning.Enabled__ (*removed in VERSION*)
```
narrieta: wouldn't removing this be a breaking change? who are the likely users of those settings? thanks

azurelinuxagent/common/conf.py
```diff
@@ -304,6 +295,14 @@ def get_password_cryptid(conf=__conf__):
     return conf.get("Provisioning.PasswordCryptId", "6")
 
 
+def get_provisioning_agent(conf=__conf__):
+    return conf.get("Provisioning.Agent", "auto")
+
+
+def get_provision_enabled(conf=__conf__):
```
narrieta: I'd add a short comment explaining that the 'enabled' setting got replace with 'agent'

README.md
jasonzio: These settings need to be made in the image from which a VM is created, and they're generally set by a user only when they create a new image. Simply updating waagent in an already-provisioned VM should have no impact, as the settings are ignored (since provisioning already happened).

I can imagine someone being impacted if they take an image they'd already created, update the waagent, and use that to create a new image. It might be worthwhile for the waagent package (deb or rpm) to have a post-install script check the config file for the presence of the `Provisioning.Enabled` config setting and emit a warning or error.

jasonzio: Might also be a good idea for the provisioning code in waagent to detect the presence of `Provisioning.Enabled` and emit a warning in the log (and via telemetry). This should make diagnosis easier in the few cases where the change gets missed.

trstringer:
@narrieta hold off on merging just yet, we're going to try to get another review of this before we move forward. Changing this to "work in progress" for the time being.

narrieta:
@trstringer OK

azurelinuxagent/pa/provision/cloudinit.py
rjschwei: I am not certain that calling systemctl from within a code path that runs as part of a systemd unit will always produce reliable results. I should but I don't know for certain nor have I found any documentation thereof.

This also makes the assumption that the generator (which is safe) has run, and properly detected Azure (there have been generator detection issues). Last but not least it assumes that all images run the generator that is part of cloud-init. There are other way to run `cloud-init-local.service`. While we can argue that that's not the "standard way" when/if someone has a "brilliant idea" and this code breaks I did want to point out at the implicit dependency that is being created by this check.

I am not advocating for a change, just making a comment.

azurelinuxagent/pa/provision/factory.py
```diff
@@ -16,19 +16,24 @@
...
                             distro_version=DISTRO_VERSION,
                             distro_full_name=DISTRO_FULL_NAME):
 
-    if conf.get_provision_cloudinit():
+    provisioning_agent = conf.get_provisioning_agent()
+
+    if provisioning_agent == 'cloud-init' or (
+            provisioning_agent == 'auto' and
+            cloud_init_is_enabled()):
+        logger.info('Using cloud-init for provisioning')
```
rjschwei: I don't think this is sufficient for testing, but I may be missing a detail about how we get here. Let's say the image is built with `auto`. When the instance is running the user goes and fiddles and sets the configuration to `cloud-init` or `waagent` and then reboots. Bang!

Since neither tool knows about the the others cached data the instance is going to be provisioned from scratch, i.e. new ssh keys, hostname, etc. From my perspective, albeit the user did something they shouldn't have, this will lead to a support call.

I think somewhere along the line we need to cache the configuration and then complain when it changes on an instance that gets rebooted. This could be handled by writing some data to `/var/run/azure` and both the agent and the Azure cloud-init data source check that location. Both complain respectively if the instance on boot is not in the expected state.

trstringer: Thanks for the review! This shouldn't be an issue, though, as the provisioning handler (both of them) check to see if provisioning is already complete. So the code path should short-circuit regardless in this case, no matter which provisioning agent originally did the provisioning work.

trstringer: Regarding the question about the safety of using systemctl in a unit, from my research this should be fine. systemctl talks to systemd through a private socket, `/run/systemd/private`. See `man init` for more information on the socket itself. And [/run is loaded early on in the boot process](https://www.freedesktop.org/wiki/Software/systemd/InitrdInterface/), so the accessibility of this socket should be guaranteed in our systemd unit.

trstringer:
@narrieta if this PR looks good it should be ready for merge.

narrieta:
running automation

trstringer:
@narrieta ready for re-running and consideration. Thanks!

narrieta:
@trstringer - Running automation

narrieta:
Automation OK

PR #1634, opened 2019-09-16
Extension update on failure return based reporting

larohra:
## Description
Currently the update sequence of an extension consists of the following operations -
Disable (OV) -> Update (NV) -> Uninstall (OV) -> Install if updateMode == UpdateWithInstall (NV) -> Enable (NV).

This PR addresses the issue where Extension Updates are blocked if there's any failure in the old version of the Extension. We've introduced a new flag 'continueOnUpdateFailure' which can be set in the new HandlerManifer of the extension to unblock the extension update.

Issue # 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1634)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -985,9 +1012,13 @@ def on_rmtree_error(_, __, exc_info):
...
         CGroupConfigurator.get_instance().remove_extension_cgroups(self.get_full_name())
 
-    def update(self, version=None):
+    def update(self, version=None, disable_failed=False):
         if version is None:
             version = self.ext_handler.properties.version
+        env = {'VERSION': version}
+
+        if disable_failed:
+            env.update({DISABLE_FAILED: "1"})
```
pgombar: Can we use `True` instead? A number might indicate we're counting the number of failures instead of setting a flag. Same comment applies for other assignments.

larohra: We specifically wanna use a number to avoid the confusion of case-sensitive value vs case-insensitive because you can only set a string value for environment variables.

PR #1635, opened 2019-09-16
Add telemetry event and log OS-specific info

pgombar:
On ext-handlers thread startup, report OS-specific parameters: agent service name, distro name and version, and which osutil class we're using based on the distro.

Added for monitoring the impact of #1580, hence creating a release-specific event to track this.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1635)


azurelinuxagent/ga/update.py
```diff
@@ -251,6 +251,15 @@ def run(self, debug=False):
             logger.info(u"Agent {0} is running as the goal state agent",
                         CURRENT_AGENT)
 
+            # Log OS-specific info, locally and as a telemetry event.
+            msg = u"Distro info: {0}:{1}, osutil class being used: {2}, " \
```
narrieta: Not sure osutil should be logged, but if it should probably it'd be better to do it in get_osutil

narrieta: could add the same in the daemon?

azurelinuxagent/ga/update.py
```diff
@@ -251,6 +251,15 @@ def run(self, debug=False):
...
+                  u"agent service name: {3}".format(DISTRO_NAME, DISTRO_VERSION,
+                                                    type(self.osutil).__name__, self.osutil.service_name)
+            add_event(AGENT_NAME,
```
narrieta: don't our telemetry events already have the OS info?

vrdmr: Yes - It is added in the [init_sysinfo](https://github.com/Azure/WALinuxAgent/blob/0105f5f4d45e051fdfc1bb78eeccb0d14967bcc6/azurelinuxagent/ga/monitor.py#L163).

azurelinuxagent/ga/update.py
vrdmr: I would also recommend adding other info - example:

```
osversion = "{0}:{1}-{2}-{3}:{4}".format(platform.system(),
                                                 DISTRO_NAME,
                                                 DISTRO_VERSION,
                                                 DISTRO_CODE_NAME,
                                                 platform.release())
```

pgombar: `get_osutil` is called all over the place, I don't think we want to emit an event every time it's called.

The purpose of this event was to confirm we were getting the correct data regarding distro and assert the changes made in #1580 were correct (hence the `osutil` parameter).

@vrdmr is correct, we are already logging distro-specific info so it might be better to make this a release-specific event then? Although I think it would be very useful to easily query Kusto for `agent service name` and see which distro uses which. It's something that required manual testing in the past.

azurelinuxagent/ga/update.py
pgombar: Correct. The purpose of this event was to assert that the correct `osutil` was created based on distro info.

PR #1637, opened 2019-09-17
Add telemetry event when cleaning up old cgroups

pgombar:
Adding a permanent telemetry event for changes made in #1595.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1637)

### DIFF
--- a/azurelinuxagent/common/cgroupapi.py
+++ b/azurelinuxagent/common/cgroupapi.py
@@ -132,9 +132,7 @@ class CGroupsApi(object):
         for controller in CGROUP_CONTROLLERS:
             try:
                 if controller not in mounted_controllers:
-                    msg = 'Cgroup controller "{0}" is not mounted. {1}'.format(controller, message)
-                    logger.warn(msg)
-                    errors.append(msg)
+                    logger.warn('Cgroup controller "{0}" is not mounted. {1}'.format(controller, message))
                 else:
                     operation(controller)
             except Exception as e:
--- a/azurelinuxagent/common/cgroupapi.py
+++ b/azurelinuxagent/common/cgroupapi.py
@@ -124,17 +124,24 @@ class CGroupsApi(object):
         """
         Executes the given operation on all controllers that need to be tracked; outputs 'message' if the controller
         is not mounted or if an error occurs in the operation
+        :return: Returns a list of error messages or an empty list if no errors occurred
         """
+        errors = []
         mounted_controllers = os.listdir(CGROUPS_FILE_SYSTEM_ROOT)
 
         for controller in CGROUP_CONTROLLERS:
             try:
                 if controller not in mounted_controllers:
-                    logger.warn('Cgroup controller "{0}" is not mounted. {1}'.format(controller, message))
+                    msg = 'Cgroup controller "{0}" is not mounted. {1}'.format(controller, message)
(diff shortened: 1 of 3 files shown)


azurelinuxagent/common/cgroupapi.py
```diff
@@ -124,17 +124,24 @@ def _foreach_controller(operation, message):
...
+        errors = []
         mounted_controllers = os.listdir(CGROUPS_FILE_SYSTEM_ROOT)
 
         for controller in CGROUP_CONTROLLERS:
             try:
                 if controller not in mounted_controllers:
-                    logger.warn('Cgroup controller "{0}" is not mounted. {1}'.format(controller, message))
+                    msg = 'Cgroup controller "{0}" is not mounted. {1}'.format(controller, message)
+                    logger.warn(msg)
+                    errors.append(msg)
```
narrieta: this shouldn't be flagged as an error

tests/common/test_cgroupapi.py
```diff
@@ -222,6 +223,12 @@ def test_cleanup_old_cgroups_should_move_daemon_pid_on_all_controllers(self):
         self.assertFalse(os.path.exists(old_cpu_cgroup))
         self.assertFalse(os.path.exists(old_memory_cgroup))
 
+        # Assert the event parameters that were sent out
+        _, kwargs = mock_add_event.call_args_list[0]
+        self.assertEquals(kwargs['op'], 'CGroupsCleanUp')
+        self.assertEquals(kwargs['is_success'], True)
+        self.assertEquals(kwargs['message'], 'Successfully cleaned up old cgroups in WALinuxAgent/WALinuxAgent.')
+
```
narrieta: can you also test the aggregation of errors?

azurelinuxagent/common/cgroupapi.py
pgombar: I had a dilemma whether to include that or not since it does point to a mismatch between the expected and actual cgroup state on disk. However, I agree it's not pertinent to the cgroup cleanup task. Will remove then.

narrieta: It is not a mismatch between the expected and actual states. This should have been once, at initialization time instead of each time we iterate over the controllers.

tests/common/test_cgroupapi.py
pgombar: Updated

tests/common/test_cgroupapi.py
```diff
@@ -222,6 +223,42 @@ def test_cleanup_old_cgroups_should_move_daemon_pid_on_all_controllers(self):
...
+    def test_cleanup_old_cgroups_should_fail_to_move_daemon_pid_on_all_controllers(self):
```
narrieta: In this naming convention the phrase after "should" should be the behavior that is being tested/asserted. In this case we are not testing that clean_old_cgroups fails to move the daemon pid on all controllers, but we are testing that it reports the errors from all controllers. A better name could be "test_cleanup_old_cgroups_should_report_errors_from_all_controllers_that_failed"

pgombar: Thanks, Norberto. This is what happens when you don't take time to think about naming :)

Issue #1479, opened 2019-03-05
FreeBSD 11.2 :  'cannot read route table [[Errno 2] No such file or directory: '/proc/net/route'


narrieta:
The freebsd based on the marketplace image is returning a large amount of errors such as the one below:
$ tail waagent.log
2019/03/04 13:51:35.200690 ERROR ExtHandler Cannot read route table [[Errno 2] No such file or directory: '/proc/net/route']
2019/03/04 13:52:35.312408 ERROR ExtHandler Cannot read route table [[Errno 2] No such file or directory: '/proc/net/route']
[...]

minusbat:
I looked at this, and started working on a fix for it actually. All it is trying to do is to read the routing table to see if it has changed. As such it should be enough to replace that code with the output of 'netstat -rn' I would thing. Just need to understand how to add a FreeBSd specific bit of code there in the way its done everywhere else. I don't do python really though (more of a C / Obj-C person)

mike-pt:
I looked into the same recently , cause without this working properly backups will never work.

The route table is easy to work around, I do know some python and I thing I've identified the 2 places this needs to change, but there's also the iptables side of it.

The main issue is that this is linux centric... Well its called "WALinuxAgent" not "WAUnix", there might be more places that need workarounds, but honestly other than adding if/else to cover both the linux and bsd cases doesn't seem to make much sense for me, I wish Microsoft could give BSD a bit more attention since they do support it on Azure and at least make sure basic things like backups work!

I would vote on having a WA BSD Agent that simply implement the bsd logic separate of this one.

onenachi:
Well I've tested and submitted a PR. Now its just a waiting game.

pgombar:
Closing as the PR got merged and will be included in the next release.

minusbat:
I just pulled master and installed this on FreeBSD 12.1 and I am still getting the original error:

2019/10/25 19:43:06.776594 ERROR ExtHandler Cannot read route table [[Errno 2] No such file or directory: '/proc/net/route']

Is there anything extra that I need to do to make this work ? I know the original PR was for FreeBSD 11, but presumably this should also work on 12 ?

onenachi:
This should just work on 12. I haven't tested it on 12, only on 11.2 and 11.3. But as long as the osutil used is the FreeBSD one, there should be no issues.

minusbat:
[...]

PR #1638, opened 2019-09-18
Fix /proc/net/route requirement that causes errors on FreeBSD

onenachi:
## Description

This is a fix for Issue https://github.com/Azure/WALinuxAgent/issues/1479).

What it does is:
- Instead of searching the linux specific `/proc/net/route` in wireserver_route_exists in dhcp.py, it calls the osutil.get_list_of_routes API and checks if the route destination matches the KNOWN_WIRESERVER_IP
- It then defines various route related functions in osutil/freebsd.py so that this works on FreeBSD
- One of these functions parses `netstat -rn` and creates RouteEntry objects that look similar enough to the Linux ones for the rest of the code to work

Issue #1479  

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1638)


onenachi:
https://github.com/Azure/WALinuxAgent/pull/1638/commits/1ee6e0152284b4cd6477e7b4d2122c23c75b3d40 is just a squish of all the other pull requests. It has all the changes.

larohra:
Thank you for your contribution @onenachi . Could you please add some test cases for the changes in freebsd.py file? We have a test file \tests\common\osutil\test_freebsd.py where you can add the specific tests for your changes to improve the code coverage. 

I've also initiated a run from our automated end-to-end testing just to ensure the changes are not breaking anything.

Apart from that the changes look good to me.

onenachi:
6c44645 is a squish of all the changes.

Unfortunately I did something stupid with 8c2077c because git thought there was a conflict and then undid it with 580e60b.

azurelinuxagent/common/osutil/freebsd.py
```diff
@@ -93,6 +97,192 @@ def get_if_mac(self, ifname):
...
+        """
+        Return a list of strings comprising the route table, including column headers. Each line is stripped of leading
+        or trailing whitespace but is otherwise unmolested.
+
+        :return: Entries in the route priority table from `netstat -rn`
+        :rtype: list(str)
+        """
+        linux_style_route_file = [ "Iface\tDestination\tGateway\tFlags\tRefCnt\tUse\tMetric\tMask\tMTU\tWindow\tIRTT" ]
+
+        cmd = "netstat -rn"
```
larohra: This function heavily relies on the current output format of this command, if somewhere in the future this format were to change, then this whole function would have to be changed.
Could you please refactor this long function into smaller private sub-functions with just one focus? It would be easier to read and understand and would be easy to modify in the future if needed

azurelinuxagent/common/osutil/freebsd.py
```diff
@@ -93,6 +97,192 @@ def get_if_mac(self, ifname):
...
+            # MTU
+            mtu = 0
+            # Window
+            window = 0
+            # Initial Round Trip Time
+            irtt = 0
+            # Add the route
+            linux_style_route_file.append("{0}\t{1}\t{2}\t{3}\t{4}\t{5}\t{6}\t{7}\t{8}\t{9}\t{10}".format(
+                    netif, dest, gw, flags, refcount, use, metric, mask, mtu, window, irtt))
+        return linux_style_route_file
```
larohra: The original implementation of this function handles all errors in function and doesn't propagate any exception further.
```python
 try:
            with open('/proc/net/route') as routing_table:
                return list(map(str.strip, routing_table.readlines()))
        except Exception as e:
            logger.error("Cannot read route table [{0}]", ustr(e))

        return []
```

After you refactor this function, could you please ensure that the `read_route_table` function doesn't raise any exception so as to not break the agent

azurelinuxagent/common/osutil/freebsd.py
larohra: Also if possible (and needed), please add more test cases for those specific methods

tests/common/osutil/test_freebsd.py
```diff
@@ -29,6 +31,103 @@ def tearDown(self):
     def test_get_dhcp_pid_should_return_a_list_of_pids(self):
         osutil_get_dhcp_pid_should_return_a_list_of_pids(self, FreeBSDOSUtil())
 
+    def test_empty_proc_net_route(self):
```
larohra: Thank you for adding the tests 👍

azurelinuxagent/common/osutil/freebsd.py
onenachi: That code doesn't depend on the output format of `nestat -rn` outside of it displaying columns for Netif, Destination, Gateway and Flags in any order similar to how _build_route_list in the default osutil parses a route in the Linux format from /proc/net/route on each iteration.

IMO it would make sense to keep this code together rather than to split it into sub functions because its essentially parsing one route in the FreeBSD netstat output and converting it to linux format on each iteration and that seems to mirror how _build_route_list reads the same for /proc/net/route on linux.

azurelinuxagent/common/osutil/freebsd.py
onenachi: done with 515b672

azurelinuxagent/common/osutil/freebsd.py
```diff
@@ -93,6 +97,192 @@ def get_if_mac(self, ifname):
...
+        cmd = "netstat -rn"
+        ret, netstat_output = shellutil.run_get_output(cmd)
+        if ret:
+            raise OSUtilError("Cannot read route table [{0}]".format(netstat_output))
+        netstat_output = netstat_output.split("\n")
+        if len(netstat_output) < 3:
+            return linux_style_route_file
+        netstat_output = netstat_output[3:]
+        # Parse the Netstat -RN header line
+        n_columns = 0
```
larohra: Could you add a comment with a sample output of `netstat -rn` is to make it easier for the reader to understand the parsing logic

azurelinuxagent/common/osutil/freebsd.py
larohra: I totally understand the functionality for this is to make the output similar to the output of /proc/net/route. My concern is that if in the future releases the header changes then we might need to modify the code.

The current code works, but is very big and unreadable unless you know what's actually happening in it. I feel if you refactor it into smaller chunks (like `_parse_netstat_header`, `_parse_routes`, `_get_dest_ip`, `_get_route_flags`, etc as per you see fit) then it makes it more readable and easily modifiable if needed in the future.

azurelinuxagent/common/osutil/freebsd.py
```diff
@@ -93,6 +97,192 @@ def get_if_mac(self, ifname):
...
 
+    @staticmethod
+    def read_route_table():
+        """
+        Return a list of strings comprising the route table, including column headers. Each line is stripped of leading
+        or trailing whitespace but is otherwise unmolested.
+
+        :return: Entries in the route priority table from `netstat -rn`
+        :rtype: list(str)
+        """
```
larohra: Could you please add an example of how the `netstat -rn` output differs with the linux based `proc/net/route` header just to make sure what this function is parsing

azurelinuxagent/common/osutil/freebsd.py
onenachi: Refactored as requested in e9d7a64

azurelinuxagent/common/osutil/freebsd.py
onenachi: Exmples added as requested in e9d7a64

azurelinuxagent/common/osutil/freebsd.py
onenachi: Exmples added as requested in e9d7a64

azurelinuxagent/common/dhcp.py
```diff
@@ -86,9 +86,8 @@ def wireserver_route_exists(self):
         route_exists = False
         logger.info("Test for route to {0}".format(KNOWN_WIRESERVER_IP))
         try:
-            route_file = '/proc/net/route'
-            if os.path.exists(route_file) and \
-                    KNOWN_WIRESERVER_IP_ENTRY in open(route_file).read():
+            route_table = self.osutil.read_route_table()
+            if any([(KNOWN_WIRESERVER_IP_ENTRY in route) for route in route_table]):
```
narrieta: In the original code there was a check for exists(route_file); if it was false, we would output a warning.
In the new code this would be reported as an error.

onenachi: I'm not sure how to work around this. read_route_table will log an error by default and then return an empty list if /proc/net/route doesn't exist.

The only thing it guarantees is not to raise an exception.

Im not sure changing that function would be appropriate. I do think using that function would be appropriate since checking route is clearly a platform specific task.

onenachi: I also do think that this is an edge case and not really an issue as it will only log an error on systems without /proc/net/route (non Linux systems) which have not overridden that function with their own osutil.

narrieta: ok

azurelinuxagent/common/osutil/freebsd.py
```diff
@@ -93,6 +97,306 @@ def get_if_mac(self, ifname):
...
+        def _get_netstat_rn_ipv4_routes():
+            """
+            Runs `netstat -rn -f inet` and parses its output and returns a list of routes where the key is the column name
+            and the value is the value in the column, stripped of leading and trailing whitespace.
+
+            :return: List of dictionaries representing routes in the ipv4 route priority list from `netstat -rn -f inet`
+            :rtype: list(dict)
+            """
+            cmd = "netstat -rn -f inet"
+            ret, output = shellutil.run_get_output(cmd)
```
narrieta: we are deprecating shellutil.run_get_output in favor of shellutil.run_command, could you change this?

onenachi: Done with a0112e2 and 72bddb7

azurelinuxagent/common/osutil/freebsd.py
```diff
@@ -93,6 +97,304 @@ def get_if_mac(self, ifname):
...
+        def _get_netstat_rn_ipv4_routes():
+            """
+            Runs `netstat -rn -f inet` and parses its output and returns a list of routes where the key is the column name
+            and the value is the value in the column, stripped of leading and trailing whitespace.
+
+            :return: List of dictionaries representing routes in the ipv4 route priority list from `netstat -rn -f inet`
+            :rtype: list(dict)
+            """
+            cmd = "netstat -rn -f inet"
+            output = shellutil.run_command(cmd, log_error=True)
```
narrieta: thank you for the update.

run_command set shell=False when invoking subprocess.Popen, so the command needs to be passed as an array of strings; see, for example https://github.com/Azure/WALinuxAgent/blob/develop/azurelinuxagent/common/utils/cryptutil.py#L55

Also, please be sure to tests this end-to-end on your side. Our test infrastructure does not validate FreeBSD.

Thanks

onenachi: Fixed in 08ba686

Tested in FreeBSD 11.2 on Azure:
image

narrieta:
LGTM; running automation - will approve and merge after that

narrieta:
Automation OK

PR #1640, opened 2019-09-20
Passing sequence number to all Extension Commands

larohra:
## Description
This PR is to address the behavior where the Guest Agent is supposed to pass the latest Sequence number from the goalstate to all extension commands

Issue # 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1640)

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -74,10 +74,13 @@ AGENT_STATUS_FILE = "waagent_status.json"
 
 NUMBER_OF_DOWNLOAD_RETRIES = 5
 
-DISABLE_FAILED = "AZURE_GUEST_AGENT_DISABLE_FAILED"
-UNINSTALL_FAILED = "AZURE_GUEST_AGENT_UNINSTALL_FAILED"
-EXTENSION_PATH = "AZURE_GUEST_AGENT_EXTENSION_PATH"
-EXTENSION_VERSION = "AZURE_GUEST_AGENT_EXTENSION_VERSION"
+
+class ExtCommandEnvVariable(object):
+    DisableFailed = "AZURE_GUEST_AGENT_DISABLE_FAILED"
+    UninstallFailed = "AZURE_GUEST_AGENT_UNINSTALL_FAILED"
+    ExtensionPath = "AZURE_GUEST_AGENT_EXTENSION_PATH"
+    ExtensionVersion = "AZURE_GUEST_AGENT_EXTENSION_VERSION"
+    ExtensionSeqNumber = "ConfigSequenceNumber"  # At par with Windows Guest Agent
 
 def get_traceback(e):
     if sys.version_info[0] == 3:
@@ -943,7 +946,7 @@ class ExtHandlerInstance(object):
     def enable(self, uninstall_failed=False):
         env = {}
         if uninstall_failed:
-            env.update({UNINSTALL_FAILED: '1'})
+            env.update({ExtCommandEnvVariable.UninstallFailed: '1'})
 
         self.set_operation(WALAEventOperation.Enable)
         man = self.load_manifest()
@@ -967,7 +970,7 @@ class ExtHandlerInstance(object):
     def install(self, uninstall_failed=False):
         env = {}
         if uninstall_failed:
-            env.update({UNINSTALL_FAILED: '1'})
+            env.update({ExtCommandEnvVariable.UninstallFailed: '1'})
 
         man = self.load_manifest()
         install_cmd = man.get_install_command()
@@ -1018,7 +1021,7 @@ class ExtHandlerInstance(object):
         env = {'VERSION': version}
 
         if disable_failed:
-            env.update({DISABLE_FAILED: "1"})
+            env.update({ExtCommandEnvVariable.DisableFailed: "1"})
 
         try:
             self.set_operation(WALAEventOperation.Update)
@@ -1207,8 +1210,9 @@ class ExtHandlerInstance(object):
                     env = {}
                 env.update(os.environ)
                 # Always add Extension Path and version to the current launch_command (Ask from publishers)
-                env.update({EXTENSION_PATH: self.get_base_dir(),
-                            EXTENSION_VERSION: self.ext_handler.properties.version})
+                env.update({ExtCommandEnvVariable.ExtensionPath: base_dir,
+                            ExtCommandEnvVariable.ExtensionVersion: self.ext_handler.properties.version,
+                            ExtCommandEnvVariable.ExtensionSeqNumber: str(self.get_seq_no())})
 
                 try:
                     # Some extensions erroneously begin cmd with a slash; don't interpret those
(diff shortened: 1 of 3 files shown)


tests/ga/test_exthandlers.py
```diff
@@ -231,6 +231,10 @@ def setUp(self):
         self.mock_get_log_dir = patch("azurelinuxagent.ga.exthandlers.ExtHandlerInstance.get_log_dir", lambda *_: self.log_dir)
         self.mock_get_log_dir.start()
 
+        mock_sleep = time.sleep
```
pgombar: Can you rebase to develop and remove the `SLEEP` mocking in the header of this file, if we're mocking it this way now? Thanks.

tests/ga/test_extension.py
```diff
@@ -1620,13 +1620,41 @@ def test_ext_path_and_version_env_variables_set_for_ever_operation(self, *args):
...
-                self.assertIn(EXTENSION_VERSION, kwargs['env'])
-                self.assertEqual("1.0.0", kwargs['env'][EXTENSION_VERSION])
+                self.assertIn(ExtCommandEnvVariable.ExtensionPath, kwargs['env'])
+                self.assertIn('OSTCExtensions.ExampleHandlerLinux-1.0.0', kwargs['env'][ExtCommandEnvVariable.ExtensionPath])
+                self.assertIn(ExtCommandEnvVariable.ExtensionVersion, kwargs['env'])
+                self.assertEqual("1.0.0", kwargs['env'][ExtCommandEnvVariable.ExtensionVersion])
 
         self._assert_handler_status(protocol.report_vm_status, "Ready", expected_ext_count=1, version="1.0.0")
 
+    def test_ext_sequence_no_should_set_for_ever_command_call(self, *args):
```
pgombar: Typo `ever` -> `every`? Also should it be `sequence_no_should_**be**_set` instead? Maybe `updated` bears more meaning since it's always set as an incrementally increasing value?

tests/ga/test_exthandlers.py
pgombar: Nevermind, wrong file.

tests/ga/test_extension.py
larohra: Thanks for the typo catch! We're not updating the seq no, just passing what we got in the goalstate to the extension commands

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1207,8 +1210,9 @@ def launch_command(self, cmd, timeout=300, extension_error_code=ExtensionErrorCo
                     env = {}
                 env.update(os.environ)
                 # Always add Extension Path and version to the current launch_command (Ask from publishers)
-                env.update({EXTENSION_PATH: self.get_base_dir(),
-                            EXTENSION_VERSION: self.ext_handler.properties.version})
+                env.update({ExtCommandEnvVariable.ExtensionPath: base_dir,
```
narrieta: Does Windows pass the sequence number to all commands? We create the settings file after install so it is meaningless in the install command

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1387,6 +1391,17 @@ def get_env_file(self):
     def get_log_dir(self):
         return os.path.join(conf.get_ext_log_dir(), self.ext_handler.name)
 
+    def get_seq_no(self):
+        runtime_settings = self.ext_handler.properties.extensions
+        # If no runtime_settings available for this ext_handler, then return 0 (this is the behavior we follow
+        # for update_settings)
```
narrieta: could we refactor this into a function (get_feault_seq_no?) and call it here and in update_settings?

azurelinuxagent/ga/exthandlers.py
larohra: Windows only sets it for Install and Enable. I didn't see the point of not passing it to the other states too (Disable/Uninstall/Update). 

We create the settings file right before install - 
```python
        if handler_state == ExtHandlerState.NotInstalled:
            ext_handler_i.set_handler_state(ExtHandlerState.NotInstalled)
            ext_handler_i.download()
            ext_handler_i.initialize()
            **ext_handler_i.update_settings()**
            if old_ext_handler_i is None:
                ***ext_handler_i.install()***
            elif ext_handler_i.version_ne(old_ext_handler_i):
                uninstall_failed = ExtHandlersHandler._update_extension_handler_and_return_if_failed(
                    old_ext_handler_i, ext_handler_i)
        else:
            ext_handler_i.update_settings()

        ext_handler_i.enable(uninstall_failed=uninstall_failed)
```

tests/ga/test_extension.py
```diff
@@ -1620,13 +1620,41 @@ def test_ext_path_and_version_env_variables_set_for_ever_operation(self, *args):
...
+                self.assertIn(ExtCommandEnvVariable.ExtensionVersion, kwargs['env'])
+                self.assertEqual("1.0.0", kwargs['env'][ExtCommandEnvVariable.ExtensionVersion])
 
         self._assert_handler_status(protocol.report_vm_status, "Ready", expected_ext_count=1, version="1.0.0")
 
+    def test_ext_sequence_no_should_be_set_for_every_command_call(self, *args):
+        test_data = WireProtocolData(DATA_FILE_MULTIPLE_EXT)
+        exthandlers_handler, protocol = self._create_mock(test_data, *args)
+
+        with patch.object(CGroupConfigurator.get_instance(), "start_extension_command") as patch_start_cmd:
```
narrieta: could we use an actual extension command to check the environment? or mock at the level of popen? this would make the test a little less dependent on implementation details

larohra: Yes makes sense, I'll add a test case for that too and leave this test as well (extra test cases never hurt)

azurelinuxagent/ga/exthandlers.py
narrieta: does windows also create the settings before install? when i wrote my extension settings were created only before enable. I am glad this has changed. I'd say it's ok to define for disable, etc.

tests/ga/test_extension.py
narrieta: ok, but then do mock at the level of Popen... I'm not convinced about taking a dependency on the fact we go thru the cgroup configurator to execute extensions. That code has already changed a couple of times.

azurelinuxagent/ga/exthandlers.py
larohra: You are correct, windows does it after install and before enable only

```c#
// Handler installation is over, now provide the plugin with runtime settings.
                    ProcessRuntimeSettings(data);
```

narrieta: but it passes the number to install, eh? oh, my! 

i'm ok with the behavior you defined for linux

azurelinuxagent/ga/exthandlers.py
narrieta: discussed offline; good as it is; my bad

tests/ga/test_extension.py

PR #1643, opened 2019-09-23
Fix flaky test for missing header

pgombar:
The elusive test came up again in this Travis run: https://travis-ci.org/Azure/WALinuxAgent/jobs/587997214

It intermittently fails because we compare two header values which contain an exact timestamp to the second. The mitigation is to remove the seconds from the timestamp (only in the scope of this test) to make the test more robust.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1643)

### DIFF
--- a/tests/protocol/test_hostplugin.py
+++ b/tests/protocol/test_hostplugin.py
@@ -473,29 +473,41 @@ class TestHostPlugin(AgentTestCase):
 
         exp_method = 'PUT'
         exp_url = hostplugin_status_url
-        exp_data = self._hostplugin_data(
-            status_blob.get_block_blob_headers(len(faux_status)),
-            bytearray(faux_status, encoding='utf-8'))
 
-        with patch.object(restutil, "http_request") as patch_http:
-            patch_http.return_value = Mock(status=httpclient.OK)
+        original_strftime = time.strftime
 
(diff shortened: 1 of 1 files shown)


tests/protocol/test_hostplugin.py
```diff
@@ -473,29 +473,41 @@ def test_validate_block_blob(self):
 
         exp_method = 'PUT'
         exp_url = hostplugin_status_url
-        exp_data = self._hostplugin_data(
-            status_blob.get_block_blob_headers(len(faux_status)),
-            bytearray(faux_status, encoding='utf-8'))
 
-        with patch.object(restutil, "http_request") as patch_http:
-            patch_http.return_value = Mock(status=httpclient.OK)
+        original_strftime = time.strftime
```
narrieta: but aren't all the tests that call _validate_hostplugin_args affected by the same issue?

narrieta: instead of mocking the formatting function, you could relax the comparison in _compare_data for this specific header

pgombar: You're correct, I was focusing on a single test that failed. I've updated the tests by removing the seconds from the timestamps when comparing in `_compare_data`.

PR #1646, opened 2019-09-25
Set env variable to return code for failure

larohra:
## Description
Currently during extension update if there is any problem in the Disable/Uninstall command of the prev version we include a boolean Environment variable to indicate the fact that there was an error. This PR adds the feature that instead of setting a boolean, we would now be setting the value of the environment variable to the exit code of the command.

Issue # 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1646)

### DIFF
--- a/azurelinuxagent/common/cgroupapi.py
+++ b/azurelinuxagent/common/cgroupapi.py
@@ -516,7 +516,7 @@ After=system-{1}.slice""".format(extension_name, EXTENSIONS_ROOT_CGROUP_NAME)
             else:
                 # There was an issue with systemd-run. We need to log it and retry the extension without systemd.
                 err_msg = 'Systemd process exited with code %s and output %s' % (e.exit_code, process_output) \
-                    if isinstance(e, ExtensionOperationError) else "Systemd Timed-out, output: %s" % process_output
+                    if isinstance(e, ExtensionOperationError) else "Systemd timed-out, output: %s" % process_output
                 add_event(AGENT_NAME,
                           version=CURRENT_VERSION,
                           op=WALAEventOperation.InvokeCommandUsingSystemd,
--- a/azurelinuxagent/common/cgroupapi.py
+++ b/azurelinuxagent/common/cgroupapi.py
@@ -515,15 +515,14 @@ After=system-{1}.slice""".format(extension_name, EXTENSIONS_ROOT_CGROUP_NAME)
                 raise
             else:
                 # There was an issue with systemd-run. We need to log it and retry the extension without systemd.
-                exit_code = e.exit_code if isinstance(e, ExtensionOperationError) else e.code
+                err_msg = 'Systemd process exited with code %s and output %s' % (e.exit_code, process_output) \
+                    if isinstance(e, ExtensionOperationError) else "Systemd Timed-out, output: %s" % process_output
                 add_event(AGENT_NAME,
                           version=CURRENT_VERSION,
                           op=WALAEventOperation.InvokeCommandUsingSystemd,
                           is_success=False,
-                          message='Failed to run systemd-run for unit {0}.scope. '
-                                  'Process exited with code {1} and output {2}'.format(scope_name,
-                                                                                       exit_code,
-                                                                                       process_output))
+                          message='Failed to run systemd-run for unit {0}.scope. {1}'
+                                  .format(scope_name, err_msg))
                 # Reset the stdout and stderr
                 stdout.truncate(0)
                 stderr.truncate(0)
--- a/azurelinuxagent/common/cgroupapi.py
+++ b/azurelinuxagent/common/cgroupapi.py
@@ -25,11 +25,11 @@ from azurelinuxagent.common.cgroup import CGroup
 from azurelinuxagent.common.cgroupstelemetry import CGroupsTelemetry
 from azurelinuxagent.common.conf import get_agent_pid_file_path
 from azurelinuxagent.common.event import add_event, WALAEventOperation
-from azurelinuxagent.common.exception import CGroupsException, ExtensionError, ExtensionErrorCodes
+from azurelinuxagent.common.exception import CGroupsException, ExtensionErrorCodes, ExtensionError, \
+    ExtensionOperationError
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.utils import fileutil, shellutil
-from azurelinuxagent.common.utils.extensionprocessutil import read_output, handle_process_completion, \
-                                                     wait_for_process_completion_or_timeout
+from azurelinuxagent.common.utils.extensionprocessutil import handle_process_completion, read_output
 from azurelinuxagent.common.version import AGENT_NAME, CURRENT_VERSION
 
 CGROUPS_FILE_SYSTEM_ROOT = '/sys/fs/cgroup'
@@ -497,36 +497,32 @@ After=system-{1}.slice""".format(extension_name, EXTENSIONS_ROOT_CGROUP_NAME)
         self.track_cgroups(extension_cgroups)
 
         # Wait for process completion or timeout
-        timed_out, return_code = wait_for_process_completion_or_timeout(process, timeout)
-        process_output = read_output(stdout, stderr)
-
-        if return_code == 0:
-            # The process terminated in time and successfully
-            return extension_cgroups, process_output
-        else:
+        try:
+            process_output = handle_process_completion(process=process,
+                                                       command=command,
+                                                       timeout=timeout,
+                                                       stdout=stdout,
+                                                       stderr=stderr,
+                                                       error_code=error_code)
+        except ExtensionError as e:
             # The extension didn't terminate successfully. Determine whether it was due to systemd errors or
             # extension errors.
+            process_output = read_output(stdout, stderr)
             systemd_failure = self._is_systemd_failure(scope_name, process_output)
 
             if not systemd_failure:
-                # There was an extension error; it either timed out or returned a non-zero exit code.
-                if timed_out:
-                    raise ExtensionError("Timeout({0}): {1}\n{2}".format(timeout, command, process_output),
-                                         code=ExtensionErrorCodes.PluginHandlerScriptTimedout)
-                else:
-                    raise ExtensionError("Non-zero exit code: {0}, {1}\n{2}".format(return_code,
-                                                                                    command,
-                                                                                    process_output),
-                                         code=error_code)
+                # There was an extension error; it either timed out or returned a non-zero exit code. Re-raise the error
+                raise
             else:
                 # There was an issue with systemd-run. We need to log it and retry the extension without systemd.
+                exit_code = e.exit_code if isinstance(e, ExtensionOperationError) else e.code
                 add_event(AGENT_NAME,
                           version=CURRENT_VERSION,
                           op=WALAEventOperation.InvokeCommandUsingSystemd,
                           is_success=False,
                           message='Failed to run systemd-run for unit {0}.scope. '
                                   'Process exited with code {1} and output {2}'.format(scope_name,
-                                                                                       return_code,
+                                                                                       exit_code,
                                                                                        process_output))
                 # Reset the stdout and stderr
                 stdout.truncate(0)
@@ -549,6 +545,9 @@ After=system-{1}.slice""".format(extension_name, EXTENSIONS_ROOT_CGROUP_NAME)
                                                            error_code=error_code)
 
                 return [], process_output
+
+        # The process terminated in time and successfully
+        return extension_cgroups, process_output
 
     def cleanup_old_cgroups(self):
         # No cleanup needed from the old daemon in the systemd case.
--- a/azurelinuxagent/common/utils/extensionprocessutil.py
+++ b/azurelinuxagent/common/utils/extensionprocessutil.py
@@ -17,7 +17,7 @@
 # Requires Python 2.6+ and Openssl 1.0+
 #
 
-from azurelinuxagent.common.exception import ExtensionError, ExtensionErrorCodes
+from azurelinuxagent.common.exception import ExtensionErrorCodes, LaunchCommandError
 from azurelinuxagent.common.future import ustr
 import os
 import signal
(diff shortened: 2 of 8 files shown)


azurelinuxagent/common/utils/extensionprocessutil.py
```diff
@@ -68,12 +68,13 @@ def handle_process_completion(process, command, timeout, stdout, stderr, error_c
...
-        raise ExtensionError("Timeout({0}): {1}\n{2}".format(timeout, command, process_output),
-                             code=ExtensionErrorCodes.PluginHandlerScriptTimedout)
+        raise LaunchCommandError("Timeout({0}): {1}\n{2}".format(timeout, command, process_output),
+                                 code=ExtensionErrorCodes.PluginHandlerScriptTimedout,
+                                 exit_code=ExtensionErrorCodes.PluginHandlerScriptTimedout)
 
     if return_code != 0:
-        raise ExtensionError("Non-zero exit code: {0}, {1}\n{2}".format(return_code, command, process_output),
-                             code=error_code)
+        raise LaunchCommandError("Non-zero exit code: {0}, {1}\n{2}".format(return_code, command, process_output),
```
pgombar: Need to change `ExtensionError` in `cgroupapi.py#start_extension_command` too?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -76,11 +76,14 @@
...
-    UninstallFailed = "AZURE_GUEST_AGENT_UNINSTALL_FAILED"
-    ExtensionPath = "AZURE_GUEST_AGENT_EXTENSION_PATH"
-    ExtensionVersion = "AZURE_GUEST_AGENT_EXTENSION_VERSION"
+    Prefix = "AZURE_GUEST_AGENT"
+    DisableReturnCode = "%s_DISABLE_CMD_EXIT_CODE" % Prefix
+    UninstallReturnCode = "%s_UNINSTALL_CMD_EXIT_CODE" % Prefix
+    ExtensionPath = "%s_EXTENSION_PATH" % Prefix
+    ExtensionVersion = "%s_EXTENSION_VERSION" % Prefix
     ExtensionSeqNumber = "ConfigSequenceNumber"  # At par with Windows Guest Agent
+    NotRun = "NotRun"
```
pgombar: Is this the default value? Can you add a comment here if it is?

tests/ga/test_extension.py
```diff
@@ -1464,7 +1462,7 @@ def test__old_handler_reports_failure_on_disable_fail_on_update(self, patch_get_
 
             self.assertFalse(old_version_kwargs['is_success'], "The last call to report event should be for a failure")
 
-            self.assertTrue(old_version_kwargs['message'].startswith('[ExtensionError]'), "No error reported")
+            self.assertTrue('Error' in old_version_kwargs['message'], "No error reported")
```
pgombar: Too loose check? Does `LaunchCommandError` show up?

azurelinuxagent/common/utils/extensionprocessutil.py
larohra: Thanks for catching it, totally missed this one! I was under the impression that we capture output only in 1 place!

azurelinuxagent/ga/exthandlers.py
larohra: It kind of is, whenever we call enable without running uninstall first, we set the env value to this.  It is to reduce ambiguity for the extension publisher where they dont have to guess if the current enable call is during an extension update case or just a normal call. Will add a comment

tests/ga/test_extension.py
larohra: I kind of wanted it to be loose, incase someone changes some implementation later on I dont want this test to be changed too. As long as there's error in the message I think its good enough for the test.

tests/ga/test_extension.py
```diff
@@ -931,7 +931,7 @@ def _helper_wait_for_handler_successful_completion(self, exthandlers_handler):
         exthandler.properties.extensions.append(extension)
 
         # Override the timeout value to minimize the test duration
-        wait_until = datetime.datetime.utcnow() + datetime.timedelta(seconds=5)
+        wait_until = datetime.datetime.utcnow() + datetime.timedelta(seconds=0.1)
```
larohra: Reduced the timeout time to make the tests faster

azurelinuxagent/common/utils/extensionprocessutil.py
```diff
@@ -68,12 +68,13 @@ def handle_process_completion(process, command, timeout, stdout, stderr, error_c
     process_output = read_output(stdout, stderr)
 
     if timed_out:
-        raise ExtensionError("Timeout({0}): {1}\n{2}".format(timeout, command, process_output),
-                             code=ExtensionErrorCodes.PluginHandlerScriptTimedout)
+        raise LaunchCommandError("Timeout({0}): {1}\n{2}".format(timeout, command, process_output),
```
narrieta: What is the intention of the new exception? How is it different to raise LaunchCommandError(code=ExtensionErrorCodes.PluginHandlerScriptTimedout) here vs ExtensionError(code=ExtensionErrorCodes.PluginHandlerScriptTimedout) @ SystemdCgroupsApi.start_command.

Also, when there is a timeout the exit code has no meaning.

If the intention was to capture the exit code of the extension then probably you could continue using ExtensionError(code=ExtensionErrorCodes.PluginHandlerScriptTimedout) for timeouts and rename the new exception to something like ExtensionCommanError and use it only for actual extension errors (which report an exit code).

Speaking of timeouts, it seems like we are raising timeout exceptions in multiple places. It may be better to raise the timeout from wait_for_process_completion_or_timeout (I would then rename it to wait_for_process_completion). That would match the design of Python's Subprocess.wait).

azurelinuxagent/common/utils/extensionprocessutil.py
narrieta: I see. My comment above still applies, though :)

azurelinuxagent/ga/exthandlers.py
narrieta: If it is a value rather than a variable name probably it should not be in ExtCommandEnvVariable

narrieta: >  It is to reduce ambiguity for the extension publisher where they dont have to guess if the current enable call is during an extension update case or just a normal call

do explain :)

larohra: >If it is a value rather than a variable name probably it should not be in ExtCommandEnvVariable
I created the `ExtCommandEnvVariable` class to contain all the environment variables in one place for better readability. I think since the `NotRun` value is set as an environment variable, makes sense to keep it there. 

 >do explain :)
It'll be easier to explain this offline. We can stay behind after scrum and I can explain the reasoning behind this

azurelinuxagent/common/utils/extensionprocessutil.py
larohra: > What is the intention of the new exception? How is it different to raise LaunchCommandError(code=ExtensionErrorCodes.PluginHandlerScriptTimedout) here vs ExtensionError(code=ExtensionErrorCodes.PluginHandlerScriptTimedout) @ SystemdCgroupsApi.start_command?

Yes the main reason for this exception was to differentiate between cases where the command execution failed with a non-0 exit code or if it failed due to some platform issue (unable to download handler manifest, etc).

> If the intention was to capture the exit code of the extension then probably you could continue using ExtensionError(code=ExtensionErrorCodes.PluginHandlerScriptTimedout) for timeouts and rename the new exception to something like ExtensionCommanError and use it only for actual extension errors (which report an exit code).

I agree, I'll modify the code accordingly. Also @vrdmr pointed out that we can reuse the `ExtensionOperationError` for this purpose as its pretty much being used for that. I'll see if I can reuse that error.

> It may be better to raise the timeout from wait_for_process_completion_or_timeout 

I totally agree to this, I was kind of under the impression we were already doing this. Let me refactor the code accordingly and modify the PR

azurelinuxagent/common/utils/extensionprocessutil.py
larohra: Totally agree to it, will modify the PR

azurelinuxagent/common/utils/extensionprocessutil.py
pgombar: `ExtensionOperationError` and `ExtensionError` are not the same. `ExtensionOperationError` indicates a failure to launch the extension and `ExtensionError` indicates an error during extension execution.

larohra: Sorry for the confusion, I meant `ExtensionOperationError` and `LaunchCommandError` look like they're there for similar purpose. I can add an extra property in the ExtensionOperationError if it doesn't break anything else.

pgombar: Also, the reason why we don't raise from `wait_for_process_completion_or_timeout` is because that would be too early in the systemd-run case (we want to determine if it was systemd that failed or the extension before raising `ExtensionError`). Check out `cgroupapi.py` starting from line 500.

pgombar: Just to keep in mind if you are to refactor this code. If you decide to raise from `wait_for_process_completion_or_timeout`, make sure to handle it appropriately in `cgroupapi.py#start_extension_command`.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1212,6 +1224,7 @@ def launch_command(self, cmd, timeout=300, extension_error_code=ExtensionErrorCo
                 # Always add Extension Path and version to the current launch_command (Ask from publishers)
                 env.update({ExtCommandEnvVariable.ExtensionPath: base_dir,
                             ExtCommandEnvVariable.ExtensionVersion: self.ext_handler.properties.version,
```
pgombar: Does this need to be cast to `str` too?

tests/common/test_cgroupapi.py
```diff
@@ -505,11 +510,21 @@ def mock_popen(*args, **kwargs):
...
+    @patch('time.sleep', side_effect=lambda _: mock_sleep)
+    def test_start_extension_command_should_use_fallback_option_if_systemd_times_out(self, _):
+        from azurelinuxagent.common.utils.extensionprocessutil import wait_for_process_completion_or_timeout
         # Mock systemd timeout and make sure the failure is only attributed to the extension if the command fails
         # using the fallback option
         original_popen = subprocess.Popen
+        mock_counter = Mock()
+        original_wait_for_completion = wait_for_process_completion_or_timeout
+
+        def mock_wait_for_process_completion(mock_counter, *args):
```
pgombar: Why was this mock needed instead of just patching the return value of `wait_for_process_completion_or_timeout`?

azurelinuxagent/ga/exthandlers.py
larohra: I tried to ensure that this is always a string but was unable to do so. I'll add a str() cast just in case. Thanks for the catch!

tests/ga/test_extension.py
```diff
@@ -1720,6 +1712,53 @@ def create_test_dir_and_script(base_dir, test_file_name, test_file):
                     self.assertIn("{0}={1}".format(ExtCommandEnvVariable.ExtensionSeqNumber, expected_seq_no),
                                   kwargs['message'])
 
+    def test_correct_exit_code_should_set_on_uninstall_cmd_failure(self, *args):
```
pgombar: `should_be_set`?

tests/common/test_cgroupapi.py
larohra: Initially you were just mocking the `wait_for_process_completion_or_timeout` call in this file (cgroupapi.py) so the 2nd call was going to the `azurelinuxagent.common.utils.extensionprocessutil.wait_for_process_completion_or_timeout` and helping your test. 
Now we needed to first return True for timeout in the first run and get the output of the command in the 2nd run. This is why I had to mock this function like this to make sure we follow both the cases.

tests/ga/test_extension.py
```diff
@@ -2308,34 +2343,149 @@ def test_failed_env_variables_should_be_set_from_within_extension_commands(self,
...
+            "uninstallCommand": test_failure_file_name})
+        new_handler_i = TestExtensionUpdateOnFailure._get_ext_handler_instance(
+            'foo', '1.0.1',
+            handler={"updateCommand": test_env_file_name + " -u", "installCommand": test_env_file_name + " -i"},
+            continue_on_update_failure=True
+        )
+
+        exit_code = 156
+        error_test_file = """
+            sleep 1m
```
pgombar: Will this actually sleep for 1 minute? If yes, can we make it shorter?

tests/ga/test_extension.py

tests/ga/test_extension.py
larohra: the script will sleep for 1m but our agent will throw a timeout exception long before that, that is the main test :)

tests/common/test_cgroupapi.py
larohra: Modified this test case, now inducing a timeout instead of mocking the return

azurelinuxagent/common/cgroupapi.py
```diff
@@ -497,36 +497,32 @@ def create_cgroup(controller):
...
                 # There was an issue with systemd-run. We need to log it and retry the extension without systemd.
+                exit_code = e.exit_code if isinstance(e, ExtensionOperationError) else e.code
                 add_event(AGENT_NAME,
                           version=CURRENT_VERSION,
                           op=WALAEventOperation.InvokeCommandUsingSystemd,
                           is_success=False,
                           message='Failed to run systemd-run for unit {0}.scope. '
                                   'Process exited with code {1} and output {2}'.format(scope_name,
-                                                                                       return_code,
+                                                                                       exit_code,
```
narrieta: but if the exception is not an ExtensionOperationError, then exit_code is not really an exit code... maybe you need to improve the message

narrieta: the idea would be to differentiate timeouts from actual errors

larohra: Makes sense, updated the code

azurelinuxagent/common/cgroupapi.py
```diff
@@ -497,37 +497,32 @@ def create_cgroup(controller):
...
-                    raise ExtensionError("Non-zero exit code: {0}, {1}\n{2}".format(return_code,
-                                                                                    command,
-                                                                                    process_output),
-                                         code=error_code)
+                # There was an extension error; it either timed out or returned a non-zero exit code. Re-raise the error
+                raise
             else:
                 # There was an issue with systemd-run. We need to log it and retry the extension without systemd.
+                err_msg = 'Systemd process exited with code %s and output %s' % (e.exit_code, process_output) \
+                    if isinstance(e, ExtensionOperationError) else "Systemd Timed-out, output: %s" % process_output
```
pgombar: nit: unnecessary capitalization in Timed-out

larohra: corrected :)

PR #1651, opened 2019-10-01
Start sending PerformanceCounter metrics for Cgroups

vrdmr:
## Description
Adding changes to start sending PerformanceCounter metrics as a part of our telemetry.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1651)


tests/ga/test_monitor.py
```diff
@@ -521,36 +534,82 @@ def test_send_extension_metrics_telemetry_for_empty_cgroup(self, patch_report_al
...
+        ext_handler.properties = ext_handler_properties
+        ext_handler_instance = ExtHandlerInstance(ext_handler=ext_handler, protocol=None)
+        ext_handler_instance.set_operation("Enable")
 
         monitor_handler = get_monitor_handler()
         monitor_handler.init_protocols()
         monitor_handler.last_cgroup_polling_telemetry = datetime.datetime.utcnow() - timedelta(hours=1)
         monitor_handler.last_cgroup_report_telemetry = datetime.datetime.utcnow() - timedelta(hours=1)
+
+        command = self.create_script("keep_cpu_busy_and_consume_memory_for_5_seconds", '''
```
pgombar: Is it 3 seconds or 5?

tests/ga/test_monitor.py
```diff
@@ -561,6 +620,80 @@ def test_send_extension_metrics_telemetry_with_actual_cgroup(self, patch_add_eve
...
+        ext_handler.properties = ext_handler_properties
+        ext_handler_instance = ExtHandlerInstance(ext_handler=ext_handler, protocol=None)
+        ext_handler_instance.set_operation("Enable")
+
+        monitor_handler = get_monitor_handler()
+        monitor_handler.init_protocols()
+        monitor_handler.last_cgroup_polling_telemetry = datetime.datetime.utcnow() - timedelta(hours=1)
+        monitor_handler.last_cgroup_report_telemetry = datetime.datetime.utcnow() - timedelta(hours=1)
+
+        command = self.create_script("keep_cpu_busy_and_consume_memory_for_5_seconds", '''
```
pgombar: Same comment, outdated name?

tests/ga/test_monitor.py
```diff
@@ -521,36 +534,82 @@ def test_send_extension_metrics_telemetry_for_empty_cgroup(self, patch_report_al
...
+                ext_handler_instance.launch_command(command)
+
+        self.assertTrue(CGroupsTelemetry.is_tracked(os.path.join(
+            BASE_CGROUPS, "cpu", "walinuxagent.extensions", "foobar_1.0.0")))
+        self.assertTrue(CGroupsTelemetry.is_tracked(os.path.join(
+            BASE_CGROUPS, "memory", "walinuxagent.extensions", "foobar_1.0.0")))
+
+        for i in range(max_num_polls):
+            metrics = CGroupsTelemetry.poll_all_tracked()
+            time.sleep(0.1)
```
pgombar: Why is this sleep needed?

tests/ga/test_monitor.py
```diff
@@ -561,6 +620,80 @@ def test_send_extension_metrics_telemetry_with_actual_cgroup(self, patch_add_eve
...
```
pgombar: Same comment.

tests/common/test_event.py
```diff
@@ -385,19 +385,49 @@ def test_elapsed_milliseconds(self):
...
+                    self.assertEqual(i["value"], "DummyCounter")
+                    counter_present = True
+                if i["name"] == "Instance":
+                    self.assertEqual(i["value"], "DummyInstance")
+                    instance_present = True
+                if i["name"] == "Value":
+                    self.assertEqual(i["value"], 100)
+                    value_present = True
+            
+            self.assertTrue(category_present & counter_present & instance_present & value_present)
```
pgombar: `&` is a bitwise operator in Python, you need `and`.

azurelinuxagent/common/event.py
```diff
@@ -37,6 +37,8 @@
 
 _EVENT_MSG = "Event: name={0}, op={1}, message={2}, duration={3}"
 TELEMETRY_EVENT_PROVIDER_ID = "69B669B9-4AF8-4C50-BDC4-6006FA76E975"
+TELEMETRY_METRICS_PROVIDER_ID = "69B669B9-4AF8-4C50-BDC4-6006FA76E975"
```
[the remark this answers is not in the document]
vrdmr: Because they are sending to two different ids (and in the future, we might send it to different locations (perf and events at different places). But yes, I can push that task into the future and remove the redundant constant.

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -114,18 +115,40 @@ def report_all_tracked():
...
+                        CGroupsTelemetry._cgroup_metrics[cgroup.name].add_memory_usage(current_memory_usage)
+                        metrics.append(MetricValue("Memory", "Total Memory Usage", cgroup.name, current_memory_usage))
+
+                        max_memory_usage = cgroup.get_max_memory_usage()
+                        CGroupsTelemetry._cgroup_metrics[cgroup.name].add_max_memory_usage(max_memory_usage)
+                        metrics.append(MetricValue("Memory", "Max Memory Usage", cgroup.name, max_memory_usage))
+                    else:
+                        raise CGroupsException('CGroup controller {0} is not supported'.format(cgroup.controller))
+                except Exception as e:
+                    if not isinstance(e, (IOError, OSError)) or e.errno != errno.ENOENT:
```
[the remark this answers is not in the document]
vrdmr: Yup. My motivation for the condition was to log if any non-IOError arrises and even if an IOError arrises, it should not be e.errno = 2.

What I did was just the boolean algebra (_DeMorgan's Law_) on 
`not (isinstance(e, (IOError, OSError)) and e.errno == errno.ENOENT)`, taking in the `not` inside, changing the `and` to `or` to come up with 
`not isinstance(e, (IOError, OSError) or not (e.errno == errno.ENOENT)`
simplified to
`not isinstance(e, (IOError, OSError) or e.errno != errno.ENOENT`

Thus, with this logic, the above does work: 
`if not isinstance(e, (IOError, OSError)) or e.errno != errno.ENOENT (2)`

Test cases:
- randomError - `not isinstance(e, (IOError, OSError))` = *True* (_log_)
- IOError (errno = 10) - `not isinstance(e, (IOError, OSError))` = *False* || `e.errno != errno.ENOENT` = *True* (_log_)
- IOError (errno = 2) - `not isinstance(e, (IOError, OSError))` = *False* || `e.errno != errno.ENOENT` = *False* (_don't log_)

tests/ga/test_monitor.py
vrdmr: Good catch. Formatted it for consistency.

tests/ga/test_monitor.py
vrdmr: Not necessary now. Removing it. Thanks for the catch.

tests/ga/test_monitor.py
vrdmr: Not necessary now. Removing it. Thanks for the catch.

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -114,18 +115,40 @@ def report_all_tracked():
...
+                    elif cgroup.controller == "memory":
+                        current_memory_usage = cgroup.get_memory_usage()
+                        CGroupsTelemetry._cgroup_metrics[cgroup.name].add_memory_usage(current_memory_usage)
+                        metrics.append(MetricValue("Memory", "Total Memory Usage", cgroup.name, current_memory_usage))
+
+                        max_memory_usage = cgroup.get_max_memory_usage()
+                        CGroupsTelemetry._cgroup_metrics[cgroup.name].add_max_memory_usage(max_memory_usage)
+                        metrics.append(MetricValue("Memory", "Max Memory Usage", cgroup.name, max_memory_usage))
+                    else:
+                        raise CGroupsException('CGroup controller {0} is not supported'.format(cgroup.controller))
```
narrieta: log a periodic ERROR instead?

azurelinuxagent/common/cgroupstelemetry.py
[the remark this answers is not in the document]
narrieta: could you add a few comments explaining when IOError/OSError happen?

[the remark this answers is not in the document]
vrdmr: Done.

tests/ga/test_monitor.py
```diff
@@ -663,36 +681,128 @@ def test_send_extension_metrics_telemetry_for_empty_cgroup(self, patch_report_al
         self.assertEqual(1, patch_poll_all_tracked.call_count)
         self.assertEqual(1, patch_report_all_tracked.call_count)
         self.assertEqual(0, patch_add_event.call_count)
+        self.assertEqual(0, patch_add_metric.call_count)
+        monitor_handler.stop()
+
+    @patch("azurelinuxagent.common.cgroup.MemoryCgroup.get_memory_usage")
+    @patch('azurelinuxagent.common.logger.Logger.periodic_warn')
+    def test_send_extension_metrics_telemetry_handling_memory_cgroup_exceptions(self, patch_periodic_warn,
```
pgombar: Can you use the same name template for this test and the next one?

tests/ga/test_monitor.py
```diff
@@ -663,36 +681,128 @@ def test_send_extension_metrics_telemetry_for_empty_cgroup(self, patch_report_al
...
+        ioerror.errno = 2
+        patch_get_memory_usage.side_effect = ioerror
+
+        CGroupsTelemetry._tracked.append(CpuCgroup("cgroup_name", "/test/path"))
+
+        monitor_handler = get_monitor_handler()
+        monitor_handler.init_protocols()
+        monitor_handler.last_cgroup_polling_telemetry = datetime.datetime.utcnow() - timedelta(hours=1)
+        monitor_handler.last_cgroup_report_telemetry = datetime.datetime.utcnow() - timedelta(hours=1)
+        monitor_handler.poll_telemetry_metrics()
```
pgombar: Wouldn't we want to assert that the returned metrics from here are empty too?

tests/ga/test_monitor.py
```diff
@@ -663,36 +681,128 @@ def test_send_extension_metrics_telemetry_for_empty_cgroup(self, patch_report_al
...
+        ioerror.errno = 2
+        patch_cpu_usage.side_effect = ioerror
+
+        CGroupsTelemetry._tracked.append(CpuCgroup("cgroup_name", "/test/path"))
+
+        monitor_handler = get_monitor_handler()
+        monitor_handler.init_protocols()
+        monitor_handler.last_cgroup_polling_telemetry = datetime.datetime.utcnow() - timedelta(hours=1)
+        monitor_handler.last_cgroup_report_telemetry = datetime.datetime.utcnow() - timedelta(hours=1)
+        monitor_handler.poll_telemetry_metrics()
```
pgombar: Wouldn't we want to assert that the returned metrics from here are empty too?

tests/ga/test_monitor.py
```diff
@@ -663,36 +681,128 @@ def test_send_extension_metrics_telemetry_for_empty_cgroup(self, patch_report_al
...
+
+    @patch('azurelinuxagent.common.logger.Logger.periodic_warn')
+    def test_send_extension_metrics_telemetry_for_unsupported_cgroup(self, patch_periodic_warn, *args):
+        CGroupsTelemetry._tracked.append(CGroup("cgroup_name", "/test/path", "io"))
+
+        monitor_handler = get_monitor_handler()
+        monitor_handler.init_protocols()
+        monitor_handler.last_cgroup_polling_telemetry = datetime.datetime.utcnow() - timedelta(hours=1)
+        monitor_handler.last_cgroup_report_telemetry = datetime.datetime.utcnow() - timedelta(hours=1)
+        monitor_handler.poll_telemetry_metrics()
```
pgombar: Wouldn't we want to assert that the returned metrics from here are empty too?

narrieta:
@vrdmr - please let me know when you've completed end-to-end testing

tests/ga/test_monitor.py
```diff
@@ -16,20 +16,24 @@
...
 from azurelinuxagent.common.protocol.restapi import VMInfo
 from azurelinuxagent.common.protocol.wire import WireProtocol
 from azurelinuxagent.common.utils import restutil
 from azurelinuxagent.common.version import AGENT_VERSION
+from azurelinuxagent.ga.exthandlers import ExtHandlerInstance
 from azurelinuxagent.ga.monitor import *
-from tests.common.test_cgroupstelemetry import make_new_cgroup, consume_cpu_time, consume_memory
+from tests.common.test_cgroupstelemetry import make_new_cgroup
 from tests.protocol.mockwiredata import WireProtocolData, DATA_FILE
 from tests.tools import *
```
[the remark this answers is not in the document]
vrdmr: Thanks. Did the import * fix in #1663. (Already changed test_event and test_logger's headers in there)
https://github.com/Azure/WALinuxAgent/pull/1663/commits/a5e1c821488bee7f8f24b9a82323bfa8db761bf9#diff-7db4a9e1a11b00430fb0e3ed0da31370R17-R49

tests/ga/test_monitor.py
```diff
@@ -16,20 +16,24 @@
...
-from azurelinuxagent.common.protocol.imds import ComputeInfo, IMDS_IMAGE_ORIGIN_ENDORSED
+from azurelinuxagent.common.osutil.default import BASE_CGROUPS
+from azurelinuxagent.common.protocol.imds import ComputeInfo
+from azurelinuxagent.common.protocol.restapi import ExtHandlerProperties, ExtHandler
 from azurelinuxagent.common.protocol.restapi import VMInfo
 from azurelinuxagent.common.protocol.wire import WireProtocol
 from azurelinuxagent.common.utils import restutil
 from azurelinuxagent.common.version import AGENT_VERSION
+from azurelinuxagent.ga.exthandlers import ExtHandlerInstance
 from azurelinuxagent.ga.monitor import *
```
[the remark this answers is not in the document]
vrdmr: Thanks. Did the import * fix in #1663. (Already changed test_event and test_logger's headers in there)
https://github.com/Azure/WALinuxAgent/pull/1663/commits/a5e1c821488bee7f8f24b9a82323bfa8db761bf9#diff-7db4a9e1a11b00430fb0e3ed0da31370R17-R49

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -112,18 +113,46 @@ def report_all_tracked():
...
+                        CGroupsTelemetry._cgroup_metrics[cgroup.name].add_memory_usage(current_memory_usage)
+                        metrics.append(MetricValue("Memory", "Total Memory Usage", cgroup.name, current_memory_usage))
+
+                        max_memory_usage = cgroup.get_max_memory_usage()
+                        CGroupsTelemetry._cgroup_metrics[cgroup.name].add_max_memory_usage(max_memory_usage)
+                        metrics.append(MetricValue("Memory", "Max Memory Usage", cgroup.name, max_memory_usage))
+                    else:
+                        msg = 'CGroup controller {0} is not supported for cgroup {1}'.format(cgroup.controller,
+                                                                                             cgroup.name)
+                        logger.periodic_warn(logger.EVERY_HOUR, msg)
```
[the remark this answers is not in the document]
vrdmr: True. Sorry. Didn't see that case. Removed re-logging stuff.

tests/common/test_event.py
```diff
@@ -385,19 +385,49 @@ def test_elapsed_milliseconds(self):
...
-        self.assertEqual(event_dictionary['providerId'], "69B669B9-4AF8-4C50-BDC4-6006FA76E975")
+        self.assertEqual(event_dictionary['providerId'], event.TELEMETRY_EVENT_PROVIDER_ID)
         for parameter in event_dictionary["parameters"]:
             if parameter['name'] == 'Counter':
                 self.assertEqual(parameter['value'], '%idle')
                 break
         else:
             self.fail("Counter '%idle' not found in event parameters: {0}".format(repr(event_dictionary)))
+
+    def test_save_event(self):
```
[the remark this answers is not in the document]
vrdmr: Done.

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -112,18 +113,46 @@ def report_all_tracked():
...
+                    elif cgroup.controller == "memory":
+                        current_memory_usage = cgroup.get_memory_usage()
+                        CGroupsTelemetry._cgroup_metrics[cgroup.name].add_memory_usage(current_memory_usage)
+                        metrics.append(MetricValue("Memory", "Total Memory Usage", cgroup.name, current_memory_usage))
+
+                        max_memory_usage = cgroup.get_max_memory_usage()
+                        CGroupsTelemetry._cgroup_metrics[cgroup.name].add_max_memory_usage(max_memory_usage)
+                        metrics.append(MetricValue("Memory", "Max Memory Usage", cgroup.name, max_memory_usage))
+                    else:
+                        msg = 'CGroup controller {0} is not supported for cgroup {1}'.format(cgroup.controller,
```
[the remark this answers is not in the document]
vrdmr: No need to add this. Once the CGroupException gets generated, the log would be generated. And it is with [PERIODIC] message added.

tests/ga/test_monitor.py
vrdmr: The metrics are returned by poll_all_tracked (From CGroupsTelemetry), and not poll_telemetry_metrics. And thus it won't be returning anything here. 

Had an offline discussion and added newer asserts (to verify sending events is not happening when there's an exception)

tests/ga/test_monitor.py
vrdmr: The metrics are returned by poll_all_tracked (From CGroupsTelemetry), and not poll_telemetry_metrics. And thus it won't be returning anything here. 

Had an offline discussion and added newer asserts (to verify sending events is not happening when there's an exception)

tests/ga/test_monitor.py
vrdmr: The metrics are returned by poll_all_tracked (From CGroupsTelemetry), and not poll_telemetry_metrics. And thus it won't be returning anything here. 

Had an offline discussion and added newer asserts (to verify sending events is not happening when there's an exception)

azurelinuxagent/ga/monitor.py
```diff
@@ -177,8 +177,8 @@ def init_sysinfo(self):
         try:
             ram = self.osutil.get_total_mem()
             processors = self.osutil.get_processor_cores()
-            self.sysinfo.append(TelemetryEventParam("RAM", ram))
-            self.sysinfo.append(TelemetryEventParam("Processors", processors))
+            self.sysinfo.append(TelemetryEventParam("RAM", int(ram)))
```
pgombar: Shouldn't we be respecting the schema in Kusto?


```
cluster("Rdos").database("rdos").GuestAgentExtensionEvents
| take 1
| project RAM, Processors, ImageOrigin | getschema 


ColumnName   | ColumnOrdinal  | DataType       | ColumnType
RAM          | 0              | System.Int64   | long
Processors   | 1              | System.Int64   | long
ImageOrigin  | 2              | System.String  | string


```

vrdmr: We translate the types in [wire.py](https://github.com/Azure/WALinuxAgent/blob/develop/azurelinuxagent/common/protocol/wire.py#L507-L523) while generating the event.

```
def event_param_to_v1(param):
    param_format = '<Param Name="{0}" Value={1} T="{2}" />'
    param_type = type(param.value)
    attr_type = ""
    if param_type is int:
        attr_type = 'mt:uint64'
    elif param_type is str:
        attr_type = 'mt:wstr'
    elif ustr(param_type).count("'unicode'") > 0:
        attr_type = 'mt:wstr'
    elif param_type is bool:
        attr_type = 'mt:bool'
    elif param_type is float:
        attr_type = 'mt:float64'
    return param_format.format(param.name,
                               saxutils.quoteattr(ustr(param.value)),
                               attr_type)
```

[In the example XML generated](https://github.com/Azure/WALinuxAgent/blob/develop/tests/ga/test_monitor.py#L433-L434) - :

```
'<Param Name="RAM" Value="10000" T="mt:uint64" />' \
'<Param Name="Processors" Value="4" T="mt:uint64" />' \
```

tests/ga/test_monitor.py
```diff
@@ -832,26 +842,33 @@ def test_send_extension_metrics_telemetry_with_actual_cgroup(self, patch_add_eve
...
-        ext_handler_instance = ExtHandlerInstance(ext_handler=ext_handler, protocol=None)
-        ext_handler_instance.set_operation("Enable")
-
-        monitor_handler = get_monitor_handler()
-        monitor_handler.init_protocols()
-        monitor_handler.last_cgroup_polling_telemetry = datetime.datetime.utcnow() - timedelta(hours=1)
-        monitor_handler.last_cgroup_report_telemetry = datetime.datetime.utcnow() - timedelta(hours=1)
-
-        command = self.create_script("keep_cpu_busy_and_consume_memory_for_{0}_seconds".format(time_to_wait), '''
+        # This test has some timing issues when systemd is managing cgroups, so we force the file system API
```
narrieta: how did all those timing issues surfaced?

azurelinuxagent/ga/monitor.py
```diff
@@ -445,7 +445,7 @@ def poll_telemetry_metrics(self):
 
             if metrics:
                 for metric in metrics:
-                    report_metric(metric.category, metric.counter, metric.instance, metric.value)
+                    report_metric(metric.category, metric.counter, metric.instance, float(metric.value))
```
narrieta: shouldn't we make the conversion when we initialize metric.value?

tests/ga/test_monitor.py
vrdmr: Initially, the systemd way of creating cgroup would not create a new cgroup when testing and I was seeing memory data (0s) and no CPU data (we throw an exception), which was messing up the data collection and the intended purpose of this test. Fixing it to FileSystem API helped me to run a 3-second running test and helped the data for the test.

azurelinuxagent/ga/monitor.py
vrdmr: The reasoning of casting the value here was to cast is at the time of writing the actual metric, as the count of sources adding metrics could change and it might miss to add the casting.

In `poll_all_tracked` or in other places in the future, where the metrics are populated, the number of metrics can change, which is why I added `float` casting here. I'll actually move it to the `report_metric` method itself, such that the data is always consistent and we don't hit schema mismatch issues later.

Eg: Didn't want to cast it at individual locations and do it at one place.
```
if cgroup.controller == "cpu":
   ...
    metrics.append(MetricValue("Process", "% Processor Time", cgroup.name, current_cpu_usage))
   ...
elif cgroup.controller == "memory":
   ...
   metrics.append(MetricValue("Memory", "Total Memory Usage", cgroup.name, current_memory_usage))
   ...
   metrics.append(MetricValue("Memory", "Max Memory Usage", cgroup.name, max_memory_usage))
   ...
```

azurelinuxagent/ga/monitor.py
```diff
@@ -178,8 +179,8 @@ def init_sysinfo(self):
         try:
             ram = self.osutil.get_total_mem()
             processors = self.osutil.get_processor_cores()
-            self.sysinfo.append(TelemetryEventParam("RAM", ram))
-            self.sysinfo.append(TelemetryEventParam("Processors", processors))
+            self.sysinfo.append(TelemetryEventParam("RAM", int(ram)))
```
narrieta: shouldn't the conversion be done by get_total_mem?

azurelinuxagent/ga/monitor.py
```diff
@@ -178,8 +179,8 @@ def init_sysinfo(self):
...
+            self.sysinfo.append(TelemetryEventParam("Processors", int(processors)))
```
narrieta: do the conversion in get_processor_cores?

azurelinuxagent/ga/monitor.py
```diff
@@ -207,7 +208,7 @@ def init_sysinfo(self):
             self.sysinfo.append(TelemetryEventParam('VMId',
                                                     vminfo.vmId))
             self.sysinfo.append(TelemetryEventParam('ImageOrigin',
-                                                    vminfo.image_origin))
+                                                    int(vminfo.image_origin)))
```
narrieta: do the conversion at the point where image_origin is initialized?

tests/common/test_cgroupstelemetry.py
```diff
@@ -32,6 +32,13 @@
                         are_cgroups_enabled, is_trusty_in_travis, i_am_root
 
 
+def raise_ioerror(*args):
```
narrieta: *_ since args is not referenced

tests/ga/test_extension.py
```diff
@@ -2218,38 +2218,50 @@ def test_ext_handler_with_cgroup_enabled(self, *args):
...
+    @patch('azurelinuxagent.common.event.EventLogger.add_metric')
     @patch('azurelinuxagent.common.event.EventLogger.add_event')
     @attr('requires_sudo')
-    def test_ext_handler_and_monitor_handler_with_cgroup_enabled(self, patch_add_event, *args):
+    def test_ext_handler_and_monitor_handler_with_cgroup_enabled(self, patch_add_event, patch_add_metric, *args):
         self.assertTrue(i_am_root(), "Test does not run when non-root")
 
-        test_data = WireProtocolData(DATA_FILE)
-        exthandlers_handler, monitor_handler, protocol = self._create_mock(test_data, *args)
+        # This test has some timing issues when systemd is managing cgroups, so we force the file system API
```
narrieta: I added this comment/code to 1 single test as a temporary workaround, but now the same hack is being propagated to other tests... can we handle this better?

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -112,18 +113,44 @@ def report_all_tracked():
 
     @staticmethod
     def poll_all_tracked():
+        metrics = []
+        MetricValue = namedtuple('Metric', ['category', 'counter', 'instance', 'value'])
```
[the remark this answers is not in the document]
vrdmr: Good idea. It could be used externally and in tests. I'll refactor it.

azurelinuxagent/ga/monitor.py
```diff
@@ -473,9 +474,13 @@ def poll_telemetry_metrics(self):
 
         if time_now >= (self.last_cgroup_polling_telemetry +
                         MonitorHandler.CGROUP_TELEMETRY_POLLING_PERIOD):
-            CGroupsTelemetry.poll_all_tracked()
+            metrics = CGroupsTelemetry.poll_all_tracked()
             self.last_cgroup_polling_telemetry = time_now
 
+            if metrics:
+                for metric in metrics:
+                    report_metric(metric.category, metric.counter, metric.instance, metric.value)
```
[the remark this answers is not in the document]
vrdmr: Done.

azurelinuxagent/ga/monitor.py
vrdmr: Similar to the comment above for `get_total_mem`. Only for schema for an event rather than changing the behavior for all the users of this method.

azurelinuxagent/ga/monitor.py
vrdmr: The casting is done for the purposes of solving current and potential schema issues. I am keeping the original type of `get_total_mem` in case we have any other uses which require its origin type (float/int, depending on the platform).

azurelinuxagent/ga/monitor.py
vrdmr: The values are already INT. The casting at the event level is making sure that the event does not evert get some ambiguous values in the future.

```
IMDS_IMAGE_ORIGIN_UNKNOWN = 0
IMDS_IMAGE_ORIGIN_CUSTOM = 1
IMDS_IMAGE_ORIGIN_ENDORSED = 2
IMDS_IMAGE_ORIGIN_PLATFORM = 3
```

tests/ga/test_extension.py
vrdmr: I'll sync on this offline. Thanks.

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -112,18 +113,44 @@ def report_all_tracked():
...
+                    # every occurrences of such case as it would be very verbose. We do want to log all the other
+                    # exceptions which could occur, which is why we do a periodic log for all the other errors.
+                    if not isinstance(e, (IOError, OSError)) or e.errno != errno.ENOENT:
+                        logger.periodic_warn(logger.EVERY_HOUR, '[PERIODIC] Could not collect metrics for cgroup '
+                                                                '{0}. Error : {1}'.format(cgroup.name, ustr(e)))
                 if not cgroup.is_active():
                     CGroupsTelemetry.stop_tracking(cgroup)
                     CGroupsTelemetry._cgroup_metrics[cgroup.name].marked_for_delete = True
 
+        return metrics
```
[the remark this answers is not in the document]
vrdmr: The `is_active` check being done later helped us get one-time running extensions earlier. But with transient scope, the cgroups would also go away once it is done. Thus, yes the `is_active` check can be done earlier.

I'll send a separate PR for that change. Don't want to extend this PR's scope.

tests/ga/test_monitor.py
narrieta: OK. Collection of cgroup data is always done using the file system, even under systemd. Instead of adding this hack to each of these tests you may simple execute them explicitly using FileSystemCgroupsApi. See, for example, the tests for the cgroup configurator.

tests/ga/test_extension.py
narrieta: Added a suggestion above

azurelinuxagent/common/cgroupstelemetry.py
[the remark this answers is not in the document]
narrieta: ok

azurelinuxagent/ga/monitor.py
narrieta: OK. Shouldn't then those casts also be done in add_default_parameters_to_event, _add_event, etc.

azurelinuxagent/common/event.py
```diff
@@ -412,7 +414,12 @@ def report_metric(category, counter, instance, value, log_event=False, reporter=
         message = "Metric {0}/{1} [{2}] = {3}".format(category, counter, instance, value)
         _log_event(AGENT_NAME, "METRIC", message, 0)
         return
-    reporter.add_metric(category, counter, instance, value, log_event)
+    try:
+        reporter.add_metric(category, counter, instance, float(value), log_event)
+    except ValueError:
+        message = "{0}/{1} [{2}] = {3}".format(category, counter, instance, value)
```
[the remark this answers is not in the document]
vrdmr: if the `value` passed cannot be cast. This is just a precaution to make sure that we are strictly enforcing the float and incorrect values passed should not go through.

azurelinuxagent/ga/monitor.py
vrdmr: Yes. Outside the purview of this PR and I'll validate if it requires any changes and send another PR for that if needed.

vrdmr: Did the change for PerfCounters and Add_Event. Will make the change for Add_log_event in the send logs PR.

tests/ga/test_extension.py
```diff
@@ -2218,38 +2218,50 @@ def test_ext_handler_with_cgroup_enabled(self, *args):
...
+        # by creating a new instance of the CGroupConfigurator
+        with patch("azurelinuxagent.common.cgroupapi.CGroupsApi._is_systemd", return_value=False):
+            cgroup_configurator_instance = CGroupConfigurator._instance
```
narrieta: since CGroupConfigurator is a singleton this may affect other tests. That was the reason I used try/finally initially. A cleaner approach would be to mock CGroupConfigurator.getinstance() (force it to return a new instance) in this single test.

narrieta: (a few more instances in the other tests)

PR #1654, opened 2019-10-03
Add support for Gen2 VM resource disks

trstringer:
## Description

The current implementation for resource disks expects (requires) that the device ID has a prefix of `00000000-0001`. For Gen2 VMs, this is not the case. They have a device ID of `f8b3781a-1e82-4818-a1c3-63d806ec15bb`. Because of this, prior to this PR waagent does not find the resource disk and therefore doesn't mount it.

Here is the output of `lsblk`:

```
    NAME   MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
    sda      8:0    0   30G  0 disk
    ├─sda1   8:1    0    2M  0 part
    ├─sda2   8:2    0  512M  0 part /boot/efi
    ├─sda3   8:3    0    1G  0 part /boot
    └─sda4   8:4    0 28.5G  0 part /
    sdb      8:16   0   64G  0 disk
    └─sdb1   8:17   0   64G  0 part
    sr0     11:0    1  628K  0 rom
```

As you can see, there is no mountpoint for `sdb1`.

This PR fixes that by looking first for a Gen1 VM resource disk device, and then trying a Gen2 VM resource disk device.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1654)

### DIFF
--- a/azurelinuxagent/common/osutil/debian.py
+++ b/azurelinuxagent/common/osutil/debian.py
@@ -33,10 +33,10 @@ import azurelinuxagent.common.utils.textutil as textutil
 from azurelinuxagent.common.osutil.default import DefaultOSUtil
 
 
-class DebianOSUtil(DefaultOSUtil):
+class DebianOSBaseUtil(DefaultOSUtil):
 
     def __init__(self):
-        super(DebianOSUtil, self).__init__()
+        super(DebianOSBaseUtil, self).__init__()
         self.jit_enabled = True
 
     def restart_ssh_service(self):
@@ -61,10 +61,10 @@ class DebianOSUtil(DefaultOSUtil):
         return self.get_endpoint_from_leases_path('/var/lib/dhcp/dhclient.*.leases')
 
 
-class DebianOS8Util(DebianOSUtil):
+class DebianOSModernUtil(DebianOSBaseUtil):
 
     def __init__(self):
-        super(DebianOS8Util, self).__init__()
+        super(DebianOSModernUtil, self).__init__()
         self.jit_enabled = True
         self.service_name = self.get_service_name()
 
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -1221,9 +1221,9 @@ class DefaultOSUtil(object):
                                 # current device. But for Gen2 VMs all of the disks (sda, sdb,
                                 # sr0) are presented in this device on the same SCSI controller.
                                 # Because of that we need to also read the LUN. It will be:
-                                #   0 - OS disk (sda)
-                                #   1 - Resource disk (sdb)
-                                #   2 - CDROM (sr0)
+                                #   0 - OS disk
+                                #   1 - Resource disk
+                                #   2 - CDROM
                                 if root_path_parts[-1] == 'block' and (
                                         device_id_or_prefix != gen2_device_id or
                                         root_path_parts[-2].split(':')[-1] == '1'):
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -97,6 +97,7 @@ IP_COMMAND_OUTPUT = re.compile('^\d+:\s+(\w+):\s+(.*)$')
 
 BASE_CGROUPS = '/sys/fs/cgroup'
 
+STORAGE_DEVICE_PATH = '/sys/bus/vmbus/devices/'
 GEN2_DEVICE_ID = 'f8b3781a-1e82-4818-a1c3-63d806ec15bb'
 
 class DefaultOSUtil(object):
@@ -1193,6 +1194,24 @@ class DefaultOSUtil(object):
         return None
 
     @staticmethod
+    def _enumerate_device_id():
+        """
+        Enumerate all storage device IDs.
+
+        Args:
+            None
+
+        Returns:
+            Iterator[Tuple[str, str]]: VmBus and storage devices.
+        """
+
+        if os.path.exists(STORAGE_DEVICE_PATH):
+            for vmbus in os.listdir(STORAGE_DEVICE_PATH):
+                deviceid = fileutil.read_file(os.path.join(STORAGE_DEVICE_PATH, vmbus, "device_id"))
+                guid = deviceid.strip('{}\n')
+                yield vmbus, guid
+
+    @staticmethod
     def search_for_resource_disk(gen1_device_prefix, gen2_device_id):
         """
         Search the filesystem for a device by ID or prefix.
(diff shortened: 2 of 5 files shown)


azurelinuxagent/common/osutil/default.py
```diff
@@ -1203,25 +1203,43 @@ def device_for_ide_port(self, port_id):
...
+                            for root, dirs, files in os.walk(path + vmbus):
+                                root_path_parts = root.split('/')
+                                # For Gen1 VMs we only have to check for the block dir in the
+                                # current device. But for Gen2 VMs all of the disks (sda, sdb,
+                                # sr0) are presented in this device on the same SCSI controller.
+                                # Because of that we need to also read the LUN. It will be:
+                                #   0 - OS disk
+                                #   1 - Resource disk
+                                #   2 - CDROM
+                                if root_path_parts[-1] == 'block' and (
```
vrdmr: Can you please add a unit test to mock the success and failure conditions, such as this code is covered?

trstringer:
@jasonzio can I get a review as well? Thanks!

azurelinuxagent/common/osutil/default.py
```diff
@@ -1203,25 +1204,42 @@ def device_for_ide_port(self, port_id):
...
+                                        if ':' in d and "block" == d.split(':')[0]:
+                                            device = d.split(':')[1]
+                                            break
+                            break
+                    if device:
                         break
-            except OSError as oe:
-                logger.warn('Could not obtain device for IDE port {0}: {1}', port_id, ustr(oe))
+                except OSError as oe:
+                    logger.warn('Could not obtain device for IDE port {0}: {1}', port_id, ustr(oe))
```
jasonzio: Please refactor this into a helper function to reduce indent levels and simplify understanding of the top-level code. This also isolates gen detection and handling so that, when Gen3 is (inevitably, someday) introduced, it'll be easier and cleaner to handle.

azurelinuxagent/common/osutil/factory.py
```diff
@@ -90,7 +90,7 @@ def _get_osutil(distro_name, distro_code_name, distro_version, distro_full_name)
             return SUSEOSUtil()
 
     if distro_name == "debian":
-        if Version(distro_version) > Version("7"):
+        if "sid" in distro_version or Version(distro_version) > Version("7"):
             return DebianOS8Util()
         else:
             return DebianOSUtil()
```
jasonzio: Given that we're now supporting up through Debian 10, I suspect this logic needs to be re-examined. You may want to rename `DebianOSUtil()` to `DebianOSOldUtil()` and `DebianOS8Util()` to `DebianOSModernUtil()` or something similar, just to make clear what's going on here.

azurelinuxagent/common/osutil/default.py
trstringer: This comment caused me to think a little more about my approach. I have refactored it again so that we now only make a single pass on the vmbus dir, so this should be more efficient. It is also greatly simplified, and the indentation and function remains very similar in structure and functionality (with additional gen2 handling, of course) than from before this code change.

Let me know if you think this works, but it should satisfy your recommendation.

azurelinuxagent/common/osutil/factory.py
trstringer: I like that idea. The reason I made this change is because I use Debian testing for my development environment. I think my preference would be more for:

`DebianOSUtil` -> `DebianOSOldUtil` and `DebianOS8Util` -> `DebianOSUtil`. What do you think about that?

trstringer: Actually on second thought, `DebianOSModernUtil` seems fine to me.

trstringer: I ended up doing `DebianOSUtil` -> `DebianOSBaseUtil`, as it seemed more natural with the relationship between `DebianOSBaseUtil` and `DebianOSModernUtil`.

vrdmr:
@trstringer - FYI - The [Py2 builds](https://travis-ci.org/Azure/WALinuxAgent/builds/593324763?utm_source=github_status&utm_medium=notification) are failing. Could you please take a look?

trstringer:
@vrdmr Yep I'll check that out and get it fixed up.

azurelinuxagent/common/osutil/debian.py
```diff
@@ -33,10 +33,10 @@
 from azurelinuxagent.common.osutil.default import DefaultOSUtil
 
 
-class DebianOSUtil(DefaultOSUtil):
+class DebianOSBaseUtil(DefaultOSUtil):
 
     def __init__(self):
-        super(DebianOSUtil, self).__init__()
+        super(DebianOSBaseUtil, self).__init__()
```
jasonzio: OK, I have to blow the whistle here. This use of super() is completely wrong-headed. I know you didn't introduce it, but it's just plain wrong.

If the class hierarchy is always single-inheritance, and I'm pretty sure it is, the correct way to invoke the superclass constructor is explicitly, by it's name.
```
class DebianOSBaseUtil(DefaultOSUtil):
    def __init__(self):
        DebianOSUtil.__init__()
```
If `DebianOSUtil` inherits from a superclass, then `DebianOSUtil.__init__()` should explicitly invoke the __init__ method in its own superclass.

I recognize I just asked for a change in every osutil class. However, this abuse/misuse of `super()` is pretty egregious. Had this stuff been done correctly the first time, renaming the classes would have required two mouseclicks in PyCharm.

azurelinuxagent/common/osutil/debian.py
```diff
@@ -61,10 +61,10 @@ def get_dhcp_lease_endpoint(self):
         return self.get_endpoint_from_leases_path('/var/lib/dhcp/dhclient.*.leases')
 
 
-class DebianOS8Util(DebianOSUtil):
+class DebianOSModernUtil(DebianOSBaseUtil):
 
     def __init__(self):
-        super(DebianOS8Util, self).__init__()
+        super(DebianOSModernUtil, self).__init__()
```
jasonzio: Same problem: this is terrible use of `super()`. Try this:
```
class DebianOSModernUtil(DebianOSBaseUtil):
    def __init__(self):
        DebianOSBaseUtil.__init__()
        self.jit_enabled = True
```

azurelinuxagent/common/osutil/default.py
jasonzio: I still think the inner code should be refactored to a separate method. Even if it only gets called once, from a maintenance perspective, it's better to pull it out. Also, if gen1 and gen2 behaviors diverge more, or if some future gen3 has even more divergent behavior here, you'll want those methods separated into helper functions.

azurelinuxagent/common/osutil/debian.py
trstringer: Let's create an issue and address this in a separate PR to keep this PR more focused on gen2 resource disk fixing.

azurelinuxagent/common/osutil/debian.py
trstringer: Let's push this issue to another unit of work to keep this PR focused on gen2 resource disk.

azurelinuxagent/common/osutil/default.py
trstringer: Ok, I pulled that code out to a separate method. Thanks!

azurelinuxagent/common/osutil/default.py
```diff
@@ -1234,7 +1236,25 @@ def device_for_ide_port(self, port_id):
                                         break
                         break
             except (OSError, IOError) as exc:
-                logger.warn('Could not obtain device for IDE port {0}: {1}', port_id, ustr(exc))
+                logger.warn('Error gettin device for {0} or {1}: {2}', gen1_device_prefix, gen2_device_id, ustr(exc))
```
jasonzio: Typo - "getting"

trstringer: Good catch :) Fixed.

azurelinuxagent/common/osutil/default.py
```diff
@@ -1192,26 +1192,28 @@ def get_mount_point(self, mountlist, device):
...
+
         device = None
         path = "/sys/bus/vmbus/devices/"
         # We have to try device IDs for both Gen1 and Gen2 VMs.
-        gen1_device_id_prefix = '{0}-000{1}'.format(g0, port_id)
         if os.path.exists(path):
             try:
                 for vmbus in os.listdir(path):
                     deviceid = fileutil.read_file(os.path.join(path, vmbus, "device_id"))
                     guid = deviceid.strip('{}')
```
jasonzio: In my testing, there were newlines at the end of the data returned from read_file; do you need to add `\n` to the `strip()` call?

jasonzio: Refactor to a generator. Something like this:
```
def enumerate_device_guids():
    path = "/sys/bus/vmbus/devices/"
    if os.path.exists(path):
        for vmbus in os.listdir(path):
            deviceid = fileutil.read_file(os.path.join(path, vmbus, "device_id"))
            guid = deviceid.strip('{}\n')
            yield guid
    raise StopIteration
```
The code then turns into something like:
```
try:
    for guid in enumerate_device_guids():
        if guid.startswith() etc
```

jasonzio: Code golf (just for your amusement):
```
def edg2():
    path = "/sys/bus/vmbus/devices/"
    if os.path.exists(path):
        return (fileutil.read_file(os.path.join(path, vmbus, "device_id")).strip('{}\n') for vmbus in os.listdir(path))
    return list()
```
Returns an iterable. If path exists, it's a generator; if path doesn't exist, it's a list. Either way, the caller just iterates.

I wouldn't suggest putting that in the code, though.

azurelinuxagent/common/osutil/default.py
```diff
@@ -1191,6 +1193,67 @@ def get_mount_point(self, mountlist, device):
...
+        """
+
+        if os.path.exists(STORAGE_DEVICE_PATH):
+            for vmbus in os.listdir(STORAGE_DEVICE_PATH):
+                deviceid = fileutil.read_file(os.path.join(STORAGE_DEVICE_PATH, vmbus, "device_id"))
+                guid = deviceid.strip('{}\n')
+                yield vmbus, guid
+
+    @staticmethod
+    def search_for_resource_disk(gen1_device_prefix, gen2_device_id):
```
anhvoms: >def search_for_resource_disk(gen1_device_prefix, gen2_device_id): [](start = 4, length = 65)

Is there an easy way to detect if this is a Gen2VM, if so would it make more sense to refactor this into a flow where we detect whether this is a Gen2 or Gen1, if it's Gen2 there is a different search_for_resource_disk function (sort of like function pointers, one for Gen1, one for Gen2).

trstringer: Good question, I'll see if I can find that answer.

trstringer:
@narrieta let me know if you anything else from my end to get this PR merged in. Thanks!

narrieta:
Automation OK

azurelinuxagent/common/osutil/default.py
lixuemin2016: Not sure you still need it, just for your reference.
 if [ -d /sys/firmware/efi/ ]; then
        os_GENERATION=2
    else
        os_GENERATION=1
    fi

Issue #233, opened 2016-06-15
[2.1] get_protocol vs get_protocol_by_file


ahmetalpbalkan:
In the daemon code [there is a `clear_protocol` call](https://github.com/Azure/WALinuxAgent/blob/e877916550b8634ff3bc406362d4e8c18c32a311/azurelinuxagent/daemon/main.py#L107) which just [deletes a file called `Protocol`](https://github.com/Azure/WALinuxAgent/blob/e877916550b8634ff3bc406362d4e8c18c32a311/azurelinuxagent/common/protocol/util.py#L209-L222). (stores either `WireProtocol` or `MetadataProtocol`).

Then the [`get_protocol`](https://github.com/Azure/WALinuxAgent/blob/e877916550b8634ff3bc406362d4e8c18c32a311/azurelinuxagent/common/protocol/util.py#L223-L249) and [`get_protocol_by_file`](https://github.com/Azure/WALinuxAgent/blob/e877916550b8634ff3bc406362d4e8c18c32a311/azurelinuxagent/common/protocol/util.py#L252-L284) refer to this file to pick the protocol.

I guess my question is
- if this is a cache, why are we [clearing](https://github.com/Azure/WALinuxAgent/blob/e877916550b8634ff3bc406362d4e8c18c32a311/azurelinuxagent/daemon/main.py#L107) it everytime agent restarts?
- why is this on a file; and not an in-memory variable?
- why does only [provisioning code use `get_protocol_by_file`](https://github.com/Azure/WALinuxAgent/blob/e877916550b8634ff3bc406362d4e8c18c32a311/azurelinuxagent/pa/provision/default.py#L61) and in 8 other places still use `get_protocol`?

😕

brendandixon:
@ahmetalpbalkan All likely remnants of the on-prem work being new. I do not know if an existing VHD image can be moved between cloud and on-prem. If so, it may be possible for it begin in one place then be rebooted in another. That would explain the conservative approach, though a file, in that case, is overkill *except* for its diagnostic advantage. And yes, our calls should be consistent: There should be just one call and *it* should figure out what to do.

PR #1656, opened 2019-10-06
Make obtaining wireserver endpoint consistent

jacwil:
## Description
There are current several consumers (direct and indirect) of the Wireserver ip address. Some have the default value hardcoded. Some obtain it when the is_dhcp_available method returns False. In one case, a consumer expects get_wireserver_endpoint method to return None in a scenario to fallback to the default, known ip address.

This PR removes the default ip address from the return of is_dhcp_available and has get_wireserver_endpoint return it when reading from the file fails, which is the case when the file does not exist.

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1656)


jacwil:
Relates to #233

azurelinuxagent/common/protocol/util.py
```diff
@@ -147,32 +149,86 @@ def _get_protocol_file_path(self):
...
+            if os.path.isfile(file_path):
+                try:
+                    self.endpoint = fileutil.read_file(file_path)
+
+                    if self.endpoint:
+                        logger.info("WireServer endpoint {0} read from file", self.endpoint)
+                        return self.endpoint
+
+                    logger.error("[GetWireserverEndpoint] Unexpected empty file {0}: {1}", file_path, str(e))
+                except IOError as e:
```
pgombar: Due to Python versions handling file exceptions differently, you should be catching both `IOError` and `OSError` here.

azurelinuxagent/common/protocol/util.py
```diff
@@ -147,32 +149,86 @@ def _get_protocol_file_path(self):
...
+
+        try:
+            self.endpoint = None
+            endpoint_file_path = self._get_wireserver_endpoint_file_path()
+            if not os.path.isfile(endpoint_file_path):
+                return
+
+            try:
+                os.remove(endpoint_file_path)
+            except IOError as e:
```
pgombar: Same comment for `OSError` and `IOError` here.

azurelinuxagent/common/protocol/util.py
```diff
@@ -147,32 +149,86 @@ def _get_protocol_file_path(self):
...
+            endpoint_file_path = self._get_wireserver_endpoint_file_path()
+            if not os.path.isfile(endpoint_file_path):
+                return
+
+            try:
+                os.remove(endpoint_file_path)
+            except (IOError, OSError) as e:
+                # Ignore file-not-found errors (since the file is being removed)
+                if e.errno == errno.ENOENT:
+                    return
```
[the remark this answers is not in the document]
jacwil: The finally block will execute when losing scope, which will release the lock. If this doesn't happen, Python has earned itself a life-long evangelical hater.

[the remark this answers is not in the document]
pgombar: You're indeed right (https://stackoverflow.com/questions/19805654/python-try-finally-block-returns). TIL!

azurelinuxagent/common/protocol/util.py
```diff
@@ -254,13 +305,14 @@ def _save_protocol(self, protocol_name):
...
+        except (IOError, OSError) as e:
             logger.error("Failed to save protocol endpoint: {0}", e)
 
     def clear_protocol(self):
         """
-        Cleanup previous saved endpoint.
+        Cleanup previous saved protocol endpoint.
         """
+        self._clear_wireserver_endpoint()
         logger.info("Clean protocol")
```
vrdmr: Is this log needed here now, with the `logger.info("Clean wireserver endpoint")` log added in `self._clear_wireserver_endpoint()`?

narrieta:
LGTM; will merge after current release.

azurelinuxagent/common/protocol/util.py
jacwil: Since nothing is expected to call _clear_wireserver_endpoint() directly, I'll make a single log entry stating both variables are getting cleared. Thank you for highlighting the excessive log entries. :)

PR #1659, opened 2019-10-08
Start sending log events to our telemetry

vrdmr:
## Description
The goal of this PR is to enable Telemetry channel for logs >= WARNING.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1659)


azurelinuxagent/common/logger.py
```diff
@@ -186,7 +186,7 @@ class AppenderType(object):
     TELEMETRY = 3
 
 
-def add_logger_appender(appender_type, level=LogLevel.INFO, path=None):
+def add_logger_appender(appender_type, level=LogLevel.INFO, path=None, ):
```
larohra: Very curious - what does just adding a `,` to a function do?

vrdmr: Sorry. I think its a typo. I don't think it should be like this.

azurelinuxagent/common/conf.py
```diff
@@ -174,15 +174,23 @@ def enable_rdma(conf=__conf__):
...
+
 def get_logs_verbose(conf=__conf__):
     return conf.get_switch("Logs.Verbose", False)
 
+
 def get_logs_console(conf=__conf__):
     return conf.get_switch("Logs.Console", True)
 
+
+def get_logs_to_telemetry(conf=__conf__):
```
narrieta: is this something we want users to be able to change? (wouldn't think so) the conf file doesn't seem to be appropriate for this.

azurelinuxagent/agent.py
```diff
@@ -65,10 +65,11 @@ def __init__(self, verbose, conf_file_path=None):
         if conf.get_logs_console():
             logger.add_logger_appender(logger.AppenderType.CONSOLE, level,
                     path="/dev/console")
-        # See issue #1035
```
narrieta: has the issue described in #1035 been fixed?

azurelinuxagent/common/logger.py
narrieta: of course the next question is why did the tests did not catch it? :)

azurelinuxagent/common/event.py
```diff
@@ -345,6 +337,30 @@ def add_metric(self, category, counter, instance, value, log_event=False):
         except EventError as e:
             logger.error("{0}", e)
 
+    @staticmethod
+    def _cleanup_message(message):
```
pgombar: Cleanup is a noun, so I thought this would be a message regarding cleanup :) I suggest either renaming to `clean_up_message` or `cleaned_message` or something similar.

azurelinuxagent/common/logger.py
```diff
@@ -155,7 +155,7 @@ def __init__(self, level, event_func):
         self.event_func = event_func
 
     def write(self, level, msg):
-        if self.level <= level:
+        if self.level >= level:
```
pgombar: Why was this reversed? What was it doing before?

azurelinuxagent/common/logger.py
vrdmr: Will change this code.

Just for info: Python allows a "," at the end (it treats the call_args as a tuple)
```
>>> def hello(yo,):
...     print(yo)
... 
>>> hello("Ping")
Ping
>>> hello("Ping", "Ting")
  File "<stdin>", line 1, in <module>
TypeError: hello() takes exactly 1 argument (2 given)
```
Also more discussion [here](https://stackoverflow.com/a/17492103).

tests/common/test_event.py
```diff
@@ -401,3 +401,36 @@ def test_report_metric(self, mock_event):
                 break
         else:
             self.fail("Counter '%idle' not found in event parameters: {0}".format(repr(event_dictionary)))
+
+    def test_cleanup_message(self):
```
pgombar: Can you add a test with a log message starting with `[PERIODIC]`?

azurelinuxagent/common/logger.py
pgombar: Honestly had no idea about this. The more you know!

azurelinuxagent/common/logger.py
vrdmr: Good catch. Fixed it and added a test to capture it.

tests/common/test_event.py
vrdmr: Good suggestion. Added in the tests.

azurelinuxagent/agent.py
vrdmr: Fixed Now. Added an addl. test `test_telemetry_logger_verify_not_logging_errors_warnings` to verify if it is fixed.

azurelinuxagent/common/event.py
```diff
@@ -430,10 +454,14 @@ def add_event(name, op=WALAEventOperation.Unknown, is_success=True, duration=0,
 
 def add_log_event(level, message, reporter=__event_logger__):
     if reporter.event_dir is None:
+        # See #1035 for not adding warning statements here.
```
narrieta: can you expand on what should not be done and why? thanks

vrdmr: I'll add the following description in the code as well - 

The issue with 1035 was never-ending machinery generating events continuously and stretching the system.

At t=0 - Assume 999 events are already in the events folder.
At t=1 - Now, assuming that a new log.WARN [1] event comes in and needs to be written - when saving this event on the disk `save_event` would generate another log.WARN [2] as it warns that there are too many events on the dist.

Thus, this new WARN [2] event again would propagate more WARN events to be written, thus introducing a never-ending cycle of WARNs being written.

Thus to prevent the chance of this happening, I introduced periodic_WARNs. Thus curtailing to one event needed to be written at most, and within the next minute, it would be cleared away.

azurelinuxagent/agent.py
narrieta: As discussed offline, we need a solid solution for the infinite recursion described in #1035

azurelinuxagent/common/event.py
```diff
@@ -429,10 +457,30 @@ def add_event(name, op=WALAEventOperation.Unknown, is_success=True, duration=0,
 
 
 def add_log_event(level, message, reporter=__event_logger__):
+    """
+    # See #1035 for not adding warning/error statements here (any where in the call-tree).
+    The issue with #1035 was never-ending machinery generating events continuously and stretching the system.
+
+    At t=0 - Assume 999 events are already in the events folder.
+    At t=1 - Now, assuming that a new log.WARN [1] event comes in and needs to be written - when saving this event on
+    the disk `save_event` would generate another log.WARN [2] as it warns that there are too many events on the dist.
```
larohra: I have a question, our normal logger just writes to the waagent.log file right? It doesnt create a new file to be saved in the events directory? Only the telemetry events are saved there. So if the `save_event` function writes a warn `logger.warn` how/why would that write to our events directory to cause this recursion?

narrieta: The PR adds a telemetry appender to the logger, so now log.warn does produce a telemetry event. When creating the event (e.g. while saving to disk) somebody does log.warn again, that will try to produce a telemetry event which...

larohra: Yes got it, @vrdmr explained it to me offline. Thanks for the clarification guys! :)

azurelinuxagent/common/event.py
```diff
@@ -219,12 +219,11 @@ def __init__(self):
         self.event_dir = None
         self.periodic_events = {}
 
-    def save_event(self, data):
+    def save_event(self, data, prevent_telemetry_logging=True):
         if self.event_dir is None:
             # To prevent the chance of #1035 happening - introduced periodic_WARNs.
```
larohra: NIT: modify the comment for the new change

azurelinuxagent/common/event.py
```diff
@@ -237,8 +236,7 @@ def save_event(self, data):
         if len(existing_events) >= 1000:
             # To prevent the chance of #1035 happening - introduced periodic_WARNs.
             # More details about it - azurelinuxagent.common.event.add_log_event's pydoc.
-            logger.periodic_warn(logger.EVERY_FIFTEEN_MINUTES,
-                                 "[PERIODIC] Too many files under: {0}, current count:  {1}, removing oldest".format(
+            logger.warn("Too many files under: {0}, current count:  {1}, removing oldest".format(
```
larohra: We wont call this function based on your check in Line 305 of this file. Can you move this `logger.warn` to that line?

larohra: Sorry my bad this is used everywhere. Disregard

larohra: Can you make this periodic warn for 1 min (since we send telemetry every min). The reason is is the events dir > 1000, we would just keep writing this warn until we clear it up which would be at the end of the min.

vrdmr: Good idea. Done. This could be caused by external events as well, which we won't have any control over.

azurelinuxagent/common/event.py
```diff
@@ -235,14 +234,13 @@ def save_event(self, data, prevent_telemetry_logging=True):
...
-            # More details about it - azurelinuxagent.common.event.add_log_event's pydoc.
-            logger.warn("Too many files under: {0}, current count:  {1}, removing oldest".format(
-                                     self.event_dir, len(existing_events)))
-            existing_events.sort()
-            oldest_files = existing_events[:-999]
+        if len(existing_events) >= MAX_NUMBER_OF_EVENTS:
+            logger.periodic_warn(logger.EVERY_MINUTE, "[PERIODIC] Too many files under: {0}, current count:  {1}, "
+                                                      "removing oldest".format(self.event_dir, len(existing_events)))
             try:
+                existing_events = os.listdir(self.event_dir)
```
larohra: NIT: `existing_events` already populated in line 236, line 241 is duplicated and not required

tests/common/test_logger.py
```diff
@@ -229,6 +229,98 @@ def test_telemetry_logger_not_on_by_default(self, mock_save):
...
+        # All 4 appenders should have been included.
+        self.assertEquals(4, counter)
+
+        lg.warn("Test Log")
+        self.assertEqual(mock_file_write.call_count, 1)
+        self.assertEqual(mock_console_write.call_count, 1)
+        self.assertEqual(mock_telem_write.call_count, 1)
+        self.assertEqual(mock_stdout_write.call_count, 1)
+
+        lg.info("Test Log")
```
larohra: Correct me if I'm wrong, but for info, shouldn't only the file_appender be called?

tests/common/test_logger.py
```diff
@@ -229,6 +229,98 @@ def test_telemetry_logger_not_on_by_default(self, mock_save):
...
+        self.assertIn(prefix, mock_file_write.call_args[0][1])
+        self.assertIn(prefix, mock_console_write.call_args[0][1])
+        self.assertIn(prefix, mock_telem_write.call_args[0][1])
+        self.assertIn(prefix, mock_stdout_write.call_args[0][1])
+
+    @patch("azurelinuxagent.common.logger.StdoutAppender.write")
+    @patch("azurelinuxagent.common.logger.TelemetryAppender.write")
+    @patch("azurelinuxagent.common.logger.ConsoleAppender.write")
+    @patch("azurelinuxagent.common.logger.FileAppender.write")
+    def test_nested_logger(self, mock_file_write, mock_console_write, mock_telem_write, mock_stdout_write):
```
larohra: I didnt quite get how this test works, can you explain a bit please?

tests/common/test_logger.py
vrdmr: Yup. The test is incorrectly checking the write. The write for all the loggers gets called, but the levels are honored in the write method itself.

Eg:
```
    def write(self, level, msg):
        if self.level <= level:
            try:
                with open(self.path, "w") as console:
                    console.write(msg)
            except IOError:
                pass
```

vrdmr: Each appender needs its own test to validate the writing of the log message for different levels.

tests/common/test_logger.py
narrieta: +1

tests/common/test_logger.py
vrdmr: Done. Added a new set of tests for each appender.

azurelinuxagent/common/event.py
```diff
@@ -224,13 +234,15 @@ def save_event(self, data):
...
-            existing_events.sort()
-            oldest_files = existing_events[:-999]
-            logger.warn("Too many files under: {0}, removing oldest".format(self.event_dir))
+        if len(existing_events) >= MAX_NUMBER_OF_EVENTS:
+            logger.periodic_warn(logger.EVERY_MINUTE, "[PERIODIC] Too many files under: {0}, current count:  {1}, "
+                                                      "removing oldest".format(self.event_dir, len(existing_events)))
             try:
-                for f in oldest_files:
-                    os.remove(os.path.join(self.event_dir, f))
+                existing_events = os.listdir(self.event_dir)
```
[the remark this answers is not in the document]
vrdmr: When this test does `add_event`, it was trying to add a new event, but the event already had 1000 events in the file, because of which it generated a `log.warn` - which again generates and adds one more event, and in the process deletes the oldest file.

Now, the earlier `add_event` call continues and does the os.remove of the oldest file, but it got deleted earlier itself, and thus hits the following issue [TravisLink](https://travis-ci.org/Azure/WALinuxAgent/jobs/614418917):
Eg:
```
======================================================================
ERROR: test_telemetry_logger_verify_maximum_recursion_depths_doesnt_happen (tests.common.test_logger.TestLogger)
----------------------------------------------------------------------
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/mock/mock.py", line 1305, in patched
    return func(*args, **keywargs)
  File "/home/travis/build/Azure/WALinuxAgent/tests/common/test_logger.py", line 354, in test_telemetry_logger_verify_maximum_recursion_depths_doesnt_happen
    logger.warn('Test Log - {0} - 1 - Warning'.format(i))
  File "/home/travis/build/Azure/WALinuxAgent/azurelinuxagent/common/logger.py", line 289, in warn
    DEFAULT_LOGGER.warn(msg_format, *args)
  File "/home/travis/build/Azure/WALinuxAgent/azurelinuxagent/common/logger.py", line 79, in warn
    self.log(LogLevel.WARNING, msg_format, *args)
  File "/home/travis/build/Azure/WALinuxAgent/azurelinuxagent/common/logger.py", line 140, in log
    write_log(appender)
  File "/home/travis/build/Azure/WALinuxAgent/azurelinuxagent/common/logger.py", line 117, in write_log
    log_appender.write(level, log_item)
  File "/home/travis/build/Azure/WALinuxAgent/azurelinuxagent/common/logger.py", line 207, in write
    self.event_func(level, msg)
  File "/home/travis/build/Azure/WALinuxAgent/azurelinuxagent/common/event.py", line 510, in add_log_event
    reporter.add_log_event(level, message)
  File "/home/travis/build/Azure/WALinuxAgent/azurelinuxagent/common/event.py", line 314, in add_log_event
    self.save_event(json.dumps(data))
  File "/home/travis/build/Azure/WALinuxAgent/azurelinuxagent/common/event.py", line 244, in save_event
    os.remove(os.path.join(self.event_dir, event_file))
OSError: [Errno 2] No such file or directory: '/tmp/TestLogger_ARDFnT/1574239919504255.tld'
```

This was specifically seen in Python2 and could be due to the event order in the list, and being picked after sorting.

FYI - @larohra

azurelinuxagent/common/event.py
vrdmr: Explained above (https://github.com/Azure/WALinuxAgent/pull/1659#discussion_r348859779)

azurelinuxagent/common/event.py
```diff
@@ -461,10 +485,30 @@ def add_event(name, op=WALAEventOperation.Unknown, is_success=True, duration=0,
 
 
 def add_log_event(level, message, reporter=__event_logger__):
+    """
+    # See #1035 for not adding warning/error statements here (any where in the call-tree).
```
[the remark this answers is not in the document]
vrdmr: Yup. Removing this comment now.

azurelinuxagent/common/event.py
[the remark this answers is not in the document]
vrdmr: So it turns out I didn't need the new list but understand when the sort is called. Once the sort is called later, it fixes the deletion problem.

azurelinuxagent/common/event.py
```diff
@@ -224,13 +234,15 @@ def save_event(self, data):
...
+            logger.periodic_warn(logger.EVERY_MINUTE, "[PERIODIC] Too many files under: {0}, current count:  {1}, "
+                                                      "removing oldest".format(self.event_dir, len(existing_events)))
             try:
-                for f in oldest_files:
-                    os.remove(os.path.join(self.event_dir, f))
+                existing_events = os.listdir(self.event_dir)
+                existing_events.sort()
+                oldest_files = existing_events[:-999]
+                for event_file in oldest_files:
+                    os.remove(os.path.join(self.event_dir, event_file))
```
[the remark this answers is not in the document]
vrdmr: The goal here was to catch any issues while reading the events folder. About the listing, yes, included that now.

Also updated the EventError message.

tests/common/test_logger.py
vrdmr: The purpose of this test is to see if the logger gets correctly created when passed it another logger. And also if the appender correctly gets the messages This is how the `ExtHandlerInstance` logger works. There was no test to see this.

What I do is initialize the default logger (logger.add_logger_appender(*)). Then create a new logger from it
```
lg = logger.Logger(logger.DEFAULT_LOGGER, child_prefix)
```

and then log using `lg`, and `logger`. And see if both logs are flowing through or not. This already works as we can see `ExtHandlerInstance` logs and `DEFAULT_LOGGER` logs are available.

tests/common/test_logger.py
```diff
@@ -376,15 +390,15 @@ def test_telemetry_logger_check_all_file_logs_written_when_events_gt_1000(self,
         exception_caught = False
         prefix = "YoloLogger"
 
-        lg = logger.Logger(logger.DEFAULT_LOGGER, prefix)
-        lg.add_appender(logger.AppenderType.FILE, logger.LogLevel.INFO, path=self.log_file)
-        lg.add_appender(logger.AppenderType.TELEMETRY, logger.LogLevel.WARNING, path=add_log_event)
+        logger.add_logger_appender(logger.AppenderType.FILE, logger.LogLevel.INFO, path=self.log_file)
```
larohra: NIT: The current test name is `test_telemetry_logger_check_all_file_logs_written_when_events_gt_**1000**` Could you please rename it to something like `*_gt_MAX_EVENTS` so that incase we change the size later we dont need to modify this test case again. Also in Line 406 we're checking explicitly for 1000, can you change it to the new constant you added for MAX_EVENTS.
Thanks!

tests/common/test_logger.py
```diff
@@ -383,10 +383,10 @@ def test_telemetry_logger_verify_maximum_recursion_depths_doesnt_happen(self, *_
...
     @patch("azurelinuxagent.common.event.send_logs_to_telemetry", return_value=True)
     @patch("azurelinuxagent.common.conf.get_lib_dir")
-    def test_telemetry_logger_check_all_file_logs_written_when_events_gt_1000(self, mock_lib_dir, *_):
+    def test_telemetry_logger_check_all_file_logs_written_when_events_gt_MAX_NUMBER_OF_EVENTS(self, mock_lib_dir, *_):
         mock_lib_dir.return_value = self.lib_dir
         __event_logger__.event_dir = self.event_dir
-        no_of_log_statements = 1100
+        no_of_log_statements = MAX_NUMBER_OF_EVENTS + 100
         exception_caught = False
         prefix = "YoloLogger"
```
larohra: You missed out 1 case in line 406 in test_logger.py :)

azurelinuxagent/common/event.py
```diff
@@ -223,24 +233,27 @@ def save_event(self, data):
...
-            except IOError as e:
-                raise EventError(e)
+        try:
+            existing_events = os.listdir(self.event_dir)
+            if len(existing_events) >= MAX_NUMBER_OF_EVENTS:
+                logger.periodic_warn(logger.EVERY_MINUTE, "[PERIODIC] Too many files under: {0}, current count:  {1}, "
+                                                          "removing oldest event files".format(self.event_dir,
+                                                                                               len(existing_events)))
+                existing_events.sort()
+                oldest_files = existing_events[:-999]
```
narrieta: (MAX_NUMBER_OF_EVENTS - 1) instead of 999

azurelinuxagent/common/event.py
```diff
@@ -342,7 +343,31 @@ def add_metric(self, category, counter, instance, value, log_event=False):
...
+        #
+        # The timestamp and the level are redundant, and should be stripped. The logging library does not schematize
+        # this data, so I am forced to parse the message using a regex.  The format is regular, so the burden is low,
+        # and usability on the telemetry side is high.
+        #
+
+        if not message:
+            return message
+
+        log_format_parser = re.compile(r"^\d+/\d+/\d+\s\d+:\d+:\d+\.?\d*(\s[A-Z]*\s)?\s?(.*)$")
```
narrieta: regex will need to be updated per #1716

azurelinuxagent/common/logger.py
```diff
@@ -81,6 +82,43 @@ def error(self, msg_format, *args):
...
+
+            Eg:
+            Assuming a logger with two appenders - FileAppender and TelemetryAppender. Here is an example of
+            how using appender_lock flag can help.
+
+            logger.warn("foo")
+                |- log.warn() (azurelinuxagent.common.logger.Logger.warn)
+                    |- log() (azurelinuxagent.common.logger.Logger.log)
+                        |- FileAppender.appender_lock is currently False not log_appender.appender_lock is True
+                            |- We sets it to True.
```
narrieta: typo: "set"

tests/common/test_logger.py
```diff
@@ -161,34 +174,336 @@ def test_periodic_forwards_message_and_args(self, mock_info, mock_error, mock_wa
...
+                                             'Microsoft.OSTCExtensions.DummyExtension-1.2.3.4')
             elif x['name'] == 'Context2':
                 self.assertEqual(x['value'], '')
-
             elif x['name'] == 'Context3':
                 self.assertEqual(x['value'], '')
+
+    @patch('azurelinuxagent.common.event.EventLogger.save_event')
+    def test_telemetry_logger_not_on_by_default(self, mock_save):
+        appender = logger.TelemetryAppender(logger.LogLevel.WARNING, add_log_event)
```
narrieta: this test should be done for all the appenders, not only the telemetry appender

tests/common/test_logger.py
narrieta: could you add the reply to the original question as a comment in the code? thanks

also for the nested logger test below

tests/common/test_logger.py
```diff
@@ -161,34 +174,336 @@ def test_periodic_forwards_message_and_args(self, mock_info, mock_error, mock_wa
...
+    @patch('azurelinuxagent.common.logger.Logger.error')
+    @patch('azurelinuxagent.common.logger.Logger.warn')
+    def test_telemetry_logger_verify_not_logging_errors_warnings(self, mock_warn, mock_error, *args):
+        appender = logger.TelemetryAppender(logger.LogLevel.WARNING, add_log_event)
+
+        with patch('azurelinuxagent.common.event.EventLogger.save_event') as mock_save:
+            appender.write(logger.LogLevel.WARNING, 'Cgroup controller "memory" is not mounted. '
+                                                    'Microsoft.OSTCExtensions.DummyExtension-1.2.3.4')
+            self.assertEqual(1, mock_save.call_count)
+            self.assertEqual(0, mock_warn.call_count)
```
narrieta: I don't quite get what the asserts in 348 & 349 are checking... could you expand a little? (and maybe add it as a comment in the code?)

tests/common/test_logger.py
```diff
@@ -161,34 +174,336 @@ def test_periodic_forwards_message_and_args(self, mock_info, mock_error, mock_wa
...
+            self.assertEqual(0, mock_error.call_count)
+
+        # Writing 2000 events should generate only one more log event due to too many files. #1035 was caused due to
+        # too many files being written in an error condition (earlier code would write 1000 files in this case).
+        for _ in range(2000):
+            appender.write(logger.LogLevel.WARNING, 'Cgroup controller "memory" is not mounted. '
+                                                    'Microsoft.OSTCExtensions.DummyExtension-1.2.3.4')
+
+        self.assertEqual(1, mock_warn.call_count)
```
narrieta: maybe we need to rename the test: I read "test_telemetry_logger_verify_not_logging_errors_warnings" as implying that no warnings/errors should be logged, but we do log a warning

tests/common/test_logger.py
```diff
@@ -161,34 +174,336 @@ def test_periodic_forwards_message_and_args(self, mock_info, mock_error, mock_wa
...
+        logger.add_logger_appender(logger.AppenderType.FILE, logger.LogLevel.INFO, path="/dev/null")
+        logger.add_logger_appender(logger.AppenderType.TELEMETRY, logger.LogLevel.WARNING, path=add_log_event)
+
+        for i in range(MAX_NUMBER_OF_EVENTS):
+            logger.warn('Test Log - {0} - 1 - Warning'.format(i))
+
+        exception_caught = False
+
+        # #1035 was caused due to too many files being written in an error condition. Adding even one more here broke
+        # the camels back earlier. This should be resolved now.
```
narrieta: can we clarify in the comment that this would go into an infinite recursion that would raise a RuntimeError?

azurelinuxagent/common/event.py
vrdmr: Done. Updated the tests as well with UTC logs

tests/common/test_logger.py
vrdmr: This test was validating specifically if Telemetry Appender is sending values or not. 

If I understand your concern, do you mean to test - just doing a log.warn should not generate any kind of log calls?

tests/common/test_logger.py
vrdmr: Discussed offline - Removing this test as it is not relevant now. The purpose of this test was to make sure no additional logs are added, especially with #1035-like issue. But with the new fix for #1035 done, it is not needed anymore.

tests/common/test_logger.py
vrdmr: Same as above - Discussed offline - Removing this test as it is not relevant now. The purpose of this test was to make sure no additional logs are added, especially with #1035-like issue. But with the new fix for #1035 done, it is not needed anymore.

tests/common/test_logger.py
vrdmr: Added a new test - `test_no_appenders_added` which validates no logs are written in any appender if no appender is added.

tests/common/test_logger.py
vrdmr: Done.

tests/common/test_cgroups.py
```diff
@@ -59,7 +60,8 @@ def test_correct_creation(self):
         self.assertEqual(test_cgroup.path, "dummy_path")
         self.assertEqual(test_cgroup.name, "test_extension")
 
-    def test_is_active(self):
+    @patch("azurelinuxagent.common.osutil.default.DefaultOSUtil._get_proc_stat")
```
narrieta: why is this needed? thanks

vrdmr: I am adding `@patch("azurelinuxagent.common.osutil.default.DefaultOSUtil._get_proc_stat")` in tests to make them run on Mac, and those tests do not need `/proc/stat` file. In tests that actually need `/proc/stat` file, I don't patch it.

vrdmr: Added the comments in the tests as well. Thanks.

PR #1660, opened 2019-10-08
Remove agent version from environment variables

larohra:
## Description
Removing the GAVersion Environment Variable from the list of helper environment variables we pass to the extensions as the Extensions should not have a dependency on a particular agent version and should follow the contract we set for them.
Issue # 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1660)

### DIFF
--- a/tests/ga/test_exthandlers.py
+++ b/tests/ga/test_exthandlers.py
@@ -233,7 +233,6 @@ class LaunchCommandTestCase(AgentTestCase):
         self.mock_get_log_dir = patch("azurelinuxagent.ga.exthandlers.ExtHandlerInstance.get_log_dir", lambda *_: self.log_dir)
         self.mock_get_log_dir.start()
 
-        mock_sleep = time.sleep
         self.mock_sleep = patch("time.sleep", lambda *_: mock_sleep(0.01))
         self.mock_sleep.start()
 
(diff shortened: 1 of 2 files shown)


larohra:
We have a test function that checks for something similar (`test_ext_path_and_version_env_variables_set_for_ever_operation`) but I like your idea of adding a new test just to check for environment variables. Let me add that test and push an update

tests/ga/test_exthandlers.py
```diff
@@ -580,3 +581,19 @@ def capture_process_output(stdout_file, stderr_file):
...
+
+        command = """
+            printenv | grep -E '(%s)'
+        """ % '|'.join(helper_env_vars.keys())
+
+        test_file = self.create_script('printHelperEnvironments.sh', command)
+        output = self.ext_handler_instance.launch_command(test_file)
+
+        for helper_var in helper_env_vars:
+            self.assertIn("%s=%s" % (helper_var, helper_env_vars[helper_var]), output)
```
vrdmr: Thanks for adding the test. 

Just a qq: I assume this test won't fail when a new Env. variable is added to the list (I see you are only checking for the env variables mentioned in `helper_env_vars`). Is there a way to make that happen as well?

tests/ga/test_exthandlers.py
```diff
@@ -231,7 +233,6 @@ def setUp(self):
         self.mock_get_log_dir = patch("azurelinuxagent.ga.exthandlers.ExtHandlerInstance.get_log_dir", lambda *_: self.log_dir)
         self.mock_get_log_dir.start()
 
-        mock_sleep = time.sleep
```
larohra: Removed this line as a utility function with the same name already exists in `tests/tools.py`

PR #1657, opened 2019-10-07
Start sending Execution Mode in the Telemetry

vrdmr:
## Description
Sending `ExecutionMode` field when sending telemetry events.

---

### PR information
- [x] The title of the PR is clear and informative.
- [x] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).
- [x] Except for special cases involving multiple contributors, the PR is started from a fork of the main repository, not a branch.
- [ ] If applicable, the PR references the bug/issue that it fixes in the description.
- [x] New Unit tests were added for the changes made and Travis.CI is passing.

### Quality of Code and Contribution Guidelines
- [x] I have read the [contribution guidelines](https://github.com/Azure/WALinuxAgent/blob/master/.github/CONTRIBUTING.md).

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1657)

PR #1663, opened 2019-10-10
Changing add_event to start sending missing fields

vrdmr:
### Description
This PR makes changes to the base implementation methods used for sending events to telemetry, to add newer telemetry fields which weren't sent earlier. Mainly TaskName, EventTid, EventPid, OpCodeName, KeywordName. The ExecutionMode has been added in #1657.

---
### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1663)


tests/common/test_event.py
```diff
@@ -269,13 +268,42 @@ def test_periodic_emits_after_elapsed_delta(self, mock_event):
...
+                                           "Yo")
+
+    @patch("azurelinuxagent.common.event.datetime")
+    @patch('azurelinuxagent.common.event.EventLogger.add_event')
+    def test_periodic_forwards_args_default_values(self, mock_event, mock_datetime):
+        event.__event_logger__.reset_periodic()
+        event_time = datetime.strptime("2019-01-01 01:30:00", '%Y-%m-%d %H:%M:%S').__str__()
+        mock_datetime.utcnow = mock.Mock(return_value=event_time)
+
+        # Doing the partial to mock the passing of default values of the
```
larohra: NIT: Incomplete sentence

azurelinuxagent/common/event.py
```diff
@@ -242,57 +242,55 @@ def is_period_elapsed(self, delta, h):
...
             (self.periodic_events[h] + delta) <= datetime.now()
 
-    def add_periodic(self,
-                     delta, name, op=WALAEventOperation.Unknown, is_success=True, duration=0,
-                     version=CURRENT_VERSION, message="", evt_type="",
-                     is_internal=False, log_event=True, force=False):
-
+    def add_periodic(self, delta, name, op=WALAEventOperation.Unknown, is_success=True, duration=0,
+                     version=str(CURRENT_VERSION), message="", evt_type="", is_internal=False, log_event=True,
+                     force=False, event_time_opcode_name=None, event_pid=0, event_tid=0, task_name="", keyword_name=""):
```
larohra: Instead of passing `event_time_opcode_name=None` this as a parameter? Why not set it in the `_add_default_parameters_to_event` function? 
I'm trying to think of scenarios where we would set it to something other than the time the event occurred. Is there some scenario that I'm missing?

vrdmr: I am already setting the event_time_opcode_name in `_add_default_parameters_to_event` [(link here)](https://github.com/Azure/WALinuxAgent/pull/1663/files#diff-58adf10cb50d782909fbf94ac6c08bf5R359) if it is not passed. Just exposing it to the caller if the caller wants to pass it.

```
event.parameters.append(TelemetryEventParam('OpcodeName', event_time_opcode_name if event_time_opcode_name else datetime.utcnow().__str__()))
```

tests/common/test_event.py
vrdmr: Thanks for catching this. Will fix it now.

azurelinuxagent/common/event.py
larohra: ohh sorry missed it, thanks for pointing it out!

narrieta: > Just exposing it to the caller if the caller wants to pass it.

Do you anticipate any scenario where this would be needed?

narrieta: (similar for pid & tid)

azurelinuxagent/common/event.py
```diff
@@ -242,57 +242,55 @@ def is_period_elapsed(self, delta, h):
...
-                                                    is_success))
+        event.parameters.append(TelemetryEventParam('OperationSuccess', is_success))
         event.parameters.append(TelemetryEventParam('Message', message))
         event.parameters.append(TelemetryEventParam('Duration', duration))
         event.parameters.append(TelemetryEventParam('ExtensionType', evt_type))
-        event.parameters.append(TelemetryEventParam('ContainerId',
-                                                    os.environ.get(CONTAINER_ID_ENV_VARIABLE, "UNINITIALIZED")))
 
+        self._add_default_parameters_to_event(event, container_id=os.environ.get(
+            CONTAINER_ID_ENV_VARIABLE, "UNINITIALIZED"), event_time_opcode_name=event_time_opcode_name,
```
narrieta: (i moved CONTAINER_ID_ENV_VARIABLE to event.py)

narrieta: let's make 

> os.environ.get(CONTAINER_ID_ENV_VARIABLE, "UNINITIALIZED")

a helper function to ensure consistency

azurelinuxagent/common/event.py
```diff
@@ -339,12 +342,27 @@ def add_metric(self, category, counter, instance, value, log_event=False):
...
         except EventError as e:
             logger.error("{0}", e)
 
+    @staticmethod
+    def _add_default_parameters_to_event(event, container_id="UNINITIALIZED", event_time_opcode_name=None, event_pid=0,
+                                         event_tid=0, task_name='', keyword_name=''):
+        event.parameters.append(TelemetryEventParam('ContainerId', container_id))
+        event.parameters.append(TelemetryEventParam('OpcodeName', event_time_opcode_name if event_time_opcode_name
+                                                    else datetime.utcnow().__str__()))
+        event.parameters.append(TelemetryEventParam('EventTid', event_tid))
```
narrieta: shouldn't this be adding the current pid/tid?

azurelinuxagent/common/event.py
```diff
@@ -339,12 +342,27 @@ def add_metric(self, category, counter, instance, value, log_event=False):
...
+        event.parameters.append(TelemetryEventParam('EventPid', event_pid))
+        event.parameters.append(TelemetryEventParam("TaskName", task_name))
```
narrieta: for task/keyword, which we currently do not set, should the default be something else than empty? is it important to distinguish whether they are populated vs whether they are not provided?

azurelinuxagent/common/event.py
vrdmr: i'll rebase against develop, and then add a new helper.

azurelinuxagent/common/event.py
vrdmr: For Task, I was thinking of "ExtHandler" (the prefix we use for Logger as well) as the default TaskName.

> is it important to distinguish whether they are populated vs whether they are not provided?

The necessity of providing at least empty values is to make sure that MA gets the schema correctly on the host side.

azurelinuxagent/common/event.py
vrdmr: Do you mean the default values here or the passed values from the top?

For default values, I was thinking -1 for default values as well as PID 0 has a special meaning.

azurelinuxagent/common/event.py
```diff
@@ -339,12 +342,27 @@ def add_metric(self, category, counter, instance, value, log_event=False):
...
+                                              event_pid=event_pid, event_tid=event_tid, task_name=task_name,
+                                              keyword_name=keyword_name)
         data = get_properties(event)
         try:
             self.save_event(json.dumps(data))
         except EventError as e:
             logger.error("{0}", e)
 
+    @staticmethod
+    def _add_default_parameters_to_event(event, container_id="UNINITIALIZED", event_time_opcode_name=None, event_pid=0,
```
pgombar: Can you extract `TelemetryEventParam("GAVersion", CURRENT_AGENT)` from `init_sysinfo` to here since it should be retrieved and populated when we are constructing the event, not only once on agent startup?

tests/common/test_event.py
```diff
@@ -269,13 +272,29 @@ def test_periodic_emits_after_elapsed_delta(self, mock_event):
...
-        mock_event.assert_called_once_with(
-            "FauxEvent",
-            duration=0, evt_type='', is_internal=False, is_success=True,
-            log_event=True, message='', op=WALAEventOperation.Unknown,
-            version=str(CURRENT_VERSION))
+    @patch("azurelinuxagent.common.event.EventLogger.add_event")
+    def test_add_event_default_variables(self, mock_add_event):
+        add_event('test', message='test event')
+        mock_add_event.assert_called_once_with('test', duration=0, evt_type='', is_internal=False, is_success=True,
+                                               log_event=True, message='test event', op='Unknown',
```
pgombar: Why `Unknown` instead of `WALAEventOperation.Unknown`?

tests/ga/test_monitor.py
```diff
@@ -428,7 +428,6 @@ def test_collect_and_send_events(self, mock_lib_dir, patch_send_event, patch_get
                          '<Param Name="ExtensionType" Value="" T="mt:wstr" />' \
                          '<Param Name="OSVersion" ' \
                          'Value="Linux:DISTRO_NAME-DISTRO_VERSION-DISTRO_CODE_NAME:platform-release" T="mt:wstr" />' \
-                         '<Param Name="GAVersion" Value="WALinuxAgent-{0}" T="mt:wstr" />' \
```
pgombar: Why wouldn't this be here? We are not removing this field, right?

tests/common/test_event.py
vrdmr: It was just for testing, wanted to see if the passed values were coming.

tests/ga/test_monitor.py
vrdmr: This particular test just tests the `collect_and_send_events` and the XML is created from the event fetched from the VM, along with the sysinfo added. The sysinfo does not populate any `GAVersion` now, which is why I removed it.

azurelinuxagent/common/event.py
```diff
@@ -341,12 +335,30 @@ def add_metric(self, category, counter, instance, value, log_event=False):
...
         try:
             self.save_event(json.dumps(data))
         except EventError as e:
             logger.error("{0}", e)
 
+    @staticmethod
+    def _add_default_parameters_to_event(event):
+        task_name = threading.current_thread().getName()
+
+        # We set thread names for the threads spawned by ExtHandler. We get the name of the thread,
```
narrieta: where do we set the thread name?

tests/common/test_event.py
narrieta: but why the change?

azurelinuxagent/common/event.py
vrdmr: Defined in the of MonitorHandler and EnvHandler `start()`.

`self.server_thread.setName("EnvHandler")` ([link](https://github.com/Azure/WALinuxAgent/pull/1663/files#diff-e553d83092bcdeb28abdda5cbb040a9aR91))
`self.event_thread.setName("MonitorHandler")` ([link](https://github.com/Azure/WALinuxAgent/pull/1663/files#diff-3d8dc864f0284d65cf57fb2b2f9799a0R165)).

tests/common/test_event.py
vrdmr: Oh. Now I understood what you both meant. Yes, I'll revert it to `WALAEventOperation.Unknown`.

azurelinuxagent/common/event.py
narrieta: the we should respect those values instead of setting them @352, right?

vrdmr: It does. In @352.

```
test_suite = ['EnvHandler', 'MonitorHandler', 'MainThread']

for i in test_suite:
    task_name = i
    task_name = task_name if task_name != "MainThread" else "ExtHandler"
    print(task_name)
```
Output:
```
/usr/bin/python3.6 /home/varad/.PyCharm2019.2/config/scratches/scratch_3.py
EnvHandler
MonitorHandler
ExtHandler
```

narrieta: I mean respect whatever value the name has, instead of changing it @352

azurelinuxagent/common/event.py
```diff
@@ -314,6 +307,7 @@ def add_log_event(self, level, message):
         event.parameters.append(TelemetryEventParam('Context2', ''))
         event.parameters.append(TelemetryEventParam('Context3', ''))
 
+        self._add_default_parameters_to_event(event)
```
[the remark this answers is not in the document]
vrdmr: Discussed the usage of `add_log_event`. It is being used in #1659.

azurelinuxagent/common/event.py
```diff
@@ -342,12 +335,36 @@ def add_metric(self, category, counter, instance, value, log_event=False):
...
+        # the event, rather than sending the event.
+        event.parameters.append(TelemetryEventParam('ContainerId', get_container_id_from_env()))
+
+        # This is used as the actual time of event generation.
+        event.parameters.append(TelemetryEventParam('OpcodeName', datetime.utcnow().__str__()))
+
+        # Other commong parameters added to the events
+        event.parameters.append(TelemetryEventParam('EventTid', threading.current_thread().ident))
+        event.parameters.append(TelemetryEventParam('EventPid', os.getpid()))
+        event.parameters.append(TelemetryEventParam("TaskName", threading.current_thread().getName()))
```
narrieta: I think we should reserve the TaskName for something else. We already have the identity of the thread in EventTid

azurelinuxagent/ga/monitor.py
```diff
@@ -329,10 +330,26 @@ def add_sysinfo(self, event):
...
-                TelemetryEventParam("ContainerId", os.environ.get(CONTAINER_ID_ENV_VARIABLE, "UNINITIALIZED")))
+                TelemetryEventParam("ContainerId", get_container_id_from_env()))
+        if "GAVersion" not in event:
+            # GAVersion is populated for agent events on the fly, but not for extension events. Add it if it's missing.
+            final_parameters.append(TelemetryEventParam("GAVersion", CURRENT_AGENT))
+
+        # Default fields are only populated by Agent and not the extension. Agent will fill up any event if they don't
+        # have the default params.
+        if "OpcodeName" not in event:
+            final_parameters.append(TelemetryEventParam("OpcodeName", datetime.datetime.utcnow().__str__()))
```
narrieta: according to _add_default_parameters_to_event, the Opcode is  of event generation.  Using now() here would be incorrect. We should not populate it, or let extension owners populated in the event file to make it accurate.

azurelinuxagent/ga/monitor.py
```diff
@@ -329,10 +330,26 @@ def add_sysinfo(self, event):
...
         final_parameters.extend(self.sysinfo)
 
-        # Container id is populated for agent events on the fly, but not for extension events. Add it if it's missing.
-        if "ContainerId" not in [param.name for param in event.parameters]:
+        if "ContainerId" not in event:
+            # ContainerId is populated for agent events on the fly, but not for extension events. Add it if it's missing.
             final_parameters.append(
-                TelemetryEventParam("ContainerId", os.environ.get(CONTAINER_ID_ENV_VARIABLE, "UNINITIALIZED")))
+                TelemetryEventParam("ContainerId", get_container_id_from_env()))
+        if "GAVersion" not in event:
```
narrieta: we should merge this list of parameters with the list in _add_default_parameters_to_event, since they must be the same.

you could, for example, add a parameter to _add_default_parameters_to_event to indicate whethet to set the values uncoditionally (as it currently does) or populate them with default values when they are not present (as this code does)

the idea is to enforce consistency and reduce the chance of somebody forgetting to add a parameter in one of those 2 places

azurelinuxagent/common/event.py
pgombar: I think there's value in sending the thread name too, since only seeing a PID in Kusto doesn't tell us much. We can't map a PID to a thread name based only on the PID, right?

azurelinuxagent/ga/monitor.py
pgombar: Or we can have a static list of parameter names (something like `AGENT_TELEMETRY_PARAMETERS = ["GAVersion", "OpcodeName", ...]` and iterate through it in both places, ensuring all of them are added.

tests/data/ext/event_from_agent.xml
```diff
@@ -0,0 +1,26 @@
+<Data>
```
pgombar: Can you add an actual event from the agent instead of this one, which seems to have come from CustomScript? Heartbeat would be the simplest one, or ProcesSGoalState.

tests/ga/test_monitor.py
```diff
@@ -122,13 +138,25 @@ def test_add_sysinfo_should_honor_sysinfo_values_from_agent_for_agent_events(sel
         sysinfo_role_name_value = "sysinfo_dummy_role"
         sysinfo_role_instance_name_value = "sysinfo_dummy_role_instance"
         sysinfo_execution_mode_value = "sysinfo_IAAS"
+        GAVersion_value = "WALinuxAgent-2.2.44"
+        OpcodeName_value = "10101010"
+        EventTid_value = 140240309798656
+        EventPid_value = 108573
```
pgombar: According to the XML example, these two are strings, not ints?

tests/ga/test_monitor.py
```diff
@@ -199,12 +257,16 @@ def test_add_sysinfo_should_honor_sysinfo_values_from_agent_for_extension_events
...
+        with patch("azurelinuxagent.ga.monitor.datetime") as patch_datetime:
+            patch_datetime.datetime.utcnow = Mock(return_value=datetime.datetime.strptime("2019-01-01 01:30:00",
+                                                                        '%Y-%m-%d %H:%M:%S'))
+            monitor_handler.add_sysinfo(event)
 
         self.assertNotEqual(None, event)
         self.assertNotEqual(0, event.parameters)
         self.assertTrue(all(param is not None for param in event.parameters))
 
+        counter = 0
```
pgombar: Duplicate of line below?

tests/ga/test_monitor.py
```diff
@@ -562,10 +644,16 @@ def test_collect_and_send_events(self, mock_lib_dir, patch_send_event, *args):
...
+                         '<Param Name="ImageOrigin" Value="1" T="mt:uint64" />' \
+                         '<Param Name="ContainerId" Value="c6d5526c-5ac2-4200-b6e2-56f2b70c5ab2" T="mt:wstr" />' \
+                         '<Param Name="GAVersion" Value="WALinuxAgent-2.2.44" T="mt:wstr" />' \
+                         '<Param Name="EventTid" Value="" T="mt:wstr" />' \
+                         '<Param Name="EventPid" Value="" T="mt:wstr" />' \
+                         '<Param Name="TaskName" Value="" T="mt:wstr" />' \
+                         '<Param Name="KeywordName" Value="" T="mt:wstr" />]]>' \
                          '</Event>'.format(AGENT_VERSION)
 
+        self.maxDiff = None
```
pgombar: Unused?

azurelinuxagent/common/event.py
narrieta: There are only 2 threads with names: the daemon and the handler, so this doesn't add a lot of value. From the messages we can easily figure out daemon vs handler.

If preserving the names is important an alternative could be to use EventTid for that.

TaskName comes from the WIndows agent, where os events have this property. 

We could consider using it later for actual tasks in the agent ("download extension", "update goal state", "collect cgroup data", etc)

azurelinuxagent/ga/monitor.py
narrieta: List by itself would not help much, I think, since the value is different for each item... or how could we use it? maybe I don't get your idea.

I think what we would like to enforce is that the same  values are set on both places, and that the correct policy (set the current value unconditionally vs set default values) is done in the correct place.

azurelinuxagent/common/event.py
```diff
@@ -342,12 +335,36 @@ def add_metric(self, category, counter, instance, value, log_event=False):
...
+        # new monitor thread would pick up the events from the disk and send it with the CURRENT_AGENT, which would have
+        # newer version of the agent. This causes confusion.
+        event.parameters.append(TelemetryEventParam("GAVersion", CURRENT_AGENT))
+
+        # ContainerId can change due to live migration and we want to preserve the container Id of the container writing
+        # the event, rather than sending the event.
+        event.parameters.append(TelemetryEventParam('ContainerId', get_container_id_from_env()))
+
+        # This is used as the actual time of event generation.
+        event.parameters.append(TelemetryEventParam('OpcodeName', datetime.utcnow().__str__()))
```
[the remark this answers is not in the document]
vrdmr: I wanted to do it, but it is a lot of change across 

In event.py, we have `from datetime import datetime`, where as in monitor.py, we only do `import datetime`. I was thinking of changing it but it would be a lot of change for this PR, which is why I kept it as it is. Will do that subsequently.

azurelinuxagent/common/event.py
narrieta: BTW, the Windows Agent uses TPL's tasks, which are different than threads, and populate this field with the task name, so for them this does add value.
If we do not have a good use for this field, I'd leave it empty for future use. One use is what I mentioned above, but it could be something totally different. For example, this PR uses the opcode for the actual timestamp of the event. We may need another field in the future and having this one available would help.

tests/ga/test_monitor.py
vrdmr: This is a flag used by the testing framework to limit how much it should present if there is a difference, on the console. I set the maxDiff=None to explicitly show the diff or else you would see a message like `Diff is 6619 characters long. Set self.maxDiff to None to see it.`

azurelinuxagent/common/event.py
pgombar: > There are only 2 threads with names: the daemon and the handler, so this doesn't add a lot of value. From the messages we can easily figure out daemon vs handler.

There are several thread names being set in this PR: MonitorHandler, ExtHandler, EnvHandler.

> If preserving the names is important an alternative could be to use EventTid for that.

If we're choosing between discarding the thread name completely or using EventTid for the thread name instead of the thread id, I vote for keeping the thread name and using EventTid.

tests/ga/test_monitor.py
vrdmr: Thanks for the catch.

tests/ga/test_monitor.py
vrdmr: In the XML, they are ints - `mt:uint64`. And are casted as ints, when reading the event.
```
    <Param Name="EventTid" Value="140240309798656" T="mt:uint64"/>
    <Param Name="EventPid" Value="108573" T="mt:uint64"/>
```

azurelinuxagent/common/event.py
narrieta: > using EventTid for the thread name instead of the thread id, I vote for keeping the thread name and using EventTid.

me too

narrieta: if doing that you may want to add an extra check to populate use the thread id if the name is not set

azurelinuxagent/ga/monitor.py
pgombar: I just realized what you were suggesting. I agree that the list wouldn't help since the values are different.
If we add a flag like you're suggesting, then this function can also call `_add_default_parameters_to_event` instead of setting them on its own.

tests/ga/test_monitor.py
pgombar: Sorry, I must have been looking at the wrong parameter :)

azurelinuxagent/ga/monitor.py
narrieta: yes, sorry I did not explain fully. You are right, the idea is that the list of values and the logic to set them would be in a single place (_add_default_parameters_to_event) instead of 2.

tests/data/ext/event_from_agent.xml
vrdmr: Done.

azurelinuxagent/ga/monitor.py
vrdmr: Okay. Will set it to empty.

azurelinuxagent/common/event.py
vrdmr: >We can't map a PID to a thread name based only on the PID, right?

I haven't investigated that route.

azurelinuxagent/ga/monitor.py
vrdmr: I will make _add_default_params into a static function in the Event class with a "defaults" flag. to make it easier.

azurelinuxagent/common/event.py
narrieta: Yes, right now EventTid is not very useful, but as we add more events it'll be good to know which thread they are coming from.
As far as this PR, I'm ok with leaving it empty, use numeric thread id, or use combo of name/numeric id.

azurelinuxagent/ga/monitor.py
vrdmr: Making it static wasn't easy as the function deals with `event` rather than an `array`. I found a better way of iterating tuples and using it now. Updates.

narrieta: if you prefer to postpone this post-release that would work too

but we should refactor this code

azurelinuxagent/common/event.py
vrdmr: Okay. We can take a new action item once we start seeing the usage of the new telemetry and usefulness of the new fields.

azurelinuxagent/common/event.py
```diff
@@ -335,35 +334,34 @@ def add_metric(self, category, counter, instance, value, log_event=False):
...
         # the event, rather than sending the event.
-        event.parameters.append(TelemetryEventParam('ContainerId', get_container_id_from_env()))
+        # OpcodeName: This is used as the actual time of event generation.
 
-        # This is used as the actual time of event generation.
-        event.parameters.append(TelemetryEventParam('OpcodeName', datetime.utcnow().__str__()))
+        default_parameters = [("GAVersion", CURRENT_AGENT), ('ContainerId', get_container_id_from_env()),
+                              ('OpcodeName', datetime.utcnow().__str__()),
+                              ('EventTid', threading.current_thread().ident),
+                              ('EventPid', os.getpid()), ("TaskName", threading.current_thread().getName()),
```
narrieta: I thought we wouldn't use TaskName?

tests/ga/test_monitor.py
```diff
@@ -174,6 +170,7 @@ def test_add_sysinfo_should_honor_sysinfo_values_from_agent_for_agent_events(sel
 
         counter = 0
         for p in event.parameters:
+            print(p.name, p.value)
```
pgombar: Remove?

PR #1666, opened 2019-10-11
Refactoring reset_logger to reset every 12 hours

vrdmr:
## Description
Refactoring the reset loggers to bring it to the MonitorHandler level. The logger now resets every 12 hours.

Issue # 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1666)

### DIFF
--- a/azurelinuxagent/ga/monitor.py
+++ b/azurelinuxagent/ga/monitor.py
@@ -114,12 +114,16 @@ class MonitorHandler(object):
     IMDS_HEARTBEAT_PERIOD = datetime.timedelta(minutes=1)
     IMDS_HEALTH_PERIOD = datetime.timedelta(minutes=3)
 
+    # Resetting loggers period
+    RESET_LOGGERS_PERIOD = datetime.timedelta(hours=12)
+
     def __init__(self):
         self.osutil = get_osutil()
         self.protocol_util = get_protocol_util()
         self.imds_client = get_imds_client()
 
         self.event_thread = None
+        self.last_reset_loggers_time = None
         self.last_event_collection = None
         self.last_telemetry_heartbeat = None
         self.last_cgroup_polling_telemetry = None
@@ -277,7 +281,24 @@ class MonitorHandler(object):
             self.send_host_plugin_heartbeat()
             self.send_imds_heartbeat()
             self.log_altered_network_configuration()
+            self.reset_loggers()
             time.sleep(min_delta)
+
+    def reset_loggers(self):
+        """
+        The loggers maintain hash-tables in memory and they need to be cleaned up from time to time.
+        For reference, please check azurelinuxagent.common.logger.Logger and
+        azurelinuxagent.common.event.EventLogger classes
+        """
+        time_now = datetime.datetime.utcnow()
+        if not self.last_reset_loggers_time:
+            self.last_reset_loggers_time = time_now
+
+        if time_now >= (self.last_reset_loggers_time +
+                        MonitorHandler.RESET_LOGGERS_PERIOD):
+            logger.reset_periodic()
+
+            self.last_reset_loggers_time = time_now
 
     def add_sysinfo(self, event):
         sysinfo_names = [v.name for v in self.sysinfo]
--- a/azurelinuxagent/ga/monitor.py
+++ b/azurelinuxagent/ga/monitor.py
@@ -296,9 +296,10 @@ class MonitorHandler(object):
 
         if time_now >= (self.last_reset_loggers_time +
                         MonitorHandler.RESET_LOGGERS_PERIOD):
-            logger.reset_periodic()
-
-            self.last_reset_loggers_time = time_now
+            try:
+                logger.reset_periodic()
+            finally:
+                self.last_reset_loggers_time = time_now
 
     def add_sysinfo(self, event):
         sysinfo_names = [v.name for v in self.sysinfo]
--- a/tests/ga/test_monitor.py
+++ b/tests/ga/test_monitor.py
@@ -279,32 +279,16 @@ class TestMonitor(AgentTestCase):
         self.assertEqual(False, args[5].call_args[1]['is_success'])
         monitor_handler.stop()
 
-    @patch("azurelinuxagent.ga.monitor.MonitorHandler.send_telemetry_heartbeat")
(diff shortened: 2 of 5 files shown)


tests/ga/test_monitor.py
```diff
@@ -30,6 +34,13 @@
 from tests.tools import *
 
 
+_MSG_INFO = "This is our test info logging message {0} {1}"
+_MSG_WARN = "This is our test warn logging message {0} {1}"
+_MSG_ERROR = "This is our test error logging message {0} {1}"
+_MSG_VERBOSE = "This is our test verbose logging message {0} {1}"
+_DATA = ["arg1", "arg2"]
+
```
vrdmr: This is not used anymore. Will remove this code. More exhaustive testing of the periodic_* is done in test_logger.

vrdmr:
*Important*: Currently, the hash map is shared across levels and won't generate a new entry into the hashmap for the same messages with different log levels. This could suppress a newer log entry with a higher-level coming in later.

tests/ga/test_monitor.py
```diff
@@ -21,6 +21,9 @@
 from azurelinuxagent.common.cgroup import CGroup
 from azurelinuxagent.common.event import EventLogger
 from azurelinuxagent.common.datacontract import get_properties
+from azurelinuxagent.common.logger import DEFAULT_LOGGER
+from azurelinuxagent.common.protocol.imds import ComputeInfo, IMDS_IMAGE_ORIGIN_ENDORSED
+from azurelinuxagent.common.protocol.restapi import VMInfo
```
pgombar: Unused imports?

azurelinuxagent/ga/monitor.py
```diff
@@ -277,8 +281,25 @@ def daemon(self):
...
+        For reference, please check azurelinuxagent.common.logger.Logger and
+        azurelinuxagent.common.event.EventLogger classes
+        """
+        time_now = datetime.datetime.utcnow()
+        if not self.last_reset_loggers_time:
+            self.last_reset_loggers_time = time_now
+
+        if time_now >= (self.last_reset_loggers_time +
+                        MonitorHandler.RESET_LOGGERS_PERIOD):
+            logger.reset_periodic()
```
narrieta: I know this is the pattern we use today for all these periodic actions; we should consider if instead we should follow this:
```
    try:
        logger.reset_periodic()
   finally:
         self.last_reset_loggers_time = time_now
```
so that if there is an exception we don't keep calling the reset code continuously (to do after release)

tests/ga/test_monitor.py
```diff
@@ -276,6 +279,38 @@ def test_failed_heartbeat_creates_telemetry(self, patch_report_heartbeat, _, *ar
         self.assertEqual(False, args[5].call_args[1]['is_success'])
         monitor_handler.stop()
 
+    @patch('azurelinuxagent.common.logger.Logger.info')
+    def test_reset_loggers(self, mock_info, *args):
```
narrieta: good test

tests/ga/test_monitor.py
```diff
@@ -276,6 +279,38 @@ def test_failed_heartbeat_creates_telemetry(self, patch_report_heartbeat, _, *ar
...
+            logger.periodic_info(logger.EVERY_DAY, "Test-Message")
+
+        self.assertIn(hash("Test-Message"), logger.DEFAULT_LOGGER.periodic_messages)
+        self.assertEqual(101, mock_info.call_count)  # 100 calls from the previous section. Adding only 1.
+        self.assertEqual(101, len(logger.DEFAULT_LOGGER.periodic_messages))  # One new message in the hash map.
+
+        # Resetting the logger time states.
+        monitor_handler = get_monitor_handler()
+        monitor_handler.last_reset_loggers_time = datetime.datetime.utcnow() - timedelta(hours=1)
+        MonitorHandler.RESET_LOGGERS_PERIOD = timedelta(milliseconds=100)
```
pgombar: Not needed?

narrieta: +1, and this part of the test should be done directly on the logger, rather than on the logger via the monitor (plus move test to the logger tests)

narrieta: (could be done separately, in the interest of moving forward with the release)

azurelinuxagent/ga/monitor.py
larohra: +1

tests/ga/test_monitor.py
vrdmr: @pgombar 
It is needed. We are not using the `RESET_LOGGERS_PERIOD` for any sleep, but for the calculation to reset the logger. I am just moving the last time sent to 1 hour, but also need to force the calculation of time_now >= _now(-1hr) + 100ms_ to be true, to hit that code.

```
if time_now >= (self.last_reset_loggers_time + MonitorHandler.RESET_LOGGERS_PERIOD):
    logger.reset_periodic()
```

@narrieta: This part of the test is testing the monitor pieces of the code and not the logger pieces of the code. We already test the periodic_* in the test_loggers (eg: `test_periodic_does_not_emit_if_previously_sent`). Yes, there needs to be a test specifically for only `logger.reset_periodic()` in the test_logger, and that can be added later (will send a new PR later for that, against develop).

vrdmr:
@larohra: The `test_periodic_emits_after_elapsed_delta` (boh for logger and events) tries to test the case you are talking about. The values of the hash are changed to represent the time spent to test that piece of code.

tests/ga/test_monitor.py
narrieta: yup, my bad... good as it is now. thanks!

larohra:
@vrdmr I mainly meant a scenario of the case we faced today (where we expected logger to log after 6hrs but it was logging after 30mins). Its more of an end to end test to ensure that no other component is affecting our current scenario. Its more as a learning/follow up item from what we discovered today and to ensure we (or some new dev on this code) dont hit this same scenario even if they're unfamiliar with the code.

vrdmr:
@larohra - Oh. Now I see what you meant. Yup - agreed. We need to actually fix the logger hash-map impl.*, because, I feel that testing it should not happen would take more time than the actual fix.

*(with either a new way to reset which only prunes the logs once they are expired, and we store the expiration time as well)

PR #1677, opened 2019-10-22
use alternate systemd detection

narrieta:
* Replacing run_get_output with run_command (errors were being ignored, since run_get_output does not throw)
* The new logic to detect systemd uses the same check as systemd itself (e.g. systemd-run).

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1677)

### DIFF
--- a/tests/common/test_cgroupapi.py
+++ b/tests/common/test_cgroupapi.py
@@ -55,82 +55,40 @@ class CGroupsApiTestCase(AgentTestCase):
         self.assertTrue(type(api) == FileSystemCgroupsApi)
 
     def test_is_systemd_should_return_true_when_systemd_manages_current_process(self):
-        fileutil_read_file = fileutil.read_file
+        path_exists = os.path.exists
 
-        def mock_read_file(filepath, asbin=False, remove_bom=False, encoding='utf-8'):
-            if filepath == "/proc/cgroups":
-                return """
-#subsys_name	hierarchy	num_cgroups	enabled
-cpuset	11	1	1
-cpu	3	77	1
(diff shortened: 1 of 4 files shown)


narrieta:
DCR run OK

tests/common/test_cgroupapi.py
```diff
@@ -55,82 +55,40 @@ def test_create_should_return_a_FileSystemCgroupsApi_on_non_systemd_platforms(se
...
-"""
-            return fileutil_read_file(filepath, asbin=asbin, remove_bom=remove_bom, encoding=encoding)
-
-        with patch("azurelinuxagent.common.cgroupapi.fileutil.read_file", mock_read_file):
+        path_exists = os.path.exists
+
+        def mock_path_exists(path):
+            if path in ("/run/systemd/system/", "/run/systemd/system"):
+                mock_path_exists.path_tested = True
+                return True
```
narrieta: this should return False

tests/common/test_cgroupapi.py
```diff
@@ -55,82 +55,40 @@ def test_create_should_return_a_FileSystemCgroupsApi_on_non_systemd_platforms(se
...
+            return path_exists(path)
+
+        mock_path_exists.path_tested = False
+
+        with patch("azurelinuxagent.common.cgroupapi.os.path.exists", mock_path_exists):
             is_systemd = CGroupsApi._is_systemd()
 
-        self.assertFalse(is_systemd)
+        self.assertTrue(is_systemd)
```
narrieta: This should assert False

PR #1682, opened 2019-10-24
Linux extensions fast track

jscalev:
## Description

Issue # 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1682)


vrdmr:
@jscalev Py2 tests failing: https://travis-ci.org/Azure/WALinuxAgent/jobs/602579959.

azurelinuxagent/common/conf.py
```diff
@@ -276,6 +276,10 @@ def get_extensions_enabled(conf=__conf__):
     return conf.get_switch("Extensions.Enabled", True)
 
 
+def get_extensions_fast_track_enabled(conf=__conf__):
+    return conf.get_switch("Extensions.FastTrackEnabled", True)
```
vrdmr: @narrieta - We do need to sort the use of configuration and exposing it to the public or not.

narrieta: discussed offline. Adding this parameter is OK

config/waagent.conf
```diff
@@ -9,6 +9,9 @@ Provisioning.Enabled=y
 # backup, monitoring, or any extension handling whatsoever.
 Extensions.Enabled=y
 
+#Enable extensions FastTrack, which improves goal state processing times
```
narrieta: Let's document the setting in README.md

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -43,6 +43,8 @@
 HEADER_HOST_CONFIG_NAME = "x-ms-host-config-name"
 HEADER_ARTIFACT_LOCATION = "x-ms-artifact-location"
 HEADER_ARTIFACT_MANIFEST_LOCATION = "x-ms-artifact-manifest-location"
+HEADER_VERIFY_FROM_ARTIFACTS_BLOB = "x-ms-verify-from-artifacts-blob"
```
narrieta: Could we add  comments explaining what this header controls? I know the current headers don't have documentation, but let's do for the new ones,

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -136,11 +138,15 @@ def get_artifact_request(self, artifact_url, artifact_manifest_url=None):
         headers = {HEADER_VERSION: API_VERSION,
                    HEADER_CONTAINER_ID: self.container_id,
                    HEADER_HOST_CONFIG_NAME: self.role_config_name,
-                   HEADER_ARTIFACT_LOCATION: artifact_url}
+                   HEADER_ARTIFACT_LOCATION: artifact_url,
+                   HEADER_VERIFY_FROM_ARTIFACTS_BLOB: "true"}
```
narrieta: I see that this header is added unconditionally. Will Host GA Plugins that do not support Fast Track handle it correctly? (i.e. ignore it rather than choke on it)

azurelinuxagent/common/protocol/wire.py
```diff
@@ -147,14 +157,14 @@ def get_vmagent_pkgs(self, vmagent_manifest):
         valid_pkg_list = self.client.filter_package_list(vmagent_manifest.family, ga_manifest, goal_state)
         return valid_pkg_list
 
-    def get_ext_handlers(self):
+    def get_ext_conf(self):
```
narrieta: WireProtocol.get_ext_handlers is an override of Protocol.get_ext_handlers. This change would need to happen across the whole class hierarchy (Protocol and MetadataProtocol need to be updated)

azurelinuxagent/common/protocol/wire.py
```diff
@@ -59,16 +61,24 @@
...
 EXT_CONF_FILE_NAME = "ExtensionsConfig.{0}.xml"
+EXT_CONFIG_FAST_TRACK_FILE_NAME = "FastTrackExtensionsConfig.{0}.xml"
 MANIFEST_FILE_NAME = "{0}.{1}.manifest.xml"
 AGENTS_MANIFEST_FILE_NAME = "{0}.{1}.agentsManifest"
 TRANSPORT_CERT_FILE_NAME = "TransportCert.pem"
 TRANSPORT_PRV_FILE_NAME = "TransportPrivate.pem"
 # Store the last retrieved container id as an environment variable to be shared between threads for telemetry purposes
 CONTAINER_ID_ENV_VARIABLE = "AZURE_GUEST_AGENT_CONTAINER_ID"
 
+FAST_TRACK_EXTENSIONS_PATTERN = re.compile("^(FastTrackExtensionsConfig)\.(\d+)\.(xml)$", re.IGNORECASE)
```
narrieta: let's make this a raw string (add r as a prefix):
```
r"^(FastTrackExtensionsConfig)\.(\d+)\.(xml)$"
```

config/waagent.conf
narrieta: We talked offline about not adding the flag explicitly in the conf file. 

Actually, to make it easier for people to change it, it may be better to leave it, probaly adding to the comment something like "Do not disable this setting unless directed by Microsoft Support"?

azurelinuxagent/common/protocol/hostplugin.py
jscalev: Added documentation

azurelinuxagent/common/protocol/hostplugin.py
jscalev: Legacy HostGAPlugins will ignore it

narrieta: thanks

jscalev:
___
*[azurelinuxagent/common/protocol/wire.py, line 72 at r4](https://reviewable.io/reviews/azure/walinuxagent/1682#-Ls-CTkivCNRGwuSBSPc-r4-72:-Lt6FG0oAYliwQKOOctU:b-n9tkhd) ([raw file](https://github.com/azure/walinuxagent/blob/6c6c3d21e72a870b444cbe6913c65e7fe76f3184/azurelinuxagent/common/protocol/wire.py#L72)):*
<details><summary><i>Previously, narrieta (Norberto Arrieta) wrote…</i></summary><blockquote>

let's make this a raw string (add r as a prefix):
```
r"^(FastTrackExtensionsConfig)\.(\d+)\.(xml)$"
```
</blockquote></details>

changed

azurelinuxagent/ga/exthandlers.py
```diff
@@ -276,6 +288,73 @@ def run(self):
...
+                      message=detailed_msg)
+
+    def process_goal_state(self, instantiation, sequence_number):
+        fabric_goal_state_changed, fast_track_changed = self.determine_what_changed(instantiation, sequence_number)
+        if fabric_goal_state_changed:
+            self.protocol.set_fast_track(False)
+
+        # Fabric goal state takes precedence over VMArtifactsProfile blob goal state
+        # So, process the Fabric goal state first
+        self.handle_ext_handlers(fabric_goal_state_changed)
```
vrdmr: Why not call `handle_ext_handlers` only when things change? The advantage would be to save the compute cycles as almost nothing is being done when it is False.

```
if fabric_goal_state_changed:
    self.handle_ext_handlers(fabric_goal_state_changed)
```

azurelinuxagent/ga/exthandlers.py
```diff
@@ -276,6 +288,73 @@ def run(self):
...
+        self.last_instantiation = instantiation
+
+        # Now process the VMArtifactsProfile if it changed. Note that just because the VMArtifactsProfile changed
+        # doesn't mean the FastTrack goal state changed
+        if fast_track_changed and self.artifacts_profile is not None:
```
vrdmr: Can be written as `if fast_track_changed and self.artifacts_profile:`.

`if x is not None` == `if x`

azurelinuxagent/ga/exthandlers.py
```diff
@@ -276,6 +288,73 @@ def run(self):
...
+            self.process_vm_artifacts_profile_for_fast_track(sequence_number)
+        elif self.protocol.client.get_use_fast_track():
+            # Our last goal state was via VMArtifactsProfile, so reset the extension handlers
+            self.ext_conf = self.protocol.client.get_ext_conf(force=True)
+            self.ext_handlers = self.ext_conf.ext_handlers
+
+    def determine_what_changed(self, instantiation, sequence_number):
```
vrdmr: Can you split this method into two methods, with one explicit task each? Something on the lines of `did_fabric_goal_state_change` and `did_new_settings_came_in`? It'll be easier to handle them when testing and debugging as well.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -276,6 +288,73 @@ def run(self):
...
+            logger.warn(msg)
+
+            add_event(AGENT_NAME,
+                      version=CURRENT_VERSION,
+                      op=WALAEventOperation.ExtensionProcessing,
+                      is_success=False,
+                      message=detailed_msg)
+
+    def process_goal_state(self, instantiation, sequence_number):
+        fabric_goal_state_changed, fast_track_changed = self.determine_what_changed(instantiation, sequence_number)
```
vrdmr: Addl. comments below in the definition of `determine_what_changed`.
```
fabric_goal_state_changed = self.determine_fabric_change(instantiation)
fast_track_changed = self.determine_fast_track_change(instantiation)
```

azurelinuxagent/ga/exthandlers.py
```diff
@@ -276,6 +288,73 @@ def run(self):
...
+        if fabric_goal_state_changed:
+            self.protocol.set_fast_track(False)
```
vrdmr: Isn't sequence number per extension? Then why is `sequence_number_file_path = os.path.join(conf.get_lib_dir(), SEQUENCE_NUMBER_FILE_NAME)` at lib level, and not inside extension level (in extension folders or some other place.)

azurelinuxagent/ga/exthandlers.py
```diff
@@ -276,6 +288,73 @@ def run(self):
...
+            vm_artifacts_ext_conf.status_upload_blob_type = self.ext_conf.status_upload_blob_type
+            if vm_artifacts_ext_conf.ext_handlers is not None and \
+                    len(vm_artifacts_ext_conf.ext_handlers.extHandlers) > 0:
+                logger.info('FastTrack goal state changed.')
+                self.protocol.set_fast_track(True, sequence_number)
+
+                # Reset the extension config and reload it, which will load it with the FastTrack extensions
+                self.ext_conf = self.protocol.client.get_ext_conf(force=True)
+                self.ext_handlers = self.ext_conf.ext_handlers
+                self.handle_ext_handlers(True)
```
vrdmr: Please add a comment - faking a goal_state_change here, by passing True.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -873,34 +926,61 @@ def get_certs(self):
...
+    def get_use_fast_track(self):
+        path = os.path.join(conf.get_lib_dir(), GOAL_STATE_SOURCE_FILE_NAME)
+        if os.path.exists(path):
+            goal_state_source = fileutil.read_file(path)
+            if goal_state_source == GOAL_STATE_SOURCE_FASTTRACK:
+                return True
+        else:
+            return False
+
+    def get_ext_conf(self, force=False):
```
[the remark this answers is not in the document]
vrdmr: Also, please add method-only tests which test the specific nature of this method.

Eg: I started adding tests for ext_conf [here](https://github.com/Azure/WALinuxAgent/pull/1691/files?file-filters%5B%5D=.py#diff-67c701dcf21f5a40ca2b611588d0e235R584) and you can pick up this to build something similar.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -276,6 +288,73 @@ def run(self):
...
+
+        if sequence_number is not None and self.last_vm_artifacts_seqno != sequence_number and \
+                conf.get_extensions_fast_track_enabled():
+            logger.info('VMArtifactsProfile changed with sequence number={0}(last:{1})'.
+                        format(str(sequence_number), str(self.last_vm_artifacts_seqno)))
+            fast_track_changed = True
+        return fabric_goal_state_changed, fast_track_changed
+
+    def process_vm_artifacts_profile_for_fast_track(self, sequence_number):
+        vm_artifacts_ext_conf = self.artifacts_profile.transform_to_extensions_config()
```
vrdmr: You can move this transformation earlier in the call stack.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1811,3 +1934,98 @@ def is_on_hold(self):
...
+        </Plugin>
+        <Plugin name="Microsoft.Flipperdoodle" version="1.0.0">
+            <RuntimeSettings seqNo="0">{"runtimeSettings":[{"handlerSettings":{"protectedSettingsCertThumbprint":"4037FBF5F1F3014F99B5D6C7799E9B20E6871CB3","protectedSettings":"MIICWgYJK","publicSettings":{"foo":"bar"}}}]}</RuntimeSettings>
+        </Plugin>
+    </PluginSettings>
+    <StatusUploadBlob statusBlobType="BlockBlob">https://putblobhere</StatusUploadBlob>
+    </Extensions>
+    """
+    def transform_to_extensions_config(self):
+        if 'extensionGoalStates' in self.__dict__:
```
vrdmr: Move the above comments into this methods pydoc.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -253,15 +258,22 @@ def run(self):
...
             logger.verbose(msg)
             # Log status report success on new config
             self.log_report = True
 
             if self.extension_processing_allowed():
-                self.handle_ext_handlers(etag)
-                self.last_etag = etag
+                sequence_number = None
+                if self.artifacts_profile is not None and conf.get_extensions_fast_track_enabled():
+                    sequence_number = self.artifacts_profile.get_sequence_number()
```
vrdmr:
```
sequence_number = None
if self.artifacts_profile is not None and conf.get_extensions_fast_track_enabled():
    sequence_number = self.artifacts_profile.get_sequence_number()
```
Does computing `sequence_number` help and used? Else push it into the `process_goal_state` method. Keeps it cleaner.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -214,20 +214,25 @@ class ExtHandlersHandler(object):
...
         self.get_artifact_error_state = ErrorState(min_timedelta=ERROR_STATE_DELTA_INSTALL)
 
     def run(self):
-        self.ext_handlers, etag = None, None
+        self.ext_handlers, instantiation = None, None
         try:
             self.protocol = self.protocol_util.get_protocol()
-            self.ext_handlers, etag = self.protocol.get_ext_handlers()
+            self.ext_conf, instantiation = self.protocol.get_ext_conf()
+            self.ext_handlers = self.ext_conf.ext_handlers
```
vrdmr: Given a choice, I would put in a method call here to a method which readies all the things needed to check if the fast track things are in place or not. Rather than dividing that piece into `get_artifacts_profile`, `get_sequence_number` and `process_vm_artifacts_profile_for_fast_track`. Most of the things done in `process_vm_artifacts_profile_for_fast_track` could have been done beforehand.

If your concern was the unnecessary processing done by us if there are no new settings coming in, then we can discuss and find a better way.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -214,20 +214,25 @@ class ExtHandlersHandler(object):
...
         self.report_status_error_state = ErrorState()
         self.get_artifact_error_state = ErrorState(min_timedelta=ERROR_STATE_DELTA_INSTALL)
 
     def run(self):
-        self.ext_handlers, etag = None, None
+        self.ext_handlers, instantiation = None, None
         try:
             self.protocol = self.protocol_util.get_protocol()
-            self.ext_handlers, etag = self.protocol.get_ext_handlers()
+            self.ext_conf, instantiation = self.protocol.get_ext_conf()
```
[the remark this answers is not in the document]
jscalev: Renamed to incarnation

azurelinuxagent/ga/exthandlers.py
jscalev: I would love to do this, but I'm concerned about modifying the existing behavior. Currently it isn't until handle_ext_handler that we determine whether a goal state occurred. Since this code is for the fabric goal state, I didn't want to change that behavior.

azurelinuxagent/ga/exthandlers.py
jscalev: Is this a strong coding guideline? I understand they're the same, but personally I prefer to be obvious - especially since Python has a habit of "guessing" to sometimes disastrous results.

narrieta: The explicit comparison against None is preferred.

azurelinuxagent/ga/exthandlers.py
jscalev: Split into two methods

azurelinuxagent/common/protocol/wire.py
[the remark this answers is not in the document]
jscalev: Renamed force to read_always and added method only tests

azurelinuxagent/ga/exthandlers.py
jscalev: ok, moved up one

azurelinuxagent/common/protocol/wire.py
jscalev: moved

azurelinuxagent/ga/exthandlers.py
jscalev: Yes. It's used in the next line when processing the goal state.

azurelinuxagent/common/protocol/wire.py
jscalev: Moved across hierarchy

azurelinuxagent/common/protocol/wire.py
jscalev: Changed

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -43,6 +43,13 @@
 HEADER_HOST_CONFIG_NAME = "x-ms-host-config-name"
 HEADER_ARTIFACT_LOCATION = "x-ms-artifact-location"
 HEADER_ARTIFACT_MANIFEST_LOCATION = "x-ms-artifact-manifest-location"
+
+# Indicates HostGAPlugin should download the VMArtifactsProfile blob
+# if it hasn't already done so to check if the manifest url is whitelisted
+HEADER_VERIFY_FROM_ARTIFACTS_BLOB = "x-ms-verify-from-artifacts-blob"
+
+# RFC 7232, section 3.2
+HEADER_IF_NONE_MATCH = "if-none-match"
```
larohra: Can you explain a bit more as to what this does? Thanks!

larohra: Saw the RFC, thanks for pointing it out :)

azurelinuxagent/common/protocol/wire.py
```diff
@@ -51,6 +51,8 @@
 
 WIRE_SERVER_ADDR_FILE_NAME = "WireServer"
 INCARNATION_FILE_NAME = "Incarnation"
+SEQUENCE_NUMBER_FILE_NAME = "SequenceNumber"
```
larohra: Since the extensions also have a `SequenceNumber`, could you rename this to something like `artifactProfileSeqNo` or something similar to avoid any unnecessary ambiguity

azurelinuxagent/common/protocol/wire.py
```diff
@@ -873,34 +926,61 @@ def get_certs(self):
...
-            # cache file is missing, nothing to do
-            logger.verbose(ustr(pe))
-        except Exception as e:
-            logger.error("Could not obtain current handlers: {0}", ustr(e))
-
-        return handler_list
+    def get_sequence_number(self):
+        path = os.path.join(conf.get_lib_dir(), SEQUENCE_NUMBER_FILE_NAME)
+        if os.path.exists(path):
+            return int(fileutil.read_file(path))
```
larohra: `fileutil.read_file` can also return a None in certain scenarios (if there is some error in reading the file). Casting `int(None)` would throw an exception. you should handle that case

azurelinuxagent/common/protocol/wire.py
```diff
@@ -873,34 +926,61 @@ def get_certs(self):
...
+        if goal_state.ext_uri is None:
+            return None
+        local_file = EXT_CONF_FILE_NAME.format(goal_state.incarnation)
+        local_file = os.path.join(conf.get_lib_dir(), local_file)
+        return local_file
+
+    def get_file_path_for_fast_track(self):
+        local_file = None
+        sequence_number = self.get_sequence_number()
+        if sequence_number > 0:
```
larohra: Not sure how the extension config sequence number works, but can it ever be 0? If so then maybe you can change the default in function `get_sequence_number` to a negative number?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -995,6 +1075,41 @@ def check_wire_protocol_version(self):
...
+        # supports the If-None-Match header that reduces how often we need to parse
+
+        # note that this method requires a lambda that returns two values
+        ret = None
+        etag = None
+        not_modified = False
+        try:
+            ret, etag, not_modified = host_func()
+
+            # Different direct channel functions report failure in different ways: by returning None, False,
```
larohra: Not relevant to this PR, but we should make all the direct_func and host_func more consistent to either use try-catch or the return value for code cleanliness and consistency

azurelinuxagent/common/protocol/wire.py
```diff
@@ -995,6 +1075,41 @@ def check_wire_protocol_version(self):
...
+        if ret or not_modified:
+            return ret, etag
+
+        try:
+            ret = direct_func()
+        except (ResourceGoneError, InvalidContainerError) as e:
+            logger.info("Request failed with direct. Error: {0}".format(ustr(e)))
+
+        if not ret:
+            # Try through the host one more time
```
larohra: Given that we call this every 3 secs and that our restutil lib has its own retry logic, the 3rd retry might cause a lot of time delay in case of some network failure (not to mention the verbosity in our logs). I wonder if it would cause more harm than good?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -995,6 +1075,41 @@ def check_wire_protocol_version(self):
...
+        ret = None
+        etag = None
+        not_modified = False
+        try:
+            ret, etag, not_modified = host_func()
+
+            # Different direct channel functions report failure in different ways: by returning None, False,
+            # or raising ResourceGone or InvalidContainer exceptions.
+            if not ret and not not_modified:
+                logger.info("Request failed using the HostGAPlugin. switching to direct.")
```
larohra: `logger.periodic_info` to avoid verbosity in our logs

azurelinuxagent/common/protocol/wire.py
```diff
@@ -995,6 +1075,41 @@ def check_wire_protocol_version(self):
...
+        except (ResourceGoneError, InvalidContainerError) as e:
+            msg = "Request failed with the current host plugin configuration." \
+                  "ContainerId: {0}, role config file: {1}." \
+                  "Error: {2}".format(self.host_plugin.container_id, self.host_plugin.role_config_name, ustr(e))
+            logger.info(msg)
```
larohra: logger.periodic_info to avoid verbosity in our logs

azurelinuxagent/common/protocol/wire.py
```diff
@@ -995,6 +1075,41 @@ def check_wire_protocol_version(self):
...
+
+        if ret or not_modified:
+            return ret, etag
+
+        try:
+            ret = direct_func()
+        except (ResourceGoneError, InvalidContainerError) as e:
+            logger.info("Request failed with direct. Error: {0}".format(ustr(e)))
```
larohra: logger.periodic_info to avoid verbosity in our logs

azurelinuxagent/common/utils/archive.py
```diff
@@ -63,34 +69,49 @@ def __init__(self, lib_dir):
                     logger.error("{0} : {1}", self._source, e.strerror)
 
     def flush(self, timestamp):
-        files = self._get_files_to_archive()
-        if len(files) == 0:
+        copy_files, move_files = self._get_files_to_archive()
```
larohra: The archive logic is called if the `update_goal_state` is not forced and if the incarnation changes. If the FastTrack extension config is modified then we might not capture it in our archive logic (which helps a lot with debugging). Before deleting the files in `save_config in wire.py (line:2012)`, you should call this function to make sure that we're capturing the FastTrackExtensionConfig too before deleting the files.

azurelinuxagent/common/utils/archive.py
```diff
@@ -38,12 +38,18 @@
...
 MAX_ARCHIVED_STATES = 50
 
+# These files we'll move - deleting them in the old location
 CACHE_PATTERNS = [
     re.compile("^(.*)\.(\d+)\.(agentsManifest)$", re.IGNORECASE),
     re.compile("^(.*)\.(\d+)\.(manifest\.xml)$", re.IGNORECASE),
     re.compile("^(.*)\.(\d+)\.(xml)$", re.IGNORECASE)
 ]
 
+# These files we'll copy only and leave them in the original location
```
larohra: If you call this function everytime the FastTrackExtensionConfig is modified then you can move these files from COPY to MOVE too

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1811,3 +1934,98 @@ def is_on_hold(self):
...
+
+            config_xml = xml.tostring(root).decode()
+            self.save_config(config_xml)
+            extensions_config = ExtensionsConfig(config_xml)
+
+            return extensions_config
+        return None
+
+    def save_config(self, config_xml):
+        # before we write the new file, remove all the old ones
```
larohra: If you call the archive function here `self.goal_state_flusher.flush(datetime.utcnow())` you dont have to implement a separate file_deletion logic

azurelinuxagent/ga/exthandlers.py
```diff
@@ -276,6 +288,78 @@ def run(self):
...
+        if sequence_number is not None and self.last_vm_artifacts_seqno != sequence_number and \
+                conf.get_extensions_fast_track_enabled():
+            logger.info('VMArtifactsProfile changed with sequence number={0}(last:{1})'.
+                        format(str(sequence_number), str(self.last_vm_artifacts_seqno)))
+            fast_track_changed = True
+        return fast_track_changed
+
+    def process_vm_artifacts_profile_for_fast_track(self, sequence_number, vm_artifacts_ext_conf):
+        if vm_artifacts_ext_conf is not None:
+            # We need to add the status upload blob from Fabric goal state because we don't receive this
```
larohra: I have concerns about this, what would happen if status blob is updated but there was no new fabric GoalState generated so `self.ext_conf.status_upload_blob` would point to the old status blob and not the new one and then we get a new ExtensionConfig through fasttrack?

larohra: Can we not make it more resilient by adding the status blob as part of the ArtifactBlob we get from FastTrack too to ensure that its always the latest?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -276,6 +288,78 @@ def run(self):
...
+            # from the VMArtifactsProfile blob
+            vm_artifacts_ext_conf.status_upload_blob = self.ext_conf.status_upload_blob
+            vm_artifacts_ext_conf.status_upload_blob_type = self.ext_conf.status_upload_blob_type
+            if vm_artifacts_ext_conf.ext_handlers is not None and \
+                    len(vm_artifacts_ext_conf.ext_handlers.extHandlers) > 0:
+                logger.info('FastTrack goal state changed.')
+                self.protocol.set_fast_track(True, sequence_number)
+
+                # Reset the extension config and reload it, which will load it with the FastTrack extensions
+                self.ext_conf = self.protocol.client.get_ext_conf(read_always=True)
```
larohra: Please correct me if I'm wrong, but all this function `self.protocol.client.get_ext_conf(read_always=True)` is doing is reading again from the ExtensionConfig file and setting the `self.protocol.client.ext_conf` to the new ExtensionConfig. But isin't the `vm_artifacts_ext_conf` same thing? Why are you reading the same config again from the file rather than just using the same object `vm_artifacts_ext_conf`?

azurelinuxagent/ga/update.py
```diff
@@ -341,7 +342,17 @@ def run(self, debug=False):
                         op=WALAEventOperation.ProcessGoalState,
                         duration=duration,
                         message="Incarnation {0}".format(exthandlers_handler.last_etag))
-
+                if last_vm_artifacts_seqno != exthandlers_handler.last_vm_artifacts_seqno:
+                    self._ensure_readonly_files()
+                    duration = elapsed_milliseconds(utc_start)
```
larohra: Apart from the Prefix - `Incarnation` vs `SeqNo`, everything from line 345 to 355 is the same as 335 to 344. 
In an effort to reduce code redundancy as much as we can, could you please make this into a small private/lambda function and pass the prefix as a parameter. Thanks!

config/waagent.conf
larohra: I dont feel we should add this flag at all here, as this feature doesnt impact customers directly at all, its just modifying the internal functioning of the whole goalstate model. We shouldn't give this power to the customer

larohra:
___
*[azurelinuxagent/common/protocol/wire.py, line 1108 at r13](https://reviewable.io/reviews/azure/walinuxagent/1682#-Ls-CTkivCNRGwuSBSPc-r13-1108:-Lu9r0J4A27a-Y0Hm2Km:bwaq4pz) ([raw file](https://github.com/azure/walinuxagent/blob/c79be1e61b7d24ca80500934c85d2867bbc49bd4/azurelinuxagent/common/protocol/wire.py#L1108)):*
<details><summary><i>Previously, jscalev (Joseph Calev) wrote…</i></summary><blockquote>

So for the Fabric case, we go direct first, then try through HostGAPlugin twice.
In this case, we try through HostGAPlugin once, then try direct, then try HostGAPlugin again.
Given the same numbers for both failure cases, and the fact that we already call this every 3 secs, I think we would have seen issues already.
</blockquote></details>

In the fabric case, before the 3rd call we're refreshing GS again. In this scenario we're not changing anything and just re-sending the same request again. Not sure if this would make anything better or not. I wonder if refreshing GoalState before the 3rd retry would help in this case?

jscalev:
___
*[azurelinuxagent/ga/exthandlers.py, line 352 at r13](https://reviewable.io/reviews/azure/walinuxagent/1682#-Ls-CTkivCNRGwuSBSPd-r13-352:-LxcPFJx78_ZpHIalE5T:b-sjsilj) ([raw file](https://github.com/azure/walinuxagent/blob/c79be1e61b7d24ca80500934c85d2867bbc49bd4/azurelinuxagent/ga/exthandlers.py#L352)):*
<details><summary><i>Previously, larohra wrote…</i></summary><blockquote>

I see that you are verifying if FastTrack extensionConfig changed by checking if it has any ext_handlers. Is this the expected behavior from CRP end too? Will there only be ExtHandlers in the ArtifactBlob only when extHandlers are modified from CRP end?
</blockquote></details>

There will only be extHandlers in the VmArtifactsProfile blob for a FastTrack change. We can receive a new blob outside of FastTrack because it contains other information (in particular OnHold - which the code already reads). So just because we receive a new VmArtifactsProfile blob doesn't mean we have a new FastTrack state.

jscalev:
___
*[azurelinuxagent/ga/exthandlers.py, line 336 at r13](https://reviewable.io/reviews/azure/walinuxagent/1682#-Ls-CTkivCNRGwuSBSPd-r13-336:-LxcOfjm5SUp2Wj-CvkA:b-usrktc) ([raw file](https://github.com/azure/walinuxagent/blob/c79be1e61b7d24ca80500934c85d2867bbc49bd4/azurelinuxagent/ga/exthandlers.py#L336)):*
<details><summary><i>Previously, larohra wrote…</i></summary><blockquote>

This function is checking if the VmArtifactBlob has changed, not necessarily if fast_track has changed (how I understand it is that fast_track_change means a new Extension_config was received).
Could we please rename it to say `VmArtifactBlob_changed` instead?
</blockquote></details>

It's a bit of a grey line in wording. We need three things to say "FastTrack" was changed
1) A new VMArtifactsProfile blob
2) A new sequence number in that blob
3) FastTrack must be enabled (in the config and because HostGAPlugin supports it)

This method checks all three of those. Added a comment with details.

jscalev:
___
*[azurelinuxagent/ga/exthandlers.py, line 273 at r13](https://reviewable.io/reviews/azure/walinuxagent/1682#-Ls-CTkivCNRGwuSBSPd-r13-273:-LxcHix6-lvfvtkLwirQ:bb7f2mf) ([raw file](https://github.com/azure/walinuxagent/blob/c79be1e61b7d24ca80500934c85d2867bbc49bd4/azurelinuxagent/ga/exthandlers.py#L273)):*
<details><summary><i>Previously, larohra wrote…</i></summary><blockquote>

NIT: Rename it to artifact_seq_no
</blockquote></details>

done

jscalev:
___
*[azurelinuxagent/ga/exthandlers.py, line 306 at r11](https://reviewable.io/reviews/azure/walinuxagent/1682#-Ls-CTkivCNRGwuSBSPd-r11-306:-LxcOWYBD6cqHCxhBYXz:b-896fix) ([raw file](https://github.com/azure/walinuxagent/blob/5e6ce7b0b10281b23495f46fe976b60796066054/azurelinuxagent/ga/exthandlers.py#L306)):*
<details><summary><i>Previously, vrdmr (Varad Meru) wrote…</i></summary><blockquote>

Addl. comments below in the definition of `determine_what_changed`.
```
fabric_goal_state_changed = self.determine_fabric_change(instantiation)
fast_track_changed = self.determine_fast_track_change(instantiation)
```
</blockquote></details>

Done.

jscalev:
___
*[azurelinuxagent/ga/exthandlers.py, line 235 at r11](https://reviewable.io/reviews/azure/walinuxagent/1682#-Ls-CTkivCNRGwuSBSPd-r11-235:-LxcGA8HBBTWpgR-2awv:bawtg9i) ([raw file](https://github.com/azure/walinuxagent/blob/5e6ce7b0b10281b23495f46fe976b60796066054/azurelinuxagent/ga/exthandlers.py#L235)):*
<details><summary><i>Previously, larohra wrote…</i></summary><blockquote>

+1, it would make code readability much easier for sure
</blockquote></details>

I'm not sure how I can do this clearly. Determining whether we can and should use FastTrack in an iteration is tricky.

jscalev:
___
*[azurelinuxagent/ga/exthandlers.py, line 348 at r13](https://reviewable.io/reviews/azure/walinuxagent/1682#-Ls-CTkivCNRGwuSBSPd-r13-348:-LxcQ1ta9ocOm-fv4MST:b-dbfshb) ([raw file](https://github.com/azure/walinuxagent/blob/c79be1e61b7d24ca80500934c85d2867bbc49bd4/azurelinuxagent/ga/exthandlers.py#L348)):*
<details><summary><i>Previously, larohra wrote…</i></summary><blockquote>

Can we not make it more resilient by adding the status blob as part of the ArtifactBlob we get from FastTrack too to ensure that its always the latest?
</blockquote></details>

These are different statuses and may be conflicting. We either receive GoalState from FastTrack or from Fabric. Only one is the latest and the other will contain outdated status.

jscalev:
___
*[azurelinuxagent/ga/exthandlers.py, line 358 at r13](https://reviewable.io/reviews/azure/walinuxagent/1682#-Ls-CTkivCNRGwuSBSPd-r13-358:-LxcRarY6dXF289jHc7J:b34o58y) ([raw file](https://github.com/azure/walinuxagent/blob/c79be1e61b7d24ca80500934c85d2867bbc49bd4/azurelinuxagent/ga/exthandlers.py#L358)):*
<details><summary><i>Previously, larohra wrote…</i></summary><blockquote>

Please correct me if I'm wrong, but all this function `self.protocol.client.get_ext_conf(read_always=True)` is doing is reading again from the ExtensionConfig file and setting the `self.protocol.client.ext_conf` to the new ExtensionConfig. But isin't the `vm_artifacts_ext_conf` same thing? Why are you reading the same config again from the file rather than just using the same object `vm_artifacts_ext_conf`? 
</blockquote></details>

I'm calling get_ext_conf to set several properties automatically. It's part of the trickery to have the rest of the code not care whether the changes are coming from FastTrack or Fabric. I'd rather not do this, but I wanted to minimize the impact.

jscalev:
___
*[azurelinuxagent/ga/update.py, line 347 at r13](https://reviewable.io/reviews/azure/walinuxagent/1682#-Ls-CTkivCNRGwuSBSPe-r13-347:-Lxcb3VC2Zc8Bki9B7ch:b-n9tkhd) ([raw file](https://github.com/azure/walinuxagent/blob/c79be1e61b7d24ca80500934c85d2867bbc49bd4/azurelinuxagent/ga/update.py#L347)):*
<details><summary><i>Previously, larohra wrote…</i></summary><blockquote>

Apart from the Prefix - `Incarnation` vs `SeqNo`, everything from line 345 to 355 is the same as 335 to 344. 
In an effort to reduce code redundancy as much as we can, could you please make this into a small private/lambda function and pass the prefix as a parameter. Thanks!  
</blockquote></details>

changed

azurelinuxagent/common/protocol/wire.py
```diff
@@ -661,6 +694,23 @@ def stream(self, uri, destination, headers=None, use_proxy=None):
 
         return success
 
+    def fetch_content_and_etag(self, uri, headers=None, use_proxy=None, decode=True):
```
larohra: NIT: Can you rename this function to be more clear as to what it does - `fetch_content_and_etag_if_modified`.
Also can you make it private?

larohra: Also in this scenario (as per the code), `etag` will be not None only if `not_modified` is True right? Can't we just use this knowledge to deduce if it was modified or not instead of passing the `not_modified` variable?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -873,34 +926,62 @@ def get_certs(self):
...
-                                                                      EXT_CONF_FILE_NAME.format(incarnation))))
-            handler_list = ext_conf.ext_handlers.extHandlers
-        except ProtocolError as pe:
-            # cache file is missing, nothing to do
-            logger.verbose(ustr(pe))
-        except Exception as e:
-            logger.error("Could not obtain current handlers: {0}", ustr(e))
-
-        return handler_list
+    def get_sequence_number(self):
```
larohra: NIT: rename to `get_artifacts_seq_no`

azurelinuxagent/common/protocol/wire.py
larohra: Also, shouldn't the name be `modified` instead of `not_modified`? It just seems counter intuitive to see `not_modified=False` when it actually means `modified=False`

narrieta:
@jscalev - is it ok to close this one now? thanks

jscalev:
Yes

PR #1683, opened 2019-10-25
Improve logging and test around systemd fallback logic

pgombar:
## Description

Explicitly log when systemd extension invocation failed and we are about to attempt falling back to a regular subprocess.Popen. Also, a test case is added that mimics a systemd timeout.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1683)

### DIFF
--- a/azurelinuxagent/common/cgroupapi.py
+++ b/azurelinuxagent/common/cgroupapi.py
@@ -510,6 +510,8 @@ After=system-{1}.slice""".format(extension_name, EXTENSIONS_ROOT_CGROUP_NAME)
                 stderr.truncate(0)
 
                 # Try invoking the process again, this time without systemd-run
+                logger.info('Extension invocation using systemd failed, falling back to regular invocation '
+                            'without cgroups tracking.')
                 process = subprocess.Popen(command,
                                            shell=shell,
                                            cwd=cwd,
--- a/tests/common/test_cgroupapi.py
+++ b/tests/common/test_cgroupapi.py
@@ -506,9 +506,9 @@ class SystemdCgroupsApiTestCase(AgentTestCase):
                         self.assertEquals(extension_cgroups, [])
 
     @patch('time.sleep', side_effect=lambda _: mock_sleep(0.001))
-    def test_start_extension_command_should_use_fallback_option_if_systemd_times_out_internally(self, _):
-        # Systemd has its own internal timeout which is not necessarily the same as the timeout we define for
-        # extension operations. When systemd times out, it will write that to stderr and exit with exit code 1.
+    def test_start_extension_command_should_use_fallback_option_if_systemd_times_out(self, _):
+        # Systemd has its own internal timeout which is shorter than what we define for extension operation timeout.
+        # When systemd times out, it will write a message to stderr and exit with exit code 1.
         # In that case, we will internally recognize the failure due to the non-zero exit code, not as a timeout.
         original_popen = subprocess.Popen
         systemd_timeout_command = "echo 'Failed to start transient scope unit: Connection timed out' >&2 && exit 1"
@@ -526,7 +526,8 @@ class SystemdCgroupsApiTestCase(AgentTestCase):
 
         with tempfile.TemporaryFile(dir=self.tmp_dir, mode="w+b") as stdout:
             with tempfile.TemporaryFile(dir=self.tmp_dir, mode="w+b") as stderr:
-                with patch("azurelinuxagent.common.cgroupapi.subprocess.Popen", side_effect=mock_popen):
+                with patch("azurelinuxagent.common.cgroupapi.subprocess.Popen", side_effect=mock_popen) \
+                        as patch_mock_popen:
                     extension_cgroups, process_output = SystemdCgroupsApi().start_extension_command(
                         extension_name="Microsoft.Compute.TestExtension-1.2.3",
                         command="echo 'success'",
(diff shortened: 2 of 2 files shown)


azurelinuxagent/common/cgroupapi.py
```diff
@@ -510,6 +510,8 @@ def create_cgroup(controller):
                 stderr.truncate(0)
 
                 # Try invoking the process again, this time without systemd-run
+                logger.info('Extension invocation using systemd failed, falling back to regular invocation '
```
larohra: We're already logging systemd failure (L502), why do we need another info log for it?

tests/common/test_cgroupapi.py
```diff
@@ -506,7 +506,42 @@ def mock_popen(*args, **kwargs):
...
+                        command="echo 'success'",
+                        timeout=300,
+                        shell=True,
+                        cwd=self.tmp_dir,
+                        env={},
+                        stdout=stdout,
+                        stderr=stderr)
+
+                    self.assertEquals(extension_cgroups, [])
+                    self.assertEquals(expected_output.format("success"), process_output)
```
narrieta: can we assert that first we tried with systemd, and then without it?

other than that LGTM, thanks!

tests/common/test_cgroupapi.py
```diff
@@ -506,7 +506,42 @@ def mock_popen(*args, **kwargs):
...
+
+    @patch('time.sleep', side_effect=lambda _: mock_sleep(0.001))
+    def test_start_extension_command_should_use_fallback_option_if_systemd_times_out_externally(self, _):
```
narrieta: actually, what is the difference of the new test with this one? (test_start_extension_command_should_use_fallback_option_if_systemd_times_out_externally)

tests/common/test_cgroupapi.py
```diff
@@ -506,7 +506,42 @@ def mock_popen(*args, **kwargs):
                         self.assertEquals(extension_cgroups, [])
 
     @patch('time.sleep', side_effect=lambda _: mock_sleep(0.001))
-    def test_start_extension_command_should_use_fallback_option_if_systemd_times_out(self, _):
+    def test_start_extension_command_should_use_fallback_option_if_systemd_times_out_internally(self, _):
+        # Systemd has its own internal timeout which is not necessarily the same as the timeout we define for
+        # extension operations. When systemd times out, it will write that to stderr and exit with exit code 1.
+        # In that case, we will internally recognize the failure due to the non-zero exit code, not as a timeout.
+        original_popen = subprocess.Popen
+        systemd_timeout_command = "echo 'Failed to start transient scope unit: Connection timed out' >&2 && exit 1"
```
larohra: Also I was going through the systemd fallback code and I noticed something, it doesnt matter if its a systemd timeout or a systemd execution failure (exit non-0), the exception is handled the same way. If that's the case then any particular reason why the tests are named 'systemd_times_out'?

narrieta: implementation vs behavior :)

though the implementation currently handles errors the same (whether they are time outs or not) we want to assert the behavior of important scenarios, timeout being one of them.

if the implementation changes, we want the behavior to remain the same.

larohra: Ohh nice, I get the motivation behind writing the test. 
1 follow up though, if we're checking the behavior then shouldn't we be testing the behavior? Like for example this test has `exit 1` to fail the test which is not necessarily a systemd timeout. If we're validation a scenario then shouldn't that actually mock systemd timeout?

narrieta: Yup, absolutely. When systemd timeout it exits with 1, no stdout and a specific message in stderr. The mock replicates this behavior.

larohra: Ahh very interesting, thanks for sharing! This is an interesting systemd behavior I was unaware of! :)

narrieta: One may argue that the mock should also sleep for the timeout period, but that would not fit with a unit tests, I think (too slow). An integration test may be worth considering.

azurelinuxagent/common/cgroupapi.py
pgombar: To make it more clear from the logs that we're in the fallback logic branch of this if; the logs I was examining yesterday were pretty confusing and I couldn't tell if we were firing it or not without going through the code.

tests/common/test_cgroupapi.py
pgombar: The `_internally` method mimics what happens when systemd times out itself, the `_externally` method mimics what happens when we time it out. I wanted to keep both since we're not in control of the timeout threshold of systemd and it might be larger or smaller than how we define our extension timeout.

tests/common/test_cgroupapi.py
pgombar: I'll add that, thanks.

vrdmr: Yup. Same comment. Checking call_count for both log.info as well as popen check.

tests/common/test_cgroupapi.py
vrdmr: Can you describe `when we time it out.`? If we are timing out, then where are we validating it?

pgombar: After inspecting logs where systemd reported "Connection timed out", it's clear that its internal timeout is significantly shorter than what we define for extension operations (~1 minute versus 5 minutes for extensions), so I'm going to drop the `_externally` test because:
1) it's currently broken (prompts for sudo but then succeeds even without sudo)
2) it doesn't test anything of value (we are not actually causing systemd to timeout/hang before starting the extension operation, but mocking it).

The `_internally` test (which I will rename to just `systemd_times_out`) covers this scenario.

tests/common/test_cgroupapi.py
pgombar: I think checking Popen's call count and arguments are strong enough, checking log.info's call count seems too dependent on the implementation of something that's not crucial to the scenario.

tests/common/test_cgroupapi.py
```diff
@@ -507,38 +507,47 @@ def mock_popen(*args, **kwargs):
...
+                        shell=True,
+                        cwd=self.tmp_dir,
+                        env={},
+                        stdout=stdout,
+                        stderr=stderr)
 
-                        self.assertEquals(extension_cgroups, [])
-                        self.assertEquals(expected_output.format("success"), process_output)
+                    # We expect two calls to Popen, first for the systemd-run call, second for the fallback option
+                    self.assertEquals(2, patch_mock_popen.call_count)
```
narrieta: you can have a stronger check if you assert that the arguments to the first call are wrapped with systemd-run and the arguments for the second call are not.

narrieta: which is done just 2 lines after my comment

oh, my

LGTM

pgombar: Happy friday, Norberto! :D

PR #1686, opened 2019-10-28
Remove upgrade extension loop for the same goal state

pgombar:
## Description

Currently, if we're upgrading an extension and the upgrade fails, we will retry the entire upgrade sequence every three seconds. This is triggered in the main loop of the extension handler thread. 

Unfortunately, when one of the commands in the upgrade sequence keeps failing (e.g. the old extension's disable command) we are stuck in a never-ending loop of trying to upgrade the extension and failing.

This PR removes the retry logic for the upgrade scenario, meaning we will treat it as any other operation -- if the operation was already processed (either successfully or unsuccessfully) for the same goal state (same incarnation number), we will not process it again.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1686)

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -817,7 +817,6 @@ class ExtHandlerInstance(object):
     def set_operation(self, op):
         self.operation = op
 
-
     def report_event(self, message="", is_success=True, duration=0, log_event=True):
         ext_handler_version = self.ext_handler.properties.version
         add_event(name=self.ext_handler.name, version=ext_handler_version, message=message,
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -428,7 +428,7 @@ class ExtHandlersHandler(object):
                 return
 
             self.get_artifact_error_state.reset()
-            if not ext_handler_i.is_upgrade and self.last_etag == etag:
+            if self.last_etag == etag:
                 if self.log_etag:
                     ext_handler_i.logger.verbose("Version {0} is current for etag {1}",
                                                  ext_handler_i.pkg.version,
@@ -684,7 +684,6 @@ class ExtHandlerInstance(object):
         self.operation = None
         self.pkg = None
         self.pkg_file = None
-        self.is_upgrade = False
         self.logger = None
         self.set_logger()
 
@@ -744,12 +743,6 @@ class ExtHandlerInstance(object):
             self.pkg = selected_pkg
             if self.pkg is not None:
                 self.ext_handler.properties.version = str(selected_pkg.version)
-
-        # Note if the selected package is different than that installed
-        if installed_pkg is None \
-                or (
-                self.pkg is not None and FlexibleVersion(self.pkg.version) != FlexibleVersion(installed_pkg.version)):
-            self.is_upgrade = True
 
         if self.pkg is not None:
             self.logger.verbose("Use version: {0}", self.pkg.version)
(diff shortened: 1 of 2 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -684,7 +684,6 @@ def __init__(self, ext_handler, protocol):
         self.operation = None
         self.pkg = None
         self.pkg_file = None
-        self.is_upgrade = False
```
larohra: I'm very curious to know as to why this property was added in the first place?

pgombar: [This PR](https://github.com/Azure/WALinuxAgent/pull/430/files#diff-f8c44add0bc67ec987eff75282ec2308R203) brought it in. It seems the flag was only used to intentionally retry the same goal state if it's an upgrade scenario. I believe the intention was to add a retry in case of transient failures during upgrade, but we've since seen how messy and costly it is when it's a non-transient upgrade failure (hint: OMS), so I don't it's a good trade-off to keep this logic.

larohra: I cant think of a scenario where this might hit, but I'm afraid of customers who might've taken dependency on this "behavior" where we keep trying to retry goalstate if there's an extension upgrade available.
Anyways I feel even if someone did take a dependency then they can modify the behavior to the correct behavior

tests/ga/test_extension.py
```diff
@@ -1481,72 +1481,50 @@ def test__extension_upgrade_failure_when_prev_version_disable_fails(self, patch_
...
-
-            self.assertEqual(loop_run + 1, patch_get_disable_command.call_count)  # counting the earlier run done to
-            # download the new update
-            self.assertEqual(0, patch_download.call_count)  # The download should never be called.
-
-        # On the next iteration, update should not be retried
-        self._assert_handler_status(protocol.report_vm_status, "NotReady", expected_ext_count=0, version="1.0.1")
-
-    @patch('azurelinuxagent.ga.exthandlers.HandlerManifest.get_disable_command')
-    def test__extension_upgrade_failure_when_prev_version_disable_fails_and_recovers(self, patch_get_disable_command,
```
larohra: I dont know how this test was recovering before (maybe by using the is_upgrade flag) - `test__extension_upgrade_failure_when_prev_version_disable_fails_and_recovers`, but maybe you could keep it just to ensure that we have this case tested too.

Everything else LGTM!

pgombar: Good point, this test was "recovering" by simply running the exthandler sequence for the same goal state, which is not valid anymore. However, there is value in having a test that exercises the recover scenario of the upgrade sequence (this time on a new incarnation). I'll add that test. Thanks!

azurelinuxagent/ga/exthandlers.py
vrdmr: @larohra: By _taking dependency on this "behavior"_, do you mean writing extensions assuming that GA would be retrying till the end of time?

@pgombar, we have to make it clear in our documentation as well to make it clear and concrete and we would only try once, if not already in our documentation.

larohra: yeah basically. Like what I meant was since we even had a test case that was pretty much trying to check if the disable failed scenario would recover itself without a new incarnation change, maybe the previous owners told the extension publishers to not worry about transient issues as they would be "auto-resolved" eventually due to which they dont have enough retry logic in their code. 

@vrdmr I agree with your point about double checking the documentation and making it very explicit that we would only try it once per goalstate.

PR #1688, opened 2019-10-31
Disable cgroups when daemon is setup incorrectly

narrieta:
Disable cgroups when, under systemd, the daemon's PID has been added to the file system explicitly.

Also disable when we cannot move the PID to walinuxagent.service (when not under systemd)

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1688)

### DIFF
--- a/azurelinuxagent/common/cgroupapi.py
+++ b/azurelinuxagent/common/cgroupapi.py
@@ -61,7 +61,7 @@ class CGroupsApi(object):
     def start_extension_command(self, extension_name, command, timeout, shell, cwd, env, stdout, stderr, error_code):
         raise NotImplementedError()
 
-    def cleanup_old_cgroups(self):
+    def cleanup_legacy_cgroups(self):
         raise NotImplementedError()
 
     @staticmethod
@@ -100,20 +100,52 @@ class CGroupsApi(object):
         is not mounted or if an error occurs in the operation
         :return: Returns a list of error messages or an empty list if no errors occurred
         """
-        errors = []
         mounted_controllers = os.listdir(CGROUPS_FILE_SYSTEM_ROOT)
 
         for controller in CGROUP_CONTROLLERS:
             try:
                 if controller not in mounted_controllers:
-                    logger.warn('Cgroup controller "{0}" is not mounted. {1}'.format(controller, message))
+                    logger.warn('Cgroup controller "{0}" is not mounted. {1}', controller, message)
                 else:
                     operation(controller)
             except Exception as e:
-                msg = 'Error in cgroup controller "{0}": {1}. {2}'.format(controller, ustr(e), message)
-                logger.warn(msg)
-                errors.append(msg)
-        return errors
+                logger.warn('Error in cgroup controller "{0}": {1}. {2}', controller, ustr(e), message)
+
+    @staticmethod
+    def _foreach_legacy_cgroup(operation):
+        """
+        Previous versions of the daemon (2.2.31-2.2.40) wrote their PID to /sys/fs/cgroup/{cpu,memory}/WALinuxAgent/WALinuxAgent;
+        starting from version 2.2.41 we track the agent service in walinuxagent.service instead of WALinuxAgent/WALinuxAgent. Also,
+        when running under systemd, the PIDs should not be explicitly moved to the cgroup filesystem. The older daemons would
+        incorrectly do that under certain conditions.
+
+        This method checks for the existence of the legacy cgroups and, if the daemon's PID has been added to them, executes the
+        given operation on the cgroups. After this check, the method attempts to remove the legacy cgroups.
+
+        :param operation:
+            The function to execute on each legacy cgroup. It must take 2 arguments: the controller and the daemon's PID
+        """
+        legacy_cgroups = []
+        for controller in ['cpu', 'memory']:
+            cgroup = os.path.join(CGROUPS_FILE_SYSTEM_ROOT, controller, "WALinuxAgent", "WALinuxAgent")
+            if os.path.exists(cgroup):
+                logger.info('Found legacy cgroup {0}', cgroup)
+                legacy_cgroups.append((controller, cgroup))
+
+        try:
+            for controller, cgroup in legacy_cgroups:
+                procs_file = os.path.join(cgroup, "cgroup.procs")
+
+                if os.path.exists(procs_file):
+                    procs_file_contents = fileutil.read_file(procs_file).strip()
+                    daemon_pid = fileutil.read_file(get_agent_pid_file_path()).strip()
+
+                    if daemon_pid in procs_file_contents:
+                        operation(controller, daemon_pid)
+        finally:
+            for _, cgroup in legacy_cgroups:
+                logger.info('Removing {0}', cgroup)
+                shutil.rmtree(cgroup, ignore_errors=True)
 
 
 class FileSystemCgroupsApi(CGroupsApi):
@@ -144,6 +176,10 @@ class FileSystemCgroupsApi(CGroupsApi):
                     raise
 
     @staticmethod
+    def _get_agent_cgroup_path(controller):
+        return os.path.join(CGROUPS_FILE_SYSTEM_ROOT, controller, VM_AGENT_CGROUP_NAME)
+
+    @staticmethod
     def _get_extension_cgroups_root_path(controller):
         return os.path.join(CGROUPS_FILE_SYSTEM_ROOT, controller, EXTENSIONS_ROOT_CGROUP_NAME)
 
@@ -158,8 +194,7 @@ class FileSystemCgroupsApi(CGroupsApi):
         return os.path.join(extensions_root, cgroup_name)
 
     def _create_extension_cgroup(self, controller, extension_name):
-        return CGroup.create(self._get_extension_cgroup_path(controller, extension_name),
-                             controller, extension_name)
+        return CGroup.create(self._get_extension_cgroup_path(controller, extension_name), controller, extension_name)
 
     @staticmethod
     def _add_process_to_cgroup(pid, cgroup_path):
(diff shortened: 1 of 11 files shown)


azurelinuxagent/common/cgroupapi.py
```diff
@@ -100,20 +100,52 @@ def _foreach_controller(operation, message):
...
                 if controller not in mounted_controllers:
-                    logger.warn('Cgroup controller "{0}" is not mounted. {1}'.format(controller, message))
+                    logger.warn('Cgroup controller "{0}" is not mounted. {1}', controller, message)
                 else:
                     operation(controller)
             except Exception as e:
-                msg = 'Error in cgroup controller "{0}": {1}. {2}'.format(controller, ustr(e), message)
-                logger.warn(msg)
-                errors.append(msg)
-        return errors
```
narrieta: now there is no need for this function to report errors to the caller

tests/common/test_cgroupapi.py
```diff
@@ -18,30 +18,37 @@
...
 from azurelinuxagent.common.utils import shellutil
 from nose.plugins.attrib import attr
+from tests.utils.cgroups_tools import CGroupsTools
 from tests.tools import *
 
 
-class CGroupsApiTestCase(AgentTestCase):
-    def setUp(self):
+class _MockedFileSystemTestCase(AgentTestCase):
+    def setUp(self, mock_cgroups_file_system_root=True):
```
narrieta: 'mock_cgroups_file_system_root' is a left over from a previous iteration... will remove it now

tests/common/test_cgroupapi.py
```diff
@@ -18,30 +18,37 @@
...
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.utils import shellutil
 from nose.plugins.attrib import attr
+from tests.utils.cgroups_tools import CGroupsTools
 from tests.tools import *
 
 
-class CGroupsApiTestCase(AgentTestCase):
-    def setUp(self):
+class _MockedFileSystemTestCase(AgentTestCase):
```
narrieta: refactored common mocks to this class

tests/common/test_cgroupapi.py
```diff
@@ -332,28 +286,17 @@ def test_start_extension_command_should_add_the_child_process_to_the_extension_c
...
-    def tearDown(self):
-        AgentTestCase.tearDown(self)
-
-    def test_it_should_return_extensions_slice_root_name(self):
+    def test_get_extensions_slice_root_name_should_return_the_root_slice_for_extensions(self):
         root_slice_name = SystemdCgroupsApi()._get_extensions_slice_root_name()
         self.assertEqual(root_slice_name, "system-walinuxagent.extensions.slice")
 
-    def test_it_should_return_extension_slice_name(self):
+    def test_get_extension_slice_name_should_return_the_slice_for_the_given_extension(self):
```
narrieta: replaced 'it' with the actual test subject in these method names

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -79,7 +79,7 @@ def enable(self):
         def disable(self):
             self._enabled = False
 
-        def _invoke_cgroup_operation(self, operation, error_message):
+        def _invoke_cgroup_operation(self, operation, error_message, on_error=None):
             """
             Ensures the given operation is invoked only if cgroups are enabled and traps any errors on the operation.
             """
```
pgombar: Can you update the comment explaining what `on_error` does and why it's there?

azurelinuxagent/daemon/main.py
```diff
@@ -69,7 +69,6 @@ def run(self, child_args=None):
         self.initialize_environment()
 
         CGroupConfigurator.get_instance().create_agent_cgroups(track_cgroups=False)
-        CGroupConfigurator.get_instance().cleanup_old_cgroups()
```
pgombar: Why not have the daemon clean up?

tests/common/test_cgroupapi.py
```diff
@@ -18,16 +18,16 @@
...
-
-from azurelinuxagent.common.cgroupapi import CGroupsApi, FileSystemCgroupsApi, SystemdCgroupsApi, VM_AGENT_CGROUP_NAME
-from azurelinuxagent.common.exception import ExtensionError, ExtensionErrorCodes
+from azurelinuxagent.common.cgroupapi import CGroupsApi, FileSystemCgroupsApi, SystemdCgroupsApi, CGROUPS_FILE_SYSTEM_ROOT, VM_AGENT_CGROUP_NAME
+from azurelinuxagent.common.exception import CGroupsException, ExtensionError, ExtensionErrorCodes
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.utils import shellutil
 from nose.plugins.attrib import attr
+from tests.utils.cgroups_tools import CGroupsTools
 from tests.tools import *
```
pgombar: Refactor `import *`?

tests/common/test_cgroupapi.py
```diff
@@ -119,59 +122,31 @@ def controller_operation(controller):
...
-
-    def setUp(self):
-        AgentTestCase.setUp(self)
-
-        self.cgroups_file_system_root = os.path.join(self.tmp_dir, "cgroup")
-        os.mkdir(self.cgroups_file_system_root)
-        os.mkdir(os.path.join(self.cgroups_file_system_root, "cpu"))
-        os.mkdir(os.path.join(self.cgroups_file_system_root, "memory"))
+            (message_format, controller, error, message) = args
+            self.assertEquals(message_format, 'Error in cgroup controller "{0}": {1}. {2}')
```
pgombar: So this will abstract away whatever's in `{}`? Cool, did not know this.

tests/common/test_cgroupconfigurator.py
```diff
@@ -19,10 +19,13 @@
...
 import subprocess
 
+import errno
+from azurelinuxagent.common.cgroupapi import VM_AGENT_CGROUP_NAME
 from azurelinuxagent.common.cgroupconfigurator import CGroupConfigurator
 from azurelinuxagent.common.cgroupstelemetry import CGroupsTelemetry
 from azurelinuxagent.common.exception import CGroupsException
 from azurelinuxagent.common.osutil.default import DefaultOSUtil
+from tests.utils.cgroups_tools import CGroupsTools
 from tests.tools import *
```
pgombar: Another `import *`

azurelinuxagent/common/cgroupapi.py
```diff
@@ -158,49 +194,28 @@ def _get_extension_cgroup_path(self, controller, extension_name):
...
+        Previous versions of the daemon (2.2.31-2.2.40) wrote their PID to /sys/fs/cgroup/{cpu,memory}/WALinuxAgent/WALinuxAgent;
+        starting from version 2.2.41 we track the agent service in walinuxagent.service instead of WALinuxAgent/WALinuxAgent. This
+        method moves the daemon's PID from the legacy cgroups to the newer cgroups.
+        """
+        def move_daemon_pid(controller, daemon_pid):
+            new_path = FileSystemCgroupsApi._get_agent_cgroup_path(controller)
+            logger.info("Writing daemon's PID ({0}) to {1}", daemon_pid, new_path)
+            fileutil.append_file(os.path.join(new_path, "cgroup.procs"), daemon_pid)
+            msg = "Moved daemon's PID from legacy cgroup to {0}".format(new_path)
+            add_event(AGENT_NAME, version=CURRENT_VERSION, op=WALAEventOperation.CGroupsCleanUp, is_success=True, message=msg)
```
pgombar: Can we explicitly say that the collection of resource usage data is still enabled? I'm looking at it from the perspective of querying for this event and parsing what the different event messages mean.

azurelinuxagent/common/exception.py
```diff
@@ -55,7 +55,11 @@ class CGroupsException(AgentError):
 
     def __init__(self, msg, inner=None):
         super(AgentError, self).__init__(msg, inner)
+        # TODO: AgentError should set the message - investigate whether doing it there would break anything
```
pgombar: So... would it? :)

narrieta:
@pgombar 
> should we explicitly check if the daemon's (and extension handler's) PID is in the correct place, and, if not, also disable cgroups?

We could add more checks on a later release

azurelinuxagent/common/cgroupapi.py
```diff
@@ -100,20 +100,52 @@ def _foreach_controller(operation, message):
...
+
+                if os.path.exists(procs_file):
+                    procs_file_contents = fileutil.read_file(procs_file).strip()
+                    daemon_pid = fileutil.read_file(get_agent_pid_file_path()).strip()
+
+                    if daemon_pid in procs_file_contents:
+                        operation(controller, daemon_pid)
+        finally:
+            for _, cgroup in legacy_cgroups:
+                logger.info('Removing {0}', cgroup)
```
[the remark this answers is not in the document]
narrieta: the whole path is in the message. a few lines up in the log there would be a message identifying the path as a legacy group (logged at line 132)

azurelinuxagent/common/cgroupapi.py
narrieta: we can talk offline... not sure I understand how that would improve parsing

azurelinuxagent/common/cgroupconfigurator.py
narrieta: doesn't the name imply a callback to execute when an error occurs? :)

azurelinuxagent/common/exception.py
narrieta: Dunno; no time to check :)

Doing it  in AgentError would change the formatting of all exceptions. Need to spend some time checking what would happen then.

azurelinuxagent/daemon/main.py
narrieta: It is not needed. Or did you have an scenario in mind when it was added?

tests/common/test_cgroupapi.py
narrieta: Looked into it when I started modifying this test. There is some funny stuff there that is going to require more time than I have the current release.

Let's see who in the team gets there first.

tests/common/test_cgroupapi.py
narrieta: This was actually needed because I changed the statement that logs this message. Previously it was logging a string formatted with string.format, now it lets the logger do the formatting so the first argument is actually the formatting string.

tests/common/test_cgroupconfigurator.py
narrieta: i know :)

azurelinuxagent/common/cgroupapi.py
vrdmr: Can you also log this locally? This could only send event to us.

azurelinuxagent/ga/update.py
```diff
@@ -457,7 +457,7 @@ def _ensure_readonly_files(self):
     def _ensure_cgroups_initialized(self):
         configurator = CGroupConfigurator.get_instance()
         configurator.create_agent_cgroups(track_cgroups=True)
-        configurator.cleanup_old_cgroups()
+        configurator.cleanup_legacy_cgroups()
```
vrdmr: Does the order matter? Creating the new cgroup and then deleting the legacy cgroup? or should we be clearing legacy cgroup first before creating the new agent cgroup.

azurelinuxagent/common/cgroupapi.py
narrieta: logged at line 213

azurelinuxagent/ga/update.py
narrieta: Yes, cleanup may need to move the PID to the new cgroup, which needs to be created first (by create_agent_cgroups)

PR #1693, opened 2019-11-03
Updating gitignore to ignore virtual environment files.

vrdmr:
## Description
Updating gitignore to ignore venv files and other extra files.

Issue # 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1693)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,12 +79,6 @@ bin/waagent2.0c
 # Virtualenv
 # http://iamzed.com/2009/05/07/a-primer-on-virtualenv/
 .Python
-[Bb]in
-[Ii]nclude
-[Ll]ib
-[Ll]ib64
-[Ll]ocal
-[Ss]cripts
 pyvenv.cfg
 .venv
 pip-selfcheck.json
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -72,3 +74,27 @@ bin/waagent2.0c
 
 # mac osx specific files
 .DS_Store
+
```
narrieta: wouldn't it be better to move your virtual environment out of the agent's folder?

in case we ever need to add any of those paths to the agent

narrieta: wouldn't it be better to move your virtual environment out of the agent's folder?

in case we ever need to add any of those paths to the agent

pgombar: It's pretty common to set up a virtual environment in the folder your project is in, I think. I was hitting this same issue but I gave up on doing `git add .` and instead try to add everything except venv.

narrieta: I setup my virtual environments outside source control and share them between projects.

Maybe you guys can set yours within a subdirectory so that we do not need to filter out bin/lib/include/scripts (which are very common and we may use in the future) at the root level?

vrdmr: @narrieta: I agree that setting up VirtualEnv outside is the ideal way. But pycharm sets it up in the folder. I did fix it for my dev-env.

But, this PR was to keep the `.gitignore` in such a way that it helps any other developers to not send `venv` by mistake.

PR #1695, opened 2019-11-04
Be more exact when detecting IMDS service health

jacwil:
## Description

This change removes the fault from IMDS when connection errors and throttling occurs.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1695)

### DIFF
--- a/azurelinuxagent/common/protocol/imds.py
+++ b/azurelinuxagent/common/protocol/imds.py
@@ -311,8 +311,6 @@ class ImdsClient(object):
 
         status, resp = self._get_metadata_from_endpoint(endpoint, resource_path, headers)
         if status == IMDS_CONNECTION_ERROR:
-            logger.periodic_warn(logger.EVERY_FIFTEEN_MINUTES,
-                                 "[PERIODIC] [IMDS_CONNECTION_ERROR] Unable to connect to primary IMDS endpoint {0}".format(endpoint))
             endpoint = self._protocol_util.get_wireserver_endpoint()
             status, resp = self._get_metadata_from_endpoint(endpoint, resource_path, headers)
 
@@ -320,9 +318,6 @@ class ImdsClient(object):
             return MetadataResult(True, False, resp)
         elif status == IMDS_INTERNAL_SERVER_ERROR:
             return MetadataResult(False, True, resp)
-        elif status == IMDS_CONNECTION_ERROR:
-            logger.periodic_warn(logger.EVERY_FIFTEEN_MINUTES,
-                                 "[PERIODIC] [IMDS_CONNECTION_ERROR] Unable to connect to backup IMDS endpoint {0}".format(endpoint))
         return MetadataResult(False, False, resp)
 
     def get_compute(self):
--- a/azurelinuxagent/common/protocol/imds.py
+++ b/azurelinuxagent/common/protocol/imds.py
@@ -2,6 +2,7 @@
 # Licensed under the Apache License, Version 2.0 (the "License");
 import json
 import re
+from collections import namedtuple
 
 import azurelinuxagent.common.utils.restutil as restutil
 from azurelinuxagent.common.exception import HttpError, ResourceGoneError
@@ -20,6 +21,7 @@ IMDS_IMAGE_ORIGIN_CUSTOM = 1
 IMDS_IMAGE_ORIGIN_ENDORSED = 2
 IMDS_IMAGE_ORIGIN_PLATFORM = 3
 
+MetadataResult = namedtuple('MetadataResult', ['success', 'service_error', 'response'])
 IMDS_RESPONSE_SUCCESS = 0
 IMDS_RESPONSE_ERROR = 1
 IMDS_CONNECTION_ERROR = 2
@@ -281,7 +283,7 @@ class ImdsClient(object):
                 return IMDS_RESPONSE_ERROR, "IMDS error in /metadata/{0}: Throttled".format(resource_path)
             if self._regex_ioerror.match(msg):
                 logger.periodic_warn(logger.EVERY_FIFTEEN_MINUTES,
-                                     "[PERIODIC] Unable to connect to IMDS endpoint {0}".format(endpoint))
+                                     "[PERIODIC] [IMDS_CONNECTION_ERROR] Unable to connect to IMDS endpoint {0}".format(endpoint))
                 return IMDS_CONNECTION_ERROR, "IMDS error in /metadata/{0}: Unable to connect to endpoint".format(resource_path)
             return IMDS_INTERNAL_SERVER_ERROR, "IMDS error in /metadata/{0}: {1}".format(resource_path, msg)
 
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/protocol/imds.py
```diff
@@ -254,38 +259,71 @@ def _http_get(self, endpoint, resource_path, headers):
...
         headers = self._health_headers if is_health else self._headers
         endpoint = IMDS_ENDPOINT
-        try:
-            resp = self._http_get(endpoint=endpoint, resource_path=resource_path, headers=headers)
-        except HttpError as e:
+
+        status, resp = self._get_metadata_from_endpoint(endpoint, resource_path, headers)
+        if status == IMDS_CONNECTION_ERROR:
             logger.periodic_warn(logger.EVERY_FIFTEEN_MINUTES,
                                  "[PERIODIC] Unable to connect to primary IMDS endpoint {0}".format(endpoint))
```
vrdmr: You might also add {IMDS_CONNECTION_ERROR} in the log message to signify the status (or some other way to describe the cause).

azurelinuxagent/common/protocol/imds.py
```diff
@@ -254,38 +259,71 @@ def _http_get(self, endpoint, resource_path, headers):
...
-                    raise
-                return False, "IMDS error in /metadata/{0}: Unable to connect to endpoint".format(resource_path)
-        if restutil.request_failed(resp):
-            return False, "IMDS error in /metadata/{0}: {1}".format(
-                resource_path, restutil.read_response_error(resp))
-        return True, resp.read()
+            status, resp = self._get_metadata_from_endpoint(endpoint, resource_path, headers)
+
+        if status == IMDS_RESPONSE_SUCCESS:
+            return True, False, resp
```
vrdmr: You could use NamedTuple to simplify the logic here.

azurelinuxagent/common/protocol/imds.py
```diff
@@ -296,9 +334,9 @@ def get_compute(self):
         """
 
         # ensure we get a 200
-        success, resp = self.get_metadata('instance/compute', is_health=False)
+        success, service_error, resp = self.get_metadata('instance/compute', is_health=False)
         if not success:
-            raise HttpError(resp)
+            raise ValueError(resp)
```
vrdmr: Why is it a value error here?

tests/protocol/test_imds.py
```diff
@@ -346,6 +359,9 @@ def _assert_validation(self, http_status_code, http_response, expected_valid, ex
                                                       response=http_response)
             validate_response = test_subject.validate()
 
+        #if http_response == 'error response':
+        #    raise Exception('Actual: {0}, Expected: {1}'.format(validate_response[1], expected_response))
+
```
vrdmr: Stale comments?

azurelinuxagent/common/protocol/imds.py
jacwil: I think it was a naïve way to simplify logic. I'll revert.

vrdmr: If you feel this needs a special error, then I would recommend creating an IMDSError class, which you can use to simplify and reuse later.

azurelinuxagent/common/protocol/imds.py
jacwil: Great idea!

azurelinuxagent/common/protocol/imds.py
```diff
@@ -254,38 +261,69 @@ def _http_get(self, endpoint, resource_path, headers):
...
+            status: one of the following response status codes: IMDS_RESPONSE_SUCCESS, IMDS_RESPONSE_ERROR,
+                    IMDS_CONNECTION_ERROR, IMDS_INTERNAL_SERVER_ERROR
+            response: IMDS response on IMDS_RESPONSE_SUCCESS, failure message otherwise
+        """
+        try:
+            resp = self._http_get(endpoint=endpoint, resource_path=resource_path, headers=headers)
+        except ResourceGoneError:
+            return IMDS_INTERNAL_SERVER_ERROR, "IMDS error in /metadata/{0}: HTTP Failed with Status Code 410: Gone".format(resource_path)
+        except HttpError as e:
+            msg = str(e)
```
pgombar: `ustr` instead of `str`?

azurelinuxagent/common/protocol/imds.py
```diff
@@ -254,38 +261,69 @@ def _http_get(self, endpoint, resource_path, headers):
...
-            resp = self._http_get(endpoint=endpoint, resource_path=resource_path, headers=headers)
-        except HttpError as e:
+
+        status, resp = self._get_metadata_from_endpoint(endpoint, resource_path, headers)
+        if status == IMDS_CONNECTION_ERROR:
             logger.periodic_warn(logger.EVERY_FIFTEEN_MINUTES,
-                                 "[PERIODIC] Unable to connect to primary IMDS endpoint {0}".format(endpoint))
-            if not self._regex_imds_ioerror.match(str(e)):
-                raise
+                                 "[PERIODIC] [IMDS_CONNECTION_ERROR] Unable to connect to primary IMDS endpoint {0}".format(endpoint))
```
pgombar: Wouldn't this message be logged already in L285?

jacwil: Yes and no. This message adds "primary" or "secondary" to the log message, but I agree that this is not actually important since we already log the endpoint. I'll remove these as duplicate.

azurelinuxagent/common/protocol/imds.py
jacwil: I'm not sure if that change would break the regex matches that follow.

vrdmr: Unless you are not logging the exact message, `str` should be good rather than `ustr`. Just do a quick check to see is the message is being translated into `ustr` before logging.

azurelinuxagent/common/protocol/imds.py
```diff
@@ -254,38 +261,64 @@ def _http_get(self, endpoint, resource_path, headers):
...
+            if self._regex_throttled.match(msg):
+                return IMDS_RESPONSE_ERROR, "IMDS error in /metadata/{0}: Throttled".format(resource_path)
+            if self._regex_ioerror.match(msg):
+                logger.periodic_warn(logger.EVERY_FIFTEEN_MINUTES,
+                                     "[PERIODIC] [IMDS_CONNECTION_ERROR] Unable to connect to IMDS endpoint {0}".format(endpoint))
+                return IMDS_CONNECTION_ERROR, "IMDS error in /metadata/{0}: Unable to connect to endpoint".format(resource_path)
+            return IMDS_INTERNAL_SERVER_ERROR, "IMDS error in /metadata/{0}: {1}".format(resource_path, msg)
+
+        if resp.status >= 500:
+            return IMDS_INTERNAL_SERVER_ERROR, "IMDS error in /metadata/{0}: {1}".format(
```
vrdmr: Not now, but you should think of creating a new class `IMDSError` which would encapsulate this better and you could have all the logic in it.

PR #1696, opened 2019-11-05
Add container id for extension telemetry events

pgombar:
## Description

PR #1628 introduced the logic of populating `ContainerId` values for telemetry event parameters on the fly, as the events were getting created, instead of once per service startup in `init_sysinfo`. 

The same telemetry pipeline is used by extension events too. Those events are dropped by extensions in the events folder, where the agent picks them up, parses them, adds sysinfo parameters and emits them. However, by having moved the `ContainerId` generation logic away from `init_sysinfo`, which was called for any event generated, now the `ContainerId` parameter is never generated for extension events.

This PR adds the `ContainerId` value from the designated environment variable, if the parameter is not already present, when adding the rest of sysinfo parameters and just before reporting the event, ensuring all emitted events have a `ContainerId` set.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1696)

### DIFF
--- a/azurelinuxagent/ga/monitor.py
+++ b/azurelinuxagent/ga/monitor.py
@@ -230,7 +230,7 @@ class MonitorHandler(object):
         Periodically read, parse, and send events located in the events folder. Currently, this is done every minute.
         Any .tld file dropped in the events folder will be emitted. These event files can be created either by the
         agent or the extensions. We don't have control over extension's events parameters, but we will override
-        any values they might have set for sys_info parameters:
+        any values they might have set for sys_info parameters.
         """
         if self.last_event_collection is None:
             self.last_event_collection = datetime.datetime.utcnow() - MonitorHandler.EVENT_COLLECTION_PERIOD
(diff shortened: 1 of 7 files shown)


azurelinuxagent/ga/monitor.py
```diff
@@ -300,15 +306,35 @@ def reset_loggers(self):
                 self.last_reset_loggers_time = time_now
 
     def add_sysinfo(self, event):
+        """
+        This method is called after parsing the event file in the events folder and before emitting it. This means
+        all events, either coming from the agent or from the extensions, are passed through this method. The purpose
+        is to add a static list of sys_info parameters such as VMName, Region, RAM, etc. If the sys_info parameters
+        are already populated in the event, they are not overwritten.
```
larohra: Shouldn't it be - `they ARE overwritten`?

pgombar: Thanks, it is the opposite of what I wrote. I'll update in the next iteration.

PR #1704, opened 2019-11-07
Remove 'from tests.tools import *'

pgombar:
## Description

The `tests.tools` module contains many helper functions used throughout other test modules. A lot of these test modules just used `from tests.tools import *`, which results in 1) importing unnecessary things, and 2) importing other agent modules implicitly via `tests.tools` instead of explicitly from the current test module.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1704)

### DIFF
--- a/tests/common/osutil/test_ubuntu.py
+++ b/tests/common/osutil/test_ubuntu.py
@@ -14,9 +14,12 @@
 #
 # Requires Python 2.6+ and Openssl 1.0+
 #
+
+import unittest
+
 from azurelinuxagent.common.osutil.ubuntu import Ubuntu12OSUtil, Ubuntu18OSUtil
 from .test_default import osutil_get_dhcp_pid_should_return_a_list_of_pids
-from tests.tools import *
+from tests.tools import AgentTestCase
 
 
 class TestUbuntu12OSUtil(AgentTestCase):
--- a/tests/ga/test_extension.py
+++ b/tests/ga/test_extension.py
@@ -22,10 +22,7 @@ from datetime import timedelta
 
 from azurelinuxagent.ga.monitor import get_monitor_handler
 from nose.plugins.attrib import attr
-from tests.protocol.mockwiredata import DATA_FILE, DATA_FILE_EXT_AUTOUPGRADE, \
-    DATA_FILE_EXT_AUTOUPGRADE_INTERNALVERSION, DATA_FILE_EXT_DELETION, DATA_FILE_EXT_INTERNALVERSION, \
-    DATA_FILE_EXT_ROLLINGUPGRADE, DATA_FILE_EXT_SEQUENCING, DATA_FILE_EXT_SINGLE, DATA_FILE_NO_EXT, \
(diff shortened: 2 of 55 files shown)


tests/ga/test_extension.py
```diff
@@ -16,13 +16,20 @@
...
 from datetime import timedelta
 
 from azurelinuxagent.ga.monitor import get_monitor_handler
 from nose.plugins.attrib import attr
-from tests.protocol.mockwiredata import *
-
+from tests.protocol.mockwiredata import DATA_FILE, DATA_FILE_EXT_AUTOUPGRADE, \
+    DATA_FILE_EXT_AUTOUPGRADE_INTERNALVERSION, DATA_FILE_EXT_DELETION, DATA_FILE_EXT_INTERNALVERSION, \
+    DATA_FILE_EXT_ROLLINGUPGRADE, DATA_FILE_EXT_SEQUENCING, DATA_FILE_EXT_SINGLE, DATA_FILE_NO_EXT, \
+    DATA_FILE_EXT_NO_PUBLIC, DATA_FILE_EXT_NO_SETTINGS, DATA_FILE_MULTIPLE_EXT, WireProtocolData
```
narrieta: since this list is likely to keep growing, maybe instead do
```
import tests.protocol.mockwiredata 
```
and then qualify each reference? e.g
```
mockwiredata.DATA_FILE
```

tests/ga/test_extension.py
```diff
@@ -16,13 +16,20 @@
...
+from tests.tools import are_cgroups_enabled, AgentTestCase, data_dir, i_am_root, MagicMock, Mock, \
```
narrieta: similarly, instead of this long list we move those functions (are_cgroups_enabled, i_am_root, etc) to a single class and import that class

data_dir should probably be moved to AgentTestCase

but we can address those on a separate PR

tests/ga/test_extension.py
pgombar: Yeah, it was getting slightly out of hand. I'll update this on this PR.

tests/ga/test_extension.py
pgombar: Okay, leaving this comment for a separate PR.

tests/common/osutil/test_ubuntu.py
```diff
@@ -14,9 +14,12 @@
 #
 # Requires Python 2.6+ and Openssl 1.0+
 #
+
+import unittest
```
vrdmr: Is this from any legacy code based on `unittest`? 

Also, as unittest is used in many places, and gets imported from `from tests.protocol.mockwiredata import *`, do you feel we should import `unittest` in `test.tools` itself?

pgombar: A lot of our test modules have this at the end of the file:
```
if __name__ == '__main__':
    unittest.main()
```
I didn't want to change this in this PR, since I'm not sure why it's even there. I'm guessing to allow running the tests even if the `nose` requirement is not satisfied.

I removed `from tests.protocol.mockwiredata import *` throughout specifically to prevent this type of shadow importing, so `unittest` needs to be explicitly imported. We can take the refactoring of it to `test.tools` along with the other modules Norberto mentioned in the next PR, I agree with the suggestion.

PR #1706, opened 2019-11-08
Fix issue with GPT disk mounting with guessing filesystem type

trstringer:
## Description

When we activate the resource disk, if the resource disk is a GPT disk with a single partition then we will try to mount the disk without formatting it. Because the resource disks are originally ntfs filesystem, if the machine has the ntfs driver then it will mount with that filesystem, completely disregarding the configuration `ResourceDisk.Filesystem`.

The fix is to explicitly specify the filesystem type with `-t` and passing in the configuration value so that we ensure that the user gets the resource disk in the desired filesystem. This should not be an ntfs-first approach.

// @anhvoms 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1706)

### DIFF
--- a/azurelinuxagent/daemon/resourcedisk/default.py
+++ b/azurelinuxagent/daemon/resourcedisk/default.py
@@ -249,14 +249,20 @@ class ResourceDiskHandler(object):
 
         return err_code, output
 
-    @staticmethod
-    def get_mount_string(mount_options, partition, mount_point):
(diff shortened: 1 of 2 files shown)


trstringer:
@narrieta thanks for the patience on this PR! Please move forward with merging.

narrieta:
running automation

PR #1714, opened 2019-11-20
Add options for log dir and log rotation

remip2:
## Description

The path for the log file waagent.log is hardcoded to /var/log and the file size can grow very fast when they are issues with extension install. NSBSD OS has a (small) RAM drive for /var and some Azure VM failed with the RAM disk full. NSBSD has a dedicated log partition mounted in /log which should be used.

The proposed new options permit to configure a custom path for waagent.log and the rotation of the file when it exceed a fixed size (with a fixed number of files to keep). 

The defaults are no file rotation and /var/log path, only the nsbsd configuration has been updated to use the new options.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1714)


vrdmr:
@remip2 Thanks for the contribution.

I recently made some changes to logger.py. Could you please rebase against develop (or merge), to get the changes in? With the new `Appender` class added, you might want to add the new fiends into it as well. 

Also, with the difference in agents (daemon, which comes in the box, and ext-handler, which gets auto-updated), there could be a situation that /var/log/waagent.log is still getting populated (by old daemon), and the new log file, at some other location, is being written by the updated ext-handler.

We'll have to handle those cases. And others that I can think of (some tools to fetch the logs form the machines depend on that path).

azurelinuxagent/common/logger.py
```diff
@@ -125,18 +126,47 @@ def write(self, level, msg):
...
+
+    def doRollover(self):
+        if self.backup_count > 0:
+            for i in range(self.backup_count-1, 0, -1):
+                src = "%s.%d" % (self.path, i)
+                dst = "%s.%d" % (self.path, i + 1)
+                if os.path.exists(src):
+                    if os.path.exists(dst):
+                        os.remove(dst)
+                    os.rename(src, dst)
```
vrdmr: Why not compress the backup?

remip2: For my use case I don't need compression, old log should be readable in plain text.

remip2:
I don't think migration is needed since I didn't change any default value. Existing agents will continue to log at the same place without rotation (or with the OS rotation). For NSBSD OS migration is handled on our side (and autoupdate is disabled).

vrdmr:
>I don't think migration is needed since I didn't change any default value. Existing agents will continue to log at the same place without rotation (or with the OS rotation).

Agreed. One possible issue is that adding new a new property for log exposes it to be changed by users without us knowing the location of the logs for collecting them.

Most of the existing agents rely on logrorate log files. And I don't want to add custom rotating logic unless specific to an OS, where it can be handled in OS-specific utils.

> For NSBSD OS migration is handled on our side (and autoupdate is disabled).

Is NSBSD OS a different flavor of BSD? And does it require its own OSUtil? We can solve the location of the Logs file issue with adding a new field to get the log file location from the OSUtil.

remip2:
> Is NSBSD OS a different flavor of BSD? And does it require its own OSUtil? We can solve the location of the Logs file issue with adding a new field to get the log file location from the OSUtil.

yes, NSBSD already have its own osutil class, I'll try to do the change

remip2:
> You should also update the location in the `Extension.LogDir` config in the config/nsbsd/waagent.conf as well.

yes, It's already have been done when I first integrated nsbsd (for Lib.Dir too)

azurelinuxagent/common/logger.py
```diff
@@ -173,18 +174,47 @@ def write(self, level, msg):
...
                 pass
 
+    def shouldRollover(self, msg):
+        if self.backup_count == 0:
+            return False
+
+        if os.path.exists(self.path) and os.stat(self.path).st_size + len(msg) > self.max_bytes:
+            return True
+
+        return False
```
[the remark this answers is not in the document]
remip2: > One qq - Does NSBSD have logrotate? I do see [logrotate](https://www.freebsd.org/cgi/man.cgi?query=logrotate) available on FreeBSD. If yes then why do we need the `shouldRollover` and `doRollover` logic?

No NSBSD doesn't have logrotate.

remip2:
@vrdmr 

> Can I add some changes to this PR to move this code to the OSUtil side as well?

Can you give me some hints about this idea, I could try to do the code ?

azurelinuxagent/agent.py
```diff
@@ -62,11 +65,25 @@ def __init__(self, verbose, conf_file_path=None):
...
+        try:
+            if os.path.isfile(log_dir):
+                raise Exception("{0} is a file".format(log_dir))
+            if not os.path.isdir(log_dir):
+                fileutil.mkdir(log_dir, mode=0o755, owner="root")
+        except Exception as e:
+            logger.error(
+                "Exception occurred while creating "
+                "log directory {0}: {1}".format(log_dir, e))
+        log_file = os.path.join(log_dir, WAAGENT_LOG_FILE)
```
larohra: In the scenario that `log_dir` is actually a file (i.e. `os.path.isfile(log_dir) == true`), the `os.path.join` would just append the log filename to the log_dir file. That inturn would throw an error later when trying to write to file. 

Eg: 
```python
with open('/home/larohra/test/1.log', 'w') as fh:
...     fh.write("test123")
...     
  File "<input>", line 1, in <module>
IOError: [Errno 20] Not a directory: '/home/larohra/test.txt/1.log'
```

We should handle this scenario too. If incase log_dir is a file, maybe just default to our `/var/log` directory?

azurelinuxagent/common/osutil/default.py
```diff
@@ -108,6 +109,12 @@ def __init__(self):
         self.jit_enabled = False
         self.service_name = self.get_service_name()
 
+        # Log-related OS configurations.
+        # self.agent_log_dir = self.get_agent_log_dir()
```
larohra: NIT: Forgot to remove the comments?

azurelinuxagent/pa/deprovision/default.py
```diff
@@ -92,7 +92,8 @@ def del_dirs(self, warnings, actions):
         actions.append(DeprovisionAction(fileutil.rm_dirs, dirs))
 
     def del_files(self, warnings, actions):
-        files = ['/root/.bash_history', '/var/log/waagent.log']
+        log_file = os.path.join(self.osutil.get_agent_log_dir(), "waagent.log*")
```
larohra: same concern, if `get_agent_log_dir()` is a file it might throw errors later

azurelinuxagent/agent.py
vrdmr: But this is under the assumption that `log_dir` is a file, and someone did not understand the word `dir`. It is supposed to be a directory :)

I didn't understand your concern here - you fear that the OSUtil will have file, in place of a directory?

larohra: We have a check above - 
```python
if os.path.isfile(log_dir):
                raise Exception("{0} is a file".format(log_dir)
```

And this exception is just being eaten and logging an error and just moving on. If this scenario is actually true, then the log_appender initialization would throw an error which (haven't checked) could bring down the agent. That is my main concern. Its a far fetched case but I feel we sjpi;d handle it.

larohra: >I didn't understand your concern here - you fear that the OSUtil will have file, in place of a directory?

Yeah basically

azurelinuxagent/pa/deprovision/default.py
```diff
@@ -92,7 +93,14 @@ def del_dirs(self, warnings, actions):
         actions.append(DeprovisionAction(fileutil.rm_dirs, dirs))
 
     def del_files(self, warnings, actions):
-        files = ['/root/.bash_history', '/var/log/waagent.log']
+        # The passed log-dir was incorrect and we defaulted to DEFAULT_LOG_DIR during the setup of log. Check
```
larohra: NIT: 
If the passed log-dir was incorrect, we default to DEFAULT_LOG_DIR.

azurelinuxagent/common/osutil/default.py
```diff
@@ -1420,3 +1424,37 @@ def _update_nic_state(self, state, ip_command, handler, description):
                     handler(state[interface_name], result.group(2))
                 else:
                     logger.error("Interface {0} has {1} but no link state".format(interface_name, description))
+
+    def supports_logrotate(self):
+        """
+        logrotate is allows automatic rotation, compression, removal, and mailing of log files.
```
pgombar: nit: extra word, "is"

azurelinuxagent/common/osutil/default.py
```diff
@@ -1420,3 +1424,37 @@ def _update_nic_state(self, state, ip_command, handler, description):
...
+        :return: bool: Does the OS support logrotate
+        """
+        return True
+    
+    def get_agent_log_dir(self):
+        """
+        Default location of the log files. This should always be a folder. Make sure if there are any changes to this,
+        the log-fetch utility needs to change as well.
+
+        :return: bool: Does the OS support logrotate
```
pgombar: Needs to be updated.

azurelinuxagent/pa/deprovision/default.py
```diff
@@ -92,7 +93,22 @@ def del_dirs(self, warnings, actions):
...
+
+        :param warnings:
+        :param actions:
+        """
+
+        # We do the following check for isfile(get_agent_log_dir()) as a precaution in case the passed log-dir was
+        # incorrect and we defaulted to DEFAULT_LOG_DIR during the setup of log. Check
+        # azurelinuxagent.agent.Agent.__init__ to see the _Init log_ section, which changes the log_dir if the passed
+        # log_dir is a file and not a directory
+        log_file = os.path.join(self.osutil.get_agent_log_dir(), WAAGENT_LOG_FILE + "*") \
```
pgombar: I recommend changing the if to `if os.path.isdir` to make this line readable, it took me several tries to understand what was happening.

tests/common/test_logger.py
```diff
@@ -466,7 +467,8 @@ def setUp(self):
         self.event_dir = os.path.join(self.lib_dir, EVENTS_DIRECTORY)
         fileutil.mkdir(self.event_dir)
 
-        self.log_file = tempfile.mkstemp(prefix="logfile-")[1]
+        self.log_folder = tempfile.mkdtemp()
+        self.log_file = tempfile.mkstemp(prefix="logfile-", dir=self.log_folder)[1]
```
pgombar: From [the docs](https://docs.python.org/3/library/tempfile.html): "mkstemp() and mkdtemp() are lower-level functions which require manual cleanup."

vrdmr: Good to know. Added the cleanup in this test.

azurelinuxagent/common/osutil/nsbsd.py
```diff
@@ -153,3 +152,15 @@ def remove_firewall(self, dst_ip=None, uid=None):
...
+        return False
+
+    def get_agent_log_dir(self):
+        return "/log/azure"
+
+    def get_agent_log_rotation_maxbytes(self):
+        return 4000000
+
+    def get_agent_log_rotation_backupcount(self):
+        return 1
```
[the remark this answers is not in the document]
vrdmr: Its the configuration that @remip2 put it. I think it was because of the limited space available on the disks for `nsbsd` systems.

@remip2 Can you clarify this better?

azurelinuxagent/common/osutil/nsbsd.py
```diff
@@ -153,3 +152,15 @@ def remove_firewall(self, dst_ip=None, uid=None):
...
         return True
+
+    def supports_logrotate(self):
+        return False
+
+    def get_agent_log_dir(self):
+        return "/log/azure"
+
+    def get_agent_log_rotation_maxbytes(self):
+        return 4000000
```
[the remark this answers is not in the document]
vrdmr: Its the configuration that @remip2 put it. I think it was because of the limited space available on the disks for `nsbsd` systems.

@remip2 Can you clarify this better?

azurelinuxagent/common/osutil/default.py
```diff
@@ -1420,3 +1424,37 @@ def _update_nic_state(self, state, ip_command, handler, description):
...
                 else:
                     logger.error("Interface {0} has {1} but no link state".format(interface_name, description))
+
+    def supports_logrotate(self):
+        """
+        logrotate is allows automatic rotation, compression, removal, and mailing of log files.
+
+        :return: bool: Does the OS support logrotate
+        """
+        return True
```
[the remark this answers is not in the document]
vrdmr: > Do most distros (or at least the ones we endorse) support `logrotate`? Have we tested this?

The log rotation behavior currently used is `logrotate` and I didn't want to change that without data. So kept the Default as true, and la

> If not, it would be better to set the default behavior to "not supported" and "support" in respective OS utils classes.

This could be done later, once we figure out the state of logrotate. But anecdotally, most endorsed distros support `logrotate` and very few special cases don't have logrotate available.

azurelinuxagent/common/osutil/nsbsd.py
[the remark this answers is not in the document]
remip2: Yes, 4Mo is just the max size before rotation for all the other log files in nsbsd system. I just want to keep it consistent for nsbsd.

azurelinuxagent/common/osutil/nsbsd.py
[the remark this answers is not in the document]
remip2: yes I confirm, my main concern for nsbsd is the disk space and in most cases only one backup is enough to get all the useful information from the log.

azurelinuxagent/common/osutil/default.py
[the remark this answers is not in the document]
pgombar: So, if a distro doesn't support `logrotate`, it will inherit `True` from the default OS util and no logs will be rotated since the manual logic for log rotation won't be invoked. Since the purpose of this PR is to add log rotation, this means we can potentially leave **endorsed distros** without log rotation still.

I think it's worth investigating which of the endorsed distros don't support `logrotate` (if any) and override their `supports_logrotate` method in order to invoke manual log rotation, if needed.

azurelinuxagent/common/osutil/nsbsd.py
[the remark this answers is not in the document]
vrdmr: @remip2 - Can you try the current fix on an NSBSD VM? (with the changes I did on top of yours) - Please let me know the findings.

Also, how can I create an NSBSD VM in Azure?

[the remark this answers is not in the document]
remip2: ok I updated the agent on the existing vm and the rotation is working. I'll do a new  complete image and test the provisioning later.

To test you can deploy the vm from the marketplace : https://azuremarketplace.microsoft.com/fr-fr/marketplace/apps/stormshield.stormshield-elastic-virtual-appliance?tab=Overview
or via a github template : https://github.com/stormshield/azure-templates/tree/master/sns/sns-2-nics
The 'root' user is admin: ssh admin@<pub ip>
The files are located in: /usr/local/lib/python2.7/azurelinuxagent
and the log: /log/azure
To stop the agent : nstop waagent
To start the agent : nstart waagent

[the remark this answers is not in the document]
remip2: I tested the provisioning from a new image, all is working as expected.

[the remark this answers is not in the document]
vrdmr: Thanks @remip2 for verifying it - I verified it on an Ubuntu VM as well to see its rotation in action (The `cat`s in the middle were to inflate the size of the log file)

image

The only observation was that the need to archive the backup files. I'll see if I can send an update for it.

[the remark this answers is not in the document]
vrdmr: @remip2  - Added support for archiving backup files - https://github.com/Azure/WALinuxAgent/pull/1714/commits/5bda83b4bc26153a12db2da013021ee4b1d6253e. Can you please let me know the default desired for NSBSD?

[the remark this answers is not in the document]
remip2: @vrdmr I don't need gz for NSBSD, thanks

azurelinuxagent/common/logger.py
```diff
@@ -246,11 +246,17 @@ def _do_rollover(self):
...
 
-                # Archive the current log file.
-                with gzip.open(dst, 'wb') as archive:
+                # Archive the current log file. Surrounding by try:except to safegaurd against gzip exceptions, if any.
+                try:
+                    archive = gzip.open(dst, 'wb')
                     with open(self.path, 'rb') as logfile:
                         archive.writelines(logfile)
 
+                    archive.close()
```
pgombar, line 255: Safer way of handling opening archive files in Py2.6: https://stackoverflow.com/questions/6086603/statement-with-and-tarfile/6086722#6086722

tests/common/test_logger.py
```diff
@@ -621,11 +621,14 @@ def test_file_appender_with_logrotate_not_supported_archiving_backup(self):
             log_file_path = os.path.join(self.log_folder, log_file)
 
             # If the files are not gzip, the gzip library throws an OSError: Not a gzipped file.
-            with gzip.open(log_file_path, 'rb') as logfile:
-                logcontent = map(lambda x: x.strip(), logfile.readlines())
-                for logline in logcontent:
-                    # Only 7-13 are in the logs. Others shouldn't be in the logs.
-                    self.assertRegex(logline.decode('utf-8'), r".*INFO\s\w+\s*test-info-1?[0-9]")
+            logfile = gzip.open(log_file_path, 'rb')
```
pgombar, line 624: Same comment: https://stackoverflow.com/questions/6086603/statement-with-and-tarfile/6086722#6086722

azurelinuxagent/common/logger.py
vrdmr: Thanks.

tests/common/test_logger.py
vrdmr: Thanks for this.

azurelinuxagent/common/logger.py
```diff
@@ -219,18 +226,60 @@ def _should_rollover(self, msg):
...
+                if os.path.exists(dst):
+                    os.remove(dst)
+
+                # Archive the current log file. Surrounding by try:except to safeguard against gzip exceptions, if any.
+                try:
+                    with contextlib.closing(gzip.open(dst, 'wb')) as archive:
+                        with open(self.path, 'rb') as logfile:
+                            archive.writelines(logfile)
+                except Exception:
+                    pass  # Didn't add warning here as we are in the process of rotating logs while logging, and
```
larohra, line 257: >Didn't add warning here as we are in the process of rotating logs while logging

I get the sentiment behind this but at the same time we would want to know if something went wrong with log rotation, how about adding a Kusto event without adding it to log file?

azurelinuxagent/common/logger.py
```diff
@@ -219,18 +226,60 @@ def _should_rollover(self, msg):
...
+                    # we don't want to add burden
+
+                # Removing the plain log file as already archived.
+                os.remove(self.path)
+            else:
+                # Rotating the backup files to make space for the current log file.
+                for i in range(self._backup_count - 1, 0, -1):
```
larohra, line 264: Line 234-249 are exactly the same as Line 263-279. Could you please make it a private function instead? Thanks!

azurelinuxagent/common/logger.py
vrdmr: yes, but it could cause other issues if the add_event causes some logs to be written.

vrdmr:
Ping @narrieta @larohra @pgombar

azurelinuxagent/common/logger.py
```diff
@@ -260,24 +263,6 @@ def _do_rollover(self):
                 # Removing the plain log file as already archived.
                 os.remove(self.path)
             else:
```
larohra, line 265: I think you forgot to call the function here in else :)

vrdmr: Not required here. I optimized it above itself, based on `_should_archive_backup_files` attribute.

larohra:
Thank you for your contribution @remip2 and apologies for this long wait. 
Though we acknowledge the need for the gaps in our logging (as pointed out by this PR), after a lot of careful consideration we have decided not to take this PR in because this feels like a band-aid to a flawed logging structure that we currently have. That being said, we are working on revamping the agent completely and we will be redesigning its logging with keeping all of these points in mind, so thank you for pointing out these pain points.

I'm closing this PR for now, but rest assured that the required features would soon be implemented in the new agent. Thank you!

remip2:
Thanks for the update, I'm keeping the patch for our own build and will keep an eye for the logging redesign.

PR #1716, opened 2019-11-21
Added utc logging

larohra:
## Description
Currently the Linux agent logs in local time of the machine. Changing the logic to log in UTC.

Issue # 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1716)

### DIFF
--- a/tests/common/test_logger.py
+++ b/tests/common/test_logger.py
@@ -18,7 +18,7 @@
 import json
 import os
 from datetime import datetime
-from datetime import timedelta
+from datetime import datetime, timedelta
 
 from azurelinuxagent.common import logger
 from azurelinuxagent.common.event import add_log_event
--- a/tests/common/test_logger.py
+++ b/tests/common/test_logger.py
@@ -211,7 +211,7 @@ class TestLogger(AgentTestCase):
                 time_in_file = datetime.strptime(log.split(LogLevel.STRINGS[logger.LogLevel.INFO])[0].strip()
                                                  , u'%Y-%m-%dT%H:%M:%S.%fZ')
             except ValueError:
-                self.fail("Ensure timestamp is of format '%Y-%m-%dT%H:%M:%S.%fZ'")
+                self.fail("Ensure timestamp follows ISO-8601 format + 'Z' for UTC")
 
             # If the time difference is > 5secs, there's a high probability that the time_in_file is in different TZ
             self.assertTrue((time_in_file-before_write_utc) <= timedelta(seconds=5))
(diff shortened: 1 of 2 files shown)


tests/common/test_logger.py
```diff
@@ -192,3 +194,18 @@ def test_telemetry_logger1(self, mock_save):
...
+        test_logger.add_appender(logger.AppenderType.FILE, logger.LogLevel.INFO, path=file_path)
+
+        before_write_utc = datetime.utcnow().strftime(u'%Y/%m/%d %H:%M:%S')
+        test_logger.info("The time should be in UTC")
+
+        with open(file_path, "r") as log_file:
+            log = log_file.read()
+            time_in_file = log.split(LogLevel.STRINGS[logger.LogLevel.INFO])[0]
+
+            self.assertIn(before_write_utc, time_in_file)   # Comparing the UTC timestamp till seconds only
```
narrieta: but what if the minute flips? xx:10:59 to xx:11:00?

narrieta: oh, i misunderstood your comment - the second may change too, right?

larohra: I considered that scenario (it has a higher probability of second flipping too), but I feel we shouldn't see this problem as it would all happen within ms. 
Another way to approach this test is to mock utcnow() to return some random time and then verify if the time matches. I didn't want to go that route as I wanted to test the scenario with correct data. Do you feel that would be a better test?

narrieta: > we shouldn't see this problem 

 famous last words :) 

maybe you can take the delta between the two times and assert that it is within a few sec (different time zone would be hours)

larohra: Good point. I'll add the delta

larohra: Done

tests/common/test_logger.py
```diff
@@ -207,8 +207,11 @@ def test_logger_should_log_in_utc(self):
 
         with open(file_path, "r") as log_file:
             log = log_file.read()
-            time_in_file = datetime.strptime(log.split(LogLevel.STRINGS[logger.LogLevel.INFO])[0].strip()
-                                             , u'%Y/%m/%d %H:%M:%S.%f')
+            try:
+                time_in_file = datetime.strptime(log.split(LogLevel.STRINGS[logger.LogLevel.INFO])[0].strip()
```
narrieta: datetime.isoformat()

narrieta: sorry, not in the test, on line 92 logger.py

larohra: Nice I didnt know something like this existed! 1 problem with this though, it doens't print the 'Z' in the end

```python
datetime.isoformat(datetime.utcnow())
'2019-11-21T19:24:40.652963'

PR #1725, opened 2019-12-04
Initialize CPU usage

narrieta:
The CPU usage for the cgroup was not initialized when creating the CpuCgroup (the system usage was).

Now initialization sets both values and is done when we start tracking the cgroup (as opposed as when it is created). We can create cgroups and never track them so this is a better place for initialization.

Now CPU is computed as a percentage of usage for all cores (e.g. if a cgroup was using 2 cores at 100% on a 4-core system, now we report 50% instead of 200%). This makes data easier to use.

Also, added field #8 of /proc/stat to the computation of CPU (see "man proc"):
> steal (since Linux 2.6.11)
     (8) Stolen time, which is the time spent in other operating systems when running in a virtualized environment

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1725)


azurelinuxagent/common/cgroup.py
```diff
@@ -118,76 +118,72 @@ def is_active(self):
...
-        return round(float(cpu_delta * self._osutil.get_processor_cores() * 100) / float(system_delta), 3)
+        self._previous_cgroup_cpu = self._current_cgroup_cpu
+        self._previous_system_cpu = self._current_system_cpu
+        self._current_cgroup_cpu = self._get_cpu_ticks()
+        self._current_system_cpu = self._osutil.get_total_cpu_ticks_since_boot()
 
-    def get_cpu_usage(self):
-        """
-        Collects and return the cpu usage.
+        cgroup_delta = self._current_cgroup_cpu - self._previous_cgroup_cpu
```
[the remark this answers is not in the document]
narrieta: The CPU usage is measured in ticks. More recent values would never be less than older values so the difference (both for cgroup and system) would always be positive (or zero).

azurelinuxagent/common/cgroup.py
```diff
@@ -118,76 +118,72 @@ def is_active(self):
...
-        return cpu_total
+                cpu_ticks = int(m.groups()[0]) + int(m.groups()[1])
+
+        return cpu_ticks
 
-    def _update_cpu_data(self):
+    def _cpu_usage_initialized(self):
+        return self._current_cgroup_cpu is not None and self._current_system_cpu is not None
+
+    def initialize_cpu_usage(self):
```
vrdmr: Can we call `initialize_cpu_usage` within the `__init__` itself?

azurelinuxagent/common/osutil/default.py
```diff
@@ -1354,7 +1354,7 @@ def get_total_cpu_ticks_since_boot():
         if proc_stat is not None:
             for line in proc_stat.splitlines():
                 if ALL_CPUS_REGEX.match(line):
-                    system_cpu = sum(int(i) for i in line.split()[1:7])
+                    system_cpu = sum(int(i) for i in line.split()[1:8])
```
vrdmr: Can you add a comment on what the columns are?

tests/common/test_cgroups.py
```diff
@@ -110,49 +98,94 @@ def test_is_active_incorrect_file(self, patch_periodic_warn, *_):
...
+
+        original_read_file = fileutil.read_file
+
+        #
+        # Tests that need to mock the contents of /proc/stat or */cpuacct/stat can set this map from
+        # the file that needs to be mocked to the mock file (each test starts with an empty map).
+        #
+        cls.mock_read_file_map = {}
+
+        def mock_read_file(filepath, **args):
```
vrdmr: Thanks for fixing this. This looks better and cleaner.

azurelinuxagent/common/cgroup.py
narrieta: The get*cgroup functions instantiate cgroups that may never be tracked... I thought it was not worth initializing the cpu usage for those (especially since the actual cgroups may not exist on disk yet --- if we initialize on init the we would need to ensure the cgroups exist before calling the getter).

azurelinuxagent/common/osutil/default.py
narrieta: OK - I'll add a comment referencing what man page can be consulted (description can be sort of long to include here)

tests/common/test_cgroups.py
narrieta: np

tests/ga/test_monitor.py
```diff
@@ -914,7 +911,7 @@ def test_send_extension_metrics_telemetry_handling_memory_cgroup_exceptions_errn
         ioerror.errno = 2
         patch_get_memory_usage.side_effect = ioerror
 
-        CGroupsTelemetry._tracked.append(CpuCgroup("cgroup_name", "/test/path"))
+        CGroupsTelemetry._tracked.append(MemoryCgroup("cgroup_name", "/test/path"))
```
vrdmr: Thanks for fixing this.

PR #1726, opened 2019-12-05
Remove marking the extension as Failed

pgombar:
## Description

Today, during an extension upgrade, if the update command of the new extension fails, the new extension will be marked as Failed. This is the only usage of that extension handler state.

However, this means that, when a new goal state comes in, since the new extension is marked as Failed, agent will try to call enable on the new extension, skipping update, because the extension handler's state is not equal to [NotInstalled](https://github.com/Azure/WALinuxAgent/blob/develop/azurelinuxagent/ga/exthandlers.py#L487).

This PR ensures that the entire upgrade scenario is called even if the previous goal state resulted in a failed update step. This is done by removing the Failed state as an option and, as a result, when the update command fails, the new extension will be removed, so when the new goal state comes in, it will be treated as a new - not previously installed - extension and the entire upgrade scenario will be exercised.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1726)

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -203,7 +203,6 @@ class ExtHandlerState(object):
     NotInstalled = "NotInstalled"
     Installed = "Installed"
     Enabled = "Enabled"
-    Failed = "Failed"
 
 
 def get_exthandlers_handler():
@@ -1023,19 +1022,14 @@ class ExtHandlerInstance(object):
         env = {'VERSION': version, ExtCommandEnvVariable.DisableReturnCode: disable_exit_code,
                ExtCommandEnvVariable.UpdatingFromVersion: updating_from_version}
 
-        try:
-            self.set_operation(WALAEventOperation.Update)
-            man = self.load_manifest()
-            update_cmd = man.get_update_command()
-            self.logger.info("Update extension [{0}]".format(update_cmd))
-            self.launch_command(update_cmd,
-                                timeout=900,
-                                extension_error_code=ExtensionErrorCodes.PluginUpdateProcessingFailed,
-                                env=env)
-        except ExtensionError:
-            # prevent the handler update from being retried
-            self.set_handler_state(ExtHandlerState.Failed)
-            raise
+        self.set_operation(WALAEventOperation.Update)
+        man = self.load_manifest()
+        update_cmd = man.get_update_command()
+        self.logger.info("Update extension [{0}]".format(update_cmd))
+        self.launch_command(update_cmd,
+                            timeout=900,
+                            extension_error_code=ExtensionErrorCodes.PluginUpdateProcessingFailed,
+                            env=env)
 
     def update_with_install(self, uninstall_exit_code=None):
         man = self.load_manifest()
--- a/tests/ga/test_extension.py
+++ b/tests/ga/test_extension.py
@@ -1464,24 +1464,32 @@ class TestExtension(ExtensionTestCase):
         self._assert_no_handler_status(protocol.report_vm_status)
 
     @patch('azurelinuxagent.ga.exthandlers.HandlerManifest.get_update_command')
-    def test_upgrade_failure(self, patch_get_update_command, *args):
+    def test_extension_upgrade_failure_when_update_fails(self, patch_get_update_command, *args):
         """
-        Extension upgrade failure should not be retried
+        When the update command of the new extension fails, it should result in the new extension removed and the
+        old extension disabled. On the next goal state, the entire upgrade scenario should be retried (once),
+        meaning the new extension is downloaded again and update is called again.
         """
         test_data, exthandlers_handler, protocol = self._set_up_update_test_and_update_gs(patch_get_update_command,
                                                                                           *args)
 
-        exthandlers_handler.run()
-        self.assertEqual(1, patch_get_update_command.call_count)
+        with patch('azurelinuxagent.ga.exthandlers.HandlerManifest.get_enable_command') as patch_get_enable_command:
+            exthandlers_handler.run()
+            self.assertEqual(1, patch_get_update_command.call_count)
+            self.assertEqual(0, patch_get_enable_command.call_count)
 
-        # On the next iteration, update should not be retried
-        exthandlers_handler.run()
-        self.assertEqual(1, patch_get_update_command.call_count)
+            # On the next iteration, update should be retried.
+            exthandlers_handler.run()
+            self.assertEqual(1, patch_get_update_command.call_count)
+            self.assertEqual(0, patch_get_enable_command.call_count)
 
-        self._assert_handler_status(protocol.report_vm_status, "NotReady", expected_ext_count=1, version="1.0.1")
+        self.assertEquals("Ready", protocol.report_vm_status.call_args[0][0].vmAgent.status)
+
+        # Since the old extension is disabled, and the new one never got enabled, there is nothing to report.
+        self._assert_no_handler_status(protocol.report_vm_status)
 
     @patch('azurelinuxagent.ga.exthandlers.HandlerManifest.get_disable_command')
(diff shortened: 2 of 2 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -1023,19 +1022,14 @@ def update(self, version=None, disable_exit_code=None, updating_from_version=Non
...
-            self.set_handler_state(ExtHandlerState.Failed)
-            raise
+        self.set_operation(WALAEventOperation.Update)
+        man = self.load_manifest()
+        update_cmd = man.get_update_command()
+        self.logger.info("Update extension [{0}]".format(update_cmd))
+        self.launch_command(update_cmd,
+                            timeout=900,
+                            extension_error_code=ExtensionErrorCodes.PluginUpdateProcessingFailed,
+                            env=env)
```
narrieta: let's double-check report_ext_handler_status --- seems like the update failure would have been reported to crp

larohra: Also double check the PR where they added the FAILED state, they did it to deal with going in the infinite loop I think. The PR would give you the motivation behind adding that and to see if this doesnt reintroduce the problem

tests/ga/test_extension.py
```diff
@@ -1464,24 +1464,32 @@ def test_uninstall_failure(self, patch_get_uninstall_command, *args):
         self._assert_no_handler_status(protocol.report_vm_status)
 
     @patch('azurelinuxagent.ga.exthandlers.HandlerManifest.get_update_command')
-    def test_upgrade_failure(self, patch_get_update_command, *args):
+    def test_extension_upgrade_failure_when_update_fails(self, patch_get_update_command, *args):
         """
-        Extension upgrade failure should not be retried
+        When the update command of the new extension fails, it should result in the new extension removed and the
+        old extension disabled. On the next goal state, the entire upgrade scenario should be retried (once),
+        meaning the new extension is downloaded again and update is called again.
```
larohra: If there's some bug in the update.cmd of the extension, wont this cause the agent to download and delete the zip file at every GS? Aesthetically that would make sense since the code is buggy, but it would still cause extra network bandwidth for the customer. Should we handle this scenario?

narrieta: Initially I was thinking along those lines, too. But since now we go in the loop only on new goal state, my thought was that re-downloading was ok. What do you guys think? @pgombar @vrdmr

tests/ga/test_extension.py
```diff
@@ -1464,24 +1464,32 @@ def test_uninstall_failure(self, patch_get_uninstall_command, *args):
...
         """
         test_data, exthandlers_handler, protocol = self._set_up_update_test_and_update_gs(patch_get_update_command,
                                                                                           *args)
 
-        exthandlers_handler.run()
-        self.assertEqual(1, patch_get_update_command.call_count)
+        with patch('azurelinuxagent.ga.exthandlers.HandlerManifest.get_enable_command') as patch_get_enable_command:
+            exthandlers_handler.run()
+            self.assertEqual(1, patch_get_update_command.call_count)
```
narrieta: I've always felt that those asserts on the get_*_command functions are too weak. As the code evolves they may be other reason why they are not called. For future tests maybe we can think of stronger asserts? (maybe mock popen and assert that the corresponding command is never invoked?)

narrieta: for example, below, at line 1483... the getter may have been called, but (as the code evolves) maybe the command was never executed

tests/ga/test_extension.py
larohra: My concern is mostly about those extensions that the customer doesnt even want or know about (like OMS) and which are extremely buggy. Extensions like this would cause unnecessary traffic which might make them mad. Not sure if that is still part of the scope of this PR

tests/ga/test_extension.py
```diff
@@ -1464,24 +1464,32 @@ def test_uninstall_failure(self, patch_get_uninstall_command, *args):
...
-        # On the next iteration, update should not be retried
-        exthandlers_handler.run()
-        self.assertEqual(1, patch_get_update_command.call_count)
+            # On the next iteration, update should be retried.
+            exthandlers_handler.run()
+            self.assertEqual(1, patch_get_update_command.call_count)
+            self.assertEqual(0, patch_get_enable_command.call_count)
 
-        self._assert_handler_status(protocol.report_vm_status, "NotReady", expected_ext_count=1, version="1.0.1")
+        self.assertEquals("Ready", protocol.report_vm_status.call_args[0][0].vmAgent.status)
```
narrieta: why the change from NotReady to Ready?

tests/ga/test_extension.py
```diff
@@ -1464,24 +1464,32 @@ def test_uninstall_failure(self, patch_get_uninstall_command, *args):
...
+
+        # Since the old extension is disabled, and the new one never got enabled, there is nothing to report.
+        self._assert_no_handler_status(protocol.report_vm_status)
```
narrieta: as discussed offline, the failure updating to the new version should be reported.

tests/ga/test_extension.py
vrdmr: > wont this cause the agent to download and delete the zip file at every GS?

The zip files for the new update would still be downloaded but only one-time per GS (not infinite). I feel it is okay. Yes, our downloads might increase in a fast-changing VM, but still would be far less than the earlier problem (similar to what @narrieta said)

The bug of infinite that came back in 2.2.44 was due to the issue that cleanup code does not have the ExtHandler models of the updated extension and thus assumes it is an outdated version. And there is no easy way to fix that problem if we still want to allow infinite attempts.

tests/ga/test_extension.py
```diff
@@ -1464,24 +1464,32 @@ def test_uninstall_failure(self, patch_get_uninstall_command, *args):
...
         """
         test_data, exthandlers_handler, protocol = self._set_up_update_test_and_update_gs(patch_get_update_command,
                                                                                           *args)
 
-        exthandlers_handler.run()
-        self.assertEqual(1, patch_get_update_command.call_count)
+        with patch('azurelinuxagent.ga.exthandlers.HandlerManifest.get_enable_command') as patch_get_enable_command:
+            exthandlers_handler.run()
+            self.assertEqual(1, patch_get_update_command.call_count)
+            self.assertEqual(0, patch_get_enable_command.call_count)
```
vrdmr: Can you also assert if the error code returned was `ExtensionErrorCodes.PluginUpdateProcessingFailed` to make the assert stronger?

tests/ga/test_extension.py
vrdmr: Why is it `Ready`? If the Update failed?

tests/ga/test_extension.py
pgombar: I think it's ok to have the extension get downloaded once per goal state since we got rid of the upgrade loop.

tests/ga/test_extension.py
pgombar: You are comparing two different things -- the old NotReady assertion was testing the handler status, this assertion is testing the agent status. Testing the agent status isn't really needed here, I added it as an extra precaution. I will update the tests to make it clearer.

pgombar: See my comment above. I will send an update to make it clearer.

tests/ga/test_extension.py
pgombar: Updating the tests to reflect this assertion earlier, after the first `exthandlers_handler.run()`, when the failure is reported.

tests/ga/test_extension.py
pgombar: Just in case the ordering of the comments changes, I'm referring to this reply:

You are comparing two different things -- the old NotReady assertion was testing the handler status, this assertion is testing the agent status. Testing the agent status isn't really needed here, I added it as an extra precaution. I will update the tests to make it clearer.

pgombar:
Abandoning in favour of #1731.

PR #1731, opened 2019-12-11
Fix upgrade sequence when update command fails

pgombar:
## Description

Today, during an extension upgrade, if the update command of the new extension fails, the new extension will be marked as Failed and on the next goal state we would directly call enable on the new version instead of calling download, initialize, and update beforehand due to [this check](https://github.com/Azure/WALinuxAgent/blob/develop/azurelinuxagent/ga/exthandlers.py#L487). This PR fixes the condition by including the `FailedUpgrade` state, ensuring we don't skip any steps in the upgrade scenario in subsequent goal states.

We need to keep the failed (new) version on disk in order to keep reporting its failed status. Since we don't clean up the failed extension, this means that we won't re-download it on the next goal state, but re-use the existing zip on disk. The old extension is also present on disk and disabled. When uninstall is called, the old version is uninstalled and removed. The new version's zip still persists (and is in a failed state). Improvements on this clean up logic can be addressed in a separate PR.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1731)

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -203,7 +203,7 @@ class ExtHandlerState(object):
     NotInstalled = "NotInstalled"
     Installed = "Installed"
     Enabled = "Enabled"
-    Failed = "Failed"
+    FailedUpgrade = "FailedUpgrade"
 
 
 def get_exthandlers_handler():
@@ -484,7 +484,9 @@ class ExtHandlersHandler(object):
         handler_state = ext_handler_i.get_handler_state()
         ext_handler_i.logger.info("[Enable] current handler state is: {0}",
                                   handler_state.lower())
-        if handler_state == ExtHandlerState.NotInstalled:
+        # We go through the entire process of downloading and initializing the extension if it's either a fresh
+        # extension or if it's a retry of a previously failed upgrade.
+        if handler_state == ExtHandlerState.NotInstalled or handler_state == ExtHandlerState.FailedUpgrade:
             ext_handler_i.set_handler_state(ExtHandlerState.NotInstalled)
             ext_handler_i.download()
             ext_handler_i.initialize()
@@ -764,17 +766,17 @@ class ExtHandlerInstance(object):
         return FlexibleVersion(self_version) != FlexibleVersion(other_version)
 
     def get_installed_ext_handler(self):
-        lastest_version = self.get_installed_version()
-        if lastest_version is None:
+        latest_version = self.get_installed_version()
+        if latest_version is None:
             return None
 
         installed_handler = ExtHandler()
         set_properties("ExtHandler", installed_handler, get_properties(self.ext_handler))
-        installed_handler.properties.version = lastest_version
+        installed_handler.properties.version = latest_version
         return ExtHandlerInstance(installed_handler, self.protocol)
 
     def get_installed_version(self):
-        lastest_version = None
+        latest_version = None
 
         for path in glob.iglob(os.path.join(conf.get_lib_dir(), self.ext_handler.name + "-*")):
             if not os.path.isdir(path):
@@ -784,17 +786,16 @@ class ExtHandlerInstance(object):
             version_from_path = FlexibleVersion(path[separator + 1:])
             state_path = os.path.join(path, 'config', 'HandlerState')
 
-            if not os.path.exists(state_path) or \
-                    fileutil.read_file(state_path) == \
-                    ExtHandlerState.NotInstalled:
-                logger.verbose("Ignoring version of uninstalled extension: "
+            if not os.path.exists(state_path) or fileutil.read_file(state_path) == ExtHandlerState.NotInstalled \
+                    or fileutil.read_file(state_path) == ExtHandlerState.FailedUpgrade:
+                logger.verbose("Ignoring version of uninstalled or failed extension: "
                                "{0}".format(path))
                 continue
 
-            if lastest_version is None or lastest_version < version_from_path:
-                lastest_version = version_from_path
+            if latest_version is None or latest_version < version_from_path:
+                latest_version = version_from_path
 
-        return str(lastest_version) if lastest_version is not None else None
+        return str(latest_version) if latest_version is not None else None
 
     def copy_status_files(self, old_ext_handler_i):
         self.logger.info("Copy status files from old plugin to new")
@@ -1033,8 +1034,8 @@ class ExtHandlerInstance(object):
                                 extension_error_code=ExtensionErrorCodes.PluginUpdateProcessingFailed,
                                 env=env)
         except ExtensionError:
-            # prevent the handler update from being retried
-            self.set_handler_state(ExtHandlerState.Failed)
+            # Mark the handler as Failed so we don't clean it up and can keep reporting its status
+            self.set_handler_state(ExtHandlerState.FailedUpgrade)
             raise
 
     def update_with_install(self, uninstall_exit_code=None):
--- a/tests/ga/test_extension.py
+++ b/tests/ga/test_extension.py
@@ -16,6 +16,7 @@
 #
 
 import os.path
+import subprocess
 import unittest
 
 from datetime import timedelta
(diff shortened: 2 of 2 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -784,17 +786,16 @@ def get_installed_version(self):
             version_from_path = FlexibleVersion(path[separator + 1:])
             state_path = os.path.join(path, 'config', 'HandlerState')
 
-            if not os.path.exists(state_path) or \
-                    fileutil.read_file(state_path) == \
-                    ExtHandlerState.NotInstalled:
-                logger.verbose("Ignoring version of uninstalled extension: "
+            if not os.path.exists(state_path) or fileutil.read_file(state_path) == ExtHandlerState.NotInstalled \
+                    or fileutil.read_file(state_path) == ExtHandlerState.FailedUpgrade:
```
narrieta: why is the extra check needed?

tests/ga/test_extension.py
```diff
@@ -1464,24 +1464,45 @@ def test_uninstall_failure(self, patch_get_uninstall_command, *args):
...
+        NotInstalled handlers), so we just re-use the existing zip.
         """
         test_data, exthandlers_handler, protocol = self._set_up_update_test_and_update_gs(patch_get_update_command,
                                                                                           *args)
 
-        exthandlers_handler.run()
-        self.assertEqual(1, patch_get_update_command.call_count)
+        with patch('azurelinuxagent.ga.exthandlers.HandlerManifest.get_enable_command') as patch_get_enable_command:
+            exthandlers_handler.run()
+            self.assertEqual(1, patch_get_update_command.call_count)
```
narrieta: let's stop adding asserts that check whether the command was executed by using the get_* functions. You could mock popen instead, for example.

tests/ga/test_extension.py
```diff
@@ -1464,24 +1464,45 @@ def test_uninstall_failure(self, patch_get_uninstall_command, *args):
         self._assert_no_handler_status(protocol.report_vm_status)
 
     @patch('azurelinuxagent.ga.exthandlers.HandlerManifest.get_update_command')
-    def test_upgrade_failure(self, patch_get_update_command, *args):
+    def test_extension_upgrade_failure_when_new_version_update_fails(self, patch_get_update_command, *args):
         """
-        Extension upgrade failure should not be retried
+        When the update command of the new extension fails, it should result in the new extension removed and the
```
narrieta: "the new extension removed"  it is not removed, right?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -764,17 +766,17 @@ def version_ne(self, other):
         return FlexibleVersion(self_version) != FlexibleVersion(other_version)
 
     def get_installed_ext_handler(self):
-        lastest_version = self.get_installed_version()
```
larohra: Wow in all this time never saw this typo! Thanks for fixing it!

tests/ga/test_extension.py
```diff
@@ -1464,24 +1464,45 @@ def test_uninstall_failure(self, patch_get_uninstall_command, *args):
...
+        old extension disabled. On the next goal state, the entire upgrade scenario should be retried (once),
```
larohra: Quick question, if the update fails then the old one is disabled, on a new GS would disable for the old one be called again?
Previously because of the Failed state we were not calling Disable again but calling the enable on the new one directly

tests/ga/test_extension.py
larohra: +1

azurelinuxagent/ga/exthandlers.py
pgombar: Me neither until I had to closely read this part of the code :D

tests/ga/test_extension.py
pgombar: Correct, this is an outdated comment, I'll update it. Thanks!

azurelinuxagent/ga/exthandlers.py
narrieta: Oy, and me thinking "lastest" was the "last of the latest" (as in "bestest")!

azurelinuxagent/ga/exthandlers.py
pgombar: As the name of the method suggests, the purpose is to find the latest installed extension version. Without the added check to ignore `FailedUpgrade` states, when a second goal state comes in (so, when the old version is disabled and the new one is failed), this method would be called through `handle_enable` and would actually retrieve the new version as the `old_exthandler` simply because it's state is **not** NotInstalled, which is incorrect.

narrieta: yes, that what it looks like, but is that happening with the current code?

tests/ga/test_extension.py
pgombar: Yes, disable would be called on the old extension again because this time we'd go into [this code path](https://github.com/Azure/WALinuxAgent/blob/develop/azurelinuxagent/ga/exthandlers.py#L495).

azurelinuxagent/ga/exthandlers.py
narrieta: (i want to double-check we are not inadvertently changing some other behavior)

I don't remember seeing a case like that, though most of the upgrade failures we see are on disable, and not on update.

pgombar: In the current code (pre-this PR), we'd retrieve the new (failed) extension as the `old_exthandler_i` in [this line](https://github.com/Azure/WALinuxAgent/blob/develop/azurelinuxagent/ga/exthandlers.py#L482), but, because we'd jump straight to [this line](https://github.com/Azure/WALinuxAgent/blob/develop/azurelinuxagent/ga/exthandlers.py#L497), `old_exthandler_i` wasn't being used anywhere. So basically one bug was hiding the other.

narrieta: ok

tests/ga/test_extension.py
larohra: I wonder if this would throw extensions in a weird state because they may or may not be handling a scenario where disable is called multiple times. We'll have to make sure they know about this change so that they can modify their code too if needed

pgombar: The contract with extension publishers is that each operation should be idempotent. From the document:

> Every time the machine restarts (which in turn will restart the Azure Agent) or a new goal state is received, the Azure Agent will try to set the machine to the latest goal state. Thus it might invoke the enabled\disabled commands multiple times even if the handler is already enabled\disabled. So the enable and disable commands need to be idempotent

narrieta: I'd say we can go ahead, since disable is supposed to be idempotent. From our publisher's doc:
> Unlike the install state, the Azure Agent will not maintain the enabled\disabled state of the handler. Every time the machine restarts (which in turn will restart the Azure Agent) or a new goal state is received, the Azure Agent will try to set the machine to the latest goal state. Thus it might invoke the enabled\disabled commands multiple times even if the handler is already enabled\disabled. So the enable and disable commands need to be idempotent i.e. if the handler is already enabled and the enable command is invoked again, the command should check if all the processes are running as expected, if yes, then the command should just exit with a success code.

larohra: Nice! I didnt know about disable, I thought we just said that enable has to be idempotent. THansk for clarifying it guys :)

tests/ga/test_extension.py
pgombar: Updated.

tests/ga/test_extension.py
```diff
@@ -1464,24 +1465,68 @@ def test_uninstall_failure(self, patch_get_uninstall_command, *args):
...
+
+            update_command_count = len([extension_call for extension_call in extension_calls
+                                        if patch_get_update_command.return_value in extension_call])
+            enable_command_count = len([extension_call for extension_call in extension_calls
+                                        if "-enable" in extension_call])
+            self.assertEquals(2, update_command_count)
+            self.assertEquals(0, enable_command_count)
+
+            # We report the failure of the new extension version
+            self._assert_handler_status(protocol.report_vm_status, "NotReady", expected_ext_count=1, version="1.0.1")
```
vrdmr: Just a minor comment - You can add another assert here (not needed in this PR), to validate both the directories in the lib_dir. This would make it concrete that we don't remove any directory.

larohra: +1

pgombar: Good point. I'll keep it in mind when the time comes for the inevitable refactoring of all these tests in `test_extension.py` since all of the assertions are quite vague.

Issue #1473, opened 2019-02-25
[BUG] Warning: the imp module is deprecated in favour of importlib


johanburati:
Getting a warning on Ubuntu 19.04 

**details:**
```
$ waagent --version
/usr/sbin/waagent:27: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import imp
WALinuxAgent-2.2.32.2 running on ubuntu 19.04
Python: 3.7.2
Goal state agent: 2.2.37

$ cat /etc/os-release
NAME="Ubuntu"
VERSION="19.04 (Disco Dingo)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu Disco Dingo (development branch)"
VERSION_ID="19.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=disco
UBUNTU_CODENAME=disco
```

vrdmr:
@roiyz-msft  - It is imported by the `waagent` file in Ubuntu's new VMs (which only have Py3 by default)

```
import os
import imp
import sys

if __name__ == '__main__' :
    import azurelinuxagent.agent as agent
    """
    Invoke main method of agent
    """
    agent.main()
```

Deprecation warning.
```
>>> import imp
__main__:1: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
>>> 
```

amkarmak:
I am also getting the same error.. 

root@ubuntu19:~# waagent --version
/usr/sbin/waagent:27: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import imp
WALinuxAgent-2.2.40 running on ubuntu 19.04
Python: 3.7.3
Goal state agent: 2.2.45
root@ubuntu19:~#

Anyone got any fix for this issue?

narrieta:
@amkarmak - This should just be a warning, is the agent somehow failing? Thanks

amkarmak:
Hi Norberto, no the agent is not failing. But this messages gets popped up.

amkarmak:
Hi Johan, thank you for the response. Sure, I will do this and update.

[...]

PR #1738, opened 2019-12-18
Replace module imp with importlib

johanburati:
## Description

Issue https://github.com/Azure/WALinuxAgent/issues/1473

https://www.python-future.org/_modules/imp.html
> In most cases it is preferred you consider using the importlib module's functionality over this module.

---

### PR information
- [x]  Except for special cases involving multiple contributors, the PR is started from a fork of the main repository, not a branch.
- [x]  If applicable, the PR references the bug/issue that it fixes in the description.

- [x]  I have read the [contribution guidelines](https://github.com/Azure/WALinuxAgent/blob/master/.github/CONTRIBUTING.md).

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1738)


bin/waagent
```diff
@@ -24,9 +24,13 @@
 #
 
 import os
-import importlib
 import sys
 
+if sys.version_info[0] == 2:
+    import imp
+else:
+    import importlib
```
rainer37, line 32: does importlib do anything or potentially going to do anything here?

pgombar: `importlib` is the preferred module since `imp` is deprecated. We define it here, the same way like `imp` was defined here so far. Does that answer your question?

rainer37: Sorry, i understand that imp is deprecated for python3, and importlib is the succeeder. But i don't see any usage of importlib in this file, and the code snippet where was using 'imp' is never supposed to be executed with python3, so i think it would be the same if we just remove the else part. Please correct me if i'm wrong, thanks!

pgombar: @rainer37 I see your point now. You are right, since we have:
```
    if sys.version_info[0] == 3:
        raise ImportError("waagent2.0 doesn't support python3")
```
`imp` would never be executed with python3.

However, the imports will always be evaluated, which means that today, in an image that runs the agent with Python3, we would see this:
```
root@ubuntu2004:~# waagent --version
/usr/sbin/waagent:27: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import imp
WALinuxAgent-2.2.46 running on ubuntu 20.04
Python: 3.8.2
Goal state agent: 2.2.46
```
The purpose of this change is to get rid of this warning.

rainer37: @pgombar thanks for the info! but sorry i think 
```
if sys.version_info[0] == 2:
    import imp
```
is enough to silence the DeprecationWarning for os with /usr/bin/env python as python3* since python3 wouldn't take this branch, and python3 is not going to do any module load using 'importlib'. S adding
```
else:
    import importlib
```
does not seem to help in any cases but only import a module that is not used. Any further explanation would be much appreciated, thanks!

pgombar: Looking only at this file's contents, you are right and there is no need for `importlib` since it is not called anywhere in this copy of `bin/waagent`. However, there are certain implicit dependencies between `bin/waagent` and VM extensions. In case the extensions need to load modules from the agent, and are using Python3, we are enabling them to do so with `importlib`.

pgombar:
@johanburati thanks for your contribution!

PR #1741, opened 2019-12-21
Do not update goal state when refreshing the host plugin

narrieta:
We are updating the goal state too frequently and that is part of the reason different components are getting out of sync. One of the reasons we were forcing updates of the goal state was to retry calls to the host plugin when the container id or the role config change. In this case updating the full goal state is not needed and it can actually introduce consistency issues.

The main change in this PR is to define an API specifically to update the host plugin without changing the goal state.

Other small changes are minor cleanup of the update goal state in the metadata protocol and the parse code of the goal state; I'll add comments in the PR to explain those changes.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1741)

### DIFF
--- a/azurelinuxagent/common/protocol/metadata.py
+++ b/azurelinuxagent/common/protocol/metadata.py
@@ -170,7 +170,7 @@ class MetadataProtocol(Protocol):
                                 "{0}.crt".format(thumbprint))
         shutil.copyfile(trans_prv_file, prv_file)
         shutil.copyfile(trans_cert_file, crt_file)
-        self.update_goal_state(forced=True)
+        self.update_goal_state()
 
     def get_vminfo(self):
         vminfo = VMInfo()
@@ -330,7 +330,7 @@ class MetadataProtocol(Protocol):
         certificates = self.get_certs()
         return certificates.cert_list
 
-    def update_goal_state(self, forced=False, max_retry=3):
+    def update_goal_state(self, max_retry=3):
         # Start updating goalstate, retry on 410
         for retry in range(0, max_retry):
             try:
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -106,10 +106,8 @@ class WireProtocol(Protocol):
         cryptutil = CryptUtil(conf.get_openssl_cmd())
         cryptutil.gen_transport_cert(trans_prv_file, trans_cert_file)
 
-        self.update_goal_state(forced=True)
-
-    def update_goal_state(self, forced=False, max_retry=3):
-        self.client.update_goal_state(forced=forced, max_retry=max_retry)
+        # Set the initial goal state
+        self.client.update_goal_state(forced=True)
 
     def get_vminfo(self):
         goal_state = self.client.get_goal_state()
@@ -136,7 +134,7 @@ class WireProtocol(Protocol):
 
     def get_vmagent_manifests(self):
         # Update goal state to get latest extensions config
-        self.update_goal_state()
+        self.client.update_goal_state()
         goal_state = self.client.get_goal_state()
         ext_conf = self.client.get_ext_conf()
         return ext_conf.vmagent_manifests, goal_state.incarnation
@@ -150,7 +148,7 @@ class WireProtocol(Protocol):
     def get_ext_handlers(self):
         logger.verbose("Get extension handler config")
         # Update goal state to get latest extensions config
-        self.update_goal_state()
+        self.client.update_goal_state()
         goal_state = self.client.get_goal_state()
         ext_conf = self.client.get_ext_conf()
         # In wire protocol, incarnation is equivalent to ETag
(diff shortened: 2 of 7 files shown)


azurelinuxagent/common/protocol/metadata.py
```diff
@@ -170,7 +170,7 @@ def detect(self):
                                 "{0}.crt".format(thumbprint))
         shutil.copyfile(trans_prv_file, prv_file)
         shutil.copyfile(trans_cert_file, crt_file)
-        self.update_goal_state(forced=True)
+        self.update_goal_state()
```
narrieta: the forced flag in metadata was not implemented so I removed it

azurelinuxagent/common/protocol/wire.py
```diff
@@ -106,10 +106,8 @@ def detect(self):
         cryptutil = CryptUtil(conf.get_openssl_cmd())
         cryptutil.gen_transport_cert(trans_prv_file, trans_cert_file)
 
-        self.update_goal_state(forced=True)
-
-    def update_goal_state(self, forced=False, max_retry=3):
```
narrieta: I removed update_goal_state() from the protocol since it can cause confusion

azurelinuxagent/common/protocol/wire.py
```diff
@@ -761,78 +759,88 @@ def update_ext_conf(self, goal_state):
...
+    # Type of update performed by _update_from_goal_state()
+    class UpdateType(object):
+        # Update the Host GA Plugin client (Contained ID and RoleConfigName)
+        HostPlugin = 0
+        # Update the full goal state only if the incarnation has changed
+        GoalState = 1
+        # Update the full goal state unconditionally
+        GoalStateForced = 2
+
+    def update_host_plugin_from_goal_state(self):
```
narrieta: new api to refresh the host plugin

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1393,15 +1399,9 @@ def parse(self, xml_text):
         self.remote_access_uri = findtext(container, "RemoteAccessInfo")
         lbprobe_ports = find(xml_doc, "LBProbePorts")
         self.load_balancer_probe_port = findtext(lbprobe_ports, "Port")
-        return self
 
 
 class HostingEnv(object):
-    """
```
narrieta: many commens in this code were totally out of whack --- maybe copy and paste errors; removed them

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1410,12 +1410,9 @@ def __init__(self, xml_text):
...
         self.deployment_name = None
         self.xml_text = None
-        self.parse(xml_text)
+        self._parse(xml_text)
 
-    def parse(self, xml_text):
-        """
-        parse and create HostingEnvironmentConfig.xml.
-        """
+    def _parse(self, xml_text):
```
narrieta: these functions should be private and the return value is never used

azurelinuxagent/daemon/main.py
```diff
@@ -150,10 +150,9 @@ def daemon(self, child_args=None):
                 #   incarnation number. A forced update ensures the most
                 #   current values.
                 protocol = self.protocol_util.get_protocol()
-                client = protocol.client
-                if client is None or type(client) is not WireClient:
+                if type(protocol) is not WireProtocol:
```
narrieta: The metadata protocol doesn't have a 'client' member so this would have raised on that protocol. Luckily rdma is not enabled there.  I changed the check to use the protocol instance instead

tests/ga/test_update.py
```diff
@@ -1717,9 +1716,8 @@ def get_vmagent_pkgs(self, manifest):
             raise ResourceGoneError()
         return self.agent_packages
 
-    def update_goal_state(self, forced=False, max_retry=3):
+    def update_goal_state(self, full_goal_state=False, max_retry=3):
```
narrieta: left-over rename; will undo

tests/protocol/mockwiredata.py
```diff
@@ -182,3 +200,52 @@ def mock_gen_trans_cert(self, trans_prv_file, trans_cert_file):
...
         with open(trans_cert_file, 'w+') as cert_file:
             cert_file.write(self.trans_cert)
+
+    #
+    # Having trouble reading the regular expressions below? you are not alone!
+    #
+    # For the use of "(?<=" "(?=" see 7.2.1 in https://docs.python.org/3.1/library/re.html
+    # For the use of "\g<1>" see backreferences in https://docs.python.org/3.1/library/re.html#re.sub
+    #
+    # Note that these regular expressions are not enough to _parse all valid XML documents (e.g. they do
```
narrieta: typo: "_parse"

tests/protocol/test_wire.py
```diff
@@ -610,73 +611,6 @@ def test_get_ext_conf_with_uri(self, *args):
         self.assertEqual("BlockBlob", ext_conf.status_upload_blob_type)
         self.assertEqual(None, ext_conf.artifacts_profile_blob)
 
-    def test_save_or_update_goal_state_should_save_new_goal_state_file(self):
```
narrieta: these were tests for a trivial function that should have been private in the first place... no need for them

tests/protocol/mockwiredata.py
```diff
@@ -182,3 +200,52 @@ def mock_gen_trans_cert(self, trans_prv_file, trans_cert_file):
 
         with open(trans_cert_file, 'w+') as cert_file:
             cert_file.write(self.trans_cert)
+
+    #
+    # Having trouble reading the regular expressions below? you are not alone!
```
larohra: Thanks for adding this small guide on regular expressions and the helper functions, they'll be really helpful! :)

tests/protocol/mockwiredata.py
```diff
@@ -182,3 +200,52 @@ def mock_gen_trans_cert(self, trans_prv_file, trans_cert_file):
...
+    #
+    # For the use of "(?<=" "(?=" see 7.2.1 in https://docs.python.org/3.1/library/re.html
+    # For the use of "\g<1>" see backreferences in https://docs.python.org/3.1/library/re.html#re.sub
+    #
+    # Note that these regular expressions are not enough to _parse all valid XML documents (e.g. they do
+    # not account for metacharacters like < or > in the values) but they are good enough for the test
+    # data. There are some basic checks, but the functions could could not match valid XML or produce
```
narrieta: typo: "could could not match" -> "may not match"

tests/protocol/test_wire.py
```diff
@@ -1114,15 +1048,170 @@ def host_func(*args):
...
+
+    def tearDown(self):
+        self.mock_crypt_util.stop()
+        self.mock_http_get.stop()
+        AgentTestCase.tearDown(self)
+
+
+class UpdateGoalStateTestCase(GoalStateTestCase):
+    """
+    Tests for WireClient.update_host_plugin_from_goal_state()
```
larohra: NIT: Typo - should be 
`Tests for WireClient.update_goal_state()`

azurelinuxagent/common/protocol/wire.py
```diff
@@ -761,78 +759,88 @@ def update_ext_conf(self, goal_state):
...
-                self.update_certs(current_goal_state_from_configuration)
-                self.update_ext_conf(current_goal_state_from_configuration)
-                self.update_remote_access_conf(current_goal_state_from_configuration)
-                self.save_cache(incarnation_file, current_goal_state_from_configuration.incarnation)
-                self.update_host_plugin(current_goal_state_from_configuration.container_id,
-                                        current_goal_state_from_configuration.role_config_name)
+                new_goal_state_xml = self.fetch_config(uri, self.get_header())
+                new_goal_state = GoalState(new_goal_state_xml)
+
+                def update_host_plugin():
```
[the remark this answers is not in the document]
narrieta: They capture state and cannot be moved unless they are parameterized...adding parameters would make them more complex that they need to be (they are just simple abbreviations for a few lines of code)

azurelinuxagent/common/protocol/wire.py
```diff
@@ -761,78 +759,88 @@ def update_ext_conf(self, goal_state):
...
-                else:
-                    logger.error("ProtocolError processing goal state, giving up [{0}]", ustr(e))
+                logger.verbose("ProtocolError processing goal state (attempt {0}/{1}) [{2}]", retry, max_retry, ustr(e))
 
             except Exception as e:
-                if retry < max_retry - 1:
-                    logger.verbose("Exception processing goal state, retrying: [{0}]", ustr(e))
-                else:
-                    logger.error("Exception processing goal state, giving up: [{0}]", ustr(e))
+                logger.verbose("Exception processing goal state (attempt {0}/{1}) [{2}]", retry, max_retry, ustr(e))
```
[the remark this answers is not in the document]
narrieta: I decided not to change the current level (verbose) since they likely would fill up the log

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1424,24 +1421,12 @@ def parse(self, xml_text):
         self.role_name = getattrib(role, "name")
         deployment = find(xml_doc, "Deployment")
         self.deployment_name = getattrib(deployment, "name")
-        return self
 
 
 class SharedConfig(object):
-    """
-    parse role endpoint server and goal state config.
```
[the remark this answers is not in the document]
narrieta: Except that the comment is not actually telling you what these objects do... they neither parse the SharedConfig nor write to SharedConfig.xml. 

I decided to keep the verbose "Load foo.xml" messages in all these classes, since they do not load the xml files either, but I was very tempted to remove those as well.

tests/protocol/mockwiredata.py
narrieta: Today, tests do simple text substitution, without adding any context for the substitution. I decided to use regexes to provide some context for the text replacement.

The new setters I created provide an abstraction over these mechanics and the implementation could be changed in the future if we find that we need full xml parsing via one of the standard libraries.

tests/protocol/test_wire.py
narrieta: thanks

azurelinuxagent/common/protocol/wire.py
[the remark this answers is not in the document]
narrieta: BTW - I also want to remove the retry logic here... but that would also be on a new iteration

azurelinuxagent/common/protocol/wire.py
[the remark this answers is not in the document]
vrdmr: Got it. Thanks.

PR #1743, opened 2020-01-06
Make protocol util a singleton per thread

larohra:
## Description
Recently we discovered a bug in the Guest Agent where the WireClient object different for different handlers which was due to the fact that a new ProtocolUtil object was being instantiated for every Hanlder (UpdateHandler, ExtHandler, etc). 
This PR addresses this issue by making sure that a single ProtocolUtil object is used for each thread. It was made a SingletonPerThread to avoid the concurrency issues that we face if we make it a generic singleton (and also because the majority of the processing is done in the ExtHandler thread and not in the other threads)

Issue # 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1743)


azurelinuxagent/common/protocol/util.py
```diff
@@ -63,16 +64,14 @@ def get_protocol_util():
...
     """
     ProtocolUtil handles initialization for protocol instance. 2 protocol types
     are invoked, wire protocol and metadata protocols.
     """
 
     def __init__(self):
-        self.lock = threading.Lock()
-        self.lock_wireserver_endpoint = threading.Lock()
-        self.protocol = None
+        self._protocol = None
```
larohra: Made the protocol object private to discourage people from using the property directly and always go through get_protocol()

azurelinuxagent/common/protocol/util.py
```diff
@@ -63,16 +64,14 @@ def get_protocol_util():
...
 
-class ProtocolUtil(object):
+class ProtocolUtil(SingletonPerThread):
     """
     ProtocolUtil handles initialization for protocol instance. 2 protocol types
     are invoked, wire protocol and metadata protocols.
     """
 
     def __init__(self):
-        self.lock = threading.Lock()
```
larohra: Removed both locks since this is a SingletonPerThread class so we dont need to worry about concurrency issues (Additionally the SingletonPerThread has its own locks to ensure concurrency)

azurelinuxagent/common/protocol/wire.py
```diff
@@ -677,19 +681,21 @@ def _fetch_response(self, uri, headers=None, use_proxy=None):
                 headers=headers,
                 use_proxy=use_proxy)
 
+            host_plugin = self.get_host_plugin()
```
larohra: Did this to encourage using the get function rather than using the property directly

azurelinuxagent/common/protocol/wire.py
```diff
@@ -109,6 +110,9 @@ def detect(self):
         # Set the initial goal state
         self.client.update_goal_state(forced=True)
```
vrdmr: Here - Do you want to use the new method you introduced? ` def update_goal_state(self, forced=False)`

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1297,7 +1303,7 @@ def get_artifacts_profile(self):
         artifacts_profile = None
 
         if self.has_artifacts_profile_blob():
-            blob = self.ext_conf.artifacts_profile_blob
+            blob = self.get_ext_conf().artifacts_profile_blob
```
vrdmr: Why not repeat the same pattern like above - `ext_conf = self.get_ext_conf()`

azurelinuxagent/ga/monitor.py
```diff
@@ -277,6 +277,8 @@ def daemon(self):
                         MonitorHandler.HOST_PLUGIN_HEARTBEAT_PERIOD,
                         MonitorHandler.IMDS_HEARTBEAT_PERIOD).seconds
         while self.should_run:
+            # Updating the goal_state periodically for the monitor thread to keep the WireClient object upto-date always
+            self.protocol.update_goal_state()
```
vrdmr: So this will update every 1m the `self.protocol` object of the monitor.py and that should not impact any protocol objects in the exthandler. Correct?

tests/ga/test_monitor.py
```diff
@@ -342,18 +344,19 @@ def test_heartbeats(self,
         self.assertEqual(0, patch_send_telemetry_metrics.call_count)
         self.assertEqual(0, patch_poll_telemetry_metrics.call_count)
 
-        monitor_handler.start()
-        time.sleep(1)
-        self.assertTrue(monitor_handler.is_alive())
+        with patch.object(monitor_handler, 'protocol'):
```
vrdmr: What is this patching? Are you forcefully returning a mock when looking at self.protocol?

vrdmr: Or you intend to test it with protocol to be None

azurelinuxagent/common/protocol/wire.py
larohra: The method `update_goal_state` I introduced is calling the `client.update_goal_state` anyways. The only reason I added that function is because monitor thread has access to the protocol object and MetadataProtocol.py doesn't have a client object in it. So to ensure that the code wont break for azure stack I added the function in wireProtocol class

azurelinuxagent/common/protocol/wire.py
larohra: In the case above host_plugin was being used in multiple places. In this case this was the only occurrence of it so directly called it from here

azurelinuxagent/ga/monitor.py
larohra: Yes that's correct, we would have 3 different protocol objects for the 3 threads that we own. The GoalState is not being used anywhere in the environment thread so I didnt add an `update_goal_state` call in that thread. The monitor thread however needs certain data (host_plugin details and incarnation) to function properly due to which I added this update call here

tests/ga/test_monitor.py
larohra: Yes, I'm forcefully returning a mock for the monitor_thread.protocol object. This was not required before my changes because we weren't calling the `protocol.update_goal_state()` before my changes. This change in behaviour was breaking the test due to which I had to mock it

azurelinuxagent/common/protocol/wire.py
```diff
@@ -109,6 +110,9 @@ def detect(self):
         # Set the initial goal state
         self.client.update_goal_state(forced=True)
 
+    def update_goal_state(self, forced=False):
```
narrieta: The update_goal_state method needs to be added to Protocol and Metadata as well. I am doing that on my PR

tests/common/test_SingletonPerThread.py
```diff
@@ -0,0 +1,144 @@
...
+        t1_object = check_obj(self.THREAD_NAME_1)
+        t2_object = check_obj(self.THREAD_NAME_2)
+
+        return t1_object, t2_object
+
+    def test_it_should_have_only_one_instance_for_same_thread(self):
+        obj1 = TestClassToTestSingletonPerThread()
+        obj2 = TestClassToTestSingletonPerThread()
+
+        self.assertEqual(obj1.uuid, obj2.uuid)
```
narrieta: minor: uuid is ok, but you could also just compare the references:

self.assertEqual(obj1, obj2)

tests/common/test_SingletonPerThread.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def test_it_should_have_multiple_instances_for_multiple_threads(self):
+        output = Queue()
```
narrieta: "instances" instead of "output"? ("output" seems a little misleading)

narrieta: just curious: why output is a local and errors an instance member?

tests/common/test_SingletonPerThread.py
```diff
@@ -0,0 +1,144 @@
...
+        self._setup_mutithread_and_execute(func1=self._get_test_class_instance,
+                                           args1=(output, self.errors),
+                                           func2=self._get_test_class_instance,
+                                           args2=(output, self.errors))
+
+        self.assertEqual(2, output.qsize())  # Assert that there are 2 objects in the queue
+        obj1, obj2 = output.get(), output.get()
+        self.assertNotEqual(obj1.uuid, obj2.uuid)
+
+    def test_it_should_remove_instance_for_calling_thread_on_clear(self):
```
narrieta: if we make clear a test function then we don't need this one

narrieta: nor get_and_clear_test_class_instance

tests/ga/test_exthandlers.py
```diff
@@ -1,21 +1,29 @@
...
-from azurelinuxagent.common.event import WALAEventOperation
-from azurelinuxagent.common.utils.extensionprocessutil import TELEMETRY_MESSAGE_MAX_LEN, format_stdout_stderr, read_output
 from tests.tools import AgentTestCase, patch, mock_sleep
 
 
 class TestExtHandlers(AgentTestCase):
+
+    def setUp(self):
+        super(TestExtHandlers, self).setUp()
+        ProtocolUtil.clear()
```
narrieta: let's add a comment explaining why this is needed?

tests/common/test_SingletonPerThread.py
```diff
@@ -0,0 +1,144 @@
...
+    THREAD_NAME_2 = 'thread-2'
+
+    def setUp(self):
+        super(TestSingletonPerThread, self).setUp()
+        # In a multi-threaded environment, exceptions thrown in the child thread will not be propagated to the parent
+        # thread. In order to achieve that, adding all exceptions to a Queue and then checking that in parent thread.
+        self.errors = Queue()
+        TestClassToTestSingletonPerThread.clear()
+
+    def _setup_mutithread_and_execute(self, func1, args1, func2, args2):
```
pgombar: nit: typo in `mutithread`

tests/protocol/test_protocol_util.py
```diff
@@ -17,18 +17,55 @@
...
+            try:
+                q.put(get_protocol_util())
+            except Exception as e:
+                err.put(str(e))
+
+        queue = Queue()
+        errors = Queue()
+        t1 = Thread(target=get_util_obj, args=(queue, errors))
+        t2 = Thread(target=get_util_obj, args=(queue, errors))
+        t1.start()
```
pgombar: nit: for extra checks, maybe consider giving the threads different names and asserting that the names match the respective singletons.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -804,9 +810,7 @@ def update_host_plugin():
                     return
 
                 def incarnation_changed():
-                    last_incarnation = None
-                    if os.path.isfile(incarnation_file):
-                        last_incarnation = fileutil.read_file(incarnation_file)
+                    last_incarnation = self.get_goal_state().incarnation
```
narrieta: should we do this until we remove the cache on disk? the current code always reads the file; get_goal_state may return the in-memory value

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1284,8 +1287,9 @@ def get_host_plugin(self):
         return self.host_plugin
 
     def has_artifacts_profile_blob(self):
-        return self.ext_conf and not \
-            textutil.is_str_none_or_whitespace(self.ext_conf.artifacts_profile_blob)
+        ext_conf = self.get_ext_conf()
```
narrieta: postpone until we remove the file cache?

azurelinuxagent/common/protocol/wire.py
narrieta: postpone?

azurelinuxagent/ga/monitor.py
narrieta: my bad... different loop... i'm not touching this on my PR, you are good

azurelinuxagent/common/protocol/wire.py
narrieta: I added update_goal_state in Wire/Metadata/Protocol without the forced flag. None of the calls to update_goal_state via protocol instances us this parameter.

The update_goal_state in WireClient does have the flag, but it is used only via the client (not the protocol)

azurelinuxagent/common/protocol/wire.py
larohra: I think we should keep this change on how we check for last_incarnation because without this the objects would again go out of sync as soon as some other thread updates the goal_state.
I can change the get_goal_state() to self.goal_state until the file cache is removed though.

azurelinuxagent/common/protocol/wire.py
larohra: Done

azurelinuxagent/common/protocol/wire.py
larohra: Done

azurelinuxagent/common/protocol/wire.py
larohra: I'll remove the flag in my PR too to avoid any merge conflicts later

tests/ga/test_monitor.py
```diff
@@ -118,7 +118,9 @@ def setUp(self):
         AgentTestCase.setUp(self)
         prefix = "UnitTest"
         logger.DEFAULT_LOGGER = Logger(prefix=prefix)
-        ProtocolUtil.clear()
+        # Since ProtocolUtil is not a singleton per thread, we need to clear it to ensure that the test cases are
```
pgombar: nit: extra word? "test cases (are) do not ..."

tests/ga/test_remoteaccess_handler.py
```diff
@@ -54,7 +53,9 @@ class TestRemoteAccessHandler(AgentTestCase):
 
     def setUp(self):
         super(TestRemoteAccessHandler, self).setUp()
-        ProtocolUtil.clear()
+        # Since ProtocolUtil is not a singleton per thread, we need to clear it to ensure that the test cases are
```
pgombar: Same comment.

tests/tools.py
```diff
@@ -543,3 +544,10 @@ def wrapper(self, *args, **kwargs):
     return decorator
 
 
+def clear_singleton_instances(cls):
```
pgombar: Can you add a comment explaining why this was introduced? For example, "some tests like XY use singleton instances of ProtocolUtil because ... ", just so it's easier to follow when reading the code.

tests/ga/test_monitor.py
larohra: Thanks for catching it! Removed

tests/ga/test_remoteaccess_handler.py
larohra: Removed

tests/tools.py
larohra: I feel adding more comments here explaining what the function does might make it extra verbose given the fact that I've already added comments everywhere ProtocolUtil was cleared. (Additionally I think the function name is pretty self-explanatory). I'll skip adding more comments for now but we can always add it later if we feel the need for it :)

pgombar: Fair enough :)

tests/common/test_SingletonPerThread.py
narrieta: ok; maybe worth a comment in the code, since this would be unexpected to the uninitiated

larohra: Makes sense, I'll add a comment

azurelinuxagent/common/protocol/wire.py
```diff
@@ -796,9 +799,7 @@ def update_host_plugin():
                     return
 
                 def incarnation_changed():
-                    last_incarnation = None
-                    if os.path.isfile(incarnation_file):
-                        last_incarnation = fileutil.read_file(incarnation_file)
+                    last_incarnation = self.goal_state.incarnation
```
narrieta: sorry, i missed this one before. this should also come in the PR that removes the files

larohra: Reverted

PR #1744, opened 2020-01-08
Do a single update_goal_state() in each iteration of the main loop

narrieta:
Several methods in WireClient were triggering an update_goal_state. 

The implementation of WireClient was inconsistent in this respect, since some methods would update and others would assume the goal state was up-to-date. This inconsistency was captured in a TODO comment in wire.py

Of more concern, if the incarnation changes in-between the calls to those functions then we may end up  processing an inconsistent goal state.

This PR moves the call to update_goal_state to the main loop.

There is also some cleanup of the Protocol interface and the Metadata implementation. I will add comments in the PR pointing those out.

I am still running DCR. I will sync with @larohra to figure out how we will merge this PR and #1743

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1744)

### DIFF
--- a/azurelinuxagent/common/protocol/metadata.py
+++ b/azurelinuxagent/common/protocol/metadata.py
@@ -170,7 +170,14 @@ class MetadataProtocol(Protocol):
                                 "{0}.crt".format(thumbprint))
         shutil.copyfile(trans_prv_file, prv_file)
         shutil.copyfile(trans_cert_file, crt_file)
-        self.update_goal_state()
+        self._update_certs_with_retry()
+
+    def update_goal_state(self):
+        # Metadata doesn't cache the goal state; nothing to do here...
+        pass
+
+    def get_endpoint(self):
+        return self.endpoint
 
     def get_vminfo(self):
         vminfo = VMInfo()
@@ -213,7 +220,7 @@ class MetadataProtocol(Protocol):
         return 0
 
     def get_vmagent_manifests(self):
-        self.update_goal_state()
+        self._update_certs_with_retry()
 
         headers = None if self.agent_etag is None else {IF_NONE_MATCH_HEADER: self.agent_etag}
 
@@ -266,7 +273,7 @@ class MetadataProtocol(Protocol):
         return vmagent_pkgs
 
     def get_ext_handlers(self, last_etag=None):
-        self.update_goal_state()
+        self._update_certs_with_retry()
         headers = {
             "x-ms-vmagent-public-x509-cert": self._get_trans_cert()
         }
@@ -330,7 +337,7 @@ class MetadataProtocol(Protocol):
         certificates = self.get_certs()
         return certificates.cert_list
 
-    def update_goal_state(self, max_retry=3):
+    def _update_certs_with_retry(self, max_retry=3):
         # Start updating goalstate, retry on 410
         for retry in range(0, max_retry):
             try:
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -81,20 +81,10 @@ class UploadError(HttpError):
 class WireProtocol(Protocol):
     """Slim layer to adapt wire protocol data to metadata protocol interface"""
 
-    # TODO: Clean-up goal state processing
-    #  At present, some methods magically update GoalState (e.g.,
-    #  get_vmagent_manifests), others (e.g., get_vmagent_pkgs)
-    #  assume its presence. A better approach would make an explicit update
-    #  call that returns the incarnation number and
-    #  establishes that number the "context" for all other calls (either by
-    #  updating the internal state of the protocol or
-    #  by having callers pass the incarnation number to the method).
-
     def __init__(self, endpoint):
         if endpoint is None:
             raise ProtocolError("WireProtocol endpoint is None")
-        self.endpoint = endpoint
-        self.client = WireClient(self.endpoint)
+        self.client = WireClient(endpoint)
 
     def detect(self):
         self.client.check_wire_protocol_version()
@@ -108,6 +98,12 @@ class WireProtocol(Protocol):
 
         # Set the initial goal state
         self.client.update_goal_state(forced=True)
+
+    def update_goal_state(self):
+        self.client.update_goal_state()
+
+    def get_endpoint(self):
+        return self.client.endpoint
 
     def get_vminfo(self):
         goal_state = self.client.get_goal_state()
@@ -133,8 +129,6 @@ class WireProtocol(Protocol):
             return 0
 
     def get_vmagent_manifests(self):
-        # Update goal state to get latest extensions config
-        self.client.update_goal_state()
         goal_state = self.client.get_goal_state()
         ext_conf = self.client.get_ext_conf()
         return ext_conf.vmagent_manifests, goal_state.incarnation
@@ -147,8 +141,6 @@ class WireProtocol(Protocol):
 
     def get_ext_handlers(self):
         logger.verbose("Get extension handler config")
-        # Update goal state to get latest extensions config
-        self.client.update_goal_state()
         goal_state = self.client.get_goal_state()
         ext_conf = self.client.get_ext_conf()
         # In wire protocol, incarnation is equivalent to ETag
@@ -164,7 +156,7 @@ class WireProtocol(Protocol):
         logger.verbose("Get In-VM Artifacts Profile")
         return self.client.get_artifacts_profile()
 
-    def download_ext_handler_pkg_through_host(self, uri, destination):
+    def _download_ext_handler_pkg_through_host(self, uri, destination):
         host = self.client.get_host_plugin()
         uri, headers = host.get_artifact_request(uri, host.manifest_uri)
         success = self.client.stream(uri, destination, headers=headers, use_proxy=False)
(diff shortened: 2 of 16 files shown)


azurelinuxagent/common/protocol/metadata.py
```diff
@@ -170,7 +170,14 @@ def detect(self):
                                 "{0}.crt".format(thumbprint))
         shutil.copyfile(trans_prv_file, prv_file)
         shutil.copyfile(trans_cert_file, crt_file)
-        self.update_goal_state()
+        self._update_certs_with_retry()
```
narrieta: The main call to update_goal_state is now in the main loop, which is protocol-independent, so I added a new method update_goal_state to the Protocol interface.

Metadata already had an update_goal_state, which just updates the certificates; I renamed that implementation to _update_certs_with_retry.

azurelinuxagent/common/protocol/metadata.py
```diff
@@ -170,7 +170,14 @@ def detect(self):
...
+
+    def update_goal_state(self):
+        # Metadata doesn't cache the goal state; nothing to do here...
+        pass
+
+    def get_endpoint(self):
```
narrieta: The code was using a data member 'endpoint' on protocol objects, but the member did not exist in the Protocol interfaces. I added it as a getter.;

azurelinuxagent/common/protocol/wire.py
```diff
@@ -164,7 +156,7 @@ def get_artifacts_profile(self):
         logger.verbose("Get In-VM Artifacts Profile")
         return self.client.get_artifacts_profile()
 
-    def download_ext_handler_pkg_through_host(self, uri, destination):
+    def _download_ext_handler_pkg_through_host(self, uri, destination):
```
narrieta: made this one private for clarity, since it is not part of the Protocol interface

azurelinuxagent/ga/exthandlers.py
```diff
@@ -206,14 +205,13 @@ class ExtHandlerState(object):
     FailedUpgrade = "FailedUpgrade"
 
 
-def get_exthandlers_handler():
-    return ExtHandlersHandler()
+def get_exthandlers_handler(protocol):
```
narrieta: Now the protocol is instantiated in the main loop and passed to the extension handler as parameter

azurelinuxagent/ga/exthandlers.py
```diff
@@ -276,7 +273,7 @@ def run(self):
                       message=detailed_msg)
             return
 
-    def cleanup_outdated_handlers(self):
+    def _cleanup_outdated_handlers(self):
```
narrieta: Made those private because i was refactoring these methods; I abandoned the refactoring, but these should stay private.

tests/ga/test_extension.py
```diff
@@ -376,185 +374,193 @@ def _set_up_update_test_and_update_gs(self, patch_command, *args):
         self._assert_ext_status(protocol.report_ext_status, "success", 0)
 
         # Next incarnation, update version
-        test_data.goal_state = test_data.goal_state.replace("<Incarnation>1<", "<Incarnation>2<")
-        test_data.ext_conf = test_data.ext_conf.replace('version="1.0.0"', 'version="1.0.1"')
-        test_data.manifest = test_data.manifest.replace('1.0.0', '1.0.1')
+        test_data.set_incarnation(2)
+        test_data.set_extensions_config_version("1.0.1")
+        test_data.set_manifest_version('1.0.1')
+        protocol.update_goal_state()
```
narrieta: now the call to update_goal_state is done in the main loop, instead of in the extension handler, so these tests need an explicit call to update_goal_state

tests/ga/test_extension.py
```diff
@@ -376,185 +374,193 @@ def _set_up_update_test_and_update_gs(self, patch_command, *args):
         self._assert_ext_status(protocol.report_ext_status, "success", 0)
 
         # Next incarnation, update version
-        test_data.goal_state = test_data.goal_state.replace("<Incarnation>1<", "<Incarnation>2<")
-        test_data.ext_conf = test_data.ext_conf.replace('version="1.0.0"', 'version="1.0.1"')
-        test_data.manifest = test_data.manifest.replace('1.0.0', '1.0.1')
+        test_data.set_incarnation(2)
```
narrieta: Since i was touching those tests, i updated them to used the utility function i created a few PRs ago. This improves the readability (I may still rename some of those utilies)

tests/ga/test_extension.py
```diff
@@ -601,16 +608,15 @@ def test_ext_handler_sequencing(self, *args):
...
-        test_data.ext_conf = test_data.ext_conf.replace("seqNo=\"0\"",
-                                                        "seqNo=\"1\"")
+        test_data.set_incarnation(2)
+        test_data.set_extensions_config_sequence_number(1)
         # Swap the dependency ordering
-        test_data.ext_conf = test_data.ext_conf.replace("dependencyLevel=\"2\"",
-                                                        "dependencyLevel=\"3\"")
-        test_data.ext_conf = test_data.ext_conf.replace("dependencyLevel=\"1\"",
-                                                        "dependencyLevel=\"4\"")
+        test_data.ext_conf = test_data.ext_conf.replace("dependencyLevel=\"2\"", "dependencyLevel=\"3\"")
```
narrieta: I didn't create utility functions for the dependency level, since it requires more sophisticated parsing of the xml... maybe next time

narrieta:
DCR run OK

azurelinuxagent/ga/update.py
```diff
@@ -297,11 +299,16 @@ def run(self, debug=False):
...
+                # Check that the parent process (the agent's daemon) is still running
+                #
                 if not debug and self._is_orphaned:
-                    logger.info("Agent {0} is an orphan -- exiting",
-                                CURRENT_AGENT)
+                    logger.info("Agent {0} is an orphan -- exiting", CURRENT_AGENT)
                     break
 
+                #
+                # Check that all thread are still running
```
pgombar: nit: threads

azurelinuxagent/common/protocol/wire.py
```diff
@@ -109,6 +99,12 @@ def detect(self):
         # Set the initial goal state
         self.client.update_goal_state(forced=True)
```
pgombar: Just checking - are we keeping this call to update goal state, even though it's not in the main loop, because it's called during set up of various threads?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1094,7 +1086,6 @@ def send_request_using_appropriate_channel(self, direct_func, host_func):
         return ret
 
     def upload_status_blob(self):
-        self.update_goal_state()
         ext_conf = self.get_ext_conf()
 
         if ext_conf.status_upload_blob is None:
```
pgombar: (github won't let me expand and comment on L1093), but why are we keeping that call to update the goal state? Because it's forced and otherwise we don't force it from the main loop? Do we even need a force update here?

tests/ga/test_extension.py
pgombar: Thanks for this!

tests/ga/test_extension.py
pgombar: We can't have it all...

azurelinuxagent/common/protocol/wire.py
narrieta: Yes, we are keeping for the moment since there may be code relying on the files this creates. As we continue to cleanup the code we may remove it if no longer needed.

azurelinuxagent/common/protocol/wire.py
narrieta: The status blob is in extensions config, so a force update would be needed if this condition was true.

As far as the if statement itself, I don't see why it is needed and I may remove it as we continue cleaning the code.

azurelinuxagent/ga/update.py
narrieta: thanks, fixed

narrieta:
Thanks, @larohra 

>  consider adding a new test case to ensure that we don't call update_goal_state() from multiple places in the main loop (essentially to assert the changes that you did in this PR).

There may still be multiple calls within 1 iteration of the loop (upload status still can force an update).

I want to enforce the single-call in code, though. On another pass of refactoring I want to detach the goal state from the protocol and then pass the goal state to the extension handler. One of the benefits of doing it is that then the extension handler would not be able to update the goal state.

larohra:
> On another pass of refactoring I want to detach the goal state from the protocol and then pass the goal state to the extension handler.

@narrieta I like the idea of passing a goalstate to ext_handler, it'll decouple the models and remove unnecessary dependencies. Looking forward to that PR

tests/test_import.py
```diff
@@ -1,3 +1,5 @@
+from mock import Mock
+
 from tests.tools import AgentTestCase
```
vrdmr: nit; Other tests are importing Mock from test.tools. Can you follow that pattern?
`from tests.tools import AgentTestCase, Mock`

narrieta: @vrdmr Thanks for pointing that out, I had not noticed it.

But, should we actually be importing from tests.tools in the other tests? Does tests.tools have its own Mock that provides extra value on top of mock.Mock? If not, I think we should actually be importing from mock in all tests.

PR #1749, opened 2020-01-09
Share protocol instance between main loop and remote access

narrieta:
Just a simple refactoring: passing the protocol instance from the main loop to remote access.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1749)


tests/test_import.py
```diff
@@ -1,31 +0,0 @@
...
-import azurelinuxagent.daemon as daemon
-import azurelinuxagent.daemon.resourcedisk as resourcedisk
-import azurelinuxagent.daemon.scvmm as scvmm
-import azurelinuxagent.ga.exthandlers as exthandlers
-import azurelinuxagent.ga.monitor as monitor
-import azurelinuxagent.ga.remoteaccess as remoteaccess
-import azurelinuxagent.ga.update as update
-
-
-class TestImportHandler(AgentTestCase):
```
larohra: Just curious what was this file for?

narrieta: Beats me :) I see no use for it and needs to be update when changing those methods. Since it provides no useful tests I decided to delete it instead of keeping it up to date.

narrieta: Beats me :) I see no use for it and needs to be update when changing those methods. Since it provides no useful tests I decided to delete it instead of keeping it up to date

PR #1754, opened 2020-01-11
Remove file dependency

larohra:
## Description
Currently the Guest Agent relies on the filesystem to fetch cached data for the GoalState if the object is None. This behavior can lead to inconsistencies in the wire client objects as some thread might update the filesystems midway of execution of goal-state.

This PR removes the dependencies of the WireClient from the file system and keeps everything in memory. We still write to files because they help in debugging. 

Issue # 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1754)


azurelinuxagent/common/protocol/util.py
```diff
@@ -175,7 +175,7 @@ def get_wireserver_endpoint(self):
                     logger.info("WireServer endpoint {0} read from file", self.endpoint)
                     return self.endpoint
 
-                logger.error("[GetWireserverEndpoint] Unexpected empty file {0}: {1}", file_path, str(e))
```
larohra: There was no error being caught here, so `e` was undefined. This faulty error message would have thrown an actual exception if this condition were true.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -528,17 +524,20 @@ class WireClient(object):
...
         logger.info("Wire server endpoint:{0}", endpoint)
-        self.endpoint = endpoint
-        self.goal_state = None
-        self.hosting_env = None
-        self.shared_conf = None
-        self.remote_access = None
-        self.certs = None
-        self.ext_conf = None
-        self.host_plugin = None
+        self._endpoint = endpoint
```
larohra: Made almost all WireClient properties private as they shouldn't be directly accessed. They should only be accessed by getter/setters

azurelinuxagent/common/protocol/wire.py
```diff
@@ -717,42 +717,33 @@ def update_certs(self, goal_state):
...
-        incarnation = self.fetch_cache(incarnation_file)
-        file_name = REMOTE_ACCESS_FILE_NAME.format(incarnation)
-        remote_access_file = os.path.join(conf.get_lib_dir(), file_name)
-        if not os.path.isfile(remote_access_file):
-            # no remote access data.
-            return None
-        xml_text = self.fetch_cache(remote_access_file)
-        remote_access = RemoteAccess(xml_text)
-        return remote_access
+        return self._remote_access
```
larohra: Original implementation was returning None if file had nothing, so just returning None too instead of error

azurelinuxagent/common/protocol/wire.py
```diff
@@ -790,35 +781,35 @@ def _update_from_goal_state(self, refresh_type):
...
 
                 if refresh_type == WireClient._UpdateType.HostPlugin:
                     update_host_plugin()
                     return
 
                 def incarnation_changed():
                     last_incarnation = None
-                    if os.path.isfile(incarnation_file):
-                        last_incarnation = fileutil.read_file(incarnation_file)
+                    if self._goal_state is not None:
```
larohra: Previously we checked the last_incarnation by checking the file values, that made different WireClient objects out of sync. Checking the incarnation value of the current object to ensure that all objects in all threads are always in sync with the latest goal state

azurelinuxagent/common/protocol/wire.py
```diff
@@ -837,70 +828,59 @@ def save_goal_state(incarnation, xml_text):
...
+        if new_remote_access is not None:
+            self._remote_access = new_remote_access
+
+    def set_host_plugin(self, new_host_plugin):
+        if new_host_plugin is not None:
+            self._host_plugin = new_host_plugin
+
+    def get_goal_state(self):
+        if self._goal_state is None:
+            raise ProtocolError("Trying to fetch goal state before initialization!")
```
larohra: In order to keep the behavior as consistent as possible, I'm throwing `ProtocolError` in places where the object is not initialized. We would've seen similar behavior earlier as the `fetch_cache` function would've thrown the same error incase the file was not found/faulty. 

Note: According to the current code the `GoalState` object would always be initialized before any of the `gets` are called but raising a `ProtocolError` ensures that the agent process doesn't die unexpectedly as these type of errors are handled gracefully in our code.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -837,70 +828,59 @@ def save_goal_state(incarnation, xml_text):
...
-                                                                      EXT_CONF_FILE_NAME.format(incarnation))))
-            handler_list = ext_conf.ext_handlers.extHandlers
-        except ProtocolError as pe:
-            # cache file is missing, nothing to do
-            logger.verbose(ustr(pe))
-        except Exception as e:
-            logger.error("Could not obtain current handlers: {0}", ustr(e))
-
-        return handler_list
+        return self._certs
```
larohra: Original implementation was returning None if file had nothing, so just returning None too instead of error

azurelinuxagent/common/protocol/wire.py
```diff
@@ -837,70 +828,59 @@ def save_goal_state(incarnation, xml_text):
...
 
     def get_ext_conf(self):
-        if self.ext_conf is None:
-            local_goal_state = self.get_goal_state()
-            if local_goal_state.ext_uri is None:
-                self.ext_conf = ExtensionsConfig(None)
+        if self._ext_conf is None:
```
larohra: Keeping behavior consistent with the older implementation

tests/protocol/test_wire.py
```diff
@@ -61,6 +60,16 @@ def get_event(message, duration=30000, evt_type="", is_internal=False, is_succes
     return event
 
 
+def setup_wireclient_ext_conf(wire_protocol_client, artifacts_profile_blob=None,
```
larohra: Helper function to create and set an Env_config object for the tests (made it global as its being used in multiple test classes)

tests/protocol/test_wire.py
```diff
@@ -596,24 +596,29 @@ def test_get_ext_conf_without_uri(self, *args):
...
-                   return_value=mockwiredata.WireProtocolData(mockwiredata.DATA_FILE).ext_conf):
-            ext_conf = wire_protocol_client.get_ext_conf()
-
-        self.assertEqual(1, len(ext_conf.ext_handlers.extHandlers))
-        self.assertEqual(2, len(ext_conf.vmagent_manifests.vmAgentManifests))
-        self.assertEqual("https://test.blob.core.windows.net/vhds/test-cs12.test-cs12.test-cs12.status?"
-                         "sr=b&sp=rw&se=9999-01-01&sk=key1&sv=2014-02-14&sig=hfRh7gzUE7sUtYwke78IOlZOrTRCYvkec4hGZ9zZzXo", ext_conf.status_upload_blob)
-        self.assertEqual("BlockBlob", ext_conf.status_upload_blob_type)
-        self.assertEqual(None, ext_conf.artifacts_profile_blob)
+        mock_data = mockwiredata.WireProtocolData(mockwiredata.DATA_FILE)
```
larohra: This test was relying deeply on the previous implementation (as it was mocking `fetch_cache` to get the ext_conf data. Since I was modifying this code, I modified the test logic to run the main function rather than the implementation. I'm a bit unsure as to what this test was trying to do, if anyone has any ideas let me know I'll modify the code accordingly

azurelinuxagent/common/protocol/wire.py
```diff
@@ -837,70 +828,59 @@ def save_goal_state(incarnation, xml_text):
 
         raise ProtocolError("Exceeded max retry updating goal state")
 
-    def get_goal_state(self):
-        if self.goal_state is None:
-            incarnation_file = os.path.join(conf.get_lib_dir(),
-                                            INCARNATION_FILE_NAME)
-            incarnation = self.fetch_cache(incarnation_file)
+    def set_goal_state(self, new_goal_state):
+        if new_goal_state is not None:
```
narrieta: why do we need the check for None? thanks

azurelinuxagent/common/protocol/wire.py
```diff
@@ -837,70 +828,59 @@ def save_goal_state(incarnation, xml_text):
...
+            if self.get_goal_state().ext_uri is None:
+                self.set_ext_conf(ExtensionsConfig(None))
             else:
-                local_file = EXT_CONF_FILE_NAME.format(local_goal_state.incarnation)
-                local_file = os.path.join(conf.get_lib_dir(), local_file)
-                xml_text = self.fetch_cache(local_file)
-                self.ext_conf = ExtensionsConfig(xml_text)
-        return self.ext_conf
+                raise ProtocolError("Trying to fetch Extension Conf before initialization!")
+        return self._ext_conf
```
narrieta: get_ext_manifest is still fetching from the local file

azurelinuxagent/common/rdma.py
```diff
@@ -29,19 +29,16 @@
...
     '/usr/local/etc/dat.conf'
 ]
 
-def setup_rdma_device(nd_version):
+
+def setup_rdma_device(nd_version, shared_conf):
     logger.verbose("Parsing SharedConfig XML contents for RDMA details")
-    xml_doc = parse_doc(
-        fileutil.read_file(os.path.join(conf.get_lib_dir(), SHARED_CONF_FILE_NAME)))
+    xml_doc = parse_doc(shared_conf.xml_text)
```
narrieta: I did not notice this parsing code before... can we move it to wire.py (SharedConfig)

azurelinuxagent/ga/monitor.py
```diff
@@ -153,6 +153,8 @@ def stop(self):
 
     def init_protocols(self):
         self.protocol = self.protocol_util.get_protocol()
+        # Update the GoalState first time to instantiate all required objects for the monitor thread
+        self.protocol.update_goal_state()
```
narrieta: we still have the issue that the different threads may update the files on disk... it would only affect debugging, though. I can fix that when I change the way we archive those files.

azurelinuxagent/common/protocol/wire.py
larohra: I think you meant the `AGENTS_MANIFEST_FILE_NAME`, I see that is being read from file and being used. I'll review that code and remove the dependency. Thanks for finding this out!

azurelinuxagent/common/rdma.py
larohra: Sure, I'll review it and move it there

azurelinuxagent/ga/monitor.py
larohra: Sounds good, thanks!

azurelinuxagent/common/protocol/wire.py
narrieta: you should probably raise; if for any reason one of those values is None then the goal state would be inconsistent: would have the previous values

azurelinuxagent/common/protocol/wire.py
narrieta: got it

azurelinuxagent/common/protocol/wire.py
narrieta: yes, the comment is for get_ext_manifest, I put here because there are no changes on that method and I can't add it there

azurelinuxagent/common/protocol/wire.py
larohra: Good point, I'll modify the code

azurelinuxagent/common/protocol/wire.py
```diff
@@ -133,7 +132,7 @@ def get_vmagent_manifests(self):
     def get_vmagent_pkgs(self, vmagent_manifest):
         goal_state = self.client.get_goal_state()
         ga_manifest = self.client.get_gafamily_manifest(vmagent_manifest, goal_state)
```
larohra: In the `get_gafamily_manifest` we were first removing the `AGENTS_MANIFEST_FILE_NAME` and then in the `filter_package_list` we were checking if the `AGENTS_MANIFEST_FILE_NAME` exists or not, and if not creating it and returning `ga_manifest.pkg_list`. 
This was unnecessarily just adding a file dependency even when it was not being used. Removed the whole code logic to avoid confusion and always return the `ga_manifest.pkg_list` we receive from GoalSate.

azurelinuxagent/common/protocol/wire.py
larohra: Raising error now to keep it consistent with other getters

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1401,6 +1350,40 @@ class SharedConfig(object):
     def __init__(self, xml_text):
         logger.verbose("Load SharedConfig.xml")
         self.xml_text = xml_text
+        self.rdma_ipv4_addr = None
+        self.rdma_mac_addr = None
+
+    def parse(self):
```
larohra: Made this function public instead of private because we only need to parse this on demand when setting up RDMA, which only happens once in the daemon handler. No need to parse it unnecessarily everytime in case the SharedConfig changes in any goalstate.

tests/ga/test_update.py
```diff
@@ -1426,64 +1426,6 @@ def test_write_pid_file_ignores_exceptions(self):
                 self.assertEqual(0, len(pid_files))
                 self.assertEqual(None, pid_file)
 
-    @patch('azurelinuxagent.common.protocol.wire.WireClient.get_goal_state',
-           return_value=GoalState(load_data('wire/goal_state.xml')))
-    def test_package_filter_for_agent_manifest(self, _):
```
larohra: Removed this test as it wasn't needed anymore.

azurelinuxagent/common/protocol/wire.py
narrieta: Thanks for the change. My intention was to achieve consistency with the rest of the code. I know this is done only once, and parsing later is not currently needed, but I do not think the time needed to parse it is significant. If perf is really a concern let's leave as is, otherwise we could have an inconsistent goal state.

larohra: Makes sense, I was debating with myself about the same thing and then in the end decided on a compromise to move the parsing logic to SharedConfig class but call it on demand. But I do agree to your point, I'll make it consistent with the remaining classes as parsing is not a perf intensive operation.

azurelinuxagent/common/protocol/wire.py
narrieta: i read that code several times and i've no idea what it was trying to do... either there was something very subtle there or it is total ----.  Your approach seems fine. Thanks!

azurelinuxagent/common/protocol/wire.py
```diff
@@ -837,70 +824,68 @@ def save_goal_state(incarnation, xml_text):
...
-                local_file = EXT_CONF_FILE_NAME.format(local_goal_state.incarnation)
-                local_file = os.path.join(conf.get_lib_dir(), local_file)
-                xml_text = self.fetch_cache(local_file)
-                self.ext_conf = ExtensionsConfig(xml_text)
-        return self.ext_conf
+                raise ProtocolError("Trying to fetch Extension Conf before initialization!")
+        return self._ext_conf
 
     def get_ext_manifest(self, ext_handler, goal_state):
         local_file = MANIFEST_FILE_NAME.format(ext_handler.name, goal_state.incarnation)
```
pgombar: Update to remove file dependency?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -790,35 +782,30 @@ def _update_from_goal_state(self, refresh_type):
                 new_goal_state = GoalState(new_goal_state_xml)
 
                 def update_host_plugin():
-                    if self.host_plugin is not None:
-                        self.host_plugin.container_id = new_goal_state.container_id
-                        self.host_plugin.role_config_name = new_goal_state.role_config_name
+                    if self._host_plugin is not None:
+                        self._host_plugin.update_container_id(new_goal_state.container_id)
+                        self._host_plugin.update_role_config_name(new_goal_state.role_config_name)
```
pgombar: Shouldn't you be using the getter for the host plugin here?

tests/protocol/test_wire.py
pgombar: nit: 'set up' is a verb needed in this scenario, 'setup' is a noun.

azurelinuxagent/common/protocol/wire.py
larohra: The getter for HostGAPlugin will create a new object if its not only present. If I call it here then the update call would technically create a new object which was not the behavior previously, I didnt want to diverge a lot from the old code as it might have some unprecedented circumstances. Mainly due to those reasons I just used the property here instead of calling the getter

azurelinuxagent/common/protocol/wire.py
vrdmr: +1

larohra: Sorry I didnt get what you meant. If you meant `MANIFEST_FILE_NAME` this file, then this is only being used to save the local cache, its never read from.
```python
    def get_ext_manifest(self, ext_handler, goal_state):
        local_file = MANIFEST_FILE_NAME.format(ext_handler.name, goal_state.incarnation)
        local_file = os.path.join(conf.get_lib_dir(), local_file)

        try:
            self.save_cache(local_file, xml_text)
            return ExtensionManifest(xml_text)
        except Exception as e:
            raise ExtensionDownloadError("Failed to retrieve extension manifest. Error: {0}".format(ustr(e)))
```

vrdmr: And this could apply for the `get_gafamily_manifest` case as well.

tests/protocol/test_wire.py
larohra: Thanks for catching this, changed the name

azurelinuxagent/ga/remoteaccess.py
```diff
@@ -62,7 +62,12 @@ def run(self):
                 if self.incarnation != current_incarnation:
                     # something changed. Handle remote access if any.
                     self.incarnation = current_incarnation
-                    self.remote_access = self.protocol.client.get_remote_access()
+                    try:
+                        self.remote_access = self.protocol.client.get_remote_access()
```
vrdmr: This is what we discussed in the meeting - that this code would be reverted. Correct?

azurelinuxagent/common/protocol/wire.py
pgombar: I don't think we could even be in the situation where `update_host_plugin()` is called without the `HostPlugin` having already been initialized. But fair point, it's better to minimize changes for this iteration and do the necessary testing and cleanup later.

azurelinuxagent/common/protocol/wire.py
pgombar: Ah sorry, the name of the method (implying a getter) and the fact that we're reading something from a file made me think this was retrieving state from a file.

azurelinuxagent/ga/remoteaccess.py
larohra: Nope not this code, that was the RDMA code. In this I added a log.warn incase there was a ProtocolError but reverted that change because this file in majorly empty

azurelinuxagent/common/protocol/wire.py
larohra: > And this could apply for the `get_gafamily_manifest` case as well.

Yes it would, both cases its the same, its just being used to save cache only

```python
    def get_gafamily_manifest(self, vmagent_manifest, goal_state):
        local_file = MANIFEST_FILE_NAME.format(vmagent_manifest.family, goal_state.incarnation)
        local_file = os.path.join(conf.get_lib_dir(), local_file)

        try:
            xml_text = self.fetch_manifest(vmagent_manifest.versionsManifestUris)
            fileutil.write_file(local_file, xml_text)
            return ExtensionManifest(xml_text)
        except Exception as e:
            raise ProtocolError("Failed to retrieve GAFamily manifest. Error: {0}".format(ustr(e)))
```

larohra:
Clean DCR run - https://tuxgold.corp.microsoft.com/job/AzLinux/job/DungeonCrawler/job/DungeonCrawler.larohra2/324/

azurelinuxagent/common/protocol/wire.py
```diff
@@ -717,42 +716,35 @@ def update_certs(self, goal_state):
...
-        file_name = REMOTE_ACCESS_FILE_NAME.format(incarnation)
-        remote_access_file = os.path.join(conf.get_lib_dir(), file_name)
-        if not os.path.isfile(remote_access_file):
-            # no remote access data.
-            return None
-        xml_text = self.fetch_cache(remote_access_file)
-        remote_access = RemoteAccess(xml_text)
-        return remote_access
+        if self._remote_access is None:
+            raise ProtocolError("Trying to fetch Remote Access before initialization!")
```
narrieta: seems like the previous code was returning None rather than raising

larohra: Yes, I was initially just setting it to None too but changed to raise a ProtocolError to keep it at par with the rest of the code. I am handling this now in the remoteaccess.py file
```python
                    try:
                        self.remote_access = self.protocol.client.get_remote_access()
                    except ProtocolError:
                        self.remote_access = None
```

PR #1755, opened 2020-01-14
Periodic HeartBeat to be logged to the file

vrdmr:
## Description
Adding Heartbeat logs to file. It logs every 12 hours currently, and helps in debugging.

Issue # 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1755)

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -273,8 +273,6 @@ class ExtHandlersHandler(object):
                       message=detailed_msg)
             return
 
-        logger.periodic_info(logger.EVERY_HALF_HOUR, "[PERIODIC] WALinuxAgent HeartBeat")
-
     def _cleanup_outdated_handlers(self):
         handlers = []
         pkgs = []
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -273,6 +273,8 @@ class ExtHandlersHandler(object):
                       message=detailed_msg)
             return
 
+        logger.periodic_info(logger.EVERY_HALF_HOUR, "[PERIODIC] WALinuxAgent HeartBeat")
+
     def _cleanup_outdated_handlers(self):
         handlers = []
         pkgs = []
--- a/azurelinuxagent/ga/monitor.py
+++ b/azurelinuxagent/ga/monitor.py
@@ -118,7 +118,6 @@ class MonitorHandler(object):
 
     def __init__(self):
         self.osutil = get_osutil()
-        self.protocol_util = get_protocol_util()
         self.imds_client = get_imds_client()
 
         self.event_thread = None
@@ -130,6 +129,7 @@ class MonitorHandler(object):
         self.last_host_plugin_heartbeat = None
         self.last_imds_heartbeat = None
         self.protocol = None
+        self.protocol_util = None
         self.health_service = None
         self.last_route_table_hash = b''
         self.last_nic_state = {}
@@ -141,9 +141,7 @@ class MonitorHandler(object):
         self.imds_errorstate = ErrorState(min_timedelta=MonitorHandler.IMDS_HEALTH_PERIOD)
 
     def run(self):
-        self.init_protocols()
-        self.init_sysinfo()
-        self.start()
+        self.start(init_data=True)
 
     def stop(self):
         self.should_run = False
@@ -151,14 +149,20 @@ class MonitorHandler(object):
             self.event_thread.join()
 
     def init_protocols(self):
+        # The initialization of ProtocolUtil for the Monitor thread should be done within the thread itself rather
+        # than initializing it in the ExtHandler thread. This is done to avoid any concurrency issues as each
+        # thread would now have its own ProtocolUtil object as per the SingletonPerThread model.
+        self.protocol_util = get_protocol_util()
         self.protocol = self.protocol_util.get_protocol()
+        # Update the GoalState first time to instantiate all required objects for the monitor thread
+        self.protocol.update_goal_state()
         self.health_service = HealthService(self.protocol.get_endpoint())
 
     def is_alive(self):
         return self.event_thread is not None and self.event_thread.is_alive()
 
-    def start(self):
-        self.event_thread = threading.Thread(target=self.daemon)
+    def start(self, init_data=False):
+        self.event_thread = threading.Thread(target=self.daemon, args=(init_data,))
         self.event_thread.setDaemon(True)
         self.event_thread.setName("MonitorHandler")
         self.event_thread.start()
@@ -268,7 +272,12 @@ class MonitorHandler(object):
 
             self.last_event_collection = datetime.datetime.utcnow()
 
-    def daemon(self):
+    def daemon(self, init_data=False):
+
+        if init_data:
+            self.init_protocols()
+            self.init_sysinfo()
+
         min_delta = min(MonitorHandler.TELEMETRY_HEARTBEAT_PERIOD,
                         MonitorHandler.CGROUP_TELEMETRY_POLLING_PERIOD,
                         MonitorHandler.CGROUP_TELEMETRY_REPORTING_PERIOD,
--- a/azurelinuxagent/ga/monitor.py
+++ b/azurelinuxagent/ga/monitor.py
@@ -426,11 +426,7 @@ class MonitorHandler(object):
 
                 add_event(name=AGENT_NAME, version=CURRENT_VERSION, op=WALAEventOperation.HeartBeat, is_success=True,
                           message=msg, log_event=False)
-
-                logger.periodic_info(logger.EVERY_HALF_DAY, "[PERIODIC] Incarnation: {0}; ContainerId: {2}",
-                                     incarnation, self.counter, get_container_id_from_env())
                 self.counter += 1
-
                 io_errors = IOErrorCounter.get_and_reset()
                 hostplugin_errors = io_errors.get("hostplugin")
                 protocol_errors = io_errors.get("protocol")
--- a/azurelinuxagent/ga/monitor.py
+++ b/azurelinuxagent/ga/monitor.py
@@ -134,7 +134,6 @@ class MonitorHandler(object):
         self.last_route_table_hash = b''
         self.last_nic_state = {}
 
-        self.counter = 0
         self.sysinfo = []
         self.should_run = True
         self.heartbeat_id = str(uuid.uuid4()).upper()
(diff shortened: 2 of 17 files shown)


azurelinuxagent/ga/monitor.py
```diff
@@ -424,14 +424,11 @@ def send_telemetry_heartbeat(self):
...
-                    name=AGENT_NAME,
-                    version=CURRENT_VERSION,
-                    op=WALAEventOperation.HeartBeat,
-                    is_success=True,
-                    message=msg,
-                    log_event=False)
+                add_event(name=AGENT_NAME, version=CURRENT_VERSION, op=WALAEventOperation.HeartBeat, is_success=True,
+                          message=msg, log_event=False)
 
+                logger.periodic_info(logger.EVERY_HALF_DAY, "[PERIODIC] Incarnation: {0}; ContainerId: {2}",
```
narrieta: Let's add more context to the message (what is it?)

Let's not include the container ID (I'm even  tempted to remove the incarnation from the heartbeat if we are not using it),  we are still working on those calls update_goal_state. If we do the heartbeat on the main thread (extension handling) then I do not have concerns about this.

narrieta: > I'm even tempted to remove the incarnation

since the heartbeat thread is separate from extension handling I think the incarnation here has no meaning since it may be out of sync with the incarnation that we are processing or processed last

vrdmr: Yup - I'll add more context.

>If we do the heartbeat on the main thread (extension handling) then I do not have concerns about this.

This sounds like a good idea. I'll add a logger in the ext-handler's run method. This would help us catch cases like the one faced by the AzureData team, where the ext-handler was down but the monitor-handler was sending messages.

> Let's not include the container ID (I'm even tempted to remove the incarnation from the heartbeat if we are not using it), 

I will keep incarnation to match it with WindowsGA, but other info can cause the periodic entry to break (Hash).
```
string message = string.Format("{0};{1};{2}",
    this.Incarnation,       // current goalstate incarantion
    this.HeartBeatCount++,  // current heart beat count 
    Utilities.ProcessId);
```

azurelinuxagent/ga/monitor.py
```diff
@@ -424,14 +424,11 @@ def send_telemetry_heartbeat(self):
...
 
-                add_event(
-                    name=AGENT_NAME,
-                    version=CURRENT_VERSION,
-                    op=WALAEventOperation.HeartBeat,
-                    is_success=True,
-                    message=msg,
-                    log_event=False)
+                add_event(name=AGENT_NAME, version=CURRENT_VERSION, op=WALAEventOperation.HeartBeat, is_success=True,
+                          message=msg, log_event=False)
```
larohra: Just a thought, why not just make the `log_event=True` for the `add_event` instead of logging it separately? We only invoke this call once every 30mins so we wouldn't have to worry about making it periodic either.

larohra: Ohh I see you wanted to log every half_day instead of 30mins. But after the discussion yesterday, you could just log the heartbeat within the add_event function directly.

azurelinuxagent/ga/monitor.py
larohra: +1 on the ProcessId, I like the idea that if something went wrong and the agent had to restart we would capture that in our logs/telemetry too.

azurelinuxagent/ga/monitor.py
```diff
@@ -424,14 +424,11 @@ def send_telemetry_heartbeat(self):
...
 
+                logger.periodic_info(logger.EVERY_HALF_DAY, "[PERIODIC] Incarnation: {0}; ContainerId: {2}",
+                                     incarnation, self.counter, get_container_id_from_env())
```
narrieta: what incarnation is this?

azurelinuxagent/ga/monitor.py
vrdmr: The Pid is now populated in the Kusto logs. I haven't added PID in the current HeartBeat, but the logs will tell if a restart happened.

azurelinuxagent/ga/monitor.py
vrdmr: I moved the logging to the extHandler now. Monitor Handler will still send the telemetry and extHandler (being the main thread), will log correctly.

Thus, the above change won't help.

Also on a side note - `log_event=True` has some weird behavior - It only logs when `is_success=False`, and thus, I didn't want to rely on that.

larohra: I feel we should keep both (telemetry and the log) in the same thread if its meant for the same purpose. I dont like the discrepancy between sending the same data from 2 different points.

Consider this scenario - 
Say ExtHandler was down (which we verify from logs) but we still see heartbeat telemetry in Kusto. The oncall engineer would think that the agent is working fine when in all reality it isint, it'll just confuse the oncall engineer.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -273,6 +273,8 @@ def run(self):
                       message=detailed_msg)
             return
 
+        logger.periodic_info(logger.EVERY_HALF_HOUR, "[PERIODIC] WALinuxAgent HeartBeat")
```
narrieta: I'd do this in the main loop (UpdateHandler.run); since that code already logs this

>             logger.info(u"Agent {0} is running as the goal state agent", CURRENT_AGENT)

we might as well just log the same (periodically) to indicate the agent is still running

azurelinuxagent/ga/monitor.py
narrieta: @larohra - we've already hit that scenario with the current code: we have hearbeat and no activity in the local log.

move the event here?

narrieta: after all, it is the main thread who reports status to CRP so the event/log may be more appropriate here.
the main thread also monitors the other 2 thread to check that they are active.

larohra: @narrieta I think we're talking about the same thing. I'm also advocating for moving the Telemetry event + log to the main thread rather than doing it in monitor thread. 

(We're commenting in the monitor thread right now)

azurelinuxagent/ga/exthandlers.py
vrdmr: This sounds like a better idea. I'll make the change.

azurelinuxagent/ga/monitor.py
vrdmr: @larohra - I agree with the scenario - and ideally, the heartbeat should come from the main extension processing thread only, but we can't overload the single process to do all the scenarios. This is, IMO, the next best thing we can get, to showcase if the main exthandler is running or not.

larohra: @vrdmr the monitor thread is the one who's going to actually send out the events to wireserver in the end, its just about who's going to add the heartbeat event file in the filesystem (which is not an expensive operation) so I dont think we're going to be overriding the main thread in any way. Additionally since we're going to be sending out heartbeats every 30mins its not going to be very expensive for the ExtHandler thread to write that file  :)

pgombar: @larohra can you clarify what you mean by "main" thread? The thread that starts other threads and checks if they're alive is the `update.py`, not `exthandlers.py`. You seem to be conflating the two.

I think the `update.py` thread would be appropriate for this kind of logging, since it's the most expensive thread to die and we should care if that one is reporting heartbeat or not.

narrieta: @larohra 

> @narrieta I think we're talking about the same thing. I'm also advocating for moving the Telemetry event + log to the main thread rather than doing it in monitor thread.

agreed

narrieta: also, if we do that I would rather check the time interval to log using the same logic as we do in the other threads rather than relying in the [PERIODIC] logging, since that can be reset (and has been reset in the past) by other components

larohra: >The thread that starts other threads and checks if they're alive is the update.py, not exthandlers.py

@pgombar they are the same threads, you might be getting confused with the name `UpdateHandler` and `ExtHandler`. Those are the classes that are running in the same `ExtHandler` thead :)

```python
    def run_exthandlers(self, debug=False):
        """
        Run the update and extension handler
        """
        logger.set_prefix("ExtHandler")
        threading.current_thread().setName("ExtHandler")
        from azurelinuxagent.ga.update import get_update_handler
        update_handler = get_update_handler()
        update_handler.run(debug)
```

pgombar: You're totally right, I just wasn't paying attention. Thanks for the correction!

azurelinuxagent/ga/update.py
```diff
@@ -345,6 +352,8 @@ def run(self, debug=False):
                         duration=duration,
                         message="Incarnation {0}".format(exthandlers_handler.last_etag))
 
+                self._send_heartbeat_telemetry(protocol)
```
narrieta: defined '_send_telemetry_heartbeat', but invoking '_send_heartbeat_telemetry'

azurelinuxagent/ga/update.py
```diff
@@ -730,6 +739,21 @@ def _write_pid_file(self):
...
+        if self.last_telemetry_heartbeat is None:
+            self.last_telemetry_heartbeat = datetime.utcnow() - UpdateHandler.TELEMETRY_HEARTBEAT_PERIOD
+
+        if datetime.utcnow() >= (self.last_telemetry_heartbeat + UpdateHandler.TELEMETRY_HEARTBEAT_PERIOD):
+            dropped_packets = self.osutil.get_firewall_dropped_packets(protocol.get_endpoint())
+            msg = "{0};{1};{2}".format(self.counter, self.heartbeat_id, dropped_packets)
+
+            add_event(name=AGENT_NAME, version=CURRENT_VERSION, op=WALAEventOperation.HeartBeat, is_success=True,
+                      message=msg, log_event=False)
+            self.counter += 1
```
narrieta: can we have a better name for this counter?

azurelinuxagent/ga/update.py
vrdmr: Thanks for catching this.

azurelinuxagent/ga/monitor.py
vrdmr: @larohra: Agreed. I liked the idea and thus moved the whole HB logic into the UpdateHandler.

azurelinuxagent/ga/update.py
narrieta: could you do a dcr run with these changes? thanks

azurelinuxagent/ga/update.py
```diff
@@ -730,6 +739,21 @@ def _write_pid_file(self):
...
+
+        if datetime.utcnow() >= (self.last_telemetry_heartbeat + UpdateHandler.TELEMETRY_HEARTBEAT_PERIOD):
+            dropped_packets = self.osutil.get_firewall_dropped_packets(protocol.get_endpoint())
+            msg = "{0};{1};{2}".format(self.heartbeat_counter, self.heartbeat_id, dropped_packets)
+
+            add_event(name=AGENT_NAME, version=CURRENT_VERSION, op=WALAEventOperation.HeartBeat, is_success=True,
+                      message=msg, log_event=False)
+            self.heartbeat_counter += 1
+
+            logger.info(u"Agent {0} is running as the goal state agent", CURRENT_AGENT)
```
pgombar: Suggestion: adding `[PERIODIC]` to this message would make it clearer that it's just a diagnostic message, not that the agent was just started (which I think might end up being an outsider's interpretation).

vrdmr: Good idea. But I'll add `[HEARTBEAT]` rather than the Periodic tag to signify that its heartbeat, because we do use Periodic for other things.

azurelinuxagent/ga/update.py
vrdmr: @narrieta - Automation OK.

PR #1756, opened 2020-01-14
Fix singleton per thread

larohra:
## Description
Currently the ProtocolUtil for the MonitorThread and the EnvThread is being initialized in the ExtHandlerThread (MainThread). Due to the SingletonPerThread approach, this would face a problem as all the different threads would still have the same ProtocolUtil object (and subsequently the same Protocol object) which would cause concurrency issues.

This PR addresses this issue and ensures that each thread has its own protocol util object.

Issue # 

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1756)


azurelinuxagent/ga/monitor.py
```diff
@@ -142,24 +142,23 @@ def __init__(self):
...
-        self.start()
+        self.start(init_data=True)
 
     def stop(self):
         self.should_run = False
         if self.is_alive():
             self.event_thread.join()
 
     def init_protocols(self):
+        self.protocol_util = get_protocol_util()
```
larohra: I wanted to remove this property altogether but didnt mainly because a lot of test cases are dependent on this. I can take this up as a separate PR once I get back from vacation

azurelinuxagent/ga/env.py
```diff
@@ -98,6 +98,7 @@ def monitor(self):
         If dhcp client process re-start has occurred, reset routes.
         Purge unnecessary files from disk cache.
         """
+        self.protocol_util = get_protocol_util()
```
larohra: I wanted to remove this property altogether but didnt mainly because a lot of test cases are dependent on this. I can take this up as a separate PR once I get back from vacation

azurelinuxagent/ga/monitor.py
```diff
@@ -269,7 +268,12 @@ def collect_and_send_events(self):
 
             self.last_event_collection = datetime.datetime.utcnow()
 
-    def daemon(self):
+    def daemon(self, init_data=False):
+
+        if init_data:
```
larohra: Now initializing the protocol as part of the new thread.
`init_sysinfo` uses certain properties from the protocol object due to which it was moved here too

larohra: So we were only doing it once in the earlier code, as part of `monitor_thread.run()` and in the main loop we were only calling `monitor_thread.start()`
```python
    def run(self):
        self.init_protocols()
        self.init_sysinfo()
        self.start()

            while self.running:
                #
                # Check that the parent process (the agent's daemon) is still running
                #
                if not debug and self._is_orphaned:
                    logger.info("Agent {0} is an orphan -- exiting", CURRENT_AGENT)
                    break

                #
                # Check that all the threads are still running
                #
                if not monitor_thread.is_alive():
                    logger.warn(u"Monitor thread died, restarting")
                    monitor_thread.start()
```

I guess they made it that way to not re-run the `init_sysinfo` and `init_protocol` to make sure some basic properties are not overwritten. With my changes, I did not want to modify that behavior so I added the flag `init_data` to ensure that we're only initializing the data once and not everytime the thread goes down.

If you want to discuss about this more we can do it offline

azurelinuxagent/ga/env.py
narrieta: let's add a comment of why this is initialized here

azurelinuxagent/ga/monitor.py
```diff
@@ -269,7 +268,12 @@ def collect_and_send_events(self):
 
             self.last_event_collection = datetime.datetime.utcnow()
 
-    def daemon(self):
+    def daemon(self, init_data=False):
```
narrieta: init_protocols instead of init_data? (data is initialized in the constructor)

larohra: I made it `init_data` because we are initializing 2 functions if this flag is true - `init_protocols` and `init_sysinfo`. Calling the flag just `init_protocols` can be a bit misleading.

Also technically this `daemon` function is kinda like the constructor for the Monitor Thread too I guess :)

narrieta: k

azurelinuxagent/ga/env.py
larohra: Done

azurelinuxagent/ga/monitor.py
pgombar: Got it, thanks!

azurelinuxagent/ga/monitor.py
```diff
@@ -130,6 +129,7 @@ def __init__(self):
         self.last_host_plugin_heartbeat = None
         self.last_imds_heartbeat = None
         self.protocol = None
+        self.protocol_util = None
```
vrdmr: Something for future - Should this be private? We dont need to expose this outside. (applies to other attributes as well)

PR #1757, opened 2020-01-15
Fix telemetry fields override for agent and extension events

pgombar:
## Description

This PR simplifies the code for creating telemetry events and when those fields are populated. With these changes, the following issues will be fixed:
1. We do not have control over extension events and how they are populated. Today, we had an issue where an extension would have the field `ContainerId`, but it would be empty, and we would send it like that. With this PR, we define "extension-specific" fields (see table below) that will be left unchanged, but the rest of the telemetry fields will be overridden with values from the agent.
2. For agent events, the `ContainerId` and `GAVersion` values were getting populated during event-creation time, but then overridden again during event-sending time. This is wrong because we want events to be a result of the environment they were created in (for more accurate debugging, continuity, etc.).
3. The `sysinfo` parameter group was defined in the `monitor` thread (which primary job is to report events, not create them). This was refactored into a separate singleton class.
4. Once this code is merged, all subsequent agent versions will have the telemetry fields populated at event-creation time, not event-sending time like before. This means older agents' events will be incomplete before sending. To patch such events, a method is called during agent startup, `update_old_events_on_disk`.

For better understanding, here is how the events are created and sent out with the changes in this PR:

_Lifecycle of an agent event:_
1. An agent event can be created through three methods - `add_event`, `add_log_event` and `add_metric`. Before the event is saved to disk, all necessary fields are added and the event is completed. These fields are agent-specific fields, common fields, and `sysinfo` fields. Refer to the bottom for a table of these fields.
2. The monitor thread reads the event from disk in `collect_and_send_events` and does no post-processing for agent events before sending it out.

_Lifecycle of an extension event:_
1. The extension drops the event in the events folder on disk. We cannot enforce the event schema before it is saved to disk, so we modify it when the event is read.
2. The monitor thread reads the event from disk in `collect_and_send_events` and first trims non-extension-specific fields.
3. Next, `finalize_event_fields` is called that first that adds common fields and `sysinfo` fields before the event is sent out. Refer to the bottom for a table of these fields.

Agent- or extension-specific field | Value
-- | --
Name | Name of the agent or extension
Version | Version of the agent or extension
Operation | Operation name
OperationSuccess | Flag for operation success
Message | Operation message
Duration | Duration of the operation

Common field | Value
-- | --
GAVersion | Version of the agent creating the event
ContainerId | Container id of the VM creating the event
OpcodeName | Used for storing the timestamp of event creation
EventTid | Current thread id
EventPid | Current process id
TaskName | Current thread name
KeywordName | Empty
ExtensionType | json or xml
IsInternal | False

Sysinfo field | Value
-- | --
OSVersion | Version of the OS
ExecutionMode | IAAS
RAM | Amount of RAM
Processors | Number of processors
VMName | Name of the VM (not present in Kusto)
TenantName | Tenant name
RoleName | Role name
RoleInstanceName | Role instance name
Location | Location of the VM (not present in Kusto)
SubscriptionId | Subscription id
ResourceGroupName | Resource group name
VMId | VM id
ImageOrigin | Image origin

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1757)


azurelinuxagent/common/event.py
```diff
@@ -406,38 +410,35 @@ def add_default_parameters_to_event(event_parameters, set_values_for_agent=True)
...
-
-        return parameters
+        DefaultParameter = namedtuple('DefaultParameter', ['name', 'value'])
+        default_params = [DefaultParameter('GAVersion', CURRENT_AGENT),
+                          DefaultParameter('ContainerId', get_container_id_from_env()),
+                          DefaultParameter('OpcodeName', datetime.utcnow().__str__() if is_agent_event else event_params_dict.get('OpcodeName', '')),
+                          DefaultParameter('EventTid', threading.current_thread().ident if is_agent_event else event_params_dict.get('EventTid', 0)),
+                          DefaultParameter('EventPid', os.getpid() if is_agent_event else event_params_dict.get('EventPid', 0)),
+                          DefaultParameter('TaskName', threading.current_thread().getName() if is_agent_event else event_params_dict.get('TaskName', '')),
+                          DefaultParameter('KeywordName', event_params_dict.get('KeywordName', ''))]
```
pgombar: For extension events, I decided to let them keep the existing value of these parameters, otherwise set them to empty values. I did this because I'm not sure if any extension publishers might rely on these fields for some sort of debugging (like we are using `OpcodeName` to tell us the event-creation time for agent events, for example).

Let me know if you think we should always override them. Currently we only always override `GAVersion` and `ContainerId`.

azurelinuxagent/common/event.py
```diff
@@ -406,38 +410,35 @@ def add_default_parameters_to_event(event_parameters, set_values_for_agent=True)
...
-
-        parameters = DataContractList(TelemetryEventParam)
-        for name, value in param_names.items():
-            parameters.append(TelemetryEventParam(name, value))
-
-        return parameters
+        DefaultParameter = namedtuple('DefaultParameter', ['name', 'value'])
+        default_params = [DefaultParameter('GAVersion', CURRENT_AGENT),
+                          DefaultParameter('ContainerId', get_container_id_from_env()),
+                          DefaultParameter('OpcodeName', datetime.utcnow().__str__() if is_agent_event else event_params_dict.get('OpcodeName', '')),
```
larohra: Instead of doing this could you use this - `datetime.utcnow().strftime(u'%Y-%m-%dT%H:%M:%S.%fZ')` to keep it constant with our logging.

azurelinuxagent/common/event.py
larohra: No I think we should let them keep what they populate (infact that is what we're advertising they do in the new telemetry pipeline too).

azurelinuxagent/common/telemetryevent.py
```diff
@@ -39,6 +39,15 @@ def __init__(self, eventId=None, providerId=None):
     def __contains__(self, param_name):
         return param_name in [param.name for param in self.parameters]
 
+    def is_extension_event(self):
+        # Events originating from the agent have "WALinuxAgent" as the Name parameter, or they don't have a Name
+        # parameter, in the case of log and metric events. So, in case the Name parameter exists and it is not
+        # "WALinuxAgent", it is an extension event.
+        for param in self.parameters:
+            if param.name == "Name":
```
larohra: Can there be any extension event that doesn't have any `Name` parameter at all either? (Not sure if its possible for them to generate such events either)

Just a thought, since we can control the events being generated by the agent, can't we add a flag like `agent_event=True` on every event generated by us and use that to recognize if its ours or not? We can remove that flag before sending out the event to wireserver to not pollute wireserver.

narrieta:
@pgombar - Thanks for describing the life cycle of events, it is useful. 

We have had several issues already when populating/overriding these telemetry fields (what we do now in add_default_parameters_to_event). I think in the past we've agreed that the current logic is too convoluted.

I think we should follow a different approach: populate the agent events correctly; we have full control over those. When picking up the events, use TelemetryEvent.is_extension_event and, if it is an extension event, respect these fields and override/add the rest:
> 

Parameter | Value
-- | --
Name | Should be the same as the extension's name.
Version | Should be the same as the extension's version.
Operation | Free form. Chose a value that is particular to your extension, and not already in use.
OperationSuccess | true or false
Message | Free form. Limit it to less than 3.9 kB
Duration | Duration of the operation in milliseconds. This is a int64.

(I just emailed you a link to the above documentation)

azurelinuxagent/common/event.py
```diff
@@ -271,37 +270,34 @@ def is_period_elapsed(self, delta, h):
...
+                  message="", log_event=True):
 
         if (not is_success) and log_event:
             _log_event(name, op, message, duration, is_success=is_success)
 
-        self._add_event(duration, evt_type, is_internal, is_success, message, name, op, version, event_id=1)
+        self._add_event(duration, is_success, message, name, op, version, event_id=1)
 
-    def _add_event(self, duration, evt_type, is_internal, is_success, message, name, op, version, event_id):
+    def _add_event(self, duration, is_success, message, name, op, version, event_id):
```
pgombar: I removed the `evt_type` and `is_internal` parameters because they always defaulted to `""` and `False`, respectively. Furthermore, these two values are for fields `ExtensionType` and `IsInternal` which are common to agent and extension events, so I had to refactor them out of this method and instead populate them in `EventLogger.add_common_parameters_to_event`.

azurelinuxagent/common/telemetryevent.py
pgombar: An extension event without the `Name` parameter already populated wouldn't be a valid event, so we don't care about "messing" them up anyway. I find this on-the-fly logic easier to maintain, having a flag introduces the possibility of human error (forgetting to set the flag in a new agent method, for example).

tests/ga/test_monitor.py
```diff
@@ -139,184 +134,146 @@ def test_parse_json_event(self, *args):
...
-        TaskName_value = "ExtHandler"
-        KeywordName_value = ""
-
+    @staticmethod
+    def create_mock_sysinfo():
+        os_version_param = "OSVersion"
+        execution_mode_param = "ExecutionMode"
+        ram_param = "RAM"
+        processors_param = "Processors"
         vm_name_param = "VMName"
```
pgombar: `VMName` is populated in `self.sysinfo` in the `MonitorHandler`, but it is not a part of the database schema, so it would always be dropped in the new telemetry pipeline.

tests/ga/test_monitor.py
```diff
@@ -139,184 +134,146 @@ def test_parse_json_event(self, *args):
...
         role_instance_name_param = "RoleInstanceName"
-        execution_mode_param = "ExecutionMode"
-        container_id_param = "ContainerId"
-        GAVersion_param = "GAVersion"
-        OpcodeName_param = "OpcodeName"
-        EventTid_param = "EventTid"
-        EventPid_param = "EventPid"
-        TaskName_param = "TaskName"
-        KeywordName_param = "KeywordName"
+        location_param = "Location"
```
pgombar: `Location` is populated in `self.sysinfo` in the `MonitorHandler`, but it is not a part of the database schema, so it would always be dropped in the new telemetry pipeline.

azurelinuxagent/common/event.py
pgombar: This is outdated; the current design only keeps extension-specific fields for extension events. Refer to the description of this PR for more details.

azurelinuxagent/common/event.py
```diff
@@ -393,11 +393,15 @@ def _clean_up_message(message):
...
-        Default fields are only populated by Agent and not the extension. Agent will fill up any event if they don't
-        have the default params. Example: GAVersion and ContainerId are populated for agent events on the fly,
-        but not for extension events. Add it if it's missing.
+        Default fields are only populated by the agent and not the extension. The agent will populate these fields for
+        events that don't already have them. Example: GAVersion and ContainerId are populated for agent events on the
+        fly, but not for extension events, so we will always override these fields with the values from the agent.
+
+        Similarly, there are some fields that have a special use case when emitted by the agent, like OpcodeName (used
+        to capture the actual time of event generation), EventTid, EventPid and TaskName. In case the event is emitted
+        from the agent, override the values. Otherwise, keep the current value, if exists, and set to empty if not.
```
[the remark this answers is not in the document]
pgombar: Outdated.

azurelinuxagent/common/event.py
pgombar: Updated, thanks.

narrieta:
@pgombar - for agent events, why add sys info when sending the event and not when creating it? thanks

pgombar:
> @pgombar - for agent events, why add sys info when sending the event and not when creating it? thanks

Because `sysinfo` is an instance attribute of `MonitorHandler` and gets populated in the constructor. This is because we depend on the protocol and IMDS to fetch some parameter values. I would need to do more refactoring to pull that out, since agent events are created from `event.py` which doesn't have a reference to `MonitorHandler`.

vrdmr:
> @pgombar - for agent events, why add sys info when sending the event and not when creating it? thanks

Also, `sys_info` fields don't change during the course of the agent and this we don't need to put it into the event file.

narrieta:
> Also, sys_info fields don't change during the course of the agent 

Today :)

narrieta:
@pgombar part of the motivation for this cleanup is to achieve consistency when populating events. We've already had multiple issues for populating some field at different times (creation vs send).

azurelinuxagent/common/telemetryevent.py
```diff
@@ -34,11 +34,21 @@ def __init__(self, eventId=None, providerId=None):
         self.eventId = eventId
         self.providerId = providerId
         self.parameters = DataContractList(TelemetryEventParam)
+        self.file_type = ""
```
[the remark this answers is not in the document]
pgombar: Actually, `extension_file_type` would be misleading, since this is a class for all telemetry events (both extension and agent).

narrieta:
@pgombar - thanks for the great description!

azurelinuxagent/common/event.py
```diff
@@ -264,6 +310,21 @@ def save_event(self, data):
             msg = "Failed to write events to file: {0}".format(e)
             raise EventError(msg)
 
+    @staticmethod
+    def collect_event(evt_file_name):
+        try:
+            logger.verbose("Found event file: {0}", evt_file_name)
+            with open(evt_file_name, "rb") as evt_file:
+                # if fail to open or delete the file, throw exception
```
narrieta: can we remove this comment?

azurelinuxagent/common/event.py
```diff
@@ -264,6 +310,21 @@ def save_event(self, data):
...
+                data_str = evt_file.read().decode("utf-8")
+            logger.verbose("Processed event file: {0}", evt_file_name)
+            os.remove(evt_file_name)
+            return data_str
+        except (IOError, OSError, UnicodeDecodeError) as e:
+            os.remove(evt_file_name)
```
narrieta: let's move the os.remove to a finally clause instead of having it in 2 places

azurelinuxagent/common/sysinfo.py
```diff
@@ -0,0 +1,100 @@
...
+            return SysInfo._instance
+
+    class __impl(object):
+        def __init__(self):
+            self._vminfo = None
+            self._compute_info = None
+            self._osutil = get_osutil()
+            self._telemetry_params = []
+
+        def _set_vminfo(self, vminfo, compute_info):
```
narrieta: you marked this as private, but it is being used from WireProtocol

azurelinuxagent/common/protocol/wire.py
```diff
@@ -108,6 +108,22 @@ def update_host_plugin_from_goal_state(self):
...
+        except ProtocolError as e:
+            logger.warn("Failed to get system info: {0}", ustr(e))
+
+        try:
+            compute_info = imds_client.get_compute()
+        except (HttpError, ValueError) as e:
+            logger.warn("Failed to get IMDS info: {0}", ustr(e))
+
+        sysinfo = SysInfo.get_instance()
+        sysinfo._set_vminfo(vminfo, compute_info)
```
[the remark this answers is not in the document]
narrieta: let's ensure that if an error happens, the telemetry fields get initialized with dummy values instead of None

azurelinuxagent/common/telemetryevent.py
narrieta: I share Laveesh's concern about extensions forgetting to set the Name parameter. If not a flag, let's try to find another way to handle this.

tests/common/test_event.py
```diff
@@ -19,23 +19,50 @@
...
 from azurelinuxagent.common.protocol.wire import GoalState
-from azurelinuxagent.common.telemetryevent import TelemetryEventParam
+from azurelinuxagent.common.telemetryevent import TelemetryEvent, TelemetryEventParam
 from azurelinuxagent.common.utils import fileutil
 from azurelinuxagent.common.utils.extensionprocessutil import read_output
-from azurelinuxagent.common.version import CURRENT_AGENT, CURRENT_VERSION
-from azurelinuxagent.ga.monitor import MonitorHandler
-from tests.tools import AgentTestCase, data_dir, load_data, Mock, patch
+from azurelinuxagent.common.version import AGENT_NAME, CURRENT_AGENT, CURRENT_VERSION
+from tests.ga.test_monitor import SysInfoMock
```
narrieta: You should refactor SysInfoMock into its own file.

BTW, currently the definition is on test_sysinfo, not test_monitor :)

tests/common/test_event.py
```diff
@@ -19,23 +19,50 @@
...
+        self.mock_sysinfo = patch("azurelinuxagent.common.sysinfo.SysInfo.get_instance", return_value=SysInfoMock)
+        self.mock_sysinfo.start()
+
+    def tearDown(self):
+        self.mock_sysinfo.stop()
+
+    def test_parse_xml_event(self, *args):
+        data_str = load_data('ext/event_from_extension.xml')
+        event = parse_xml_event(data_str)
+        self.assertNotEqual(None, event)
```
narrieta: since we are here, can we improve test_parse_* to do more substantial checks?

azurelinuxagent/common/event.py
```diff
@@ -394,51 +454,126 @@ def _clean_up_message(message):
...
+                if event.is_extension_event() or FlexibleVersion(version) >= FlexibleVersion("2.2.47"):
+                    continue
+
+                # The event filename is <epoch_time>.tld. The factor 1000000 is defined in save_event.
+                epoch_time = float(event_file[:-4]) / 1000000.0
+                event_creation_time = datetime.fromtimestamp(epoch_time).strftime(u'%Y-%m-%dT%H:%M:%S.%fZ')
+                EventLogger._update_old_event_schema(event, event_creation_time)
+
+                data = get_properties(event)
+                fileutil.write_file(event_file_path, json.dumps(data))
```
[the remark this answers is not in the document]
pgombar: This cleanup method is being called even before the monitor handler starts up, so we should be good.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -108,6 +108,22 @@ def update_host_plugin_from_goal_state(self):
     def get_endpoint(self):
         return self.client.get_endpoint()
 
+    def init_sysinfo(self, imds_client):
+        vminfo = VMInfo()
```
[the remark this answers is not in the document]
pgombar:
````
vminfo = VMInfo()
compute_info = ComputeInfo()
````
These constructors don't call any other methods and can't throw. My intention was to pass an empty object (with member variables set to `None`, but the references still exist) to the `sysinfo.set_vminfo(vminfo, compute_info` call, since then the references to the (empty) attributes wouldn't throw.

File: tests/common/test_event.py (hunk 14 above)
Comment (pgombar) [reply tests/common/test_event.py]: I'd really prefer doing it on a separate, smaller PR. I can add a task for myself.

File: azurelinuxagent/common/telemetryevent.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (pgombar) [reply azurelinuxagent/common/telemetryevent.py]: We are using the value of this member variable to populate the `ExtensionType` parameter, but it signifies the type of event file this telemetry event originated from, be it from agent or extension. It is add as part of the `_add_common_parameters_to_event` method invoked for all events.

File: azurelinuxagent/common/protocol/wire.py (hunk 16 above)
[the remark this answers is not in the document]
Comment (narrieta) [reply azurelinuxagent/common/protocol/wire.py]: ok

File: azurelinuxagent/common/sysinfo.py [hunk 17]
```diff
@@ -0,0 +1,110 @@
...
+from azurelinuxagent.common.future import ustr
+from azurelinuxagent.common.osutil import get_osutil
+from azurelinuxagent.common.telemetryevent import TelemetryEventParam
+from azurelinuxagent.common.version import DISTRO_NAME, DISTRO_VERSION, DISTRO_CODE_NAME, AGENT_EXECUTION_MODE
+
+
+class SysInfo(object):
+
+    _instance = None
+    _lock = Lock()
```
narrieta: not used

azurelinuxagent/common/protocol/wire.py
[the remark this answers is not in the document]
narrieta: so if we fail to create a valid object all the values will be None? we should try to choose some appropriate default since code using VMInfo may choke on None.

azurelinuxagent/common/sysinfo.py
```diff
@@ -0,0 +1,110 @@
...
+
+    @staticmethod
+    def get_instance():
+        if SysInfo._instance is None:
+            # The instance gets initialized only once, on extension handler process start up in update.py. As part of
```
narrieta: let's remove "in update.py" from the comment

tests/common/test_sysinfo.py
```diff
@@ -0,0 +1,85 @@
...
+class TestSysInfo(AgentTestCase):
+
+    def setUp(self):
+        AgentTestCase.setUp(self)
+        SysInfo._instance = None
+
+    @staticmethod
+    def build_vminfo():
+        vminfo = VMInfo()
+        vminfo.vmName = "TEST_VMName"
```
larohra: NIT: You can add all these parameters as part of the constructor itself, no need to do it separately 
```python
class VMInfo(DataContract):
    def __init__(self,
                 subscriptionId=None,
                 vmName=None,
                 containerId=None,
                 roleName=None,
                 roleInstanceName=None,
                 tenantName=None):

File: tests/common/test_sysinfo.py [hunk 20]
```diff
@@ -0,0 +1,85 @@
...
+        vminfo.tenantName = "TEST_TenantName"
+        vminfo.roleName = "TEST_RoleName"
+        vminfo.roleInstanceName = "Test_RoleInstanceName"
+
+        compute_info = ComputeInfo()
+        compute_info.location = "TEST_Location"
```
larohra: Same for this

tests/tools.py
```diff
@@ -41,12 +41,12 @@
 from azurelinuxagent.common.version import PY_VERSION_MAJOR
 
 try:
-    from unittest.mock import Mock, patch, MagicMock, ANY, DEFAULT, call
+    from unittest.mock import Mock, patch, MagicMock, ANY, DEFAULT, call, PropertyMock
```
larohra: What is `PropertyMock` used for?

azurelinuxagent/common/event.py
```diff
@@ -394,51 +455,131 @@ def _clean_up_message(message):
...
+        EventLogger._add_common_parameters_to_event(event, event_creation_time)
+        EventLogger._add_sysinfo_parameters_to_event(event)
 
-        # Converting the event_parameters into a dictionary as it helps to easily look up and get values
-        param_names = OrderedDict([(param.name, param.value) for param in event_parameters])
+    @staticmethod
+    def trim_extension_event_parameters(event):
+        """
+        This method is called for extension events before they are sent out. Per the agreement with extension
+        publishers, the parameters that belong to extensions and will be reported intact are Name, Version, Operation,
```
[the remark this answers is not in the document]
larohra: Spoke to @narrieta and looks like there is a formal document about it. We should get rid of that once we bring in the Telemetry pipeline for extensions.

pgombar:
Closing in favour of #1791

PR #1758, opened 2020-01-17
Handle errors from update_goal_state

narrieta:
When I moved the call to update_goal_state from get_vmagent_manifests/get_ext_handlers to the main loop I did not notice it does not handle exceptions.

The purpose of this PR is to handle exceptions mimicking what the original code did. 3 TODOs for subsequent PRs:
* other calls to update_goal_state should generate the same telemetry event
* the main loop should handle more exceptions
* the main loop needs some serious refactoring to make it amenable to unit tests

Verified the fix manually raising an exception from the debugger on a live agent.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1758)

### DIFF
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -317,7 +317,6 @@ class UpdateHandler(object):
                     goal_state_fetched = True
                 except Exception as e:
                     msg = u"Exception retrieving the goal state: {0}".format(ustr(traceback.format_exc()))
-                    logger.warn(msg)
                     add_event(AGENT_NAME, op=WALAEventOperation.FetchGoalState, version=CURRENT_VERSION, is_success=False, message=msg)
 
                 if goal_state_fetched:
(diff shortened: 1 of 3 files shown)


azurelinuxagent/ga/update.py
```diff
@@ -311,39 +311,47 @@ def run(self, debug=False):
...
-                            available_agent.name)
-                    break
+                goal_state_fetched = False
+                try:
+                    protocol.update_goal_state()
+                    goal_state_fetched = True
+                except Exception as e:
+                    msg = u"Exception retrieving the goal state: {0}".format(ustr(traceback.format_exc()))
+                    logger.warn(msg)
+                    add_event(AGENT_NAME, op=WALAEventOperation.FetchGoalState, version=CURRENT_VERSION, is_success=False, message=msg)
```
pgombar: Wouldn't `add_event` also log the message if `is_success=False`?

narrieta: thanks; removed

PR #1761, opened 2020-01-20
Handling errors while reading extension status files (Limiting Size and Transient issues)

vrdmr:
## Description
We need to limit the size of individual extension status files to <128KB.

Sample messages that are logged now - 

Json Parsing issues
```
2020-01-31T00:18:09.174246Z WARNING ExtHandler Failed to read any status for extension - Microsoft.OSTCExtensions.VMAccessForLinux-1.5.4. Failed due to Expecting ',' delimiter: line 1 column 10 (char 9)
2020-01-31T00:18:45.824286Z WARNING ExtHandler Failed to read any status for extension - Microsoft.OSTCExtensions.VMAccessForLinux-1.5.4. Failed due to Expecting property name enclosed in double quotes: line 1 column 3 (char 2)
```
Max status file size crossed
```
2020-01-31T00:22:08.113671Z WARNING ExtHandler [ExtensionStatusError] Handler - Microsoft.OSTCExtensions.VMAccessForLinux, status file /var/lib/waagent/Microsoft.OSTCExtensions.VMAccessForLinux-1.5.4/status/0.status of size 140267 bytes is too big. Max Limit allowed is 131072 bytes
```
Status file not found 
```
2020-01-31T01:56:51.706109Z WARNING ExtHandler Failed to find any status for extension - Microsoft.OSTCExtensions.VMAccessForLinux-1.5.4. Failed due to [Errno 2] No such file or directory: '/var/lib/waagent/Microsoft.OSTCExtensions.VMAccessForLinux-1.5.4/status/3.status'
```

In the portal, for really long status messages, the following will be shown:

Status message | Lorem ipsum   dolor sit amet, consectetur adipiscing elit. Nam neque tellus, eleifend in   massa eget, aliquam sagittis lacus. Aenean suscipit odio quis pharetra   efficitur. Aliquam viverra leo lectus, vit... TRUNCATED
-- | --

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1761)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -38,24 +38,24 @@
...
+from azurelinuxagent.common.version import AGENT_NAME, CURRENT_VERSION, DISTRO_NAME, DISTRO_VERSION, \
+    GOAL_STATE_AGENT_VERSION, PY_VERSION_MAJOR, PY_VERSION_MICRO, PY_VERSION_MINOR
 
 # HandlerEnvironment.json schema version
 HANDLER_ENVIRONMENT_VERSION = 1.0
 
 EXTENSION_STATUS_ERROR = 'error'
 EXTENSION_STATUS_SUCCESS = 'success'
+EXTENSION_STATUS_WARNING = 'warning'
+
```
narrieta: can we use EXTENSION_STATUS_* in VALID_EXTENSION_STATUS and EXTENSION_TERMINAL_STATUSES? thanks

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1099,20 +1102,27 @@ def collect_ext_status(self, ext):
...
+                raise IOError(msg)
+
             data = json.loads(data_str)
             parse_ext_status(ext_status, data)
         except IOError as e:
-            ext_status.message = u"Failed to get status file {0}".format(e)
+            ext_status.message = u"Failed to get status file: {0}".format(e)
             ext_status.code = -1
-            ext_status.status = "error"
+            ext_status.status = EXTENSION_STATUS_ERROR
```
narrieta: I believe these were going to change to warning?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1099,20 +1102,27 @@ def collect_ext_status(self, ext):
         ext_status = ExtensionStatus(seq_no=seq_no)
         try:
             data_str = fileutil.read_file(ext_status_file)
+
+            if len(data_str) > MAX_STATUS_FILE_SIZE_IN_BYTES:
+                msg = "Handler - {0}, status file {1} of size {2} bytes is too big. Max Limit allowed is {3} " \
```
narrieta: if we are propagating those messages to users, they need to be more oriented to them. you may want to add a detailed message for the publisher (telemetry would also be good) and a friendlier message for the user

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1099,20 +1102,27 @@ def collect_ext_status(self, ext):
...
             parse_ext_status(ext_status, data)
         except IOError as e:
-            ext_status.message = u"Failed to get status file {0}".format(e)
+            ext_status.message = u"Failed to get status file: {0}".format(e)
             ext_status.code = -1
-            ext_status.status = "error"
+            ext_status.status = EXTENSION_STATUS_ERROR
         except ExtensionError as e:
-            ext_status.message = u"Malformed status file {0}".format(e)
+            ext_status.message = u"Malformed status file: {0}".format(e)
```
narrieta: we should send telemetry for the publishers to become aware of ExtensionError and ValueError

narrieta: These messages propagated to the user need to improved

azurelinuxagent/ga/exthandlers.py
vrdmr: yes. I'll replace this with a _namedset to make it easy to use.

azurelinuxagent/ga/exthandlers.py
vrdmr: Yes. I may need to send a periodic event for this - knowing that this will get called every 15 seconds. Will update in the next iteration.

vrdmr: Yes. Mainly the (e) messages should be updated in the case as well. Will take a look at the error reporting in this scenario.

azurelinuxagent/ga/exthandlers.py
vrdmr: Yes. Will sync with the Windows side and check how to make it work. They don't send `ext_status.code = -1` and only send `ext_status.code = ExtensionErrorCodes.PluginSettingsStatusInvalid`. I am currently seeing where `-1` comes into play.

tests/ga/test_extension.py
```diff
@@ -1170,11 +1170,166 @@ def test_get_ext_handling_status(self, *args):
 
         os.path.exists = orig_state
 
+    @patch("azurelinuxagent.ga.exthandlers.ExtHandlerInstance.get_status_file_path")
+    def test_collect_ext_status(self, patch_get_status_file_path, *args):
+        """
+        Testing is_ext_handling_complete() with various input and
```
narrieta: can we have a better description of this test. Current description says it tests 'is_ext_handling_complete' with some inputs checking some outputs... no idea what that does :)

tests/ga/test_extension.py
```diff
@@ -1170,11 +1170,166 @@ def test_get_ext_handling_status(self, *args):
...
+        self.assertEqual(len(ext_status.substatusList), 1)
+        sub_status = ext_status.substatusList[0]
+        self.assertEqual(sub_status.code, "0")
+        self.assertEqual(sub_status.message, None)
+        self.assertEqual(sub_status.status, EXTENSION_STATUS_SUCCESS)
+
+    @patch("azurelinuxagent.ga.exthandlers.ExtHandlerInstance.get_status_file_path")
+    def test_collect_ext_status_very_large_status_message(self, patch_get_status_file_path, *args):
+        """
+        Testing collect_ext_status() with a very large status file (>128K) to what it re
```
narrieta: incomplete sentence?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -97,12 +102,14 @@ def get_traceback(e):
 
 def validate_has_key(obj, key, fullname):
     if key not in obj:
-        raise ExtensionError("Missing: {0}".format(fullname))
+        raise ExtensionStatusError("Invalid status format by extension: Missing {0} key".format(fullname),
```
pgombar: What is `fullname` here in relation to `key`? `key` is missing from the dictionary, shouldn't the error message be `"Missing {0} key".format(key)`?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1091,28 +1097,61 @@ def get_status_file_path(self, extension=None):
...
-            ext_status.code = -1
-            ext_status.status = "error"
-        except ExtensionError as e:
-            ext_status.message = u"Malformed status file {0}".format(e)
-            ext_status.code = ExtensionErrorCodes.PluginSettingsStatusInvalid
-            ext_status.status = "error"
+            msg = u"Failed to find any status for extension - {0}-{1}, Sequence number {2}. Failed due to {3}" \
+                .format(ext.name, self.ext_handler.properties.version, seq_no, e)
+            ext_status.message = msg
+            ext_status.code = ExtensionErrorCodes.PluginUnknownFailure
```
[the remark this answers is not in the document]
pgombar: It might not be the extension's fault we couldn't find the status file (e.g. low disk space so the file couldn't be saved), right?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1091,28 +1097,61 @@ def get_status_file_path(self, extension=None):
...
-            ext_status.code = -1
-            ext_status.status = "error"
+            msg = u"Failed to read any status for extension - {0}-{1}, Sequence number {2}. Failed due to {3}"\
+                .format(ext.name, self.ext_handler.properties.version, seq_no, e)
+            ext_status.message = msg
+            ext_status.code = ExtensionErrorCodes.PluginUnknownFailure
+            ext_status.status = EXTENSION_STATUS_WARNING
+
+        if ext_status.code == ExtensionErrorCodes.PluginUnknownFailure:
+            logger.periodic_warn(logger.EVERY_HALF_HOUR, ext_status.message)
```
pgombar: Add `[PERIODIC]` to the message if it's periodically logged?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1091,28 +1097,61 @@ def get_status_file_path(self, extension=None):
...
+            try:
+                parse_ext_status(ext_status, data)
+                if len(data_str) > MAX_STATUS_FILE_SIZE_IN_BYTES:
+                    raise ExtensionStatusError("For Extension Handler {0}-{1}, the status file {2} of size {3} bytes is"
+                                               " too big. Max Limit allowed is {4} bytes"
+                                               .format(ext.name, self.ext_handler.properties.version, ext_status_file,
+                                                       len(data_str), MAX_STATUS_FILE_SIZE_IN_BYTES),
+                                               code=ExtensionStatusError.MaxSizeExceeded)
+            except ExtensionStatusError as e:
+                logger.periodic_warn(logger.EVERY_HALF_HOUR, ustr(e))
```
pgombar: Add [PERIODIC] to the message if it's periodically logged?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1091,28 +1097,61 @@ def get_status_file_path(self, extension=None):
...
+                add_periodic(delta=logger.EVERY_HALF_HOUR, name=ext.name, version=self.ext_handler.properties.version,
+                             op=WALAEventOperation.StatusProcessing, is_success=False, message=ustr(e), log_event=False)
+
+                if e.code == ExtensionStatusError.MaxSizeExceeded:
+                    # Emptying the substatus to reduce the size, and preserve other fields of the text
+                    ext_status.substatusList = []
+                    ext_status.message = ext_status.message if len(ext_status.message) < 200 \
+                        else ext_status.message[:200] + " ... TRUNCATED MESSAGE"
```
pgombar: Suggestion - `... TRUNCATED MESSAGE` seems a bit verbose, I think the same effect could be achieved more concisely with `[TRUNCATED]` at the end of the message.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1091,28 +1097,61 @@ def get_status_file_path(self, extension=None):
...
+                elif e.code == ExtensionStatusError.StatusFileMalformed:
+                    ext_status.message = "Could not get a valid status from the extension {0}-{1}. Encountered the " \
+                                         "following error: {0}".format(ext.name, self.ext_handler.properties.version,
+                                                                       ustr(e))
+                    ext_status.code = ExtensionErrorCodes.PluginSettingsStatusInvalid
+                    ext_status.status = EXTENSION_STATUS_WARNING
```
pgombar: What happens when `e.code == ExtensionStatusError.StatusFileNotFound`? We would still be returning `ext_status`? Do you need to set the message to explicitly say the status file was not found? How would this be reported to the user?

One other thing; this method is called in `report_ext_status` where there is a check:
````
ext_status = self.collect_ext_status(ext)            
if ext_status is None:
    continue
````

However, this is never true since we will always return `ext_status` which is defined in the beginning of the method as `ExtensionStatus(seq_no=seq_no)`.

File: tests/ga/test_extension.py [hunk 13]
```diff
@@ -1170,11 +1170,166 @@ def test_get_ext_handling_status(self, *args):
...
+        extension = Extension(name=handler_name)
+        exthandler.properties.extensions.append(extension)
+
+        ext_handler_i = ExtHandlerInstance(exthandler, protocol)
+
+        with patch('azurelinuxagent.common.utils.fileutil.read_file') as patch_read_file:
+            patch_read_file.side_effect = IOError("No such file or directory: {0}".format(status_file_path))
+            ext_status = ext_handler_i.collect_ext_status(extension)
+
+            self.assertEqual(ext_status.code, -1)
```
pgombar: `ExtensionErrorCodes.PluginUnknownFailure` instead of -1?

tests/ga/test_extension.py
```diff
@@ -1170,11 +1170,166 @@ def test_get_ext_handling_status(self, *args):
...
+        handler_name = "TestHandler"
+        exthandler = ExtHandler(name=handler_name)
+        exthandler.properties.version = "1.0.0"
+        extension = Extension(name=handler_name)
+        exthandler.properties.extensions.append(extension)
+
+        ext_handler_i = ExtHandlerInstance(exthandler, protocol)
+        ext_status = ext_handler_i.collect_ext_status(extension)
+
+        self.assertEqual(ext_status.code, -1)
```
pgombar: `ExtensionErrorCodes.PluginUnknownFailure` instead of -1?

azurelinuxagent/ga/exthandlers.py
vrdmr: > However, this is never true since we will always return ext_status which is defined in the beginning of the method as ExtensionStatus(seq_no=seq_no).

This would be true [only] in the case of 
```
seq_no, ext_status_file = self.get_status_file_path(ext)
if seq_no == -1:
    return None
```

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1091,28 +1097,61 @@ def get_status_file_path(self, extension=None):
...
+            logger.periodic_warn(logger.EVERY_HALF_HOUR, ext_status.message)
+            add_periodic(delta=logger.EVERY_HALF_HOUR, name=ext.name, version=self.ext_handler.properties.version,
+                         op=WALAEventOperation.StatusProcessing, is_success=False, message=ext_status.message,
+                         log_event=False)
+        else:
+            # We did not encounter any PluginUnknownFailures and thus the status file was correctly written and has
+            # valid json.
+            try:
+                parse_ext_status(ext_status, data)
+                if len(data_str) > MAX_STATUS_FILE_SIZE_IN_BYTES:
```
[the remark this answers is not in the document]
vrdmr: There is no individual status size check but a limit to the aggregated status.

```
const long MaxVMStatusBlobSizeInMB = 1;  // 1MB limit on VM aggregate status blob
```

The limit to the entire file length is just a precaution on our end to not have an extension status go out of proportions.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -35,36 +35,38 @@
...
 
-EXTENSION_STATUS_ERROR = 'error'
-EXTENSION_STATUS_SUCCESS = 'success'
-VALID_EXTENSION_STATUS = ['transitioning', 'error', 'success', 'warning']
-EXTENSION_TERMINAL_STATUSES = ['error', 'success']
+VALID_EXTENSION_STATUS = NamedSet(('transitioning', 'warning', 'error', 'success'))
+
+EXTENSION_STATUS_ERROR = VALID_EXTENSION_STATUS.error
+EXTENSION_STATUS_SUCCESS = VALID_EXTENSION_STATUS.success
+EXTENSION_STATUS_WARNING = VALID_EXTENSION_STATUS.warning
```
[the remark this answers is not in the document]
vrdmr: I wanted to make it explicit. But yes, VALID_EXTENSION_STATUS.* is explicit enough. I'll refactor this.

azurelinuxagent/ga/exthandlers.py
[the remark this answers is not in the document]
vrdmr: @pgombar Yes. That was the idea as we don't know if it is the extension at fault or something on the box which is causing this issue.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1091,28 +1097,61 @@ def get_status_file_path(self, extension=None):
...
+            msg = u"Failed to find any status for extension - {0}-{1}, Sequence number {2}. Failed due to {3}" \
+                .format(ext.name, self.ext_handler.properties.version, seq_no, e)
+            ext_status.message = msg
+            ext_status.code = ExtensionErrorCodes.PluginUnknownFailure
+            ext_status.status = EXTENSION_STATUS_WARNING
         except ValueError as e:
-            ext_status.message = u"Malformed status file {0}".format(e)
-            ext_status.code = -1
-            ext_status.status = "error"
+            msg = u"Failed to read any status for extension - {0}-{1}, Sequence number {2}. Failed due to {3}"\
```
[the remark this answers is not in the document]
vrdmr: Yes. Invalid json throws Value error.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1091,28 +1097,61 @@ def get_status_file_path(self, extension=None):
...
-            parse_ext_status(ext_status, data)
         except IOError as e:
-            ext_status.message = u"Failed to get status file {0}".format(e)
-            ext_status.code = -1
-            ext_status.status = "error"
-        except ExtensionError as e:
-            ext_status.message = u"Malformed status file {0}".format(e)
-            ext_status.code = ExtensionErrorCodes.PluginSettingsStatusInvalid
-            ext_status.status = "error"
+            msg = u"Failed to find any status for extension - {0}-{1}, Sequence number {2}. Failed due to {3}" \
```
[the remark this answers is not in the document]
vrdmr: Yes. When the file was unavailable, or we couldn't read it due to some permissions issue.

azurelinuxagent/ga/exthandlers.py
vrdmr: The full name is the full path of the key that you are searching for. The use case was better to error reporting.

It essentially means - "Missing {0} key from the {1} node".format(key, obj), but the obj node is currently a dictionary, and we needed a way to explain that. Thus, the fullname helps in clarifying whats missing from where (eg: formattedMessage/lang).

Eg: `validate_has_key(formatted_message, 'lang', 'formattedMessage/lang')`

azurelinuxagent/ga/exthandlers.py
vrdmr: This looks cleaner.

azurelinuxagent/common/exception.py
```diff
@@ -100,6 +100,19 @@ def __init__(self, msg=None, inner=None, code=-1):
         super(ExtensionDownloadError, self).__init__(msg, inner, code)
 
 
+class ExtensionStatusError(ExtensionError):
+    """
+    When extension failed to provide a valid status file
+    """
+    Unknown = -1
```
narrieta: What is the use of Unknown? Currently I don't see it being used anywhere. You may want to force the  caller to provide a value.

azurelinuxagent/common/protocol/util.py
```diff
@@ -46,6 +46,13 @@
 WIRE_PROTOCOL_NAME = "WireProtocol"
 
 
+class NamedSet(set):
```
narrieta: We may want to rename this class. 'name' not being an item in the NamedSet raises an exception, which is not quite the behavior of a set in Python.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -35,36 +35,38 @@
 import azurelinuxagent.common.conf as conf
 import azurelinuxagent.common.logger as logger
 import azurelinuxagent.common.utils.fileutil as fileutil
+from azurelinuxagent.common.protocol.util import NamedSet
```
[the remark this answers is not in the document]
narrieta: > Since NamedSet doesn't actually have anything to do with protocol, it should be defined elsewhere. Is there a better place for it in one of the common modules?

+1

azurelinuxagent/ga/exthandlers.py
```diff
@@ -35,36 +35,33 @@
...
+    GOAL_STATE_AGENT_VERSION, PY_VERSION_MAJOR, PY_VERSION_MICRO, PY_VERSION_MINOR
 
 # HandlerEnvironment.json schema version
 HANDLER_ENVIRONMENT_VERSION = 1.0
 
-EXTENSION_STATUS_ERROR = 'error'
-EXTENSION_STATUS_SUCCESS = 'success'
-VALID_EXTENSION_STATUS = ['transitioning', 'error', 'success', 'warning']
-EXTENSION_TERMINAL_STATUSES = ['error', 'success']
+VALID_EXTENSION_STATUS = NamedSet(('transitioning', 'warning', 'error', 'success'))
```
narrieta: We need to start making a distinction between public and private module members. Could you review all of the new definitions and mark the privates?

azurelinuxagent/ga/exthandlers.py
narrieta: > What happens when e.code == ExtensionStatusError.StatusFileNotFound? We would still be returning ext_status? Do you need to set the message to explicitly say the status file was not found? How would this be reported to the user?

+1

narrieta: > > This would be true [only] in the case of
> 
> seq_no, ext_status_file = self.get_status_file_path(ext)
> if seq_no == -1:
>     return None

But we still need to report status, right?

azurelinuxagent/common/exception.py
narrieta: Also, I think we need to start defining those exceptions on the module that raises them. They are part of the module's interface.

tests/data/ext/sample-status-invalid-format.json
```diff
@@ -0,0 +1,36 @@
+[
```
narrieta: can we add a comment somewhere in the file to explain what is wrong with it? maybe a string in caps as part of the message? for an example, see tests/data/events/custom_script_invalid_json.tld

narrieta: same for the other files

tests/ga/test_extension.py
```diff
@@ -1171,11 +1171,167 @@ def test_get_ext_handling_status(self, *args):
 
         os.path.exists = orig_state
 
+    @patch("azurelinuxagent.ga.exthandlers.ExtHandlerInstance.get_status_file_path")
```
narrieta: ExtHandlerInstance.get_status_file_path should be private. Let's try not to mock it.

Other tests in this file (e.g. test_ext_handler_no_reporting_status) run an extension handler to produce a status file. That seems a better approach.

narrieta: Also, by mocking it we are bypassing the logic implemented there. If we introduce a bug there we may miss it.

For example, your description of this test ("This test validates that collect_ext_status correctly picks up the status file") seems to imply you want to test the part that picks up the correct status, but you are mocking it out.

narrieta: > Other tests in this file (e.g. test_ext_handler_no_reporting_status) run an extension handler to produce a status file. 

Alternatively, you could refactor the parsing logic into its own class (which would take a file as input) and then test that class separately.

azurelinuxagent/ga/exthandlers.py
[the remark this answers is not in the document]
vrdmr: Yes. The Json could be malformed (which could be the issue caused by extension.). I'll change this.

azurelinuxagent/ga/exthandlers.py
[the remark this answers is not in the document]
vrdmr: > Then we need to improve "Failed to find any status"

I can improve the wording just a little to make the extension and the fault clearer. But apart from this, we don't have much context to be sent to the customer on what could be the failure, which could be preventing the extension to generate the status.
```
msg = u"We couldn't find any status for {0}-{1} extension, for the sequence number {2}. It failed due to {3}".format(ext.name, self.ext_handler.properties.version, seq_no, e)
```

> Also, it seems like those kind of errors are permanent, so we should fail the extension instead of setting it up to warning.

The need for making them warning came after an extension (backup or encryption) requested that we retry status reading or send non-terminal status because they couldn't generate the status file in time. Even if we want to fail them early, there have been cases that it was temporary.

azurelinuxagent/ga/exthandlers.py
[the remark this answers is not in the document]
vrdmr: > Since NamedSet doesn't actually have anything to do with protocol, it should be defined elsewhere. Is there a better place for it in one of the common modules?

The apt only place I could find was in future.py (bringing in future python-supported feature to present). Other common utility modules didn't seem to be the correct place. And I didn't want to create a new utility module, which might not be used again, only for this one function.

azurelinuxagent/ga/exthandlers.py
vrdmr: Done - renamed to `full_key_path`. To describe the path.

azurelinuxagent/ga/exthandlers.py
vrdmr: >But we still need to report status, right?

Yes - and we do send the status.
```
if ext_status.code == ExtensionErrorCodes.PluginUnknownFailure or ext_status.code == \
                ExtensionErrorCodes.PluginSettingsStatusInvalid:
   ... # Log stuff and send telemetry.
else:
   ... # Parse status.

return ext_status
```

azurelinuxagent/common/protocol/util.py
vrdmr: This can be named as AttributeSet, where each member of the set actually becomes an attribute. I am open to other ideas as well.

azurelinuxagent/common/exception.py
vrdmr: > What is the use of Unknown? Currently I don't see it being used anywhere. You may want to force the caller to provide a value.

Yes. Unknown was useless - Removed it now and made code compulsory.

> Also, I think we need to start defining those exceptions on the module that raises them. They are part of the module's interface.

I partially agree with this. I don't want to break away the current exceptions in this PR and it is better to see all the exceptions together.

azurelinuxagent/ga/exthandlers.py
vrdmr: Done. I didn't change it for all the cases where tests and multiple modules need it.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1085,28 +1086,61 @@ def get_status_file_path(self, extension=None):
...
-            ext_status.code = -1
-            ext_status.status = "error"
+            msg = u"Failed to read any status for extension - {0}-{1}, Sequence number {2}. Failed due to {3}"\
+                .format(ext.name, self.ext_handler.properties.version, seq_no, e)
+            ext_status.message = msg
+            ext_status.code = ExtensionErrorCodes.PluginUnknownFailure
+            ext_status.status = VALID_EXTENSION_STATUS.warning
+
+        if ext_status.code == ExtensionErrorCodes.PluginUnknownFailure:
+            logger.periodic_warn(logger.EVERY_HALF_HOUR, "[PERIODIC] " + ext_status.message)
```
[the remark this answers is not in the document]
vrdmr: Yes. This does not change previous behavior (which was sending these three fields). 
Also, I will double-check with a prod VM and edited status.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1085,28 +1086,61 @@ def get_status_file_path(self, extension=None):
...
+            try:
+                parse_ext_status(ext_status, data)
+                if len(data_str) > MAX_STATUS_FILE_SIZE_IN_BYTES:
+                    raise ExtensionStatusError("For Extension Handler {0}-{1}, the status file {2} of size {3} bytes is"
+                                               " too big. Max Limit allowed is {4} bytes"
+                                               .format(ext.name, self.ext_handler.properties.version, ext_status_file,
+                                                       len(data_str), MAX_STATUS_FILE_SIZE_IN_BYTES),
+                                               code=ExtensionStatusError.MaxSizeExceeded)
+            except ExtensionStatusError as e:
+                logger.periodic_warn(logger.EVERY_HALF_HOUR, "[PERIODIC] " + ustr(e))
```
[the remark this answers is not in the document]
vrdmr: Add Seq no.

azurelinuxagent/common/protocol/util.py
vrdmr: Renamed to AttributeEnum.

azurelinuxagent/ga/exthandlers.py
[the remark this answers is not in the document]
vrdmr: Placed it to `types.py`.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1085,28 +1086,61 @@ def get_status_file_path(self, extension=None):
...
+                add_periodic(delta=logger.EVERY_HALF_HOUR, name=ext.name, version=self.ext_handler.properties.version,
+                             op=WALAEventOperation.StatusProcessing, is_success=False, message=ustr(e), log_event=False)
+
+                if e.code == ExtensionStatusError.MaxSizeExceeded:
+                    # Emptying the substatus to reduce the size, and preserve other fields of the text
```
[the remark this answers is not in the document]
vrdmr: Done. Fixed.

tests/data/ext/sample-status-invalid-format.json
vrdmr: Done.

vrdmr: Renamed the files to point out the issues.

tests/ga/test_extension.py
vrdmr: I've gone to make the extension folder with config and status folders in it, with the files dropped in. This exercised all the parts of that method correctly. The only thing that is mocked is a new lib file. So no internal things mocked out.

azurelinuxagent/common/types.py
```diff
@@ -0,0 +1,26 @@
+# Microsoft Azure Linux Agent
```
narrieta: I would recommend not adding this type now.

Seems to be a copy of _nameset in protocol util, which we will remove along with the metadata protocol.

This type is very simplistic and lacks a lot of functionality provided by enums (like assigning values to the items in the enum). Its design is odd since it is an extension of a set, etc. Also it may introduce subtle coding errors if we use it extensively because it doesn't really behave as an enum. For example

```
>>> Foo = Enum('Foo', 'one two three')
>>> Bar = Enum('Foo', 'one four five')
>>> f = Foo.one
>>> b = Bar.one
>>> f == b
False

```
vs

```
>>> Foo = AttributeEnum(('one', 'two', 'three'))
>>> Bar = AttributeEnum(('one', 'four', 'five'))
>>> f = Foo.one
>>> b = Bar.one
>>> f == b
True
```

If we really want an enum type we could consider using the backport of enum instead (https://pypi.org/project/enum34/), but we would need to pack it along with the agent.

azurelinuxagent/common/exception.py
```diff
@@ -100,6 +100,19 @@ def __init__(self, msg=None, inner=None, code=-1):
         super(ExtensionDownloadError, self).__init__(msg, inner, code)
 
 
+class ExtensionStatusError(ExtensionError):
```
narrieta: we talked about moving this to exthandlers.py

azurelinuxagent/ga/exthandlers.py
```diff
@@ -38,44 +38,52 @@
...
 
 # HandlerEnvironment.json schema version
-HANDLER_ENVIRONMENT_VERSION = 1.0
+_HANDLER_ENVIRONMENT_VERSION = 1.0
 
-EXTENSION_STATUS_ERROR = 'error'
-EXTENSION_STATUS_SUCCESS = 'success'
-VALID_EXTENSION_STATUS = ['transitioning', 'error', 'success', 'warning']
-EXTENSION_TERMINAL_STATUSES = ['error', 'success']
+VALID_EXTENSION_STATUS = AttributeEnum(('transitioning', 'warning', 'error', 'success'))
```
narrieta: seems like this could be private as well

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1399,6 +1439,43 @@ def get_seq_no(self):
...
+            except IOError as e:
+                failed_to_read = True
+                raised_exception = e
+            except ValueError as e:
+                failed_to_parse_json = True
+                raised_exception = e
+            time.sleep(_STATUS_FILE_RETRY_DELAY)
+
+        if failed_to_read:
+            msg = u"We couldn't find any status for {0}-{1} extension, for the sequence number {2}. It failed due to " \
```
narrieta: we talked about changing this from "couldn't find" to "couldn't read"

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1399,6 +1439,43 @@ def get_seq_no(self):
...
+                  u"{3}".format(ext.name, self.ext_handler.properties.version, seq_no, raised_exception)
+            raise ExtensionStatusError(msg=msg, code=ExtensionStatusError.StatusFileNotFound)
+        elif failed_to_parse_json:
+            msg = u"Failed to read any status for extension - {0}-{1}, Sequence number {2}. Failed due to {3}" \
```
narrieta: similarly, this would be "Failed to parse", and also try to make it more user-friendly. Maybe something like "Extension foo reported its status using an incorrect format", or something along those lines.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1084,29 +1090,63 @@ def get_status_file_path(self, extension=None):
...
             return None
 
+        data = None
+        data_str = None
         ext_status = ExtensionStatus(seq_no=seq_no)
+
+        try:
+            data_str, data = self._read_and_parse_json_status_file(ext, seq_no, ext_status_file)
+        except ExtensionStatusError as e:
+            logger.periodic_warn(logger.EVERY_HALF_HOUR, u"[PERIODIC] " + ustr(e))
```
narrieta: we talked about periodic logging may hide errors when the incarnation changes and, *at the very least*, document this so that people don't mess up the messages unintentionally.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1084,29 +1090,63 @@ def get_status_file_path(self, extension=None):
...
+        except ExtensionStatusError as e:
+            msg = u"For Extension Handler {0}-{1} for the sequence number {2}, the status file {3}. " \
+                  u"Encountered the following error: {4}".format(ext.name, self.ext_handler.properties.version, seq_no,
+                                                                 ext_status_file, ustr(e))
+            logger.periodic_warn(logger.EVERY_DAY, u"[PERIODIC] " + msg)
+            add_periodic(delta=logger.EVERY_HALF_HOUR, name=ext.name, version=self.ext_handler.properties.version,
+                         op=WALAEventOperation.StatusProcessing, is_success=False, message=msg, log_event=False)
+
+            if e.code == ExtensionStatusError.MaxSizeExceeded:
+                ext_status.message = self._truncate_message(ext_status.message)
```
narrieta: we discussed how truncating from > 128k to 200 may not be the best approach. And also the loop below wouldn't catch the case where there are many, many, substatuses. Why not truncating each field (message and then each substatus) to 128K and keep appending until you exceed 128K?

azurelinuxagent/ga/exthandlers.py
narrieta: also 'StatusFileNotFound' is misleading

tests/data/ext/sample-status-invalid-status.json
```diff
@@ -0,0 +1,35 @@
+[
+    {
```
narrieta: we talked about adding some kind of comment (maybe as part of one of the strings) to point out what is wrong with this file

tests/ga/test_extension.py
```diff
@@ -387,6 +390,27 @@ def _set_up_update_test_and_update_gs(self, patch_command, *args):
...
+        handler_version = "1.0.0"
+        mock_lib_dir.return_value = self.lib_dir
+        fileutil.mkdir(os.path.join(self.lib_dir, handler_name + "-" + handler_version, "config"))
+        fileutil.mkdir(os.path.join(self.lib_dir, handler_name + "-" + handler_version, "status"))
+        shutil.copy(os.path.join(data_dir, "ext", status_file),
+                    os.path.join(self.lib_dir, handler_name + "-" + handler_version, "status", "0.status"))
+        shutil.copy(tempfile.mkstemp(prefix="test-file")[1],
+                    os.path.join(self.lib_dir, handler_name + "-" + handler_version, "config", "0.settings"))
+
+        test_data = mockwiredata.WireProtocolData(mockwiredata.DATA_FILE)
```
narrieta: for new tests let's use mock_wire_protocol instead of mockwiredata.

the problem with using mockwiredata is that each test needs to create their mocks for http get and crypt utils. When there is need to change something on the mock (e.g. when i had to add support for POST or mock http request instead of http get) the impact on tests can be huge.

narrieta: btw, as i continue refactoring protocol very likely i will have to change those mocks again... i am not looking forward to all the tests update i will have to do

tests/data/ext/sample-status-invalid-status.json
vrdmr: Sorry - Missed this change. The status/status key was missing in this case. Will send in an update again.

azurelinuxagent/ga/exthandlers.py
vrdmr: As we discussed, the underlying messages raised by `_read_and_parse_json_status_file` for exception cases contain the information about the sequence number and other relevant information required in the logs.

But I understand your concern that the messages should be crafted when the logs are written, such that the messages are clear at the place they are logged. I'll bring the `msg` building up one level.
```
"We couldn't find any status for {0}-{1} extension, for the sequence number {2}. It failed due to {3}".format(ext.name, self.ext_handler.properties.version, seq_no, raised_exception)
```
```
msg = u"Failed to read any status for extension - {0}-{1}, Sequence number {2}. Failed due to {3}".format(ext.name, self.ext_handler.properties.version, seq_no, raised_exception)
```

tests/ga/test_extension.py
```diff
@@ -1171,11 +1195,121 @@ def test_get_ext_handling_status(self, *args):
...
+    def test_collect_ext_status(self, mock_lib_dir, *args):
+        """
+        This test validates that collect_ext_status correctly picks up the status file (sample-status.json) and then
+        parses it correctly.
+        """
+        ext_handler_i, extension = self._setup_extension_for_validating_collect_ext_status(mock_lib_dir,
+                                                                                           "sample-status.json", *args)
+        ext_status = ext_handler_i.collect_ext_status(extension)
+
+        self.assertEqual(ext_status.code, 1)
```
narrieta: Let's add strings describing what the error is. Not only they document the code, but also help a lot when debugging test failures. When this test fails you would just see
> 2 != 1
For sample strings look at test_wire.py in #1800

azurelinuxagent/ga/exthandlers.py
narrieta: My concern is different. We can discuss offline.

tests/ga/test_extension.py
```diff
@@ -1171,11 +1195,121 @@ def test_get_ext_handling_status(self, *args):
...
+
+    @patch("azurelinuxagent.common.conf.get_lib_dir")
+    def test_collect_ext_status_read_file_read_exceptions(self, mock_lib_dir, *args):
+        """
+        Testing collect_ext_status to validate the readfile exceptions.
+        """
+        ext_handler_i, extension = self._setup_extension_for_validating_collect_ext_status(mock_lib_dir,
+                                                                                           "sample-status.json", *args)
+
+        with patch('azurelinuxagent.common.utils.fileutil.read_file') as patch_read_file:
```
narrieta: let's do this patch conditional so that it applies only to the status file. If the implementation of collect_ext_status ever needs to read another file likely this test will break.

For example, see #1800 

We need to be careful when mocking "basic" functions such as reading a file, logging a message, producing telemetry, etc

azurelinuxagent/ga/exthandlers.py
vrdmr: Done.

azurelinuxagent/ga/exthandlers.py
vrdmr: Done.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1085,28 +1086,61 @@ def get_status_file_path(self, extension=None):
...
+                                               code=ExtensionStatusError.MaxSizeExceeded)
+            except ExtensionStatusError as e:
+                logger.periodic_warn(logger.EVERY_HALF_HOUR, "[PERIODIC] " + ustr(e))
+                add_periodic(delta=logger.EVERY_HALF_HOUR, name=ext.name, version=self.ext_handler.properties.version,
+                             op=WALAEventOperation.StatusProcessing, is_success=False, message=ustr(e), log_event=False)
+
+                if e.code == ExtensionStatusError.MaxSizeExceeded:
+                    # Emptying the substatus to reduce the size, and preserve other fields of the text
+                    ext_status.substatusList = []
+                    ext_status.message = ext_status.message if len(ext_status.message) < 200 \
```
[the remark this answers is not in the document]
vrdmr: Solved in the new approach.

azurelinuxagent/ga/exthandlers.py
vrdmr: Done.

azurelinuxagent/ga/exthandlers.py
vrdmr: Done.

tests/ga/test_extension.py
vrdmr: Fixed. Added better documentation to describe what the numbers mean.

azurelinuxagent/ga/exthandlers.py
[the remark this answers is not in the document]
vrdmr: Validated this. If we populate the three fields, CRP uses the three and posts it to the portal.

tests/ga/test_extension.py
```diff
@@ -2693,5 +2722,179 @@ def test_success_code_should_set_in_env_variables_on_cmd_success(self, _):
                 ExtCommandEnvVariable.DisableReturnCode, exit_code) in update_kwargs['message'])
 
 
+@patch('time.sleep', side_effect=lambda _: mock_sleep(0.001))
+class TestCollectExtensionStatus(ExtensionTestCase):
```
vrdmr: Main chunk of tests.

tests/ga/test_exthandlers.py
```diff
@@ -159,8 +159,38 @@ def test_parse_ext_status_should_parse_null_substatus_as_empty(self):
         self.assertTrue(isinstance(extension_status.substatusList, list), 'substatus was not parsed correctly')
         self.assertEqual(0, len(extension_status.substatusList))
 
+    def test_parse_extension_status_with_empty_status(self):
```
vrdmr: Tests for parsing empty status files

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1084,29 +1098,71 @@ def get_status_file_path(self, extension=None):
         return seq_no, path
 
     def collect_ext_status(self, ext):
```
vrdmr: mainset of changes are in the `collect_ext_status` method

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1084,29 +1098,71 @@ def get_status_file_path(self, extension=None):
...
         seq_no, ext_status_file = self.get_status_file_path(ext)
         if seq_no == -1:
             return None
 
+        data = None
+        data_str = None
         ext_status = ExtensionStatus(seq_no=seq_no)
+
+        try:
+            data_str, data = self._read_and_parse_json_status_file(ext_status_file)
```
vrdmr: failing to read/parse json

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1084,29 +1098,71 @@ def get_status_file_path(self, extension=None):
...
+            ext_status.status = ValidHandlerStatus.error
+
+            return ext_status
+
+        # We did not encounter InvalidJsonFile/CouldNotReadStatusFile and thus the status file was correctly written
+        # and has valid json.
         try:
-            data_str = fileutil.read_file(ext_status_file)
-            data = json.loads(data_str)
             parse_ext_status(ext_status, data)
```
vrdmr: parsing errors can raise exceptions from this method.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1084,29 +1098,71 @@ def get_status_file_path(self, extension=None):
...
+        except ExtensionStatusError as e:
+            msg = u"For Extension Handler {0}-{1} for the sequence number {2}, the status file {3}. " \
+                  u"Encountered the following error: {4}".format(ext.name, self.ext_handler.properties.version, seq_no,
+                                                                 ext_status_file, ustr(e))
+            logger.periodic_warn(logger.EVERY_DAY, u"[PERIODIC] " + msg)
+            add_periodic(delta=logger.EVERY_HALF_HOUR, name=ext.name, version=self.ext_handler.properties.version,
+                         op=WALAEventOperation.StatusProcessing, is_success=False, message=msg, log_event=False)
+
+            if e.code == ExtensionStatusError.MaxSizeExceeded:
+                ext_status.message, field_size = self._truncate_message(ext_status.message, _MAX_STATUS_MESSAGE_LENGTH)
```
vrdmr: Handling very large status files in here.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1084,29 +1098,71 @@ def get_status_file_path(self, extension=None):
...
+                ext_status.substatusList = self._process_substatus_list(ext_status.substatusList, field_size)
```
vrdmr: handling substatus lists here.

tests/ga/test_extension.py
```diff
@@ -2848,5 +2722,179 @@ def test_success_code_should_set_in_env_variables_on_cmd_success(self, _):
...
+        ext_status = ext_handler_i.collect_ext_status(extension)
+
+        self.assertEqual(ext_status.code, SUCCESS_CODE_FROM_STATUS_FILE)
+        self.assertEqual(ext_status.configurationAppliedTime, None)
+        self.assertEqual(ext_status.operation, "Enable")
+        self.assertEqual(ext_status.sequenceNumber, 0)
+        self.assertRegex(ext_status.message, "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vestibulum non lacinia urna, sit .*")
+        self.maxDiff = None
+        self.assertEqual(ext_status.status, ValidHandlerStatus.success)
+        self.assertEqual(len(ext_status.substatusList), 1) # NUM OF SUBSTATUS PARSED
```
pgombar: Where exactly do we see that the status was truncated? Can you have a more explicit check of even just write a comment explaining what happens?

For example, in your next test, you specify that the original file has 41 substatuses, but we only capture 12.

narrieta:
Fixes #1784

PR #1770, opened 2020-01-31
Monitor thread errors

narrieta:
Added code to handle exceptions in the monitor thread main loop.

Minor refactoring of the functions called by the main loop to ensure consistency on how they handle exceptions. I decided to leave log_altered_network_configuration alone for now (it is setup as a high-priority operation).

Tested the refactoring changes using the debugger on a live VM.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1770)

### DIFF
--- a/azurelinuxagent/ga/monitor.py
+++ b/azurelinuxagent/ga/monitor.py
@@ -234,11 +234,11 @@ class MonitorHandler(object):
         agent or the extensions. We don't have control over extension's events parameters, but we will override
         any values they might have set for sys_info parameters.
         """
-        if self.last_event_collection is None:
-            self.last_event_collection = datetime.datetime.utcnow() - MonitorHandler.EVENT_COLLECTION_PERIOD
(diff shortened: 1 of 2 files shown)


azurelinuxagent/ga/monitor.py
```diff
@@ -461,29 +465,29 @@ def poll_telemetry_metrics(self):
...
                 metrics = CGroupsTelemetry.poll_all_tracked()
-                self.last_cgroup_polling_telemetry = time_now
 
                 if metrics:
                     for metric in metrics:
                         report_metric(metric.category, metric.counter, metric.instance, metric.value)
         except Exception as e:
             logger.warn("Could not poll all the tracked telemetry due to {0}", ustr(e))
 
+        self.last_cgroup_polling_telemetry = datetime.datetime.utcnow()
```
vrdmr: Good catch.

PR #1777, opened 2020-02-11
Separate goal state from protocol

narrieta:
if an error occurs while _update_from_goal_state is populating the goal state the retry loop may end up producing a goal state that is only partially populated.

This PR addresses that issue and starts to refactor the goal state out of the protocol. I will do further refactoring to improve that separation once we remove the metadata protocol.

The code refactored in this PR is heavily exercised by the current tests. I will add tests specific to the goal state class in my next round of refactoring.

I did manual verification using the debugger on a live VM, and automation reported no errors.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1777)


azurelinuxagent/common/protocol/__init__.py
```diff
@@ -15,7 +15,3 @@
 # Requires Python 2.6+ and Openssl 1.0+
 #
 
-from azurelinuxagent.common.protocol.util import get_protocol_util, \
-                                                 OVF_FILE_NAME, \
-                                                 TAG_FILE_NAME
-
```
narrieta: This import was creating a dependency on protocol.util to code that imported anything under protocol and was leading to a couple of circular dependencies. I remove the statement and updated the code to always import  get_procotol_util from protocol.util instead of protocol.

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -0,0 +1,411 @@
+# Microsoft Azure Linux Agent
```
narrieta: Most of the code here comes from wire.py, but please do a careful review. Thanks!

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -0,0 +1,411 @@
...
+            xml_text = wire_client.fetch_config(uri, wire_client.get_header())
+            self.ext_conf = ExtensionsConfig(xml_text)
+
+        uri = findtext(container, "RemoteAccessInfo")
+        if uri is None:
+            self.remote_access = None
+        else:
+            xml_text = wire_client.fetch_config(uri, wire_client.get_header_for_cert())
+            self.remote_access = RemoteAccess(xml_text)
+
```
narrieta: I'm planning on moving the fetch_* methods below to protocol, and remove the current interface that exposes the goal state from the protocol class. I.e. code would do "goal_state = protocol.fetch_goal_state()" and then use "goal_state" directly instead of using "protocol" to retrieve the properties of the goal state.

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -0,0 +1,411 @@
...
+PEM_FILE_NAME = "Certificates.pem"
+TRANSPORT_CERT_FILE_NAME = "TransportCert.pem"
+TRANSPORT_PRV_FILE_NAME = "TransportPrivate.pem"
+
+# Store the last retrieved container id as an environment variable to be shared between threads for telemetry purposes
+CONTAINER_ID_ENV_VARIABLE = "AZURE_GUEST_AGENT_CONTAINER_ID"
+
+
+class GoalState(object):
+    def __init__(self, wire_client, full_goal_state=False, base_incarnation=None):
```
narrieta: Now the constructor of GoalState has a dependency on the protocol, which makes unit testing a little harder. To compensate for that I created mock_wire_protocol.py in the tests directory)

azurelinuxagent/common/protocol/wire.py
```diff
@@ -20,32 +20,30 @@
...
 from azurelinuxagent.common.utils import fileutil, restutil
 from azurelinuxagent.common.utils.archive import StateFlusher
 from azurelinuxagent.common.utils.cryptutil import CryptUtil
 from azurelinuxagent.common.utils.textutil import parse_doc, findall, find, \
-    findtext, getattrib, gettext, remove_bom, get_bytes_from_pem, parse_json
+    findtext, gettext, remove_bom, get_bytes_from_pem, parse_json
 from azurelinuxagent.common.version import AGENT_NAME, CURRENT_VERSION
 
 VERSION_INFO_URI = "http://{0}/?comp=versions"
-GOAL_STATE_URI = "http://{0}/machine/?comp=goalstate"
```
narrieta: These variables and methods were moved to goal_state.py

azurelinuxagent/common/protocol/wire.py
```diff
@@ -774,42 +710,25 @@ def _update_from_goal_state(self, refresh_type):
...
-                    self.save_cache(os.path.join(conf.get_lib_dir(), INCARNATION_FILE_NAME), new_goal_state.incarnation)
-
-                    update_host_plugin()
+                if self._goal_state is None or refresh_type == WireClient._UpdateType.GoalStateForced:
+                    new_goal_state = GoalState.fetch_full_goal_state(self)
+                else:
+                    new_goal_state = GoalState.fetch_full_goal_state_if_incarnation_different_than(self, self._goal_state.incarnation)
+
+                if new_goal_state is not None:
+                    self._goal_state = new_goal_state
```
narrieta: this fixes the original issue: the entire goal state is updated or no updates are done

azurelinuxagent/common/protocol/wire.py
```diff
@@ -826,35 +745,39 @@ def save_goal_state(incarnation, xml_text):
...
+            def save_if_not_none(goal_state_property, file_name):
+                file_path = os.path.join(conf.get_lib_dir(), file_name)
 
-    def _set_ext_conf(self, new_ext_conf):
-        if new_ext_conf is None:
-            logger.warn("Setting empty Extension Config object!")
-        self._ext_conf = new_ext_conf
+                if goal_state_property is not None and goal_state_property.xml_text is not None:
+                    self.save_cache(file_path, goal_state_property.xml_text)
+                else:  # remove the file from the previous goal state if it exists
```
narrieta: i forgot to remove this else (removing the older files is not needed, since self.goal_state_flusher.flush() above moves the files to the history folder

tests/data/wire/ext_conf.xml
```diff
@@ -3,19 +3,19 @@
     <GAFamily>
       <Name>Prod</Name>
       <Uris>
-          <Uri>http://manifest_of_ga.xml</Uri>
+          <Uri>http://mock-goal-state/manifest_of_ga.xml</Uri>
```
narrieta: see mock_wire_protocol.py for the motivation of these updates in the test data

azurelinuxagent/common/protocol/wire.py
```diff
@@ -826,35 +745,39 @@ def save_goal_state(incarnation, xml_text):
...
 
-    def _set_remote_access(self, new_remote_access):
-        if new_remote_access is None:
-            logger.warn("Setting empty Remote Access object!")
-        self._remote_access = new_remote_access
+            save_if_not_none(self._goal_state, GOAL_STATE_FILE_NAME.format(self._goal_state.incarnation))
+            save_if_not_none(self._goal_state.hosting_env, HOSTING_ENV_FILE_NAME)
+            save_if_not_none(self._goal_state.shared_conf, SHARED_CONF_FILE_NAME)
+            save_if_not_none(self._goal_state.ext_conf, EXT_CONF_FILE_NAME.format(self._goal_state.incarnation))
+            save_if_not_none(self._goal_state.remote_access, REMOTE_ACCESS_FILE_NAME.format(self._goal_state.incarnation))
```
larohra: Just curious, why is "Certificates.xml" handled differently in the goal_state.py file whereas all these other files are handled here?

azurelinuxagent/common/protocol/wire.py
larohra: Thanks for clarifying that! This did confuse me a bit

azurelinuxagent/common/protocol/wire.py
```diff
@@ -867,29 +790,29 @@ def get_goal_state(self):
         return self._goal_state
 
     def get_hosting_env(self):
-        if self._hosting_env is None:
+        if self._goal_state is None:
```
larohra: shouldn't this be `self._goal_state.hosting_env` rather than just `self._goal_state`?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -867,29 +790,29 @@ def get_goal_state(self):
...
             raise ProtocolError("Trying to fetch Hosting Environment before initialization!")
-        return self._hosting_env
+        return self._goal_state.hosting_env
 
     def get_shared_conf(self):
-        if self._shared_conf is None:
+        if self._goal_state is None:
```
larohra: Same as above

azurelinuxagent/common/protocol/wire.py
```diff
@@ -867,29 +790,29 @@ def get_goal_state(self):
...
             raise ProtocolError("Trying to fetch Shared Conf before initialization!")
-        return self._shared_conf
+        return self._goal_state.shared_conf
 
     def get_certs(self):
         # This property can be None in the GoalState, so not returning a ProtocolError here
-        return self._certs
+        return self._goal_state.certs
```
larohra: This will throw an error if `self._goal_state` is None, we should handle that case

tests/data/wire/ext_conf.xml
larohra: Thanks for creating that file, really helpful!

azurelinuxagent/common/protocol/wire.py
narrieta: Yes, I forgot to add a comment here.

Certificates doesn't hold a reference to xml_text. I did not add one since I'm guessing the original intention was not to hold the data on memory for too long.

azurelinuxagent/common/protocol/wire.py
narrieta: Thanks, will update it.

narrieta: The intention of this check was to ensure that update_goal_state was called before retrieving any of the properties of the goal state, wasn't it?

azurelinuxagent/common/protocol/wire.py
narrieta: Thanks, added

tests/protocol/mock_wire_protocol.py
```diff
@@ -0,0 +1,83 @@
...
+
+    try:
+        # To minimize the impact of mocking restutil.http_get we only use the mock data for requests
+        # to the wireserver or requests starting with "mock-goal-state"
+        original_http_get = http_get
+
+        wire_server_endpoint = 'http://{0}'.format(KNOWN_WIRESERVER_IP)
+
+        def mock_http_get(url, *args, **kwargs):
+            if not (url.startswith(wire_server_endpoint) or url.startswith('http://mock-goal-state/') or url.startswith('https://mock-goal-state/')):
```
narrieta: this should be case insensitive

azurelinuxagent/common/protocol/wire.py
larohra: Ohh sorry I missed the raise in the next line. When I introduced this code we werent raising error rather just logging a warning which is what I thought was still happening here. Thanks for clarifying!

tests/protocol/test_hostplugin.py
```diff
@@ -237,355 +243,326 @@ def test_ensure_initialized(self):
...
+            wire_protocol.client.upload_status_blob()
 
-        # act
-        wire_protocol.client.upload_status_blob()
+            # assert direct route is called
+            self.assertEqual(1, patch_upload.call_count, "Direct channel was not used")
 
-        # assert direct route is called
-        self.assertEqual(1, patch_upload.call_count, "Direct channel was not used")
+            # assert host plugin route is called
```
pgombar: Should be "host plugin route is **NOT** called"?

PR #1791, opened 2020-02-26
Populate telemetry events at creation time

narrieta:
Replaces #1757

Updates: 

- Instead of patching old agent events at startup, it patches them at collection time
- To distinguish between events from old agents and current agent, saves telemetry files as *.waagent.tld instead of *.tld
- Initializes common fields in the daemon (they were initialized only on the extension handler)
- Merges SysInfo into EventLogger
- Uses a static member (GoalState.ContainerID) instead of an environment variable to keep the latest container ID
- Simplifies MonitorHandle.collect_and_send_events and improves error handling collecting events
- Removes the call to update_goal_state() from the intialization of the MonitorHandle
- Internal refactoring of the logic that adds common parameters to events; merged a few methods together
- Many test improvements to avoid testing via private implementation details

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1791)


azurelinuxagent/common/event.py
```diff
@@ -18,30 +18,34 @@
...
+from azurelinuxagent.common.protocol.imds import get_imds_client
+
+EVENTS_DIRECTORY = "events"
 
 _EVENT_MSG = "Event: name={0}, op={1}, message={2}, duration={3}"
 TELEMETRY_EVENT_PROVIDER_ID = "69B669B9-4AF8-4C50-BDC4-6006FA76E975"
 TELEMETRY_METRICS_EVENT_ID = 4
 
-# Store the last retrieved container id as an environment variable to be shared between threads for telemetry purposes
-CONTAINER_ID_ENV_VARIABLE = "AZURE_GUEST_AGENT_CONTAINER_ID"
```
narrieta: Now this is in GoalState.ContainerID

azurelinuxagent/common/event.py
```diff
@@ -53,15 +57,13 @@
 
 MAX_NUMBER_OF_EVENTS = 1000
 
+EVENT_FILE_EXTENSION = '.waagent.tld'
```
narrieta: Events produced by >= 2.2.47 will be saved as *.waagent.tld

azurelinuxagent/common/event.py
```diff
@@ -179,6 +181,53 @@ def _save(self):
     ]
 
 
+def parse_json_event(data_str):
```
narrieta: These were previously on the monitor handle. Note that I change the exception handling in parse_event.

azurelinuxagent/common/event.py
```diff
@@ -229,6 +278,97 @@ def __init__(self):
...
+            imds_client = get_imds_client(protocol.get_endpoint())
+            imds_info = imds_client.get_compute()
+            parameters['Location'].value = imds_info.location
+            parameters['SubscriptionId'].value = imds_info.subscriptionId
+            parameters['ResourceGroupName'].value = imds_info.resourceGroupName
+            parameters['VMId'].value = imds_info.vmId
+            parameters['ImageOrigin'].value = int(imds_info.image_origin)
+        except Exception as e:
+            logger.warn("Failed to get IMDS info; will be missing from telemetry: {0}", ustr(e))
+        logger.info('')
```
narrieta: will remove this empty message

azurelinuxagent/common/event.py
```diff
@@ -272,37 +412,33 @@ def is_period_elapsed(self, delta, h):
...
         event = TelemetryEvent(event_id, TELEMETRY_EVENT_PROVIDER_ID)
 
         event.parameters.append(TelemetryEventParam('Name', str(name)))
         event.parameters.append(TelemetryEventParam('Version', str(version)))
-        event.parameters.append(TelemetryEventParam('IsInternal', bool(is_internal)))
         event.parameters.append(TelemetryEventParam('Operation', str(op)))
         event.parameters.append(TelemetryEventParam('OperationSuccess', bool(is_success)))
         event.parameters.append(TelemetryEventParam('Message', str(message)))
         event.parameters.append(TelemetryEventParam('Duration', int(duration)))
-        event.parameters.append(TelemetryEventParam('ExtensionType', str(evt_type)))
```
narrieta: this is now populated in _add_common_event_parameters

azurelinuxagent/common/event.py
```diff
@@ -393,52 +529,106 @@ def _clean_up_message(message):
...
+                         TelemetryEventParam('KeywordName', ''),
+                         TelemetryEventParam('ExtensionType', event.file_type),
+                         TelemetryEventParam('IsInternal', False)]
+
+        event.parameters.extend(common_params)
+        event.parameters.extend(self._common_parameters)
+
     @staticmethod
-    def add_default_parameters_to_event(event_parameters, set_values_for_agent=True):
+    def _trim_extension_event_parameters(event):
```
narrieta: This comes from #1757; we need to improve the way we handle parameters to avoid all those sequential traversals of the array

azurelinuxagent/common/event.py
```diff
@@ -393,52 +529,106 @@ def _clean_up_message(message):
...
-        return parameters
+                try:
+                    logger.verbose("Processing event file: {0}", event_file_path)
+
+                    with open(event_file_path, "rb") as fd:
+                        event_data = fd.read().decode("utf-8")
+
+                    event = parse_event(event_data)
+
+                    is_legacy_event = match.group('agent_event') is None
```
narrieta: I'll add a comment here to explain why i mean with "legacy" (older agent/extensions)

azurelinuxagent/common/event.py
```diff
@@ -393,52 +529,106 @@ def _clean_up_message(message):
...
+                    os.remove(event_file_path)
+            except Exception as e:
+                logger.warn("Failed to process event file {0}: {1}", event_file, ustr(e))
+
+        return event_list
+
+    def _update_old_event_schema(self, event, event_creation_time):
+        # Ensure that if an agent event is missing a field from the schema defined since 2.2.47, the missing fields
+        # will be appended, ensuring the event schema is complete before the event is reported.
+        new_event = TelemetryEvent()
```
narrieta: This comes from #1757. I discussed this @pgombar - we should not just add any missing fields since we may hide bugs elsewhere or populate with invalid values. Some missing fields must be added, though. For example, OpCode (timestamp) will be in 2.2.46, but not on previous agents.

azurelinuxagent/common/protocol/imds.py
```diff
@@ -28,8 +27,8 @@
 IMDS_INTERNAL_SERVER_ERROR = 3
 
 
-def get_imds_client():
-    return ImdsClient()
+def get_imds_client(wireserver_endpoint):
```
narrieta: The IMDS client had a dependency on the protocol just to get the endpoint; Instead, now we pass the endpoint as parameter.

azurelinuxagent/common/protocol/restapi.py
```diff
@@ -27,13 +27,11 @@ class VMInfo(DataContract):
     def __init__(self,
                  subscriptionId=None,
                  vmName=None,
-                 containerId=None,
```
narrieta: This should have been removed when we move the container ID to the env variable

azurelinuxagent/common/telemetryevent.py
```diff
@@ -34,11 +35,27 @@ def __init__(self, eventId=None, providerId=None):
         self.eventId = eventId
         self.providerId = providerId
         self.parameters = DataContractList(TelemetryEventParam)
+        self.file_type = ""
 
     # Checking if the particular param name is in the TelemetryEvent.
     def __contains__(self, param_name):
         return param_name in [param.name for param in self.parameters]
 
+    def is_extension_event(self):
```
narrieta: From #1757 - Needs improvement to avoid traversing the array

azurelinuxagent/daemon/main.py
```diff
@@ -155,10 +162,7 @@ def daemon(self, child_args=None):
                 # - Changes to RDMA state may not increment the goal state
                 #   incarnation number. A forced update ensures the most
                 #   current values.
-                protocol = self.protocol_util.get_protocol()
-                if type(protocol) is not WireProtocol:
```
narrieta: Metadata is gone, so the check is no longer needed

azurelinuxagent/ga/monitor.py
```diff
@@ -155,9 +104,12 @@ def init_protocols(self):
         self.protocol_util = get_protocol_util()
         self.protocol = self.protocol_util.get_protocol()
         # Update the GoalState first time to instantiate all required objects for the monitor thread
-        self.protocol.update_goal_state()
```
narrieta: No longer needed. This removes the last known race condition updating the goal state

azurelinuxagent/ga/monitor.py
```diff
@@ -155,9 +104,12 @@ def init_protocols(self):
         self.protocol_util = get_protocol_util()
         self.protocol = self.protocol_util.get_protocol()
         # Update the GoalState first time to instantiate all required objects for the monitor thread
```
narrieta: will remove this comment

azurelinuxagent/ga/monitor.py
```diff
@@ -167,106 +119,22 @@ def start(self, init_data=False):
         self.event_thread.setName("MonitorHandler")
         self.event_thread.start()
 
-    def init_sysinfo(self):
```
narrieta: moved to EventLogger

azurelinuxagent/ga/monitor.py
```diff
@@ -167,106 +119,22 @@ def start(self, init_data=False):
...
-                                                    vminfo.resourceGroupName))
-            self.sysinfo.append(TelemetryEventParam('VMId',
-                                                    vminfo.vmId))
-            self.sysinfo.append(TelemetryEventParam('ImageOrigin',
-                                                    int(vminfo.image_origin)))
-        except (HttpError, ValueError) as e:
-            logger.warn("failed to get IMDS info: {0}", ustr(e))
-
-    @staticmethod
-    def collect_event(evt_file_name):
```
narrieta: moved to EventLogger

tests/common/events
```diff
@@ -0,0 +1,30 @@
+{
```
narrieta: will remove this file

tests/common/mocksysinfo.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2020 Microsoft Corporation
```
narrieta: Need to remove this file

tests/common/test_event.py
```diff
@@ -14,84 +15,111 @@
...
 from tests.protocol import mockwiredata, mock_wire_protocol
-from tests.tools import AgentTestCase, data_dir, Mock, patch
+from azurelinuxagent.common.version import CURRENT_AGENT, CURRENT_VERSION, AGENT_EXECUTION_MODE
+from azurelinuxagent.common.osutil import get_osutil
+from tests.tools import AgentTestCase, data_dir, load_data, Mock, patch, skip_if_predicate_true
+from tests.utils.event_logger_tools import EventLoggerTools
 
 
 class TestEvent(AgentTestCase):
-    def test_add_event_should_read_container_id_from_process_environment(self):
```
narrieta: replaced by test_add_event_should_use_the_container_id_from_the_most_recent_goal_state

tests/common/test_event.py
```diff
@@ -270,107 +298,81 @@ def test_periodic_emits_after_elapsed_delta(self, mock_event):
...
     def test_save_event(self):
         add_event('test', message='test event')
-        self.assertTrue(len(os.listdir(self.tmp_dir)) == 1)
+        self.assertTrue(len(os.listdir(self.event_dir)) == 1)
 
         # checking the extension of the file created.
-        for filename in os.listdir(self.tmp_dir):
-            self.assertEqual(".tld", filename[-4:])
-
-    def test_save_event_message_with_non_ascii_characters(self):
```
narrieta: included as a test case in test_collect_events_should_ignore_invalid_event_files

tests/common/test_event.py
```diff
@@ -270,107 +298,81 @@ def test_periodic_emits_after_elapsed_delta(self, mock_event):
...
-                    self.assertEqual(i['value'], 'HelloWorldTask')
-                elif i['name'] == 'KeywordName':
-                    self.assertEqual(i['value'], '')
-                elif i['name'] == 'GAVersion':
-                    self.assertEqual(i['value'], str(CURRENT_AGENT))
-                else:
-                    self.assertFalse(True, "Contains a field outside the defaults expected. Field Name: {0}".
-                                     format(i['name']))
-
-    def test_save_event_message_with_decode_errors(self):
```
narrieta: included as a test case in test_collect_events_should_ignore_invalid_event_files

tests/common/test_event.py
```diff
@@ -437,107 +439,214 @@ def test_elapsed_milliseconds(self):
         utc_start = datetime.utcnow() + timedelta(days=1)
         self.assertEqual(0, elapsed_milliseconds(utc_start))
 
-    def _assert_default_params_get_correctly_added(self, param_list_actual, parameters_expected):
```
narrieta: replaced by _assert_event_includes_all_parameters_in_the_telemetry_schema

tests/common/test_event.py
```diff
@@ -437,107 +439,214 @@ def test_elapsed_milliseconds(self):
...
-        counter = 0
-        for p in default_parameters_expected_names:
-            self.assertIn(p, param_list_dict)
-            self.assertEqual(param_list_dict[p], parameters_expected[p])
-            counter += 1
-
-        self.assertEqual(len(default_parameters_expected_names), counter)
-
-    @patch("azurelinuxagent.common.event.get_container_id_from_env", return_value="TEST_CONTAINER_ID")
-    def test_add_default_parameters_to_extension_event(self, *args):
```
narrieta: replaced by test_collect_events_should_add_all_the_parameters_in_the_telemetry_schema_to_extension_events

tests/common/test_event.py
```diff
@@ -437,107 +439,214 @@ def test_elapsed_milliseconds(self):
...
-        extension_param_list_with_defaults = EventLogger.add_default_parameters_to_event(extension_param_list_populated,
-                                                                                         set_values_for_agent=False)
-        self._assert_default_params_get_correctly_added(extension_param_list_with_defaults,
-                                                        parameters_expected)
-
-    @patch("threading.Thread.getName", return_value="HelloWorldTask")
-    @patch('os.getpid', return_value=42)
-    @patch("azurelinuxagent.common.event.get_container_id_from_env", return_value="TEST_CONTAINER_ID")
-    @patch("azurelinuxagent.common.event.datetime")
-    def test_add_default_parameters_to_agent_event(self, patch_datetime, *args):
```
narrieta: replaced by test_add_event_should_create_events_that_have_all_the_parameters_in_the_telemetry_schema (and its sibling tests for the other methods to create an event)

tests/common/test_event.py
```diff
@@ -558,35 +667,6 @@ def test_report_metric(self, mock_event):
         else:
             self.fail("Counter '%idle' not found in event parameters: {0}".format(repr(event_dictionary)))
 
-    def test_save_metric(self):
```
narrieta: replaced by test_report_metric_should_create_events_that_have_all_the_parameters_in_the_telemetry_schema

tests/common/test_logger.py
```diff
@@ -237,35 +237,6 @@ def test_telemetry_logger(self):
 
         self.assertEqual(5, mock.call_count)  # Only ERROR should be called.
 
-    @patch("azurelinuxagent.common.event.send_logs_to_telemetry", return_value=True)
-    @patch('azurelinuxagent.common.event.EventLogger.save_event')
-    def test_telemetry_logger_sending_correct_fields(self, mock_save, patch_conf_get_logs_to_telemetry):
```
narrieta: replaced by test_add_log_event_should_create_events_that_have_all_the_parameters_in_the_telemetry_schema

tests/data/events/collect_and_send_events_invalid_data/1560752429123264-1.tld
```diff
@@ -1,32 +0,0 @@
- {
```
narrieta: those files are replaced with new test cases

tests/data/ext/event_from_agent.json
```diff
@@ -1 +1,28 @@
-{"eventId": 1, "providerId": "69B669B9-4AF8-4C50-BDC4-6006FA76E975", "parameters": [{"name": "Name", "value": "WALinuxAgent"}, {"name": "Version", "value": "2.2.44"}, {"name": "IsInternal", "value": false}, {"name": "Operation", "value": "ProcessGoalState"}, {"name": "OperationSuccess", "value": true}, {"name": "Message", "value": "Incarnation 12"}, {"name": "Duration", "value": 16610}, {"name": "ExtensionType", "value": ""}, {"name": "GAVersion", "value": "WALinuxAgent-2.2.44"}, {"name": "ContainerId", "value": "TEST-CONTAINER-ID-ALREADY-PRESENT-GUID"}, {"name": "OpcodeName", "value": "2019-11-02 01:42:49.188030"}, {"name": "EventTid", "value": 140240384030528}, {"name": "EventPid", "value": 108573}, {"name": "TaskName", "value": "ExtHandler"}, {"name": "KeywordName", "value": ""}]}
\ No newline at end of file
+{"eventId": 1, "providerId": "69B669B9-4AF8-4C50-BDC4-6006FA76E975", "parameters": [{"name": "Name", "value": "dummy_name"},
```
narrieta: VS Code messed the formatting of these 2 files. Will see if I can improve it

tests/ga/test_monitor.py
```diff
@@ -118,238 +79,14 @@ def setUp(self):
...
         logger.DEFAULT_LOGGER = Logger(prefix=prefix)
+
         # Since ProtocolUtil is a singleton per thread, we need to clear it to ensure that the test cases do not
         # reuse a previous state
         clear_singleton_instances(ProtocolUtil)
 
     def tearDown(self):
         AgentTestCase.tearDown(self)
 
-    def test_parse_xml_event(self, *args):
```
narrieta: moved to test_event

tests/ga/test_monitor.py
```diff
@@ -118,238 +79,14 @@ def setUp(self):
...
-            elif p.name == TaskName_param:
-                self.assertEqual(TaskName_value, p.value)
-                counter += 1
-            elif p.name == KeywordName_param:
-                self.assertEqual(KeywordName_value, p.value)
-                counter += 1
-
-        self.assertEqual(12, counter)
-
-    def test_add_sysinfo_should_honor_sysinfo_values_from_agent_for_extension_events(self, *args):
```
narrieta: no longer needed

tests/ga/test_monitor.py
```diff
@@ -118,238 +79,14 @@ def setUp(self):
...
-                self.assertEqual(TaskName_value, p.value)
-                counter += 1
-            elif p.name == KeywordName_param:
-                self.assertEqual(KeywordName_value, p.value)
-                counter += 1
-
-        self.assertEqual(12, counter)
-        os.environ.pop(CONTAINER_ID_ENV_VARIABLE)
-
-    def test_main_loop_should_handle_exceptions_retrieving_goal_state(self, *_):
```
narrieta: removed by #1757

tests/tools.py
```diff
@@ -333,6 +334,23 @@ def emulate_raises_regex(self, exception_type, regex, function, *args, **kwargs)
                     exception_type, regex, str(e)))
         self.fail("No exception was thrown.  Expected exception {0} matching {1}".format(exception_type, regex))
 
+    def emulate_assertDictEqual(self, first, second, msg=None):
```
narrieta: A bare bones implementation not even close to the assert in >= 2.7 (in terms of debugging test failures). We can improve it as needed.

tests/common/test_event.py
```diff
@@ -270,107 +298,81 @@ def test_periodic_emits_after_elapsed_delta(self, mock_event):
...
-        for filename in os.listdir(self.tmp_dir):
-            self.assertEqual(".tld", filename[-4:])
-
-    def test_save_event_message_with_non_ascii_characters(self):
-        test_data_dir = os.path.join(data_dir, "events", "collect_and_send_extension_stdout_stderror")
-        msg = ""
-
-        with open(os.path.join(test_data_dir, "dummy_stdout_with_non_ascii_characters"), mode="r+b") as stdout:
-            with open(os.path.join(test_data_dir, "dummy_stderr_with_non_ascii_characters"), mode="r+b") as stderr:
-                msg = read_output(stdout, stderr)
```
narrieta: This test was passing for the wrong reason. It beats me why we were going thru read_output to prepare the test case, but by doing so an error was happening during the call to read_output and the string in the test event was actually the error message from that exception.

azurelinuxagent/common/event.py
```diff
@@ -393,52 +529,106 @@ def _clean_up_message(message):
...
                 return message
 
+    def _add_common_event_parameters(self, event, event_timestamp):
+        """
+        This method is called for all events and ensures all telemetry fields are added before the event is sent out.
+        Note that the event timestamp is saved in the OpcodeName field.
+        """
+        common_params = [TelemetryEventParam('GAVersion', CURRENT_AGENT),
+                         TelemetryEventParam('ContainerId', GoalState.ContainerID),
+                         TelemetryEventParam('OpcodeName', event_timestamp.strftime(u'%Y-%m-%dT%H:%M:%S.%fZ')),
```
narrieta: #1757 changed the format of the timestamp. I talked to @pgombar about this and she confirmed it was not intentional. I was going to revert to the previous format but 2.2.46 was just converting the datetime to string, with no formatting. This new format is safer, and plays well in Kusto queries, so I left it alone. Just be aware of the change from 2.2.46 in case it ever matters.

tests/tools.py
```diff
@@ -333,6 +334,23 @@ def emulate_raises_regex(self, exception_type, regex, function, *args, **kwargs)
...
+
+        for k in first.keys():
+            if k not in second:
+                fail("'{0}' is missing from second".format(k))
+            if first[k] != second[k]:
+                fail("'{0}' != '{1}' (key: {2})".format(first[k], second[k], k))
+            found.add(k)
+
+        for k in second.keys():
+            if k not in first:
```
narrieta: this should be 'if k not in found"

azurelinuxagent/common/event.py
```diff
@@ -393,52 +529,106 @@ def _clean_up_message(message):
...
+
+                        if event.is_extension_event():
+                            EventLogger._trim_extension_event_parameters(event)
+                            self._add_common_event_parameters(event, event_file_creation_time)
+                        else:
+                            self._update_old_event_schema(event, event_file_creation_time)
+
+                    event_list.events.append(event)
+                finally:
+                    os.remove(event_file_path)
```
pgombar: Suggestion: before this change, we had logged when we were done with an event (basically, when we deleted it):

> logger.verbose("Processed event file: {0}", evt_file_name)

I found it useful when debugging.

azurelinuxagent/common/event.py
```diff
@@ -393,52 +529,106 @@ def _clean_up_message(message):
...
+                    is_legacy_event = match.group('agent_event') is None
+
+                    if is_legacy_event:
+                        # We'll use the file creation time for the event's timestamp
+                        event_file_creation_time_epoch = os.path.getmtime(event_file_path)
+                        event_file_creation_time = datetime.fromtimestamp(event_file_creation_time_epoch)
+
+                        if event.is_extension_event():
+                            EventLogger._trim_extension_event_parameters(event)
+                            self._add_common_event_parameters(event, event_file_creation_time)
```
pgombar: We would want to modify extension events always, right? It doesn't matter if it's a legacy event or not.

azurelinuxagent/common/event.py
```diff
@@ -393,52 +529,106 @@ def _clean_up_message(message):
...
+        new_event.parameters = []
+        self._add_common_event_parameters(new_event, event_creation_time)
+
+        event_params = dict([(param.name, param.value) for param in event.parameters])
+        new_event_params = dict([(param.name, param.value) for param in new_event.parameters])
+
+        missing_params = set(new_event_params.keys()).difference(set(event_params.keys()))
+        params_to_add = []
+        for param_name in missing_params:
+            params_to_add.append(TelemetryEventParam(param_name, new_event_params[param_name]))
```
pgombar: Didn't we discuss not adding the (potentially wrong) value for missing telemetry fields? We can safely populate `OpcodeName` as it's a run-time value, but not others like `ContainerId` or `GAVersion`. This was a bug in #1757.

azurelinuxagent/common/event.py
```diff
@@ -393,52 +529,106 @@ def _clean_up_message(message):
...
+            if param.name in params_to_keep:
+                trimmed_params.append(param)
 
-        # Converting the event_parameters into a dictionary as it helps to easily look up and get values
-        param_names = OrderedDict([(param.name, param.value) for param in event_parameters])
+        event.parameters = trimmed_params
+
+    def collect_events(self):
+        """
+        Retuns a list of events that need to sent to the telemetry pipeline and deletes the corresponding files
```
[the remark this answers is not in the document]
narrieta: thanks, fixed

azurelinuxagent/common/event.py
narrieta: this is related to the comment I posted above, on line 595 - I need to clarify what I mean by "legacy". This is the comment I added: 

>                   # "legacy" events are events produced by previous versions of the agent (<= 2.2.46) and extensions;
>                   # they do not include all the telemetry fields, so we add them here

azurelinuxagent/common/event.py
narrieta: Yes, this was intentional. We already log "Processing event file:" at line 588 so having a "I'm starting" and a "I completed" for each event is way too verbose.

azurelinuxagent/common/event.py
narrieta: Yes, that is the comment I posted at line 619. Looking forward for your PR fixing this (as we discussed) :)

tests/tools.py
narrieta: Actually 'found' is not needed, since if we get here then all the keys in first are also in second. Will delete it.

azurelinuxagent/common/event.py
pgombar: But my point is that all extension events need to have common parameters added, not just legacy ones? Why is ` self._add_common_event_parameters(event, event_file_creation_time)` only called for legacy extension events?

azurelinuxagent/common/event.py
pgombar: After our discussion, my understanding was that you weren't going to keep this logic in. Ok, will patch in a future PR.

azurelinuxagent/common/event.py
narrieta: all extension events are legacy events according to my definition.
> is_legacy_event = match.group('agent_event') is None
only (new) agent event events are not legacy.

pgombar: Ah, got it now. Thanks.

tests/tools.py
pgombar: Thanks for adding this!

tests/ga/test_monitor.py
```diff
@@ -546,339 +283,236 @@ def test_reset_loggers_ensuring_timestamp_gets_updated(self, *args):
...
+                               version=CURRENT_VERSION,
+                               message="DummyMessage"):
+        event_data = TestEventMonitoring._get_event_data(name=size if size != 0 else name,
+                op=op,
+                is_success=is_success,
+                duration=duration,
+                version=version,
+                message=random_generator(size) if size != 0 else message)
+        event_file = os.path.join(self.event_dir, "{0}.tld".format(int(time.time() * 1000000)))
+        with open(event_file, 'wb+') as fd:
```
pgombar: Could we use `save_event` here? It handles the naming and `event_dir` creation if needed.

tests/ga/test_monitor.py
```diff
@@ -546,339 +283,236 @@ def test_reset_loggers_ensuring_timestamp_gets_updated(self, *args):
...
+            fd.write(event_data.encode('utf-8'))
+
+    @staticmethod
+    def _get_event_data(duration, is_success, message, name, op, version, eventId=1):
```
pgombar: PyCharm identified that this method is nearly identical to `tests.protocol.test_wire.get_event`. In fact, the one in `test_wire` is outdated since we are also adding the `IsInternal` parameter which is removed with this change. Can you please merge the two methods?

tests/ga/test_monitor.py
narrieta: save_event should be a private of EventLogger, don't want to introduce dependencies on private members

narrieta: Also, see my comment below

tests/ga/test_monitor.py
narrieta: I had to deal with so many test breaks that I did not cleanup all the tests and some of them (like this one) I only modified to allow them to pass. The original code was creating a weird event that did not follow the schema for an extension or an agent event, but was saving it using the "*.tld" extension so it looked to the new code as an extension event, causing many of these tests to break. Had it been done using the public interface (add_event, for example) the internal refactoring would have not caused any issues with it.

As we do improvements on the existing code we are going to have to decide at what point to stop. I had already done a fair amount of work investigating test failures and rewriting many of them and I did not go further on this one on specific. Each one of us will have to make that call. I stopped here.

When writing new code though, we should make every effort to avoid this kind of issues, though. I believe these tests are relatively recent, but to be fair we have not been paying enough attention to design as a team. Let's try to improve from now on.

tests/common/test_event.py
```diff
@@ -270,107 +298,81 @@ def test_periodic_emits_after_elapsed_delta(self, mock_event):
...
+
+        self.assertEquals(len(event_list.events), 1)
+        self.assertEquals(TestEvent._get_event_message(event_list.events[0]), u'World\u05e2\u05d9\u05d5\u05ea \u05d0\u05d7\u05e8\u05d5\u05ea\u0906\u091c')
+
+    def test_collect_events_should_ignore_invalid_event_files(self):
+        self._create_test_event_file("custom_script.tld")
+        self._create_test_event_file("custom_script_utf-16.tld")
+        self._create_test_event_file("custom_script_invalid_json.tld")
+        os.chmod(self._create_test_event_file("custom_script_no_read_access.tld"), 0o200)
+        self._create_test_event_file("custom_script_valid_json.tld")
```
pgombar: What's the difference between `custom_script_valid_json.tld` and `custom_script.tld`?

tests/common/test_event.py
```diff
@@ -270,107 +298,81 @@ def test_periodic_emits_after_elapsed_delta(self, mock_event):
...
+        with patch("azurelinuxagent.common.event.logger.warn") as mock_warn:
+            event_list = event.collect_events()
+
+            self.assertEquals(len(event_list.events), 2)
+            self.assertTrue(
+                all(TestEvent._get_event_message(evt) == "A test telemetry message." for evt in event_list.events),
+                "The valid events were not found")
+
+            # expected_warn_message = r'Failed to process event file.*custom_script_with_invalid_json.tld.*Error parsing event'
+            # self.assertIsNotNone(re.search(expected_warn_message, str(mock_warn.call_args[0])))
```
pgombar: Remove?

tests/common/test_event.py
narrieta: none, i just want 2 valid events -- i'll rename to custom-script-1 & 2

tests/common/test_event.py
narrieta: will do

narrieta: actually it is an assert i forgot to put back. will do

tests/common/test_event.py
```diff
@@ -437,107 +439,214 @@ def test_elapsed_milliseconds(self):
...
+                'ExtensionType': 'json'},
+            assert_timestamp=lambda timestamp:
+                self.assertEquals(timestamp, event_creation_time, "The event timestamp (opcode) is incorrect")
+            )
+
+    def test_collect_events_should_add_all_the_parameters_in_the_telemetry_schema_to_extension_events(self):
+        self._assert_extension_event_includes_all_parameters_in_the_telemetry_schema('custom_script.tld')
+
+    def test_collect_events_should_ignore_extra_parameters_in_extension_events(self):
+        self._assert_extension_event_includes_all_parameters_in_the_telemetry_schema('custom_script_extra_parameters.tld')
```
pgombar: We should also test extension events that are XML, not just JSON. The original need for this PR was because we were incorrectly handling XML extension events, because they already had `ContainerId` present, but empty, which we honored instead of overriding the value.

There is an existing (real) VMAccess XML event in `\tests\data\ext\event_from_extension.xml`.

pgombar: The filename `event_from_extension.xml` should be updated too, since extension events are `.tld`, regardless of if they're XML or JSON.

azurelinuxagent/common/event.py
larohra: NIT: Rename this to `AGENT_EVENT_FILE_EXTENSION`

azurelinuxagent/common/event.py
```diff
@@ -53,15 +58,13 @@
 
 MAX_NUMBER_OF_EVENTS = 1000
 
+EVENT_FILE_EXTENSION = '.waagent.tld'
+EVENT_FILE_REGEX = re.compile(r'(?P<agent_event>\.waagent)?\.tld$')
```
larohra: Same, rename this to `AGENT_EVENT_FILE_REGEX` for clarity

azurelinuxagent/common/event.py
narrieta: yup, thanks!

azurelinuxagent/common/event.py
narrieta: this one is different: it matches all event files (new agent, old agent and extensions)

tests/common/test_event.py
narrieta: Discussed offline - this is covered in the test case using custom_script_extra_parameters.tld

PR #1800, opened 2020-03-06
Enable WireClient tests

narrieta:
Re-enables the tests disabled by #1799 

I'll annotate the PR with the relevant changes.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1800)

### DIFF
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -520,11 +520,6 @@ class WireClient(object):
         logger.info("Wire server endpoint:{0}", endpoint)
         self._endpoint = endpoint
         self._goal_state = None
-        self._hosting_env = None
-        self._shared_conf = None
-        self._remote_access = None
-        self._certs = None
-        self._ext_conf = None
         self._host_plugin = None
         self.status_blob = StatusBlob(self)
         self.goal_state_flusher = StateFlusher(conf.get_lib_dir())
--- a/tests/common/test_event.py
+++ b/tests/common/test_event.py
@@ -28,7 +28,8 @@ from azurelinuxagent.common.event import add_event, add_periodic, add_log_event,
     WALAEventOperation, parse_xml_event, parse_json_event, AGENT_EVENT_FILE_EXTENSION, EVENTS_DIRECTORY
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.protocol.goal_state import GoalState
-from tests.protocol import mockwiredata, mock_wire_protocol
+from tests.protocol import mockwiredata
+from tests.protocol.mocks import mock_wire_protocol
 from azurelinuxagent.common.version import CURRENT_AGENT, CURRENT_VERSION, AGENT_EXECUTION_MODE
 from azurelinuxagent.common.osutil import get_osutil
 from tests.tools import AgentTestCase, data_dir, load_data, Mock, patch, skip_if_predicate_true
@@ -98,7 +99,7 @@ class TestEvent(AgentTestCase):
 
             self.fail("Could not find Contained ID on event")
 
-        with mock_wire_protocol.create(mockwiredata.DATA_FILE) as protocol:
+        with mock_wire_protocol(mockwiredata.DATA_FILE) as protocol:
             contained_id = create_event_and_return_container_id()
             # The expect value comes from DATA_FILE
             self.assertEquals(contained_id, 'c6d5526c-5ac2-4200-b6e2-56f2b70c5ab2', "Incorrect container ID")
--- a/tests/protocol/mock_wire_protocol.py
+++ b/tests/protocol/mock_wire_protocol.py
@@ -18,7 +18,7 @@
 import contextlib
 import re
 from azurelinuxagent.common.protocol.wire import WireProtocol
-from azurelinuxagent.common.utils.restutil import KNOWN_WIRESERVER_IP, http_get
+from azurelinuxagent.common.utils.restutil import KNOWN_WIRESERVER_IP, http_request, DEFAULT_RETRIES, RETRY_CODES, DELAY_IN_SECONDS
 from tests.tools import patch
 from tests.protocol.mockwiredata import WireProtocolData
 
@@ -28,21 +28,22 @@ def create(mock_wire_data_file):
     Creates a mock WireProtocol object that will return the data specified by 'mock_wire_data_file' (which must
     follow the structure of the data files defined in tests/protocol/mockwiredata.py).
 
-    NOTE: This function creates mocks for azurelinuxagent.common.utils.restutil.http_get and
+    NOTE: This function creates mocks for azurelinuxagent.common.utils.restutil.http_request and
           azurelinuxagent.common.protocol.wire.CryptUtil. These mocks can be stopped using the
           methods stop_mock_http_get() and stop_mock_crypt_util().
 
     The return value is an instance of WireProtocol augmented with these properties/methods:
 
         * mock_wire_data - the WireProtocolData constructed from the mock_wire_data_file parameter.
-        * stop_mock_http_get() - stops the mock for restutil.http_get
+        * stop_mock_http_request() - stops the mock for restutil.http_request
         * stop_mock_crypt_util() - stops the mock for CrypUtil
+        * stop() - stops both mocks
     """
-    def stop_mock_http_get():
-        if stop_mock_http_get.mock is not None:
-            stop_mock_http_get.mock.stop()
-            stop_mock_http_get.mock = None
-    stop_mock_http_get.mock = None
+    def stop_mock_http_request():
+        if stop_mock_http_request.mock is not None:
+            stop_mock_http_request.mock.stop()
+            stop_mock_http_request.mock = None
+    stop_mock_http_request.mock = None
 
     def stop_mock_crypt_util():
         if stop_mock_crypt_util.mock is not None:
@@ -50,26 +51,33 @@ def create(mock_wire_data_file):
             stop_mock_crypt_util.mock = None
     stop_mock_crypt_util.mock = None
 
+    def stop():
+        stop_mock_crypt_util()
+        stop_mock_http_request()
+
     protocol = WireProtocol(KNOWN_WIRESERVER_IP)
     protocol.mock_wire_data = WireProtocolData(mock_wire_data_file)
-    protocol.stop_mock_http_get = stop_mock_http_get
+    protocol.stop_mock_http_request = stop_mock_http_request
     protocol.stop_mock_crypt_util = stop_mock_crypt_util
+    protocol.stop = stop
 
     try:
-        # To minimize the impact of mocking restutil.http_get we only use the mock data for requests
+        # To minimize the impact of mocking restutil.http_request we only use the mock data for requests
         # to the wireserver or requests starting with "mock-goal-state"
-        original_http_get = http_get
-
         mock_data_re = re.compile(r'https?://(mock-goal-state|{0}).*'.format(KNOWN_WIRESERVER_IP.replace(r'.', r'\.')), re.IGNORECASE)
 
-        def mock_http_get(url, *args, **kwargs):
-            if mock_data_re.match(url) is None:
-                return original_http_get(url, *args, **kwargs)
-            return protocol.mock_wire_data.mock_http_get(url, *args, **kwargs)
+        original_http_request = http_request
 
-        p = patch("azurelinuxagent.common.utils.restutil.http_get", side_effect=mock_http_get)
+        def mock_http_request(method, url, data, headers=None, use_proxy=False, max_retry=DEFAULT_RETRIES, retry_codes=RETRY_CODES, retry_delay=DELAY_IN_SECONDS):
+            if method == 'GET' and mock_data_re.match(url) is not None:
+                return protocol.mock_wire_data.mock_http_get(url, headers, use_proxy, max_retry, retry_codes, retry_delay)
+            elif method == 'POST':
+                return protocol.mock_wire_data.mock_http_post(url, data, headers, use_proxy, max_retry, retry_codes, retry_delay)
+            return original_http_request(method, url, data, headers, use_proxy, max_retry, retry_codes, retry_delay)
+
+        p = patch("azurelinuxagent.common.utils.restutil.http_request", side_effect=mock_http_request)
         p.start()
-        stop_mock_http_get.mock = p
+        stop_mock_http_request.mock = p
 
         p = patch("azurelinuxagent.common.protocol.wire.CryptUtil", side_effect=protocol.mock_wire_data.mock_crypt_util)
         p.start()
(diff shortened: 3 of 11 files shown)


azurelinuxagent/common/protocol/wire.py
```diff
@@ -520,11 +520,6 @@ def __init__(self, endpoint):
         logger.info("Wire server endpoint:{0}", endpoint)
         self._endpoint = endpoint
         self._goal_state = None
-        self._hosting_env = None
```
narrieta: I forgot to remove these on #1777

tests/protocol/mock_wire_protocol.py
```diff
@@ -28,48 +28,56 @@ def create(mock_wire_data_file):
     Creates a mock WireProtocol object that will return the data specified by 'mock_wire_data_file' (which must
     follow the structure of the data files defined in tests/protocol/mockwiredata.py).
 
-    NOTE: This function creates mocks for azurelinuxagent.common.utils.restutil.http_get and
+    NOTE: This function creates mocks for azurelinuxagent.common.utils.restutil.http_request and
```
narrieta: I had to mock 1 level deeper (http_request instead of http_get) to address all the scenarios in the tests,

tests/protocol/mock_wire_protocol.py
```diff
@@ -28,48 +28,56 @@ def create(mock_wire_data_file):
...
-        def mock_http_get(url, *args, **kwargs):
-            if mock_data_re.match(url) is None:
-                return original_http_get(url, *args, **kwargs)
-            return protocol.mock_wire_data.mock_http_get(url, *args, **kwargs)
+        original_http_request = http_request
+
+        def mock_http_request(method, url, data, headers=None, use_proxy=False, max_retry=DEFAULT_RETRIES, retry_codes=RETRY_CODES, retry_delay=DELAY_IN_SECONDS):
+            if method == 'GET' and mock_data_re.match(url) is not None:
+                return protocol.mock_wire_data.mock_http_get(url, headers, use_proxy, max_retry, retry_codes, retry_delay)
+            elif method == 'POST':
```
narrieta: added support for POST

tests/protocol/mock_wire_protocol.py
```diff
@@ -28,48 +28,56 @@ def create(mock_wire_data_file):
...
+    def stop():
+        stop_mock_crypt_util()
+        stop_mock_http_request()
+
     protocol = WireProtocol(KNOWN_WIRESERVER_IP)
     protocol.mock_wire_data = WireProtocolData(mock_wire_data_file)
-    protocol.stop_mock_http_get = stop_mock_http_get
+    protocol.stop_mock_http_request = stop_mock_http_request
     protocol.stop_mock_crypt_util = stop_mock_crypt_util
+    protocol.stop = stop
```
narrieta: those stop functions are mainly for debugging purposes; they come handy when there are conflicts with the mocks used by tests and the ones used here.

tests/protocol/test_hostplugin.py
```diff
@@ -33,7 +33,7 @@
 from azurelinuxagent.common.utils import restutil
 from tests.protocol import mock_wire_protocol
 from tests.protocol.mockwiredata import WireProtocolData, DATA_FILE, DATA_FILE_NO_EXT
-from tests.protocol.test_wire import MockResponse
+from tests.protocol.test_wire import MockResponse as TestWireMockResponse
```
narrieta: This one was so much fun to debug! This test uses 2 entities named MockResponse: the one imported here and another at the bottom of the file.

Which one gets used depends on whether patch is used as a decorator or as a function. Ah, the tangled web we weave!

The import alias takes care of this issue.

tests/protocol/test_hostplugin.py
```diff
@@ -403,6 +403,7 @@ def test_validate_http_request(self):
 
         with mock_wire_protocol.create(DATA_FILE) as protocol:
             test_goal_state = protocol.client._goal_state
+            plugin = protocol.client.get_host_plugin()
```
narrieta: Since now get_host_plugin() always fetches the goal state, this initialization was breaking the code below under the scope of the http_request mock.

I may come back (and also ask your help) to update more tests. Similar issues are all over the code.

tests/protocol/test_hostplugin.py
```diff
@@ -564,151 +562,153 @@ def test_validate_get_extension_artifacts(self):
...
-            self.fail('IO error expected to be raised')
-        except IOError:
-            # expected
-            pass
-
-    @patch("azurelinuxagent.common.utils.restutil.http_get",
-           return_value=MockResponse(status_code=200, body=b''))
-    @patch("azurelinuxagent.common.protocol.healthservice.HealthService.report_host_plugin_versions")
-    def test_ensure_health_service_called(self, patch_http_get, patch_report_versions):
+        with patch("azurelinuxagent.common.utils.restutil.http_get") as patch_http_get:
```
narrieta: I moved these patch decorators to be statements while I was debugging the issue with the mock request. Although this change is no longer needed I'm leaving there should we need to debug more.

tests/protocol/test_wire.py
```diff
@@ -209,26 +209,28 @@ def test_get_host_ga_plugin(self, *args):
...
             self.assertEqual(goal_state.role_config_name, host_plugin.role_config_name)
 
-    @skip_if_predicate_true(lambda: True, "Needs to be re-enabled before release 2.2.47")
-    def test_upload_status_blob_default(self, *args):
-        """
-        Default status blob method is HostPlugin.
-        """
-        with create_mock_protocol(status_upload_blob=testurl, status_upload_blob_type=testtype) as protocol:
-            protocol.client.status_blob.vm_status = VMStatus(message="Ready", status="Ready")
+    def test_upload_status_blob_should_use_the_host_channel_by_default(self, *_):
```
narrieta: The changes below are the meat of this PR. Now we mock the interaction with the host/storage in terms of http requests, instead of reaching into implementation internals.

tests/protocol/test_wire.py
```diff
@@ -516,227 +521,235 @@ def test_get_ext_conf_with_uri(self, *args):
...
-        with patch("azurelinuxagent.common.utils.restutil._http_request",
-                   side_effect=[mock_failed_response, mock_successful_response]):
-            with patch("azurelinuxagent.common.protocol.wire.WireClient.update_goal_state") as mock_update_goal_state:
-                with patch("azurelinuxagent.common.protocol.wire.WireClient.fetch", wraps=client.fetch) as patch_direct:
-                    with patch("azurelinuxagent.common.protocol.wire.WireClient.fetch_manifest_through_host",
-                               wraps=client.fetch_manifest_through_host) as patch_host:
-                        ret = client.fetch_manifest([VMAgentManifestUri(uri="uri1")])
-                        self.assertEquals(ret, "OK")
+        with mock_wire_protocol.create(mockwiredata.DATA_FILE) as protocol:
+            protocol.client.get_host_plugin()  # force initialization of the host plugin
```
pgombar: Why is this needed here and not in the test above?

tests/protocol/test_wire.py
```diff
@@ -516,227 +521,235 @@ def test_get_ext_conf_with_uri(self, *args):
...
+            with patch("azurelinuxagent.common.utils.restutil.http_request", side_effect=http_request) as mock_request:
+                HostPluginProtocol.set_default_channel(False)
 
-        HostPluginProtocol.set_default_channel(False)
+                manifest = protocol.client.fetch_manifest([VMAgentManifestUri(uri=manifest_url)])
 
-        mock_failed_response = MockResponse(body=b"", status_code=httpclient.GONE)
-        mock_successful_response = MockResponse(body=b"OK", status_code=200)
+                self.assertEquals(manifest, manifest_xml)
+                self.assertTrue(any(args[0][0] == 'GET' and args[0][1] == manifest_url for args in mock_request.call_args_list))  # direct channel
```
pgombar: Don't we expect the first GET to succeed since we're mocking it? Why are we expecting several requests?

tests/protocol/test_wire.py
narrieta: It is not needed, but simplifies the mock below. Initialization of the host plugin does a request for the goal state. Since the mock below is tracking goal state requests I would need to add code to filter out this one request. It is not too hard, but adding that code there would make the code less readable since it would distract from the actual intention of the test.

Some of the other tests don't track goal state requests so I do it only when it simplifies the code.

I'll add a coment.

tests/protocol/test_wire.py
narrieta: Note that this test follows a different pattern than the other ones. This one was one of the first test I started with.

The underlying implementation of the protocol can trigger multiple requests while executing those fetch* and upload* methods. For example, if we get to use the host plugin we will trigger a request for the goal state, followed by puts to the health service for the plugin, etc.

The mocks in the other tests keep track of the interesting requests in http_request.urls and discard the rest. In this one test I am using mock_request.call_args_list instead. If you run the test under the debugger and set a breakpoint on the first assert you can check the call list and you will see a request for the goal state on top of the the GET for the manifest.

I think the pattern I used on the other tests allows for stronger assertions, and also makes tests a lot more readable so I'll change this one to follow that pattern as well.

Also, post-release, I'll see if I can capture that pattern in a more general tool that we can use for this kind of tests.

narrieta: I captured the pattern I was mentioning as mock_http_request in mocks.py

tests/common/test_event.py
```diff
@@ -28,7 +28,8 @@
     WALAEventOperation, parse_xml_event, parse_json_event, AGENT_EVENT_FILE_EXTENSION, EVENTS_DIRECTORY
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.protocol.goal_state import GoalState
-from tests.protocol import mockwiredata, mock_wire_protocol
+from tests.protocol import mockwiredata
+from tests.protocol.mocks import mock_wire_protocol
```
narrieta: I moved mock_wire_protocol to a different file (mocks.py)

narrieta: I also changed the usage from

` with mock_wire_protocol.create(...)... 
`

to

` with mock_wire_protocol(...)... 
`

As I'm adding more functionality to the mocks the latter feels better.

tests/protocol/mock_wire_protocol.py
```diff
@@ -1,84 +0,0 @@
...
-
-import contextlib
-import re
-from azurelinuxagent.common.protocol.wire import WireProtocol
-from azurelinuxagent.common.utils.restutil import KNOWN_WIRESERVER_IP, http_get
-from tests.tools import patch
-from tests.protocol.mockwiredata import WireProtocolData
-
-@contextlib.contextmanager
-def create(mock_wire_data_file):
```
narrieta: moved this to mocks.py as mock_wire_protocol

tests/protocol/mocks.py
```diff
@@ -0,0 +1,142 @@
...
+import contextlib
+import re
+from azurelinuxagent.common.protocol.wire import WireProtocol
+from azurelinuxagent.common.utils import restutil
+from tests.tools import patch
+from tests.protocol.mockwiredata import WireProtocolData
+
+
+@contextlib.contextmanager
+def mock_wire_protocol(mock_wire_data_file):
```
narrieta: moved here from the previous mock_wire_protocol.py, with a few changes I am highlighting below

tests/protocol/mocks.py
```diff
@@ -0,0 +1,142 @@
...
+        # To minimize the impact of mocking restutil.http_request we only use the mock data for requests
+        # to the wireserver or requests starting with "mock-goal-state"
+        mock_data_re = re.compile(r'https?://(mock-goal-state|{0}).*'.format(restutil.KNOWN_WIRESERVER_IP.replace(r'.', r'\.')), re.IGNORECASE)
+
+        original_http_request = restutil.http_request
+
+        def http_request(method, url, data, headers=None, use_proxy=False, max_retry=restutil.DEFAULT_RETRIES, retry_codes=restutil.RETRY_CODES, retry_delay=restutil.DELAY_IN_SECONDS):
+            if method == 'GET' and mock_data_re.match(url) is not None:
+                return protocol.mock_wire_data.mock_http_get(url, headers, use_proxy, max_retry, retry_codes, retry_delay)
+            elif method == 'POST':
```
narrieta: added support for POST

tests/protocol/mocks.py
```diff
@@ -0,0 +1,142 @@
...
+            stop_mock_http_request.mock = None
+    stop_mock_http_request.mock = None
+
+    def stop_mock_crypt_util():
+        if stop_mock_crypt_util.mock is not None:
+            stop_mock_crypt_util.mock.stop()
+            stop_mock_crypt_util.mock = None
+    stop_mock_crypt_util.mock = None
+
+    def stop():
```
narrieta: those stop functions are mainly for debugging purposes; they come handy when there are conflicts with the mocks used by tests and the ones used here.

tests/protocol/mocks.py
```diff
@@ -0,0 +1,142 @@
...
+        protocol.detect()
+
+        yield protocol
+
+    finally:
+        protocol.stop()
+
+
+@contextlib.contextmanager
+def mock_http_request(http_get_handler=None, http_post_handler=None, http_put_handler=None):
```
narrieta: New mock useful for tests that need to interact with the host/storage at the HTTP request level

tests/common/test_event.py
vrdmr: Yup - Looks cleaner. Thanks

tests/protocol/test_wire.py
```diff
@@ -516,227 +515,209 @@ def test_get_ext_conf_with_uri(self, *args):
...
-    def test_download_ext_handler_pkg_should_not_change_default_channel_if_host_fails(self, mock_get_artifact_request,
-                                                                                      *args):
-        mock_get_artifact_request.return_value = "dummy_url", "dummy_header"
-        protocol = WireProtocol("foo.bar")
-        HostPluginProtocol.set_default_channel(False)
+                urls = http_request.get_tracked_urls()
+                self.assertEquals(success, True, 'The download should have succeeded')
+                self.assertEquals(len(urls), 2, "Unexpected number of HTTP requests: [{0}]".format(urls))
+                self.assertEquals(urls[0], extension_url, "The first attempt should have been over the direct channel")
+                self.assertTrue(urls[1].endswith('/extensionArtifact'), "The retry attempt should have been over the host channel")
```
vrdmr: Do we need to validate the IP address to validate it went through the HGAP? I know the `/extensionArtifact` should be good enough to validate if the correct URI was hit.

narrieta: yes, that is one of the changes i want to do post-release

tests/protocol/test_wire.py
pgombar: Got it, thanks for the update.

PR #1806, opened 2020-03-11
[DNM] Drop MDS Support and Cleanup MDS Certificates/Firewall Rule

byue:
## Description

Ensure agent cleans up MDS certificates, and MDS firewall rule at 169.254.169.254 so that IMDS is not blocked (they use the same IP). This PR also ensures wireserver certificates and firewall rules are re-initialized through the detect pathway when Linux Agent migrates from MDS protocol to WireServer Protocol.

### Verification Plan
- [x] Test WireServer to WireServer update (manual/auto), MetdataServer to WireServer update (manual/auto), and new WireServer VM scenarios. Also test extension operations (create, re-run, delete). Collect logs to see if 169.254.169.254 firewall rule is removed and verify certificates in /var/lib/waagent are cleaned up.
- [x] Examine Kusto Logs and verify AzureStack internals are ok.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1806)


azurelinuxagent/common/protocol/util.py
```diff
@@ -317,34 +270,33 @@ def clear_protocol(self):
...
-            protocols.append(prots.MetadataProtocol)
-        protocol_name, protocol = self._detect_protocol(protocols)
+        protocol_file_path = self._get_protocol_file_path()
+        if os.path.isfile(protocol_file_path):
+            protocol_name = fileutil.read_file(protocol_file_path)
+            if protocol_name == prots.WireProtocol:
+                endpoint = self.get_wireserver_endpoint()
+                self._protocol = WireProtocol(endpoint)
+                return self._protocol
+            elif protocol_name == prots.MetadataProtocol:
```
byue: Checking Protocol file for MetadataServer is not sufficient for the manual upgrade case where we upgrade from MetadataServer to WireServer by restarting the agent (WireServer is written to protocol file so we never enter cleanup branch).

azurelinuxagent/common/protocol/util.py
```diff
@@ -317,34 +262,33 @@ def clear_protocol(self):
...
-
         if self._protocol is not None:
             return self._protocol
 
-        try:
-            self._protocol = self._get_protocol()
+        # If metadataserver certificates is present we clean certificates
+        # and remove MetadataServer firewall rule. It is possible
+        # Wire Protocol is being used due to a previous intermediate upgrade
+        # before 2.2.47 but metadata artifacts were not cleaned up.
```
narrieta: this would now be 2.2.48

azurelinuxagent/common/protocol/util.py
```diff
@@ -317,34 +262,33 @@ def clear_protocol(self):
...
+        if is_metadata_server_artifact_present():
+            cleanup_metadata_server_artifacts(self.osutil)
+
+        protocol_file_path = self._get_protocol_file_path()
+        if os.path.isfile(protocol_file_path) and fileutil.read_file(protocol_file_path) == WIRE_PROTOCOL_NAME:
```
narrieta: do you have an scenario where the metadata cert is not present, but the protocol file still contains "Metadata"?

if not, i think we can remove "and fileutil.read_file(protocol_file_path) == WIRE_PROTOCOL_NAME" here.

azurelinuxagent/common/protocol/util.py
```diff
@@ -212,82 +183,56 @@ def _clear_wireserver_endpoint(self):
...
-            self.dhcp_handler.skip_cache = True
-            raise e
-
     def _detect_metadata_protocol(self):
         protocol = MetadataProtocol()
         protocol.detect()
         return protocol
 
-    def _detect_protocol(self, protocols):
+    def _detect_protocol(self):
```
narrieta: I think _detect_metadata_protocol above is no longer needed

azurelinuxagent/common/protocol/metadata_server_migration_util.py
```diff
@@ -0,0 +1,81 @@
...
+def _reset_firewall_rules(osutil):
+    """
+    Removes MetadataServer firewall rule so IMDS can be used. Enables
+    WireServer firewall rule based on if firewall is configured to be on.
+    """
+    osutil.remove_rules_files()
+    osutil.remove_firewall(dst_ip=_KNOWN_METADATASERVER_IP, uid=os.getuid())
+    if conf.enable_firewall():
+        success = osutil.enable_firewall(dst_ip=KNOWN_WIRESERVER_IP, uid=os.getuid())
+        add_periodic(
```
narrieta: this should be add_event; we use periodic in the env thread because that one runs continuously, but this is a 1-time deal

tests/protocol/test_metadata_server_migration_util.py
```diff
@@ -0,0 +1,85 @@
...
+
+from azurelinuxagent.common.protocol.metadata_server_migration_util import _LEGACY_METADATA_SERVER_TRANSPORT_PRV_FILE_NAME, \
+                                                                           _LEGACY_METADATA_SERVER_TRANSPORT_CERT_FILE_NAME, \
+                                                                           _LEGACY_METADATA_SERVER_P7B_FILE_NAME, \
+                                                                           _KNOWN_METADATASERVER_IP
+from azurelinuxagent.common.utils.restutil import KNOWN_WIRESERVER_IP
+from tests.tools import AgentTestCase, patch, call
+
+class TestMetadataServerMigrationUtil(AgentTestCase):
+    def test_ensure_file_removed(self):
```
narrieta: we try to avoid testing private methods
_ensure_file_removed should be covered by the other tests

narrieta: similar for tests below

tests/protocol/test_protocol_util.py
```diff
@@ -138,64 +107,66 @@ def test_detect_wire_protocol_no_dhcp(self, WireProtocol, mock_get_lib_dir, _):
...
-        protocol = protocol_util.get_protocol()
-        self.assertEquals(MetadataProtocol.return_value, protocol)
-        protocol_util.get_wireserver_endpoint.assert_not_called()
+        protocol_util.get_wireserver_endpoint.assert_any_call()
 
-        # Test for unknown protocol
-        protocol_util.clear_protocol()
-        protocol_util._save_protocol("Not_a_Protocol")
-        protocol_util._detect_protocol.side_effect = NotImplementedError()
+    @patch("os.path.isfile")
```
narrieta: this is an excessive use of mocks that make those tests too fragile. let's sync offline

tests/protocol/test_protocol_util.py
```diff
@@ -138,64 +107,66 @@ def test_detect_wire_protocol_no_dhcp(self, WireProtocol, mock_get_lib_dir, _):
...
+        protocol_util.get_wireserver_endpoint = Mock()
+        mock_isfile.return_value = True
+        wire_protocol.return_value = MagicMock()
+        artifact_present.return_value = False
+        mock_read_file.return_value = WIRE_PROTOCOL_NAME
+        cleanup_util.return_value = None
+        protocol_util = get_protocol_util()
+        protocol_util._detect_protocol = MagicMock()
+        protocol_util.get_protocol()
+        cleanup_util.assert_not_called()
```
narrieta: could you add test cases for the 3 scenarios we discussed over Teams?

azurelinuxagent/common/protocol/util.py
byue: This will break the migration scenario from MDS to WireServer because we will return early and not generate WireServer certificates here. If we don't check the file contents and the file contains MetadataServer, we will not do _detect_protocol which takes care of generating the WireServer certificates.

azurelinuxagent/common/protocol/util.py
```diff
@@ -317,34 +257,35 @@ def clear_protocol(self):
...
         if self._protocol is not None:
             return self._protocol
 
-        try:
-            self._protocol = self._get_protocol()
+        # If metadataserver certificates is present we clean certificates
+        # and remove MetadataServer firewall rule. It is possible
+        # Wire Protocol is being used due to a previous intermediate upgrade
+        # before 2.2.48 but metadata artifacts were not cleaned up.
+        if is_metadata_server_artifact_present():
```
byue: need to move this after _protocol_detect, if _protocol_detect fails we shouldn't delete MetadataServer certificates in the case where upgrade fails and we roll back to previous version.

azurelinuxagent/common/protocol/util.py
narrieta: it's ok now, the comment you added addresses my question

azurelinuxagent/common/protocol/util.py
narrieta: ok, could you also capture that scenario on a unit test? thanks

azurelinuxagent/common/protocol/util.py
```diff
@@ -34,36 +32,25 @@
...
 from azurelinuxagent.common.protocol.ovfenv import OvfEnv
 from azurelinuxagent.common.protocol.wire import WireProtocol
-from azurelinuxagent.common.protocol.metadata import MetadataProtocol
 from azurelinuxagent.common.utils.restutil import KNOWN_WIRESERVER_IP, \
                                                   IOErrorCounter
 
 OVF_FILE_NAME = "ovf-env.xml"
-TAG_FILE_NAME = "useMetadataEndpoint.tag"
 PROTOCOL_FILE_NAME = "Protocol"
 MAX_RETRY = 360
```
byue: Are we able to reduce MAX_RETRY? In the case that we update from MetadataServer to WireServer, and WireServer is not reachable, we will keep trying to contact it 360 times (1 hour if we probe every 10 seconds) before failing update and reverting to MetadataServer. Previously the retry loop would also try contacting MetadataServer, but since this is stripped out we are now only trying to contact WireServer in _detect.

narrieta: If you want to change the number of retries only during the migration then it should be fine.

Changing it for all other scenarios may not be right, we would need to evaluate whether that would affect public. For that I would recommend doing it on a separate PR.

byue: In_detect_protocol in util.py, can we do something this?
retries =  5 if is_metadata_server_artifact_present() else MAX_RETRY

narrieta:
LGTM - Running automation

@vrdmr @pgombar - can i have a review, please?

tests/protocol/test_metadata_server_migration_util.py
```diff
@@ -0,0 +1,131 @@
...
+from azurelinuxagent.common.protocol.metadata_server_migration_util import _LEGACY_METADATA_SERVER_TRANSPORT_PRV_FILE_NAME, \
+                                                                           _LEGACY_METADATA_SERVER_TRANSPORT_CERT_FILE_NAME, \
+                                                                           _LEGACY_METADATA_SERVER_P7B_FILE_NAME, \
+                                                                           _KNOWN_METADATASERVER_IP
+from azurelinuxagent.common.utils.restutil import KNOWN_WIRESERVER_IP
+from tests.tools import AgentTestCase, patch, MagicMock
+
+class TestMetadataServerMigrationUtil(AgentTestCase):
+    @patch('azurelinuxagent.common.conf.get_lib_dir')
+    def test_is_metadata_server_artifact_not_present(self, mock_get_lib_dir):
```
pgombar: Should the name of this test and the next one be swapped?

tests/protocol/test_metadata_server_migration_util.py
```diff
@@ -0,0 +1,131 @@
...
+        dir = tempfile.gettempdir()
+        metadata_server_transport_cert_file = os.path.join(dir, _LEGACY_METADATA_SERVER_TRANSPORT_CERT_FILE_NAME)
+        open(metadata_server_transport_cert_file, 'w').close()
+        mock_get_lib_dir.return_value = dir
+        assert migration_util.is_metadata_server_artifact_present()
```
pgombar: nit: in the majority of this repo, the syntax for assert is `self.assertEquals`/`self.assertTrue`

azurelinuxagent/common/protocol/metadata_server_migration_util.py
```diff
@@ -0,0 +1,80 @@
...
+    if conf.enable_firewall():
+        success = osutil.enable_firewall(dst_ip=KNOWN_WIRESERVER_IP, uid=os.getuid())
+        add_event(
+            AGENT_NAME,
+            version=CURRENT_VERSION,
+            op=WALAEventOperation.Firewall,
+            is_success=success,
+            log_event=False)
+
+def _ensure_file_removed(directory, file_name):
```
larohra: Why not add this function in `utils/fileutils.py` and make it public since its a helper function.

tests/protocol/test_protocol_util.py
```diff
@@ -138,64 +129,162 @@ def test_detect_wire_protocol_no_dhcp(self, WireProtocol, mock_get_lib_dir, _):
...
         protocol_util.get_wireserver_endpoint.assert_any_call()
 
-        # Test to ensure protocol persists
-        protocol_util.get_wireserver_endpoint.reset_mock()
-        protocol_util._save_protocol("MetadataProtocol")
+    @patch('azurelinuxagent.common.conf.get_lib_dir')
+    @patch('azurelinuxagent.common.conf.enable_firewall')
+    def test_get_protocol_wireserver_to_wireserver_update_removes_metadataserver_artifacts(self, mock_enable_firewall, mock_get_lib_dir, _):
+        """
+        This is for testing that agent upgrade from WireServer to WireServer protocol
```
larohra: I'm sorry I couldn't understand the context of this test, could you please explain briefly this scenario
>agent upgrade from WireServer to WireServer protocol

Thanks!

azurelinuxagent/common/protocol/metadata_server_migration_util.py
byue: Are there future plans to use this? I put it here so that this code is easier to remove post-migration.

tests/protocol/test_protocol_util.py
byue: This is testing the normal agent upgrade with both the old agent and new agent using WireServer protocol. However I am testing that MetadataServer certificates are cleaned up during this upgrade. MetadataServer certificates could be leftover because there was a previous upgrade from MetadataServer to WireServer, which does not have the logic for cleaning up MetadataServer certificates.

MDS Agent -> WS Agent (does not clean up MDS certificates, older version) -> WS Agent (version this PR is, has cleanup logic)

tests/protocol/test_metadata_server_migration_util.py
byue: thanks for the catch, not sure why I did that

tests/protocol/test_metadata_server_migration_util.py
byue: fixed

azurelinuxagent/common/protocol/metadata_server_migration_util.py
larohra: No current plans to use it as of now but since its a utility function, might as well have it in the utils directory incase we ever need it later :)

tests/protocol/test_protocol_util.py
larohra: Ahh now I get it, thanks for adding such a verbose test! :)

azurelinuxagent/common/protocol/metadata_server_migration_util.py
larohra: Nevermind, had a small discussion about this with my team and since its going to be removed anyways no need to put it to a Util function. Thanks!

tests/protocol/test_metadata_server_migration_util.py
```diff
@@ -0,0 +1,131 @@
...
+    def test_cleanup_metadata_server_artifacts_firewall_disabled(self, mock_os_getuid, mock_get_lib_dir, mock_enable_firewall):
+        # Setup Certificate Files
+        dir = tempfile.gettempdir()
+        metadata_server_transport_prv_file = os.path.join(dir, _LEGACY_METADATA_SERVER_TRANSPORT_PRV_FILE_NAME)
+        metadata_server_transport_cert_file = os.path.join(dir, _LEGACY_METADATA_SERVER_TRANSPORT_CERT_FILE_NAME)
+        metadata_server_p7b_file = os.path.join(dir, _LEGACY_METADATA_SERVER_P7B_FILE_NAME)
+        open(metadata_server_transport_prv_file, 'w').close()
+        open(metadata_server_transport_cert_file, 'w').close()
+        open(metadata_server_p7b_file, 'w').close()
+
```
vrdmr: You could refactor all the common code into a `_setup_metadata_protocol_setup`-type method.

vrdmr: Or just override the `setUp` method of `AgentTestCase`.

azurelinuxagent/common/protocol/util.py
```diff
@@ -216,82 +188,51 @@ def _clear_wireserver_endpoint(self):
...
+                    raise e
+    
+            except ProtocolError as e:
+                logger.info("Protocol endpoint not found: {0}", e)
 
             if retry < MAX_RETRY - 1:
-                logger.info("Retry detect protocols: retry={0}", retry)
+                logger.info("Retry detect protocol: retry={0}", retry)
                 time.sleep(PROBE_INTERVAL)
         raise ProtocolNotFoundError("No protocol found.")
```
vrdmr: So - It can wait - in theory- for 1 hour before failing to raise ProtocolNotFoundError()? Just confirming.

I understand Telemetry events won't work either as telemetry requires the protocol to be enabled, for the Wireserver to receive the `/telemetrydata` request

vrdmr: Is there guidance on how much we should wait? or we decide the wait time period.

vrdmr: I'll investigate and update if reqd.

vrdmr: @byue - Dont worry about this comment.

tests/protocol/test_metadata_server_migration_util.py
vrdmr: @byue - Dont worry about this comment - You can do it in a different PR.

PR #1815, opened 2020-03-17
Add lock around protocol detection

narrieta:
Protocol detection creates a few files on disk, which need to be protected with a lock.

@byue - FYI

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1815)

### DIFF
--- a/azurelinuxagent/common/protocol/util.py
+++ b/azurelinuxagent/common/protocol/util.py
@@ -74,6 +74,7 @@ class ProtocolUtil(SingletonPerThread):
     """
 
     def __init__(self):
+        self._lock = threading.RLock()  # protects the files on disk created during protocol detection
         self._protocol = None
         self.endpoint = None
         self.osutil = get_osutil()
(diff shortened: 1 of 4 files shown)


tests/protocol/test_protocol_util.py
```diff
@@ -44,29 +44,24 @@ def test_get_protocol_util_should_return_same_object_for_same_thread(self, _):
         self.assertEqual(protocol_util1, protocol_util2)
 
     def test_get_protocol_util_should_return_different_object_for_different_thread(self, _):
-        def get_util_obj(q, err):
+        protocol_util_instances = []
```
narrieta: now the protocol object has a lock so it cannot be serialized into the queue

byue:
Are there multiple readers? If so we could use read locks to allow multiple readers and write locks.

narrieta:
@byue  This was an existing lock that we are putting back temporarily and will remove later again. Performance is not an issue here.

config/freebsd/waagent.conf
```diff
@@ -48,10 +48,10 @@ ResourceDisk.Filesystem=ufs
 ResourceDisk.MountPoint=/mnt/resource
 
 # Create and use swapfile on resource disk.
-ResourceDisk.EnableSwap=n
+ResourceDisk.EnableSwap=y
```
larohra: I think changing the target branch picked up an unwanted commit with this PR

azurelinuxagent/common/protocol/util.py
```diff
@@ -74,6 +74,7 @@ class ProtocolUtil(SingletonPerThread):
     """
 
     def __init__(self):
+        self._lock = threading.RLock()  # protects the files on disk created during protocol detection
```
pgombar: What is the idea behind using a reentrant thread lock instead of a regular thread lock?

narrieta:
Replaced by #1816, which will be part of release 2.2.47

PR #1820, opened 2020-03-18
Reduce the verbosity of goal state errors

narrieta:
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1820)

### DIFF
--- a/azurelinuxagent/common/protocol/restapi.py
+++ b/azurelinuxagent/common/protocol/restapi.py
@@ -238,6 +238,9 @@ class Protocol(DataContract):
     def update_goal_state(self):
         raise NotImplementedError()
 
+    def try_update_goal_state(self):
+        raise NotImplementedError()
+
     def update_host_plugin_from_goal_state(self):
         raise NotImplementedError()
 
--- a/tests/protocol/test_wire.py
+++ b/tests/protocol/test_wire.py
@@ -16,12 +16,13 @@
 # Requires Python 2.6+ and Openssl 1.0+
 #
 
+import contextlib
 import json
 import os
+import re
 import time
 import unittest
 import uuid
-import contextlib
 
 from azurelinuxagent.common.exception import InvalidContainerError, ResourceGoneError, ProtocolError, \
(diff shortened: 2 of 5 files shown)


tests/protocol/test_wire.py
```diff
@@ -1056,6 +1057,109 @@ def test_forced_update_should_update_the_goal_state_and_the_host_plugin_when_the
...
+                #
+                with create_log_and_telemetry_mocks() as (logger, add_event):
+                    protocol.client.try_update_goal_state()
+                    protocol.client.try_update_goal_state()
+                    protocol.client.try_update_goal_state()
+
+                    w = warnings()
+                    pw = periodic_warnings()
+                    self.assertTrue(len(w) == 0, "Subsequent failures should not produce warnings: [{0}]".format(w))
+                    self.assertEquals(len(pw), 3, "Subsequent failures should produce periodic warnings: [{0}]".format(pw))
```
pgombar: Why would periodic warnings be 3 instead of 1 if we are logging them every 6 hours?

narrieta: we always call periodic_warning... the underlying implementation should handle the period

pgombar: Got it, this just asserts the periodic is being called.

azurelinuxagent/common/protocol/restapi.py
```diff
@@ -238,6 +238,9 @@ def detect(self):
     def update_goal_state(self):
         raise NotImplementedError()
 
+    def try_update_goal_state(self):
```
larohra: Small suggestion, why not retire the public `update_goal_state` method for `try_update_goal_state` one to ensure everyone only uses this public function.

narrieta: both of them will go away in favor of the get_goal_state function we've talked about. The goal state is still stored in the protocol client and we need to get it our of there. Will continue the refactoring once metadata is gone. Not only Metadata is a pain to update, but I do not want to take the risk of breaking it, since we do not automation for it.

PR #1827, opened 2020-03-20
Update test_wire unit tests to use mock_wire_protocol

pgombar:
## Description

Update leftover tests from #1800.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1827)

### DIFF
--- a/tests/protocol/test_wire.py
+++ b/tests/protocol/test_wire.py
@@ -316,68 +316,6 @@ class TestWireProtocol(AgentTestCase):
             self.assertEqual(dict(onHold='true'), in_vm_artifacts_profile.__dict__)
             self.assertTrue(in_vm_artifacts_profile.is_on_hold())
 
-    def test_fetch_manifest_fallback(self, *args):
-        uri1 = ExtHandlerVersionUri()
-        uri1.uri = 'ext_uri'
-        uris = DataContractList(ExtHandlerVersionUri)
-        uris.append(uri1)
-        host_uri = 'host_uri'
-        mock_host = HostPluginProtocol(host_uri,
-                                       'container_id',
-                                       'role_config')
-        client = WireProtocol(WIRESERVER_URL).client
-
-        with patch.object(WireClient, "fetch", return_value=None) as patch_fetch:
-            with patch.object(WireClient, "get_host_plugin", return_value=mock_host):
-                with patch.object(HostPluginProtocol, "get_artifact_request", return_value=[host_uri, {}]):
-                    HostPluginProtocol.set_default_channel(False)
-                    self.assertRaises(ExtensionDownloadError, client.fetch_manifest, uris)
-                    self.assertEqual(patch_fetch.call_count, 2)
-                    self.assertEqual(patch_fetch.call_args_list[0][0][0], uri1.uri)
-                    self.assertEqual(patch_fetch.call_args_list[1][0][0], host_uri)
-
-    # This test checks if the manifest_uri variable is set in the host object of WireClient
-    # This variable is used when we make /extensionArtifact API calls to the HostGA
-    def test_fetch_manifest_ensure_manifest_uri_is_set(self, *args):
-        uri1 = ExtHandlerVersionUri()
-        uri1.uri = 'ext_uri'
-        uris = DataContractList(ExtHandlerVersionUri)
-        uris.append(uri1)
-        host_uri = 'host_uri'
-        mock_host = HostPluginProtocol(host_uri, 'container_id', 'role_config')
-        client = WireProtocol(WIRESERVER_URL).client
-        manifest_return = "manifest.xml"
-
-        with patch.object(WireClient, "get_host_plugin", return_value=mock_host):
-            mock_host.get_artifact_request = MagicMock(return_value=[host_uri, {}])
-
-            # First test tried to download directly from blob and asserts manifest_uri is set
-            with patch.object(WireClient, "fetch", return_value=manifest_return) as patch_fetch:
-                fetch_manifest_mock = client.fetch_manifest(uris)
-                self.assertEqual(fetch_manifest_mock, manifest_return)
-                self.assertEqual(patch_fetch.call_count, 1)
-                self.assertEqual(mock_host.manifest_uri, uri1.uri)
-
-            # Second test tries to download from the HostGA (by failing the direct download)
-            # and asserts manifest_uri is set
-            with patch.object(WireClient, "fetch") as patch_fetch:
-                patch_fetch.side_effect = [None, manifest_return]
-                fetch_manifest_mock = client.fetch_manifest(uris)
-                self.assertEqual(fetch_manifest_mock, manifest_return)
-                self.assertEqual(patch_fetch.call_count, 2)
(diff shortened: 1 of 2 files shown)


tests/protocol/test_wire.py
```diff
@@ -316,68 +316,6 @@ def test_get_in_vm_artifacts_profile_default(self, *args):
             self.assertEqual(dict(onHold='true'), in_vm_artifacts_profile.__dict__)
             self.assertTrue(in_vm_artifacts_profile.is_on_hold())
 
-    def test_fetch_manifest_fallback(self, *args):
```
pgombar: Already covered in `test_fetch_manifest_should_use_host_channel_when_direct_channel_fails_and_set_it_to_default`

tests/protocol/test_wire.py
```diff
@@ -316,68 +316,6 @@ def test_get_in_vm_artifacts_profile_default(self, *args):
...
-                with patch.object(HostPluginProtocol, "get_artifact_request", return_value=[host_uri, {}]):
-                    HostPluginProtocol.set_default_channel(False)
-                    self.assertRaises(ExtensionDownloadError, client.fetch_manifest, uris)
-                    self.assertEqual(patch_fetch.call_count, 2)
-                    self.assertEqual(patch_fetch.call_args_list[0][0][0], uri1.uri)
-                    self.assertEqual(patch_fetch.call_args_list[1][0][0], host_uri)
-
-    # This test checks if the manifest_uri variable is set in the host object of WireClient
-    # This variable is used when we make /extensionArtifact API calls to the HostGA
-    def test_fetch_manifest_ensure_manifest_uri_is_set(self, *args):
```
pgombar: Already covered in `test_fetch_manifest_should_use_host_channel_when_direct_channel_fails_and_set_it_to_default`

tests/protocol/test_wire.py
```diff
@@ -316,68 +316,6 @@ def test_get_in_vm_artifacts_profile_default(self, *args):
...
-            # and asserts manifest_uri is set
-            with patch.object(WireClient, "fetch") as patch_fetch:
-                patch_fetch.side_effect = [None, manifest_return]
-                fetch_manifest_mock = client.fetch_manifest(uris)
-                self.assertEqual(fetch_manifest_mock, manifest_return)
-                self.assertEqual(patch_fetch.call_count, 2)
-                self.assertEqual(mock_host.manifest_uri, uri1.uri)
-                self.assertTrue(HostPluginProtocol.is_default_channel())
-
-    def test_get_in_vm_artifacts_profile_host_ga_plugin(self, *args):
```
pgombar: Already covered in `test_get_artifacts_profile_should_use_host_channel_when_direct_channel_fails`

tests/protocol/test_wire.py
```diff
@@ -492,24 +430,20 @@ def test_get_ext_conf_without_uri(self, *args):
...
-                self.assertEqual(1, len(ext_conf.ext_handlers.extHandlers))
-                self.assertEqual(2, len(ext_conf.vmagent_manifests.vmAgentManifests))
-                self.assertEqual("https://test.blob.core.windows.net/vhds/test-cs12.test-cs12.test-cs12.status?"
-                                 "sr=b&sp=rw&se=9999-01-01&sk=key1&sv=2014-02-14&sig=hfRh7gzUE7sUtYwke78IOlZOrTRCYvkec4hGZ9zZzXo",
-                                 ext_conf.status_upload_blob)
-                self.assertEqual("BlockBlob", ext_conf.status_upload_blob_type)
-                self.assertEqual(None, ext_conf.artifacts_profile_blob)
+    def test_get_ext_conf_with_uri(self):
+        with mock_wire_protocol(mockwiredata.DATA_FILE) as protocol:
+            protocol.detect()
```
narrieta: mock_wire_protocol returns a fully initialized protocol. No need to call detect() again

tests/protocol/test_wire.py
```diff
@@ -492,24 +430,20 @@ def test_get_ext_conf_without_uri(self, *args):
...
+            wire_protocol_client = protocol.client
+            ext_conf = wire_protocol_client.get_ext_conf()
+
+            self.assertEqual(1, len(ext_conf.ext_handlers.extHandlers), "Unexpected number of extension handlers")
```
narrieta: what is the intention of this test? The name of the test doesn't tell me what it is trying to do, there are no comments explaining what is is trying to do, and the messages in the asserts don't help me there either.

Looking at the code, it gives me the impression that it is just a basic test for get_ext_conf() and all we are doing here is checking that the data returned by the function matches mockwiredata.DATA_FILE. If that is indeed the intention of the test a simple docstring might help the reader of this test. The "with_uri" part in the test name I cannot explain? Can you?

pgombar: Yes, the test is just asserting the values from the `ext_conf.xml` are properly populated. The name `with_uri` contrasts the name of the previous test, `without_uri`, where the extension config has no extensions set.

I will add a comment for both tests and improve the naming.

tests/protocol/test_wire.py
pgombar: Tnx, leftover from old test.

tests/protocol/test_wire.py
narrieta: Thanks. 

Let's also improve the assertion messages. Since that is indeed the intention of the test then "The number of handlers in the extension config does not match the test data: [{0}]" might be better than "Unexpected number of extension handlers" (where {0} is the list of handlers that the test fetched).

Similar for other asserts

pgombar: Initially I did that, but the objects in the list then show up as:
`AssertionError: 2 != 1 : Unexpected number of extension handlers: [[<azurelinuxagent.common.protocol.restapi.ExtHandler object at 0x7f71d4d20048>]]`

which I didn't think was valuable.

narrieta: Simple iteration over the handlers extracting the name would do it

narrieta: but if you think it's an overkill you can leave as is, but do review the other messages for improvements

pgombar: Updated.

tests/protocol/test_wire.py
```diff
@@ -482,34 +420,41 @@ def test_report_event_large_event(self, patch_send_event, *args):
 
 
 class TestWireClient(HttpRequestPredicates, AgentTestCase):
-    def test_get_ext_conf_without_uri(self, *args):
+    def test_get_ext_conf_without_extensions_should_retrieve_vmagent_manifests_info(self, *args):
+        # Asserts that WireClient correctly populates the extension config values given in the XML file when extensions
```
narrieta: The comment is misleading. WireClient no longer fetches the goal state  from the XML files on disk.

The mock happens to keep the test data in XML files, but that is irrelevant to what we are testing here.

I would suggest something like "Basic test for get_ext_conf(); verifies that it fetches the correct data by comparing the returned data with the test data provided the mock_wire_protocol" or similar.

pgombar: My bad, updated.

PR #1840, opened 2020-03-27
Remove Protocol interface

narrieta:
Since the Metadata protocol is gone, the Protocol interface is no longer needed.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1840)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -348,13 +348,10 @@ def _extension_processing_allowed(self):
             return False
 
         if conf.get_enable_overprovisioning():
-            if not self.protocol.supports_overprovisioning():
```
narrieta: The default implementation of supports_overprovisioning was True

PR #1847, opened 2020-04-07
Add log collector utility

pgombar:
## Description

First step of the feature to enable WALinuxAgent to periodically upload relevant log files to the host. 

This PR adds the module that contains the functionality to collect logs based on a given manifest file. The format and contents of the manifest files is taken from [Azure Inspect Disk Service](https://github.com/Azure/azure-diskinspect-service/tree/master/pyServer/manifests/linux).

The log collector module supports prioritization of important files in case there is not enough space for all files to be collected. Singe large files are not dropped, but truncated, where the freshest entries are kept.

During collection, the files are kept in a tarball that is incrementally updated on each collection. This means that in subsequent calls, only new files are added to the tarball, deleted files are removed, and modified files are updated, the rest stays the same. Finally, when all the files are collected, the tarball is compressed into a zip before it is sent to the host.

Future work entails adding a thread that will periodically call this utiilty (with limits on resource consumption) and send the zip to the host.

---

### PR information

---
This change is [Reviewable](https://reviewable.io/reviews/azure/walinuxagent/1847)


config/logcollector_manifest_full
```diff
@@ -0,0 +1,73 @@
+echo,### Probing Directories ###
```
pgombar, line 1: This file is identical to IID's [manifest file](https://github.com/Azure/azure-diskinspect-service/blob/master/pyServer/manifests/linux/diagnostic), collection option diagnostic.

config/logcollector_manifest_normal
```diff
@@ -0,0 +1,33 @@
+echo,### Probing Directories ###
```
pgombar, line 1: This file is identical to IID's [manifest file](https://github.com/Azure/azure-diskinspect-service/blob/master/pyServer/manifests/linux/agents), collection option agents.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+        The tarball is then compressed into a zip.
+        :return: Returns True if the log collection succeeded
+        """
+        try:
+            # Clear previous run's output and create base directories if they don't exist already
+            rm_files(OUTPUT_RESULTS_FILE_PATH)
+            mkdir(TRUNCATED_FILES_DIR)
+            mkdir(LOG_COLLECTOR_DIR)
+
+            files_to_collect = self._create_list_of_files_to_collect()
```
pgombar, line 338: We first need to decide which files to collect, then collect them. This line does the former.

This list will contain all files we will collect in the archive. The list is crafted from the manifest file, matched with existing files on disk, prioritized by the `MUST_COLLECT_FILES` list, and cut-off once the overall size of the collected files grows too big.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+            self._log_to_results_file("\n### Creating archive ###")
+
+            self._remove_deleted_files_from_archive(files_to_collect)
+            self._update_files_in_archive(files_to_collect)
```
pgombar, line 342: After we decided which files to collect, the archive is created. Since we don't recreate the entire archive every single collection call (as there will be many of the same files to collect), we only remove the files that have been deleted from disk since the last collection call, add new files and update existing files if they've been modified.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+
+    @staticmethod
+    def _get_list_of_files_in_archive():
+        with tarfile.open(OUTPUT_ARCHIVE_PATH, "a") as tarball:
+            return tarball.getnames()
+
+    @staticmethod
+    def _delete_file_from_archive(file):
+        try:
+            command_string = "tar --file {0} --delete {1}".format(OUTPUT_ARCHIVE_PATH, file)
```
pgombar, line 162: `tarfile` Python module doesn't have a method to delete a file from the archive, so command line option is used. Elsewhere, the Python API is preferred.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+
+
+LOG_COLLECTOR_DIR = '/var/lib/waagent/logcollector'
+TRUNCATED_FILES_DIR = '/var/truncated'
+
+OUTPUT_RESULTS_FILE_PATH = os.path.join(LOG_COLLECTOR_DIR, "results.txt")
+OUTPUT_ARCHIVE_PATH = os.path.join(LOG_COLLECTOR_DIR, "logs.tar")
+COMPRESSED_ARCHIVE_PATH = os.path.join(LOG_COLLECTOR_DIR, "logs.zip")
+
+MUST_COLLECT_FILES = [
```
pgombar, line 40: These files will be given priority if the total size of files to be collected exceeds `ARCHIVE_SIZE_LIMIT`. The higher the file is in this list, the higher its priority. For example, `/var/log/waagent*` files will be collected before `/var/lib/waagent/GoalState.*.xml' files.

azurelinuxagent/common/logcollector.py
pgombar: We expect the difference in files between two collection times (given a collection interval of 1h) to be mostly new files (new GoalState.xml, etc.) or updated files (any log file will have newer entries).

Most of the files will remain the same (for example, old GoalState.xml of Config.xml files), so we avoid collecting them again every single call. They are kept in the archive from last time.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+
+        return files_to_collect
+
+    @staticmethod
+    def _truncate_large_file(file_path):
+        # Truncate large file to size limit (keep freshest entries of the file), copy file to a temporary location
+        # and update file path in list of files to collect
+        try:
+            # Binary files cannot be truncated, don't include large binary files
+            if os.path.splitext(file_path)[1] == ".gz":
```
pgombar, line 243: TODO: need to discard `.zip` files too (our history folder contains zip files).

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+        LogCollector._log_to_results_file(message)
+
+    @staticmethod
+    def _parse_copy_command(path):
+        file_paths = LogCollector._expand_path(path)
+        LogCollector._log_to_results_file(file_paths)
+        return file_paths
+
+    @staticmethod
+    def _convert_file_name_to_archive(file_name):
```
narrieta, line 111: can you add comments as to what "file name" and "archive name" are? i know what they are, but it is not obvious from the code

config/logcollector_manifest_normal
larohra: Why not keep it simple and just collect full everytime? We already have the MUST_HAVE files given priority

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+TRUNCATED_FILES_DIR = '/var/truncated'
+
+OUTPUT_RESULTS_FILE_PATH = os.path.join(LOG_COLLECTOR_DIR, "results.txt")
+OUTPUT_ARCHIVE_PATH = os.path.join(LOG_COLLECTOR_DIR, "logs.tar")
+COMPRESSED_ARCHIVE_PATH = os.path.join(LOG_COLLECTOR_DIR, "logs.zip")
+
+MUST_COLLECT_FILES = [
+    '/var/log/waagent*',
+    '/var/lib/waagent/GoalState.*.xml',
+    '/var/lib/waagent/waagent_status.json',
```
narrieta, line 43: we should avoid hardcoding /var/lib/waagent, etc. maybe the idea of matching the Inspect Disk manifest does not work as well as we expected. ID uses static paths, but our log collector should be more dynamic, allowing us to change those directories should we wish to (customer have been asking us to get out of /var, for example)

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+
+    @staticmethod
+    def _convert_archive_name_to_file_name(archive_name):
+        truncated_prefix = "truncated_"
+        if archive_name.startswith(truncated_prefix):
+            # truncated_var_log_syslog.1 becomes /var/truncated/var/log/syslog.1
+            file_name = archive_name[len(truncated_prefix):].replace("_", os.path.sep)
+            original_file_name = os.path.join(TRUNCATED_FILES_DIR, file_name.lstrip(os.path.sep))
+            return original_file_name
+        else:
```
narrieta, line 128: i don't get the intention of the else part... what kind of paths are we expecting here?

```
>>> '/var/log/waagent.log'.lstrip(os.path.sep)
'var/log/waagent.log'
>>> '/waagent.log'.lstrip(os.path.sep)
'waagent.log'
>>> 'waagent.log'.lstrip(os.path.sep)
'waagent.log'
>>> 

```

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+            return file_name.lstrip(os.path.sep)
+
+    @staticmethod
+    def _convert_archive_name_to_file_name(archive_name):
+        truncated_prefix = "truncated_"
+        if archive_name.startswith(truncated_prefix):
+            # truncated_var_log_syslog.1 becomes /var/truncated/var/log/syslog.1
+            file_name = archive_name[len(truncated_prefix):].replace("_", os.path.sep)
+            original_file_name = os.path.join(TRUNCATED_FILES_DIR, file_name.lstrip(os.path.sep))
```
narrieta, line 126: os.path.split instead?

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+        truncated_files = os.listdir(TRUNCATED_FILES_DIR)
+
+        for file_path in truncated_files:
+            full_path = os.path.join(TRUNCATED_FILES_DIR, file_path)
+            if full_path not in files_to_collect:
+                rm_files(full_path)
+
+    @staticmethod
+    def _is_file_updated(file_name, archive_file):
+        # A file is updated if either its size or last modified time changed.
```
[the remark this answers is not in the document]
narrieta: +1

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+        mtime_tarball = datetime.fromtimestamp(archive_file.mtime).replace(microsecond=0)
```
narrieta, line 146: why do you need the conversion to datetime?

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+    @staticmethod
+    def _get_list_of_files_in_archive():
+        with tarfile.open(OUTPUT_ARCHIVE_PATH, "a") as tarball:
+            return tarball.getnames()
+
+    @staticmethod
+    def _delete_file_from_archive(file):
+        try:
+            command_string = "tar --file {0} --delete {1}".format(OUTPUT_ARCHIVE_PATH, file)
+            command = safe_shlex_split(command_string)
```
narrieta, line 163: why safe_shlex_split?

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+        return files_to_collect
+
+    def collect_logs(self):
+        """
+        Public method that collects necessary log files in a tarball that is updated each time this method is invoked.
+        The tarball is then compressed into a zip.
+        :return: Returns True if the log collection succeeded
+        """
+        try:
+            # Clear previous run's output and create base directories if they don't exist already
```
narrieta, line 333: you should clean up as soon as you're done creating the zip instead of cleaning on the next run

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+            self._log_to_results_file("\n### Compressing archive ###")
+            with zipfile.ZipFile(COMPRESSED_ARCHIVE_PATH, "w", compression=zipfile.ZIP_DEFLATED) as compressed_archive:
+                compressed_archive.write(OUTPUT_ARCHIVE_PATH, arcname="logs.tar")
+
+            tar_size = os.path.getsize(OUTPUT_ARCHIVE_PATH)
+            zip_size = os.path.getsize(COMPRESSED_ARCHIVE_PATH)
+            self._log_to_results_file("Uncompressed archive {0} size: {1}b".format(OUTPUT_ARCHIVE_PATH, tar_size))
+            self._log_to_results_file("Compressed archive {0} size: {1}b".format(COMPRESSED_ARCHIVE_PATH, zip_size))
+
+            self._remove_uncollected_truncated_files(files_to_collect)
```
narrieta, line 353: should be done within a finally clause

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+            expanded_paths = self._expand_path(path)
+            manifest.extend(expanded_paths)
+
+        return manifest
+
+    def _read_manifest_file(self):
+        return read_file(self.manifest_file_path).splitlines()
+
+    @staticmethod
+    def _log_to_results_file(entry):
```
narrieta, line 75: why not use python's logging facilities for this (e.g. errors should be clearly marked)

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+    '/var/lib/waagent/history/*.zip',
+    '/var/log/azure/*/*',
+    '/var/log/azure/*/*/*',
+    '/var/lib/waagent/ExtensionsConfig.*.xml',
+    '/var/lib/waagent/HostingEnvironmentConfig.xml',
+    '/var/lib/waagent/error.json'
+]
+
+FILE_SIZE_LIMIT = 30 * 1024 * 1024  # 30 MB
+ARCHIVE_SIZE_LIMIT = 150 * 1024 * 1024  # 150 MB
```
narrieta, line 53: let's add 'uncompressed' to the name

azurelinuxagent/common/logcollector.py
narrieta, line 243: i've also seen '.xz'

azurelinuxagent/common/logcollector.py
narrieta, line 338: we need to be careful on how we use disk space, especially since you are doing this on /var, which we know can be small and the agent tends to fill up

it looks to me that at some point we may have up to 4 copies of some files: the original files, truncated versions, archived versions, and compressed versions.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+        try:
+            # Clear previous run's output and create base directories if they don't exist already
+            rm_files(OUTPUT_RESULTS_FILE_PATH)
+            mkdir(TRUNCATED_FILES_DIR)
+            mkdir(LOG_COLLECTOR_DIR)
+
+            files_to_collect = self._create_list_of_files_to_collect()
+            self._log_to_results_file("\n### Creating archive ###")
+
+            self._remove_deleted_files_from_archive(files_to_collect)
```
narrieta, line 341: what is the value added of the archive? i'm not sure it's buying us anything, and probably it is adding a little overhead plus extra code

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+            self._update_files_in_archive(files_to_collect)
+
+            self._log_to_results_file("\n### Compressing archive ###")
+            with zipfile.ZipFile(COMPRESSED_ARCHIVE_PATH, "w", compression=zipfile.ZIP_DEFLATED) as compressed_archive:
+                compressed_archive.write(OUTPUT_ARCHIVE_PATH, arcname="logs.tar")
+
+            tar_size = os.path.getsize(OUTPUT_ARCHIVE_PATH)
+            zip_size = os.path.getsize(COMPRESSED_ARCHIVE_PATH)
+            self._log_to_results_file("Uncompressed archive {0} size: {1}b".format(OUTPUT_ARCHIVE_PATH, tar_size))
```
narrieta, line 350: we should add the results file to the zip

tests/common/test_logcollector.py
```diff
@@ -0,0 +1,440 @@
...
+        self._assert_files_are_not_in_archive(unexpected_files)
+
+        self._assert_archive_created(second_archive)
+
+        no_files = self._get_number_of_files_in_archive()
+        self.assertEquals(5, no_files, "Expected 5 files in archive, found {0}!".format(no_files))
+        self._assert_archive_sizes()
+
+    def test_log_collector_should_update_archive_when_files_are_new_or_modified_or_deleted(self):
+        # Ensure the archive reflects the state of files on the disk at collection time. If a file was updated, it
```
narrieta, line 330: seems to me like the archive is an internal implementation detail... should this test be done in terms of the zip?

tests/tools.py
```diff
@@ -168,6 +168,10 @@ def i_am_root():
     return os.geteuid() == 0
 
 
+def is_python_version_26():
```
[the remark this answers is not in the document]
pgombar: I needed this function for the tests I was writing and noticed it already existed in `test_crypt_util.py` so I refactored it into this common place for easier re-usability. `tools.py` seemed like a good place since it already contains other helper methods for our tests. Do you have a different idea in mind?

azurelinuxagent/common/utils/fileutil.py
```diff
@@ -242,3 +242,9 @@ def clean_ioerror(e, paths=[]):
             except Exception:
                 # swallow exception
                 pass
+
+
+def append_items_to_file(file_path, iterable):
```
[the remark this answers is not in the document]
pgombar: Thanks for the comment, I totally missed adding documentation and unit tests, will update this, as well as add encoding.

> how would be the value added of this function versus write_file(append=True)?

My use case is writing a list of items to a file where each entry is in its own line. The method `write_file` accepts a string, `contents`, meaning I would have to pre-construct a string of all list items (and newlines) before passing it to the method. I prefer avoiding constructing a (potentially) large string in place and instead iterate over members and output them to the file separately.

azurelinuxagent/common/utils/shellutil.py
```diff
@@ -141,7 +144,7 @@ def format_command(cmd):
         return " ".join(cmd) if isinstance(cmd, list) else command
 
     try:
-        process = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=False)
+        process = subprocess.Popen(command, stdin=stdin, stdout=stdout, stderr=stderr, shell=False)
```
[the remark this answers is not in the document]
pgombar: As discussed offline, I won't use this `run_command` to invoke shell commands from the log collector utility, I will create a designated one.

azurelinuxagent/common/utils/shellutil.py
```diff
@@ -111,7 +111,10 @@ def run_get_output(cmd, chk_err=True, log_cmd=True, expected_errors=[]):
 
 
 def _encode_command_output(output):
-    return ustr(output, encoding='utf-8', errors="backslashreplace")
+    if output:
```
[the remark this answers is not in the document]
pgombar: Removing this since I won't be using `run_command` for running shell commands from the log collector utility. I needed this because `stdout` or `stderr` were `None` in some of my use cases.

azurelinuxagent/common/utils/shellutil.py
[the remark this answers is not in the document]
pgombar: I will be adding my own wrapper around `subprocess.Popen` to invoke shell commands for the log collector utility. 
Can you elaborate on this please? I don't understand the question, what do you mean by log output? Maybe you are pointing out something I oversaw and that would affect the design of my new wrapper function.

config/logcollector_manifest_normal
pgombar: The periodic collection should not be full due to resource usage on the VM, but we want to provide the full option for on-demand use of the tool as a command line utility for support engineers to use to pull a full set of logs from the VM. These two options are already given by Windows Agent and its log collection tool.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+        #    the size limit.
+        parsed_file_paths = self._parse_manifest_file()
+        prioritized_file_paths = self._get_priority_files_list(parsed_file_paths)
+        files_to_collect = self._get_final_list_for_archive(prioritized_file_paths)
+        return files_to_collect
+
+    def collect_logs(self):
+        """
+        Public method that collects necessary log files in a tarball that is updated each time this method is invoked.
+        The tarball is then compressed into a zip.
```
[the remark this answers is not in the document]
pgombar: Yes, 7zip should do it. That is true for IID payload today, so this is not a new requirement.

azurelinuxagent/common/logcollector.py
pgombar: I agree with the sentiment and will add support for that, but I would like to keep the hard-coded values for the v1 of this feature for two reasons:
1) `/var/lib/waagent` is still the most commonly used location for agent files (I haven't seen this modified once so far)
2) it would be beneficial to release the v1 for this feature as soon as possible so we can get real feedback and telemetry from live VMs and users quicker and then decide on areas for improvement.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+import tarfile
+import zipfile
+
+from azurelinuxagent.common.future import ustr
+from azurelinuxagent.common.utils.fileutil import mkdir, read_file, rm_files, append_file, append_items_to_file
+from azurelinuxagent.common.utils.shellutil import run_command
+from azurelinuxagent.common.utils.textutil import safe_shlex_split
+
+
+LOG_COLLECTOR_DIR = '/var/lib/waagent/logcollector'
```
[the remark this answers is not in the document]
pgombar: See my other comment, I would like to keep this hard-coded for v1 and add support for this in the next iteration.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+TRUNCATED_FILES_DIR = '/var/truncated'
+
+OUTPUT_RESULTS_FILE_PATH = os.path.join(LOG_COLLECTOR_DIR, "results.txt")
+OUTPUT_ARCHIVE_PATH = os.path.join(LOG_COLLECTOR_DIR, "logs.tar")
+COMPRESSED_ARCHIVE_PATH = os.path.join(LOG_COLLECTOR_DIR, "logs.zip")
```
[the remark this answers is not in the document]
pgombar: Good catch, no. Just the final compressed archive.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+MUST_COLLECT_FILES = [
+    '/var/log/waagent*',
+    '/var/lib/waagent/GoalState.*.xml',
+    '/var/lib/waagent/waagent_status.json',
+    '/var/lib/waagent/history/*.zip',
+    '/var/log/azure/*/*',
+    '/var/log/azure/*/*/*',
+    '/var/lib/waagent/ExtensionsConfig.*.xml',
+    '/var/lib/waagent/HostingEnvironmentConfig.xml',
+    '/var/lib/waagent/error.json'
```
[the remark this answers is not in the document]
pgombar: It's a small JSON file where we store when an agent version is blacklisted. I thought it couldn't hurt since it's a small file, but if you think it's redundant, I can remove it from the must-haves.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+
+LOG_COLLECTOR_DIR = '/var/lib/waagent/logcollector'
+TRUNCATED_FILES_DIR = '/var/truncated'
+
+OUTPUT_RESULTS_FILE_PATH = os.path.join(LOG_COLLECTOR_DIR, "results.txt")
+OUTPUT_ARCHIVE_PATH = os.path.join(LOG_COLLECTOR_DIR, "logs.tar")
+COMPRESSED_ARCHIVE_PATH = os.path.join(LOG_COLLECTOR_DIR, "logs.zip")
+
+MUST_COLLECT_FILES = [
+    '/var/log/waagent*',
```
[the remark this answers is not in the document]
pgombar: Hm, there can be quite a few since they are rotated. For the must-have then, I can just limit it to `waagent.log`.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+    '/var/lib/waagent/GoalState.*.xml',
+    '/var/lib/waagent/waagent_status.json',
+    '/var/lib/waagent/history/*.zip',
+    '/var/log/azure/*/*',
+    '/var/log/azure/*/*/*',
+    '/var/lib/waagent/ExtensionsConfig.*.xml',
```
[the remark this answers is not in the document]
pgombar: I can move `ExtensionsConfig` to be just after `GoalState` files, this wasn't intentional. What files do you see missing for must-haves?

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+import zipfile
+
+from azurelinuxagent.common.future import ustr
+from azurelinuxagent.common.utils.fileutil import mkdir, read_file, rm_files, append_file, append_items_to_file
+from azurelinuxagent.common.utils.shellutil import run_command
+from azurelinuxagent.common.utils.textutil import safe_shlex_split
+
+
+LOG_COLLECTOR_DIR = '/var/lib/waagent/logcollector'
+TRUNCATED_FILES_DIR = '/var/truncated'
```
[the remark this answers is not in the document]
pgombar: According to https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard, it seemed like the best candidate:

/var | Variable files—files whose content is expected to continually change during normal operation of the system—such as logs, spool files, and temporary e-mail files.
-- | --


It seemed better than `/tmp` or any other alternative:

/tmp | Temporary files (see also /var/tmp). Often not preserved between system reboots, and may be severely size restricted.
-- | --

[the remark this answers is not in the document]
pgombar: See my reply below, this seemed like the best fit according to https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+            manifest.extend(expanded_paths)
+
+        return manifest
+
+    def _read_manifest_file(self):
+        return read_file(self.manifest_file_path).splitlines()
+
+    @staticmethod
+    def _log_to_results_file(entry):
+        if type(entry) == ustr or type(entry) == str:
```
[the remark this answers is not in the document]
pgombar: Made the check for arrays instead, looks much cleaner, thanks!

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+            append_file(OUTPUT_RESULTS_FILE_PATH, entry + "\n")
+        else:
+            append_items_to_file(OUTPUT_RESULTS_FILE_PATH, entry)
+
+    @staticmethod
+    def _expand_path(path):
```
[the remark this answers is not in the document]
pgombar: You're right, simplified it now.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+        if os.path.exists(path):
+            expanded.append(path)
+        else:
+            paths = glob.glob(path)
+            if len(paths) > 0:
+                expanded.extend(sorted(paths))
+        return expanded
+
+    @staticmethod
+    def _parse_ll_command(folder):
```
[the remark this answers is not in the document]
pgombar: From the perspective of the manifest, it's parsing the line with `ll` command. I'll rename it to `_process_ll_command` to make it clearer.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+        output = run_command(["ls", "-alF", folder])
+        header = "Output of \"ll {0}\":\n".format(folder)
+        LogCollector._log_to_results_file(header + output)
+
+    @staticmethod
+    def _parse_echo_command(message):
```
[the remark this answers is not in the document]
pgombar: The method is processing the command `echo $message` from the manifest, so we just output the `$message` to the log file. I will rename this method to `_process_echo_command` to make it clearer.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+        LogCollector._log_to_results_file(message)
+
+    @staticmethod
+    def _parse_copy_command(path):
```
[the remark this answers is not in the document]
pgombar: Renaming to `_process_copy_command`.

azurelinuxagent/common/logcollector.py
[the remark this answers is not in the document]
pgombar: Good point, updating it to only look at last modified time.

azurelinuxagent/common/logcollector.py
pgombar: The granularity is different, `os.path.getmtime` returns microseconds, whereas `zipfile` and `tarfile` do not.
```
os.path.getmtime('/var/lib/waagent/logcollector/logs.tar')
1586245322.2551062

tarfile.mtime
1586245157

zipinfo.date_time
(2020, 4, 7, 0, 42, 2)
```

azurelinuxagent/common/logcollector.py
pgombar: Removing. While trying out these commands, it was quicker to copy-paste from command line and do the splitting  separately.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+
+                # If the original file hasn't been updated since the truncated file, it means there were no changes
+                # and we don't need to truncate it again.
+                if original_file_mtime < truncated_file_mtime:
+                    return truncated_file_path
+
+            # Get the last N bytes of the file
+            with open(truncated_file_path, "w+") as fh:
+                command_string = "tail -c {0} {1}".format(FILE_SIZE_LIMIT, file_path)
+                command = safe_shlex_split(command_string)
```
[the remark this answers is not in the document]
pgombar: Removing, tokenizing manually instead.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+                run_command(command, stdout=fh)
```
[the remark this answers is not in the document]
pgombar: Updating. It was easier to copy-paste from command line while testing and let the function tokenize it instead.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+            expanded.append(path)
+        else:
+            paths = glob.glob(path)
+            if len(paths) > 0:
+                expanded.extend(sorted(paths))
+        return expanded
+
+    @staticmethod
+    def _parse_ll_command(folder):
+        output = run_command(["ls", "-alF", folder])
```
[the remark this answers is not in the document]
pgombar: Refactored this part, now the exceptions will be handled. Thanks for catching this!

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+        manifest_entries = self._read_manifest_file()
+
+        for entry in manifest_entries:
+            # The entry can be one of the four flavours:
+            # 1) ll,/etc/udev/rules.d -- list out contents of the folder and store to results file
+            # 2) echo,### Gathering Configuration Files ### -- print message to results file
+            # 3) copy,/var/lib/waagent/provisioned -- add file to list of files to be collected
+            # 4) diskinfo, -- ignore other commands for now
+
+            contents = entry.split(",")
```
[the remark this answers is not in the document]
pgombar: I skip anything that is not a `ll`, `echo` or `copy` command.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+
+            truncated_file_path = os.path.join(TRUNCATED_FILES_DIR, file_path.replace(os.path.sep, "_"))
+            if os.path.exists(truncated_file_path):
+                original_file_mtime = os.path.getmtime(file_path)
+                truncated_file_mtime = os.path.getmtime(truncated_file_path)
+
+                # If the original file hasn't been updated since the truncated file, it means there were no changes
+                # and we don't need to truncate it again.
+                if original_file_mtime < truncated_file_mtime:
+                    return truncated_file_path
```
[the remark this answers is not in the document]
pgombar: :smile:

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+
+            # Get the last N bytes of the file
+            with open(truncated_file_path, "w+") as fh:
+                command_string = "tail -c {0} {1}".format(FILE_SIZE_LIMIT, file_path)
+                command = safe_shlex_split(command_string)
+                run_command(command, stdout=fh)
+
+            return truncated_file_path
+        except OSError as e:
```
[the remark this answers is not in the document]
pgombar: Refactored this so now these exceptions are handled. Thanks for catching this!

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+    def _parse_manifest_file(self):
+        files_to_collect = set()
+        manifest_entries = self._read_manifest_file()
+
+        for entry in manifest_entries:
+            # The entry can be one of the four flavours:
+            # 1) ll,/etc/udev/rules.d -- list out contents of the folder and store to results file
+            # 2) echo,### Gathering Configuration Files ### -- print message to results file
+            # 3) copy,/var/lib/waagent/provisioned -- add file to list of files to be collected
+            # 4) diskinfo, -- ignore other commands for now
```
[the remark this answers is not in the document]
pgombar: Since we are using IID's manifest, it may contain other commands needed for IID, but not needed for log collection. I am ignoring those commands until/if we find they are useful for our purposes. I will clarify the comment in the code.

azurelinuxagent/common/logcollector.py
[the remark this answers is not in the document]
pgombar: I will add a log statement to the local file saying there was an error parsing those lines.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+
+            # Get the last N bytes of the file
+            with open(truncated_file_path, "w+") as fh:
+                command_string = "tail -c {0} {1}".format(FILE_SIZE_LIMIT, file_path)
+                command = safe_shlex_split(command_string)
+                run_command(command, stdout=fh)
+
+            return truncated_file_path
+        except OSError as e:
+            LogCollector._log_to_results_file("Failed to truncate large file: {0}".format(ustr(e)))
```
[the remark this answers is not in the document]
pgombar: This module shouldn't depend on the agent's `logger` or waagent.log, it is a standalone module. The results file is added to the zip (even in case of unsuccessful log collection).

azurelinuxagent/common/logcollector.py
pgombar: The idea was to leave the previous run's output for debugging purposes. For example, in a situation where the tool is invoked and the archive couldn't be created, not even the results file will be added to the archive. In that case, there is value in leaving that file on disk instead of deleting it.

azurelinuxagent/common/logcollector.py
pgombar: Both `zipfile` and `tarfile` remove the leading path separator when adding the file to the archive. For example, file `var/log/waagent.log` in the archive is actually `/var/log/waagent.log` on disk.

azurelinuxagent/common/logcollector.py
pgombar: I'm joining paths here, not splitting. A file in the archive called `truncated_var_log_syslog.1` is actually `/var/truncated/var/log/syslog.1` on disk.

azurelinuxagent/common/logcollector.py
pgombar: Added comments.

azurelinuxagent/common/logcollector.py
pgombar: I just ran some basic perf tests and you're right - the main load is the compressing part, not the collecting files part, and since we compress every single time the collection runs, there is no benefit in keeping an intermediate archive.

Reminder, we cannot avoid compressing every collection call since `zipfile` doesn't support removing files from a compressed archive. This means we lose the ability to modify the compressed archive and make it incrementally updated, hence the need to create a compressed archive from scratch every single time.

I am changing the approach and removing the intermediate archive. Thanks for pointing it out.

azurelinuxagent/common/logcollector.py
pgombar: Replied below - changing the approach to remove the intermediate archive, so at worst we will now keep 3 copies of the file: original, truncated, compressed.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,360 @@
...
+                LogCollector._log_to_results_file("Updating archive, removing deleted file {0}".format(archive_file))
+                LogCollector._delete_file_from_archive(archive_file)
+
+    @staticmethod
+    def _add_file_to_archive(file_name, archive_file_name):
+        with tarfile.open(OUTPUT_ARCHIVE_PATH, "a") as archive:
+            archive.add(file_name, arcname=archive_file_name)
+
+    @staticmethod
+    def _get_file_from_archive(archive_file_name):
```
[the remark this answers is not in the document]
pgombar: Removed.

azurelinuxagent/common/logcollector.py
pgombar: Removed.

azurelinuxagent/common/logcollector.py
pgombar: Removed.

azurelinuxagent/common/logcollector.py
pgombar: Removed.

azurelinuxagent/common/logcollector.py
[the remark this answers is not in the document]
pgombar: This list is meant to be static and internal to the module. I don't think we should be exposing this in a separate file.

azurelinuxagent/common/logcollector.py
pgombar: Didn't occur to me to use that module, honestly. I was mimicking the output of IID and their results file which is just a simple outline of operations performed. I don't think there's added value in using Python's logger in this case.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,302 @@
...
+        total_uncompressed_size = 0
+        final_files_to_collect = []
+
+        while priority_file_queue:
+            file_path = heappop(priority_file_queue)[1]  # (priority, file_path)
+            file_size = min(os.path.getsize(file_path), _FILE_SIZE_LIMIT)
+
+            if total_uncompressed_size + file_size > _UNCOMPRESSED_ARCHIVE_SIZE_LIMIT:
+                self._log_to_results_file("Archive too big, done with adding files.")
+                break
```
narrieta, line 240: you need to truncate first, then do the size check, right?

azurelinuxagent/common/logcollector.py
narrieta: what's the logic for postponing it?

azurelinuxagent/common/logcollector.py
narrieta: k

tests/tools.py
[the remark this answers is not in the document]
narrieta: Yes, create a separate module.

tests/common/test_logcollector.py
```diff
@@ -0,0 +1,435 @@
...
+
+
+@skip_if_predicate_true(is_python_version_26, "Disabled on Python 2.6")
+class TestLogCollector(AgentTestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        AgentTestCase.setUpClass()
+
+        prefix = "{0}_".format(cls.__class__.__name__)
```
narrieta, line 42: what's the purpose of the prefix?

tests/common/test_logcollector.py
pgombar: Forgot to resolve, archive step is gone so the tests assert on the zip anyway.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,302 @@
...
+            process = subprocess.Popen(command, stdin=subprocess.PIPE, stdout=stdout, stderr=subprocess.PIPE, shell=False)
+            stdout, stderr = process.communicate()
+            return_code = process.returncode
+        except Exception as e:
+            error_msg = u"Command [{0}] raised unexpected exception: [{1}]".format(format_command(command), ustr(e))
+            LogCollector._log_to_results_file(error_msg)
+            return
+
+        if return_code != 0:
+            encoded_stdout = _encode_command_output(stdout)
```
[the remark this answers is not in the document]
larohra: I think errors should always be logged irrespective of the value of `output` variable.

azurelinuxagent/common/logcollector.py
larohra: It might be worth checking out the logging library though mainly because it gives you a lot of additional functionalities for free. 
For example I noticed that you're not logging the timestamp to the `results_file` which would be really helpful during debugging if something goes wrong. You can always add it here but it might be a simpler change to do it if using a module vs modifying the custom implementation.

azurelinuxagent/common/logcollector.py
```diff
@@ -73,141 +102,50 @@ def _read_manifest_file(self):
 
     @staticmethod
     def _log_to_results_file(entry):
-        if type(entry) == ustr or type(entry) == str:
-            append_file(OUTPUT_RESULTS_FILE_PATH, entry + "\n")
+        if isinstance(entry, list):
+            append_items_to_file(_OUTPUT_RESULTS_FILE_PATH, entry)
         else:
-            append_items_to_file(OUTPUT_RESULTS_FILE_PATH, entry)
+            append_file(_OUTPUT_RESULTS_FILE_PATH, entry + "\n")
```
larohra, line 108: I noticed that you're not logging the timestamp in the log file, it would be awesome if you could add that too. Thanks! :)

azurelinuxagent/common/logcollector.py
pgombar: Already handled in L236, `file_size = min(os.path.getsize(file_path), _FILE_SIZE_LIMIT)`.

azurelinuxagent/common/logcollector.py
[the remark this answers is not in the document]
pgombar: This was my intention, to always log if the subprocess exited with non-zero.

tests/common/test_logcollector.py
pgombar: Just to help with debugging, I find it easier to match the temp folder to the tests I'm interested in (when using breakpoints) since it's named after the class. We use it in a couple of other test classes too.

tests/tools.py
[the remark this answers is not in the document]
pgombar: Will send a separate PR for this.

azurelinuxagent/common/logcollector.py
pgombar: Added since we already have the API for this

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,302 @@
...
+
+_LOG_COLLECTOR_DIR = '/var/lib/waagent/logcollector'
+_TRUNCATED_FILES_DIR = '/var/truncated'
+
+_OUTPUT_RESULTS_FILE_PATH = os.path.join(_LOG_COLLECTOR_DIR, "results.txt")
+COMPRESSED_ARCHIVE_PATH = os.path.join(_LOG_COLLECTOR_DIR, "logs.zip")
+
+_MUST_COLLECT_FILES = [
+    '/var/log/waagent.log',
+    '/var/lib/waagent/GoalState.*.xml',
```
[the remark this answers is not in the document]
pgombar: Expanded to all but Certificates

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,302 @@
...
+    def _process_ll_command(folder):
+        LogCollector.run_shell_command(["ls", "-alF", folder], output=True)
+
+    @staticmethod
+    def _process_echo_command(message):
+        LogCollector._log_to_results_file(message)
+
+    @staticmethod
+    def _process_copy_command(path):
+        file_paths = glob.glob(path)
```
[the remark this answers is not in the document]
pgombar: The returning value of this function (glob of the given path) is used to expand the list of files to collect. This is done by the caller of this function.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,302 @@
...
+        LogCollector._log_to_results_file(file_paths)
+        return file_paths
```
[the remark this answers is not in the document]
pgombar: `copy` comes from the IID command in the manifest, e.g. `copy,/var/log/waagent.log` where the intended usage is to collect that file. In my implementation, that means adding that file to the file list and logging that addition to the results file, since I collect all the files in the end, after finalizing the file list. I've named all these helper processing functions to match the commands from the manifest (copy, echo, ll).

azurelinuxagent/common/logcollector.py
pgombar: Thanks for catching that, definitely agree it's useful. Added.

azurelinuxagent/common/logcollector.py
pgombar: Added.

azurelinuxagent/common/logcollector.py
```diff
@@ -19,59 +19,83 @@
...
-from azurelinuxagent.common.utils.fileutil import mkdir, read_file, rm_files, append_file, append_items_to_file
+from azurelinuxagent.common.utils.fileutil import mkdir, read_file, write_file, rm_files
 
+_EXTENSION_LOG_DIR = get_ext_log_dir()
+_AGENT_LIB_DIR = get_lib_dir()
 
-_LOG_COLLECTOR_DIR = '/var/lib/waagent/logcollector'
-_TRUNCATED_FILES_DIR = '/var/truncated'
+_LOG_COLLECTOR_DIR = os.path.join(_AGENT_LIB_DIR, "logcollector")
+_TRUNCATED_FILES_DIR = "/var/truncated"
```
narrieta, line 35: why keeping this outside the agent's directory?

azurelinuxagent/common/logcollector.py
```diff
@@ -19,59 +19,83 @@
...
-    '/var/log/waagent.log',
-    '/var/lib/waagent/GoalState.*.xml',
-    '/var/lib/waagent/ExtensionsConfig.*.xml',
-    '/var/lib/waagent/waagent_status.json',
-    '/var/lib/waagent/history/*.zip',
-    '/var/log/azure/*/*',
-    '/var/log/azure/*/*/*',
-    '/var/lib/waagent/HostingEnvironmentConfig.xml',
-    '/var/log/waagent*',
+    "/var/log/waagent.log",
```
narrieta, line 41: shouldn't hardcode this path

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,340 @@
...
+import glob
+from heapq import heappush, heappop
+import logging
+import os
+import subprocess
+import zipfile
+
+from azurelinuxagent.common.conf import get_lib_dir, get_ext_log_dir
+from azurelinuxagent.common.future import ustr
+from azurelinuxagent.common.utils.fileutil import mkdir, read_file, write_file, rm_files
```
narrieta, line 29: We need to be very careful adding dependencies to other modules in the agent, since the agent uses several globals. For example, if at some point we add a dependency on a function that uses logger then that output would be in the agent's log and may be confusing. Telemetry is also global. Your current dependencies look pretty safe, but if it was me I would get rid of the dependency on fileutil since I do not think those functions are buying you much.

Also, please add this note as a comment so that people changing the code do not add undesired dependencies.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,340 @@
...
+        _f_handler.setFormatter(_f_format)
+        _LOGGER.addHandler(_f_handler)
+        _LOGGER.setLevel(logging.INFO)
+
+    @staticmethod
+    def _run_shell_command(command, stdout=subprocess.PIPE, output=False):
+        def format_command(cmd):
+            return " ".join(cmd) if isinstance(cmd, list) else command
+
+        def _encode_command_output(output):
```
narrieta, line 89: why is this encoding needed?

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,340 @@
...
+    def _expand_must_collect_files():
+        # Match the regexes from the MUST_COLLECT_FILES list to existing file paths on disk.
+        manifest = []
+        for path in _MUST_COLLECT_FILES:
+            manifest.extend(sorted(glob.glob(path)))
+
+        return manifest
+
+    @staticmethod
+    def _parametrize_filepath(path):
```
narrieta, line 125: your code would be simpler and less prone to mistakes if you allow something like %lib_dir%/foo.xml instead of /var/lib/foo.xml. then when you read the manifest you can expand %lib_dir% instead of having to remember that one needs to do this parameterization each time a file path is used

azurelinuxagent/common/logcollector.py
```diff
@@ -19,59 +19,83 @@
...
+    def _create_base_dirs():
+        mkdir(_TRUNCATED_FILES_DIR)
+        mkdir(_LOG_COLLECTOR_DIR)
 
     @staticmethod
-    def run_shell_command(command, stdout=subprocess.PIPE, output=False):
+    def _set_logger():
+        _f_handler = logging.FileHandler(_OUTPUT_RESULTS_FILE_PATH, encoding="utf-8")
+        _f_format = logging.Formatter(fmt='%(asctime)s %(levelname)s %(message)s',
+                                      datefmt='%Y-%m-%d %H:%M:%S')
```
larohra, line 79: Could you make the dateformat consistent to our logs pls - 
`(u'%Y-%m-%dT%H:%M:%S.%fZ')`

larohra: You should also set this to ensure that the time is always in UTC - 
`logging.Formatter.converter = time.gmtime`

azurelinuxagent/common/logcollector.py
```diff
@@ -97,28 +121,36 @@ def _expand_must_collect_files():
...
         return manifest
 
-    def _read_manifest_file(self):
-        return read_file(self.manifest_file_path).splitlines()
-
     @staticmethod
-    def _log_to_results_file(entry):
-        if isinstance(entry, list):
-            append_items_to_file(_OUTPUT_RESULTS_FILE_PATH, entry)
+    def _parametrize_filepath(path):
```
larohra, line 125: Suggestion: Why not add macros (like `lib`, `log` and `extensionLog`) and just make the manifest relative to those.
Eg: 
```
copy,lib/provisioned -> points to /var/lib/waagent/provisioned
copy,lib/GoalState.*.xml -> points to ,/var/lib/waagent/GoalState.*.xml
copy,extensionLog/custom-script/handler.log -> points to /var/log/azure/custom-script/handler.log
```

This way we can always just replace the macros with whatever the standard in that particular VM

azurelinuxagent/common/logcollector.py
```diff
@@ -274,28 +313,27 @@ def collect_logs(self):
         files_to_collect = []
 
         try:
-            # Clear previous run's output and create base directories if they don't exist already
-            rm_files(_OUTPUT_RESULTS_FILE_PATH)
-            mkdir(_TRUNCATED_FILES_DIR)
-            mkdir(_LOG_COLLECTOR_DIR)
+            # Clear previous run's output and create base directories if they don't exist already.
+            self._create_base_dirs()
+            write_file(_OUTPUT_RESULTS_FILE_PATH, "")
```
larohra, line 318: Wont the logger take care of creating this?

azurelinuxagent/common/logcollector.py
pgombar: Thanks, good suggestions! Updated.

azurelinuxagent/common/logcollector.py
pgombar: The logger wouldn't clear the file, just keep appending to it. I am keeping the `results.txt` file once per run.

azurelinuxagent/common/logcollector.py
pgombar: This is hardcoded throughout the agent. Unlike `/var/lib/waagent` and `/var/log/azure`, we don't expose a parameter for agent log file.

larohra: I remember this was part of the log-rotate PR that Varad was working on. It might be useful for you to parameterize it now and reuse the same one for the other PR (or do it the other way round), whatever works for you I guess.

azurelinuxagent/common/logcollector.py
larohra: Ahh ok. 
A very small NIT: Add a wrapper function over this called `reset_file` or something to make it a bit more clearer. Totally fine if you dont do it though, just looks more cleaner to me :)

azurelinuxagent/common/logcollector.py
pgombar: Thanks, this is a very good point that hadn't crossed my mind. I agree that the fileutil dependencies are not worth keeping, I will refactor them.

azurelinuxagent/common/logcollector.py
pgombar: I maintain utf-8 encoding throughout the module (when reading files and when writing to them), so I wanted to preserve this behaviour for outputs out of my control, e.g. outputs from shell commands or stack traces. Seems safe to keep it?

azurelinuxagent/common/logcollector.py
narrieta: > This is hardcoded throughout the agent. 

It shouldn't. Please add a function similar to get_ext_log_dir and let's start using it in new code

azurelinuxagent/common/logcollector.py
narrieta: But you are creating a utf8-encoded log file already. Why would this extra encoding would be needed?

azurelinuxagent/common/logcollector.py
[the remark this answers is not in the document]
narrieta: can you add a docstring describing what the 'output' parameter means?

azurelinuxagent/common/logcollector.py
pgombar: Done, thanks.

azurelinuxagent/common/logcollector.py
pgombar: Done, thanks.

azurelinuxagent/common/logcollector.py
pgombar: `get_ext_log_dir` reads from the config file, and currently we don't have an entry in the config file for agent log file. I wouldn't want to do that change in the scope of this PR. What I can do is parameterize it for the log collector (add `$AGENT_LOG` to the manifests) and then define the the value in this module as `/var/log/waagent.log` until we make the change throughout the agent, including the config file.

azurelinuxagent/common/logcollector.py
pgombar: Renamed for clarity.

azurelinuxagent/common/logcollector.py
pgombar: Moved to /var/lib/waagent/logcollector/truncated/

azurelinuxagent/common/logcollector.py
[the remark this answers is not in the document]
pgombar: Done.

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,377 @@
...
+    def _expand_must_collect_files():
+        # Match the regexes from the MUST_COLLECT_FILES list to existing file paths on disk.
+        manifest = []
+        for path in _MUST_COLLECT_FILES:
+            manifest.extend(sorted(glob.glob(path)))
+
+        return manifest
+
+    @staticmethod
+    def _parametrize_filepath(path):
```
larohra, line 147: Why is this function needed now since you've already parameterized it earlier?

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,377 @@
...
+
+# Please note: be careful when adding agent dependencies in this module.
+# This module uses its own logger and logs to its own file, not to the agent log.
+
+from azurelinuxagent.common.conf import get_lib_dir, get_ext_log_dir
+from azurelinuxagent.common.future import ustr
+
+_EXTENSION_LOG_DIR = get_ext_log_dir()
+_AGENT_LIB_DIR = get_lib_dir()
+_AGENT_LOG = "/var/log/waagent.log"
```
larohra, line 36: Similar to `get_lib_dir()` function, could you also add a function `get_agent_log_file()` or something and replace all occurances of `/var/log/waagent.log` with it? 
Or should that be done in a separate PR?

azurelinuxagent/common/logcollector.py
```diff
@@ -0,0 +1,377 @@
...
+
+    @staticmethod
+    def _expand_parameters(manifest_data):
+        _LOGGER.info("Using {0} as $LIB_DIR".format(_AGENT_LIB_DIR))
+        _LOGGER.info("Using {0} as $LOG_DIR".format(_EXTENSION_LOG_DIR))
+        _LOGGER.info("Using {0} as $AGENT_LOG".format(_AGENT_LOG))
+
+        new_manifest = []
+        for line in manifest_data:
+            line.replace("$LIB_DIR", _AGENT_LIB_DIR)
```
larohra, line 218: Don't you wanna replace the `$AGENT_LOG` too?

azurelinuxagent/common/logcollector.py
narrieta: +1

azurelinuxagent/common/logcollector.py
```diff
@@ -51,7 +52,7 @@
     os.path.join(_AGENT_LIB_DIR, "history", "*.zip"),
     os.path.join(_EXTENSION_LOG_DIR, "*", "*"),
     os.path.join(_EXTENSION_LOG_DIR, "*", "*", "*"),
-    "/var/log/waagent*",
+    "{0}*".format(_AGENT_LOG)
```
narrieta, line 55: this will also capture waagent.log.. you may want to use 'waagent\..+' instead

azurelinuxagent/common/logcollector.py
narrieta: +1

pgombar: Not needed, good catch. Forgot to remove it.

azurelinuxagent/common/logcollector.py
pgombar: Thanks, added!

azurelinuxagent/common/logcollector.py
pgombar: Thanks, it's more specific with your suggestion. Updated.

azurelinuxagent/common/logcollector.py
pgombar: I initially wanted to tackle it in a separate PR, but might as well.

azurelinuxagent/common/logcollector.py
```diff
@@ -52,7 +52,7 @@
     os.path.join(_AGENT_LIB_DIR, "history", "*.zip"),
     os.path.join(_EXTENSION_LOG_DIR, "*", "*"),
     os.path.join(_EXTENSION_LOG_DIR, "*", "*", "*"),
-    "{0}*".format(_AGENT_LOG)
+    "{0}.+".format(_AGENT_LOG)  # any additional waagent.log files (e.g., waagent.log.1.gz)
```
narrieta, line 55: `{0}\..+`

azurelinuxagent/common/conf.py
```diff
@@ -243,6 +243,10 @@ def get_ext_log_dir(conf=__conf__):
     return conf.get("Extension.LogDir", "/var/log/azure")
 
 
+def get_agent_log_file():
```
larohra, line 246: Just a bit concerned about putting this here since its not part of the conf (Maybe osutil would be a better place since the other PR eventually plans to move it there).

pgombar: I still have to revisit the logic in the other PR, and if the ultimate goal is to expose the agent log parameter in the config file, this would be the correct place. Of course, this can change, so for now this I settled for this solution.

Issue #1859, opened 2020-04-20
Upgrade WALinuxAgent to support Python 3.8


pgombar:
**Describe the bug: A clear and concise description of what the bug is.**

WALinuxAgent should support Python 3.8.

pgombar:
Fixed in #1860.

PR #1860, opened 2020-04-21
Upgrade unit test to be compatible with Python 3.8

pgombar:
## Description

- Fixed tests that were failing in Python 3.8
- Removed coverage output in Travis runs since the output was getting in the way of readability for unit test runs (codecov still runs on PRs and commits)
- Disabled tests that rely on mocking `platform.dist`. These tests and the method they are testing need to be updated once we figure out a good strategy to support `distro` and get away from `platform`.

Related to #1859.

---

### PR information

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -4,6 +4,6 @@ flake8; python_version >= '2.7'
 mock==2.0.0; python_version == '2.6'
 mock==3.0.5; python_version >= '2.7' and python_version <= '3.5'
 mock==4.0.2; python_version >= '3.6'
-distro; python_version >= '3.7'
+distro; python_version >= '3.8'
 nose
 nose-timer; python_version >= '2.7'
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -4,5 +4,6 @@ flake8; python_version >= '2.7'
 mock==2.0.0; python_version == '2.6'
 mock==3.0.5; python_version >= '2.7' and python_version <= '3.5'
 mock==4.0.2; python_version >= '3.6'
+distro; python_version >= '3.7'
 nose
 nose-timer; python_version >= '2.7'
--- a/tests/common/test_version.py
+++ b/tests/common/test_version.py
@@ -25,8 +25,8 @@ import azurelinuxagent.common.conf as conf
 from azurelinuxagent.common.event import EVENTS_DIRECTORY
 from azurelinuxagent.common.version import set_current_agent, \
     AGENT_LONG_VERSION, AGENT_VERSION, AGENT_NAME, AGENT_NAME_PATTERN, \
-    get_f5_platform, get_distro
-from tests.tools import AgentTestCase, open_patch, patch
+    get_f5_platform, get_distro, PY_VERSION_MAJOR, PY_VERSION_MINOR
+from tests.tools import AgentTestCase, open_patch, patch, skip_if_predicate_false
 
 
 def freebsd_system():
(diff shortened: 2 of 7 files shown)


tests/common/test_version.py
```diff
@@ -57,6 +57,13 @@ def default_system_exception():
     raise Exception
 
 
+def is_platform_supported():
```
larohra, line 60: NIT: rename to `is_platform_dist_supported()`. `is_platform_supported` sounded more like if the platform supports it

test-requirements.txt
```diff
@@ -4,5 +4,6 @@ flake8; python_version >= '2.7'
 mock==2.0.0; python_version == '2.6'
 mock==3.0.5; python_version >= '2.7' and python_version <= '3.5'
 mock==4.0.2; python_version >= '3.6'
+distro; python_version >= '3.7'
```
larohra, line 7: >distro is supported and tested on Python 2.7, 3.4+ 

Got this from their page. We should modify the line to download distro for python_version >= 2.7

tests/common/test_version.py
```diff
@@ -78,6 +85,7 @@ def test_distro_is_correct_format_when_openbsd(self, platform_system_name, mock_
         self.assertListEqual(openbsd_list, osinfo)
         return
 
+    @skip_if_predicate_false(is_platform_supported, "platform.dist() is deprecated in Python 3.7+")
```
larohra, line 88: Instead of skipping these tests shouldn't we fix them to test using the distro module instead? (Since we dont have the Canonical deadline anymore might as well do it right)

larohra:
>Removed coverage output in Travis runs since the output was getting in the way of readability for unit test runs (codecov still runs on PRs and commits)

Thanks for removing coverage, it was definitely annoying to see the red cross even when Travis passed! That being said, I dont see the codecov results for in the PR anywhere. Is there a specific place I need to look for it?

test-requirements.txt
narrieta: i'm thinking we should match this to the images tested in DCR to make UTs closer to actual VM environments. We know Ubuntu 20 will be 3.8 and includes 'distro' by default... other distros?

tests/common/test_version.py
narrieta: or should we have versions for platform and distro?

pgombar: Updating them to cover both scenarios: when platform.dist is supported and when it isn't.

test-requirements.txt
pgombar: Since our `requirements.txt` file already specify `distro` is a dependency, this requirements in `test-requirements.txt` is in sync.

Note that I've also updated the requirement in `requirements.txt` to start with py3.7 (as opposed to 3.8) since `platform.dist` is deprecated starting 3.7.

pgombar: Ignore this, it's deprecated starting 3.8, not 3.7, reverting that back.

pgombar:
> > Removed coverage output in Travis runs since the output was getting in the way of readability for unit test runs (codecov still runs on PRs and commits)
> 
> Thanks for removing coverage, it was definitely annoying to see the red cross even when Travis passed! That being said, I dont see the codecov results for in the PR anywhere. Is there a specific place I need to look for it?

No, you're definitely right, by removing it from the yaml file, it removes the hook to run on each commit. I will revert the change then, I don't think there's a way where we can remove the output from individual Travis runs but still keep a hook on GitHub.

PR #1861, opened 2020-04-21
Prepend python path to extensions 

larohra:
## Description
Adding the python interpreter path of the agent to extensions.

Issue # 

---

### PR information


azurelinuxagent/ga/exthandlers.py
```diff
@@ -1215,6 +1217,11 @@ def launch_command(self, cmd, timeout=300, extension_error_code=ExtensionErrorCo
                             ExtCommandEnvVariable.ExtensionVersion: str(self.ext_handler.properties.version),
                             ExtCommandEnvVariable.ExtensionSeqNumber: str(self.get_seq_no())})
 
+                python_symlink_path = get_python_symlink_path_if_ubuntu_20_04_image()
+                if python_symlink_path is not None:
```
pgombar, line 1221: Should we be checking that it's a symlink here like you are in `update.py` before appending it to the PATH?

azurelinuxagent/common/osutil/factory.py
```diff
@@ -38,6 +38,13 @@
 
 from distutils.version import LooseVersion as Version
 
+UBUNTU_20_04_IMAGE_PATH = "/etc/os-release/ub20-image"
+
+def get_python_symlink_path_if_ubuntu_20_04_image():
```
narrieta, line 43: should be in osutils, rather than in the factory

azurelinuxagent/ga/update.py
```diff
@@ -728,6 +730,36 @@ def _write_pid_file(self):
...
+            return
+
+        add_event(
+            AGENT_NAME,
+            op=WALAEventOperation.Ubuntu2004_Image,
+            version=CURRENT_VERSION,
+            is_success=True,
+            message="Agent running on Ubuntu20.04, setting a python symlink in agent directory")
+
+        if os.path.islink(agent_dir_python_symlink_path):
```
narrieta, line 748: path.exists instead of path.islink?

azurelinuxagent/ga/update.py
```diff
@@ -728,6 +730,36 @@ def _write_pid_file(self):
...
+            logger.info("Symlink to python interpreter already exists in agent dir {0}".format(agent_dir_python_symlink_path))
+            return
+
+        try:
+            # Create a symlink in agent directory which points to the Python interpreter path that the agent is running on
+            os.symlink(sys.executable, agent_dir_python_symlink_path)
+        except OSError as e:
```
narrieta, line 755: does symlink *always* throw an OSError?

azurelinuxagent/ga/update.py
```diff
@@ -728,6 +730,36 @@ def _write_pid_file(self):
...
+            version=CURRENT_VERSION,
+            is_success=True,
+            message="Agent running on Ubuntu20.04, setting a python symlink in agent directory")
+
+        if os.path.islink(agent_dir_python_symlink_path):
+            logger.info("Symlink to python interpreter already exists in agent dir {0}".format(agent_dir_python_symlink_path))
+            return
+
+        try:
+            # Create a symlink in agent directory which points to the Python interpreter path that the agent is running on
```
narrieta, line 753: let's also add a log message (add_event(is_success=True) doesn't log)

tests/ga/test_extension.py
```diff
@@ -1942,6 +1942,58 @@ def test_correct_exit_code_should_be_set_on_uninstall_cmd_failure(self, *args):
...
+
+                        for _, kwargs in mock_report_event.call_args_list:
+                            # The output is of the format - 'testfile.sh\n[stdout]PATH=N\n[stderr]'
+                            if test_file_name not in kwargs['message']:
+                                continue
+
+                            python_symlink_path = get_python_symlink_path_if_ubuntu_20_04_image()
+                            if python_symlink_path is not None:
+                                self.assertEqual(file, file_that_exists_in_fs, "File should exist in Filesystem if ubuntu image exists")
+                                self.assertIn("\nPATH={0}".format(conf.get_lib_dir()), kwargs['message'], "Agent lib directory should be prepended to path")
```
narrieta, line 1989: i think we should assert that the full path of the link is in path, rather than just the agent's dir

azurelinuxagent/ga/update.py
larohra: Initially I read that path.islink checks if its a symlink but wont return False if its broken but path.exists would even return false if the link is broken which is awesome! I'll replace it with path.exists

azurelinuxagent/ga/update.py
larohra: I couldn't find the exact list in the official documentation but read in a blog it reports FilenotFoundError, or FileExistsError as OSError. I can do a blanket catch too just for piece of mind I guess

azurelinuxagent/ga/exthandlers.py
larohra: I'm checking there only to verify if it exists or not. I dont want to add that check here because I dont want to manipulate with the filesystem in this method (mainly coz its called so frequently). This check is mainly to verify if its a ub20 image or not because the function `get_python_symlink_path_if_ubuntu_20_04_image` would return not None only if its a ub20 image.

azurelinuxagent/ga/update.py
larohra: done

tests/ga/test_extension.py
larohra: We dont need to provide the whole link path when we prepend to the path before invoking the command. Rather, we just need to provide the directory of where its stored and the shebang walks the PATH and finds it accordingly. We're writing the symlink in the agent dir so we're just prepending the agent dir to the PATH variable rather than the whole link. 

Eg:

```sh
larohra@larohra-ub16dev:~/test_import/modsDir$ file python
python: symbolic link to /usr/bin/python3

# With just directory prepended to path
larohra@larohra-ub16dev:~/test_import/modsDir$ PATH=`pwd`:$PATH ./mods.py
sys.version_info(major=3, minor=5, micro=2, releaselevel='final', serial=0)
Hello test

# With Dir/python prepended to path
larohra@larohra-ub16dev:~/test_import/modsDir$ PATH=`pwd`/python:$PATH ./mods.py
sys.version_info(major=2, minor=7, micro=12, releaselevel='final', serial=0)
Hello test

```

azurelinuxagent/common/osutil/factory.py
larohra: Moved to osutil/ubuntu.py, looked like the perfect place for it to be

larohra:
Closing in favour or - https://github.com/Azure/WALinuxAgent/pull/1863

azurelinuxagent/common/event.py
```diff
@@ -110,6 +110,7 @@ class WALAEventOperation:
     Unknown = "Unknown"
     Upgrade = "Upgrade"
     Update = "Update"
+    Ubuntu2004_Image = "Ubuntu2004_Image"
```
kikoreis, line 113: Shouldn't the event name make it clear this is a mitigation for a missing /usr/bin/python?

azurelinuxagent/common/osutil/ubuntu.py
```diff
@@ -15,15 +15,23 @@
...
 import time
 
 import azurelinuxagent.common.logger as logger
 import azurelinuxagent.common.utils.shellutil as shellutil
+from azurelinuxagent.common import conf
 
 from azurelinuxagent.common.osutil.default import DefaultOSUtil
 
 
+UBUNTU_20_04_IMAGE_PATH = "/etc/ub20-image"
```
kikoreis, line 28: There may be a better way of us checking whether the mitigation needs to be applied. I guess there's the obvious question -- why don't you just check to see whether /usr/bin/python exists and if not, put a python to python3 symlink in anyway? It's not bound to make anything much worse -- if the extension has been fixed it will no longer assume /usr/bin/python, and if it does find /usr/bin/python then it can go ahead assuming it's python2.

kikoreis:
Apart from my two comments above, I also have a general question here: how useful is this fix? I assume it is important because there are extensions that assume /usr/bin/python exists, but in our investigation the extensions we found using /usr/bin/python also had Python2 assumptions built in, so aren't we simply delaying the point at which it crashes?

azurelinuxagent/common/osutil/ubuntu.py
kikoreis: If you do want to rely some on-disk metadata, you may be able to use the serial from /etc/cloud/build.info, which for a Bionic VM booted ~90 days ago looks like:

build_name: server
serial: 20191218

A simple numeric comparison of serial will allow you to decide at which point you stop applying this mitigation. I'm not sure this approach is valid but wanted to record it here.

larohra: >why don't you just check to see whether /usr/bin/python exists and if not, put a python to python3 symlink in anyway? 

We did discuss this idea too but dropped it because that would impact all extensions (on all distros) but for the purpose of this PoC we decided to just restrict it to Ub20 instead.

> you may be able to use the serial from /etc/cloud/build.info

Ahh interesting, did not know about this! Thanks for informing us about this. Its always good to know more about the internals of a distro.

larohra:
@kikoreis you have a very valid point. The whole purpose of this PR was to create a PoC with the symlink and see if the extensions (if any) were compatible with Python 3 once their dependency on `/usr/bin/env python` was fulfilled. Unfortunately though, as you mentioned, looks like all the extensions that did depend on that shebang are not compatible with Py3 as of now. So based on that we're dropping the PR ( #1863 ) and not going ahead with the fix.

Thanks for taking a look at this though!

PR #1873, opened 2020-04-28
Fix wrongful dir deletion

larohra:
## Description
This PR focuses on the error where if we have a bad extension zip existing on the machine, then we delete the whole directory including the state which causes problems later. Now we ensure that the HandlerState for extension is always maintained in the extension directory 

Issue # 

---

### PR information


larohra:
Successful DCR run (except for the known issues) - https://tuxgold.corp.microsoft.com/job/AzLinux/job/DungeonCrawler/job/DungeonCrawler.larohra2/460/artifact/results/

azurelinuxagent/ga/exthandlers.py
```diff
@@ -842,14 +843,30 @@ def _download_extension_package(self, source_uri, target_file):
...
+            self.report_event(message=err_msg, is_success=False, log_event=False)
             os.remove(source_file)
-            if os.path.exists(target_directory):
-                shutil.rmtree(target_directory)
+            if os.path.exists(temp_dir):
+                shutil.rmtree(temp_dir)
             return False
+
+        try:
+            copy_tree(temp_dir, target_directory)
```
pgombar, line 860: We never clean up the temp dir when extraction succeeds. From the [docs](https://docs.python.org/2/library/tempfile.html):

> The user of mkdtemp() is responsible for deleting the temporary directory and its contents when done with it.

larohra: That's a good point! Thanks for catching it! I think I just thought that since its a temp directory the OS would eventually clean it up lol but I'll do it just to be sure.

larohra: Done

azurelinuxagent/ga/exthandlers.py
```diff
@@ -842,14 +843,33 @@ def _download_extension_package(self, source_uri, target_file):
...
+        # Blanket exception catch to capture any errors during copying the extension files
+        except Exception as e:
+            # Incase there is an error while copying, we want to clean the directory to avoid inconsistent state
+            fileutil.clean_ioerror(e, paths=[target_directory])
```
narrieta, line 864: but by removing the target_directory we are back to the original issue, right?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -925,6 +925,9 @@ def initialize(self):
             status_dir = self.get_status_dir()
             fileutil.mkdir(status_dir, mode=0o700)
 
+            conf_dir = self.get_conf_dir()
```
larohra, line 928: The main issue was that incase of a bad zip, the conf directory was being deleted and `get_status_file_path` function was throwing because it has a dependency on it. So moved it one step up to ensure that the config directory is always present.

larohra:
This PR needs 2 follow ups - 
1- Always create a new placeholder status file for extensions for if doesnt exist - [Bug#7273136](https://msazure.visualstudio.com/One/_workitems/edit/7273136)
2 - Separate the HandlerState and HandlerStatus file outside Extension Directory to manage the lifecycle outside extension events - [Bug#7273698](https://msazure.visualstudio.com/One/_workitems/edit/7273698)

Both these would be done later as separate PRs

PR #1874, opened 2020-04-29
Removing infinite download of extension manifest without a new GS

larohra:
## Description
This PR contains the fix where we dont download the extension manifest without a new GoalSatate

Issue # 

---

### PR information


larohra:
DCR ok (except for the known errors) - https://tuxgold.corp.microsoft.com/job/AzLinux/job/DungeonCrawler/job/DungeonCrawler.larohra2/461/artifact/results/

azurelinuxagent/ga/exthandlers.py
```diff
@@ -423,6 +423,14 @@ def handle_ext_handler(self, ext_handler, etag):
 
         try:
             state = ext_handler.properties.state
+
+            self.get_artifact_error_state.reset()
```
narrieta, line 427: the error state should be reset after the "if decide_version" (because download succeeded)

now, since with this change we short-circuit the logic if there is not a new goal state, it seems to me that the error state is not needed and we should always report errors (since it is a new goal state) -- could you review the code to check if this is true?

larohra: good point, I didnt really check this part. Let me take a look and verify the need for it

azurelinuxagent/ga/exthandlers.py
```diff
@@ -474,12 +472,8 @@ def handle_ext_handler_download_error(self, ext_handler_i, e, code=-1):
         msg = ustr(e)
         ext_handler_i.set_handler_status(message=msg, code=code)
 
-        self.get_artifact_error_state.incr()
```
larohra, line 477: All the operations that raise `ExtensionDownloadError` only happen on a per GoalState basis. It doesn't make sense to not report them everytime if it causes the GoalState to fail. Changing the logging to report always rather than every 5 mins

narrieta: Seems like get_artifact_error_state can be removed altogether. The only usage I see is in

```
            self.get_artifact_error_state.reset()
```

and get_ext_handler doesn't download anything

larohra: Verified that get_artifact_error_state is no longer needed. We already log the error as part of `try_update_gs` function. Removed

azurelinuxagent/ga/exthandlers.py
larohra: Done, removed the error state completely

tests/ga/test_exthandlers.py
```diff
@@ -249,29 +249,6 @@ def test_extension_sequence_number(self):
                                               disk_sequence_number=3,
                                               expected_sequence_number=-1)
 
-    @patch("azurelinuxagent.ga.exthandlers.add_event")
-    @patch("azurelinuxagent.common.errorstate.ErrorState.is_triggered")
-    def test_it_should_report_an_error_if_the_wireserver_cannot_be_reached(self, patch_is_triggered, patch_add_event):
```
larohra, line 254: This is a redundant test, this scenario is being tested by the tests for `try_update_goal_state` functionality

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1388,6 +1361,8 @@ def set_handler_status(self, status="NotReady", message="", code=0):
         try:
             handler_status_json = json.dumps(get_properties(handler_status))
             if handler_status_json is not None:
+                if not os.path.exists(state_dir):
```
larohra, line 1364: Currently if there is any exceptions in extension processing and the config directory doesnt exist, the error is never reported back to CRP and the customer has to wait for 90 mins before the operation timesout. This is a bad Customer Experience. To mitigate it, I'm creating the config directory incase it doesnt exist while reporting status.

tests/ga/test_extension.py
```diff
@@ -920,39 +950,37 @@ def test_ext_handler_download_failure_permanent_ProtocolError(self, mock_add_eve
...
 
-        mock_error_state.return_value = False
+        exthandlers_handler.run()
+        _assert_mock_add_event_call(expected_call_count=1)
+        self._assert_handler_status(protocol.report_vm_status, "NotReady", 0, "1.0.0")
 
+        # Re-run exthandler.run without updating the GS and ensure we dont report error
         exthandlers_handler.run()
+        _assert_mock_add_event_call(expected_call_count=1)
+        self._assert_handler_status(protocol.report_vm_status, "NotReady", 0, "1.0.0")
```
larohra, line 975: This line will fail this test currently because the cleanup logic cleans up the state and we dont report it again. Once the PR #1889  is merged, this test should pass.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -236,32 +236,11 @@ def __init__(self, protocol):
...
-        except Exception as e:
-            msg = u"Exception retrieving extension handlers: {0}".format(ustr(e))
-            detailed_msg = '{0} {1}'.format(msg, traceback.extract_tb(get_traceback(e)))
-
-            self.get_artifact_error_state.incr()
-
-            if self.get_artifact_error_state.is_triggered():
-                add_event(AGENT_NAME,
-                          version=CURRENT_VERSION,
-                          op=WALAEventOperation.GetArtifactExtended,
```
narrieta, line 255: i think GetArtifactExtended is not used anymore and can be deleted

azurelinuxagent/ga/exthandlers.py
```diff
@@ -423,6 +402,13 @@ def handle_ext_handler(self, ext_handler, etag):
 
         try:
             state = ext_handler.properties.state
+
```
narrieta, line 405: state = ext_handler.properties.state should come after this if statement

tests/ga/test_extension.py
```diff
@@ -920,39 +950,37 @@ def test_ext_handler_download_failure_permanent_ProtocolError(self, mock_add_eve
...
 
-        self.assertEquals(0, mock_add_event.call_count)
+        # Change incarnation and then re-check we report error
+        test_data.set_incarnation(2)
+        protocol.update_goal_state()
+
+        exthandlers_handler.run()
+        _assert_mock_add_event_call(expected_call_count=2)
```
narrieta, line 982: checking for a fixed number of calls to add_event is fragile. If we happen to add an unrelated event on this path the test will break

azurelinuxagent/ga/exthandlers.py
larohra: Good catch! removed it

azurelinuxagent/ga/exthandlers.py
larohra: Done

PR #1879, opened 2020-04-30
Fix timestamp for periodic operations in the monitor thread

narrieta:
#1770 reset the timestamp of the tasks in the monitor thread on each iteration.

As a result, tasks with periods greater than the minimum period (e.g. perf counters, reset periodic messages) will be executed only once, at agent startup.

### DIFF
--- a/azurelinuxagent/ga/monitor.py
+++ b/azurelinuxagent/ga/monitor.py
@@ -33,6 +33,7 @@ from azurelinuxagent.common.protocol.imds import get_imds_client
 from azurelinuxagent.common.utils.restutil import IOErrorCounter
 from azurelinuxagent.common.utils.textutil import hash_strings
 from azurelinuxagent.common.version import AGENT_NAME, CURRENT_VERSION
+from azurelinuxagent.ga.periodic_operation import PeriodicOperation
 
 
 def generate_extension_metrics_telemetry_dictionary(schema_version=1.0,
@@ -51,9 +52,11 @@ def get_monitor_handler():
 
 
 class MonitorHandler(object):
+    # telemetry
     EVENT_COLLECTION_PERIOD = datetime.timedelta(minutes=1)
+    # host health
     TELEMETRY_HEARTBEAT_PERIOD = datetime.timedelta(minutes=30)
-    # extension metrics period
+    # cgroup data period
     CGROUP_TELEMETRY_POLLING_PERIOD = datetime.timedelta(minutes=5)
     CGROUP_TELEMETRY_REPORTING_PERIOD = datetime.timedelta(minutes=30)
     # host plugin
@@ -62,7 +65,8 @@ class MonitorHandler(object):
     # imds
     IMDS_HEARTBEAT_PERIOD = datetime.timedelta(minutes=1)
     IMDS_HEALTH_PERIOD = datetime.timedelta(minutes=3)
-
+    # log network configuration
+    LOG_NETWORK_CONFIGURATION_PERIOD = datetime.timedelta(minutes=1)
     # Resetting loggers period
     RESET_LOGGERS_PERIOD = datetime.timedelta(hours=12)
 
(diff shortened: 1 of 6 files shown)


azurelinuxagent/ga/monitor.py
```diff
@@ -122,22 +133,10 @@ def collect_and_send_events(self):
...
-            if self.last_event_collection is None:
-                self.last_event_collection = datetime.datetime.utcnow() - MonitorHandler.EVENT_COLLECTION_PERIOD
+        event_list = collect_events()
 
-            if datetime.datetime.utcnow() >= (self.last_event_collection + MonitorHandler.EVENT_COLLECTION_PERIOD):
-                try:
-                    event_list = collect_events()
-
-                    if len(event_list.events) > 0:
-                        self.protocol.report_event(event_list)
```
larohra, line 134: Initially we were reporting errors everytime it happened if unable to send telemetry (i.e. every minute). Now we will only log once per hour as per PeriodicOperation. Should we revisit that?
(Also added another concern with the way PeriodicOperation logs in the PeriodicOperation file)

azurelinuxagent/ga/periodic_operation.py
```diff
@@ -0,0 +1,51 @@
...
+    def run(self):
+        try:
+            if self._last_run is None or datetime.datetime.utcnow() >= self._last_run + self._period:
+                try:
+                    self._operation()
+                finally:
+                    self._last_run = datetime.datetime.utcnow()
+        except Exception as e:
+            if self._last_log_warning is None or datetime.datetime.utcnow() >= self._last_log_warning + self._LOG_WARNING_PERIOD:
+                logger.warn("[PERIODIC] Failed to {0}: {1}", self._name, ustr(e))
```
larohra, line 49: I am a bit concerned about how the Periodic logger is set up here. If the _operation function fails for multiple reasons within the same hour and it does not have its own try-except, we might miss out on the errors as this implementation logs errors based on the operation rather than the type of the error (as we do it in the PeriodicLogger). 

I feel instead of setting a periodic logger, we should just log at the time frequency that the operation is called in so that we dont miss out on key errors. Or maybe have a cache per error maybe?

azurelinuxagent/ga/monitor.py
```diff
@@ -266,33 +242,22 @@ def send_host_plugin_heartbeat(self):
...
+                                                                 other_errors)
+            add_event(
+                name=AGENT_NAME,
+                version=CURRENT_VERSION,
+                op=WALAEventOperation.HttpErrors,
+                is_success=True,
+                message=msg,
+                log_event=False)
 
     def poll_telemetry_metrics(self):
```
larohra, line 262: NIT: Forgot to reformat this function :)

azurelinuxagent/ga/periodic_operation.py
```diff
@@ -0,0 +1,51 @@
...
+
+import datetime
+
+from azurelinuxagent.common import logger
+from azurelinuxagent.common.future import ustr
+
+
+class PeriodicOperation(object):
+    '''
+    Instances of PeriodicOperation are tasks that are executed only after the given
```
larohra, line 26: Thanks for adding this class! Will definitely make it easier to setup a periodic operation now! 

NIT: Consider adding a small note here saying that the `run` will never throw so make sure to add your custom try-except if needed

tests/ga/test_monitor.py
```diff
@@ -24,49 +25,67 @@
...
+        if len(enabled_operations) == 0 or self._name in enabled_operations:
+            run.original_definition(self)
+    run.original_definition = PeriodicOperation.run
+
+    with mock_wire_protocol(DATA_FILE) as protocol:
+        protocol_util = MagicMock()
+        protocol_util.get_protocol = Mock(return_value=protocol)
+        with patch("azurelinuxagent.ga.monitor.get_protocol_util", return_value=protocol_util):
+            with patch.object(PeriodicOperation, "run", side_effect=run, autospec=True):
+                with patch("azurelinuxagent.ga.monitor.MonitorHandler.stopped", side_effect=[False] * iterations + [True]):
```
larohra, line 76: `side_effect=[False] * iterations + [True]` this is awesome, so Pythonic!

tests/ga/test_monitor.py
```diff
@@ -24,49 +25,67 @@
...
-        self.status = status
-        self.reason = reason
-        self.response = response
+def random_generator(size=6, chars=string.ascii_uppercase + string.digits + string.ascii_lowercase):
+    return ''.join(random.choice(chars) for x in range(size))
 
-    def read(self):
-        return self.response
+@contextlib.contextmanager
+def _create_monitor_handler(enabled_operations=[], iterations=1):
```
larohra, line 53: Thanks for the major rewrite!

azurelinuxagent/ga/monitor.py
narrieta: (I'll address your 2 comments here)

While refactoring the code I realized we were always logging on error. Given the frequency of some of those operations this has the potential of producing very large logs. I agree that we may be hiding errors if they are different but our current periodic logging has issues of its own.

To be honest, I've never seen those operations fail, so as far as this PR for release 2.2.48 I see 2 options: revert to the previous behavior of always logging or use this simple approach (which could be improved in subsequent PRs if one of us feels it is needed). I am inclined towards the latter, but can revert if you prefer so.

azurelinuxagent/ga/periodic_operation.py
narrieta: (addresed in the reply to your first comment)

azurelinuxagent/ga/monitor.py
narrieta: Thanks; done.

azurelinuxagent/ga/periodic_operation.py
narrieta: Done

azurelinuxagent/ga/monitor.py
```diff
@@ -51,9 +52,11 @@ def get_monitor_handler():
 
 
 class MonitorHandler(object):
+    # telemetry
     EVENT_COLLECTION_PERIOD = datetime.timedelta(minutes=1)
+    # host health
     TELEMETRY_HEARTBEAT_PERIOD = datetime.timedelta(minutes=30)
```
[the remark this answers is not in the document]
narrieta: Agree. This code still needs a lot of refactoring. I am trying to minimize changes in the agent for this release, though.

PR #1889, opened 2020-05-18
Fix the cleanup-outdated-handlers to only delete handlers that are not present in the GS

larohra:
## Description

This PR focuses on fixing the cleanup-outdated-handlers functions to only delete handlers that are not present in the GS rather than cleaning up all the Extension Handlers with State==NotInstalled.

This PR addresses 1st part of the issue where the LinuxAgent does not report install failures properly back to CRP which leads to a bad customer experience.

Issue # 

---

### PR information


tests/data/wire/ext_conf_multiple_extensions.xml
```diff
@@ -16,31 +16,11 @@
...
-        <Plugin name="Microsoft.OSTCExtensions.Edp.ExampleExtensionLinuxInTest" version="1.0.0"
-                location="https://mock-goal-state/zrdfepirv2sn6prdstr05a.blob.core.windows.net/b01058962be54ceca550a390fa5ff064/Microsoft.OSTCExtensions.Edp_ExampleExtensionLinuxInTest_ussouth_manifest.xml"
-                state="enabled" autoUpgrade="true"
-                failoverlocation="https://mock-goal-state/rdfepirv2sn3prdstr01.blob.core.windows.net/b01058962be54ceca550a390fa5ff064/Microsoft.OSTCExtensions.Edp_ExampleExtensionLinuxInTest_ussouth_manifest.xml"
-                runAsStartupTask="false" isJson="true" useExactVersion="false"/>
+        <Plugin name="OSTCExtensions.ExampleHandlerLinux" version="1.0.0" location="http://mock-goal-state/rdfepirv2hknprdstr03.blob.core.windows.net/b01058962be54ceca550a390fa5ff064/Microsoft.OSTCExtensions_ExampleHandlerLinux_asiaeast_manifest.xml" config="" state="enabled" autoUpgrade="false" failoverlocation="http://mock-goal-state/rdfepirv2hknprdstr04.blob.core.windows.net/b01058962be54ceca550a390fa5ff064/Microsoft.OSTCExtensions_ExampleHandlerLinux_asiaeast_manifest.xml" runAsStartupTask="false" isJson="true"/>
```
larohra, line 19: Had to reformat this file as it was breaking the regex for the `mockwiredata.set_extensions_config_state` function.

tests/ga/test_extension.py
```diff
@@ -79,103 +81,129 @@ def raise_ioerror(*args):
     e.errno = EIO
     raise e
 
-
 class TestExtensionCleanup(AgentTestCase):
+
```
larohra, line 85: Reformatted the `TestExtensionCleanup` class to test the public interfaces rather than testing the internal private functions.

tests/ga/test_extension.py
```diff
@@ -79,103 +81,129 @@ def raise_ioerror(*args):
...
+    @staticmethod
+    def _is_extension_dir(path):
+        return re.match(HANDLER_NAME_PATTERN, os.path.basename(path))
 
-        self._install_handlers(start=0, count=5, handler_state=ExtHandlerState.Installed)
-        self._install_handlers(start=5, count=5, handler_state=ExtHandlerState.Enabled)
 
-        self.assertEqual(self._count_packages(), 10)
-        self.assertEqual(self._count_installed(), 10)
+    def _assert_ext_handler_status(self, report_vm_status, expected_status, version, expected_ext_hanlder_count=0):
```
pgombar, line 110: typo in `hanlder`

tests/data/wire/ext_conf_multiple_extensions.xml
pgombar: It was breaking an existing regex or it started breaking due to your changes? Can you expand on the changes in this file?

tests/ga/test_extension.py
```diff
@@ -79,103 +81,129 @@ def raise_ioerror(*args):
...
+        self.assertTrue(report_vm_status.called)
+        args, kw = report_vm_status.call_args
+        vm_status = args[0]
+        self.assertEqual(expected_ext_hanlder_count, len(vm_status.vmAgent.extensionHandlers))
+        for ext_handler in vm_status.vmAgent.extensionHandlers:
 
-        self.ext_handlers._cleanup_outdated_handlers()
+            self.assertEquals(expected_status, ext_handler.status)
+            self.assertEquals(version, ext_handler.version)
+            # self.assertEquals(expected_handler_name, handler_status.name)
```
pgombar, line 119: Remove?

pgombar: Rather, uncomment?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -302,19 +302,19 @@ def _cleanup_outdated_handlers(self):
                 if re.match(HANDLER_NAME_PATTERN, item) is None:
                     continue
                 try:
-                    eh = ExtHandler()
-
                     separator = item.rfind('-')
+                    handler_name = item[0:separator]
+                    if any([handler_name == ext_handler.name for ext_handler in self.ext_handlers.extHandlers]):
```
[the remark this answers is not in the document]
larohra: Good point, I thought of doing this initially but realised that the ext_handlers wont be that many but then again saving any computation is good I guess.

tests/data/wire/ext_conf_multiple_extensions.xml
larohra: It was breaking an existing regex. I think we never tested the regex on this file before so that's why we didnt know. I mainly just reformatted it and removed the extra spaces and new line chars and made it 1 coherent line (per Plugin line).

tests/ga/test_extension.py
larohra: Thanks for catching this, you should really charge for linting code like this :p

tests/ga/test_extension.py
larohra: Removed it, we dont really need it. Mainly just checking if the cleanup was successful or not.

tests/ga/test_extension.py
```diff
@@ -79,103 +81,128 @@ def raise_ioerror(*args):
...
-
-            if not handler_state is None:
-                zipfile.ZipFile(dst).extractall(handler.get_base_dir())
-                handler.set_handler_state(handler_state)
+        self.mock_sleep = patch("time.sleep", lambda *_: mock_sleep(0.0001))
+        self.mock_sleep.start()
 
-            version += 1
+    def tearDown(self):
+        AgentTestCase.tearDown(self)
```
narrieta, line 92: the order of cleanup should be the reverse of setup (i.e. stop the mock first, then invoke the base class)

tests/ga/test_extension.py
```diff
@@ -79,103 +81,128 @@ def raise_ioerror(*args):
...
-    def _is_installed(self, path):
-        path = os.path.join(path, 'config', 'HandlerState')
-        return fileutil.read_file(path) != "NotInstalled"
-
-    @patch("azurelinuxagent.common.conf.get_lib_dir")
-    def test_cleanup_leaves_installed_extensions(self, mock_conf):
-        mock_conf.return_value = self.lib_dir
+    @staticmethod
+    def _is_extension_dir(path):
+        return re.match(HANDLER_NAME_PATTERN, os.path.basename(path))
```
narrieta, line 107: return re.match(HANDLER_NAME_PATTERN, os.path.basename(path)) != None

(the method name implies a bool return value)

tests/ga/test_extension.py
```diff
@@ -79,103 +81,128 @@ def raise_ioerror(*args):
...
+            self.assertEquals(version, ext_handler.version)
+        return
 
-        self.assertEqual(self._count_packages(), 10)
-        self.assertEqual(self._count_installed(), 10)
-        self.assertEqual(self._count_uninstalled(), 0)
+    @contextlib.contextmanager
+    def _setup_test_env(self, test_data):
+        with mock_wire_protocol(test_data) as protocol:
+            protocol.report_vm_status = MagicMock()
```
narrieta, line 124: instead of mocking report_vm_status, can we mock the put request instead?

tests/ga/test_extension.py
larohra: Didnt, know that, thanks for letting me know! Done

tests/ga/test_extension.py
larohra: good catch, I was inherently using the None comparison checks as bool. Done

tests/ga/test_extension.py
larohra: Good point, let me make that change.

tests/ga/test_extension.py
narrieta: to clarify - it is just a best practice; in this case it does not make a difference

tests/ga/test_extension.py

PR #1891, opened 2020-05-21
Expose periods of environment thread in waagent.conf

narrieta:
Added parameters in waagent.conf to control how often operations in the environment thread are performed.


README.md
```diff
@@ -236,6 +238,28 @@ without the agent. In order to do that, the `provisionVMAgent` flag must be set
 provisioning time, via whichever API is being used. We will provide more details on
 this on our wiki when it is generally available. 
 
+#### __Extensions.GoalStatePeriod__
```
larohra, line 241: I'm not sure if we should expose this in the conf mainly because small things like the VM status is dependent on this.

azurelinuxagent/common/conf.py
```diff
@@ -164,6 +171,18 @@ def enable_firewall(conf=__conf__):
     return conf.get_switch("OS.EnableFirewall", False)
 
 
+def get_enable_firewall_period(conf=__conf__):
+    return conf.get_int("OS.EnableFirewallPeriod", 30)
```
larohra, line 175: A lot of the new flags dont have an entry in the README, any particular reason for that?

azurelinuxagent/ga/env.py
```diff
@@ -92,56 +101,50 @@ def start(self):
...
+            # than initializing it in the ExtHandler thread. This is done to avoid any concurrency issues as each
+            # thread would now have its own ProtocolUtil object as per the SingletonPerThread model.
+            self.protocol_util = get_protocol_util()
+            self._protocol = self.protocol_util.get_protocol()
+            while not self.stopped:
+                try:
+                    for op in self._periodic_operations:
+                        op.run()
+                except Exception as e:
+                    logger.error("An error occurred in the environment thread main loop; will skip the current iteration.\n{0}", ustr(e))
```
larohra, line 115: We might never hit this path as no operation has their own error handling so the `PeriodicOperation.run()` would just log it and let it go. 

I wonder if it would be cleaner to just keep only the uber except @Line:119 to ensure that we log incase the thread goes down

azurelinuxagent/ga/monitor.py
```diff
@@ -137,29 +139,24 @@ def collect_and_send_events(self):
...
-                self._log_altered_network_configuration_op.run()
-                self._reset_loggers_op.run()
-            except Exception as e:
-                logger.warn("An error occurred in the monitor thread main loop; will skip the current iteration.\n{0}", ustr(e))
-            time.sleep(min_delta)
+                    for op in self._periodic_operations:
+                        op.run()
+
+                except Exception as e:
+                    logger.error("An error occurred in the monitor thread main loop; will skip the current iteration.\n{0}", ustr(e))
```
larohra, line 155: Same for this, all functions have their own error handling or the PeriodicFunction.run will take care of it, this code path will never hit

README.md
```diff
@@ -236,6 +238,28 @@ without the agent. In order to do that, the `provisionVMAgent` flag must be set
...
+
+_Type: Integer_  
+_Default: 6_
```
larohra, line 244: Also just curious, why did you change the default from 3 to 6secs? I think Joe had a dependency on checking the artifact blob every 3 secs for Fasttrack.

README.md
narrieta: There is a legitimate need to change the default. I added a note in the documentation. Also, changes are logged locally and to telemetry.

README.md
narrieta: We need to slow down status reporting. until it is moved to a different thread i'm slowing it down here. can you post a comment in joe's PR to not hardcode 3 sec

azurelinuxagent/common/conf.py
narrieta: This should be the only one that is not in README. EnableFirewall is not there either so I am following the same

azurelinuxagent/ga/update.py
```diff
@@ -421,6 +422,28 @@ def _emit_restart_event(self):
 
         return
 
+    @staticmethod
+    def _emit_changes_in_default_configuration():
```
larohra, line 426: Thanks for adding this, much needed! 
I have a suggestion, why not just log all parameters that changed from default, that way we will always have an idea of what the customers are using and maybe later on we can leverage that to change our defaults if needed.

azurelinuxagent/ga/env.py
narrieta: a typo (for example) could produce an exception. since this is the main loop i am protecting it

azurelinuxagent/ga/monitor.py
narrieta: same as above

azurelinuxagent/common/conf.py
larohra: So there is a `yml` list and then there is a description for each flag in the conf file. I found a couple that have a description in the README but have not been added to the `yml` list in the README - 
```
OS.RemovePersistentNetRulesPeriod
OS.RootDeviceScsiTimeoutPeriod
Provisioning.MonitorHostNamePeriod
```   

Missing completely from README - 
```
OS.MonitorDhcpClientRestartPeriod
```

Also I personally feel we should add the Firewall rules to the readme too, it makes more sense to be as transparent as possible especially because its an open source project.

README.md
larohra: Sure, I'll add a comment once I do that PR

azurelinuxagent/common/conf.py
narrieta: I'll add MonitorDhcpClientRestartPeriod to README. The other list is not in sync with README, I think it is just meant as a sample, not a comprehensive list of options. I'm not planning on listing all the missing options there.

azurelinuxagent/ga/periodic_operation.py
```diff
@@ -35,22 +36,40 @@ class PeriodicOperation(object):
...
+        and sleeps until it is time to execute that operation.
+        """
+        next_operation_time = min([op.next_run_time() for op in operations])
+
+        sleep_timedelta = next_operation_time - datetime.datetime.utcnow()
+        # timedelta.total_seconds() is not available on Python 2.6, do the computation manually
+        sleep_seconds = ((sleep_timedelta.days * 24 * 3600 + sleep_timedelta.seconds) * 10.0 ** 6 + sleep_timedelta.microseconds) / 10.0 ** 6
+
+        if sleep_seconds > 0:
+            time.sleep(sleep_seconds)
```
pgombar, line 74: I had no idea sleep() could take floats...

azurelinuxagent/ga/update.py
```diff
@@ -422,6 +422,28 @@ def _emit_restart_event(self):
...
+        try:
+            def log_if_int_changed_from_default(name, current):
+                default = conf.get_int_default_value(name)
+                if default != current:
+                    msg = "{0} changed from its default; new value: {1}".format(name, current)
```
pgombar, line 431: Suggestion, log the default value here too?

narrieta: i had it initially; too verbose, so i removed it

azurelinuxagent/ga/monitor.py
narrieta: BTW - the reason i added the try/except here is because sleep_until_next_operation in the finally was throwing on Python 2.6 (initially I was using timedelta.totalseconds() in it

larohra: Ahhh that makes sense, finally can throw too didnt think of that lol. Thanks for adding it! :)

PR #1896, opened 2020-05-28
Improvements in setup of cgroups

narrieta:
Previously the cgroups code would make several assumptions as to how cgroups were setup.

These changes discover the current setup instead of making those assumptions. Also, it sends telemetry for important parts of that setup.


azurelinuxagent/common/cgroupapi.py
```diff
@@ -64,6 +63,16 @@ def start_extension_command(self, extension_name, command, timeout, shell, cwd,
...
 
+    @staticmethod
+    def cgroups_supported():
+        distro_info = get_distro()
+        distro_name = distro_info[0]
+        try:
+            distro_version = FlexibleVersion(distro_info[1])
+        except ValueError:
+            return False
+        return distro_name.lower() == 'ubuntu' and distro_version.major >= 16
```
narrieta, line 74: I am focusing on Ubuntu 16, 18 & 20 so this is hardcoded for now

azurelinuxagent/common/cgroupapi.py
```diff
@@ -202,6 +212,58 @@ def _add_process_to_cgroup(pid, cgroup_path):
         fileutil.append_file(tasks_file, "{0}\n".format(pid))
         logger.info("Added PID {0} to cgroup {1}".format(pid, cgroup_path))
 
+    @staticmethod
```
narrieta, line 215: moved this from osutils

azurelinuxagent/common/cgroupapi.py
```diff
@@ -386,23 +538,6 @@ def _get_extensions_slice_root_name():
     def _get_extension_slice_name(self, extension_name):
         return "system-{0}-{1}.slice".format(EXTENSIONS_ROOT_CGROUP_NAME, self._get_extension_cgroup_name(extension_name))
 
-    def create_agent_cgroups(self):
```
narrieta, line 389: systemd creates the cgroups for the agent so this function was misleading. now, initialize() discovers the cgroups for the agent and starts tracking for them

azurelinuxagent/daemon/main.py
```diff
@@ -149,9 +148,6 @@ def daemon(self, child_args=None):
         # that require the goal state and IMDS
         self._initialize_telemetry()
 
-        # Initialize the agent cgroup
```
narrieta, line 152: I am doing all the cgroups stuff in the extension handler. Systemd does all the setup and the handler simply discovers the setup and starts tracking the cgroups. When I get to non-systemd distros I will have to do some of the setup in the daemon, but for now it is not needed.

azurelinuxagent/common/event.py
```diff
@@ -688,7 +689,7 @@ def initialize_event_logger_vminfo_common_parameters(protocol, reporter=__event_
     reporter.initialize_vminfo_common_parameters(protocol)
 
 
-def add_event(name, op=WALAEventOperation.Unknown, is_success=True, duration=0, version=str(CURRENT_VERSION),
+def add_event(name=AGENT_NAME, op=WALAEventOperation.Unknown, is_success=True, duration=0, version=str(CURRENT_VERSION),
```
narrieta, line 692: I defaulted the 'name' parameter to the agent; makes the calls to add_event a little more readable

tests/common/osutil/test_default_osutil.py
```diff
@@ -17,167 +17,8 @@
 
 from azurelinuxagent.common.osutil.default import DefaultOSUtil, shellutil
 from tests.tools import AgentTestCase, patch
-import os
 
 
 class DefaultOsUtilTestCase(AgentTestCase):
-
```
narrieta, line 24: Moved those to test_cgroupsapi.py/MountCgroupsTestCase

tests/common/test_cgroupconfigurator.py
```diff
@@ -17,117 +17,85 @@
...
-
-    def tearDown(self):
-        self.mock_cgroups_file_system_root.stop()
+    @staticmethod
+    def _get_new_cgroup_configurator_instance(initialize=True):
+        CGroupConfigurator._instance = None
+        configurator = CGroupConfigurator.get_instance()
+        if initialize:
+            with patch('azurelinuxagent.common.cgroupapi.CGroupsApi.cgroups_supported', return_value=True):
+                with patch('azurelinuxagent.common.cgroupapi.CGroupsApi._is_systemd', return_value=True):
```
pgombar, line 44: Before this PR, we were forcing FileSystemCgroupsApi here. Any reason for the change? I forget why filesystem was favoured before.

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -36,48 +37,131 @@ class CGroupConfigurator(object):
...
+            try:
+                if self._initialized:
+                    return
+
+                #
+                # check whether cgroup monitoring is supported on the current distro
+                #
+                self._cgroups_supported = CGroupsApi.cgroups_supported()
+                if not self._cgroups_supported:
+                    logger.info("Cgroup monitoring is not supported on {0}", get_distro())
```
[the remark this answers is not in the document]
narrieta: The list of supported distros is hardcoded (ubuntu 16, 18, 20) so no need for telemetry

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -36,48 +37,131 @@ class CGroupConfigurator(object):
...
+                if cpu_cgroup_relative_path is None:
+                    log_cgroup_warn("The agent's process is not within a CPU cgroup")
+                else:
+                    cpu_accounting = self._cgroups_api.get_unit_property(agent_unit_name, "CPUAccounting")
+                    log_cgroup_info('CPUAccounting: {0}', cpu_accounting)
+
+                if memory_cgroup_relative_path is None:
+                    log_cgroup_warn("The agent's process is not within a memory cgroup")
+                else:
+                    cpu_accounting = self._cgroups_api.get_unit_property(agent_unit_name, "MemoryAccounting")
```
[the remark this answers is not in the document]
narrieta: thanks!

tests/common/test_cgroupconfigurator.py
narrieta: Yes, with these changes FileSystemCgroupsApi is dead code. I do not want to delete it yet since we will eventually get to distros without systemd but for now all the code is going with the systemd path. I updated these tests to follow the same

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -36,48 +37,131 @@ class CGroupConfigurator(object):
...
 
-            add_event(
-                AGENT_NAME,
-                version=CURRENT_VERSION,
-                op=WALAEventOperation.InitializeCGroups,
-                is_success=self._enabled,
-                message=status,
-                log_event=False)
+            except Exception as e:
+                message = "Error initializing cgroups: {0}".format(ustr(e))
```
larohra, line 148: Not sure if needed, but shouldn't we mark `self._cgroups_enabled = False` here?

tests/common/mock_cgroup_commands.py
```diff
@@ -0,0 +1,104 @@
...
+@contextlib.contextmanager
+def mock_cgroup_commands():
+    original_popen = subprocess.Popen
+    original_read_file = fileutil.read_file
+    original_path_exists = os.path.exists
+
+    def mock_popen(command, *args, **kwargs):
+        if isinstance(command, list):
+            command_string = " ".join(command)
+            for c in _default_commands:
```
larohra, line 84: Curious to know why not just use a dict inplace of a list for `_default_commands`. Would definitely save you the trouble of walking the list everytime

azurelinuxagent/common/cgroupconfigurator.py
narrieta: initially I was setting enabled to True at the very end of the try clause, but then I realized cgroups should be ready to use if we reach line 133 so I moved the assignment there. If for any reason we cannot track the agent cgroups we should still track extensions; if for any reason we cannot track the agent's memory cgroup we should still track its cpu cgroup

tests/common/mock_cgroup_commands.py
narrieta: initially it was a dict, but then i realized strict string comparison was not enough for I need to do and change it to a sequential list of regular expressions so each time I do a sequential match.

now, most of those regular expressions are simple strings so at some point I considered having a dict search followed by a sequential regex match, but the tests actually run fast enough as it is, so I gave up the idea.

PR #1899, opened 2020-06-01
Use controller mount point for extension cgroup path

narrieta:
A small change in how we create the path for extensions cgroups. Instead of a hardcoded path now we use the mount point for the controllers.

Also, some test updates.

### DIFF
--- a/azurelinuxagent/common/cgroupapi.py
+++ b/azurelinuxagent/common/cgroupapi.py
@@ -22,7 +22,7 @@ import subprocess
 import uuid
 
 from azurelinuxagent.common import logger
-from azurelinuxagent.common.cgroup import CGroup
+from azurelinuxagent.common.cgroup import CGroup, CpuCgroup, MemoryCgroup
 from azurelinuxagent.common.cgroupstelemetry import CGroupsTelemetry
 from azurelinuxagent.common.conf import get_agent_pid_file_path
 from azurelinuxagent.common.event import add_event, WALAEventOperation
@@ -430,6 +430,8 @@ class SystemdCgroupsApi(CGroupsApi):
     """
     Cgroups interface via systemd
     """
+    def __init__(self):
+        self._cgroup_mountpoints = None
 
     @staticmethod
     def get_systemd_version():
@@ -440,8 +442,7 @@ class SystemdCgroupsApi(CGroupsApi):
         #
         return shellutil.run_command(['systemctl', '--version'])
 
-    @staticmethod
-    def get_cpu_and_memory_mount_points():
+    def get_cgroup_mount_points(self):
         """
         Returns a tuple with the mount points for the cpu and memory controllers; the values can be None
         if the corresponding controller is not mounted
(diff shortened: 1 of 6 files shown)


azurelinuxagent/common/cgroupapi.py
```diff
@@ -606,27 +627,45 @@ def _is_systemd_failure(scope_name, process_output):
...
+        try:
+            # systemd-run creates the scope under the system slice by default
+            cgroup_relative_path = os.path.join('system.slice', scope_name)
+
+            cpu_cgroup_mountpoint, memory_cgroup_mountpoint = self.get_cgroup_mount_points()
+
+            if cpu_cgroup_mountpoint is None:
+                logger.info("The CPU controller is not mounted; will not track resource usage")
+            else:
+                cpu_cgroup_path = os.path.join(cpu_cgroup_mountpoint, cgroup_relative_path)
```
larohra, line 654: Just curious, would this path and the path from `get_process_cgroup_paths` be same?

tests/common/test_cgroupconfigurator.py
```diff
@@ -130,7 +130,8 @@ def raise_exception(*_):
                     message = args[0]
                     self.assertIn("A TEST EXCEPTION", message)
 
-    def test_start_extension_command_should_not_use_systemd_when_groups_are_not_enabled(self):
+    @patch('time.sleep', side_effect=lambda _: mock_sleep())
+    def test_start_extension_command_should_not_use_systemd_when_groups_are_not_enabled(self, _):
```
larohra, line 134: NIT typo - 
`test_start_extension_command_should_not_use_systemd_when_**cgroups**_are_not_enabled`

tests/common/test_cgroupconfigurator.py
```diff
@@ -150,45 +151,30 @@ def test_start_extension_command_should_not_use_systemd_when_groups_are_not_enab
...
-            if command in command_arg:
-                return original_popen("date", *args, **kwargs)
-            else:
-                return original_popen(command_arg, *args, **kwargs)
-
-        with patch("azurelinuxagent.common.cgroupapi.subprocess.Popen", side_effect=mock_popen) as patcher:
-            patcher.get_command_calls = lambda: [args[0] for args, _ in patcher.call_args_list if len(args) > 0 and command in args[0]]
-            yield patcher
+    @patch('time.sleep', side_effect=lambda _: mock_sleep())
+    def test_start_extension_command_should_use_systemd_run_when_groups_are_enabled(self, _):
```
larohra, line 155: NIT typo - 
`test_start_extension_command_should_use_systemd_run_when_**cgroups**_are_enabled`

azurelinuxagent/common/cgroupapi.py
narrieta: Yes. I was planning on using that function, but then I realize I do not have the process of the extension (I do have the process id of systemd-run, which the agent calls, but not of the extension itself.)

The documentation of systemd-run specifies that the cgroup is created under the system slice (unless changed by the --slice argument) so this should be safe.

tests/common/test_cgroupconfigurator.py
narrieta: thanks

tests/common/test_cgroupconfigurator.py
narrieta: thanks

azurelinuxagent/common/cgroupapi.py
larohra: Ahh ok makes sense. 
I have another question (possibly outside the scope of this PR), why dont we create a new slice for each extension when invoking them? Using the `--slice` argument that you mentioned? Or were we actually doing that at some point but reverted to the system slice for some unknown issues?

narrieta: Yes, we already have the code that creates the slices and starts extensions in them, but it is not enabled because some versions of systemd simply ignore the --slice argument. As I start working on extensions I will add code to use those features when they are available.

tests/common/mock_cgroup_commands.py
```diff
@@ -64,13 +64,14 @@
 '''
 Running scope as unit: TEST_UNIT.scope
 Thu 28 May 2020 07:25:55 AM PDT
-''')
+'''),
```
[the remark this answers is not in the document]
narrieta: yes, makes adding more items a little simple

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -239,15 +241,15 @@ def start_extension_command(self, extension_name, command, timeout, shell, cwd,
...
-                extension_cgroups, process_output = self._cgroups_api.start_extension_command(extension_name,
-                                                                                              command,
-                                                                                              timeout,
-                                                                                              shell=shell,
-                                                                                              cwd=cwd,
-                                                                                              env=env,
-                                                                                              stdout=stdout,
-                                                                                              stderr=stderr,
-                                                                                              error_code=error_code)
+                process_output = self._cgroups_api.start_extension_command(extension_name,
```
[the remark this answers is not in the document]
narrieta: FileSystemCgroupsApi is at this point dead code. I am keeping there because I will use some of that code when I get to distros without systemd

PR #1902, opened 2020-06-04
Add API for uploading logs via host plugin

pgombar:
## Description

Continuation of the work started in #1847. These changes update the wire client, wire protocol and host plugin with an API for uploading logs.

---

### PR information


azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -60,7 +70,7 @@ def __init__(self, endpoint, container_id, role_config_name):
         self.api_versions = None
         self.endpoint = endpoint
         self.container_id = container_id
-        self.deployment_id = None
+        self.deployment_id = self._extract_deployment_id(role_config_name)
```
larohra, line 73: I wonder if we should make the deployment ID as an parameter of `AgentGlobal` class to ensure that its always uptodate. Or is this ID dependent on each GS? Any idea?

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -202,7 +217,29 @@ def should_report(is_healthy, error_state, last_timestamp, period):
...
     def put_vm_log(self, content):
-        raise NotImplementedError("Unimplemented")
+        """
+        Try to upload VM logs, a compressed zip file, via the host plugin /vmAgentLog channel.
+        :param content: the binary content of the zip file to upload
+        """
+        if not self.ensure_initialized():
+            raise ProtocolError("HostGAPlugin: HostGAPlugin is not available")
+
+        if content is None or self.container_id is None or self.deployment_id is None:
```
larohra, line 227: Not sure if you already investigated this, but we should pick up containerId (and maybe deploymentId) from AgentGlobals class instead as that data is always up-to-date

larohra: Or do we handle that externally (outside this class)?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1195,14 +1196,73 @@ def get_artifacts_profile(self):
...
+            msg = "Upload VM logs request succeeded using the host plugin channel for " \
+                  "container id {0} and role config file {1}".format(host_plugin.container_id,
+                                                                     host_plugin.role_config_name)
+            add_event(name=AGENT_NAME,
+                      version=CURRENT_VERSION,
+                      op=WALAEventOperation.HostPlugin,
+                      is_success=True,
+                      message=msg,
+                      log_event=True)
+        except (ResourceGoneError, InvalidContainerError) as e:
```
larohra, line 1220: If we use the AgentGlobal class here, we can avoid the extra refresh-retry because that object always has the latest guids :)

tests/data/wire/goal_state.xml
```diff
@@ -12,15 +12,15 @@
      <ContainerId>c6d5526c-5ac2-4200-b6e2-56f2b70c5ab2</ContainerId>
      <RoleInstanceList>
        <RoleInstance>
-         <InstanceId>MachineRole_IN_0</InstanceId>
+         <InstanceId>b61f93d0-e1ed-40b2-b067-22c243233448.MachineRole_IN_0</InstanceId>
```
larohra, line 15: From these tests it seems that the might not always be present in the RoleConfigName. Is this guid necessary for putting logs in the host? Is it always guaranteed to be there? Any idea?

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -20,29 +20,39 @@
 import base64
 import datetime
 import json
+import uuid
```
larohra, line 23: I didnt see any use of this in your changes (maybe I missed something). Just confirming if this is an empty import or not

tests/protocol/test_hostplugin.py
```diff
@@ -536,6 +538,44 @@ def test_validate_page_blobs(self):
                         test_goal_state,
                         exp_method, exp_url, exp_data)
 
+    def test_validate_http_request_when_uploading_logs(self):
+        with mock_wire_protocol(DATA_FILE) as protocol:
+            test_goal_state = protocol.client._goal_state
```
larohra, line 543: NIT: Why not use `protocol.client.get_goal_state()` instead.

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -20,29 +20,39 @@
...
 
 from azurelinuxagent.common import logger
 from azurelinuxagent.common.errorstate import ErrorState, ERROR_STATE_HOST_PLUGIN_FAILURE
+from azurelinuxagent.common.event import WALAEventOperation, report_event
```
narrieta, line 27: report_event is practically identical to add_event. we should use only one of them to avoid confussions (add_event seems to be the most common)

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -20,29 +20,39 @@
...
+
 URI_FORMAT_GET_API_VERSIONS = "http://{0}:{1}/versions"
 URI_FORMAT_GET_EXTENSION_ARTIFACT = "http://{0}:{1}/extensionArtifact"
 URI_FORMAT_PUT_VM_STATUS = "http://{0}:{1}/status"
 URI_FORMAT_PUT_LOG = "http://{0}:{1}/vmAgentLog"
 URI_FORMAT_HEALTH = "http://{0}:{1}/health"
+
 API_VERSION = "2015-09-01"
+
+HEADER_CLIENT_NAME = "x-ms-client-name"
```
narrieta, line 46: private

azurelinuxagent/common/protocol/hostplugin.py
narrieta, line 227: why this part of the check here?

>  self.container_id is None or self.deployment_id is None

that should probably be where those properties are set

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -202,7 +217,29 @@ def should_report(is_healthy, error_state, last_timestamp, period):
...
+            raise ProtocolError("HostGAPlugin: HostGAPlugin is not available")
+
+        if content is None or self.container_id is None or self.deployment_id is None:
+            raise ProtocolError("HostGAPlugin: Invalid arguments passed to upload VM logs. "
+                                "Content: {0}, container id: {1}, deployment id: {2}".format(content,
+                                                                                             self.container_id,
+                                                                                             self.deployment_id))
+
+        url = URI_FORMAT_PUT_LOG.format(self.endpoint, HOST_PLUGIN_PORT)
+        response = restutil.http_put(url,
```
narrieta, line 234: Can http_put raise?

tests/protocol/test_hostplugin.py
```diff
@@ -536,6 +538,44 @@ def test_validate_page_blobs(self):
...
+                                "x-ms-client-version": AGENT_VERSION,
+                                "x-ms-client-correlationid": correlation_id}
+
+            host_client = wire.HostPluginProtocol(wireserver_url,
+                                                  test_goal_state.container_id,
+                                                  test_goal_state.role_config_name)
+
+            self.assertFalse(host_client.is_initialized, "Host plugin should not be initialized!")
+
+            with patch.object(restutil, "http_request") as patch_http:
```
narrieta, line 560: try using the http_handlers in mock_wire_protocol instead of mocking http_request

tests/protocol/test_hostplugin.py
```diff
@@ -536,6 +538,44 @@ def test_validate_page_blobs(self):
...
+                with patch.object(wire.HostPluginProtocol, "get_api_versions", return_value=api_versions):
```
narrieta, line 561: mock_wire_protocol should already handle the call to api versions

tests/protocol/test_hostplugin.py
```diff
@@ -536,6 +538,44 @@ def test_validate_page_blobs(self):
...
+                    with patch("azurelinuxagent.common.protocol.hostplugin.uuid.uuid4", return_value=correlation_id):
```
narrieta, line 562: this mock goes too deep into the implementation in my opinion, plus in the future uuid may be used elsewhere in the plugin making this mock too fragile. not sure it is worth checking the exact value; I believe what you want to test is that it uses a unique id? that may be too hard to test, maybe just that it uses a guid (i.e. the value matches the pattern for a guid)

azurelinuxagent/common/protocol/wire.py
narrieta: We should try to minimize dependencies on those globals. I think the host ga plugin doesn't need that dependency since it only needs those values on initialization when an http request returns ResourceGone. Telemetry is different in that we would need  to fetch the goal state for each event in order to get the updated values; in that case the dependency on the global is a good trade  off.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1195,14 +1196,73 @@ def get_artifacts_profile(self):
...
+
+            msg = "Upload VM logs request succeeded using the host plugin channel for " \
+                  "container id {0} and role config file {1}".format(host_plugin.container_id,
+                                                                     host_plugin.role_config_name)
+            add_event(name=AGENT_NAME,
+                      version=CURRENT_VERSION,
+                      op=WALAEventOperation.HostPlugin,
+                      is_success=True,
+                      message=msg,
+                      log_event=True)
```
narrieta, line 1219: log_event=True is misleading, since the message is logged only if is_success==False

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1195,14 +1196,73 @@ def get_artifacts_profile(self):
...
                                  log_event=False)
 
         return artifacts_profile
 
+    def upload_logs(self, content):
+        try:
+            host_plugin = self.get_host_plugin()
+            host_plugin.put_vm_log(content)
+
+            msg = "Upload VM logs request succeeded using the host plugin channel for " \
```
narrieta, line 1211: "using the host plugin channel with" ?

azurelinuxagent/common/protocol/wire.py
narrieta: or did you actually intend to log the message? most likely yes: you will need a separate call to the logger

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1195,14 +1196,73 @@ def get_artifacts_profile(self):
...
+            self.update_host_plugin_from_goal_state()
+
+            new_container_id = host_plugin.container_id
+            new_role_config_name = host_plugin.role_config_name
+            msg = "Host plugin reconfigured with new parameters. " \
+                  "ContainerId: {0}, role config file: {1}.".format(new_container_id, new_role_config_name)
+            logger.info(msg)
+
+            try:
+                host = self.get_host_plugin()
```
narrieta, line 1239: there is too much code repetition here

tests/protocol/test_wire.py
```diff
@@ -815,6 +815,41 @@ def http_get_handler(url, *_, **kwargs):
             self.assertTrue(self.is_host_plugin_extension_artifact_request(urls[3]), "The retry request should have been over the host channel")
             self.assertEquals(HostPluginProtocol.is_default_channel(), False, "The default channel should not have changed")
 
+    def test_upload_logs_should_not_refresh_plugin_when_first_attempt_succeeds(self):
+        def http_put_handler(url, *_, **__):
+            if protocol.get_endpoint() in url and url.endswith('/vmAgentLog'):
```
narrieta, line 820: let's create a predicate for this condition in HttpRequestPredicates

tests/protocol/test_wire.py
```diff
@@ -815,6 +815,41 @@ def http_get_handler(url, *_, **kwargs):
...
+                return MockResponse(body=b'', status_code=200)
+            self.fail('The upload logs request was sent to the wrong url: {0}'.format(url))
```
narrieta, line 822: but at this point you do not know that it was an upload log request, you only know that it was a PUT request

azurelinuxagent/common/protocol/wire.py
larohra: On the contrary, any HostGA request would fail if the containerId is not correct. Wouldn't it make more sense to always send the most up-to-date value we have for containerId by using AgentGlobals rather than waiting for it to fair and then retrying again. This adds a delay in the turn-around-time for the customer at the end of the day (even though minimal) so I feel we can go for that.

Also just for my understanding, why should we try to avoid dependencies on AgentGlobals? Like the way I see it is the values which always need to be up-to-date and are independent of different objects can be part of the AgentGlobals class and should be used wherever possible.

tests/protocol/test_wire.py
narrieta: seems like you added it, but did not use it

azurelinuxagent/common/protocol/wire.py
narrieta: As a general rule globals should be avoided/used sparingly. I don't buy the concern about turnaround time since the extra requests should be rare.

azurelinuxagent/common/protocol/hostplugin.py
pgombar: It's used in one of the headers for the log upload request: `HEADER_CORRELATION_ID: str(uuid.uuid4())`.

azurelinuxagent/common/protocol/hostplugin.py
pgombar: I moved this import from inline to the top without paying attention to it. I made the change now to add_event.

azurelinuxagent/common/protocol/hostplugin.py
pgombar: Thanks, made all headers private.

tests/data/wire/goal_state.xml
pgombar: The tests didn't contain this parameter simply because we never knew about it / needed it until now. Real world goal states come with this parameter.

azurelinuxagent/common/protocol/wire.py
pgombar: Thanks, I did want to log this. Updated.

azurelinuxagent/common/protocol/wire.py
pgombar: I thought about refactoring this into its own little method:
```
def _get_host_and_put_logs(content, msg):
    host = self.get_host_plugin()
    host.put_vm_log(content)

    logger.info(msg)
    add_event(name=AGENT_NAME,
                version=CURRENT_VERSION,
                op=WALAEventOperation.HostPlugin,
                is_success=True,
                message=msg,
                log_event=False)
```
I didn't because I didn't think it added a lot of value. If you this it does or you have better suggestions how to make this method more readable, please let me know!

azurelinuxagent/common/protocol/wire.py
pgombar: Keep in mind this method would be called every hour by default, not more often. I think it makes more sense to use AgentGlobals in the telemetry code where we rely on fresh and up-to-date info instead of here.

What if the AgentGlobals happen to be out of sync here? (We just exposed a polling interval for goal state updates, so we can't rely on it being every 3 seconds and having such a narrow window.) Should this method be responsible for updating the globals? It seems cleaner to let that responsibility lie in the constructor of the goal state and let host plugin update itself and its dependencies.

azurelinuxagent/common/protocol/hostplugin.py
pgombar: Deployment ID is a part of each goal state, but I'm not sure when/how it changes. See discussion below about adding AgentGlobals as a dependency on this API.

azurelinuxagent/common/protocol/hostplugin.py
pgombar: You are right, I wanted to check for `content` as it's a passed argument and accidentally included these.

azurelinuxagent/common/protocol/hostplugin.py
pgombar: It can, thanks for catching that. I added a catch for that (and raise `ProtocolError`) in the caller of this method.

tests/protocol/test_hostplugin.py
pgombar: Thanks!

tests/protocol/test_hostplugin.py
pgombar: Thanks, updated.

tests/protocol/test_wire.py
pgombar: Thanks, updated.

tests/protocol/test_wire.py
pgombar: I was looking at how the test `test_upload_status_blob_should_use_the_host_channel_by_default` was written:
`self.fail('The upload status request was sent to the wrong url: {0}'.format(url))`.

Since this inline put handler is defined in the test method that's (only) calling the upload log (or upload status) method, isn't it safe to assume we are only expecting PUT requests to those URLs?

azurelinuxagent/common/protocol/wire.py
larohra: That's a very fair point, I did not think of that. Given that, I agree that we should refresh host plugin object on demand rather than using the agent global in this case.

azurelinuxagent/common/protocol/hostplugin.py
larohra: Got it. Thanks for explaining it! :)

azurelinuxagent/common/protocol/wire.py
larohra: Suggestion: Can't you make the function a lambda and make the whole retry-and-update-host-logic a helper function that you use here and in the send_appropriate_request_function too?

larohra: I think Joe added a helper function like that in his PR. You could take a look at that too for inspiration

tests/protocol/test_hostplugin.py
```diff
@@ -539,42 +539,48 @@ def test_validate_page_blobs(self):
                         exp_method, exp_url, exp_data)
 
     def test_validate_http_request_when_uploading_logs(self):
```
larohra, line 541: Probably not a good test for the current changes but just putting this comment here incase I forget later, make sure to add a test to ensure that failures are handled gracefully by the agent (for example if the PUT call fails and you return a ProtocolError).

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -202,7 +217,26 @@ def should_report(is_healthy, error_state, last_timestamp, period):
         return datetime.datetime.utcnow() >= (last_timestamp + period)
 
     def put_vm_log(self, content):
-        raise NotImplementedError("Unimplemented")
+        """
+        Try to upload VM logs, a compressed zip file, via the host plugin /vmAgentLog channel.
+        :param content: the binary content of the zip file to upload
```
narrieta, line 222: wow, so the host ga plugin requires to load the whole file to memory? what is the max size of the zip, again?

azurelinuxagent/common/protocol/hostplugin.py
narrieta: the caller coming on a different PR?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1195,14 +1196,78 @@ def get_artifacts_profile(self):
...
```
narrieta, line 1241: just curious: why do we keep calling self.get_host_plugin() and why use a different variable here? (multiple calls are ok, but the use of a new variable threw me off)

azurelinuxagent/common/protocol/wire.py
narrieta: we have multiple versions of the "try the host plugin, catch resource gone/invalid container, get a new goal state, update the plugin, and now retry" logic. it'd be good to capture that logic in a single method in the plugin and use it from the other places. 

or use joe's function if appropriate

tests/protocol/test_hostplugin.py
narrieta: +1

the code to handle the error is there already so we should include the test in this pr

tests/protocol/test_hostplugin.py
```diff
@@ -536,6 +538,50 @@ def test_validate_page_blobs(self):
                         test_goal_state,
                         exp_method, exp_url, exp_data)
 
+    def test_validate_http_request_when_uploading_logs(self):
+        def http_put_handler(url, *args, **kwargs):
+            if self.is_host_plugin_put_logs_request(url):
+                http_put_handler.args, http_put_handler.kwargs = args, kwargs
+                return MockResponse(body=b'', status_code=200)
+            self.fail('The upload logs request was sent to the wrong url: {0}'.format(url))
```
narrieta, line 546: the mock handler should handle the requests it is interested in and let the others go thru (instead of failing the test). this allows for cases in which, as the code evolves, additional requests are done. the mock protocol would raise if nobody handles the request

tests/protocol/test_wire.py
narrieta: I think I misread your code initially, you are checking is_host_plugin_put_logs_request

as far as the self.fail, let's remove it (see my previous comment). when i initially wrote the mock protocol, it did not raise on unhandled  requests so the fail in test_upload_status_blob_should_use_the_host_channel_by_default must be a left over; you can remove it, too, if you want

azurelinuxagent/common/protocol/hostplugin.py
pgombar: I haven't figured out a way to not load it. Everything we send using `restutil` loads the object in-memory. The restriction on the agent side for the uncompressed archive is 150 MB (from my testing on actual log data, this will result in a compressed archive of 13-15 MB). On the host, the restriction is 100 MB.

azurelinuxagent/common/protocol/wire.py
pgombar: Used Joe's approach.

tests/protocol/test_hostplugin.py
pgombar: Added a test to verify the host plugin method raises on a failed request. Since this exception is propagated to the protocol, which propagates it to the log uploading thread (coming in the next PR), I will add another test then to make sure we are handling it gracefully.

azurelinuxagent/common/protocol/wire.py
pgombar: This is outdated now. I didn't have a reason why I was doing it like that.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -875,41 +879,8 @@ def check_wire_protocol_version(self):
...
-            except (ResourceGoneError, InvalidContainerError) as e:
-                logger.periodic_info(logger.EVERY_HOUR, "[PERIODIC] Request failed using the direct channel, "
-                                                        "switching to host plugin. Error: {0}".format(ustr(e)))
-
-            if ret:
-                return ret
-        else:
-            logger.periodic_info(logger.EVERY_HALF_DAY, "[PERIODIC] Using host plugin as default channel.")
-
+    def call_hostplugin_with_container_check(self, host_func):
```
narrieta, line 882: is this function being used?

narrieta, line 882: private?

azurelinuxagent/common/protocol/hostplugin.py
pgombar: In theory, it could. However, from the memory point of view, all this is adding is an overhead of ~15 MB (worst case scenario) every hour. Disk I/O could be a bigger problem, but we don't have any plans for limiting that in the near future.

PR #1903, opened 2020-06-06
Handle failed HTTP responses and setting host as default channel

pgombar:
## Description

Today, if we get an unsuccessful HTTP response, we will not handle it and continue parsing it thinking the data in the response is correct. The methods vulnerable to this are:
1) `azurelinuxagent.common.protocol.wire.WireProtocol.download_ext_handler_pkg`
2) `azurelinuxagent.common.protocol.wire.WireClient.fetch_manifest`
3) `azurelinuxagent.common.protocol.wire.WireClient.get_artifacts_profile`

Method 1) calls `azurelinuxagent.common.protocol.wire.WireClient.stream` and methods 2) and 3) call `azurelinuxagent.common.protocol.wire.WireClient.fetch` to get the content from the HTTP response.

This PR fixes two bugs:
1) Fix the handling of unsuccessful responses in `azurelinuxagent.common.protocol.wire.WireClient.stream` and `azurelinuxagent.common.protocol.wire.WireClient.fetch` to first check if the response was successful before parsing the data, and
2) Fix setting the host plugin channel as default in `azurelinuxagent.common.protocol.wire.WireClient.send_request_using_appropriate_channel`. The host plugin will only be set as default if the request succeed.

---

### PR information

### DIFF
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -961,7 +961,8 @@ class WireClient(object):
                              log_event=True)
                 raise
 
-        if ret is not None and not HostPluginProtocol.is_default_channel():
+        # Depending on the lambda, the unsuccessful return value can be None or False
+        if ret and not HostPluginProtocol.is_default_channel():
             logger.info("Setting host plugin as default channel from now on. "
                         "Restart the agent to reset the default channel.")
             HostPluginProtocol.set_default_channel(True)
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -155,7 +155,7 @@ class WireProtocol(DataContract):
         success = self.client.stream(uri, destination, headers=headers, use_proxy=False)
         return success
 
-    def download_ext_handler_pkg(self, uri, destination, headers=None, use_proxy=True):
+    def download_ext_handler_pkg(self, uri, destination):
         direct_func = lambda: self.client.stream(uri, destination, headers=None, use_proxy=True)
         # NOTE: the host_func may be called after refreshing the goal state, be careful about any goal state data
         # in the lambda.
@@ -628,7 +628,7 @@ class WireClient(object):
         logger.verbose("Fetch [{0}] with headers [{1}] to file [{2}]", uri, headers, destination)
 
         response = self._fetch_response(uri, headers, use_proxy)
-        if response is not None:
+        if not restutil.request_failed(response):
             chunk_size = 1024 * 1024  # 1MB buffer
             try:
                 with open(destination, 'wb', chunk_size) as destination_fh:
@@ -647,7 +647,7 @@ class WireClient(object):
         logger.verbose("Fetch [{0}] with headers [{1}]", uri, headers)
         content = None
         response = self._fetch_response(uri, headers, use_proxy)
-        if response is not None:
+        if not restutil.request_failed(response):
             response_content = response.read()
             content = self.decode_config(response_content) if decode else response_content
         return content
@@ -961,7 +961,7 @@ class WireClient(object):
                              log_event=True)
                 raise
 
-        if not HostPluginProtocol.is_default_channel():
+        if ret and not HostPluginProtocol.is_default_channel():
             logger.info("Setting host plugin as default channel from now on. "
                         "Restart the agent to reset the default channel.")
             HostPluginProtocol.set_default_channel(True)
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -961,7 +961,7 @@ class WireClient(object):
                              log_event=True)
                 raise
 
-        if ret and not HostPluginProtocol.is_default_channel():
+        if ret is not None and not HostPluginProtocol.is_default_channel():
             logger.info("Setting host plugin as default channel from now on. "
                         "Restart the agent to reset the default channel.")
             HostPluginProtocol.set_default_channel(True)
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -163,6 +163,9 @@ class WireProtocol(DataContract):
 
         try:
             success = self.client.send_request_using_appropriate_channel(direct_func, host_func)
+
+            if not success:
+                logger.warn("Failed to download extension handler package from URI {0}".format(uri))
         except Exception:
             success = False
 
@@ -618,6 +621,8 @@ class WireClient(object):
                     host = self.get_host_plugin()
                     host.update_manifest_uri(version.uri)
                     return response
+                else:
+                    logger.warn("Failed to fetch manifest from URI {0}".format(version.uri))
             except Exception as e:
                 logger.warn("Exception when fetching manifest. Error: {0}".format(ustr(e)))
 
@@ -1201,6 +1206,8 @@ class WireClient(object):
                                  is_success=False,
                                  message=msg,
                                  log_event=False)
+            else:
+                logger.warn("Failed to retrieve artifacts profile from blob {0}".format(blob))
 
         return artifacts_profile
 
--- a/tests/protocol/test_wire.py
+++ b/tests/protocol/test_wire.py
@@ -481,9 +481,11 @@ class TestWireClient(HttpRequestPredicates, AgentTestCase):
             urls = protocol.get_tracked_urls()
             self.assertEquals(success, True, 'The download should have succeeded')
             self.assertEquals(len(urls), 1, "Unexpected number of HTTP requests: [{0}]".format(urls))
-            self.assertEquals(urls[0], extension_url, "The extension should have been downloaded over the direct channel")
+            self.assertEquals(urls[0], extension_url,
+                              "The extension should have been downloaded over the direct channel")
             self.assertTrue(os.path.exists(target_file), 'The extension package was not downloaded')
-            self.assertEquals(HostPluginProtocol.is_default_channel(), False, "The host channel should not have been set as the default")
+            self.assertEquals(HostPluginProtocol.is_default_channel(), False,
+                              "The host channel should not have been set as the default")
 
     def test_download_ext_handler_pkg_should_use_host_channel_when_direct_channel_fails_and_set_host_as_default(self):
         extension_url = 'https://fake_host/fake_extension.zip'
@@ -505,13 +507,15 @@ class TestWireClient(HttpRequestPredicates, AgentTestCase):
             self.assertEquals(success, True, 'The download should have succeeded')
             self.assertEquals(len(urls), 2, "Unexpected number of HTTP requests: [{0}]".format(urls))
             self.assertEquals(urls[0], extension_url, "The first attempt should have been over the direct channel")
-            self.assertTrue(self.is_host_plugin_extension_artifact_request(urls[1]), "The retry attempt should have been over the host channel")
+            self.assertTrue(self.is_host_plugin_extension_artifact_request(urls[1]),
+                            "The retry attempt should have been over the host channel")
             self.assertTrue(os.path.exists(target_file), 'The extension package was not downloaded')
-            self.assertEquals(HostPluginProtocol.is_default_channel(), True, "The host channel should have been set as the default")
+            self.assertEquals(HostPluginProtocol.is_default_channel(), True,
+                              "The host channel should have been set as the default")
 
     def test_download_ext_handler_pkg_should_retry_the_host_channel_after_refreshing_host_plugin(self):
-        extension_url = 'https://fake_host/fake_extension.zip'
-        target_file = os.path.join(self.tmp_dir, 'fake_extension.zip')
+        extension_url = "https://fake_host/fake_extension.zip"
+        target_file = os.path.join(self.tmp_dir, "fake_extension.zip")
 
         def http_get_handler(url, *args, **kwargs):
             if url == extension_url:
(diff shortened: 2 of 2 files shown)


azurelinuxagent/common/protocol/wire.py
```diff
@@ -155,7 +155,7 @@ def _download_ext_handler_pkg_through_host(self, uri, destination):
         success = self.client.stream(uri, destination, headers=headers, use_proxy=False)
         return success
 
-    def download_ext_handler_pkg(self, uri, destination, headers=None, use_proxy=True):
+    def download_ext_handler_pkg(self, uri, destination):
```
pgombar, line 158: These headers were not used so I removed them.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -628,7 +628,7 @@ def stream(self, uri, destination, headers=None, use_proxy=None):
         logger.verbose("Fetch [{0}] with headers [{1}] to file [{2}]", uri, headers, destination)
 
         response = self._fetch_response(uri, headers, use_proxy)
-        if response is not None:
+        if not restutil.request_failed(response):
```
pgombar, line 631: Main change of the PR. The response can be "404: not found" which is not `None`, so according to the old condition, we would continue parsing the data here.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -647,7 +647,7 @@ def fetch(self, uri, headers=None, use_proxy=None, decode=True):
         logger.verbose("Fetch [{0}] with headers [{1}]", uri, headers)
         content = None
         response = self._fetch_response(uri, headers, use_proxy)
-        if response is not None:
+        if not restutil.request_failed(response):
```
pgombar, line 650: Main change of the PR. The response can be "404: not found" which is not `None`, so according to the old condition, we would continue parsing the data here.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -961,7 +961,7 @@ def send_request_using_appropriate_channel(self, direct_func, host_func):
                              log_event=True)
                 raise
 
-        if not HostPluginProtocol.is_default_channel():
+        if ret and not HostPluginProtocol.is_default_channel():
```
pgombar, line 964: Main change of the PR. Previously, the host plugin channel would be set as default even if `ret` was None.

tests/protocol/test_wire.py
```diff
@@ -576,10 +586,46 @@ def http_get_handler(url, *_, **kwargs):
...
+            self.assertTrue(self.is_host_plugin_extension_artifact_request(urls[1]),
+                            "The second attempt should have been over the host channel")
+            self.assertTrue(self.is_goal_state_request(urls[2]),
+                            "The host channel should have been refreshed the goal state")
+            self.assertTrue(self.is_host_plugin_extension_artifact_request(urls[3]),
+                            "The third attempt should have been over the host channel")
+            self.assertEquals(HostPluginProtocol.is_default_channel(), False,
+                              "The host channel should not have been set as the default")
+
+    def test_download_ext_handler_pkg_should_fail_if_both_channels_return_an_unsuccessful_response(self):
```
pgombar, line 598: New test. This test would fail with the old code (tested it) since the download would "succeed" if the response was 404.

tests/protocol/test_wire.py
```diff
@@ -705,6 +756,38 @@ def http_get_handler(url, *_, **kwargs):
 
             self.assertEquals(HostPluginProtocol.is_default_channel(), False)
 
+    def test_fetch_manifest_should_raise_if_both_channels_return_an_unsuccessful_response(self):
```
pgombar, line 759: New test. This test would fail with the old code (tested it) since the fetch_manifest function would return the failed response thinking it's a valid manifest.

tests/protocol/test_wire.py
```diff
@@ -809,11 +905,46 @@ def http_get_handler(url, *_, **kwargs):
...
+            self.assertTrue(self.is_host_plugin_extension_artifact_request(urls[1]),
+                            "The second request should have been over the host channel")
+            self.assertTrue(self.is_goal_state_request(urls[2]),
+                            "The goal state should have been refreshed before retrying the host channel")
+            self.assertTrue(self.is_host_plugin_extension_artifact_request(urls[3]),
+                            "The retry request should have been over the host channel")
+            self.assertEquals(HostPluginProtocol.is_default_channel(), False,
+                              "The default channel should not have changed")
+
+    def test_get_artifacts_profile_should_fail_if_both_channels_return_an_unsuccessful_response(self):
```
pgombar, line 919: New test. This test would fail with the old code (tested it) since the get_artifacts_profile function would return the failed response thinking it's a valid artifacts profile.

tests/protocol/test_wire.py
```diff
@@ -576,10 +586,46 @@ def http_get_handler(url, *_, **kwargs):
...
+            # initialization of the host plugin triggers a request for the goal state;
+            # do it here before we start tracking those requests.
+            protocol.client.get_host_plugin()
+
+            protocol.set_http_handlers(http_get_handler=http_get_handler)
+
+            success = protocol.download_ext_handler_pkg(extension_url, target_file)
+
+            urls = protocol.get_tracked_urls()
+            self.assertEquals(success, False, "The download should have failed")
```
larohra, line 621: NIT: use `self.assertFalse` instead

tests/protocol/test_wire.py
```diff
@@ -900,15 +1031,44 @@ def host_func(*args):
...
+
+            direct_func.counter = 0
+            host_func.counter = 0
+
+            # Assert we've called both the direct channel function (once) and the host channel function (twice).
+            # The second call through host channel resulted in None, so we shouldn't set the host channel as default.
+            with patch("azurelinuxagent.common.protocol.wire.WireClient.update_host_plugin_from_goal_state") \
+                    as mock_update_host_plugin_from_goal_state:
+                ret = protocol.client.send_request_using_appropriate_channel(direct_func, host_func)
+                self.assertEquals(None, ret)
```
larohra, line 1066: NIT: use `self.assertIsNone` instead

tests/protocol/test_wire.py
```diff
@@ -900,15 +1031,44 @@ def host_func(*args):
...
+                self.assertEquals(1, direct_func.counter)
+                self.assertEquals(2, host_func.counter)
+                self.assertEquals(1, mock_update_host_plugin_from_goal_state.call_count)
+                self.assertEquals(False, protocol.client.get_host_plugin().is_default_channel())
```
larohra, line 1070: NIT: self.assertFalse

tests/protocol/test_wire.py
```diff
@@ -900,15 +1031,44 @@ def host_func(*args):
...
+            direct_func.counter = 0
+            host_func.counter = 0
+
+            # Assert we've called both the direct channel function (once) and the host channel function (twice).
+            # The second call through host channel resulted in None, so we shouldn't set the host channel as default.
+            with patch("azurelinuxagent.common.protocol.wire.WireClient.update_host_plugin_from_goal_state") \
+                    as mock_update_host_plugin_from_goal_state:
+                ret = protocol.client.send_request_using_appropriate_channel(direct_func, host_func)
+                self.assertEquals(None, ret)
+                self.assertEquals(1, direct_func.counter)
```
larohra, line 1067: NIT: Could you add some error message incase this test fails just for improved debuggability

azurelinuxagent/common/protocol/wire.py
narrieta, line 631: _fetch_response can return None so it seems like the previous check is still needed

azurelinuxagent/common/protocol/wire.py
narrieta: _fetch_response can return None so it seems like the previous check is still needed

azurelinuxagent/common/protocol/wire.py
narrieta, line 964: some of the lambdas passed to send_request_using_appropriate_channel do not have an explicit return; since you are using the return value, could add an explicit return to all the lambdas?

narrieta, line 964: if this is a check against None (as opposed to a bool or an empty collection) let's be explicit: if ret is not None

azurelinuxagent/common/protocol/wire.py
narrieta, line 650: even with this change we still don't handle the error in some paths. now instead of returning a reponse with an error we return None. If you follow the call stack upwards you will see paths that ignore the None value and do not report the error.

tests/protocol/test_wire.py
pgombar: Done, thanks for catching that.

azurelinuxagent/common/protocol/wire.py
pgombar: Sounds good.

pgombar: I inspected the lambdas and couldn't find any which doesn't return. Which one are you referring to?
The lambdas in question are `azurelinuxagent.common.protocol.wire.WireClient.stream` and `azurelinuxagent.common.protocol.wire.WireClient.fetch`.

narrieta: e.g. direct_func, host_func in
```
def download_ext_handler_pkg(self, uri, destination, headers=None, use_proxy=True):
        direct_func = lambda: self.client.stream(uri, destination, headers=None, use_proxy=True)
        # NOTE: the host_func may be called after refreshing the goal state, be careful about any goal state data
        # in the lambda.
        host_func = lambda: self._download_ext_handler_pkg_through_host(uri, destination)

        try:
            success = self.client.send_request_using_appropriate_channel(direct_func, host_func)
        except Exception:
            success = False

        return success
```

narrieta: never mind, dunno what i was thinking, you are ok

pgombar: Maybe fasttrack changes are seeping in :)

pgombar: Actually, `azurelinuxagent.common.protocol.wire.WireClient.stream` is a lambda that can return False. I will make the check more explicit to include both None and False

pgombar: Decided that `if ret` looks better than `if ret is not None and if ret is not False`.

azurelinuxagent/common/protocol/wire.py
pgombar: So, the error on the request level will always be logged as a warning since this is checked in `azurelinuxagent.common.protocol.wire.WireClient._fetch_response` (called just above this line).

Inspecting the scenarios that rely on `azurelinuxagent.common.protocol.wire.WireClient.fetch`:
1) Fetching manifest: an ExtensionDownloadError will be raised since the response will not be None instead of failed response body. The exception is raised in `azurelinuxagent.common.protocol.wire.WireClient.fetch_manifest` if response is None.
2) Download extension handler package: an exception will be raised due to:
```
if not self.protocol.download_ext_handler_pkg(source_uri, target_file):
    raise Exception("Failed to download extension package - no error information is available")
```
3) Get artifacts profile: there is a check against None values in `azurelinuxagent.ga.exthandlers.ExtHandlersHandler._extension_processing_allowed`:
```
artifacts_profile = self.protocol.get_artifacts_profile()
if artifacts_profile and artifacts_profile.is_on_hold():
```

So the handling of None is already covered. However, I agree the error reporting can be improved, since we only have the warning on the request level and then all the way on the top of the call stack in cases 1) and 2), and nothing in 3). I'm adding an error log in the middle functions, `azurelinuxagent.common.protocol.wire.WireClient.get_artifacts_profile`, `azurelinuxagent.common.protocol.wire.WireClient.fetch_manifest` and `azurelinuxagent.common.protocol.wire.WireProtocol.download_ext_handler_pkg`.

pgombar: One thing that worries me about adding a log warning when requests fail in `azurelinuxagent.common.protocol.wire.WireClient.get_artifacts_profile` is that we call this method every 3 seconds. However, even today we have this potential verbosity in case of (consistent) failure since the failed requests would be logged lower in the call stack, in `azurelinuxagent.common.protocol.wire.WireClient._fetch_response`.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -155,14 +155,17 @@ def _download_ext_handler_pkg_through_host(self, uri, destination):
...
         direct_func = lambda: self.client.stream(uri, destination, headers=None, use_proxy=True)
         # NOTE: the host_func may be called after refreshing the goal state, be careful about any goal state data
         # in the lambda.
         host_func = lambda: self._download_ext_handler_pkg_through_host(uri, destination)
 
         try:
             success = self.client.send_request_using_appropriate_channel(direct_func, host_func)
+
+            if not success:
+                logger.warn("Failed to download extension handler package from URI {0}".format(uri))
```
narrieta, line 168: can we add some context as to what the problem was?

narrieta, line 168: do you need a similar message in the except clause below?

narrieta: similar comments for fetch_manifest below

narrieta, line 168: similar comments for fetch_manifest below

azurelinuxagent/common/protocol/wire.py
narrieta: let's sync offline on this

larohra: Suggestion: Since you're touching this code anyways, why not make the 3 calls consistent? Like make sure to raise an error if download fails and then handle them in their callers to reduce ambiguity. 
This is definitely a bigger change than what you started with but since you're touching the code anyways might as well make some efforts into solving it permanently :)

pgombar:
Ignore this PR for now, will come up with better changes.

pgombar:
Closing in favour of #2004

PR #1908, opened 2020-06-11
Report processes that do not belong to the agent's cgroup

narrieta:
As part of collecting resource usage metrics we now report processes that should not be in the agent's cgroup.

### DIFF
--- a/azurelinuxagent/common/cgroup.py
+++ b/azurelinuxagent/common/cgroup.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 #
 # Requires Python 2.6+ and Openssl 1.0+
+from collections import namedtuple
+
 import errno
 import os
 import re
@@ -22,6 +24,21 @@ from azurelinuxagent.common.exception import CGroupsException
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.osutil import get_osutil
 from azurelinuxagent.common.utils import fileutil
+
+
+MetricValue = namedtuple('Metric', ['category', 'counter', 'instance', 'value'])
+
+
+class MetricsCategory(object):
+    MEMORY_CATEGORY = "Memory"
+    PROCESS_CATEGORY = "Process"
+
+
+class MetricsCounter(object):
+    PROCESSOR_PERCENT_TIME = "% Processor Time"
+    TOTAL_MEM_USAGE = "Total Memory Usage"
+    MAX_MEM_USAGE = "Max Memory Usage"
+
 
 re_user_system_times = re.compile(r'user (\d+)\nsystem (\d+)\n')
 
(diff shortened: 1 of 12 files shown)


azurelinuxagent/ga/monitor.py
```diff
@@ -36,58 +36,148 @@
...
-        return telemetry_dict
-    else:
-        return None
-
-
 def get_monitor_handler():
     return MonitorHandler()
 
 
+class PollResourceUsageOperation(PeriodicOperation):
```
narrieta, line 43: moved the periodic tasks into their own classes, which can be used/tested independently of the monitor thread

azurelinuxagent/ga/monitor.py
```diff
@@ -36,58 +36,148 @@
...
         self.imds_client = None
 
         self.event_thread = None
         self._periodic_operations = [
-            PeriodicOperation("reset_loggers", self.reset_loggers, self.RESET_LOGGERS_PERIOD),
+            ResetPeriodicLogMessagesOperation(),
             PeriodicOperation("collect_and_send_events", self.collect_and_send_events, self.EVENT_COLLECTION_PERIOD),
-            PeriodicOperation("send_telemetry_heartbeat", self.send_telemetry_heartbeat, self.TELEMETRY_HEARTBEAT_PERIOD),
-            PeriodicOperation("poll_telemetry_metrics usage", self.poll_telemetry_metrics, self.CGROUP_TELEMETRY_POLLING_PERIOD),
+            ReportNetworkErrorsOperation(),
```
narrieta, line 172: I left these 3 operations in the monitor class because I still need to unravel their dependency on the protocol

tests/common/test_cgroupconfigurator.py
```diff
@@ -130,6 +127,21 @@ def raise_exception(*_):
                     message = args[0]
                     self.assertIn("A TEST EXCEPTION", message)
 
+    def test_get_processes_in_agent_cgroup_should_return_the_processes_within_the_agent_cgroup(self):
```
pgombar, line 130: How is this test different from `test_get_processes_in_cgroup_should_return_the_processes_within_the_cgroup` in `test_cgroupapi.py`?

azurelinuxagent/common/cgroupapi.py
```diff
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
narrieta: yup

tests/common/test_cgroupconfigurator.py
narrieta: code-wise they are the same, but test 2 different components (configurator vs api)

azurelinuxagent/common/cgroup.py
```diff
@@ -23,6 +25,21 @@
 from azurelinuxagent.common.osutil import get_osutil
 from azurelinuxagent.common.utils import fileutil
 
+
+MetricValue = namedtuple('Metric', ['category', 'counter', 'instance', 'value'])
+
+
+class MetricsCategory(object):
+    MEMORY_CATEGORY = "Memory"
+    PROCESS_CATEGORY = "Process"
```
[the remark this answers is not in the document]
narrieta: I moved that code from elsewhere, but CPU does make more sense. We are not using the existing metric in kusto so I'll go ahead with the rename.

PR #1910, opened 2020-06-17
Do not use --unit with systemd-cgls

narrieta:
systemd-cgls doesn't support --unit on ubuntu 16; using the cgroup path instead.

also, improved error handling and reporting.

### DIFF
--- a/azurelinuxagent/common/cgroupconfigurator.py
+++ b/azurelinuxagent/common/cgroupconfigurator.py
@@ -41,6 +41,8 @@ class CGroupConfigurator(object):
             self._cgroups_supported = False
             self._cgroups_enabled = False
             self._cgroups_api = None
+            self._agent_cpu_cgroup_path = None
+            self._agent_memory_cgroup_path = None
             self._get_processes_in_agent_cgroup_last_error = None
             self._get_processes_in_agent_cgroup_error_count = 0
 
@@ -137,16 +139,16 @@ class CGroupConfigurator(object):
                 if cpu_controller_root is None or cpu_cgroup_relative_path is None:
                     logger.info("Will not track CPU for the agent's cgroup")
                 else:
-                    cpu_cgroup_path = os.path.join(cpu_controller_root, cpu_cgroup_relative_path)
-                    CGroupsTelemetry.track_cgroup(CpuCgroup(agent_unit_name, cpu_cgroup_path))
+                    self._agent_cpu_cgroup_path = os.path.join(cpu_controller_root, cpu_cgroup_relative_path)
+                    CGroupsTelemetry.track_cgroup(CpuCgroup(agent_unit_name, self._agent_cpu_cgroup_path))
 
                 if memory_controller_root is None or memory_cgroup_relative_path is None:
                     logger.info("Will not track memory for the agent's cgroup")
                 else:
-                    memory_cgroup_path = os.path.join(memory_controller_root, memory_cgroup_relative_path)
-                    CGroupsTelemetry.track_cgroup(MemoryCgroup(agent_unit_name, memory_cgroup_path))
+                    self._agent_memory_cgroup_path = os.path.join(memory_controller_root, memory_cgroup_relative_path)
+                    CGroupsTelemetry.track_cgroup(MemoryCgroup(agent_unit_name, self._agent_memory_cgroup_path))
 
-                log_cgroup_info("Agent cgroups: CPU: {0} -- MEMORY: {1}", cpu_cgroup_path, memory_cgroup_path)
+                log_cgroup_info("Agent cgroups: CPU: {0} -- MEMORY: {1}", self._agent_cpu_cgroup_path, self._agent_memory_cgroup_path)
 
             except Exception as e:
                 message = "Error initializing cgroups: {0}".format(ustr(e))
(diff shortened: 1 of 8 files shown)


azurelinuxagent/common/utils/shellutil.py
```diff
@@ -120,14 +120,14 @@ class CommandError(Exception):
     Exception raised by run_command when the command returns an error
     """
     @staticmethod
-    def _get_message(command, returncode):
+    def _get_message(command, return_code, stderr):
         command_name = command[0] if isinstance(command, list) and len(command) > 0 else command
-        return "'{0}' failed: {1}".format(command_name, returncode)
+        return "'{0}' failed: {1} ({2})".format(command_name, return_code, stderr.rstrip())
```
narrieta, line 125: debugging failures with just the error code in the exception message can be hard; added stderr

azurelinuxagent/ga/monitor.py
```diff
@@ -58,25 +58,30 @@ def _operation_impl(self):
...
-            if len(unexpected_processes) > 0 and self._error_count < 5:
-                unexpected_processes.sort()
-                error = ustr(unexpected_processes)
-                if error != self._last_error:
-                    self._error_count += 1
-                    self._last_error = error
-                    message = "The agent's cgroup includes unexpected processes: {0}".format(error)
-                    logger.info(message)
-                    add_event(op=WALAEventOperation.CGroupsDebug, message=message)
+        processes_check_error = None
```
narrieta, line 61: any exception in the code to check processes should not prevent us from reporting metrics

azurelinuxagent/common/utils/shellutil.py
larohra: Can there be a case where stderr is None? If it is `stderr.rstrip()` would throw

azurelinuxagent/ga/monitor.py
```diff
@@ -58,25 +58,30 @@ def _operation_impl(self):
...
+                        unexpected_processes.append(command_line)
+
+                if len(unexpected_processes) > 0:
+                    unexpected_processes.sort()
+                    processes_check_error = ustr(unexpected_processes)
+        except Exception as e:
+            processes_check_error = ustr(e)
+
+        # Report a small sample of errors
+        if processes_check_error != self._last_error and self._error_count < 5:
```
larohra, line 79: I missed this in the previous PR, but I noticed we're not resetting the error count ever. I think we should reset it once  a day or something to also get newer errors that might occur

azurelinuxagent/common/utils/shellutil.py
narrieta: no, it'd be an empty string

azurelinuxagent/ga/monitor.py
narrieta: it was intentional; there is no need for that, i just want a sample of possible errors

azurelinuxagent/common/utils/shellutil.py
larohra: Awesome (Y)

tests/utils/test_shell_util.py
```diff
@@ -140,7 +139,8 @@ def test_run_command_should_raise_an_exception_when_the_command_fails(self):
             shellutil.run_command(command)
 
         exception = context_manager.exception
-        self.assertEquals(str(exception), "'ls' failed: 2")
+        self.assertIn("'ls' failed: 2", str(exception))
```
narrieta, line 142: btw - python 2.6 doesn't have an assert to match a regex, I need to add that to the test utilities.

i'll do that on a separate PR, in the meanwhile I split the check on 2 asserts

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -223,8 +225,9 @@ def get_processes_in_agent_cgroup(self):
             The return value can be None if cgroups are not enabled or if an error occurs during the operation.
             """
             def __impl():
-                agent_unit = self._cgroups_api.get_agent_unit_name()
-                return self._cgroups_api.get_processes_in_cgroup(agent_unit)
+                if self._agent_cpu_cgroup_path is None:
```
pgombar, line 228: Wouldn't it be better to use the memory cgroup here since we know CPU is not mounted by default in some distros, whereas memory is?

azurelinuxagent/ga/monitor.py
```diff
@@ -58,25 +58,30 @@ def _operation_impl(self):
...
+            self._error_count += 1
+            self._last_error = processes_check_error
+            message = "The agent's cgroup includes unexpected processes: {0}".format(processes_check_error)
```
pgombar, line 82: The error message now doesn't match the intention when `process_check_error` just contains the stack trace of an exception that occurred when we were trying to check processes in the agent cgroup. I know you are only using this event to gather diagnostics, so it's up to you if you want to make it clearer.

azurelinuxagent/common/cgroupconfigurator.py
narrieta: No, it is CPU that we are interested in.

azurelinuxagent/ga/monitor.py
narrieta: thanks; fixed

azurelinuxagent/common/cgroupconfigurator.py
pgombar: Why CPU specifically? Aren't we only using the cgroup path to get the PIDs? They are also stored in the memory cgroup.

narrieta: We want to enforce CPU, so it is the CPU cgroup that we need to check.

PR #1911, opened 2020-06-18
Fix generic logs event code

larohra:
## Description
This PR contains - 
- Adding timestamp to Context2 for GAGenericLogs table
- adding a force flag to add_log_event to force send log event in certain cases
- Not adding parameters to events that do not need it
- Refactoring: Created Enums for TelemetryEventSchemas and reformatted code 

Issue # 

---

### PR information


azurelinuxagent/common/event.py
```diff
@@ -305,23 +306,23 @@ def __init__(self):
...
+        self._common_parameters.append(TelemetryEventParam(CommonTelemetryEventSchema.ExecutionMode, AGENT_EXECUTION_MODE))
+        self._common_parameters.append(TelemetryEventParam(CommonTelemetryEventSchema.RAM, int(EventLogger._get_ram(osutil))))
+        self._common_parameters.append(TelemetryEventParam(CommonTelemetryEventSchema.Processors, int(EventLogger._get_processors(osutil))))
 
         # Parameters from goal state
-        self._common_parameters.append(TelemetryEventParam("VMName", "VMName_UNINITIALIZED"))
-        self._common_parameters.append(TelemetryEventParam("TenantName", "TenantName_UNINITIALIZED"))
-        self._common_parameters.append(TelemetryEventParam("RoleName", "RoleName_UNINITIALIZED"))
-        self._common_parameters.append(TelemetryEventParam("RoleInstanceName", "RoleInstanceName_UNINITIALIZED"))
+        # self._common_parameters.append(TelemetryEventParam("VMName", "VMName_UNINITIALIZED"))
```
narrieta, line 315: why commented out?

azurelinuxagent/common/event.py
```diff
@@ -533,15 +534,20 @@ def _add_common_event_parameters(self, event, event_timestamp):
...
+                         TelemetryEventParam(CommonTelemetryEventSchema.EventTid, threading.current_thread().ident),
+                         TelemetryEventParam(CommonTelemetryEventSchema.EventPid, os.getpid()),
+                         TelemetryEventParam(CommonTelemetryEventSchema.TaskName, threading.current_thread().getName()),
+                         TelemetryEventParam(CommonTelemetryEventSchema.KeywordName, '')]
+
+        if event.eventId in (None, TELEMETRY_EVENT_EVENT_ID) and event.providerId in (None, TELEMETRY_EVENT_PROVIDER_ID):
+            # Currently only the GuestAgentExtensionEvents has these columns, the other tables dont have them so skipping
+            # this data in those tables.
+            # By default, if no eventId/providerId is specified, add this data too.
+            common_params.extend([TelemetryEventParam(GuestAgentExtensionEventsSchema.ExtensionType, event.file_type),
```
narrieta, line 549: From our conversation, these 2 should not be None and it does not really matter whether those columns are added either way (None and not None), so removing the explicit check for None makes the condition a little clearer.

azurelinuxagent/common/event.py
larohra: Sorry forgot to remove, thanks for catching it!

azurelinuxagent/common/event.py
larohra: Done

azurelinuxagent/common/event.py
narrieta: Thanks for the update; forgot to mention there was another instance:
```

$ find . -name '*.py' -exec grep -l '#.*"VMName"' {} \;
./azurelinuxagent/common/event.py
$ find . -name '*.py' -exec grep -n '#.*"VMName"' {} \;
357:            # parameters["VMName"].value = vminfo.vmName

```

tests/common/test_event.py
```diff
@@ -50,31 +53,43 @@ def setUp(self):
...
         }
 
+        self.expected_extension_events_params = {
+            GuestAgentExtensionEventsSchema.IsInternal: False,
+            GuestAgentExtensionEventsSchema.ExtensionType: ""
+        }
+
+    @staticmethod
+    def _is_guest_extension_event(event):
+        return event.eventId in (None, TELEMETRY_EVENT_EVENT_ID) and \
```
pgombar, line 86: Why is None a valid option here?

azurelinuxagent/common/event.py
```diff
@@ -446,11 +445,11 @@ def add_event(self, name, op=WALAEventOperation.Unknown, is_success=True, durati
...
-        event.parameters.append(TelemetryEventParam('EventName', WALAEventOperation.Log))
-        event.parameters.append(TelemetryEventParam('CapabilityUsed', logger.LogLevel.STRINGS[level]))
-        event.parameters.append(TelemetryEventParam('Context1', self._clean_up_message(message)))
-        event.parameters.append(TelemetryEventParam('Context2', ''))
-        event.parameters.append(TelemetryEventParam('Context3', ''))
+        event.parameters.append(TelemetryEventParam(GuestAgentGenericLogsSchema.EventName, WALAEventOperation.Log))
+        event.parameters.append(TelemetryEventParam(GuestAgentGenericLogsSchema.CapabilityUsed, logger.LogLevel.STRINGS[level]))
+        event.parameters.append(TelemetryEventParam(GuestAgentGenericLogsSchema.Context1, self._clean_up_message(message)))
+        event.parameters.append(TelemetryEventParam(GuestAgentGenericLogsSchema.Context2, datetime.utcnow().strftime(logger.Logger.LogTimeFormatInUTC)))
+        event.parameters.append(TelemetryEventParam(GuestAgentGenericLogsSchema.Context3, ''))
```
pgombar, line 452: Would you want to use this field for the timestamp of when the event is sent? Asking since Context2 now contains the timestamp when the event is created.

azurelinuxagent/common/telemetryevent.py
```diff
@@ -20,6 +20,57 @@
 from azurelinuxagent.common.datacontract import DataContract, DataContractList
 from azurelinuxagent.common.version import AGENT_NAME
 
+class CommonTelemetryEventSchema(object):
+
+    # Common schema keys for both tables
```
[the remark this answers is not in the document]
larohra: This is the common schemas for all the tables that we have in Kusto (GAExtensionEvents, GAGenericLogs, GAPerfMetrics). Do you want me to add a comment for these 3 tables? I'm not sure if that's required mainly because we might add/remove tables later on so the comment would get outdated. We can discuss this offline if you want.

tests/common/test_event.py
larohra: A lot of the tests were using None value by default so I just kept that behavior for tests. Earlier we used to always add those 2 columns for all tables so this check was not needed before. I can check if removing this check breaks a lot of tests. If it does I would rather keep the None check here rather than fixing the tests as they dont add any logical value to our testing.

azurelinuxagent/common/event.py
larohra: Well logically it should always have the timestamp of when the event was created right? The main reason why I added timestamp here is to keep it in sync with the current Windows agent behavior. Also it would be easier for publishers if we have consistency in our schema to make it easy for them to setup their queries.

tests/common/test_event.py
narrieta: I think this comment is similar to the other one I posted on the check against None... let's do try to remove it if not needed and not keep it there just for the sake of the tests.

narrieta: (if needed for functionality of the agent of course it can stay there)

larohra: Done, thankfully wasnt that big a pain that I thought it would be

azurelinuxagent/common/telemetryevent.py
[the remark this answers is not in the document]
larohra: nvm added table names to the comment.

azurelinuxagent/common/event.py
larohra: Done

azurelinuxagent/common/event.py
pgombar: My comment was more about potentially adding the timestamp of sending the event to Context3 and if there were plans for that (does Windows have it?). I totally agree the timestamp of event creation is important.

larohra: Ohh sorry misunderstood that. Windows uses Context3 as the Name equivalent of who created the event. 

Eg: Context3
Microsoft-WindowsAzure-Diagnostics/GuestAgent
Microsoft-WindowsAzure-Diagnostics/Heartbeat
Microsoft-WindowsAzure-Diagnostics/GuestAgent

We can do something similar. We can maybe repurpose the OpCode name for the time that we send the event for GenericLogs table maybe? But that's outside the scope of this PR

pgombar: Gotcha.

Issue #1890, opened 2020-05-19
[BUG] Can't run waagent -deprovision


Bug-Reaper:
**Describe the bug: A clear and concise description of what the bug is.**

Note: Please add some context which would help us understand the problem better
1. Install the AUR for WALinuxAgent https://aur.archlinux.org/packages/walinuxagent/
2. Attempt to run  `waalinux -deprovision`
3. ???
![Screenshot_2020-05-19_05-09-46](https://user-images.githubusercontent.com/31227781/82325292-d9b34900-998f-11ea-9e6b-f94060b2693a.png)

**Distro and WALinuxAgent details:**
 - Arch Linux (Latest)
 - WALinuxAgent version (2.2.46)
- Python 3.8.2

Bug-Reaper:
Can I just manually delete all this stuff?
```

All SSH host keys (if Provisioning.RegenerateSshHostKeyPair is 'y' in the configuration file)

Nameserver configuration in /etc/resolv.conf

Root password from /etc/shadow (if Provisioning.DeleteRootPassword is 'y' in the configuration file)

Cached DHCP client leases

Resets host name to localhost.localdomain
```

Bug-Reaper:
Alright, I got this figured out. Not sure if this is specific to Arch Linux, I'd imagine it applies to any *Nix system using the latest dhcpcd.

What you need to do is (as root):
```
systemctl stop dhcpcd
umount /var/lib/dhcpcd/run/systemd/journal
umount /var/lib/dhcpcd/run/udev
umount /var/lib/dhcpcd/sys
umount /var/lib/dhcpcd/dev
umount /var/lib/dhcpcd/proc
```
Apparently those are now mounted by dhcpcd ONTOP of where they originally were. As in, unmounting them doesn't remove files from those locations. Very strange.

I'd be happy to make a PR which automates this process as part of `agent.py`. None of those commands are going to break the `waagent -deprovison` process regardless of what *nix OS you're on and we could obviously try to detect that we're on a system that requires this first.

Cheers

pgombar:
Hi @Bug-Reaper thank you for your analysis. Feel free to submit a PR to improve this and reference this issue. cc @trstringer

Bug-Reaper:
@pgombar @trstringer I'll try to get a PR up for this and test it out over the next couple of days. Thanks :)

Bug-Reaper:
The PR is up, let me know what you think @trstringer :+1:

PR #1920, opened 2020-06-26
Fix DHCPCD --deprovision bug

Bug-Reaper:
## Description

Issue #1890 
https://github.com/Azure/WALinuxAgent/issues/1890

Fixes bug where `waagent --deprovision` would error/crash when trying to delete cached DHCPCD leases.

This error occurs because DHCPCD mounts files onto sub-directories of `/var/lib/dhcpcd`. These mounted DHCPCD files cannot be deleted whilst mounted (even with the all mighty power of `sudo rm`). This causes `waagent --deprovision` to crash when attempting to delete cached DHCPCD leases.

This PR adds the new method `umount(path)` to common/utils/fileutil.py. If the `path` argument does not pass os.path.ismount() then nothing happens. Otherwise the argument `path` is unmounted.

Right before we delete cached DHCPCD leases in the `del_dhcp_lease`, we call the method `fileutil.umount()` against paths of known DHCPCD mountpoints that cause this error. Because some mountpoints cannot be unmounted while DHCPCD is in use, we also first make sure that the DHCPCD service is not running. (pa/deprovision/default.py)

In doing so this PR address the crash/error and has been tested to work.

---

### PR information
- [ x ] The title of the PR is clear and informative.
- [ x ] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).
- [ x ] If applicable, the PR references the bug/issue that it fixes in the description.

- [ x ] I have read the [contribution guidelines](https://github.com/Azure/WALinuxAgent/blob/master/.github/CONTRIBUTING.md).

``` 
___      ___            __       ___      ___       __      
|"  |    |"  |          /""\     |"  \    /"  |     /""\     
||  |    ||  |         /    \     \   \  //   |    /    \    
|:  |    |:  |        /' /\  \    /\\  \/.    |   /' /\  \   
 \  |___  \  |___    //  __'  \  |: \.        |  //  __'  \  
( \_|:  \( \_|:  \  /   /  \\  \ |.  \    /:  | /   /  \\  \ 
 \_______)\_______)(___/    \___)|___|\__/|___|(___/    \___)                                                                                                                          
```

### DIFF
--- a/azurelinuxagent/common/utils/fileutil.py
+++ b/azurelinuxagent/common/utils/fileutil.py
@@ -123,6 +123,14 @@ def chmod(path, mode):
         os.chmod(path, mode)
 
 
+def umount(*args):
+    for paths in args:
+        # find all possible mounts
+            for path in glob.glob(paths):
+                if os.path.ismount(path):
+                    os.unmount(path)
+
+
 def rm_files(*args):
     for paths in args:
         # find all possible file paths
--- a/azurelinuxagent/common/utils/fileutil.py
+++ b/azurelinuxagent/common/utils/fileutil.py
@@ -30,6 +30,7 @@ import shutil
 
 import azurelinuxagent.common.logger as logger
 import azurelinuxagent.common.utils.textutil as textutil
+import azurelinuxagent.common.utils.shellutil as shellutil
 
 from azurelinuxagent.common.future import ustr
 
@@ -128,7 +129,8 @@ def umount(*args):
         # find all possible mounts
             for path in glob.glob(paths):
                 if os.path.ismount(path):
-                    os.unmount(path)
+                    command=["umount",path]
+                    shellutil.run_command(command)
 
 
 def rm_files(*args):
--- a/azurelinuxagent/pa/deprovision/default.py
+++ b/azurelinuxagent/pa/deprovision/default.py
@@ -74,7 +74,7 @@ class DeprovisionHandler(object):
         username = ovfenv.username
         warnings.append(("WARNING! {0} account and entire home directory "
                          "will be deleted.").format(username))
-        actions.append(DeprovisionAction(self.osutil.del_account, 
+        actions.append(DeprovisionAction(self.osutil.del_account,
                                          [username]))
 
 
@@ -111,6 +111,11 @@ class DeprovisionHandler(object):
 
     def del_dhcp_lease(self, warnings, actions):
         warnings.append("WARNING! Cached DHCP leases will be deleted.")
+
+        # For Arch based systems. Files mounted by DHCPCD can't be deleted whilst mounted, not even by root.
+        dirs_to_umount = ["/var/lib/dhcpcd/run/systemd/journal", "/var/lib/dhcpcd/run/udev", "/var/lib/dhcpcd/sys", "/var/lib/dhcpcd/dev", "/var/lib/dhcpcd/proc"]
+        actions.append(DeprovisionAction(fileutil.umount, dirs_to_umount))
+
         dirs_to_del = ["/var/lib/dhclient", "/var/lib/dhcpcd", "/var/lib/dhcp"]
         actions.append(DeprovisionAction(fileutil.rm_dirs, dirs_to_del))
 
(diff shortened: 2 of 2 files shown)


azurelinuxagent/common/utils/fileutil.py
```diff
@@ -123,6 +124,15 @@ def chmod(path, mode):
         os.chmod(path, mode)
 
 
+def umount(*args):
+    for paths in args:
+        # find all possible mounts
+            for path in glob.glob(paths):
```
trstringer, line 130: Unnecessary indentation here after the comment.

azurelinuxagent/common/utils/fileutil.py
```diff
@@ -123,6 +124,15 @@ def chmod(path, mode):
...
+                if os.path.ismount(path):
+                    command=["umount",path]
+                    shellutil.run_command(command)
```
trstringer, line 133: Is the usage of the `command` var necessary here? Looks like a single-use variable, consider combining this into a single line.

azurelinuxagent/pa/deprovision/default.py
```diff
@@ -111,6 +111,14 @@ def del_resolv(self, warnings, actions):
 
     def del_dhcp_lease(self, warnings, actions):
         warnings.append("WARNING! Cached DHCP leases will be deleted.")
+
+        # For Arch based systems. Kill DHCPCD service. Needed to clear cached leases.
+        shellutil.run("systemctl stop dhcpcd", chk_err=False)
+
+        # For Arch based systems. Files mounted by DHCPCD can't be deleted whilst mounted, not even by root.
+        dirs_to_umount = ["/var/lib/dhcpcd/run/systemd/journal", "/var/lib/dhcpcd/run/udev", "/var/lib/dhcpcd/sys", "/var/lib/dhcpcd/dev", "/var/lib/dhcpcd/proc"]
+        actions.append(DeprovisionAction(fileutil.umount, dirs_to_umount))
```
trstringer, line 120: If this is Arch-specific logic, then it should go into the `ArchDeprovisionHandler`.

trstringer (changes requested):
Along with the individual suggestions, how are you testing this solution? What are your steps? Thanks for the contribution!

azurelinuxagent/common/utils/fileutil.py
Bug-Reaper: I agree, we can shorten this to one line.

azurelinuxagent/pa/deprovision/default.py
Bug-Reaper: So perhaps I should rephrase my comment here. While it's true this is for Arch systems, it's also needed for any *Nix distro that's running the latest version of `dhcpcd`. A quick check on wiki shows that a large number  of distros use `systemd/dhcpcd` by default or as an optional configuration:
Screenshot_2020-06-26_17-18-33


The systems that can be configured with `systemd/dhcpcd` include: Arch Linux, CentOS, CoreOS, Fedora, Mageia, Manjaro Linux, openSUSE, SUSE Linux Enterprise Server, Red Hat Enterprise Linux, Solus, Linux Mint, Ubuntu and Debian.


Among those, many use `systemd/dhcpcd` by default and with none/few alternatives.


I believe this is the correct place to run this, but I can update my comment to make it clear this is a DHCPCD specific issue and not an Arch one.

azurelinuxagent/common/utils/fileutil.py
Bug-Reaper: Yep, the indentation is unnecessary here. I'll amend that.

Bug-Reaper:
Initial testing is covered in the issue.

Testing methodology for this PR has gone something like this:
(On a virtual system that's using the latest `systemd/dhcpcd`)
1. Install the latest waagent and run `waagent --deprovision` to confirm crash is still present on the latest version.
2. Clone my fork to my home directory.
3. Copy my forked version of the `azurelinuxagent` libs to replace the version installed by `waagent`.
4. Run `waagent --deprovision` again and confirm it completes successfully without crashing.
5. Shutdown my VM.
6. Upload the corresponding VHD file into Azure storage.
7. Create an Azure disk from that VHD file and deploy it as an Azure VM.
8. Confirm that VM is running normally (by SSHing in) and also that valid information is being reported about that VM in the Azure web portal.

PR #1926, opened 2020-06-29
Fix return value of start_extension_command

narrieta:
#1899 changed the return value of start_extension_command, but did not update the retry path.

The test for retry did not check the return value, so the issue was undetected.

Updated code and existing test. Added new test for normal path (no retry)


narrieta:
needs to go against release branch

PR #1935, opened 2020-07-06
Do not retrieve users in each goal state

narrieta:
The remote access handler retrieves the list of users on the VM when the goal state doesn't include any jit users (this is done to remove the jit accounts).

Retrieving the users can be extremely slow in some environments and currently it is done an each goal state.

I introduced a flag so that now we retrieve the users only on the first goal state after service start or when the agent has actually created jit accounts.


tests/common/osutil/mock_osutil.py
```diff
@@ -1,53 +0,0 @@
-# Copyright Microsoft Corporation
```
narrieta, line 1: this mock is specialized for the remote access tests so i moved it there

tests/ga/test_remoteaccess_handler.py
```diff
@@ -58,478 +82,431 @@ def setUp(self):
         # Since ProtocolUtil is a singleton per thread, we need to clear it to ensure that the test cases do not
         # reuse a previous state
         clear_singleton_instances(ProtocolUtil)
-        del info_messages[:]
-        del error_messages[:]
         for data in TestRemoteAccessHandler.eventing_data:
             del data
 
     # add_user tests
-    @patch('azurelinuxagent.common.logger.Logger.info', side_effect=log_info)
```
narrieta, line 67: No point in testing the log statements, removed these checks from the tests

azurelinuxagent/ga/remoteaccess.py
```diff
@@ -48,108 +47,107 @@ def get_remote_access_handler(protocol):
...
             for user in existing_jit_users:
                 try:
                     if user not in goal_state_users:
                         # user explicitly removed
-                        logger.info("User {0} removed from remote_access".format(user))
-                        self.remove_user(user)
-                except RemoteAccessError as rae:
-                    self.err_message = self.err_message + "Error removing user {0}. Exception: {1}"\
-                        .format(user, ustr(rae))
+                        self._remove_user(user)
```
pgombar, line 105: Don't we want to keep the log when we are removing the user?

pgombar: You are logging it when removing below on L122.

narrieta: _remove_user does the logging; 122 is a left-over, will remove it

azurelinuxagent/ga/remoteaccess.py
```diff
@@ -48,108 +47,107 @@ def get_remote_access_handler(protocol):
...
-                error += "and error cleaning up {0}".format(str(rae))
-            raise RemoteAccessError("Error adding user {0} cleanup successful".format(username), ustr(e))
-
-    def handle_failed_create(self, username):
-        try:
-            self.delete_user(username)
-        except Exception as e:
-            raise RemoteAccessError("Failed to clean up after account creation for {0}.".format(username), e)
+            logger.info("Adding remote access user '{0}' with expiration date {1}", username, expiration_date)
+            self._os_util.useradd(username, expiration_date, REMOTE_ACCESS_ACCOUNT_COMMENT)
```
kevinclark19a, line 137: Wouldn't it be better to move this line to outside of the try catch? We don't actually have any handling logic for Exceptions that occur before the user_added flag is set (line 138), so any Exception raised by this line has to be caught by a calling function's try: except anyway. This way we wouldn't have to keep track of the user_added flag at all.

azurelinuxagent/ga/remoteaccess.py
```diff
@@ -48,108 +47,107 @@ def get_remote_access_handler(protocol):
...
-            raise RemoteAccessError("Failed to delete user {0}".format(username), e)
+            logger.info("Adding remote access user '{0}' to sudoers", username)
+            prv_key = os.path.join(conf.get_lib_dir(), TRANSPORT_PRIVATE_CERT)
+            pwd = self._cryptUtil.decrypt_secret(encrypted_password, prv_key)
+            self._os_util.chpasswd(username, pwd, conf.get_password_cryptid(), conf.get_password_crypt_salt_len())
+            self._os_util.conf_sudoer(username)
+        except Exception:
+            if user_added:
+                self._remove_user(username)
+            raise
```
kevinclark19a, line 148: This except: block (lines 145-148) looks like it would be more natural as a finally: block.

azurelinuxagent/ga/remoteaccess.py
narrieta: I wrote the code like that initially but then I thought it looked more natural if the exception handling covers the entire function and we have a flag for when to clean up.

tests/ga/test_remoteaccess_handler.py
```diff
@@ -14,21 +14,53 @@
...
+        self.all_users = {}
+        self.sudo_users = set()
+        self.jit_enabled = True
+
+    def useradd(self, username, expiration=None, comment=None):
+        if username == "":
+            raise Exception("test exception for bad username")
+        if username in self.all_users:
+            raise Exception("test exception, user already exists")
+        self.all_users[username] = (username, None, None, None, comment, None, None, expiration)
```
kevinclark19a, line 41: Would a named tuple be better here (line 41)? I'm having trouble understanding what all the "None"s are.

narrieta: yes, probably. the existing tests need many improvements. i did several improvements as part of this pr and i may do more if i touch this code in the future.

narrieta: (btw i moved this mock from the common utils directory to this specific test, since its functionality is limited to the remote access tests)

tests/ga/test_remoteaccess_handler.py
```diff
@@ -58,478 +82,431 @@ def setUp(self):
...
-    @patch('azurelinuxagent.common.logger.Logger.info', side_effect=log_info)
-    @patch('azurelinuxagent.common.logger.Logger.error', side_effect=log_error)
+    def test_add_user_bad_creation_data(self, *_):
+        with patch("azurelinuxagent.ga.remoteaccess.get_osutil", return_value=MockOSUtil()):
+            rah = RemoteAccessHandler(Mock())
+            tstpassword = "]aPPEv}uNg1FPnl?"
+            tstuser = ""
+            expiration = datetime.utcnow() + timedelta(days=1)
+            pwd = tstpassword
+            error = "test exception for bad username".format(tstuser)
```
kevinclark19a, line 113: This format (line 113) doesn't seem to do anything. Is the intent here to include the test username in the error string? or is the format call extraneous?

narrieta: thanks; i forgot to remove the call to format -- there may be some other instances below; i'll check all of them

narrieta: yes, i did find a few more. all fixed now

tests/ga/test_remoteaccess_handler.py
```diff
@@ -58,478 +82,431 @@ def setUp(self):
...
-    @patch('azurelinuxagent.common.logger.Logger.info', side_effect=log_info)
-    @patch('azurelinuxagent.common.logger.Logger.error', side_effect=log_error)
+    def test_add_user(self, *_):
+        with patch("azurelinuxagent.ga.remoteaccess.get_osutil", return_value=MockOSUtil()):
+            rah = RemoteAccessHandler(Mock())
+            tstpassword = "]aPPEv}uNg1FPnl?"
+            tstuser = "foobar"
+            expiration_date = datetime.utcnow() + timedelta(days=1)
+            pwd = tstpassword
+            rah._add_user(tstuser, pwd, expiration_date)
```
kevinclark19a, line 97: Probably (definitely) out of the scope of this PR, but this should ideally call "run()" instead of a private method (lines 97, 114, 126, etc...). Is it worth capturing this as a future task, or should we just make the changes if these tests break?

narrieta: Yes, most of those tests need rewriting and unfortunately that is very common across most of the code base so opening tasks could be unmanageable. I tend to do improvements as I touch related code, many times I span the improvements over several iterations of the code.

narrieta: Of course, if a test breaks while changing the prod code we need to address that immediately. When that happens to you, try to evaluate whether the broken test adds any value, on several occasions I've found fixing broken tests is not worth the effort (they may be redundant, test private implementation, etc)

azurelinuxagent/ga/remoteaccess.py
```diff
@@ -48,108 +47,106 @@ def get_remote_access_handler(protocol):
...
+            # There are no JIT users in the goal state; that may mean that they were removed or that they
+            # were never added. Enumerating the users on the current vm can be very slow and this path is hit
+            # on each goal state; we use self._check_existing_jit_users to avoid enumerating the users
+            # every single time.
+            if self._check_existing_jit_users:
+                logger.info("Looking for existing remote access users.")
+
+                self._check_existing_jit_users = False
+
+                existing_jit_users = self._get_existing_jit_users()
```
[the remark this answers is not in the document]
narrieta: good point; i'll do this with try/finally

PR #1936, opened 2020-07-07
Disable only when enabled for version update

kevinclark19a:
## Description

Issue # 

PR to approach feature parity with Windows Guest Agent; on extension version update, agent now only invokes DisableCommand if the updating extension is already in an Enabled state. Two tests were added to verify this change, and the existing TestExtensionUpdateOnFailure tests were rewritten where the change broke them, exposing some issues.

The bulk of the changes in this PR are to facilitate the creation of new tests (including those mentioned above). The newest standard in the code base for testing (as of this PR) is to utilize the mock_wire_protocol function to generate a WireProtocol object that pulls test data from the */tests/data directory, and then to mock subprocess.Popen to modify default extension behavior. The ExtensionActor class and friends (within tests/ga/extension_actor.py) provide an interface to utilize mock_wire_protocol that gets its data not from external files, but dynamically from the test itself. 

An ExtensionActor can be built with any set of functions (including lambdas) that match the required function signature. Each action (e.g. InstallCommand, EnableCommand, etc.) are then wrapped as Mocks that can be queried as normal. HTTP PUT /status API calls are also tracked, and can be accessed at ExtensionActor.statusBlobs.

---

### PR information


tests/ga/test_extension.py
```diff
@@ -2535,304 +2536,285 @@ def test_ext_handler_with_systemd_cgroup_enabled(self, *args):
...
-            "reportHeartbeat": False,
-            "continueOnUpdateFailure": continue_on_update_failure
-        }
+            # Run the handler with only the first version to install it
+            exthandlers_handler.run()
+            
+            # Verify the extension was picked up correctly.
+            # NOTE: this won't work for tests that want install to fail.
+            firstVersionActor.get_command("installCommand").assert_called_once()
+            firstVersionActor.get_command("enableCommand").assert_called_once()
```
narrieta, line 2570: I'd throw rather than assert. This function seems to be a helper for test setup rather than verification of behavior of entity being tested.

kevinclark19a: I'm not sure I understand. What exactly would be thrown?

tests/ga/test_extension.py
```diff
@@ -2535,304 +2536,285 @@ def test_ext_handler_with_systemd_cgroup_enabled(self, *args):
...
-        with patch.object(CGroupConfigurator.get_instance(), "start_extension_command",
-                          side_effect=ExtensionError('disable Failed')) as patch_start_cmd:
-            with self.assertRaises(ExtensionError):
-                ExtHandlersHandler._update_extension_handler_and_return_if_failed(old_handler_i, new_handler_i)
+        TestExtensionUpdateOnFailure._do_upgrade_scenario(first_ext, second_ext)
 
-            args, kwargs = patch_start_cmd.call_args
+        first_ext.get_command("disableCommand").assert_called_once()
+        second_ext.get_command("installCommand").assert_called_once()
+        second_ext.get_command("enableCommand").assert_called_once()
```
narrieta, line 2588: Do we have a test for the entire update sequence? if we do, then this test may be redundant. If we don't, can we turn this test into one that verifies the full sequence of extension commands? (and also assert that the commands are called in the right order instead of just called once)

kevinclark19a: I do not believe there is a "nothing failed" test, at least in this class. It would be trivial to add checks for the entire sequence of commands, and I think I have an idea for guaranteeing ordering... I'll toy around with it and commit a change.

tests/ga/test_extension.py
narrieta: Sorry, i was trying to say asserts should be used to check the behavior you are testing, e.g " first_ext.get_command("disableCommand").assert_called_once()" in test_enabled_ext_should_be_disabled_at_ver_update

_do_upgrade_scenario is more of a setup method; it is preparing the extensions for the test to act on and do its checking. A setup failure should raise an exception rather than assert. In this case probably "raise Exception("The install command was not invoked when installing the extension")" or something similar would do it.

kevinclark19a: That makes more sense. Thanks. I'll make those changes

tests/ga/test_extension.py
kevinclark19a: I've definitely got the "guaranteed ordering" working. Do you think it would be worth it to integrate the functionality into extension_actor, enabled by default? I.e. is this something we'll want for all the tests?

narrieta: Not sure yet, let me make more progress in the review to get a feeling of what the actor is responsible for

tests/ga/extension_actor.py
```diff
@@ -0,0 +1,465 @@
...
+
+        # The keys below will be used to configure attributes for a mock; periods in such attributes are treated
+        # as attribute trees, which would break introspection.
+        formattedId = Formats.FormatExtensionDir(name, version).replace(".", "_")
+
+        self.delegate ={
+            "installCommand": dict(key="{0}_install".format(formattedId), action=installAction),
+            "uninstallCommand": dict(key="{0}_uninstall".format(formattedId), action=uninstallAction),
+            "updateCommand": dict(key="{0}_update".format(formattedId), action=updateAction),
+            "enableCommand": dict(key="{0}_enable".format(formattedId), action=Actions._wrap_with_write_status(enableAction, name, version)),
```
narrieta, line 281: All extension commands (not only enable) report status

tests/ga/extension_actor.py
```diff
@@ -0,0 +1,465 @@
...
+        """
+        for key in self.delegate.keys():
+            yield key
+    
+    def getKeyForCommand(self, cmd):
+        """
+        Returns the unique ID associated with a specific command name in this ActionSet.
+        cmd may be "installCommand", "enableCommand", etc.
+        """
+        return self.delegate.get(cmd, {}).get("key", None)
```
narrieta, line 308: is None an appropriate return value, or should this raise? seems like if somebody passes a command that does not exist that should be an error

tests/ga/extension_actor.py
kevinclark19a: The agent only seems to check the status file after enable is called. Is this behavior expected?

tests/ga/extension_actor.py
kevinclark19a: I think it might as well raise; nothing checks the returned value against None before attempting to access an attribute.

tests/ga/test_extension.py
```diff
@@ -2535,304 +2536,285 @@ def test_ext_handler_with_systemd_cgroup_enabled(self, *args):
...
-            args, kwargs = patch_start_cmd.call_args
+        first_ext = get_extension_actor(disableAction=unique_fail)
+        second_ext = get_extension_actor(version="1.1.0", continueOnUpdateFailure=True)
+
+        TestExtensionUpdateOnFailure._do_upgrade_scenario(first_ext, second_ext)
+
+        first_ext.get_command("disableCommand").assert_called_once()            
+        second_ext.get_command("updateCommand").assert_called_once()
+        first_ext.get_command("uninstallCommand").assert_called_once()
+        second_ext.get_command("installCommand").assert_called_once()
```
pgombar, line 2615: I assume that at this point in time, when we want to assert if the command was called and how many times, we're past the point of checking if it suceeded? If the command failed (and it shouldn't have), we would get notified beforehand, correct?

tests/ga/test_extension.py
```diff
@@ -2535,304 +2536,285 @@ def test_ext_handler_with_systemd_cgroup_enabled(self, *args):
...
+
+        first_ext = get_extension_actor(uninstallAction=unique_fail)
+        second_ext = get_extension_actor(version="1.1.0", continueOnUpdateFailure=True)
+
+        TestExtensionUpdateOnFailure._do_upgrade_scenario(first_ext, second_ext)
+
+        first_ext.get_command("disableCommand").assert_called_once()            
+        second_ext.get_command("updateCommand").assert_called_once()
+        first_ext.get_command("uninstallCommand").assert_called_once()
+        second_ext.get_command("installCommand").assert_called_once()
```
pgombar, line 2635: How are we validating that the order of these calls is correct?

tests/ga/test_extension.py
```diff
@@ -2535,304 +2536,285 @@ def test_ext_handler_with_systemd_cgroup_enabled(self, *args):
...
+        second_ext = get_extension_actor(version="1.1.0", continueOnUpdateFailure=False)
+
+        TestExtensionUpdateOnFailure._do_upgrade_scenario(first_ext, second_ext)
+            
+        first_ext.get_command("disableCommand").assert_called_once()
+        second_ext.get_command("updateCommand").assert_not_called()
+        first_ext.get_command("uninstallCommand").assert_not_called()
+        second_ext.get_command("installCommand").assert_not_called()
+
+        assert len(second_ext.statusBlobs) == 1
```
pgombar, line 2656: Please use self.assertEquals, self.assertTrue, etc instead of this syntax. The files we own use self.*, so let's be consistent.

tests/ga/extension_actor.py
```diff
@@ -0,0 +1,465 @@
...
+    uninstallAction=Actions.SucceedAction, updateAction=Actions.SucceedAction, enableAction=Actions.SucceedAction,
+    disableAction=Actions.SucceedAction):
+    """
+    Factory method for ExtensionActor class. Note that the provided name and version needs to match a plugin listed
+    in the xml doc returned by data_fetcher_base["manifest"]; otherwise, the agent won't be able to properly download
+    this extension.
+    """
+    
+    actionSet = ActionSet(installAction, uninstallAction, updateAction, enableAction, disableAction, name, version)
+    info = ExtensionInfo(name, version, continueOnUpdateFailure, updateMode)
```
narrieta, line 117: The action set and info are basically the extension manifest + name & version.

Consider merging them into a single entity with "ExtensionManifest" somewhere in its name, that way the test code is a little closer to the prod code. If you do that, please add reportHeartbeat to have a full manifest (no need to add rebootAfterInstall since it is not implemented). Also, get_extension_actor could take an instance of this entity as parameter (right now it takes quite a few parameters).

tests/ga/test_extension.py
```diff
@@ -2535,304 +2536,285 @@ def test_ext_handler_with_systemd_cgroup_enabled(self, *args):
...
+        first_ext.get_command("uninstallCommand").assert_called_once()
+        second_ext.get_command("installCommand").assert_called_once()
+
+        _, update_kwargs = second_ext.get_command("updateCommand").call_args
+        _, install_kwargs = second_ext.get_command("installCommand").call_args
+
+        second_extension_dir = "/".join([
+            conf.get_lib_dir(),
+            Formats.FormatExtensionDir(second_ext.extension_info.name, second_ext.extension_info.version)
+        ])
```
pgombar, line 2742: Would it be cleaner to move this utility to the extension actor's class? I saw you used it several times in the tests.

tests/protocol/mockwiredata.py
```diff
@@ -23,76 +23,17 @@
...
-        "ext_conf": "wire/ext_conf.xml",
-        "manifest": "wire/manifest.xml",
-        "ga_manifest": "wire/ga_manifest.xml",
-        "trans_prv": "wire/trans_prv",
-        "trans_cert": "wire/trans_cert",
-        "test_ext": "ext/sample_ext-1.3.0.zip",
-        "remote_access": None,
-        "in_vm_artifacts_profile": None
-}
+from tests.protocol.mockwiredata_filenames import *
```
pgombar, line 26: Avoid using `import *`, import only what you need.

tests/protocol/mockwiredata.py
```diff
@@ -289,44 +208,125 @@ def replace_xml_attribute_value(xml_document, element_name, attribute_name, attr
             raise Exception("Could not match attribute '{0}' of element '{1}'".format(attribute_name, element_name))
         return new_xml_document
 
+    @staticmethod
+    def replace_xml_attribute_value_on_matching_element(xml_document, element_name, identifying_attr_str, attribute_name, attribute_value):
+
+        # Two cases: the attribute we're using to identify the element might be before the attribute we're replacing,
+        # or it might be after. We need to handle both.
```
pgombar, line 215: Could you add a quick example in the comment section of this method? I'm not sure what you're referring to just based on reading this sentence.

tests/ga/extension_actor.py
narrieta: The current implementation checks for status at the end of the loop, but this should actually be async to extension execution. For example, extensions may be reporting progress of their install, and the agent should report that status to users.

Even in today's implementation, if install succeeds and reports status and then enable crashes before reporting status, the agent would report the status from install.

And lastly, the REST API allows users to disable extensions. The disable command should also report status and the agent should pick it up and report it up the chain.

tests/ga/extension_actor.py
```diff
@@ -0,0 +1,465 @@
...
+        A function factory for an action that executes the given command after writing 
+        a simple (successful) status file in the appropriate directory. 
+        
+        Useful in particular for wrapping "enableCommand"s, as the agent expects any 
+        extension to write such a status after enabling.
+        """
+        import azurelinuxagent.common.utils.fileutil as fileutil
+        import azurelinuxagent.common.conf as conf
+        
+        filedir = "/".join([
```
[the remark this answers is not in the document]
kevinclark19a: Would `os.path.join` be better?

tests/ga/test_extension.py
kevinclark19a: We are not validating ordering as of now. In another thread, I've mentioned that I can get that working, but wasn't sure if that was something we wanted to be part of the ExtensionActor class.

tests/ga/test_extension.py
kevinclark19a: `Formats.FormatExtensionDir` is actually from that file; it might be best to simply bring this functionality into that method.

tests/ga/extension_actor.py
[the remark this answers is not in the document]
pgombar: I believe so, I'm not entirely familiar with the difference, but let's be safe and use `os` methods for OS/filesystem-related use cases. If anything, for consistency with the rest of our code.

tests/ga/extension_actor.py
```diff
@@ -0,0 +1,465 @@
...
+        dataFetcher["test_ext"] = mockwiredata.generate_ext_fetcher_func(base_manifest)
+
+
+    def __init__(self, dataFetcherBase, actionSet, extensionInfo):
+        """
+        Creates a ExtensionActor emulating an extension specified by the provided
+        extensionInfo and actionSet. Copies the dataFetcherBase to serve as the dynamic
+        loader of info required by a WireProtocolData* object.
+        """
+        self.action_scope = Mock()
```
narrieta, line 403: seems like these members should be private

tests/ga/extension_actor.py
```diff
@@ -0,0 +1,465 @@
...
+        ExtensionActor._configure_action_scope(self.action_scope, actionSet)
+
+        self.data_fetcher = dataFetcherBase.copy()
+        ExtensionActor._configure_data_fetcher(self.data_fetcher, actionSet, extensionInfo)
+
+        self.get_command = lambda cmd: getattr(self.action_scope, actionSet.getKeyForCommand(cmd), None)
```
narrieta, line 409: just curious: why is this defined as a lambda instead of a fuction?

tests/ga/extension_actor.py
```diff
@@ -0,0 +1,465 @@
...
+        replaced subprocess.Popen, which is handled by the start() function.
+        """
+
+        original_popen = subprocess.Popen
+
+        def mock_popen(command, *args, **kwargs):
+            # The command can be specified as a list of args or one single string. Handle that here.
+            format_command = lambda cmd: " ".join(cmd) if isinstance(cmd, list) else cmd
+
+            script_name = format_command(command).split("/")[-1] # The command passed here (if it corresponds to one within
```
narrieta, line 455: this assumes a specific format for the command. it may be working now in those tests, but if we happen to add another call to popen somewhere in the call stack this will break

tests/ga/extension_actor.py
```diff
@@ -0,0 +1,465 @@
...
+    simplify the logic of adding the new actors, but is a bigger change. Ultimately, this feature isn't currently (i.e. as of
+    writing) needed.
+    """
+
+    new_actors = list(filter(lambda actor: actor not in protocol._actors, actors))  # list() applies the lambda before we edit protocol._actors
+    protocol._actors.extend(new_actors)
+
+    for actor in new_actors:
+        patched_popen = actor.patch_popen() # Like in get_protocol_and_handler, the patch_popen() and start() calls must
+        patched_popen.start()               # be interleaved.
```
narrieta, line 246: where do we stop the mock?

tests/ga/extension_actor.py
kevinclark19a: The ExtensionActor doesn't hold onto actionSet, so a closure within init is the only time it is in scope. Basically I made a decision not to hold on to actionSet (since we're computing action_scope), and getKeyFromCommand was the only thing required off of actionSet.

tests/ga/extension_actor.py
kevinclark19a: Good point. In fact, line 453 is a mitigation of one such situation (i.e. another call to popen in the call stack), wherein popen was passed a list instead of a string. However, I think there's only a finite set of possible args to popen (I think consisting of list and string).

Note, though, that a string (or list) that doesn't follow the specific format will simply fallback to the original popen (via the default on line 463). Is there another type(command) I need to account for?

tests/ga/extension_actor.py
kevinclark19a: Let me know if you just missed this comment, or if you saw it but it wasn't clear enough:

line 248: `# Keep track of the patches so the finally clause within get_protocol_and_handler can stop them.`
line 249: `protocol._patched_popens.append(patched_popen)`

and the finally clause referenced in the comment (line 211-214):
```
finally:
    for patched_popen in protocol._patched_popens:
        patched_popen.stop()
```

To answer your question, we're keeping track of all of the mock popen objects, and stopping them when the `with get_protocol_and_handler` goes out of scope.

tests/ga/extension_actor.py
```diff
@@ -0,0 +1,465 @@
...
+    Note that this function does not update xml for ext_conf or manifest, meaning that added extensions must already be present
+    within those files as provided by the firstActor's data source. This could probably be implemented with the
+    WireProtocolDataBase.replace_xml_element_value function, but it would probably make sense to first implement autogeneration
+    of those files within ExtensionActor.__init__, as we would then know exactly the xml elements in those files. This would
+    simplify the logic of adding the new actors, but is a bigger change. Ultimately, this feature isn't currently (i.e. as of
+    writing) needed.
+    """
+
+    new_actors = list(filter(lambda actor: actor not in protocol._actors, actors))  # list() applies the lambda before we edit protocol._actors
+    protocol._actors.extend(new_actors)
```
narrieta, line 242: attaching the actor to the protocol seemms odd, even if this test code. can we find a better place for them?

kevinclark19a: I'm actually running into this with the "guaranteed ordering of command invocations" we mentioned earlier. I'm leaning towards some sort of "manager" class that would basically take on the three base-namespace functions (`get_extension_actor`, `get_protocol_and_handler`, and `add_extension_actors`) into itself as methods. Thoughts?

tests/ga/extension_actor.py
narrieta: I see; the cleanup is on a different function. Can we improve this (maybe use a context manager) so  that initialization/cleanup are together?

tests/ga/extension_actor.py
kevinclark19a: Unfortunately, all of the parameters of `get_extension_actor` are necessary bits of info required to create an `ActionSet` and `ExtensionInfo`. I agree that the two classes ought to be merged, but giving `get_extension_actor` such a merged object simply shifts the parameters from a factory method to an init method...

It seems better to simply leave the parameters where they are. Thoughts?

tests/ga/extension_actor.py
kevinclark19a: Just so we're on the same page, you mean make the `add_extension_actors` function a context manager?

tests/protocol/mockwiredata.py
kevinclark19a: In this case, we actually _do_ want to import everything, though I do realize that more things might be added to that module we don't want (e.g. private implementation details, mainly). I don't think that's likely though, as that file only has data in it.

Is this an example that proves the rule? or do we still want to avoid the `import *`?

((The only real benefit to keeping the star is, I guess, readability, though maybe that's outweighed by the code smell))

tests/protocol/mockwiredata.py
kevinclark19a: I've pushed a change; let me know if the examples help.

tests/ga/extension_actor.py
kevinclark19a: They're referenced by external functions inside the encompassing module; is there a standard way of indicating package-private?

tests/ga/extension_actor.py
```diff
@@ -30,35 +34,49 @@
...
         """
-        A simple fail action with the correct function signature for ActionSet actions.
+        A simple fail action with the correct function signature for ExtensionManifestInfo actions.
         """
         raise ExtensionError("FailAction called.")
 
+    @staticmethod
+    def generate_unique_fail():
+        """
+        TODO: Desc
```
pgombar, line 71: Add desc? 😄

tests/ga/extension_actor.py
```diff
@@ -87,58 +99,55 @@ def _wrap_with_write_status(cmd, name, version):
         """
 
         def wrapped_cmd(*args, **kwargs):
-            filename = "/".join([filedir, "{0}.status".format(wrapped_cmd.inc)])
+            filename = "/".join([filedir, "status", "{0}.status".format(wrapped_cmd.inc)])
```
pgombar, line 102: os.path.join with os.path.sep instead of "/"

tests/ga/extension_actor.py
```diff
@@ -0,0 +1,465 @@
...
+        
+        # We save the patched popens within the protocol to enable additions in future function calls
+        # (e.g. update_extension_actors).
+        protocol._patched_popens = [ ] # Populated later; timing matters, and a patched_popen needs to be started immediately after creation.
+        # We also need actor objs to, amoung other things, guarantee that the same actor doesn't have
+        # two different patch_popen's running at the same time.
+        protocol._actors = [firstActor]
+        protocol._actors.extend(remainingActors)
+
+        protocol.set_http_handlers(http_get_handler=_generate_mock_http_get(protocol._actors[1:]),
```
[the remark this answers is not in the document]
pgombar: bump

tests/ga/extension_actor.py
```diff
@@ -0,0 +1,465 @@
...
+            for patched_popen in protocol._patched_popens:
+                patched_popen.stop()
+
+
+def add_extension_actors(protocol, incarnation, *actors):
+    """
+    Given a protocol obj returned by a extension_actor.get_protocol_and_handler(firstActor, *remainingActors) call, injects the
+    extensions emulated by (the objects within) actors.
+    
+    It accomplishes this though applying the following updates onto the provided protocol:
```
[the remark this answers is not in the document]
pgombar: bump

tests/ga/extension_actor.py
```diff
@@ -251,70 +261,89 @@ def add_extension_actors(protocol, incarnation, *actors):
...
+            "uninstallCommand": dict(key="{0}_uninstall".format(formatted_id), action=Actions._wrap_with_write_status(uninstall_action, name, version)),
+            "updateCommand": dict(key="{0}_update".format(formatted_id), action=Actions._wrap_with_write_status(update_action, name, version)),
+            "enableCommand": dict(key="{0}_enable".format(formatted_id), action=Actions._wrap_with_write_status(enable_action, name, version)),
+            "disableCommand": dict(key="{0}_disable".format(formatted_id), action=Actions._wrap_with_write_status(disable_action, name, version))
         }
 
-    def items(self):
+    def as_generator(self):
+        """
+        TODO: Desc
```
pgombar, line 304: Add desc?

tests/ga/extension_actor.py
```diff
@@ -354,69 +383,61 @@ def return_mock_popen_obj(*args, **kwargs):
...
-        self.get_command = lambda cmd: getattr(self.action_scope, actionSet.getKeyForCommand(cmd), None)
-        self.extension_info = extensionInfo
+        self.manifest_info = manifest_info
 
-        self.statusBlobs = []
+        self.status_blobs = []
+    
+    def get_command(self, cmd):
+        """
+        TODO: Desc
```
pgombar, line 431: Add desc?

tests/protocol/mockwiredata.py
pgombar: Gotcha! I actually didn't know the sequence of the attributes was arbitrary, but it's something we can't control. Nice catch!

tests/protocol/mockwiredata.py
pgombar: We want to be intentional with what we import. Today, this is fine, since you are using everything from that module so it's simpler to do `import *`. However, this is bad practice for several reasons, some being:
1) You don't know exactly what's imported and it's not easy to find which module this was imported from and reduces readability.
2) Let's say a module does an `import *` on this module. Then it would implicitly also import everything this module is importing, which can lead to clashes with methods defind in the current module.
3) The linter will complain about this 😄 

We've had actual issues due to this bad practice before, tests were mocking objects imported from the wrong module, etc. These two PRs did the cleanup of such code: #1704 and #1900.

tests/ga/extension_actor.py
```diff
@@ -0,0 +1,465 @@
...
+                first_patch = first_ext.patch_popen()
+                first_patch.start()
+                second_patch = second_ext.patch_popen()
+                second_patch.start()
+
+                # Execute code that might try to invoke either extension.
+                {more code}
+
+            finally:
+                first_patch.stop()
```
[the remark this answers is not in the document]
pgombar: bump

tests/ga/test_extension.py
pgombar: bump

tests/ga/test_extension.py
pgombar: We need to validate the order of commands, not just that they were called.

tests/ga/extension_actor.py
kevinclark19a: lol right. Thanks for spotting this

tests/ga/extension_actor.py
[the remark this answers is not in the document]
kevinclark19a: Weirdly, this first comment doesn't show up in the Conversation's tab for me, even if "bump" does... Anyways, this code probably needs a comment; the reasoning here is that we're instantiating the `WireProtocol` object with the `first_actor`'s data (line 198: `first_actor.data_fetcher`), so that data gets queried after the `http_get_handler` returns, assuming that said handler doesn't actually handle the request. 

I suppose that we could add the first actor to the handler: it would be redundant, but wouldn't (shouldn't) change the overall effect.

tests/ga/extension_actor.py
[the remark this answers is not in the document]
kevinclark19a: Sorry, this comment also didn't show up on the conversations tab. It's definitely supposed to be "through"; I'll make sure to update.

tests/ga/extension_actor.py
kevinclark19a: Forgot to check my TODO's before pushing!

tests/ga/extension_actor.py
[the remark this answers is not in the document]
kevinclark19a: ((same story with the conversations tab as before)) 
I'm not actually sure if it matters; this is the way it's done in the code and nothing crashes-- I realize that's a low bar, but it does grant a sliver of confidence.

Maybe you're right on the semantics, though; `second_patch` captures `first_patch` in its closure (defined as `original_popen` on line 470 in this file), so stopping first_patch first might be problematic. As for why it doesn't matter in the current code base, a `subprocess.Popen` call would need to happen between the two `stop()`s to expose the potential error. 

To be on the safe side, I'll push a fix to this docstring and the actual codebase; it's a simple list reversal anyways.

tests/ga/test_extension.py
kevinclark19a: Sorry, didn't see this in the conversations tab.

Actually, since we're running the `run()` command, the `exthanders_handler` object would've swallowed any exceptions caused by the commands failing. Our only mechanism for checking a failed extension command is to check the `PUT /status` http endpoint (this is not unique to this PR's solution; rather, its intrinsic to the wholistic style of unit tests we have).

I hadn't thought to include a `ExtensionActor.status_blobs` check for every test, though there are some that have it. Do you think this would be a worthwhile addition to the tests?

tests/ga/test_extension.py
kevinclark19a: I agree, and I believe @narrieta does as well. As I see it, there are two options:

1. Do it manually in every single test.
2. Bake it into the `extension_actor` module as a function/method.

My idea for option 1 is a list that can be appended to from within lambdas local to the test; the keys added to the list can then be checked for the correct order after the fact.

It seems like this is something we'd want to do on basically every test dealing with extensions, though, so it would be great to extend that idea into something that works with option 2. I'm running into some scoping problems, however. I think I need to think about this some more. If you have an alternative, let me know.

tests/protocol/mockwiredata.py
kevinclark19a: Sure, that makes sense. I'll go ahead and push a fix. Thanks!

tests/protocol/mocks.py
```diff
@@ -27,13 +27,15 @@
 
 
 @contextlib.contextmanager
-def mock_wire_protocol(mock_wire_data_file, http_get_handler=None, http_post_handler=None, http_put_handler=None, fail_on_unknown_request=True):
+def mock_wire_protocol(mock_wire_data_source, http_get_handler=None, http_post_handler=None, http_put_handler=None, fail_on_unknown_request=True,
```
narrieta, line 30: it's odd to take both mock_wire_data_source and mockwiredata_factory

it seems like the expectation of the code is that when the source is a file the factory must be aget_file_based_wire_protocol_data and when a fetcher it must be get_dynamic_wire_protocol_data

let's try to improve this design

tests/protocol/mockwiredata.py
```diff
@@ -23,76 +23,23 @@
...
-        "test_ext": "ext/sample_ext-1.3.0.zip",
-        "remote_access": None,
-        "in_vm_artifacts_profile": None
-}
+from tests.protocol.mockwiredata_fetcher import DEFAULT_FETCHER, data_files_to_fetcher, generate_ext_fetcher_func
 
-DATA_FILE_IN_VM_ARTIFACTS_PROFILE = DATA_FILE.copy()
-DATA_FILE_IN_VM_ARTIFACTS_PROFILE["ext_conf"] = "wire/ext_conf_in_vm_artifacts_profile.xml"
-DATA_FILE_IN_VM_ARTIFACTS_PROFILE["in_vm_artifacts_profile"] = "wire/in_vm_artifacts_profile.json"
+from tests.protocol.mockwiredata_filenames import DATA_FILE, DATA_FILE_IN_VM_ARTIFACTS_PROFILE, \
```
narrieta, line 28: only DATA_FILE is being referenced in this file

tests/protocol/mockwiredata_fetcher.py
```diff
@@ -0,0 +1,65 @@
...
+        finally:
+            file.close()
+
+        return zip_file_buffer.getvalue()
+
+    return ext_fetcher_func
+
+
+
+DEFAULT_FETCHER = data_files_to_fetcher(mockwiredata_filenames.DATA_FILE)
```
larohra, line 65: Correct me if I'm wrong, but the purpose of this is to mainly delay-load reading from the file right? How does this fit into the whole dynamic loading scenario? 
(Sorry I'm still a bit lost with these changes)

PR #1937, opened 2020-07-07
Fix telemetry unicode errors

larohra:
## Description
This PR mainly deals with ensuring that the telemetry events sent by the agent are encoded properly in UTF-8 for both Py2 and Py3. This is to support special characters in telemetry.

Issue # 

---

### PR information


azurelinuxagent/common/protocol/wire.py
```diff
@@ -1109,12 +1107,16 @@ def report_event(self, event_list):
...
+                buf[event.providerId] = b''
+                logger.verbose("No of events this request = {0}".format(events_per_request))
+                events_per_request = 0
             buf[event.providerId] = buf[event.providerId] + event_str
+            events_per_request += 1
 
         # Send out all events left in buffer.
         for provider_id in list(buf.keys()):
             if len(buf[provider_id]) > 0:
+                logger.verbose("No of events this request = {0}".format(events_per_request))
```
larohra, line 1119: Added extra verbose log to capture the amount of events that we send per request, this might be useful in figuring out telemetry pipeline issues if needed.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1093,10 +1090,11 @@ def send_event(self, provider_id, event_str):
 
     def report_event(self, event_list):
         buf = {}
+        events_per_request = 0
         # Group events by providerId
         for event in event_list.events:
             if event.providerId not in buf:
-                buf[event.providerId] = ""
+                buf[event.providerId] = b''
```
larohra, line 1097: Converted the buffer strings to `bytes`. This is mainly needed for py3 as `str.encode()` converts to `bytes`. In Py2, `type(bytes) == type(str)` so this is a no-op.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -510,10 +510,8 @@ def event_to_v1(event):
     params = ""
     for param in event.parameters:
         params += event_param_to_v1(param)
-    event_str = ('<Event id="{0}">'
-                 '<![CDATA[{1}]]>'
-                 '</Event>').format(event.eventId, params)
-    return event_str
+    event_str = ustr('<Event id="{0}"><![CDATA[{1}]]></Event>').format(event.eventId, params)
+    return event_str.encode("utf-8")
```
kevinclark19a, line 514: Wouldn't we want to keep event_str as a ustr object? That way, we can limit encoding (and therefore, raw strings that will except on calls to formats with unicode args) to writing to disk and telemetry.

tests/common/test_event.py
```diff
@@ -764,6 +767,43 @@ def http_post_handler(url, body, **__):
...
+                "Sævör grét áðan því úlpan var ónýt",
+                "いろはにほへとちりぬるを わかよたれそつねならむ うゐのおくやまけふこえて あさきゆめみしゑひもせす",
+                "? דג סקרן שט בים מאוכזב ולפתע מצא לו חברה איך הקליטה"
+                "Pchnąć w tę łódź jeża lub ośm skrzyń fig",
+                "Normal string event"
+            ]
+            for msg in test_messages:
+                add_event('TestEventEncoding', message=msg)
+                event_list = event.collect_events()
+                protocol.client.report_event(event_list)
```
kevinclark19a, line 800: Could there be issues within exthandlers.py (i.e. leading up to ExtHandlerInstance.add_event) and monitor.py (i.e. leading up to MonitorHandler.collect_and_sent_events) that are hidden by the test calling the client directly?

azurelinuxagent/common/protocol/wire.py
larohra: This is actually the step before sending it out to telemetry. After this we directly call the common restutil function which actually sends it out but I didn't want to change it there because that is being called from a lot of places and that introduces a bigger scope for an error.

tests/common/test_event.py
larohra: There possibly can be, but that is something that is not being addressed by this PR. In this I'm mainly stabilizing the channel which we use to transmit the events, for the other things, we would have to analyse the scenarios and fix them respectively :)

tests/protocol/test_wire.py
```diff
@@ -362,7 +362,7 @@ def test_send_event(self, mock_http_request, *args):
 
         event_str = u'a test string'
         client = WireProtocol(WIRESERVER_URL).client
-        client.send_event("foo", event_str)
+        client.send_event("foo", event_str.encode('utf-8'))
```
kevinclark19a, line 365: Should inputs to send_event be of type str, or of type ustr? This maybe harkens back to my prior comment about keeping encode secluded within communicating with the outside world.

kevinclark19a: I'm assuming that this is, like explained in the other comment, designed to keep changes out of the send_event function.

larohra: Technically it should be ustr (which is the same as str for Py3) here and only be encoded in the restutil function which actually sends the data out. But since right now I'm only focusing the changes for the telemetry event, we're encoding the string in this function which is just 1 level up from the restutil function to limit the blast radius of this change. 

Does that make sense? We can discuss more on this offline if you have additional doubts.

azurelinuxagent/common/protocol/wire.py
narrieta: can we add utf_8 somewhere in the function name to be more explicit on what it returns ?

azurelinuxagent/common/utils/textutil.py
```diff
@@ -396,3 +398,20 @@ def format_memory_value(unit, value):
         raise TypeError('Value must be convertible to a float')
 
     return int(value * units[unit])
+
+def str_to_encoded_ustr(s, encoding='utf-8'):
+    from azurelinuxagent.common.version import PY_VERSION_MAJOR
```
narrieta, line 403: can we move the import to the top of the file (easier to see module dependencies when reading the code)

azurelinuxagent/common/utils/textutil.py
```diff
@@ -396,3 +398,20 @@ def format_memory_value(unit, value):
         raise TypeError('Value must be convertible to a float')
 
     return int(value * units[unit])
+
+def str_to_encoded_ustr(s, encoding='utf-8'):
```
narrieta, line 402: not sure i understand what the function is trying to do. could you add a doctring? thanks

azurelinuxagent/common/utils/textutil.py
larohra: I tried doing that initially but it started giving me a lot of cyclic dependency errors :) 
I gave up fixing it because it required extra refactoring (and to be brutally honest I also got lazy because I didnt want to deal with it). I can take a stab and fix it though if you feel we should start making that effort to make our agent better.

azurelinuxagent/common/utils/textutil.py
larohra: Make sense, I'll add that.

larohra: Done

azurelinuxagent/ga/monitor.py
```diff
@@ -247,10 +249,14 @@ def collect_and_send_events(self):
...
-        event_list = collect_events()
+        try:
+            event_list = collect_events()
 
-        if len(event_list.events) > 0:
-            self.protocol.report_event(event_list)
+            if len(event_list.events) > 0:
+                self.protocol.report_event(event_list)
+        except Exception as e:
+            err_msg = "Failure in collecting/sending Agent events: {0}".format(textutil.str_to_encoded_ustr(e))
```
larohra, line 258: PeriodicOperation was already handling exceptions for us here, I mainly added the explicit checks here to send out a Kusto event incase the `collect_and_send_events` ever throws some unknown error

azurelinuxagent/common/event.py
```diff
@@ -77,6 +77,8 @@ class WALAEventOperation:
     CGroupsInfo = "CGroupsInfo"
     CGroupsInitialize = "CGroupsInitialize"
     CGroupsLimitsCrossed = "CGroupsLimitsCrossed"
+    CollectEventErrors = "CollectEventErrors"
```
larohra, line 79: @pgombar gave me the idea of reporting the number of dropped events per run to Kusto so that we have some data on how many events our agent is dropping. To achieve this I broke down the errors into these 4 categories. The events can either be dropped either while collecting them from filesystem or while sending them to Wireserver. The sum of all 4 of these can specify the total no of dropped events. 

I broke it down into 4 different parts mainly to have more granularity over the errors that we see and hence can work towards understanding the problem better.

azurelinuxagent/common/event.py
```diff
@@ -581,9 +585,12 @@ def collect_events(self):
         Retuns a list of events that need to be sent to the telemetry pipeline and deletes the corresponding files
         from the events directory.
         """
+        max_errors_to_report_per_run = 5
```
larohra, line 586: To not overload Kusto with all possible types of errors, just capturing the first 5 and reporting those. The count would still reflect the exact count but we would just be capturing the first 5 errors. We can modify this number based on the data that we receive as we move forward.

azurelinuxagent/common/utils/textutil.py
pgombar: Should we move this function to `future.py`? `textutil` seems to have been designed as a common utility class, so ideally it shouldn't be relying on anything else in the agent.

azurelinuxagent/common/event.py
```diff
@@ -574,14 +578,25 @@ def _trim_extension_event_parameters(event):
 
         event.parameters = trimmed_params
 
+    @staticmethod
+    def report_dropped_events_error(count, errors, op, max_errors_to_report):
+        err_msg_format = "DroppedEventsCount: {0}\nReasons (first {1}): {2}"
```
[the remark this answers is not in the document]
larohra: Done

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1092,29 +1089,53 @@ def send_event(self, provider_id, event_str):
...
-                                     "Single event too large: {0}, with the length: {1} more than the limit({2})"
-                                     .format(str(details_of_event), len(event_str), MAX_EVENT_BUFFER_SIZE))
-                continue
-            if len(buf[event.providerId] + event_str) >= MAX_EVENT_BUFFER_SIZE:
-                self.send_event(event.providerId, buf[event.providerId])
-                buf[event.providerId] = ""
-            buf[event.providerId] = buf[event.providerId] + event_str
+            try:
+                if event.providerId not in buf:
+                    buf[event.providerId] = b''
```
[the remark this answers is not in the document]
larohra: So to avoid double encoding and also to ensure that the `len(encoded_string) < MAX_EVENT_BUFFER_SIZE`, I changed the `buf[event.providerId]` to save the encoded string. Encode() in py3 converts the string to bytes and keeps it as str in py2. This is the reason I changed it to bytes.

azurelinuxagent/common/protocol/wire.py
larohra: The main use case is to check on a request level how many events we're sending per request and if we dropped any and how long each log takes to respond (since its a single threaded environment). This gives us insight into how long each request to wireserver telemetry takes to respond and also what's the average number of events per thread. 

That was the main motivation behind adding this. If you guys feel this is an overkill I can remove it.

azurelinuxagent/common/utils/restutil.py
```diff
@@ -319,10 +319,11 @@ def _http_request(method, host, rel_uri, port=None, data=None, secure=False,
                                          conn_port,
                                          timeout=10)
 
+    # Logger requires the data to be a ustr to log properly, ensuring that the data string that we log is always ustr.
```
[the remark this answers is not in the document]
larohra: So the function `str_to_encoded_ustr` is encoding the string to utf-8 if no other encoding is specified. The only difference here is that its converting the encoded string back to ustr for logging purposes. 

In an ideal world we should only encode the request right before sending it out (in the L:329 of this file) but to limit the surface area for errors, I'm encoding the data earlier and just decoding it here for logging.

azurelinuxagent/common/utils/textutil.py
larohra: I totally understand your concern but a lot of different functions in textutil have similar checks for py2 :)

Apart from the py-version check I dont have any other dependency and since this is a function that deals with strings in general I feel textutil is better suited for the job. 

But yeah if you guys feel strongly about it being in future.py, I'll move it not a big deal

pgombar: Ahh, I just noticed we use inline imports for `PY_VERSION_MAJOR` all over the place in this module. Fun! :smile:

pgombar: Actually, we mix it up... function `gen_password_hash` in this module is querying `sys.version_info` instead of relying on `PY_VERSION_MAJOR` (which does the same thing). We will eventually need to address this since I'm guessing the linter will complain about imports not being at the top of the file and then either break the circular dependency or remove `PY_VERSION_MAJOR` and use `sys.version_info` directly.

Up to you if you want to take this up or not.

azurelinuxagent/common/utils/textutil.py
narrieta: thanks

azurelinuxagent/common/utils/textutil.py
narrieta: sounds good

azurelinuxagent/common/protocol/wire.py
```diff
@@ -488,7 +488,7 @@ def put_page_blob(self, url, data):
 
 
 def event_param_to_v1(param):
-    param_format = '<Param Name="{0}" Value={1} T="{2}" />'
+    param_format = ustr('<Param Name="{0}" Value={1} T="{2}" />')
```
narrieta, line 491: can we add utf_8 somewhere in the function name to be more explicit on what it returns ?

larohra: Done

PR #1943, opened 2020-07-13
Fix check for systemd-run failure when invoking extensions

pgombar:
## Description

We were erroneously classifying systemd-run failures when invoking extensions which produce a long stdout and/or stderr. The reason is because we check for the unit name in the beginning of stderr as a signal that systemd-run was invoked successfully. The change in this PR is to check for the unit name in stderr before truncating it for telemetry purposes.

Other fixes:
1) We had a bug in `extensionprocessutils.py` where the `TELEMETRY_MESSAGE_MAX_LEN` wasn't being patched since it was being assigned in the function signature instead of inside of the function, which produced misleading results in the unit tests.
2) Fixes a bug in `test_start_extension_command_should_invoke_the_command_directly_if_systemd_times_out` where because of the indentation the asserts weren't being executed since they were part of the context manager instead of being outside of its scope.

---

### PR information


azurelinuxagent/common/utils/extensionprocessutil.py
```diff
@@ -99,7 +99,7 @@ def read_output(stdout, stderr):
         return format_stdout_stderr("", "Cannot read stdout/stderr: {0}".format(ustr(e)))
 
 
-def format_stdout_stderr(stdout, stderr, max_len=TELEMETRY_MESSAGE_MAX_LEN):
+def format_stdout_stderr(stdout, stderr):
```
pgombar, line 102: Lesson to learn: don't assign constants in the function signature since this is loaded before the function body, so patching the constant will not work.

tests/utils/test_extension_process_util.py
```diff
@@ -164,29 +164,26 @@ def test_handle_process_completion_should_raise_on_nonzero_exit_code(self):
                 self.assertEquals(context_manager.exception.code, error_code)
                 self.assertIn("Non-zero exit code:", ustr(context_manager.exception))
 
-    def test_read_output_it_should_return_no_content(self):
+    def test_read_output_should_return_no_content(self):
```
pgombar, line 167: Once I fixed the patching for `TELEMETRY_MESSAGE_MAX_LEN`, these tests produced different (correct) results :)

tests/common/test_cgroupapi.py
```diff
@@ -27,6 +27,8 @@
 from azurelinuxagent.common.future import ustr
```
pgombar, line 27: At some point I think I clicked on "format file" so there's a lot of whitespace updates, I'll point out relevant updates to the tests.

tests/common/test_cgroupapi.py
```diff
@@ -27,6 +27,8 @@
...
 from azurelinuxagent.common.utils import shellutil, fileutil
 from nose.plugins.attrib import attr
+
+from azurelinuxagent.common.utils.extensionprocessutil import TELEMETRY_MESSAGE_MAX_LEN
```
pgombar, line 31: At some point I think I clicked on "format file" so there's a lot of whitespace updates, I'll point out relevant updates to the tests.

tests/common/test_cgroupapi.py
```diff
@@ -727,11 +752,14 @@ def mock_popen(*args, **kwargs):
...
+                    extension_calls = [args[0] for (args, _) in popen_patch.call_args_list if
+                                       "echo 'success'" in args[0]]
 
                     self.assertEquals(2, len(extension_calls), "The extension should have been invoked exactly twice")
-                    self.assertIn("systemd-run --unit=Microsoft.Compute.TestExtension_1.2.3", extension_calls[0], "The first call to the extension should have used systemd")
-                    self.assertEquals("echo 'success'", extension_calls[1], "The second call to the extension should not have used systemd")
+                    self.assertIn("systemd-run --unit=Microsoft.Compute.TestExtension_1.2.3", extension_calls[0],
+                                  "The first call to the extension should have used systemd")
+                    self.assertEquals("echo 'success'", extension_calls[1],
+                                      "The second call to the extension should not have used systemd")
```
pgombar, line 762: I spent several hours trying to figure out why the breakpoint in L755 was never hit... turns out we had a bug checked in where the indentation was such that the asserts were inside of the context manager, instead of outside it.

This is a nasty bug since it's so easy to miss. Could linters figure this one out? Thoughts, @kevinclark19a?

tests/common/test_cgroupapi.py
```diff
@@ -740,29 +768,70 @@ def mock_popen(*args, **kwargs):
...
+                    self.assertIn("Non-zero exit code", ustr(context_manager.exception))
+                    # The scope name should appear in the process output since systemd-run was invoked and stderr
+                    # wasn't truncated.
+                    self.assertIn("Microsoft.Compute.TestExtension_1.2.3", ustr(context_manager.exception))
+
+    @attr('requires_sudo')
+    @patch("azurelinuxagent.common.cgroupapi.add_event")
+    @patch('time.sleep', side_effect=lambda _: mock_sleep())
+    @patch("azurelinuxagent.common.utils.extensionprocessutil.TELEMETRY_MESSAGE_MAX_LEN", 5)
+    def test_start_extension_command_should_not_use_fallback_option_if_extension_fails_with_long_output(self, *args):
```
pgombar, line 803: New test to cover long extension outputs. This would fail before the changes in this PR.

tests/common/test_cgroupapi.py
kevinclark19a: Ah. Interesting. I'm not sure a linter could actually help with this... I don't see how it would differentiate between the legitimate caught raise statement presiding (presumably) inside of the `start_extension_command` call, and the erroneously caught one inside the `assertEquals` call. As far as the AST is concerned, they're basically identical. 

Still, there might be some way of using attributes to let the linter know what we want caught and what we want to error out... I'll have to look into it more.

tests/common/test_cgroupapi.py
```diff
@@ -740,29 +768,70 @@ def mock_popen(*args, **kwargs):
...
                             stderr=stderr)
 
-                        # We should have invoked the extension command only once, in the systemd-run case
-                        self.assertEquals(1, patch_mock_popen.call_count)
-                        args = patch_mock_popen.call_args[0][0]
-                        self.assertIn("systemd-run --unit", args)
+                    extension_calls = [args[0] for (args, _) in popen_patch.call_args_list
+                                       if long_stdout_stderr_command in args[0]]
+
+                    self.assertEquals(1, len(extension_calls), "The extension should have been invoked exactly twice")
```
larohra, line 826: NIT: "The extension should have been invoked exactly **once**"

tests/common/test_cgroupapi.py
narrieta: can you point out the change in indentation? I'm having a hard time finding it in the diff

tests/common/test_cgroupapi.py
narrieta: so we need to review only the lines where you added comments?

tests/common/test_cgroupapi.py
pgombar: I thought I was blind.. turns out I commented on the wrong method. The content of the comment is correct, but the method in question is `test_start_extension_command_should_not_use_fallback_option_if_extension_fails` instead of this one.

tests/common/test_cgroupapi.py
```diff
@@ -740,29 +768,70 @@ def mock_popen(*args, **kwargs):
...
+                            extension_name="Microsoft.Compute.TestExtension-1.2.3",
+                            command=command,
+                            timeout=300,
+                            shell=True,
+                            cwd=self.tmp_dir,
+                            env={},
+                            stdout=stdout,
+                            stderr=stderr)
+
+                    extension_calls = [args[0] for (args, _) in popen_patch.call_args_list if command in args[0]]
```
pgombar, line 787: *This* is where the indentation was wrong. It was at the same level with the context manager, instead of outside, like now.

tests/common/test_cgroupapi.py
pgombar: Added a comment on the right method.

tests/common/test_cgroupapi.py
pgombar: For this file, that's the case, yes.

tests/common/test_cgroupapi.py
narrieta: But this new code... you mentioned it was in existing code

narrieta: ok, i see it now

narrieta: the assert for root was hiding the issue

larohra: but it should've failed in travis then right?

pgombar: It didn't fail because these asserts were never executed. I ran the test both from sudo shell and from sudo PyCharm and could never hit the asserts. I'm guessing it has something to do with the context manager which just exits the current scope when the object is released?

tests/common/test_cgroupapi.py
narrieta: your difftool should allow you selectively undo the changes you don't want (meld does). In the future let's try to avoid global formatting changes, since it makes the review and reading the history of the code harder,

narrieta: (you can even undo the changes after commiting, if you feel like doing it here)

tests/common/test_cgroupapi.py
larohra: ohh dang. Good catch though! Hopefully we can catch these moving on

narrieta: > I'm guessing it has something to do with the context manager which just exits the current scope when the object is released?

The test case raises, so the call to start_extension_command will exit the with statement.

I seem to remember at some point I formatted a few tests in this file so that may have moved the assert to the with statement.

PR #1944, opened 2020-07-13
Use run_command to execute iptables

narrieta:
iptables was using the deprecated run/run_get_output functions

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -60,17 +60,17 @@ _IPTABLES_VERSION_PATTERN = re.compile("^[^\d\.]*([\d\.]+).*$")
 _IPTABLES_LOCKING_VERSION = FlexibleVersion('1.4.21')
 
 
-def _get_iptables_version_command():
-    return ["iptables", "--version"]
-
-
 def _add_wait(wait, command):
     """
-    If 'wait' is True, adds the wait option (w) to the given iptables command line
+    If 'wait' is True, adds the wait option (-w) to the given iptables command line
     """
     if wait:
         command.insert(1, "-w")
     return command
+
+
+def _get_iptables_version_command():
+    return ["iptables", "--version"]
 
 
 def _get_firewall_accept_command(wait, command, destination, owner_uid):
@@ -162,6 +162,7 @@ class DefaultOSUtil(object):
                     # Transient error  that we ignore.  This code fires every loop
                     # of the daemon (60m), so we will get the value eventually.
                     return 0
+                logger.error("Failed to get firewall packets: {0}", ustr(e))
                 return -1
 
             return 0
@@ -249,17 +250,14 @@ class DefaultOSUtil(object):
             try:
                 drop_rule = _get_firewall_drop_command(wait, "-C", dst_ip)
                 shellutil.run_command(drop_rule)
-                firewall_established = True
-            except Exception as e:
-                if isinstance(e, CommandError) and e.returncode == 2:
+                logger.verbose("Firewall appears established")
+                return True
+            except CommandError as e:
+                if e.returncode == 2:
                     self.remove_firewall(dst_ip, uid)
                     msg = "please upgrade iptables to a version that supports the -C option"
                     logger.warn(msg)
                     raise Exception(msg)
-
-            if firewall_established:
-                logger.verbose("Firewall appears established")
-                return True
 
             # Otherwise, append both rules
             try:
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -89,10 +89,6 @@ def _get_firewall_packets_command(wait):
     return _add_wait(wait, ["iptables", "-t", "security", "-L", "OUTPUT", "--zero", "OUTPUT", "-nxv"])
 
 
-def _get_firewall_flush_command(wait):
-    return _add_wait(wait, ["iptables", "-t", "security", "-flush"])
-
-
 # Precisely delete the rules created by the agent.
 # this rule was used <= 2.2.25.  This rule helped to validate our change, and determine impact.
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/osutil/default.py
```diff
@@ -167,11 +198,13 @@ def _delete_rule(self, rule):
...
-            rc = shellutil.run(rule, chk_err=False)
-            if rc == 1:
-                return
-            elif rc == 2:
-                raise Exception("invalid firewall deletion rule '{0}'".format(rule))
+            try:
+                rc = shellutil.run_command(rule)
+            except CommandError as e:
+                if e.returncode == 1:
+                    return
```
kevinclark19a, line 206: Why is a returncode of 1 not an error? Should we document this as a comment?

narrieta: I image the code was written like this to avoid flagging an error if the rule does not exist, though this is not very robust.
I was tempted to change the logic, but at the end decided against it. I am not sure about the intention of the loop either.

tests/common/osutil/test_default.py
```diff
@@ -633,277 +634,252 @@ def test_conf_sudoer(self, mock_dir):
...
 
-        mock_run.assert_has_calls([
-            call(osutil.FIREWALL_DELETE_CONNTRACK_ACCEPT.format(wait, dst_ip), chk_err=False),
-        ])
-        mock_output.assert_has_calls([
-            call(osutil.IPTABLES_VERSION)
-        ])
-        self.assertFalse(osutil._enable_firewall)
+        with TestOSUtil._mock_iptables() as mock_iptables:
+            delete_conntrack_accept_command = mock_iptables.set_command(osutil._get_firewall_delete_conntrack_accept_command(mock_iptables.wait, mock_iptables.destination), exit_code=2)
```
pgombar, line 874: Why is this an invalid rule? It's the one we phased out after 2.2.25?

azurelinuxagent/common/osutil/default.py
```diff
@@ -56,25 +56,50 @@
...
-
-FIREWALL_ACCEPT = "iptables {0} -t security -{1} OUTPUT -d {2} -p tcp -m owner --uid-owner {3} -j ACCEPT"
-# Note:
-# -- Initially "flight" the change to ACCEPT packets and develop a metric baseline
-#    A subsequent release will convert the ACCEPT to DROP
-# FIREWALL_DROP = "iptables {0} -t security -{1} OUTPUT -d {2} -p tcp -m conntrack --ctstate INVALID,NEW -j ACCEPT"
-FIREWALL_DROP = "iptables {0} -t security -{1} OUTPUT -d {2} -p tcp -m conntrack --ctstate INVALID,NEW -j DROP"
-FIREWALL_LIST = "iptables {0} -t security -L -nxv"
-FIREWALL_PACKETS = "iptables {0} -t security -L OUTPUT --zero OUTPUT -nxv"
-FIREWALL_FLUSH = "iptables {0} -t security --flush"
```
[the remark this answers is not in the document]
pgombar: Ah, I see it's not being used anywhere else.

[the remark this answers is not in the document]
narrieta: yup

azurelinuxagent/common/osutil/default.py
```diff
@@ -119,21 +144,23 @@ def get_firewall_dropped_packets(self, dst_ip=None):
...
+                    m = pattern.match(line)
+                    if m is not None:
+                        return int(m.group(1))
+
+            except Exception as e:
+                if isinstance(e, CommandError) and e.returncode == 3:
+                    # Transient error  that we ignore.  This code fires every loop
+                    # of the daemon (60m), so we will get the value eventually.
+                    return 0
+                logger.warn("Failed to get firewall packets: {0}", ustr(e))
```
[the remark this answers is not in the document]
narrieta: we did: run_get_output logs by the default (that's why I had to add the "expected_errors" stuff in the old function)

tests/common/osutil/test_default.py
narrieta: It is confusing, I'll add a comment.

The test tries to hit the path where we check for exit code 2 (invalid rule), but the agent uses only only valid rules, so the mocked command specifies a valid rule, but the return code indicates invalid. When I ported the tests I thought this was odd; should have added a comment then.

narrieta: Now, in the previous version of the code this was not apparent, since the mock using magic (deep knowledge of the impementation) and mocks like this:


```
        mock_run.side_effect = [2]	
        mock_output.side_effect = [(0, version), (1, "Output")]

```

tests/common/osutil/test_default.py
```diff
@@ -633,277 +634,252 @@ def test_conf_sudoer(self, mock_dir):
...
 
 Chain OUTPUT (policy ACCEPT 104 packets, 43628 bytes)
     pkts      bytes target     prot opt in     out     source               destination
        0        0 ACCEPT     tcp  --  any    any     anywhere             168.63.129.16        owner UID match daemon
       32     1920 DROP       tcp  --  any    any     anywhere             168.63.129.16
 
-''')]
-        dst = '168.63.129.16'
+''')
+            self.assertEqual(32, osutil.DefaultOSUtil().get_firewall_dropped_packets('168.63.129.16'))
```
narrieta, line 736: this should use 'destination'

azurelinuxagent/common/osutil/default.py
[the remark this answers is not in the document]
pgombar: My bad, I saw `log_cmd=False` in the old function and interpreted it as not logging the error, whereas that's the `chk_err` flag.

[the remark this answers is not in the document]
narrieta: yes, those 2 flags are confusing

PR #1951, opened 2020-07-17
- SUSE implementation for handling dhcp configuration for hostname push

rjschwei:
+ The default implementation is based on the assumption that a distribution
    uses dhclient, but SUSE uses wicked and has a different config to push
    the hostname to a DHCP server.
- Do not use ifdown when the hostname changes
  + When accelerated networking is enabled ifdown may time out, suspected
    kernel issue. In any event it is not necessary to bring the interface
    down to push a hostname change to the DHCP server, ifup is sufficient.

## Description

Issue # 

---

### PR information


Moustafa-Moustafa (dismissed):
LGTM

anhvoms (dismissed):

trstringer (dismissed):

narrieta:
Code LGTM; running automation.

narrieta:
Automation failed on SUSE 12. We will follow up and post an update here (fyi: @larohra)

larohra:
@rjschwei looks like these changes are breaking the nslookup test we have in our automation suite. 

```
[03_check_nslookup.py]

ret_code: 1

std_out:
Hostname: edp5zvnsaf
[ERROR] Ran into exception: [Errno -2] Name or service not known
'nslookup edp5zvnsaf' returned exit code '1'
```

We use `ip = socket.gethostbyname(hostname)` to get the ip for the hostname.
We get the hostname by using the `hostname` command. 

Please let me know if you need any more info from me to debug this issue further.

rjschwei:
> @rjschwei looks like these changes are breaking the nslookup test we have in our automation suite.
> 
> ```
> [03_check_nslookup.py]
> 
> ret_code: 1
> 
> std_out:
> Hostname: edp5zvnsaf
> [ERROR] Ran into exception: [Errno -2] Name or service not known
> 'nslookup edp5zvnsaf' returned exit code '1'
> ```
> 
> We use `ip = socket.gethostbyname(hostname)` to get the ip for the hostname.
> We get the hostname by using the `hostname` command.
> 
> Please let me know if you need any more info from me to debug this issue further.
> 
> Its only failing for suse-12 and is failing consistently (I did multiple runs to ensure it wasn't a transient issue)

Hard to judge wit this little context but with of failing on SLE 12 and not SLE 15 it could very well be a behavior difference of wicked in that as expected in SLE 15 ifup pushes the hostname and in SLE 12 using ifup alone does not push the hostname.

This is good to know and I can go  back to the network team. This might take a while. Thanks for running the test, I'll come back to this.

rjschwei:
@larohra Can you confirm that the wicked versions in the images you are using between SLES 15 and SLES 12 are the same? They should both be 0.6.60, i.e. the behavior is expected to be the same in SLES 12 and SLES 15. Due to different refresh cycles it is possible that the test suite ended up running with images that have different wicked versions and that may have caused the failure in SLES 12.

To verify that the hostname is being sent by the client the image you are using for testing can be modified by adding and enabling a new systemd service.

```
[Unit]
Description=Enable tcpdump for dhcp traffic tracing
Before=wicked.service

[Service]
Type=oneshot
ExecStart=ip link set up dev eth0
ExecStart=tcpdump -w "dhcp4.pcap" -s 65535 -Uvnfi eth0 'port 67 or port 68'
```
should do the trick.

Anyway, long story short, given that wicked is the same on SLES 12 and SLES 15 the behavior is expected to be identical.

larohra:
@rjschwei doesn't look like that's the case. I did another run on a suse12 machine and printed out the wicked version (which is 0.6.60) but the nslookup test is still failing - 

>16:17:23 [agent-bvt_suse_12_dczbn50hgp] 2020-09-09T23:17:23Z.903 Remote execution std_out: Hostname: edpthk3lr9
16:17:23 [agent-bvt_suse_12_dczbn50hgp] Wicked: wicked 0.6.60
16:17:23 [agent-bvt_suse_12_dczbn50hgp] [ERROR] Ran into exception: [Errno -2] Name or service not known
16:17:23 [agent-bvt_suse_12_dczbn50hgp] 'nslookup edpthk3lr9' returned exit code '1'

Is there anything else I can try? 

Btw we're using basic Azure Marketplace images for our testing without any specific modification on the image, this is the details for the Suse12 machine - 
```json
 "suse_12": {
      "imagePublisher": "SUSE",
      "imageOffer": "sles-12-sp5",
      "sku": "gen1"
    }
```

Please let me know if you need any other data from me

linuxelf001:
@rjschwei - When I deploy SLES 12 sp5 VM using the current image, I notice local file (/usr/lib/python2.7/site-packages/azurelinuxagent/common/osutil/suse.py) on the VM has all this patch code changes except one line

What is the reason behind missing this one line?

narrieta:
@rjschwei for reference, this is the test failing in our automation:

```
import subprocess
import sys
import socket


def main():
    pipe = subprocess.Popen(['hostname'],
                            stdout=subprocess.PIPE,
                            stderr=subprocess.PIPE)
    hostname = list(map(lambda s: s.decode('utf-8').strip(), pipe.stdout.readlines()))[0]
    if hostname[-1:] == '\n':
        hostname = hostname[:-1]

    print("Hostname: {0}".format(hostname))

    exit_code = 0
    try:
        ip = socket.gethostbyname(hostname)
        print("Resolved IP: {0}".format(ip))
    except Exception as e:
        print("[ERROR] Ran into exception: {0}".format(e))
        exit_code = 1

    print("'nslookup {0}' returned exit code '{1}'".format(hostname, exit_code))
    sys.exit(exit_code)


if __name__ == "__main__":
    main()

```

And the error referenced above:

```
16:17:23 [agent-bvt_suse_12_dczbn50hgp] 2020-09-09T23:17:23Z.903 Remote execution std_out: Hostname: edpthk3lr9
16:17:23 [agent-bvt_suse_12_dczbn50hgp] Wicked: wicked 0.6.60
16:17:23 [agent-bvt_suse_12_dczbn50hgp] [ERROR] Ran into exception: [Errno -2] Name or service not known
16:17:23 [agent-bvt_suse_12_dczbn50hgp] 'nslookup edpthk3lr9' returned exit code '1'

```

rjschwei:
> @rjschwei - When I deploy SLES 12 sp5 VM using the current image, I notice local file (/usr/lib/python2.7/site-packages/azurelinuxagent/common/osutil/suse.py) on the VM has this patch code except below one line
> 
> import azurelinuxagent.common.logger as logger
> 
> What is the reason behind missing this one line?

The logger was dropped, this PR has to be re-based.

rjschwei:
> @rjschwei for reference, this is the test failing in our automation:
> 
<snip>
> And the error referenced above:
> 
> ```
> 16:17:23 [agent-bvt_suse_12_dczbn50hgp] 2020-09-09T23:17:23Z.903 Remote execution std_out: Hostname: edpthk3lr9
> 16:17:23 [agent-bvt_suse_12_dczbn50hgp] Wicked: wicked 0.6.60
> 16:17:23 [agent-bvt_suse_12_dczbn50hgp] [ERROR] Ran into exception: [Errno -2] Name or service not known
> 16:17:23 [agent-bvt_suse_12_dczbn50hgp] 'nslookup edpthk3lr9' returned exit code '1'
> ```

Well error code 2 (-2) means `TRY_AGAIN` which according to the man page means:

"""
This is usually a temporary error and means that the local server did not receive a response from an authoritative server. A retry at some later time may succeed.
"""

Given tat I cannot reproduce this with the current image:

```
~> python
Python 2.7.17 (default, Nov 04 2019, 16:08:54) [GCC] on linux2
Could not open PYTHONSTARTUP
IOError: [Errno 2] No such file or directory: '/etc/pythonstart'
>>> import subprocess
>>> import sys
>>> import socket
>>> pipe = subprocess.Popen(['hostname'],
...                             stdout=subprocess.PIPE,
...                             stderr=subprocess.PIPE)
>>> hostname = list(map(lambda s: s.decode('utf-8').strip(), pipe.stdout.readlines()))[0]
>>> hostname
u'rjs-nettest'
>>> if hostname[-1:] == '\n':
...     hostname = hostname[:-1]
... 
>>> hostname
u'rjs-nettest'
>>> ip = socket.gethostbyname(hostname)
>>> ip
'172.16.0.8'
>>> 
```
 and also tested Python 3

```
~> python3
Python 3.4.10 (default, Dec 13 2019, 16:20:47) [GCC] on linux
Could not open PYTHONSTARTUP
FileNotFoundError: [Errno 2] No such file or directory: '/etc/pythonstart'
>>> import subprocess
>>> import sys
>>> import socket
>>> pipe = subprocess.Popen(['hostname'],
...                             stdout=subprocess.PIPE,
...                             stderr=subprocess.PIPE)
>>> hostname = list(map(lambda s: s.decode('utf-8').strip(), pipe.stdout.readlines()))[0]
>>> hostname
'rjs-nettest'
>>> if hostname[-1:] == '\n':
...     hostname = hostname[:-1]
... 
>>> hostname
'rjs-nettest'
>>> ip = socket.gethostbyname(hostname)
>>> ip
'172.16.0.8'
>>> 
```

I am going to say that this may be a timing issue for when the test gets run. It would be interesting to see if the nameserver is configured as expected when the test runs

```
grep nameserver /etc/resolv.conf
```

Also note that the test is misleading when there is an error. When the exception is caught the real exit code is written, to stdout which is not good practice, and then the exit code is forced to be `1` and that is printed with a message that implies the nslookup returned that code, i.e. `1`. But that is not true, in this case gethostbyname actually returned `2` as the error code, well represented as `-2` per convention. 

Last but not least consider that the current images use cloud-init and as such provisioning is not done by the agent. In order to test the code change a hostname change has to take place as only the daemon process of the agent would trigger the push of the hostname to the server. cloud-init does not push the system hostname back to the server as the network configuration received from the server is considered to be consistent.

narrieta:
@rjschwei - thanks for looking into thiis. Let use take a closer look at that test and we will take this PR once we fix the issue.

kevinclark19a:
@rjschwei after some testing, I've verified both that the test error isn't temporary (i.e., isn't fixed by a delay and retries) and that the nameserver seems correctly configured. 

The issue seems to be with cloud-init not properly publishing the hostname; as you pointed out, the provisioning code isn't called for suse12. There _is_ in fact a code path through which `publish_hostname` is called in the extension handler itself, but only on hostname changes. My current theory is that our testing VMs are too short-lived for that scenario, whereas any user-created VM isn't.

I've identified a code change in the agent that fixes the issue, but I think our team needs more discussion on this issue.

kevinclark19a:
Aha! We've located the issue and have an easy fix for it, @rjschwei.

First, here's the bug I've found causing our testing issue. L#113 compares the existing `DHCLIENT_HOSTNAME_OPTION` value to `AUTO`, but the default seems to be `"AUTO"`. Changing that allows our testing to pass and, more importantly, doesn't overwrite the default value when it is there.

### An explanation of the behavior:

As a part of our integration testing process, we take a default azure image, deprovision the existing agent, and install and provision with the new agent from whatever branch we're testing. In the case of suse12, provisioning basically just invokes cloud init, and this code isn't invoked.

Deprovisioning, though, involves calling `set_dhcp_hostname` with a placeholder parameter of `localhost.localdomain`. The code in this PR then appends the line `DHCLIENT_HOSTNAME_OPTION=localhost.localdomain` to the config file. The dh client on the VM then fails to find the real hostname, because it thinks that "`localhost.localdomain`" is supposed to be the hostname (I think, but we're pretty sure).

### As for why this is only happening on a suse12 VM:

I was wondering why this wasn't an issue on, e.g. an ubuntu VM, seeing as deprovisioning there also would write an analogous line to the relevant dhclient config (setting the hostname to "`localhost.localdomain`"). Turns out that the default code checks for a line like "`send host-name gethostname()`". This functions the same as "`DHCLIENT_HOSTNAME_OPTION="AUTO"`", but PR has a typo and checks for `AUTO` instead of `"AUTO"` like our suse 12 image seems to have by default.

rjschwei:
@kevinclark19a Fantastic thanks for chasing this :D

narrieta (approved):

kevinclark19a (approved):

PR #1953, opened 2020-07-17
Disable only when enabled for version update

kevinclark19a:
## Description

Fixed bug to further feature parity with the windows guest agent; during version update, `disableCommand` is now only called of the preexisting extension is in an `ExtHandlerState.Enabled` state.

This change broke some tests in `tests.ga.test_extension.TestExtensionUpdateOnFailure` that did not correctly set the preexisting extension's state before simulating an upgrade. This PR addresses this by manually setting the extension's state, but this isn't an ideal solution; the root cause of these tests breaking is that they call a private method inside of `ga.exthandlers.ExtHandlersHandler` instead of the `run` method of that same class. The private method relies on state assigned by earlier calls in `run`, and this is something that should probably be revisited in the future.

Issue # 

---

### PR information

### DIFF
--- a/tests/ga/test_extension.py
+++ b/tests/ga/test_extension.py
@@ -2580,6 +2580,7 @@ class TestExtensionUpdateOnFailure(ExtensionTestCase):
     def test_uninstall_failed_env_variable_should_set_for_install_when_continue_on_update_failure_is_true(
             self, *args):
         old_handler_i = self._get_ext_handler_instance('foo', '1.0.0')
+        old_handler_i.set_handler_state(ExtHandlerState.Enabled)
         new_handler_i = self._get_ext_handler_instance('foo', '1.0.1', continue_on_update_failure=True)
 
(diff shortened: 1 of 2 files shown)


tests/ga/test_extension.py
```diff
@@ -2834,6 +2841,19 @@ def test_success_code_should_set_in_env_variables_on_cmd_success(self, _):
             self.assertTrue(test_env_file_name + " -u" in update_kwargs['message'] and "%s=%s" % (
                 ExtCommandEnvVariable.DisableReturnCode, exit_code) in update_kwargs['message'])
 
+    @patch('time.sleep', side_effect=lambda _: mock_sleep(0.0001))
+    def test_disable_should_not_be_called_on_disabled_ext_during_version_upgrade(self, _):
```
larohra, line 2845: NIT: The test name is a bit confusing, it should be more specific, something like - 
`test_disable_should_not_be_called_during_version_upgrade_if_not_enabled`

PR #1958, opened 2020-07-22
Use run_command for ip route

narrieta:
ip route was using run/run_get_output


azurelinuxagent/common/osutil/default.py
```diff
@@ -1018,15 +1022,18 @@ def get_ip4_addr(self):
         return self.get_first_if()[1]
 
     def set_route_for_dhcp_broadcast(self, ifname):
-        route_cmd = "ip route add"
-        return shellutil.run("{0} 255.255.255.255 dev {1}".format(
-            route_cmd, ifname),
-                             chk_err=False)
+        try:
+            route_cmd = ["ip", "route", "add", "255.255.255.255", "dev", ifname]
+            return shellutil.run_command(route_cmd)
```
pgombar, line 1027: Why is this method even returning? The callers do nothing with the return value.

azurelinuxagent/common/osutil/default.py
```diff
@@ -1062,8 +1069,11 @@ def route_add(self, net, mask, gateway):
         """
         Add specified route 
         """
-        cmd = "ip route add {0} via {1}".format(net, gateway)
-        return shellutil.run(cmd, chk_err=False)
+        try:
+            cmd = ["ip", "route", "add", net, "via", gateway]
+            return shellutil.run_command(cmd)
```
pgombar, line 1074: Same comment here; return is not used anywhere.

azurelinuxagent/common/osutil/default.py
narrieta: yes, i noticed that but won't fix it now. there are several implementations of the method

azurelinuxagent/common/osutil/default.py
narrieta: idem

azurelinuxagent/common/osutil/default.py
pgombar: It's an easy change since none of the callers do anything with the return value, but I won't block your PR for it, up to you.

PR #1918, opened 2020-06-24
Extension telemetry pipeline

larohra:
## Description
This PR contains the major changes for enabling Extension Telemetry pipeline for Extensions. 
It adds a new thread to the Agent which comes alive every 5 mins and reads the events directory per extension and sends those extensions to Wireserver.

Issue # 

---

### PR information
- [ ] The title of the PR is clear and informative.
- [ ] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).
- [ ] Except for special cases involving multiple contributors, the PR is started from a fork of the main repository, not a branch.
- [ ] If applicable, the PR references the bug/issue that it fixes in the description.
- [ ] New Unit tests were added for the changes made and Travis.CI is passing.

### Quality of Code and Contribution Guidelines
- [ ] I have read the [contribution guidelines](https://github.com/Azure/WALinuxAgent/blob/master/.github/CONTRIBUTING.md).

PR #1960, opened 2020-07-23
Recreate handler environment file on service startup

larohra:
## Description
This PR addresses the following use cases- 
- Recreate HandlerEnvironment.json file on service startup for existing extensions. This is needed in these scenarios - 

1. When the Extension Telemetry Pipeline is enabled, the HandlerEnvironment.json file will reflect that its enabled and the existing extensions can start sending out telemetry right away.
2. If the feature is disabled in the future, we will remove the EventsFolder option from the HandlerEnvironment.json file to ensure extensions stop sending out telemetry. 
- This PR also deletes all existing events/ directories for extensions incase the feature is disabled. This is to ensure that no extension is writing data to a pre-exisiting directory in the hopes that the agent would send it eventually.

Issue # 

---

### PR information


tests/ga/test_update.py
```diff
@@ -1479,6 +1486,118 @@ def test_telemetry_heartbeat_creates_event(self, patch_add_event, patch_info, *_
         self.assertTrue(any(call_args[0] == "[HEARTBEAT] Agent {0} is running as the goal state agent {1}"
                             for call_args in patch_info.call_args), "The heartbeat was not written to the agent's log")
 
+    @contextlib.contextmanager
+    def _get_update_handler(self, iterations=1, test_data=DATA_FILE):
```
larohra, line 1490: Even though I modified an existing test class for these tests, I didnt change any of the existing tests because most of them are calling/mocking specific functions of the UpdateHandler and have the tests build around it (not to mention TestUpdate class has a huge amount of tests in it). The time investment would be huge to refactor these tests because I would have to fist understand what the test is doing and then modify the tests accordingly. In the end I chose not to do that. If you guys feel it should be done as part of this PR, I can start working on it.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1317,19 +1332,19 @@ def update_settings(self):
...
-                "configFolder": self.get_conf_dir(),
-                "statusFolder": self.get_status_dir(),
-                "heartbeatFile": self.get_heartbeat_file()
+                HandlerEnvironment.logFolder: self.get_log_dir(),
+                HandlerEnvironment.configFolder: self.get_conf_dir(),
+                HandlerEnvironment.statusFolder: self.get_status_dir(),
+                HandlerEnvironment.heartbeatFile: self.get_heartbeat_file()
             }
 
         if is_extension_telemetry_pipeline_enabled():
```
vivlingaiah, line 1341: This seems to return value of _ENABLE_EXTENSION_TELEMETRY_PIPELINE, which needs a new deployment if changed, correct ? Is it possible to dynamically change the config so that it doesn't require deployment ? (more flexibility). Just thinking

azurelinuxagent/ga/exthandlers.py
```diff
@@ -232,6 +232,16 @@ def get_exthandlers_handler(protocol):
     return ExtHandlersHandler(protocol)
 
 
+def list_agent_lib_directory(skip_agent_package=True):
+    for name in os.listdir(conf.get_lib_dir()):
+        path = os.path.join(conf.get_lib_dir(), name)
```
vivlingaiah, line 237: call conf.get_lib_dir() once and reuse ?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1317,19 +1332,19 @@ def update_settings(self):
...
-                "heartbeatFile": self.get_heartbeat_file()
+                HandlerEnvironment.logFolder: self.get_log_dir(),
+                HandlerEnvironment.configFolder: self.get_conf_dir(),
+                HandlerEnvironment.statusFolder: self.get_status_dir(),
+                HandlerEnvironment.heartbeatFile: self.get_heartbeat_file()
             }
 
         if is_extension_telemetry_pipeline_enabled():
-            handler_env["eventsFolder"] = self.get_extension_events_dir()
+            handler_env[HandlerEnvironment.eventsFolder] = self.get_extension_events_dir()
```
vivlingaiah, line 1342: this looks correct

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1515,23 +1530,14 @@ def _truncate_message(field, truncate_size=_MAX_SUBSTATUS_FIELD_LENGTH):
...
-
-    def get_heartbeat_file(self):
-        return self.data["handlerEnvironment"]["heartbeatFile"]
+    fileName = "HandlerEnvironment.json"
+    handlerEnvironment = "handlerEnvironment"
+    logFolder = "logFolder"
+    configFolder = "configFolder"
+    statusFolder = "statusFolder"
+    heartbeatFile = "heartbeatFile"
+    eventsFolder = "eventsFolder"
```
vivlingaiah, line 1540: Is this the value of the folder name or property ? Anyways, name used for the folder in Windows is "Events"

azurelinuxagent/ga/exthandlers.py
larohra: This is actually an atomic operation but good catch, this makes more sense, I'll implement this!

azurelinuxagent/ga/exthandlers.py
larohra: It was meant to be that specific and not that dynamic :)
Basically I can also add the flag to the agent config file but we didnt want to give the customers the option to turn this off as this is a very basic workflow needed by the extensions. If we do find an issue, we will turn it off from our end rather and do a successive deployment. That was the whole idea behind it.

Do you have any other ideas by any chance for making it more dynamic but at the same time not opening up the flag to the customers?

azurelinuxagent/ga/exthandlers.py
larohra: This is the name for the property. The name for the folder would be `events` (lower case) in linux too.

larohra: According to our document - 
>/var/log/azure/ExtensionName/events/ and C:\WindowsAzure\Logs\Plugin\ExtensionName \Events  

Looks like for Linux we use lower-case and for windows we use Capitalised-case. Do you want to keep this or converge the names? I personally dont mind the difference because both of them are specific to the specific OSes (like in linux the standard is to have lower-case names and in windows the standard is to have Capitalised-case.

azurelinuxagent/ga/update.py
```diff
@@ -766,6 +769,34 @@ def _send_heartbeat_telemetry(self, protocol):
...
+                                                                                                  path=path,
+                                                                                                  protocol=protocol)
+                if handler_instance is not None:
+                    # Recreate the HandlerEnvironment for existing extensions on startup.
+                    # This is to ensure that existing extensions can start using the telemetry pipeline if they support
+                    # it and also ensures that the extensions are not sending out telemetry if the Agent has to disable the feature.
+                    handler_instance.create_handler_env()
+
+            except Exception:
+                # Ignore errors if any
```
narrieta, line 787: why ignore errors here? maybe log them?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -232,6 +232,17 @@ def get_exthandlers_handler(protocol):
     return ExtHandlersHandler(protocol)
 
 
+def list_agent_lib_directory(skip_agent_package=True):
```
narrieta, line 235: suggestion: this function doesn't seem to have a well defined functionality (doesn't seem very useful by itself). You are using it in conjunction with get_ext_handler_instance_from_path_if_valid (the "if_valid" is also kind of funny).

why don't you combine both into a single function to get the extension handlers on disk? (in my opinion, such function would be more useful that the other 2).

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1317,19 +1333,19 @@ def update_settings(self):
...
+                HandlerEnvironment.heartbeatFile: self.get_heartbeat_file()
             }
 
         if is_extension_telemetry_pipeline_enabled():
-            handler_env["eventsFolder"] = self.get_extension_events_dir()
+            handler_env[HandlerEnvironment.eventsFolder] = self.get_extension_events_dir()
 
         env = [{
             "name": self.ext_handler.name,
             "version": _HANDLER_ENVIRONMENT_VERSION,
```
pgombar, line 1347: Why not parameterize these also?

azurelinuxagent/ga/update.py
```diff
@@ -766,6 +769,34 @@ def _send_heartbeat_telemetry(self, protocol):
...
+                continue
+
+        try:
+            if not is_extension_telemetry_pipeline_enabled():
+                # If extension telemetry pipeline is disabled, ensure we delete all existing extension events directory
+                # because the agent will not be listening on those events.
+                extension_event_dirs = glob.glob(os.path.join(conf.get_ext_log_dir(), "*", EVENTS_DIRECTORY))
```
pgombar, line 794: Why is `*` needed here, I thought the path for extension events is: `/var/log/azure/extension_name/events`?

pgombar: Ah, I think I got it. The `*` corresponds to any `extension_name`.

azurelinuxagent/ga/update.py
```diff
@@ -766,6 +769,34 @@ def _send_heartbeat_telemetry(self, protocol):
...
+                for ext_dir in extension_event_dirs:
+                    shutil.rmtree(ext_dir, ignore_errors=True)
+        except Exception as e:
+            logger.warn("Error when trying to delete existing Extension events directory. Error: {0}".format(ustr(e)))
```
pgombar, line 798: Just a thought: do you want to also send out telemetry while the feature is being stabilized to know what the issues are?

tests/ga/test_update.py
```diff
@@ -1479,6 +1486,118 @@ def test_telemetry_heartbeat_creates_event(self, patch_add_event, patch_info, *_
...
+                    with patch("time.sleep", side_effect=lambda _: mock_sleep(0.001)):
+                        with patch('sys.exit'):
+                            # Setup the initial number of iterations
+                            update_handler.set_iterations(iterations)
+                            try:
+                                yield update_handler, protocol
+                            finally:
+                                # Since PropertyMock requires us to mock the type(ClassName).property of the object,
+                                # reverting it back to keep the state of the test clean
+                                type(update_handler).running = True
```
pgombar, line 1531: Couldn't you also reset the iterations counter here?

tests/ga/test_update.py
```diff
@@ -1479,6 +1486,118 @@ def test_telemetry_heartbeat_creates_event(self, patch_add_event, patch_info, *_
...
+    def test_it_should_delete_extension_events_directory_if_extension_telemetry_pipeline_disabled(self):
+
+            # Disable extension telemetry pipeline and ensure events directory got deleted
+            with self._setup_test_for_ext_event_dirs_retention() as (update_handler, expected_events_dirs):
+                with patch('azurelinuxagent.ga.exthandlers._ENABLE_EXTENSION_TELEMETRY_PIPELINE', False):
+                    update_handler.run(debug=True)
+                    for ext_dir in expected_events_dirs:
+                        self.assertFalse(os.path.exists(ext_dir), "Extension directory {0} still exists!".format(ext_dir))
+
+    def test_it_should_retain_events_directories_if_extension_telemetry_pipeline_enabled(self):
```
pgombar, line 1593: nit: "extension events" instead of just "events" in the method name

azurelinuxagent/ga/update.py
larohra: Yeah exactly, its to capture every extension event dir and delete it

azurelinuxagent/ga/update.py
larohra: Yup, the error specific telemetries are added to the #1918 PR :)

tests/ga/test_update.py
larohra: This block (`finally`) is only hit when we exit the block where we initialize the update handler. The whole idea was to re-use the same object without creating a new mock_update_handler for every small run. Does that answer your question? I can explain more offline if needed

azurelinuxagent/ga/exthandlers.py
larohra: The main reason why I separated out the 2 is because I didnt want to loop through the directory again for checking and removing the orphaned packages in the `_cleanup_outdated_handlers` functions, especially since we call this every 3 secs. 

```python
            if os.path.isfile(path) and not os.path.isdir(path[0:-len(HANDLER_PKG_EXT)]):
                if not re.match(_HANDLER_PKG_PATTERN, item):
                    continue
                pkgs.append(path)
```

If you feel strongly about this though I can combine the 2 functions and then add another loop for the orphaned packages.

azurelinuxagent/ga/update.py
larohra: Good point, we should ignore any errors that we get from `get_ext_handler_instance_from_path_if_valid` function but we should log errors if we're unable to re-create hanlder_env for whatever reason. Will add the change

azurelinuxagent/ga/exthandlers.py
larohra: >(the "if_valid" is also kind of funny).

This basically meant return ext-handler-instance if its a valid handler, but I'll just rename it to something simpler.

tests/ga/test_update.py
larohra: Done

azurelinuxagent/ga/exthandlers.py
narrieta: it was just as suggestion, stroke me as a little odd, but far from "feeling strongly about it" :) ok as is

larohra: haha fair enough, thanks for the comment :)

tests/ga/test_update.py
pgombar: Got it! Thanks.

azurelinuxagent/ga/exthandlers.py
pgombar: bump

larohra: Ohh sorry missed this comment! I didnt do it because I thought there were already too many refactoring being done and these keys are pretty generic (name and version) but I guess they should be consistent. Let me quickly fix these up too.

larohra:
Successful DCR run - https://tuxgold.corp.microsoft.com/job/AzLinux/job/DungeonCrawler/job/DungeonCrawler.larohra/304/
(has 1 known issue)

PR #1969, opened 2020-07-30
Simplified changes to Unit Test Framework

kevinclark19a:
## Description

Simplified code changes that achieve basically the same testing experience as the prior PR.

Revamped the `TestExtensionUpdateOnFailure` test suite.

---

### PR information


test-requirements.txt
```diff
@@ -7,3 +7,4 @@ mock==4.0.2; python_version >= '3.6'
 distro; python_version >= '3.8'
 nose
 nose-timer; python_version >= '2.7'
+enum34; python_version < '2.7'
```
kevinclark19a, line 10: This is a python >2.6 module backported to 2.6. It contains the `Enum` class, which I'm using for convenience  in `tests.ga.extension_emulator.ExtensionCommandNames`.

larohra: This is technically a backport for Python versions < 3.4 - https://pypi.org/project/enum34/
Change it to 3.4 instead.

larohra: I'm surprised the tests passed for 2.7 given you were only installing it on Py2.6! I wonder if the py2.7 travis image already has this module installed in it

tests/ga/test_extension.py
```diff
@@ -2322,324 +2324,320 @@ def test_it_should_parse_boolean_values_encoded_as_strings(self):
...
+                    # Note that if installCommand is supposed to fail, this will erroneously raise.
+                    (first_ext, ExtensionCommandNames.ENABLE)
+                )
 
-    def test_uninstall_failed_env_variable_should_set_for_install_when_continue_on_update_failure_is_true(
-            self, *args):
-        old_handler_i = self._get_ext_handler_instance('foo', '1.0.0')
-        old_handler_i.set_handler_state(ExtHandlerState.Enabled)
-        new_handler_i = self._get_ext_handler_instance('foo', '1.0.1', continue_on_update_failure=True)
+            protocol.mock_wire_data.set_extensions_config_version("1.1.0")
```
larohra, line 2365: Shouldn't you be picking up this version from `upgraded_ext.version` instead?

tests/ga/test_extension.py
```diff
@@ -2322,324 +2324,320 @@ def test_it_should_parse_boolean_values_encoded_as_strings(self):
...
+        self.mock_sleep = patch("time.sleep", lambda *_: mock_sleep(0.0001))
+        self.mock_sleep.start()
+
+    def tearDown(self):
+        self.mock_sleep.stop()
+        AgentTestCase.tearDown(self)
 
     @staticmethod
-    def _get_ext_handler_instance(name, version, handler=None, continue_on_update_failure=False):
+    def _do_upgrade_scenario(first_ext, upgraded_ext):
```
larohra, line 2338: NIT: Rename to `_do_upgrade_scenario_and_get_order` or something moer verbose for more clarity

tests/ga/test_extension.py
```diff
@@ -2322,324 +2324,320 @@ def test_it_should_parse_boolean_values_encoded_as_strings(self):
...
+
+        invocation_record.compare(
+            (first_ext, ExtensionCommandNames.DISABLE),
+            (second_ext, ExtensionCommandNames.UPDATE),
+            (first_ext, ExtensionCommandNames.UNINSTALL),
+            (second_ext, ExtensionCommandNames.INSTALL),
+            (second_ext, ExtensionCommandNames.ENABLE)
+        )
+
+        _, kwargs = second_ext.actions[ExtensionCommandNames.UPDATE].call_args
```
larohra, line 2405: From the ExtensionEmulator class, `self.actions` is a dict of actions. And according to `Actions` class, the functions are just plain old objects, not Mocks. 

```python
self.actions = {
            ExtensionCommandNames.INSTALL: install_action,
            ExtensionCommandNames.UNINSTALL: uninstall_action,
            ExtensionCommandNames.UPDATE: update_action,
            ExtensionCommandNames.ENABLE: enable_action,
            ExtensionCommandNames.DISABLE: disable_action
        }

class Actions(object):
    """
    A collection of static methods providing some basic functionality for the ExtensionEmulator
    class' actions.
    """

    @staticmethod
    def succeed_action(*args, **kwargs):
        """
        A nop action with the correct function signature for ExtensionEmulator actions.
        """
        return 0
```

In the patched_popen, you're just calling the `Action` of the emulator directly with whatever arguments are being passed to it - 
```python
return next(
                emulator.actions[ExtensionCommandNames(command_name)]
                for emulator in emulators
                if emulator.matches(ext_name, ext_version)
            )(cmd, *args, **kwargs)
```

My main question is how is this line working if this is not a mock? I feel like I'm missing something
```python
_, kwargs = second_ext.actions[ExtensionCommandNames.UPDATE].call_args
```

tests/ga/test_extension.py
```diff
@@ -2322,324 +2324,320 @@ def test_it_should_parse_boolean_values_encoded_as_strings(self):
...
+
+        invocation_record.compare(
+            (first_ext, ExtensionCommandNames.DISABLE),
+            (second_ext, ExtensionCommandNames.UPDATE),
+            (first_ext, ExtensionCommandNames.UNINSTALL),
+            (second_ext, ExtensionCommandNames.INSTALL),
+            (second_ext, ExtensionCommandNames.ENABLE)
+        )
+
+        _, kwargs = second_ext.actions[ExtensionCommandNames.INSTALL].call_args
```
larohra, line 2427: A suggestion, you could also pass a lambda as an action to verify if a certain env variable is set. That way you will have one less state to maintain in your Mocks.

For eg: 
```python
install_action = lambda *args, *kwargs: return 0 if "Uninstall_failed" in args['env'] else 1 
second_ext = extension_emulator(version="1.1.0", continue_on_update_failure=True, install_action=install_action)
```

Something like this (this exact example might not work exactly as it, just wanted to illustrate my point)

tests/ga/test_extension.py
```diff
@@ -48,7 +48,9 @@
 from tests.protocol.mocks import mock_wire_protocol, HttpRequestPredicates
 from tests.protocol.mockwiredata import DATA_FILE
 from tests.tools import AgentTestCase, data_dir, i_am_root, MagicMock, Mock, patch
-
+from tests.ga.extension_emulator import Actions, ExtensionCommandNames, extension_emulator, \
+    enable_invocations, generate_put_handler
```
larohra, line 52: I see some unused imports in this file (not related to your changes). Could you please clean those up too since you're modifying this code anyways. Thanks!

test-requirements.txt
kevinclark19a: Hm. That's odd. At least for 2.6, travis doesn't have this module pre installed-- the 2.6 test actually failed before adding this line. (and passed for 2.7+. hence why I only added this dependency for 2.7!)

The docs do very obviously state that enum is "New in version 3.4", so I'll definitely make this change.

tests/ga/extension_emulator.py
```diff
@@ -0,0 +1,315 @@
...
+        """
+        return 0
+    
+    
+    @staticmethod
+    def fail_action(*args, **kwargs):
+        """
+        A simple fail action with the correct function signature for ExtensionEmulator actions.
+        """
+        raise ExtensionError("FailAction called.")
```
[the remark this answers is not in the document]
kevinclark19a: My thoughts would be delayed parsing; I don't think any tests in the test suite actually use this method; instead, they use `generate_unique_fail` (below). Do you think a better solution would be to remove this all together? 

Nice catch, btw.

tests/ga/extension_emulator.py
```diff
@@ -0,0 +1,315 @@
...
+
+class InvocationRecord:
+
+    def __init__(self):
+        self._queue = []
+
+    def add(self, ext_name, ext_ver, ext_cmd):
+        self._queue.append((ext_name, ext_ver, ext_cmd))
+
+    def compare(self, *cmds):
```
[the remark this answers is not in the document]
kevinclark19a: Never actually seen "NIT." Is this like in the phrase "nit pick"?

Regardless, I agree that this rename should happen.

tests/ga/extension_emulator.py
```diff
@@ -0,0 +1,315 @@
...
+    return patched_popen
+
+def generate_mock_load_manifest(*emulators):
+
+    original_load_manifest = ExtHandlerInstance.load_manifest
+
+    def mock_load_manifest(self):
+
+        try:
+            matching_emulator = next(
```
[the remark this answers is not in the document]
kevinclark19a: I can see how this could lead to clarity issues. Your assumption is correct: `next` returns the "next" element in a list, which will always be the first matching emulator-- subsequent calls will not iterate the same list.

I personally think a better name for this function would be `first`, but I guess whomever wrote the builtins disagreed... a helper function would certainly help with this.

tests/ga/test_extension.py
kevinclark19a: To clarify: the intent here is to make explicit the returned value, correct?

tests/ga/test_extension.py
kevinclark19a: You're right. Good catch.

tests/ga/test_extension.py
kevinclark19a: Sorry, this is definitely something that could be more explicit in the code, looking back over it.

The thing you're missing is the private function `_extend_func` in `extension_emulator.py`. This is called within the factory method (`extension_emulator` in that same file), and wraps not only the result of the passed function in a mock popen object, but also the function itself in a mock. It's near the bottom of the page, if I remember correctly.

I think the lack of clarity is actually pointing to a issue with how the code is written, though; I don't think it's ideal to call `_extend_func` within only the factory method. Do you have any thoughts on alternatives?

Could it be as simple as moving the `_extend_func` call into the `ExtensionEmulator` class and better documenting it? I think it would be nice if it was relatively opaque to the user of the test (but clear enough if someone is to dive into the `ExtensionEmulator` class), because that way, a regular lambda can be used, and the test writer wouldn't have to care about the necessary massaging to get that lambda into the correct format.

tests/ga/test_extension.py
kevinclark19a: This could maybe work. I'd love to sync with you offline to discuss this further-- I'm mostly concerned that this method would prematurely exit the `run` call, and the ramifications that might have in our coverage...

test-requirements.txt
```diff
@@ -7,3 +7,4 @@ mock==4.0.2; python_version >= '3.6'
 distro; python_version >= '3.8'
 nose
 nose-timer; python_version >= '2.7'
+enum34; python_version < '3.4'
```
narrieta, line 10: where is this being used?

kevinclark19a: `tests.ga.extension_emulator.ExtensionCommandName` inherits from `enum.Enum`.

larohra: +1 to @pgombar's comment. If its not adding too much extra benefit, might as well skip it

tests/ga/extension_emulator.py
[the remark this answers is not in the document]
larohra: I see this is being used by this test - `test_non_enabled_ext_should_not_be_disabled_at_ver_update` but yeah if its not really needed then might as well remove it.

tests/ga/extension_emulator.py
```diff
@@ -114,6 +113,8 @@ def generate_put_handler(*emulators):
     For use with tests.protocol.mocks.mock_wire_protocol.
     """
 
+    first_matching_emulator = lambda matches_func: next(emulator for emulator in emulators if matches_func(emulator))
```
larohra, line 116: I see multiple copies of this helper function. Why not make it a private function of the `ExtensionEmulator` class and call that from everywhere. The uber goal is to avoid code duplication as much as possible

tests/ga/extension_emulator.py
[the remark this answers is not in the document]
larohra: It is exactly that, like its not really functional but better cosmetically lol.
I think this is open source PR lingo, I picked it up from comments to my PR and then it just kinda caught on haha.

tests/ga/extension_emulator.py
```diff
@@ -0,0 +1,315 @@
...
+        self._queue = []
+
+    def add(self, ext_name, ext_ver, ext_cmd):
+        self._queue.append((ext_name, ext_ver, ext_cmd))
+
+    def compare(self, *cmds):
+        """
+        Verifies that any and all recorded invocations appear in the provided command list in that exact ordering.
+
+        Each cmd in cmds should be a tuple of the form (ExtensionEmulator, ExtensionCommandNames).
```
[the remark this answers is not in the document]
larohra: I think you missed this one (from what I understood in the code, it is expected it to be a single command and not a list of commands)

tests/ga/test_extension.py
larohra: Yes that is correct, the function name should be explicit enough to tell you exactly what its doing. (Feel free to choose another name if you like too, I'm not that great with names lol)

tests/ga/test_extension.py
larohra: Ahh I see it now! I definitely missed that one lol. And yes I agree with your idea, it would be better to add it to the `ExtensionEmulator` class and make the `_extend_func` an internal lambda so that anyone reading the code can understand what's happening there. 
Also while you're at it, make sure to rename the function to something more specific like - `_extend_action_to_popen_mock`  or something.

tests/ga/test_extension.py
larohra: Sure let's sync about it. I dont understand exactly what you mean by premature exit, Maybe you could explain more over in our sync

tests/ga/extension_emulator.py
```diff
@@ -18,22 +18,21 @@
...
 from tests.tools import Mock, patch
 
 from tests.protocol.mocks import HttpRequestPredicates
 
-from azurelinuxagent.ga.exthandlers import ExtHandlerInstance, HandlerManifest
+from azurelinuxagent.ga.exthandlers import ExtHandlerInstance
 
 
-class ExtensionCommandNames(Enum):
+class ExtensionCommandName(Enum):
```
larohra, line 35: Wait I think you misread my comment @kevinclark19a. My comment was to make the comment in Line 152 singular as compared to making the whole class name singular lol. The class name should be plural, and now I understand what exactly you meant in the comment in line 152. Let me add another comment there.

tests/ga/extension_emulator.py
```diff
@@ -145,25 +144,26 @@ def __init__(self):
     def add(self, ext_name, ext_ver, ext_cmd):
         self._queue.append((ext_name, ext_ver, ext_cmd))
 
-    def compare(self, *cmds):
+    def compare(self, *expected_cmds):
         """
         Verifies that any and all recorded invocations appear in the provided command list in that exact ordering.
 
-        Each cmd in cmds should be a tuple of the form (ExtensionEmulator, ExtensionCommandNames).
+        Each cmd in expected_cmds should be a tuple of the form (ExtensionEmulator, ExtensionCommandNames).
```
larohra, line 151: NIT: Rename to - 
`Each cmd in expected_cmds should be a tuple of the form (ExtensionEmulator, ExtensionCommandNames object).`

tests/ga/extension_emulator.py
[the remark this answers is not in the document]
larohra: Nevermind this, I saw what you did. Left another comment, will resolve this one :)

test-requirements.txt
kevinclark19a: I was at one point using the enum features to iterate over the enum values, something that is hacky at best without an enum, and that I figured would be worth a test import.

However, I'm pretty sure I'm not actually doing that anymore; I'm going to try reworking this to not have the import, and, if it works, will have no argument for keeping this!

tests/ga/extension_emulator.py
```diff
@@ -0,0 +1,312 @@
...
+from azurelinuxagent.common.future import ustr
+from azurelinuxagent.common.exception import ExtensionError
+from tests.tools import Mock, patch
+
+from tests.protocol.mocks import HttpRequestPredicates
+
+from azurelinuxagent.ga.exthandlers import ExtHandlerInstance
+
+
+class ExtensionCommandName(Enum):
```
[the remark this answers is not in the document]
kevinclark19a: I think you're probably right; I'll try removing the dependency and see how much everything breaks.
I provide a little more context in my reply above.

tests/ga/extension_emulator.py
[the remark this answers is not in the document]
kevinclark19a: Oh, interesting. The "name not found" error was basically acting as a stand in for the `ExtensionError`, which is why this was working.

I think it would make sense to remove this, if you guys agree.

tests/ga/extension_emulator.py
```diff
@@ -0,0 +1,312 @@
...
+    Create a HTTP handler to store status blobs for each provided emulator.
+    For use with tests.protocol.mocks.mock_wire_protocol.
+    """
+
+    first_matching_emulator = lambda matches_func: next(emulator for emulator in emulators if matches_func(emulator))
+
+    def mock_put_handler(url, *args, **kwargs):
+
+        if HttpRequestPredicates.is_host_plugin_status_request(url):
+            return None
```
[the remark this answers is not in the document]
kevinclark19a: So the `host_plugin_status_request`s are actually encrypted from what I remember, and aren't what we want anyways.
The `mock_wire_protocol` function this is intended to be passed into checks if the result is `None` and, if so, falls back to its own method.

Basically the `return None` here serves as a fallback to whatever the `mock_wire_protocol` does by default, from my understanding.

tests/ga/extension_emulator.py
kevinclark19a: Is this supposed to say `...ExtensionCommandName object...`?

((Note the lack of an s))

tests/ga/test_extension.py
```diff
@@ -2322,324 +2321,320 @@ def test_it_should_parse_boolean_values_encoded_as_strings(self):
...
-            msg = str(error.exception)
-            self.assertIn("Disable Failed", msg, "Update should fail with Disable Failed error")
-            self.assertIn("ExtensionError", msg, "The Exception should initially be propagated as ExtensionError")
+        invocation_record = TestExtensionUpdateOnFailure._do_upgrade_scenario_and_get_order(first_ext, second_ext)
+
+        invocation_record.compare(
+            (first_ext, ExtensionCommandName.DISABLE)
+        )
+
+        self.assertEqual(len(second_ext.status_blobs), 1, "The second extension should have a single submitted status.")
```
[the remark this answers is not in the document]
kevinclark19a: I'm pretty sure you're right about that.
I don't know if that behavior falls into the scope of this particular test, but I don't really have any strong feelings about it; it would be easy enough to add, anyway, if that's what we want to do.

tests/ga/extension_emulator.py
larohra: No like ExtensionCommandNames  (as in the name of the class) and object. Essentially `ExtensionCommandNames object` would mean an object of the ExtensionCommandNames class. Does that make sense? I'm confused too now lol

tests/ga/extension_emulator.py
```diff
@@ -32,7 +31,7 @@
 from azurelinuxagent.ga.exthandlers import ExtHandlerInstance
 
 
-class ExtensionCommandName(Enum):
+class ExtensionCommandNames:
```
larohra, line 34: NIT: Dont you need to inherit it from Object (like `ExtensionCommandNames(object)`). 
Not sure if this is needed for py2 vs py3, but we follow this convention for all our classes, it would be cleaner if you could just follow it here too just to be consistent with the other code. Thanks!

tests/ga/test_extension.py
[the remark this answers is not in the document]
larohra: If its easy to add, might as well add it. Its always good to have more testing rather than less testing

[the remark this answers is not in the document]
pgombar: Bump, what was the decision?
The current tests don't have this check, so I'm okay if you don't want to add it in this PR.

tests/ga/test_extension.py
pgombar: Not sure if I follow, did you guys sync about this?

kevinclark19a: We did! we ultimately decided that both ways of doing things (i.e. the way in the code, and the way @larohra suggested) are functionally the same for this particular use case, but that there could conceivably be a scenario which would better fit the suggested code.

tests/ga/extension_emulator.py
kevinclark19a: This is annoying, because SOP for py2 is to inherit directly from `object`, but SOP for py3 is to _not_ directly inherit from `object`. 

Regardless, internal consistency is more important!

tests/ga/test_extension.py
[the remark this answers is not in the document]
kevinclark19a: This check was added! The first extension does actually report for `install`/`enable` command success, but doesn't afterward. That's why we expect a single status.

tests/ga/extension_emulator.py
```diff
@@ -197,6 +199,27 @@ def _extend_func(func):
...
+
+                match = re.fullmatch(regex, config_file)
+                if not match:
+                    continue
+
+                if seq < int(match.group("seq")):
+                    seq = int(match.group("seq"))
+
+            fileutil.write_file(os.path.join(os.path.dirname(cmd), "status", "{seq}.status".format(seq=seq)),
+                json.dumps([{ "status": {"status": "success"} }]))
```
larohra, line 221: NIT: If you always report success, there might be a scenario where the status being reported is success but the process failed with some error. Might be worth making this a bit more strict to report success only if return_code == 0 else "failed" or something.

tests/ga/test_extension.py
```diff
@@ -2461,6 +2461,7 @@ def test_extension_error_should_be_raised_when_continue_on_update_failure_is_fal
             (first_ext, ExtensionCommandNames.DISABLE)
         )
 
+        self.assertEqual(len(first_ext.status_blobs), 1, "The first extension should not have submitted a second status.")
```
larohra, line 2461: A follow up on my previous comment, might also be worth adding if the status was actually successful rather than just checking how many times it was reported. This goes for the `second_ext.status_blobs` too.

tests/ga/extension_emulator.py
narrieta: did we decide not to address this comment?

kevinclark19a: I looked into this, but forgot to document what I found here.

Basically, the object written to file here is only ever sent to the `HTTP PUT /status` endpoint when there is not some sort of failure. In other words, it doesn't seem possible that `return_code == 0` and that status reporting success is conveyed at the same time.

A perfect example of this is the test `tests.ga.test_extension.test_extension_error_should_be_raised_when_continue_on_update_failure_is_true_on_disable_and_update_failure`, wherein update fails and the only status reported for that extension is the exit code we outputted (instead of the `status: success` stuff). This is verified by an `assertTrue` within the test, and I went back and stepped through the code, and it is in fact the case.

larohra: >wherein update fails and the only status reported for that extension is the exit code we outputted (instead of the status: success stuff).

There are multiple levels to our status report, we have the HandlerStatus and ExtensionSettings settings. The HandlerStatus failure is set independently of the extension status. I think since you aren't checking the individual setting status of any extension you will probably not see this. Even though this might not be needed scenario for your changes, I'm wondering if we'd need this later (since you are creating a new framework).

kevinclark19a: Ah, that sounds like it very well might be the case. Do you mind pointing me to where in `exthandlers.py` we do the reporting of the extension status (i.e., the status always being reported as `success` here)?

I'd love to take a look at it. If this other status is submitted somewhere and has a use-case, I agree the code here should be flexible; even though I don't think it would be particularly tricky to add the if after the fact (when/if we run into the need for it), I don't have any problems with adding it in this PR if there is a use-case we'll want to test for.

tests/ga/test_extension.py
kevinclark19a: I think that `second_ext.status_blobs` is in fact checking for the error code returned by `disable`; a visual inspection (within the debugger) shows that this is the only status reported to the `HTTP PUT /status` endpoint.

tests/ga/extension_emulator.py
larohra: This function `report_ext_status` in exthandlers.py does that parsing. You can follow the thread here

In your test code, in line: 121 in `extension_emulator.py`, I think you're getting the handler_status. If you debug the whole `aggregateState` that you pull from the status blob, you should see the extension status too (not just the handler status).

I can walk the debug with you if you want if it gets too confusing, let me know

larohra: >If this other status is submitted somewhere and has a use-case

There might be instances where we need this, like if we want to verify the extension status for a specific sequence no for that extension. Given that you are creating a cool new framework, this would definitely be helpful

tests/ga/extension_emulator.py
```diff
@@ -216,9 +219,16 @@ def wrapped_func(cmd, *args, **kwargs):
...
                 if seq < int(match.group("seq")):
                     seq = int(match.group("seq"))
+            
+            status_file = os.path.join(os.path.dirname(cmd), "status", "{seq}.status".format(seq=seq))
+            
+            if return_value == 0:
+                status_contents = ExtensionEmulator._EXT_STATUS_SUCCESS
+            else:
+                status_contents = ExtensionEmulator._EXT_STATUS_ERROR_BASE.copy()
+                status_contents[0]["status"]["substatus"] = return_value
```
larohra, line 229: Thanks for adding this! A small request, can you make the `substatus` a json instead with something like - 
`{'exit_code': return_value}` just to make it more clearer and easier to understand and to eventually test.

PR #1976, opened 2020-08-11
Enabled linter on loose files in common.

kevinclark19a:
## Description

2nd of 7 PRs to enable linter.

Error Codes can be found here: http://pylint.pycqa.org/en/latest/technical_reference/features.html

---

### PR information


azurelinuxagent/common/AgentGlobals.py
```diff
@@ -1,4 +1,4 @@
-# Microsoft Azure Linux Agent
+# Microsoft Azure Linux Agent # pylint: disable=C0103
```
narrieta, line 1: should this one be here? it's within a comment

azurelinuxagent/common/conf.py
```diff
@@ -19,11 +19,11 @@
 
 """
 Module conf loads and parses configuration file
-"""
+""" # pylint: disable=W0105
```
narrieta, line 22: does this affect the doctring?

azurelinuxagent/common/future.py
```diff
@@ -6,38 +6,40 @@
...
 
 if sys.version_info[0] == 3:
-    import http.client as httpclient
-    from urllib.parse import urlparse
+    import http.client as httpclient # pylint: disable=W0611,import-error
+    from urllib.parse import urlparse # pylint: disable=W0611,import-error,no-name-in-module
 
-    """Rename Python3 str to ustr"""
-    ustr = str
+    """Rename Python3 str to ustr""" # pylint: disable=W0105
```
narrieta, line 23: ok with doctring?

azurelinuxagent/common/future.py
```diff
@@ -6,38 +6,40 @@
...
 
 elif sys.version_info[0] == 2:
-    import httplib as httpclient
-    from urlparse import urlparse
+    import httplib as httpclient # pylint: disable=E0401,W0611
+    from urlparse import urlparse # pylint: disable=E0401
 
-    """Rename Python2 unicode to ustr"""
-    ustr = unicode
+    """Rename Python2 unicode to ustr""" # pylint: disable=W0105
```
narrieta, line 34: docstring

azurelinuxagent/common/logger.py
```diff
@@ -145,7 +145,7 @@ def write_log(log_appender):
         for appender in self.appenders:
             appender.write(level, log_item)
             #
-            # TODO: we should actually call
+            # TODO: we should actually call # pylint: disable=W0511
```
narrieta, line 148: ok within comment?

azurelinuxagent/common/logger.py
```diff
@@ -158,15 +158,15 @@ def write_log(log_appender):
             for appender in self.logger.appenders:
                 appender.write(level, log_item)
                 #
-                # TODO: call write_log instead (see comment above)
+                # TODO: call write_log instead (see comment above) # pylint: disable=W0511
```
narrieta, line 161: comment

azurelinuxagent/common/AgentGlobals.py
kevinclark19a: Ah. This in particular might be problematic.
I think the exact error this comment is suppressing is that the module `AgentGlobals` does not follow correct naming conventions.

Though this does in fact disable that error (showing that pylint happily accepts this command within a comment), any additional `invalid-name<C0103>` errors later within this file would also be disabled, which is less than ideal...

I do not see a work around here though.

Note: `<code> # pylint: disable=<some_err> # pylint: disable=<some_other_err>` actually only disables `<some_other_err>` (and not also `<some_err>`).  This is obviously something we have to be weary of, but it makes me think that pylint is simply looking at the end of each line for this pattern.

azurelinuxagent/common/conf.py
kevinclark19a: The error it is suppressing is:
`# pointless-string-statement<W0105>: (hi-pri) Used when a string is used as a statement (which of course has no effect). This is a particular case of W0104 with its own message so you can easily disable it if you're using those strings as documentation, instead of comments.`

I'm not quite sure what you mean by "affect[ing] the docstring," though.

azurelinuxagent/common/logger.py
kevinclark19a: See the comment I left above.

2.7.pylintrc
```diff
@@ -100,8 +100,6 @@ disable=bad-option-value, # pylint does not recognize the error code/symbol (nee
     useless-object-inheritance, # Used when a class inherit from object, which under python3 is implicit (3+ only)
     duplicate-code # Indicates that a set of similar lines has been detected among multiple file. TODO: this is on our "fix" list, and is thus left last.
 
-
-# This section controls the duplicate-code error.
```
narrieta, line 104: we shouldn't remove this comment

azurelinuxagent/common/AgentGlobals.py
narrieta: OK, we can leave it as is for now. This will be one easy to fix.

azurelinuxagent/common/conf.py
narrieta: I wasn't sure if the doctring can have an inline comment. I just double-checked it can,

BTW - This one may be just easier to fix (make this a comment instead of a docstring) than to suppress and fix later

2.7.pylintrc
kevinclark19a: Sorry, this was a mistake with the merging.

azurelinuxagent/common/future.py
narrieta, line 23: might as well fix it now (comment instead of docstring)

azurelinuxagent/common/future.py
narrieta: let's fix it

azurelinuxagent/common/conf.py
kevinclark19a: all of these `W0105` were auto-marked by the script I wrote, which is part of the reason why they weren't fixed.

If you want, I can go through and fix them. Is this potentially beyond the scope of this PR, though? Since there are so many changes, I'd prefer to limit the number of changes that aren't just the insertion of a comment.

narrieta: sure, you can leave them as is.

narrieta (dismissed):

.travis.yml
```diff
@@ -23,25 +23,25 @@ matrix:
         - >-
           NOSEOPTS="--verbose --with-timer" SETUPOPTS=""
           PYLINTOPTS="--rcfile=2.7.pylintrc"
-          PYLINTFILES="azurelinuxagent/daemon azurelinuxagent/agent.py azurelinuxagent/ga setup.py makepkg.py"
+          PYLINTFILES="--ignore=osutil,protocol,utils azurelinuxagent/common/ azurelinuxagent/daemon azurelinuxagent/agent.py azurelinuxagent/ga setup.py makepkg.py"
```
larohra, line 26: Why the ignore?

kevinclark19a: This PR was supposed to be _just_ the loose files in common, not the other folders. Somehow, they made it into the PR, though. I can go ahead and remove the ignore.

PR #1977, opened 2020-08-11
Enabled linter for azurelinuxagent.common.osutil module.

kevinclark19a:
## Description

3rd of 7 PRs enabling linter in Travis.

Error Codes can be found here: http://pylint.pycqa.org/en/latest/technical_reference/features.html

---

### PR information


2.7.pylintrc
```diff
@@ -100,8 +100,6 @@ disable=bad-option-value, # pylint does not recognize the error code/symbol (nee
     useless-object-inheritance, # Used when a class inherit from object, which under python3 is implicit (3+ only)
     duplicate-code # Indicates that a set of similar lines has been detected among multiple file. TODO: this is on our "fix" list, and is thus left last.
 
-
```
narrieta, line 103: merge issue?

azurelinuxagent/common/osutil/default.py
```diff
@@ -31,15 +31,18 @@
 import struct
 import sys
 import time
-from pwd import getpwall
+from pwd import getpwall # pylint: disable=ungrouped-imports
 
 import array
 
+#pylint: disable=R0801
```
larohra, line 38: Why is this error disabled only for a few set of imports? Also I couldn't find a link for R0801 in the docs you added. Could you please add a link for it too, thanks!

azurelinuxagent/common/osutil/default.py
```diff
@@ -31,15 +31,18 @@
...
 import azurelinuxagent.common.conf as conf
 import azurelinuxagent.common.logger as logger
 import azurelinuxagent.common.utils.fileutil as fileutil
 import azurelinuxagent.common.utils.shellutil as shellutil
 import azurelinuxagent.common.utils.textutil as textutil
+#pylint: enable=R0801
+
```
larohra, line 45: NIT: Extra line

azurelinuxagent/common/osutil/default.py
larohra: Also NIT: no space between # and pylint

kevinclark19a: The symbol for this error code is `duplicate-code`. The reason why we have this within the codebase but not within the comments at the beginning of `*.pylintrc` is because I ended up suppressing `duplicate-code` globally. That we want to eventually fix this is captured within a `TODO` beside `duplicate-code` within the disabled section, and the links at the beginning of both files do in fact have an entry for `duplicate-code (R0801)`.

I agree, though, that having a line-item with the others at the start of the `*.pylintrc` files would be ideal.

Let's figure out the next PR we want approved, so I can make the change just one time, there.

larohra: so if this code is suppressed globally we dont need this disable/enable here then do we?

kevinclark19a: Not for suppression, no.

However, I would assume we want want to leave this comment in so we know where to fix it.

larohra: But to fix it we would need to remove the global filter right? And once that is removed this line will start throwing errors in the linter right? Wouldn't that make this redundant?

kevinclark19a: Yeah, that's true. I think just searching in the source code is preferred to running the linter and picking out the filename and line numbers one by one, though.

I feel like that's a good enough value add to keep these.

larohra: I see. Is this the only place that particular error is raised? If not then you should add the same suppress line in every line that error would be raised by the linter if it was not globally suppressed. Whatever route you take at the end of the day, it should be consistent :)

kevinclark19a: Right, everything is in fact consistent!

larohra: So if this is the only place where its used, then why even add it to the global filter?

kevinclark19a: It is not the only place it is used, sorry.

There are other places, and they are also marked. There seems to be a bug with `duplicate-code` that makes inline error suppression not work, hence the global disabling as well.

larohra: Ahh alright, that makes sense! Thanks for adding it! I think just a description for this in the linters.rc file would be all I need. :)

PR #1987, opened 2020-08-20
Add log collection tool and thread

pgombar:
## Description

This work is the continuation of #1902 and #1847. In this PR there are two main updates:
1) Command-line utility for collecting logs is added.
2) A thread is added that periodically calls the command-line utility to collect logs and upload them to the host. The log collection is invoked with `systemd-run` and is ran with CPU and memory limits.

Note that the feature is disabled by default for now.

---

### PR information


azurelinuxagent/agent.py
```diff
@@ -151,6 +151,23 @@ def show_configuration(self):
         for k in sorted(configuration.keys()):
             print("{0} = {1}".format(k, configuration[k]))
 
+    def collect_logs(self, log_collector_mode):
+        from azurelinuxagent.common.logcollector import LogCollector, OUTPUT_RESULTS_FILE_PATH
+        if log_collector_mode == "full":
```
kevinclark19a, line 156: Would it be useful to make this string a constant? I'm mainly concerned with additional modes being added and the potential therein to introduce typos, especially because this string is repeated down below (line 162).

azurelinuxagent/agent.py
```diff
@@ -151,6 +151,23 @@ def show_configuration(self):
...
+            print("Running log collector mode full")
+        else:
+            print("Running log collector mode normal")
+
+        try:
+            lc = LogCollector(log_collector_mode == "full")
```
kevinclark19a, line 162: In the vein of the above comment, is it worth passing a constant instead of a bool here?

azurelinuxagent/agent.py
```diff
@@ -159,7 +176,7 @@ def main(args=[]): # pylint: disable=R0912,W0102
     """
     if len(args) <= 0: # pylint: disable=len-as-condition
         args = sys.argv[1:]
-    command, force, verbose, debug, conf_file_path = parse_args(args)
+    command, force, verbose, debug, conf_file_path, log_collector_mode = parse_args(args)
```
kevinclark19a, line 179: This seems like we're returning a lot of arguments, and might need to return more later on (with the addition of more options). Would the return type be better as a specific class, or maybe a dict?

azurelinuxagent/agent.py
```diff
@@ -197,45 +217,58 @@ def parse_args(sys_args): # pylint: disable=R0912
     verbose = False
     debug = False
     conf_file_path = None
-    for a in sys_args: # pylint: disable=C0103
-        m = re.match("^(?:[-/]*)configuration-path:([\w/\.\-_]+)", a) # pylint: disable=W1401,C0103
+    log_collector_mode = None
+
+    for arg in sys_args: # pylint: disable=C0103
```
kevinclark19a, line 222: I believe the `C0103` error suppression was because the variable name (`a`) doesn't conform to PEP. I think we can remove it now!

azurelinuxagent/agent.py
```diff
@@ -197,45 +217,58 @@ def parse_args(sys_args): # pylint: disable=R0912
...
             conf_file_path = m.group(1)
             if not os.path.exists(conf_file_path):
                 print("Error: Configuration file {0} does not exist".format(
                         conf_file_path), file=sys.stderr) # pylint: disable=C0330
                 usage()
                 sys.exit(1)
-        
-        elif re.match("^([-/]*)deprovision\\+user", a):
+
+        m = re.match("^(?:[-/]*)-mode:([\w/\.\-_]+)", arg)
```
kevinclark19a, line 232: The linter is complaining here because `\w` is not a valid escape. I think this should be a `r"..."` string.

azurelinuxagent/agent.py
```diff
@@ -258,11 +292,12 @@ def usage():
     s += ("usage: {0} [-verbose] [-force] [-help] " # pylint: disable=C0103
            "-configuration-path:<path to configuration file>" # pylint: disable=C0330
            "-deprovision[+user]|-register-service|-version|-daemon|-start|"
-           "-run-exthandlers|-show-configuration]"
+           "-run-exthandlers|-show-configuration|-collect-logs [-mode:full]"
```
kevinclark19a, line 295: I can't quite tell what's happening here; it seems you've added _just_ an open straight brace, and the corresponding close brace was already there. 

Do we now have mismatched braces? Or did we just have them before? I can't find the original open brace, so I'm leaning towards the second, but I just want to be sure!

tests/ga/test_collect_logs.py
```diff
@@ -0,0 +1,187 @@
...
+
+    @patch("azurelinuxagent.ga.collect_logs.CollectLogsHandler.collect_and_send_logs")
+    def test_it_should_only_collect_logs_if_conditions_are_met(self, patch_collect_and_send_logs):
+        # In order to collect logs, three conditions have to be met:
+        # 1) the flag must be set in the conf file
+        # 2) systemd must be managing services
+        # 3) python version 2.7+ which is automatically true for these tests since they are disabled on py2.6
+
+        with _create_collect_logs_handler() as collect_logs_handler:
+            with patch("azurelinuxagent.ga.collect_logs.os.path.exists", return_value=False):
```
kevinclark19a, line 110: I'm worried about the scope of this mock; if any other code branches happen to call `os.path.exists` on an unrelated file, could this patch cause issues?

Maybe making a inner function that only special cases the exact file we're looking for, and passing the rest on to the original function?

azurelinuxagent/agent.py
pgombar: It would be cleaner, but I will leave that change out from this PR since it's tangential to the feature. I also don't really expect this to grow in the future, to be honest.

azurelinuxagent/agent.py
pgombar: Nice, will do!

azurelinuxagent/agent.py
pgombar: Looking at the current version of this code in develop, it seems like we did have mismatched braces (the closing one `]`). Not that big of a deal luckily because it was just an example string of various command-line options, but now it should be fixed.

azurelinuxagent/common/logcollector.py
```diff
@@ -339,26 +335,35 @@ def collect_logs(self):
...
+            _LOGGER.info("Starting log collection at {0}".format(start_time.strftime("%Y-%m-%dT%H:%M:%SZ")))
 
             files_to_collect = self._create_list_of_files_to_collect()
             _LOGGER.info("### Creating compressed archive ###")
 
-            with zipfile.ZipFile(_COMPRESSED_ARCHIVE_PATH, "w", compression=zipfile.ZIP_DEFLATED) as compressed_archive:
-                for file in files_to_collect:
-                    archive_file_name = LogCollector._convert_file_name_to_archive_name(file)
-                    compressed_archive.write(file, arcname=archive_file_name)
+            with zipfile.ZipFile(COMPRESSED_ARCHIVE_PATH, "w", compression=zipfile.ZIP_DEFLATED) as compressed_archive:
```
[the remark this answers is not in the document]
pgombar: Thanks for the analysis! Since we are first targeting VMs with systemd (in order to enable resource limiting), none of them actually run py2.6, so I will take this change when we are expanding to non-systemd VMs. This will be useful for sure!

tests/ga/test_collect_logs.py
pgombar: The scope of the mock is just the `collect_logs` module. So, if any other code branches outside of that module hits `os.path.exists`, it would return the actual function, not the mock. However, I like your suggestion of scoping down the mock to the specific file, it will be clearer what the mock is trying to achieve. I will make the change, thanks!

kevinclark19a (dismissed):

azurelinuxagent/agent.py
```diff
@@ -151,6 +151,23 @@ def show_configuration(self):
         for k in sorted(configuration.keys()):
             print("{0} = {1}".format(k, configuration[k]))
 
+    def collect_logs(self, log_collector_mode):
+        from azurelinuxagent.common.logcollector import LogCollector, OUTPUT_RESULTS_FILE_PATH
```
larohra, line 155: NIT: Avoid local importing if possible

azurelinuxagent/agent.py
larohra: +1 to this, it would be helpful to have an Enum or something here just to make this a bit more clearer as to what this is doing and what the multiple options are.

azurelinuxagent/agent.py
```diff
@@ -197,45 +217,58 @@ def parse_args(sys_args): # pylint: disable=R0912
...
             conf_file_path = m.group(1)
             if not os.path.exists(conf_file_path):
                 print("Error: Configuration file {0} does not exist".format(
                         conf_file_path), file=sys.stderr) 
                 usage()
                 sys.exit(1)
-        
-        elif re.match("^([-/]*)deprovision\\+user", a):
+
+        m = re.match("^(?:[-/]*)-mode:([\w/\.\-_]+)", arg) # pylint: disable=W1401,C0103
```
larohra, line 232: NIT: Rename to `log-collector-mode` maybe? Or something more verbose than just `mode` to make it clearer its for the log collection

azurelinuxagent/agent.py
```diff
@@ -197,45 +217,58 @@ def parse_args(sys_args): # pylint: disable=R0912
...
+        if m is not None:
+            log_collector_mode = m.group(1)
+            if log_collector_mode != "full":
```
larohra, line 235: I thought there are 2 options - full and normal?

azurelinuxagent/agent.py
larohra: nvm saw this is an addition to the `collect-logs` command

azurelinuxagent/agent.py
```diff
@@ -183,11 +200,14 @@ def main(args=[]): # pylint: disable=R0912,W0102
                 agent.run_exthandlers(debug)
             elif command == "show-configuration":
                 agent.show_configuration()
+            elif command == "collect-logs":
```
larohra, line 203: Maybe a bit tangential to this PR, but could we also make an Enum of all these commands just to make the code a bit cleaner. Thanks!

larohra (commented):
Initial comments

azurelinuxagent/ga/collect_logs.py
```diff
@@ -0,0 +1,225 @@
...
+        # There are three conditions that need to be met in order to allow periodic log collection:
+        # 1) It should be enabled in the configuration.
+        # 2) The system must be using systemd to manage services. Needed for resource limiting of the log collection.
+        # 3) The python version must be greater than 2.6 in order to support the ZipFile library used when collecting.
+        conf_enabled = conf.get_collect_logs()
+        systemd_present = os.path.exists("/run/systemd/system/")
+        supported_python = PY_VERSION_MINOR >= 7 if PY_VERSION_MAJOR == 2 else PY_VERSION_MAJOR == 3
+        is_allowed = conf_enabled and systemd_present and supported_python
+
+        if self.last_state != is_allowed:
```
larohra, line 74: Is there a chance that any of these requirements can be changed without restarting the agent? If not then we should remove the `self.last_state` and just log this once (maybe before adding the thread handler in the `update.py` file)

azurelinuxagent/ga/collect_logs.py
```diff
@@ -0,0 +1,225 @@
...
+    def stopped(self):
+        return not self.should_run
+
+    def stop(self):
+        self.should_run = False
+        if self.is_alive():
+            self.join()
+
+    def init_protocols(self):
+        # The initialization of ProtocolUtil for the log collection thread should be done within the thread itself
```
larohra, line 116: Why not use the same ProtocolUtil from the ExtHandler thread? Is there any internal call that can cause multithreading issues? If not then reusing the same object would be better since that is the direction we are going towards, removing all the extra ProtocolUtil objects from different places. That would also ensure not having to refresh the protocol object independently as the main loop would already do it for us.

azurelinuxagent/ga/collect_logs.py
```diff
@@ -0,0 +1,225 @@
...
+        start_time = datetime.datetime.utcnow()
+        try:
+            shellutil.run_command(final_command, log_error=True)
+            duration = elapsed_milliseconds(start_time)
+            archive_size = os.path.getsize(COMPRESSED_ARCHIVE_PATH)
+
+            msg = "Successfully collected logs. Archive size: {0} b, elapsed time: {1} ms.".format(archive_size,
+                                                                                                   duration)
+            logger.info(msg)
+            add_event(
```
larohra, line 179: NIT: maybe add the `add_event` in `finally` clause with `msg and is_success` parameterized to make it look a bit cleaner and less redundant?

azurelinuxagent/ga/collect_logs.py
```diff
@@ -0,0 +1,225 @@
...
+        return True
+
+    def _send_logs(self):
+        try:
+            with open(COMPRESSED_ARCHIVE_PATH, "rb") as fh: # pylint: disable=C0103
+                archive_content = fh.read()
+                self.protocol.upload_logs(archive_content)
+                msg = "Successfully uploaded logs."
+                logger.info(msg)
+                add_event(
```
larohra, line 208: NIT: Same comment here, adding the `add_event` to finally

larohra (commented):
Some comments on the main code, will review the tests now

azurelinuxagent/agent.py
```diff
@@ -197,45 +217,58 @@ def parse_args(sys_args): # pylint: disable=R0912
...
-        elif re.match("^([-/]*)deprovision\\+user", a):
+
+        m = re.match("^(?:[-/]*)-mode:([\w/\.\-_]+)", arg) # pylint: disable=W1401,C0103
+        if m is not None:
+            log_collector_mode = m.group(1)
+            if log_collector_mode != "full":
+                print("Error: Invalid value for log collector mode: {0}. Accepted value: full. "
+                      "If mode is not specified, will use normal mode.".format(log_collector_mode))
+                print(usage())
+                sys.exit(1)
```
larohra, line 239: I'm not sure if exiting in the `parse_args` function is the right approach since its just a utility function. Wouldn't it make more sense to just raise an Exception here and let the program die from the `main` ? There's no error handling there anyways so wouldn't make a lot of difference.

larohra (commented):
Final comments

azurelinuxagent/common/logcollector.py
```diff
@@ -195,9 +191,9 @@ def _remove_uncollected_truncated_files(files_to_collect):
 
     @staticmethod
     def _expand_parameters(manifest_data):
-        _LOGGER.info("Using {0} as $LIB_DIR".format(_AGENT_LIB_DIR)) # pylint: disable=W1202
-        _LOGGER.info("Using {0} as $LOG_DIR".format(_EXTENSION_LOG_DIR)) # pylint: disable=W1202
-        _LOGGER.info("Using {0} as $AGENT_LOG".format(_AGENT_LOG)) # pylint: disable=W1202
+        _LOGGER.info("Using %s as $LIB_DIR", _AGENT_LIB_DIR)
```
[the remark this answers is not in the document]
pgombar: Yeah, I did what the linter told me here. :smile:

azurelinuxagent/agent.py
pgombar: Moved to top of file, thanks.

azurelinuxagent/agent.py
pgombar: Yes, but the default option is normal, when you don't specify anything else.
so waagent -collect-logs (normal mode)
or waagent -collect-logs -mode:full

azurelinuxagent/agent.py
pgombar: The log collector is modelled after IID, which has two modes as well which have been sufficient for our needs, so I don't expect other modes to be added here, to be honest. On the other hand, the only valid option for `log_collector_mode` is "full". If the parameter isn't specified, normal mode will be used. I can make the "full" command-line option as a constant to make it cleaner, though.

azurelinuxagent/agent.py
pgombar: I agree, the entire `agent.py` module seems to be a bit wacky. I will have to defer this change for later, though :)

azurelinuxagent/common/conf.py
```diff
@@ -223,6 +225,14 @@ def get_logs_console(conf=__conf__):
     return conf.get_switch("Logs.Console", True)
 
 
+def get_collect_logs(conf=__conf__):
+    return conf.get_switch("Logs.Collect", False)
+
+
+def get_collect_logs_period(conf=__conf__):
+    return conf.get_int("Logs.CollectPeriod", 3600)
```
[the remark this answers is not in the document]
pgombar: We should keep UX in mind. I think having a dedicated boolean flag to enable/disable the feature is more natural for the user, and it really doesn't cost us much to maintain both flags. The collection interval one is simply ignored if the feature is disabled.

azurelinuxagent/ga/collect_logs.py
pgombar: That is a good point, and yes, the agent needs to be restarted in order for any of the changes to the three conditions to take effect. Will log it once in the `update.py` module. Thanks!

azurelinuxagent/ga/update.py
```diff
@@ -276,7 +271,8 @@ def run(self, debug=False): # pylint: disable=R0912,R0914
             # Get all thread handlers
             all_thread_handlers = [
                 get_monitor_handler(),
-                get_env_handler()
+                get_env_handler(),
+                get_collect_logs_handler()
```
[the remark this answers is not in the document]
pgombar: Very good point, made the change. Thanks!

azurelinuxagent/common/utils/textutil.py
```diff
@@ -427,5 +427,8 @@ def str_to_encoded_ustr(s, encoding='utf-8'): # pylint: disable=C0103
         except Exception:
             # If some issues in decoding, just return the string
             return ustr(s)
-    # For Py2, explicitly convert the string to unicode with the specified encoding
-    return ustr(s, encoding=encoding)
+    try:
+        # For Py2, explicitly convert the string to unicode with the specified encoding
+        return ustr(s, encoding=encoding)
+    except Exception as e: # pylint: disable=C0103
+        return ustr(e)
```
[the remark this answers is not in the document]
pgombar: That's totally true. The only caller of this method is `logger.verbose` in `restutil.py`, so basically just for our own debugging purposes. The reason I had to add exception handling is that we are sending the entire content of logs.zip as loaded bytes in memory and we don't control its encoding, so this was throwing for py2.

I have made it cleaner by sanitizing the parameter in the caller by redacting `data` in the caller only if it's the PUT logs API being called. Therefore there's no need for the exception handling I introduced.

azurelinuxagent/common/logcollector.py
```diff
@@ -339,26 +335,34 @@ def collect_logs(self):
...
             files_to_collect = self._create_list_of_files_to_collect()
             _LOGGER.info("### Creating compressed archive ###")
 
-            with zipfile.ZipFile(_COMPRESSED_ARCHIVE_PATH, "w", compression=zipfile.ZIP_DEFLATED) as compressed_archive:
-                for file in files_to_collect: # pylint: disable=redefined-builtin
-                    archive_file_name = LogCollector._convert_file_name_to_archive_name(file)
-                    compressed_archive.write(file, arcname=archive_file_name)
+            with zipfile.ZipFile(COMPRESSED_ARCHIVE_PATH, "w", compression=zipfile.ZIP_DEFLATED) as compressed_archive:
+                for file_to_collect in files_to_collect:
+                    archive_file_name = LogCollector._convert_file_name_to_archive_name(file_to_collect)
```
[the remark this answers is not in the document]
pgombar: Thanks for catching that!

azurelinuxagent/common/logcollector.py
```diff
@@ -339,26 +335,34 @@ def collect_logs(self):
...
+                _LOGGER.info("Elapsed time: %s ms", elapsed_ms)
 
-                compressed_archive_size = os.path.getsize(_COMPRESSED_ARCHIVE_PATH)
-                _LOGGER.info("Successfully compressed files. " # pylint: disable=W1202
-                             "Compressed archive size is {0}b".format(compressed_archive_size))
-                compressed_archive.write(_OUTPUT_RESULTS_FILE_PATH, arcname="results.txt")
+                compressed_archive.write(OUTPUT_RESULTS_FILE_PATH.encode("utf-8"), arcname="results.txt")
 
-            return True
+            return COMPRESSED_ARCHIVE_PATH
```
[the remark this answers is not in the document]
pgombar: Thanks for catching that! Updated and renamed.

azurelinuxagent/common/logcollector_manifests.py
```diff
@@ -1,4 +1,58 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Requires Python 2.6+ and Openssl 1.0+
+#
+
+MANIFEST_NORMAL = """echo,### Probing Directories ###
```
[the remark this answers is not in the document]
pgombar: I did have that approach initially, but if it's a file, the import works a little bit differently than Python modules. I would have to give it the relative file path so the current module, and that gets messed up when running from an egg, from my testing experience. I can spend more time when I'm back on experimenting with this, but my initial results with files didn't yield a working prototype :/

azurelinuxagent/ga/collect_logs.py
```diff
@@ -0,0 +1,225 @@
...
+            add_event(
+                name=AGENT_NAME,
+                version=CURRENT_VERSION,
+                op=WALAEventOperation.LogCollection,
+                is_success=True,
+                message=msg,
+                log_event=False)
+        except Exception as e: # pylint: disable=C0103
+            duration = elapsed_milliseconds(start_time)
+            msg = "Failed to collect logs. Elapsed time: {0} ms. Error: {1}".format(duration, ustr(e))
```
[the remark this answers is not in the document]
pgombar: Thanks, added a special check for `CommandError` and now I am appending `e.stderr` in that case too.

azurelinuxagent/ga/collect_logs.py
pgombar: Thanks, looks much cleaner.

azurelinuxagent/ga/collect_logs.py
```diff
@@ -0,0 +1,225 @@
...
+        systemd_cmd = ["systemd-run", "--unit={0}".format(scope_name), "--scope"]
+
+        # More info on resource limits properties in systemd here:
+        # https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/resource_management_guide/sec-modifying_control_groups
+        cpu_limit, memory_limit = CollectLogsHandler._get_resource_limits()
+        resource_limits = ["--property=CPUAccounting=1", "--property=CPUQuota={0}".format(cpu_limit),
+                           "--property=MemoryAccounting=1", "--property=MemoryLimit={0}".format(memory_limit)]
+
+        # The log tool is invoked from the current agent's egg with the command line option
+        collect_logs_cmd = [get_python_cmd(), "-u", sys.argv[0], "-collect-logs"]
```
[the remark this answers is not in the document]
pgombar: Hmm, I used the same logic as the daemon when deciding to run the latest agent available, check out `update.py#run_latest`. Therefore `sys.executable` will always return the same interpreter as the `get_python_cmd` method since it also uses the platform one to start the extension handler.

azurelinuxagent/ga/collect_logs.py
pgombar: Although I did review the log collection thread code and I don't see any "writes" to a shared state, I think it's still better to use a separate ProtocolUtil just in case. The cost is very low, but the potential gain is high since we definitely avoid any mess with the protocol in case the log collector does modify state.

I would keep it as is since my plan when coming back is to move away from the separate thread and instead make the log collection a non-blocking call in an existing thread (yet to be decided whether monitor's or environment's thread), per Norberto's suggestion.

tests/ga/test_collect_logs.py
```diff
@@ -0,0 +1,192 @@
...
+        # 2) systemd must be managing services
+        # 3) python version 2.7+ which is automatically true for these tests since they are disabled on py2.6
+
+        with _create_collect_logs_handler(systemd_present=False) as collect_logs_handler:
+            with patch("azurelinuxagent.ga.collect_logs.conf.get_collect_logs", return_value=False):
+                collect_logs_handler.run_and_wait()
+
+        self.assertEqual(patch_collect_and_send_logs.call_count, 0, "Log collection should not have been enabled")
+
+        with _create_collect_logs_handler(systemd_present=False) as collect_logs_handler:
```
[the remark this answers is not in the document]
pgombar: Yep, seems like it. Thanks!

tests/ga/test_collect_logs.py
```diff
@@ -0,0 +1,192 @@
...
+            with patch("azurelinuxagent.ga.collect_logs.conf.get_collect_logs", return_value=False):
+                collect_logs_handler.run_and_wait()
+
+        self.assertEqual(patch_collect_and_send_logs.call_count, 0, "Log collection should not have been enabled")
+
+        with _create_collect_logs_handler(systemd_present=False) as collect_logs_handler:
+            with patch("azurelinuxagent.ga.collect_logs.conf.get_collect_logs", return_value=True):
```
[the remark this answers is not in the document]
pgombar: Thanks, I added comments for clarity and removed this redundant mock.

tests/ga/test_collect_logs.py
```diff
@@ -0,0 +1,192 @@
...
+                collect_logs_handler.run_and_wait()
+
+        self.assertEqual(patch_collect_and_send_logs.call_count, 0, "Log collection should not have been enabled")
+
+        with _create_collect_logs_handler() as collect_logs_handler:
+            collect_logs_handler.run_and_wait()
+
+        self.assertEqual(patch_collect_and_send_logs.call_count, 1, "Log collection should have been enabled")
+
+    def test_log_collection_is_invoked_with_resource_limits(self):
```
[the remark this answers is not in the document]
pgombar: Probably is, but that wouldn't be a unit test... We should stick to testing our code, not systemd's inner workings I think.

Currently I have tested this (verified resource limiting works) manually. In the future, it might be worth adding a DCR test for this.

tests/ga/test_update.py
```diff
@@ -1256,25 +1257,27 @@ def iterator(*args, **kwargs): # pylint: disable=useless-return,unused-argument
...
                 with patch('azurelinuxagent.ga.update.get_monitor_handler') as mock_monitor:
                     with patch('azurelinuxagent.ga.update.get_env_handler') as mock_env:
-                        with patch('azurelinuxagent.ga.update.initialize_event_logger_vminfo_common_parameters'):
-                            with patch('time.sleep', side_effect=iterator) as mock_sleep: # pylint: disable=redefined-outer-name
-                                with patch('sys.exit') as mock_exit:
-                                    if isinstance(os.getppid, MagicMock):
-                                        self.update_handler.run()
-                                    else:
-                                        with patch('os.getppid', return_value=42):
+                        with patch('azurelinuxagent.ga.update.get_collect_logs_handler') as mock_collect_logs:
```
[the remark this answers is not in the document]
pgombar: I felt the pain when adding new tests for sure :smile: that would help a lot!

tests/ga/test_update.py
```diff
@@ -1591,14 +1594,14 @@ def test_it_should_delete_extension_events_directory_if_extension_telemetry_pipe
...
-                for ext_dir in expected_events_dirs: # pylint: disable=bad-indentation
-                    self.assertTrue(os.path.exists(ext_dir), "Extension directory {0} should exist!".format(ext_dir)) # pylint: disable=bad-indentation
+        # Rerun update handler again with extension telemetry pipeline enabled to ensure we dont delete events directories
+        with self._setup_test_for_ext_event_dirs_retention() as (update_handler, expected_events_dirs):
+            update_handler.run(debug=True)
+            for ext_dir in expected_events_dirs:
+                self.assertTrue(os.path.exists(ext_dir), "Extension directory {0} should exist!".format(ext_dir))
 
 
+@patch('azurelinuxagent.ga.update.get_collect_logs_handler')
```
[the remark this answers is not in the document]
pgombar: I thought it was an easy thing to do, but when adding these patches to the class setup, I ran into some issues like "call_count not present, return_value not present" for the patches. After about 30 minutes I gave up, since it's not crucial for this PR. Could be done in the future though.

tests/test_agent.py
```diff
@@ -82,13 +84,13 @@ class TestAgent(AgentTestCase):
 
     def test_accepts_configuration_path(self):
         conf_path = os.path.join(data_dir, "test_waagent.conf")
-        c, f, v, d, cfp = parse_args(["-configuration-path:" + conf_path]) # pylint: disable=unused-variable,invalid-name
+        c, f, v, d, cfp, lcm = parse_args(["-configuration-path:" + conf_path]) # pylint: disable=unused-variable,invalid-name
```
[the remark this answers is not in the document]
pgombar: None of the other params are used (except cfp), but I decided to keep all of them named for 1) consistency with other tests, 2) better readability of the test.

tests/test_agent.py
```diff
@@ -179,6 +180,43 @@ def test_agent_get_configuration(self):
...
+        c, f, v, d, cfp, lcm = parse_args(["-collect-logs"]) # pylint: disable=unused-variable,invalid-name
+        self.assertEqual(c, "collect-logs")
+        self.assertEqual(lcm, None)
+
+    @patch("sys.stderr")
+    @patch("sys.exit", side_effect=Exception)
+    def test_rejects_invalid_log_collector_mode(self, mock_exit, mock_stderr): # pylint: disable=unused-argument
+        try:
+            c, f, v, d, cfp, lcm = parse_args(["-collect-logs", "-mode:invalid"]) # pylint: disable=unused-variable,invalid-name
+            self.assertTrue(False) # pylint: disable=redundant-unittest-assert
```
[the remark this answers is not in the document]
pgombar: You're right, I was mimicking the way the other test that rejects an invalid option was written. I've updated both tests, thanks!

azurelinuxagent/ga/collect_logs.py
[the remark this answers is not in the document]
pgombar: Update: I changed this to use `sys.executable` instead.

azurelinuxagent/ga/collect_logs.py
```diff
@@ -121,7 +120,7 @@ def init_protocols(self):
 
     def daemon(self, init_data=False):
         try:
-            if not self._log_collection_allowed():
+            if not is_log_collection_allowed():
```
larohra, line 123: This check is redundant now, not really needed

azurelinuxagent/common/utils/restutil.py
```diff
@@ -312,11 +312,17 @@ def _http_request(method, host, rel_uri, port=None, data=None, secure=False, # p
                                          conn_port,
                                          timeout=10)
 
+
+    if method == "PUT" and "vmAgentLog" in url:
```
larohra, line 316: I dont particularly like this condition here. What's the use case? Since we're just logging it to the verbose log, does it really matter if we redact it or not?

azurelinuxagent/agent.py
larohra: +1 on @kevinclark19a's point. This would definitely make it cleaner and would also make our tests more cleaner (all the tests in `test_agent.py` where we are returning all args even if we dont need it - 
Eg: 
`c, f, v, d, cfp, lcm = parse_args(["-collect-logs", "-mode:invalid"])`

azurelinuxagent/common/utils/textutil.py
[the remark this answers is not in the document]
larohra: >so this was throwing for py2.

What was the error exactly? Was it some encoding thing?

azurelinuxagent/common/logcollector_manifests.py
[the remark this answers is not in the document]
larohra: oh no worries! Thanks for checking though :)

azurelinuxagent/ga/collect_logs.py
larohra: >I would keep it as is since my plan when coming back is to move away from the separate thread and instead make the log collection a non-blocking call in an existing thread

Fair enough, if you plan on moving this to an existing thread anyways then you can explore this point then (make sure to capture a task for this just to ensure we dont forget about it)

tests/ga/test_update.py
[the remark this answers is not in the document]
larohra: awesome thanks for trying!

tests/test_agent.py
```diff
@@ -179,6 +179,38 @@ def test_agent_get_configuration(self):
...
+        # Defaults to None if mode not specified
+        c, f, v, d, cfp, lcm = parse_args(["-collect-logs"]) # pylint: disable=unused-variable,invalid-name
+        self.assertEqual(c, "collect-logs")
+        self.assertEqual(lcm, False)
+
+    @patch("sys.stderr")
+    @patch("sys.exit", side_effect=Exception)
+    def test_rejects_invalid_log_collector_mode(self, mock_exit, mock_stderr): # pylint: disable=unused-argument
+        try:
+            c, f, v, d, cfp, lcm = parse_args(["-collect-logs", "-mode:invalid"]) # pylint: disable=unused-variable,invalid-name
```
larohra, line 197: Does this test need modification? (`-mode:invalid` is not a valid param anymore)

azurelinuxagent/ga/collect_logs.py
pgombar: Yup, you're right! Will remove it.

tests/test_agent.py
pgombar: Thanks! I missed it.

azurelinuxagent/agent.py
kevinclark19a: I realize this was already resolved, but it looks like the linter error is still being suppressed. Was this maybe a merge error?

azurelinuxagent/ga/collect_logs.py
```diff
@@ -39,6 +39,33 @@ def get_collect_logs_handler():
     return CollectLogsHandler()
 
 
+def is_log_collection_allowed():
+    # There are three conditions that need to be met in order to allow periodic log collection:
+    # 1) It should be enabled in the configuration.
+    # 2) The system must be using systemd to manage services. Needed for resource limiting of the log collection.
+    # 3) The python version must be greater than 2.6 in order to support the ZipFile library used when collecting.
+    conf_enabled = conf.get_collect_logs()
+    systemd_present = os.path.exists("/run/systemd/system/")
```
kevinclark19a, line 48: Could this systemd path ever be distro dependent?

pgombar: Not from our knowledge, but let me make it a constant.

azurelinuxagent/common/utils/restutil.py
pgombar: Synced offline, using a flag approach.

azurelinuxagent/agent.py
pgombar: Ah, could be, let me remove it.

larohra (approved):
Thanks for the changes, this would be really helpful once enabled! :)

kevinclark19a (approved):

azurelinuxagent/common/utils/shellutil.py
```diff
@@ -17,10 +17,18 @@
...
+import platform
 import subprocess
+
 import azurelinuxagent.common.logger as logger
 from azurelinuxagent.common.future import ustr
 
+
+def get_python_cmd():
+    major_version = platform.python_version_tuple()[0]
+    return "python" if int(major_version) <= 2 else "python{0}".format(major_version)
```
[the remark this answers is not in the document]
pgombar: I didn't understand your comment, we check the python version and evaluate to `python` if it's py2 and `pythonX` if it's pyX.

azurelinuxagent/common/conf.py
```diff
@@ -97,6 +97,7 @@ def load_conf_from_file(conf_file_path, conf=__conf__):
     "OS.CheckRdmaDriver": False,
     "Logs.Verbose": False,
     "Logs.Console": True,
+    "Logs.Collect": False,
```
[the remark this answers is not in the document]
pgombar: Will send a PR to update this.

azurelinuxagent/common/logcollector.py
```diff
@@ -65,8 +67,9 @@ class LogCollector(object): # pylint: disable=R0903
 
     _TRUNCATED_FILE_PREFIX = "truncated_"
 
-    def __init__(self, manifest_file_path):
-        self._manifest_file_path = manifest_file_path
+    def __init__(self, is_full_mode=False):
```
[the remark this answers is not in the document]
pgombar: "normal" mode is used when the invocation is periodic from the agent. Full mode is only used on-demand and can be specified from the command line as:

`waagent -collect-logs [-full]`

azurelinuxagent/common/logcollector.py
```diff
@@ -88,7 +91,7 @@ def _create_base_dirs():
 
     @staticmethod
     def _set_logger():
-        _f_handler = logging.FileHandler(_OUTPUT_RESULTS_FILE_PATH, encoding="utf-8")
+        _f_handler = logging.FileHandler(OUTPUT_RESULTS_FILE_PATH, encoding="utf-8")
```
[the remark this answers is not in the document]
pgombar: Just did; the output doesn't go to serial console nor syslog.

azurelinuxagent/common/utils/shellutil.py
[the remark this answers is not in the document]
narrieta: laveesh already fixed this

PR #1992, opened 2020-08-26
Travis speedup bug fixed; separate folder for CI stuff.

kevinclark19a:
## Description

The last TravisSpeedup PR introduced a bug where the python 2.6 scenario could fail (and correctly report the errors in the log) but the run would still be marked as a success. 

This PR fixes that in addition to moving all CI specific stuff to its own folder.

---

### PR information


ci/nosetests_only.sh
```diff
@@ -1,12 +1,16 @@
 #!/usr/bin/env bash
 
+EXIT_CODE=0
+
 echo
 echo "========================================="
 echo "nosetests -a '!requires_sudo' output"
 echo "========================================="
-nosetests -a '!requires_sudo' tests
+nosetests -a '!requires_sudo' tests || EXIT_CODE=$(($EXIT_CODE || $?))
```
kevinclark19a, line 9: Before, we were simply ignoring the exit code reported by the `nosetests` command; now, we record it in a local variable.

ci/nosetests_only.sh
```diff
@@ -1,12 +1,16 @@
...
 
 echo "========================================="
 echo "nosetests -a 'requires_sudo' output"
 echo "========================================="
-sudo env "PATH=$PATH" nosetests -a 'requires_sudo' tests
\ No newline at end of file
+sudo env "PATH=$PATH" nosetests -a 'requires_sudo' tests || EXIT_CODE=$(($EXIT_CODE || $?))
```
kevinclark19a, line 14: Same thing as above. Note that we or the command's exit code with our variable so as not to overwrite any exit code we grabbed from the prior `nosetests` command.

ci/nosetests_only.sh
narrieta: Should that be "&& EXIT_CODE=..."?

pgombar: ^ same comment

kevinclark19a: No, because of short circuiting. If the `nosetests` command did indeed fail, any command after `&&` wouldn't execute, and we wouldn't set the exit code. You can try this yourself in bash:
```
$ $(exit 1) || echo "hello"
hello
$ $(exit 1) && echo "hello"
$ $(exit 0) || echo "hello"
$ $(exit 0) && echo "hello"
hello
```

In this case, we don't care about an exit code of zero (because we would default to 0 anyway), so using `||` makes sense.

Actually, now that I think about it, the `$(($EXIT_CODE || $?))` isn't strictly necessary; we could just do `EXIT_CODE=$?` and call it a day, because that command would only execute when the `nosetests` command fails. Same thing with below.

larohra: A question, is the intention here to exit on error? There's an easy way to do that - 
[`set -e`](https://stackoverflow.com/questions/3474526/stop-on-first-error)

If that is not the intention, then why not? At the end of the day we want everything to pass right? If even one of it is failing we should fail fast and return no?

kevinclark19a: If we did it that way, we wouldn't get the output of the next `nosetests` command. This way, we can see _all_ of the failing tests, instead of having to fix these and rerun travis.

ci/nosetests_only.sh
narrieta: @kevinclark19a - we are ok, somehow i missed the initialization of EXIT_CODE on line 3.

could add set -u to the 2 scripts? thanks

ci/nosetests_only.sh
narrieta: @kevinclark19a - we are ok, somehow i missed the initialization of EXIT_CODE on line 3.

could add set -u to the 2 scripts? thanks

pgombar: For reference:
>-u
Treat unset variables and parameters other than the special parameters ‘@’ or ‘*’ as an error when performing parameter expansion. An error message will be written to the standard error, and a non-interactive shell will exit.

kevinclark19a: Thanks! `help set` has this info.

PR #1993, opened 2020-08-26
Fix `arguments-differ<W0221>` linter errors.

kevinclark19a:
## Description

---

### PR information


azurelinuxagent/common/osutil/alpine.py
```diff
@@ -34,7 +34,7 @@ def is_dhcp_enabled(self):
     def get_dhcp_pid(self):
         return self._get_dhcp_pid(["pidof", "dhcpcd"])
 
-    def restart_if(self, ifname): # pylint: disable=W0221
+    def restart_if(self, ifname, unused_retries=None, unused_wait=None):
```
narrieta, line 37: we should keep the parameter names consistent, too, in case somebody ever does "restart_if(..., retries=...)

azurelinuxagent/common/osutil/bigip.py
```diff
@@ -222,7 +222,14 @@ def get_dvd_device(self, dev_dir='/dev'):
...
         raise OSUtilError("Failed to get dvd device")
-
+    
+    # The linter reports that this function's arguments differ from those
+    # of the function this overrides. This doesn't seem to be a problem, however,
+    # because this function accepts any option that could'be been specified for
+    # the original (and, by forwarding the kwargs to the original, will reject any
+    # option _not_ accepted by the original). Additionally, this method allows us
+    # to keep the defaults for mount_dvd in one place (the original function) instead
+    # of having to duplicate it here as well.
```
narrieta, line 232: But with this signature this function will accept arguments the base function doesn't

kevinclark19a: Sort of; this function basically immediately passes the `kwargs` to its super, which would then error out on invalid input.

I realize that that doesn't completely localize the error to this function, though; I felt like this was better than repeating the defaults here as well, and having to keep both up to date. Maybe this isn't the case, though.

pgombar (dismissed):

azurelinuxagent/common/osutil/bigip.py
narrieta: ok as is, thanks

narrieta (dismissed):

narrieta (approved):

azurelinuxagent/common/osutil/alpine.py
```diff
@@ -34,7 +34,7 @@ def is_dhcp_enabled(self):
     def get_dhcp_pid(self):
         return self._get_dhcp_pid(["pidof", "dhcpcd"])
 
-    def restart_if(self, ifname): # pylint: disable=W0221
+    def restart_if(self, ifname, retries=None, wait=None):
```
larohra, line 37: I'm surprised pylint is not complaining about unused variables now lol

larohra (approved):
LGTM

azurelinuxagent/common/osutil/alpine.py
kevinclark19a: I was also confused with this, lol. My understanding is that pylint is smart enough to realize that these parameters are used in the base classes' method, that this method is overriding.

PR #1967, opened 2020-07-25
Fixed delays for HTTP retries rather than exponential delays

larohra:
## Description

This PR addresses the issue where we can see upto 5mins of provisioning delay due to exponential delays in the restutil logic. Now, we will only have sequential delays. Changed the defaults to be at par with the Windows Guest Agent.

Issue # 

---

### PR information
- [ ] The title of the PR is clear and informative.
- [ ] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).
- [ ] Except for special cases involving multiple contributors, the PR is started from a fork of the main repository, not a branch.
- [ ] If applicable, the PR references the bug/issue that it fixes in the description.
- [ ] New Unit tests were added for the changes made and Travis.CI is passing.

### Quality of Code and Contribution Guidelines
- [ ] I have read the [contribution guidelines](https://github.com/Azure/WALinuxAgent/blob/master/.github/CONTRIBUTING.md).

PR #1996, opened 2020-08-28
Remove "default channel" notion

pgombar:
## Description

Repair item for various IcMs where at some point the direct channel would fail, and we "stick" to the HGAP channel forever, since we never unset it from default.

Two major changes:

- remove the option of a default channel, each scenario will try both channels if one of them fails
- switch the order where now HGAP is primary, direct is fallback, like Windows Agent

Note that #1967 significantly reduced the total time for HTTP request retries.

---

### PR information


kevinclark19a (dismissed):

azurelinuxagent/common/protocol/wire.py
```diff
@@ -932,47 +932,36 @@ def _call_hostplugin_with_container_check(self, host_func):
...
-        else:
-            logger.periodic_info(logger.EVERY_HALF_DAY, "[PERIODIC] Using host plugin as default channel.")
-
-        ret = self._call_hostplugin_with_container_check(host_func)
-
-        if not HostPluginProtocol.is_default_channel():
-            logger.info("Setting host plugin as default channel from now on. "
-                        "Restart the agent to reset the default channel.")
-            HostPluginProtocol.set_default_channel(True)
+        ret = self._call_direct_channel(direct_func)
```
larohra, line 962: Incase this raises (anything apart from ResourceGoneError or InvalidContainerError), this would throw and not retry with host. Is that the expected behavior? I personally feel we should always retry with Host irrespective of the error.

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -57,8 +57,6 @@
 
 
 class HostPluginProtocol(object): # pylint: disable=R0902
-    _is_default_channel = False
```
larohra, line 60: The repercussion of removing this property is that incase of a proxy, we would still always try direct first for every request and hence add additional delays per request. This can add up to a lot of unnecessary waiting and might end up adding valuable provision time for the VM (which will lead to TDPR issues for us). I think a better solution would be to keep our current logic where we set the HostGA to default and try HostGA first and if it fails, retry with direct rather than not do it at all. 

So essentially - 

try direct -> Pass (very good) -> Stop
                -> Fail -> Try HostGA -> Set HostGA as default

Next request, 
If HostGA default, try HostGA first -> Fail -> Try Direct again (we dont have this currently)

We can sync more about this offline if you want

azurelinuxagent/ga/update.py
```diff
@@ -937,31 +933,17 @@ def _download(self):
         uris_shuffled = self.pkg.uris
         random.shuffle(uris_shuffled)
         for uri in uris_shuffled:
-            if not HostPluginProtocol.is_default_channel() and self._fetch(uri.uri): # pylint: disable=R1723
```
larohra, line 940: Why not move this function to use the `send_request_using_appropriate_channel` function too?

azurelinuxagent/common/protocol/hostplugin.py
pgombar: >always try direct first for every request and hence add additional delays per request. This can add up to a lot of unnecessary waiting and might end up adding valuable provision time for the VM
I thought about this and that's why I clarified in the description of this PR that your change #1967 significantly reduced the overall time it takes for a request to finally fail (after all retries).

I think it's better to remove a "default state" from HGAP because:
1) keeping state is dangerous in our not-so-safe multithreaded code
2) Windows does no such thing and we should converge the two behaviours. I believe they have a similar retry logic to ours, but I have to double check that.
3) Purely from a supportability perspective, it's better to be deterministic about our scenarios instead of having it rely on a case by case basis. Downloading manifests? Always direct first, then HGAP, not "depends on what the default is, which depends on what happened potentially a long time ago when one direct request failed".

These are my arguments for removing this altogether but I'm happy to discuss more.

pgombar: Plus, the concern to adding more time to TDPR is unchanged. There is no scenario where the HGAP would already be set as default during provisioning since there were no previous requests to fail, correct?

pgombar: Posting here for continuity:
>2. Windows does no such thing and we should converge the two behaviours. I believe they have a similar retry logic to ours, but I have to double check that.

This is true, but the important difference is that Windows always uses HGAP first, then direct. It's worth exploring that change in this PR too, I will do it shortly.

>Plus, the concern to adding more time to TDPR is unchanged. There is no scenario where the HGAP would already be set as default during provisioning since there were no previous requests to fail, correct?

I oversaw that the provisioning steps contains a lot of requests, so before my change one of the requests would fail, then we would be using HGAP throughout (meaning less retries/time to fail). If we change our behaviour to use HGAP first, we achieve the same thing.

azurelinuxagent/common/protocol/wire.py
pgombar: This is true, but it's also the current behavior. Let me check what Windows does.

pgombar: Apparently, Windows invokes the fallback (direct channel) *always*, irrespective of the type of failed request. Let me update the code then.

azurelinuxagent/ga/update.py
pgombar: `self` here is the `GuestAgent` object which does not have the reference to `WireClient` where the method `send_request_using_appropriate_channel` lives.

pgombar: Even the `_fetch` method called here is a private method of `GuestAgent`, not the `WireClient` (like other methods that fetch things).

tests/ga/test_extension.py
```diff
@@ -364,6 +364,7 @@ def tearDownClass(cls):
 @patch('time.sleep', side_effect=lambda _: mock_sleep(0.001))
 @patch("azurelinuxagent.common.protocol.wire.CryptUtil")
 @patch("azurelinuxagent.common.utils.restutil.http_get")
+@patch("azurelinuxagent.common.utils.restutil.http_post")
```
pgombar, line 367: This was needed because, now, we go through HGAP and some of its methods send a health signal via the HealthStore using POST.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -913,47 +913,41 @@ def _call_hostplugin_with_container_check(self, host_func):
...
-        # When using the host plugin channel, regardless if it's set as default or not, try sending the request first.
-        # On specific failures that indicate a stale goal state (such as resource gone or invalid container parameter),
-        # refresh the goal state and try again. If successful, set the host plugin channel as default. If failed,
-        # raise the exception.
+            # Different direct channel functions report failure in different ways: by returning None, False,
+            # or raising ResourceGone or InvalidContainer exceptions.
+            if not ret:
+                logger.periodic_info(logger.EVERY_HOUR, "[PERIODIC] Request failed using the direct channel, "
+                                                        "switching to host plugin.")
+        except (ResourceGoneError, InvalidContainerError) as e: # pylint: disable=C0103
```
larohra, line 926: NIT: missed Pylint error

azurelinuxagent/common/protocol/wire.py
```diff
@@ -913,47 +913,41 @@ def _call_hostplugin_with_container_check(self, host_func):
...
+            ret = direct_func()
 
-        # When using the host plugin channel, regardless if it's set as default or not, try sending the request first.
-        # On specific failures that indicate a stale goal state (such as resource gone or invalid container parameter),
-        # refresh the goal state and try again. If successful, set the host plugin channel as default. If failed,
-        # raise the exception.
+            # Different direct channel functions report failure in different ways: by returning None, False,
+            # or raising ResourceGone or InvalidContainer exceptions.
+            if not ret:
+                logger.periodic_info(logger.EVERY_HOUR, "[PERIODIC] Request failed using the direct channel, "
```
larohra, line 924: NIT: The log messages need refactoring as per the new changes

azurelinuxagent/common/protocol/wire.py
```diff
@@ -913,47 +913,41 @@ def _call_hostplugin_with_container_check(self, host_func):
...
-            logger.periodic_info(logger.EVERY_HALF_DAY, "[PERIODIC] Using host plugin as default channel.")
+        # NOTE: direct_func and host_func are passed as lambdas. Be careful about capturing goal state data in them as
+        # they will not be refreshed even if a goal state refresh is called before retrying the host_func.
 
-        ret = self._call_hostplugin_with_container_check(host_func)
+        ret = None
+        try:
+            ret = self._call_hostplugin_with_container_check(host_func)
+        except Exception:
+            pass
```
larohra, line 947: In the `_call_hostplugin_with_container_check` function, we only catch `(ResourceGoneError, InvalidContainerError)` and log/reraise those. Incase its some other random exception, this would not even log it. We should log the errors incase the exception is not in `(ResourceGoneError, InvalidContainerError)`

azurelinuxagent/ga/update.py
```diff
@@ -941,31 +937,17 @@ def _download(self):
...
+            if self._fetch(uri.uri): # pylint: disable=R1723
```
larohra, line 940: NIT: Pylint

larohra: Do we want to keep similar behaviour for this too? Basically going hostga first and then direct?

azurelinuxagent/ga/update.py
```diff
@@ -941,31 +937,17 @@ def _download(self):
...
-                    else:
-                        logger.warn("Host plugin download failed")
-
-                # If the HostPlugin rejects the request,
-                # let the error continue, but set to use the HostPlugin
-                except ResourceGoneError:
-                    HostPluginProtocol.set_default_channel(True)
-                    raise
-
+                if self._fetch(uri, headers=headers, use_proxy=False): # pylint: disable=R1723
```
larohra, line 947: NIT: Pylint

tests/ga/test_extension.py
```diff
@@ -398,8 +399,9 @@ def _assert_no_handler_status(self, report_vm_status): # pylint: disable=useless
         self.assertEqual(0, len(vm_status.vmAgent.extensionHandlers))
         return
 
-    def _create_mock(self, test_data, mock_http_get, MockCryptUtil, *args): # pylint: disable=unused-argument,invalid-name
+    def _create_mock(self, test_data, mock_http_post, mock_http_get, MockCryptUtil, *args): # pylint: disable=unused-argument,invalid-name
```
larohra, line 402: NIT: Pylint

tests/protocol/test_wire.py
```diff
@@ -218,10 +214,9 @@ def test_get_host_ga_plugin(self, *args): # pylint: disable=unused-argument
     def test_upload_status_blob_should_use_the_host_channel_by_default(self, *_):
         def http_put_handler(url, *_, **__): # pylint: disable=inconsistent-return-statements
             if protocol.get_endpoint() in url and url.endswith('/status'):
-                return MockResponse(body=b'', status_code=200)
+                return MockResponse(body=b'OK', status_code=200)
```
larohra, line 217: I see `b'OK'` being used in multiple places, why not make it a constant of the `MockResponse` class?

tests/protocol/test_wire.py
```diff
@@ -464,213 +456,153 @@ def test_get_ext_conf_with_extensions_should_retrieve_ext_handlers_and_vmagent_m
...
+            self.assertTrue(self.is_host_plugin_extension_artifact_request(urls[0]), "The first attempt should have been over the direct channel")
+            self.assertEqual(urls[1], manifest_url, "The retry should have been over the direct channel")
 
-    def test_fetch_manifest_should_retry_the_host_channel_after_refreshing_the_host_plugin_and_set_the_host_as_default(self):
+    def test_fetch_manifest_should_retry_the_host_channel_after_refreshing_the_host_plugin(self):
         manifest_url = 'https://fake_host/fake_manifest.xml'
         manifest_xml = '<?xml version="1.0" encoding="utf-8"?><PluginVersionManifest/>'
 
-        def http_get_handler(url, *_, **kwargs):
+        def http_get_handler(url, *args, **kwargs): # pylint: disable=unused-argument
```
larohra, line 577: NIT: Pylint

tests/protocol/test_wire.py
```diff
@@ -851,99 +755,62 @@ def http_put_handler(url, *_, **__):
...
-                raise InvalidContainerError()
-
             def host_func(*args): # pylint: disable=unused-argument
                 host_func.counter += 1
                 if host_func.counter == 1:
                     raise ResourceGoneError("Resource is gone")
                 return 42
 
-            direct_func.counter = 0
+            def direct_func(*args): # pylint: disable=unused-argument
```
larohra, line 802: NIT: Pylint

tests/protocol/test_wire.py
```diff
@@ -851,99 +755,62 @@ def http_put_handler(url, *_, **__):
...
-            def direct_func(*args): # pylint: disable=unused-argument
-                direct_func.counter += 1
-                return 42
-
             def host_func(*args): # pylint: disable=useless-return,unused-argument
                 host_func.counter += 1
-                return None
+                return 42
+
+            def direct_func(*args): # pylint: disable=unused-argument
```
larohra, line 764: NIT: Pylint

pgombar:
Closing for #2046

PR #2007, opened 2020-09-11
Start exthandler with the same python interpreter

larohra:
## Description

Currently we do this to start the exthandler service –
`python3 -u /usr/sbin/waagent -run-exthandlers`

 instead of doing this –
`/usr/bin/python3 -u /usr/sbin/waagent -run-exthandlers`

This was causing issues in VMs where python3 and /usr/bin/python3 were different interpreters (multiple ICMs reported)

This PR basically starts the exthandler process using the same python interpreter as the daemon process to ensure that there can be no discrepency there.

Here's the link to the TFS bug - https://msazure.visualstudio.com/One/_workitems/edit/8240001

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/common/utils/shellutil.py
+++ b/azurelinuxagent/common/utils/shellutil.py
@@ -24,10 +24,9 @@ import azurelinuxagent.common.logger as logger
 from azurelinuxagent.common.future import ustr
 
 
-def get_current_python_cmd():
+def get_complete_python_cmd():
+    # Returns the complete python command (with path) for the current python interpreter
     return sys.executable
-    # major_version = platform.python_version_tuple()[0]
-    # return "python" if int(major_version) <= 2 else "python{0}".format(major_version)
 
 
 if not hasattr(subprocess, 'check_output'):
--- a/azurelinuxagent/common/utils/shellutil.py
+++ b/azurelinuxagent/common/utils/shellutil.py
@@ -17,16 +17,17 @@
 # Requires Python 2.6+ and Openssl 1.0+
 #
 
-import platform
 import subprocess
+import sys
 
 import azurelinuxagent.common.logger as logger
 from azurelinuxagent.common.future import ustr
 
 
-def get_python_cmd():
-    major_version = platform.python_version_tuple()[0]
-    return "python" if int(major_version) <= 2 else "python{0}".format(major_version)
+def get_current_python_cmd():
+    return sys.executable
+    # major_version = platform.python_version_tuple()[0]
+    # return "python" if int(major_version) <= 2 else "python{0}".format(major_version)
 
 
 if not hasattr(subprocess, 'check_output'):
--- a/azurelinuxagent/common/utils/shellutil.py
+++ b/azurelinuxagent/common/utils/shellutil.py
@@ -18,15 +18,9 @@
 #
 
 import subprocess
-import sys
 
 import azurelinuxagent.common.logger as logger
 from azurelinuxagent.common.future import ustr
-
-
-def get_complete_python_cmd():
-    # Returns the complete python command (with path) for the current python interpreter
-    return sys.executable
 
 
 if not hasattr(subprocess, 'check_output'):
--- a/tests/ga/test_update.py
+++ b/tests/ga/test_update.py
@@ -1090,7 +1090,7 @@ class TestUpdate(UpdateTestCase): # pylint: disable=too-many-public-methods
 
         self.assertEqual(args, cmds)
         self.assertTrue(len(args) > 1)
-        self.assertRegex(args[0], r"(/*/python\d*)$", "The command doesn't contain full python path")
+        self.assertRegex(args[0], r"(/.*/python[\d.]*)$", "The command doesn't contain full python path")
         self.assertEqual("-run-exthandlers", args[len(args) - 1])
         self.assertEqual(True, 'cwd' in kwargs)
         self.assertEqual(agent.get_agent_dir(), kwargs['cwd'])
@@ -1104,7 +1104,7 @@ class TestUpdate(UpdateTestCase): # pylint: disable=too-many-public-methods
         args = args[0]
 
         self.assertTrue(len(args) > 1)
-        self.assertRegex(args[0], r"(/*/python\d*)$", "The command doesn't contain full python path")
+        self.assertRegex(args[0], r"(/.*/python[\d.]*)$", "The command doesn't contain full python path")
         self.assertEqual("AnArgument", args[len(args) - 1])
 
     def test_run_latest_polls_and_waits_for_success(self):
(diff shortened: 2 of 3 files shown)


tests/ga/test_update.py
```diff
@@ -1086,11 +1086,11 @@ def test_run_latest(self):
         args = args[0]
         cmds = textutil.safe_shlex_split(agent.get_agent_cmd())
         if cmds[0].lower() == "python":
-            cmds[0] = get_python_cmd()
+            cmds[0] = get_complete_python_cmd()
 
         self.assertEqual(args, cmds)
         self.assertTrue(len(args) > 1)
-        self.assertTrue(args[0].startswith("python"))
+        self.assertRegex(args[0], r"(/*/python\d*)$", "The command doesn't contain full python path")
```
kevinclark19a, line 1093: This regex doesn't seem right; `/*/` is equivalent to `/+` and matches one or more `/` in a row. Seems like we're missing a `.`? Aka `/.*/python\d*`.

Additionally, if we're only wanting to match one or zero digits at the end, `\d?` might be a better option.

tests/ga/test_update.py
```diff
@@ -1104,7 +1104,7 @@ def test_run_latest_passes_child_args(self):
         args = args[0]
 
         self.assertTrue(len(args) > 1)
-        self.assertTrue(args[0].startswith("python"))
+        self.assertRegex(args[0], r"(/*/python\d*)$", "The command doesn't contain full python path")
```
kevinclark19a, line 1107: Same as above.

tests/ga/test_update.py
larohra: That is a very good point, I wonder how this is working lol (I am clearly not a regex expert lol). Let me play around with it to see what's happening there.

>Additionally, if we're only wanting to match one or zero digits at the end, \d? might be a better option.
Sometimes the interpreter can also be something like `/usr/bin/python3.6` or something, essentially not necessary that it'll always be just a digit.

larohra: I just realized the regex doesn't work for `python3.6` either lol. Let me fix that too.

larohra: Fixed

tests/ga/test_update.py
larohra: done

tests/ga/test_update.py
kevinclark19a: Lol sure, didn't realize that was a possible name for the python exec. I think the original was working because the regex is just doing a search instead of a match all, i.e. `.../python3` is matched, but not the whole path.

larohra: Yes that is absolutely correct, good catch! I know who to come to now for regex related issues :)

tests/ga/test_update.py
```diff
@@ -1104,7 +1104,7 @@ def test_run_latest_passes_child_args(self):
         args = args[0]
 
         self.assertTrue(len(args) > 1)
-        self.assertTrue(args[0].startswith("python"))
+        self.assertRegex(args[0], r"(/.*/python[\d.]*)$", "The command doesn't contain full python path")
```
kevinclark19a, line 1107: A couple of minor quibbles: `[\d.]` might be better as `[\d\.]` in order to only match the `.` literal; you can preface the regex with a `^` to force the match to be the entire command (since it would then start with `^` and end with `$`).

tests/ga/test_update.py
kevinclark19a: Always happy to help!

tests/ga/test_update.py
larohra: Agreed on using the `^`, will add it.

For the regex, looks like it still considers `.` as a literal rather than a wildcard.
```python

re.match(r"^(/.*/python[\d.]*)$", "/usr/bin/python3.6...").group()
'/usr/bin/python3.6...'

re.match(r"^(/.*/python[\d.]*)$", "/usr/bin/python3.6...qwqwq").group()
  File "<input>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'group'
```

kevinclark19a (approved):

tests/ga/test_update.py
```diff
@@ -1086,11 +1086,11 @@ def test_run_latest(self):
         args = args[0]
         cmds = textutil.safe_shlex_split(agent.get_agent_cmd())
         if cmds[0].lower() == "python":
-            cmds[0] = get_python_cmd()
+            cmds[0] = get_complete_python_cmd()
 
         self.assertEqual(args, cmds)
         self.assertTrue(len(args) > 1)
-        self.assertTrue(args[0].startswith("python"))
+        self.assertRegex(args[0], r"^(/.*/python[\d.]*)$", "The command doesn't contain full python path")
```
narrieta, line 1093: the period in [\d.] needs to be escaped ('.' by itself means any character)

azurelinuxagent/common/utils/shellutil.py
```diff
@@ -17,16 +17,16 @@
...
 import azurelinuxagent.common.logger as logger
 from azurelinuxagent.common.future import ustr
 
 
-def get_python_cmd():
-    major_version = platform.python_version_tuple()[0]
-    return "python" if int(major_version) <= 2 else "python{0}".format(major_version)
+def get_complete_python_cmd():
+    # Returns the complete python command (with path) for the current python interpreter
+    return sys.executable
```
narrieta, line 29: minor suggestion: I would use sys.executable directly instead of defining this function

narrieta (commented):
A minor comment and a suggestion

tests/ga/test_update.py
larohra: A dot (`.`) inside a character class (square brackets `[]`) is considered a literal, it doesn't need to be escaped separately - https://stackoverflow.com/questions/19976018/does-a-dot-have-to-be-escaped-in-a-character-class-square-brackets-of-a-regula

larohra: I also did a small experiment for it - 

```python
re.match(r"^(/.*/python[\d.]*)$", "/usr/bin/python3.6...").group()
'/usr/bin/python3.6...'

re.match(r"^(/.*/python[\d.]*)$", "/usr/bin/python3.6...qwqwq").group()
  File "<input>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'group'
```

kevinclark19a: Good to see confirmation from an external source (stack overflow). Don't get me wrong, I 100% believed you from the experiment alone (in the earlier, now-outdated thread), but it's still nice to see it spelled out.

larohra: Haha yeah, even for me I didnt believe the experiment completely because there can be some discrepancy in the different versions so just finding some online link is definitely more assuring :)

azurelinuxagent/common/utils/shellutil.py
larohra: Done

kevinclark19a (approved):

narrieta (approved):

PR #2010, opened 2020-09-15
Verify that the extension status is an array

narrieta:
If the status for one extension is not an array, the agent fails to report status for all extensions

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -153,7 +153,8 @@ def parse_ext_status(ext_status, data):
     if data is None:
         return
     if not isinstance(data, list):
-        raise ExtensionStatusError(msg="The extension status must be an array: {0}".format(data), code=ExtensionStatusError.StatusFileMalformed)
+        data_string = ustr(data)[:4096]
+        raise ExtensionStatusError(msg="The extension status must be an array: {0}".format(data_string), code=ExtensionStatusError.StatusFileMalformed)
     if not data:
         return
 
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -150,8 +150,13 @@ def parse_ext_substatus(substatus):
 
 
 def parse_ext_status(ext_status, data):
-    if data is None or len(data) == 0: # pylint: disable=len-as-condition
+    if data is None:
         return
+    if not isinstance(data, list):
+        raise ExtensionStatusError(msg="The extension status must be an array: {0}".format(data), code=ExtensionStatusError.StatusFileMalformed)
+    if not data:
+        return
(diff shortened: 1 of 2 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -150,8 +150,13 @@ def parse_ext_substatus(substatus):
 
 
 def parse_ext_status(ext_status, data):
-    if data is None or len(data) == 0: # pylint: disable=len-as-condition
+    if data is None:
         return
+    if not isinstance(data, list):
+        raise ExtensionStatusError(msg="The extension status must be an array: {0}".format(data), code=ExtensionStatusError.StatusFileMalformed)
```
larohra, line 156: Data would contain the complete content of the status file, wont that be too verbose if the extension status is huge (multiple substatuses, etc)?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -150,8 +150,13 @@ def parse_ext_substatus(substatus):
...
+    if not data:
```
larohra, line 157: NIT: Redundant check (line 153 and 157)

narrieta: it not redundant: the first check is for None, the second for length of the list (somehow PEP 8 doesn't like an explicit comparison len(foo) == 0 and asks for not foo (that is the reason for the pylint warning in the previous code)

btw - our code and unit tests allow for an empty status, which i get the feeling is not the right behavior. I decided to keep that behavior for now, I may come back to it late.

azurelinuxagent/ga/exthandlers.py
narrieta: Yes, I thought about truncating it, but I really want to capture the extension name if it is there. The verbosity wouldn't matter too much, I think. I can always truncate it to 1 or 2 K.

narrieta: actually, the error message also goes to telemetry... let me check if we truncate or if a large message would drop the event

narrieta: yes, greater that 64512 we would drop the event. i'll truncate at 4K

kevinclark19a (approved):

azurelinuxagent/ga/exthandlers.py
larohra: >btw - our code and unit tests allow for an empty status, which i get the feeling is not the right behavior. I decided to keep that behavior for now, I may come back to it later.

Ahh I see, no wonder you needed 2 extra checks for it. Good catch though, make sure to open a tfs bug for it to ensure we dont forget :)

larohra (approved):

kevinclark19a (approved):

PR #2011, opened 2020-09-18
Add LIS version to OSInfo.message

andschwa:
## Description
This uses the Linux kernel’s `modinfo` command to retrieve the “version”
field for the “hv_vmbus” kernel module (the LIS drivers). This is the
documented method to retrieve the LIS version.

This command should not fail, given that it’s part of the Linux kernel.
Moreover, the existing RDMA code already relies on `modinfo` without a
timeout etc. It is also distribution-agnostic.

If for some reason the module’s version is unavailable, or `modinfo`
somehow doesn’t exist, we catch the exception and return “Unknown”
instead.

While a timeout would be nice, the existing `run_command` doesn’t
currently support it (and adding it will require some work because we
can’t use Python 3.3’s `timeout` due to Python 2.6 being our minimum
version). Again, there is already existing code which calls `modinfo`
without a timeout, so if this were to fail, it seems it would fail
regardless of this patch.

---

### PR information

### DIFF
--- a/azurelinuxagent/common/version.py
+++ b/azurelinuxagent/common/version.py
@@ -21,6 +21,7 @@ import platform
 import sys
 
 import azurelinuxagent.common.conf as conf
+import azurelinuxagent.common.utils.shellutil as shellutil
 from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
 from azurelinuxagent.common.future import ustr, get_linux_distribution
 
@@ -88,7 +89,7 @@ def get_distro():
         try:
             # dist() removed in Python 3.8
             osinfo = list(platform.dist()) + [''] # pylint: disable=W1505,E1101
-        except: # pylint: disable=W0702
+        except Exception:
             osinfo = ['UNKNOWN', 'FFFF', '', '']
 
     # The platform.py lib has issue with detecting oracle linux distribution.
@@ -114,6 +115,37 @@ def get_distro():
     # Remove trailing whitespace and quote in distro name
     osinfo[0] = osinfo[0].strip('"').strip(' ').lower()
     return osinfo
+
+
+def get_lis_version():
+    """
+    This uses the Linux kernel's 'modinfo' command to retrieve the
+    "version" field for the "hv_vmbus" kernel module (the LIS
+    drivers). This is the documented method to retrieve the LIS module
+    version. Every Linux guest on Hyper-V will have this driver, but
+    it may not be installed as a module (it could instead be built
+    into the kernel). In that case, this will return "Absent" instead
+    of the version, indicating the driver version can be deduced from
+    the kernel version. It will only return "Failed" in the presence
+    of an exception.
+
+    This function is used to generate telemetry for the version of the
+    LIS drivers installed on the VM. The function and associated
+    telemetry can be removed after a few releases.
+    """
+    try:
+        modinfo_output = shellutil.run_command(["modinfo", "-F", "version", "hv_vmbus"])
+        if modinfo_output:
+            return modinfo_output
+        # If the system doesn't have LIS drivers, 'modinfo' will
+        # return nothing on stdout, which will cause 'run_command'
+        # to return an empty string.
+        return "Absent"
+    except Exception:
+        # Ignore almost every possible exception because this is in a
+        # critical code path. Unfortunately the logger isn't already
+        # imported in this module or we'd log this too.
+        return "Failed"
 
 
 AGENT_NAME = "WALinuxAgent"
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -49,8 +49,8 @@ from azurelinuxagent.common.protocol.util import get_protocol_util
 from azurelinuxagent.common.protocol.hostplugin import HostPluginProtocol
 from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
 from azurelinuxagent.common.version import AGENT_NAME, AGENT_VERSION, AGENT_DIR_PATTERN, CURRENT_AGENT,\
-    CURRENT_VERSION, DISTRO_NAME, DISTRO_VERSION, is_current_agent_installed, PY_VERSION_MAJOR, PY_VERSION_MINOR, \
-    PY_VERSION_MICRO
+    CURRENT_VERSION, DISTRO_NAME, DISTRO_VERSION, is_current_agent_installed, get_lis_version, PY_VERSION_MAJOR, \
+    PY_VERSION_MINOR, PY_VERSION_MICRO
 from azurelinuxagent.ga.collect_logs import get_collect_logs_handler, is_log_collection_allowed
 from azurelinuxagent.ga.env import get_env_handler
 from azurelinuxagent.ga.extension_telemetry import get_extension_telemetry_handler
@@ -262,9 +262,9 @@ class UpdateHandler(object): # pylint: disable=R0902
             initialize_event_logger_vminfo_common_parameters(protocol)
 
             # Log OS-specific info.
-            os_info_msg = u"Distro: {0}-{1}; OSUtil: {2}; AgentService: {3}; Python: {4}.{5}.{6}".format(
+            os_info_msg = u"Distro: {0}-{1}; OSUtil: {2}; AgentService: {3}; Python: {4}.{5}.{6}; LISDrivers: {7}".format(
                 DISTRO_NAME, DISTRO_VERSION, type(self.osutil).__name__, self.osutil.service_name, PY_VERSION_MAJOR,
-                PY_VERSION_MINOR, PY_VERSION_MICRO)
+                PY_VERSION_MINOR, PY_VERSION_MICRO, get_lis_version())
             logger.info(os_info_msg)
             add_event(AGENT_NAME, op=WALAEventOperation.OSInfo, message=os_info_msg)
 
--- a/tests/common/test_version.py
+++ b/tests/common/test_version.py
@@ -17,9 +17,10 @@
 
 from __future__ import print_function
 
+import os
+import textwrap
+
 import mock
-import os # pylint: disable=wrong-import-order
-import textwrap # pylint: disable=wrong-import-order
 
 import azurelinuxagent.common.conf as conf
 from azurelinuxagent.common.event import EVENTS_DIRECTORY
@@ -60,34 +61,31 @@ def default_system_exception():
 
 def is_platform_dist_supported():
     # platform.dist() and platform.linux_distribution() is deprecated from Python 3.8+
-    if PY_VERSION_MAJOR == 3 and PY_VERSION_MINOR >= 8: # pylint: disable=no-else-return
+    if PY_VERSION_MAJOR == 3 and PY_VERSION_MINOR >= 8:
         return False
-    else:
-        return True
+    return True
+
 
 class TestAgentVersion(AgentTestCase):
-    def setUp(self): # pylint: disable=useless-return
+    def setUp(self):
         AgentTestCase.setUp(self)
-        return
 
     @mock.patch('platform.system', side_effect=freebsd_system)
     @mock.patch('re.sub', side_effect=freebsd_system_release)
-    def test_distro_is_correct_format_when_freebsd(self, platform_system_name, mock_variable): # pylint: disable=useless-return,unused-argument
+    def test_distro_is_correct_format_when_freebsd(self, platform_system_name, mock_variable): # pylint: disable=unused-argument
         osinfo = get_distro()
         freebsd_list = ['freebsd', "10.0", '', 'freebsd']
         self.assertListEqual(freebsd_list, osinfo)
-        return
 
     @mock.patch('platform.system', side_effect=openbsd_system)
     @mock.patch('re.sub', side_effect=openbsd_system_release)
-    def test_distro_is_correct_format_when_openbsd(self, platform_system_name, mock_variable): # pylint: disable=useless-return,unused-argument
+    def test_distro_is_correct_format_when_openbsd(self, platform_system_name, mock_variable): # pylint: disable=unused-argument
         osinfo = get_distro()
         openbsd_list = ['openbsd', "20.0", '', 'openbsd']
         self.assertListEqual(openbsd_list, osinfo)
-        return
 
     @mock.patch('platform.system', side_effect=default_system)
-    def test_distro_is_correct_format_when_default_case(self, *args): # pylint: disable=useless-return,unused-argument
+    def test_distro_is_correct_format_when_default_case(self, *args): # pylint: disable=unused-argument
         default_list = ['', '', '', '']
         unknown_list = ['unknown', 'FFFF', '', '']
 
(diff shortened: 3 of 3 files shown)


azurelinuxagent/common/version.py
```diff
@@ -146,6 +159,8 @@ def get_distro():
 DISTRO_CODE_NAME = __distro__[2]
 DISTRO_FULL_NAME = __distro__[3]
 
+LIS_VERSION = get_lis_version()
```
narrieta, line 162: why add a new module global instead of just calling the function?

tests/common/test_version.py
```diff
@@ -115,6 +116,10 @@ def test_distro_is_correct_for_exception_case(self, *args): # pylint: disable=us
         self.assertListEqual(default_list, osinfo)
         return
 
+    def test_lis_version(self, *args): # pylint: disable=useless-return,unused-argument
```
narrieta, line 119: please address the pylint warnings (remove the return statement and use *_ instead of *_args)

narrieta (changes requested):

azurelinuxagent/common/version.py
andschwa: For consistency with the existing pattern of code. I can change it.

tests/common/test_version.py
andschwa: This is consistent with all the above unit tests. Do you want those changed too?

azurelinuxagent/common/version.py
andschwa: (I actually had written it the way you suggest first, but changed it because it was inconsistent.)

tests/common/test_version.py
larohra: We actually recently enabled the linters on our repo. The above tests haven't been touched in a while so we have it commented, but we will eventually fix them. All the new code though is supposed to be without any linter errors so it would be great if you could address these errors! :)

azurelinuxagent/common/version.py
narrieta: Thanks; we are rewriting most of this code since some of the patterns it follows tend to be problematic. Do not worry about consistency.

tests/common/test_version.py
andschwa: Ok, since it wouldn't make sense to have just one new unit test dissimilar to all the others, I went ahead and fixed the trivial pylint errors in this file.

azurelinuxagent/common/version.py
andschwa: Ok, I personally think it's less consistent now, but it is as you asked.

azurelinuxagent/common/version.py
```diff
@@ -116,6 +117,24 @@ def get_distro():
...
+    """
+    try:
+        lis_version = shellutil.run_command(["modinfo", "-F", "version", "hv_vmbus"], log_error=True)
+        # If the system doesn't have LIS drivers, 'modinfo' will
+        # return nothing on stdout, which will cause 'run_command' to
+        # return an empty string.
+        if lis_version:
+            return lis_version
+        return "Unknown"
+    except: # pylint: disable=W0702
```
larohra, line 134: NIT: A missed linter comment here

larohra: Also, pure synthetic comment, but you could have a single `return "Unknown"` after the `except` call rather than 2 returns for the same value. Functionally its the same just looks a bit more cleaner. (I'm good with the code either way)

Eg: 

```python
try:
     do_something()
     return_val = "val"
except: 
     # Error handling
     return_val = "error"
return return_val
 ```

```

azurelinuxagent/common/version.py
andschwa: Ah I missed your message 😂

larohra (changes requested):

tests/common/test_version.py
larohra: Awesome thanks for that :)

azurelinuxagent/common/version.py
andschwa: It's intentionally catching and ignoring all exceptions here (which results in a pylint warning) since this code runs really early at startup (I followed the pattern in `get_distro()` above which implied these code paths want _zero_ exceptions being raised).

andschwa:
```python
        try:
            # dist() removed in Python 3.8
            osinfo = list(platform.dist()) + [''] # pylint: disable=W1505,E1101
        except: # pylint: disable=W0702
            osinfo = ['UNKNOWN', 'FFFF', '', '']
```

larohra: Ahh I see. Could you still just have a blanket catch and just log it as a warning or something? It would be useful info incase we ever need to debug. That would also take care of the pylint warning

andschwa: I'm not a fan of that style because it relies on Python's weird concept of scope. For full clarity it would need to be:

```python
    lis_version = "Unknown"
    try:
        command_output = shellutil.run_command(["modinfo", "-F", "version", "hv_vmbus"], log_error=True)
        # If the system doesn't have LIS drivers, 'modinfo' will
        # return nothing on stdout, which will cause 'run_command' to
        # return an empty string.
        if command_output:
            lis_version = command_output
    except: # pylint: disable=W0702
        # Ignore every possible exception because this is in a
        # critical code path.
        pass
    return lis_version
```

Do you prefer that?

larohra: As @narrieta mentioned in another comment, this is all legacy code that we're in the process of rewriting bit by bit, so dont worry about maintaining consistency with this code :)

andschwa: It might be reasonable in both this function and `get-distro()` to catch `Exception`, which in the [hierarchy](https://docs.python.org/3.5/library/exceptions.html#exception-hierarchy) is the base exception for everything but SystemExit, KeyboardInterrupt, and GeneratorExit.

andschwa: Yeah I see that, lmao, I was doing my best to fit my code into the existing as closely as possible, rather than creating new patterns and precedents in code I don't maintain 😅

larohra: > Do you prefer that?

This looks good to me! 

Also catching `Exception` (not `BaseException`) sounds good 👍

andschwa: You got it! 😀

larohra: > Yeah I see that, lmao, I was doing my best to fit my code into the existing as closely as possible, rather than creating new patterns and precedents in code I don't maintain 😅

haha no worries, thanks for being considerate towards our code :)

azurelinuxagent/common/version.py
```diff
@@ -116,6 +117,28 @@ def get_distro():
...
 
+def get_lis_version():
+    """
+    This uses the Linux kernel's 'modinfo' command to retrieve the "version"
+    field for the "hv_vmbus" kernel module (the LIS drivers). This is the
+    documented method to retrieve the LIS version.
+    """
+    lis_version = "Unknown"
+    try:
+        command_output = shellutil.run_command(["modinfo", "-F", "version", "hv_vmbus"], log_error=True)
```
narrieta, line 128: we need better logging for errors here. since exceptions are being ignored, should the command fail we would get a message saying that modinfo failed; the message should be more explicit, something saying that the version of the lis driver could not be retrieved.

azurelinuxagent/common/version.py
```diff
@@ -116,6 +117,28 @@ def get_distro():
...
+        # If the system doesn't have LIS drivers, 'modinfo' will
+        # return nothing on stdout, which will cause 'run_command' to
+        # return an empty string.
+        if command_output:
+            lis_version = command_output
```
narrieta, line 133: we also need a more explicit message here. if we use empty then the log/telemetry will just include "LIS: "

many "not_installed"

azurelinuxagent/ga/update.py
```diff
@@ -261,9 +261,9 @@ def run(self, debug=False): # pylint: disable=R0912,R0914
             initialize_event_logger_vminfo_common_parameters(protocol)
 
             # Log OS-specific info.
-            os_info_msg = u"Distro: {0}-{1}; OSUtil: {2}; AgentService: {3}; Python: {4}.{5}.{6}".format(
+            os_info_msg = u"Distro: {0}-{1}; OSUtil: {2}; AgentService: {3}; Python: {4}.{5}.{6}; LIS: {7}".format(
```
narrieta, line 264: small change: can we use "LIS driver" instead of just "LIS"

tests/common/test_version.py
```diff
@@ -113,39 +111,43 @@ def test_distro_is_correct_for_exception_case(self, *args): # pylint: disable=us
             osinfo = get_distro()
 
         self.assertListEqual(default_list, osinfo)
-        return
+
+    def test_lis_version(self):
```
narrieta, line 115: for new tests we are following this as naming convention: test_<subject_under_test>_should_<behaviour_being_tested>

this test could be named "test_get_lis_version_should_return_a_string"

tests/common/test_version.py
```diff
@@ -256,8 +257,8 @@ def test_get_f5_platform_bigiq_5_1_0(self):
         Changelist: 1907534
         JobID: 726344""")
 
-        mo = mock.mock_open(read_data=version_file) # pylint: disable=invalid-name
-        with patch(open_patch(), mo):
+        mocked_open = mock.mock_open(read_data=version_file)
+        with patch(open_patch(), mocked_open):
```
narrieta, line 261: thanks for fixing those linter warnings!

narrieta (changes requested):

azurelinuxagent/common/version.py
andschwa: The current logic is, if we don't get a version string back from `modprobe`, we return "Unknown." So actually this will include `LIS: Unknown`.

azurelinuxagent/common/version.py
andschwa: The logger module is not currently imported in `version.py`. However, as far as I can tell, `run_command` imports your logger module and with `log_error=True` records the logs. I am actually unsure what cautions this codebase currently takes with `run_command`. At least `get_systemd_version()` simply returns whatever `run_command` returns, with no wrapping try/catch, but `_collect_logs()` catches `Exception` and then includes it in event message (but does not log it with the logger module).

How would you like exceptions thrown by `run_command` handled here?

azurelinuxagent/common/version.py
narrieta: ok - not sure if you guys want to differentiate between unknown and no installed... if no need, ok as is

andschwa: Following up with you offline about this.

andschwa: While I've been unable to confirm if there's any case where the drivers would be absent on an Azure Linux VM, I've gone ahead and differentiated this anyway.

andschwa:
Good morning @narrieta, is there anything else needed here?

narrieta:
@andschwa - Please add a comment specifying that this is just a temporary change and can be removed in a few releases

azurelinuxagent/common/version.py
```diff
@@ -116,6 +117,32 @@ def get_distro():
...
+def get_lis_version():
+    """
+    This uses the Linux kernel's 'modinfo' command to retrieve the
+    "version" field for the "hv_vmbus" kernel module (the LIS
+    drivers). This is the documented method to retrieve the LIS
+    drivers' version. Every Linux guest on Hyper-V will have this
+    module (and so this version field).
+
+    This telemetry can be moved to a separate extension when the Linux
+    Integration Services team (@lisval) develops such.
```
andschwa, line 129: @narrieta already done!

narrieta: Please replace
    This telemetry can be moved to a separate extension when the Linux
    Integration Services team (@lisval) develops such.

With something similar to
    This function is used to generate telemetry for the version of the LIS drivers installed on the VM. The function and associated telemetry can be removed after a few releases.

andschwa: Done.

narrieta:
code looks good; running automation

andschwa:
You can just auto-rebase this on top of develop, the only new commit to develop was my fix for the readme. There's no need to merge develop back into this. (I believe I have "allow edits by maintainers" checked too, so you don't have to wait on me to do that.)

larohra (approved):
LGTM

narrieta:
Automation failed on several distros

    std_out:
    waagent.log contains the following ERROR log(s): 
    ['2020-09-23T19:44:13.176091Z ERROR ExtHandler ExtHandler Command: [modinfo -F version hv_vmbus], return code: [1], 
    stdout: [] stderr: [modinfo: ERROR: Module hv_vmbus not found.\n']

will contact @andschwa offline

azurelinuxagent/common/version.py
andschwa: @narrieta So this was the cause of those failures. Based on my understanding of the other tests, I've changed `log_error` to `False` (well, I removed the argument and am using its default now).

tests/common/test_version.py
```diff
@@ -113,39 +111,45 @@ def test_distro_is_correct_for_exception_case(self, *args): # pylint: disable=us
...
+    def test_get_lis_version_should_return_a_string(self):
+        """
+        On a Hyper-V guest, this function should return a string of the
+        version, like '4.3.5'. Anywhere else it should return
+        'Absent', and should not return 'Failed', so we check that it
+        returns a string but not 'Failed'.
+        """
+        lis_version = get_lis_version()
+        self.assertIsInstance(lis_version, str)
+        self.assertNotEqual(lis_version, "Failed")
```
narrieta, line 124: how is the test ensuring that the function will never return "Failed"? We run unit tests in a variety of platforms (OS X, Linux, Windows, physical machines, VMs, containers)

andschwa: I don't understand your question. It will only return `Failed` if an exception is raised when calling `run_command`. As I asked you above, I'm not sure how you want that handled, so I went with catching the exception and returning "Failed." As that should never be the case, this test should pass on all those platforms. If the test fails, something weird has happened that needs debugging (which is the point of the assertion).

andschwa: (Also I've updated the doc strings to be more clear about this.)

narrieta: My concern here is whether this assert will ever fail and force us to break our development cycle to debug the failure.

The production code doesn't seem to care at all about the failure. Why does the test do? If you are trying to get a notification of the failure you would be better handling this in the production code with, for example, a telemetry event that you guys would monitor.

andschwa: > The production code doesn't seem to care at all about the failure.

This is correct, but wouldn't you be? I was trying to write as extensive of a unit test as possible (given the contributor guidelines) especially where it does not make sense to mock the function.

I'll remove the assertion for you.

andschwa:
Hello @narrieta and @larohra, is there anything left for this?

larohra:
> Hello @narrieta and @larohra, is there anything left for this?

@andschwa I'm just running a sanity test automation run, will merge once it passes.

andschwa:
Thanks for the update @larohra!

larohra (approved):

larohra:
DCR was successful

narrieta (approved):

PR #2015, opened 2020-09-22
Verify that the CPU and Memory cgroups for the agent are properly initialized; disabled cgroups if they are not active.

narrieta:
On a small number of machines the controllers for CPU and/or memory were not active and the agent's cgroups were initialized to None. The resource usage data was not being collected, as expected, but the check for processes in the agent's cgroup was still being done, generating false positives about cgroups not working properly.

Also, added a check to disable cgroups if both cgroups are None.

On another set of machines, the agent was within the init.scope, or directly under system.slice; added a check to disable cgroups if the agent is not in system.slice/walinuxagent.service.

I also converted some WARNING messages to INFO, since there were a couple of questions when people noticed them in the agent log.

Lastly, addressed linter warnings in the code I touched.

### DIFF
--- a/azurelinuxagent/common/cgroupconfigurator.py
+++ b/azurelinuxagent/common/cgroupconfigurator.py
@@ -35,7 +35,9 @@ class CGroupConfigurator(object):
 
     NOTE: with the exception of start_extension_command, none of the methods in this class raise exceptions (cgroup operations should not block extensions)
     """
-    class __impl(object): # pylint: disable=R0902,C0103
+    # too-many-instance-attributes<R0902> Disabled: class complexity is OK
+    # invalid-name<C0103> Disabled: class is private, so name starts with __
+    class __Impl(object):  # pylint: disable=R0902,C0103
         def __init__(self):
             self._initialized = False
             self._cgroups_supported = False
@@ -46,7 +48,8 @@ class CGroupConfigurator(object):
             self._get_processes_in_agent_cgroup_last_error = None
             self._get_processes_in_agent_cgroup_error_count = 0
 
-        def initialize(self): # pylint: disable=R0912
+        # too-many-branches<R0912> Disabled: branches are sequential, not nested
+        def initialize(self):  # pylint: disable=R0912
             # pylint: disable=too-many-locals
             try:
                 if self._initialized:
@@ -76,12 +79,7 @@ class CGroupConfigurator(object):
                     logger.info(message)
                     add_event(op=WALAEventOperation.CGroupsInfo, message=message)
 
-                def log_cgroup_warn(format_string, *args):
-                    message = format_string.format(*args)
-                    logger.warn(message)
-                    add_event(op=WALAEventOperation.CGroupsInfo, message=message, is_success=False, log_event=False)
-
-                log_cgroup_info("systemd version: {0}", self._cgroups_api.get_systemd_version()) # pylint: disable=E1101
+                log_cgroup_info("systemd version: {0}", self._cgroups_api.get_systemd_version())  # pylint: disable=E1101
 
                 #
                 # Older versions of the daemon (2.2.31-2.2.40) wrote their PID to /sys/fs/cgroup/{cpu,memory}/WALinuxAgent/WALinuxAgent.  When running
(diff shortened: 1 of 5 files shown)


azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -76,12 +79,7 @@ def log_cgroup_info(format_string, *args):
                     logger.info(message)
                     add_event(op=WALAEventOperation.CGroupsInfo, message=message)
 
-                def log_cgroup_warn(format_string, *args):
```
narrieta, line 79: converted warnings into infos

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -90,69 +88,79 @@ def log_cgroup_warn(format_string, *args):
...
-                cpu_cgroup_relative_path, memory_cgroup_relative_path = self._cgroups_api.get_process_cgroup_relative_paths("self") # pylint: disable=E1101
+                agent_unit_name = self._cgroups_api.get_agent_unit_name()  # pylint: disable=E1101
+                cpu_cgroup_relative_path, memory_cgroup_relative_path = self._cgroups_api.get_process_cgroup_relative_paths("self")  # pylint: disable=E1101
+                expected_relative_path = os.path.join('system.slice', agent_unit_name)
                 if cpu_cgroup_relative_path is None:
-                    log_cgroup_warn("The agent's process is not within a CPU cgroup")
+                    log_cgroup_info("The agent's process is not within a CPU cgroup")
                 else:
-                    cpu_accounting = self._cgroups_api.get_unit_property(agent_unit_name, "CPUAccounting") # pylint: disable=E1101
+                    if cpu_cgroup_relative_path != expected_relative_path:
```
narrieta, line 125: Do not enable cgroups if agent is not in system.slice/walinuxagent.service

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -90,69 +88,79 @@ def log_cgroup_warn(format_string, *args):
...
 
                 if memory_controller_root is None or memory_cgroup_relative_path is None:
                     logger.info("Will not track memory for the agent's cgroup")
                 else:
                     self._agent_memory_cgroup_path = os.path.join(memory_controller_root, memory_cgroup_relative_path)
+                    log_cgroup_info("Agent Memory cgroup: {0}", self._agent_memory_cgroup_path)
                     CGroupsTelemetry.track_cgroup(MemoryCgroup(agent_unit_name, self._agent_memory_cgroup_path))
 
-                log_cgroup_info("Agent cgroups: CPU: {0} -- MEMORY: {1}", self._agent_cpu_cgroup_path, self._agent_memory_cgroup_path)
+                if self._agent_cpu_cgroup_path is not None or self._agent_memory_cgroup_path is not None:
```
narrieta, line 157: Do not enable cgroups if both cpu and memory are None

azurelinuxagent/ga/monitor.py
```diff
@@ -181,11 +181,13 @@ def __init__(self):
             ResetPeriodicLogMessagesOperation(),
             PeriodicOperation("collect_and_send_events", self.collect_and_send_events, self.EVENT_COLLECTION_PERIOD),
             ReportNetworkErrorsOperation(),
-            PollResourceUsageOperation(),
             PeriodicOperation("send_host_plugin_heartbeat", self.send_host_plugin_heartbeat, self.HOST_PLUGIN_HEARTBEAT_PERIOD),
             PeriodicOperation("send_imds_heartbeat", self.send_imds_heartbeat, self.IMDS_HEARTBEAT_PERIOD),
             ReportNetworkConfigurationChangesOperation(),
         ]
+        if CGroupConfigurator.get_instance().enabled():
```
narrieta, line 188: don't check for cgroups stuff if they are disabled

azurelinuxagent/ga/update.py
```diff
@@ -267,6 +268,24 @@ def run(self, debug=False): # pylint: disable=R0912,R0914
             logger.info(os_info_msg)
             add_event(AGENT_NAME, op=WALAEventOperation.OSInfo, message=os_info_msg)
 
+            #
```
narrieta, line 271: initialization of these thread was racing against initialization of the main thread

azurelinuxagent/ga/update.py
```diff
@@ -267,6 +268,24 @@ def run(self, debug=False): # pylint: disable=R0912,R0914
...
+            # Perform initialization tasks
+            #
+            from azurelinuxagent.ga.exthandlers import get_exthandlers_handler, migrate_handler_state
```
kevinclark19a, line 274: Should this import be up top of the file?

narrieta: ya; i tried to fix that and it ends up breaking up the tests... we have several of those problematic imports across the code base... yet another cleanup task to do

kevinclark19a: Hm. What were the errors the tests were having? Should we document this?

narrieta: you can try moving it in the current code and check the errors... i don't think we need to document it, all of us are aware of this (now you are :))

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -183,13 +191,13 @@ def _invoke_cgroup_operation(self, operation, error_message, on_error=None):
...
                 return operation()
-            except Exception as e: # pylint: disable=C0103
-                logger.warn("{0} Error: {1}".format(error_message, ustr(e)))
+            except Exception as exception:
+                logger.warn("{0} Error: {1}".format(error_message, ustr(exception)))
                 if on_error is not None:
                     try:
-                        on_error(e)
-                    except Exception as ex: # pylint: disable=W0612
-                        logger.warn("CGroupConfigurator._invoke_cgroup_operation: {0}".format(ustr(e)))
```
narrieta, line 192: BTW on this one the linter caught me logging the wrong exception

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -90,69 +88,79 @@ def log_cgroup_warn(format_string, *args):
...
+                agent_unit_name = self._cgroups_api.get_agent_unit_name()  # pylint: disable=E1101
+                cpu_cgroup_relative_path, memory_cgroup_relative_path = self._cgroups_api.get_process_cgroup_relative_paths("self")  # pylint: disable=E1101
+                expected_relative_path = os.path.join('system.slice', agent_unit_name)
                 if cpu_cgroup_relative_path is None:
-                    log_cgroup_warn("The agent's process is not within a CPU cgroup")
+                    log_cgroup_info("The agent's process is not within a CPU cgroup")
                 else:
-                    cpu_accounting = self._cgroups_api.get_unit_property(agent_unit_name, "CPUAccounting") # pylint: disable=E1101
+                    if cpu_cgroup_relative_path != expected_relative_path:
+                        log_cgroup_info("The Agent is not in the expected cgroup; will not enable cgroup monitoring. CPU relative path:[{0}] Expected:[{1}]", cpu_cgroup_relative_path, expected_relative_path)
```
larohra, line 126: Shouldn't it be a warning here?

larohra (approved):
1 minor comment, else LGTM

tests/common/test_cgroupconfigurator.py
```diff
@@ -31,29 +31,105 @@
...
+        configurator = CGroupConfiguratorSystemdTestCase._get_new_cgroup_configurator_instance()
 
-        self.assertTrue(
-            any(cg for cg in tracked if cg.name == 'walinuxagent.service' and 'cpu' in cg.path),
-            "The Agent's CPU is not being tracked")
-        self.assertTrue(
-            any(cg for cg in tracked if cg.name == 'walinuxagent.service' and 'memory' in cg.path),
-            "The Agent's memory is not being tracked")
+        # protected-access<W0212> Disabled: OK to access CGroupConfigurator._tracked from unit test for CGroupConfigurator
+        tracked = CGroupsTelemetry._tracked  # pylint: disable=protected-access
```
kevinclark19a, line 59: Is this protected access ok here? It would be nice to keep unit tests agnostic of the internals of our prod modules.

narrieta: Yes, in an ideal situation. I believe in this situation it is OK to break the rule; the unit tests do not make heavy use of internals, and this is unit test for A using A's internals. Things get dicey in other parts of the code where we have unit tests for A using internals of D when A uses B uses C uses D.
Exposing the internals of CGroupsTelemetry just the test's benefit doesn't seem right, and looking more more complicated way to do this check doesn't seem to be worth it.

kevinclark19a (approved):

PR #2016, opened 2020-09-22
Gs fix for no config

kevinclark19a:
## Description

DCR runs are failing when wireserver returns a no-op goal state. This change adds some error handling to skip such goal states.

---

### PR information


azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -135,19 +131,41 @@ def fetch_goal_state(wire_client):
...
         """
         Fetches the full goal state, including nested properties (such as extension config).
         """
-        return GoalState(wire_client, full_goal_state=True)
+        goal_state = GoalState(wire_client, full_goal_state=True)
+
+        try:
+            GoalState._verify_gs_for_noop(goal_state)
+        except ValueError:
+            return None
```
narrieta, line 153: we shouldn't return None... the caller needs to handle the missing role instance. (e.g. the logic to refresh the host ga plugin should probably retry; the extension handler should skip extension processing, etc)

narrieta: similar comment below (fetch_full_goal_state_if_incarnation_different_than)

narrieta (changes requested):

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -75,15 +75,11 @@ def __init__(self, wire_client, full_goal_state=False, base_incarnation=None):
 
         AgentGlobals.update_container_id(self.container_id)
 
-        fetch_full_goal_state = False
-        if full_goal_state:
-            fetch_full_goal_state = True
+        if full_goal_state and role_config:
```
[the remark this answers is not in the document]
kevinclark19a: gotcha. You mean before we do the `find` call on the children of the role instance?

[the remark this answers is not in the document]
narrieta: yup, i think we need to skip this if role instance is None:

        self.role_instance_id = findtext(role_instance, "InstanceId")
        role_config = find(role_instance, "Configuration")
        self.role_config_name = findtext(role_config, "ConfigName")
        container = find(xml_doc, "Container")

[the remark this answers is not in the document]
kevinclark19a: The idea was to use the `role_config` as a guard on attempting to continue, regardless of the reason provided by the caller. My thoughts were that there were 3 distinct "happy paths":

- the `full_goal_state` was requested, and we do in fact have something other than a no-op
- an old `base_incarnation` was specified, and, (again) we have something other than a no-op
- We have no reason to fetch the entire goal state (and we therefore do not care if we have a no-op or not)

I'm now worried that the last of the three is not actually true; if neither `full_goal_state` nor `base_incarnation` were specified, (the third case above) I'm pretty sure we're doing a `HostPlugin` update, and you mentioned somewhere on this PR that we need the `role_config_name` to be populated for that.

I think the other two cases are fine as is, though I think from a semantic standpoint (and also to satisfy your other comment) I need to move the `role_config` check to above.

Let me know if there's anything I missed above.

azurelinuxagent/common/protocol/goal_state.py
kevinclark19a: I think the precedent for these methods was to return `None`. e.g. 

> the extension handler should skip extension processing

already happens with the `None` returned. Would the simplest answer be to add the host ga refresh retry logic based on a `is None` check?

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -60,14 +61,21 @@ def __init__(self, wire_client, full_goal_state=False, base_incarnation=None):
         directly.
 
         """
-        try:
-            uri = GOAL_STATE_URI.format(wire_client.get_endpoint())
+        uri = GOAL_STATE_URI.format(wire_client.get_endpoint())
+
+        for _ in range(1, 6):
```
kevinclark19a, line 66: Anyone have any strong feelings about where to put a constant with the number of retries?

narrieta: I'd add as a private member of the GoalState class, or if we want all constants at the top of the file as a private of the module

narrieta, line 66: Consider reducing the number of retries (2-3?) just in case (we do not want to send too many requests to the wire server.)

azurelinuxagent/common/exception.py
```diff
@@ -153,6 +153,14 @@ class ProtocolNotFoundError(ProtocolError):
     def __init__(self, msg=None, inner=None): # pylint: disable=W0235
         super(ProtocolNotFoundError, self).__init__(msg, inner)
 
+class IncompleteGoalStateError(ProtocolError):
+    """
+    Goal state is returned incomplete.
+    """
+
+    def __init__(self, msg=None, inner=None): # pylint: disable=W0235
```
narrieta, line 161: we need to document the reason of all new pylint supressions

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -60,14 +61,21 @@ def __init__(self, wire_client, full_goal_state=False, base_incarnation=None):
...
             self.xml_text = wire_client.fetch_config(uri, wire_client.get_header())
             xml_doc = parse_doc(self.xml_text)
 
+            role_instance = find(xml_doc, "RoleInstance")
+            if role_instance:
+                break
+        else:
+            raise IncompleteGoalStateError("Goal State doesn't have a RoleInstance.")
```
narrieta, line 74: we need to log the incarnation somewhere

narrieta: (at least for the extension handler)

tests/protocol/test_wire.py
```diff
@@ -1063,6 +1065,26 @@ def test_it_should_return_true_on_success(self):
         with mock_wire_protocol(mockwiredata.DATA_FILE) as protocol:
             self.assertTrue(protocol.client.try_update_goal_state(), "try_update_goal_state should have succeeded")
 
+    def test_noop_gs_should_skip(self):
+        first_ext = extension_emulator()
```
narrieta, line 1069: why do we need extensions at all in this test?

narrieta (changes requested):
Added some comments

azurelinuxagent/common/exception.py
kevinclark19a: Good catch. Will do.

azurelinuxagent/common/protocol/goal_state.py
kevinclark19a: I'm leaning towards constant of the file

azurelinuxagent/common/exception.py
narrieta: thanks, also add 2 paces before the comment. pycharm is noisy about single space

azurelinuxagent/common/protocol/goal_state.py
narrieta: sounds good

tests/protocol/test_wire.py
kevinclark19a: The intention was to verify that the invalid goal state didn't affect the extensions; now that I think about it, I think there should be a check that no other extension commands were called.

narrieta: we should have a test for the GoalState class here; if you consider this is important enough to add tests specific for extensions we should move this test along with the other extension tests

azurelinuxagent/common/protocol/goal_state.py
narrieta: Logging here has the potential of being very verbose  since this code is called in each iteration of the main loop. From what we've seen so far this error does not seem to last for long, so likely we are ok as is.

Sorry, for my previous comment, I mean logging in the extension handler code (see try_update_goal_state). The verbosity issue is already handled in that code. Logging the incarnation number could be accomplished by simply adding it to the message in the exception (the code in try_update_goal_state logs only the first error, so if a different kind of error happens we wouldn't know until the operation succeeds and the error state is reset, but that is a different issue.)

If we decide to leave the logging here, let's change the error message to something like

> Fetched goal state without a RoleInstance [incarnation {inc}]".format(inc=self.incarnation))

narrieta (changes requested):
added one small comment

narrieta (approved):

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -60,14 +63,21 @@ def __init__(self, wire_client, full_goal_state=False, base_incarnation=None):
         directly.
 
         """
-        try:
-            uri = GOAL_STATE_URI.format(wire_client.get_endpoint())
+        uri = GOAL_STATE_URI.format(wire_client.get_endpoint())
+
+        for _ in range(1, NUM_GS_FETCH_RETRIES + 1):
```
larohra, line 68: Do we need a small sleep in this retry? Just to give the platform some time to actually try updating the GS

larohra (approved):

PR #2018, opened 2020-09-23
Remove enum _UpdateType and retry fetching goal state

narrieta:
Removing enum _UpdateType simplifies the code a little bit.

The retry logic fetching goal state is not needed; the underlying http APIs aready handle retry.

### DIFF
--- a/azurelinuxagent/common/protocol/goal_state.py
+++ b/azurelinuxagent/common/protocol/goal_state.py
@@ -42,10 +42,11 @@ PEM_FILE_NAME = "Certificates.pem"
 TRANSPORT_CERT_FILE_NAME = "TransportCert.pem"
 TRANSPORT_PRV_FILE_NAME = "TransportPrivate.pem"
 
+
 # too-many-instance-attributes<R0902> Disabled: The goal state consists of a good number of properties
 class GoalState(object):  # pylint: disable=R0902
-
-    def __init__(self, wire_client, full_goal_state=False, base_incarnation=None):
+    # too-many-branches<R0912> Disable: Branches are sequential, not nested
+    def __init__(self, wire_client, full_goal_state=False, base_incarnation=None):  # pylint: disable=R0912
         """
         Fetches the goal state using the given wire client.
 
--- a/azurelinuxagent/common/protocol/goal_state.py
+++ b/azurelinuxagent/common/protocol/goal_state.py
@@ -25,6 +25,7 @@ import azurelinuxagent.common.logger as logger
 from azurelinuxagent.common.AgentGlobals import AgentGlobals
 from azurelinuxagent.common.datacontract import set_properties
 from azurelinuxagent.common.event import add_event, WALAEventOperation
+from azurelinuxagent.common.exception import ProtocolError
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.protocol.restapi import Cert, CertList, Extension, ExtHandler, ExtHandlerList, \
     ExtHandlerVersionUri, RemoteAccessUser, RemoteAccessUsersList, \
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/protocol/wire.py
```diff
@@ -686,27 +686,30 @@ def _fetch_response(self, uri, headers=None, use_proxy=None):
...
+                new_goal_state = GoalState.fetch_full_goal_state(self)
+            else:
+                new_goal_state = GoalState.fetch_full_goal_state_if_incarnation_different_than(self, self._goal_state.incarnation)
+
+            if new_goal_state is not None:
+                self._goal_state = new_goal_state
+                self._save_goal_state()
+                self._update_host_plugin(new_goal_state.container_id, new_goal_state.role_config_name)
+
+        except Exception as e: # pylint: disable=C0103
```
larohra, line 711: NIT: Missed pylint error

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -58,42 +60,46 @@ def __init__(self, wire_client, full_goal_state=False, base_incarnation=None):
...
+            if not fetch_full_goal_state:
+                self.hosting_env = None
+                self.shared_conf = None
+                self.certs = None
+                self.ext_conf = None
+                self.remote_access = None
+                return
+        except Exception as exception:
+            # We don't log the error here since fetching the goal state is done every few seconds
+            raise ProtocolError(msg="Error fetching goal state", inner=exception)
```
larohra, line 98: Would it be useful to add a traceback here? Might be useful in figuring out what exactly failed when trying to fetch GS

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -122,9 +128,9 @@ def __init__(self, wire_client, full_goal_state=False, base_incarnation=None):
             else:
                 xml_text = wire_client.fetch_config(uri, wire_client.get_header_for_cert())
                 self.remote_access = RemoteAccess(xml_text)
-        except Exception as e: # pylint: disable=C0103
-            logger.warn("Fetching the goal state failed: {0}", ustr(e))
-            raise
+        except Exception as exception:
+            logger.warn("Fetching the goal state failed: {0}", ustr(exception))
+            raise ProtocolError(msg="Error fetching goal state", inner=exception)
```
larohra, line 133: same comment here

kevinclark19a (approved):

azurelinuxagent/common/protocol/goal_state.py
narrieta: I thought you were adding the traceback to error messages? (one way to do that could be in the conversion to string of the base class for exceptions)

azurelinuxagent/common/protocol/wire.py
narrieta: thanks, will fix it

azurelinuxagent/common/protocol/goal_state.py
narrieta: btw - since this is just raising the exception, the traceback shouln't be added here, but in whoever handles the exception

narrieta: and another note:

    one way to do that could be in the conversion to string of the base class for exceptions)

may lead to bad (but useful) formatting. An alternative would be to create a function in the logger to log an exception. That function could also add traceback for the inner exceptions.

kevinclark19a (approved):

azurelinuxagent/common/protocol/goal_state.py
larohra: >I thought you were adding the traceback to error messages? 

I was primarily talking about doing it on a case to case basis, but I like your idea better of adding it for every error.

> An alternative would be to create a function in the logger to log an exception.

That would be useful for sure,  maybe reuse the logger.error? But anyways not sure when I'll get time to work on that. For now do you wanna add the trace here until we have a better solution in place?

larohra (approved):
1 minor comment but else LGTM

azurelinuxagent/common/protocol/goal_state.py
narrieta: The trace should not be added here, that should be done at the place where the exception is handled.

PR #2019, opened 2020-09-23
One thread for telemetry

larohra:
## Description
This PR mainly converges sending telemetry out from a single thread instead of multiple threads sending them out separately (Monitor thread and Extension Telemetry thread). 

Main changes in the PR - 
- Send telemetry from one single thread 
- Move collection of telemetry events to one thread (unify extension telemetry thread and agent extension event collector from monitor thread)

Apart from the major changes, there are some code refactoring, minor bug fixes and linter fixes included in this PR.

Issue # 

---

### PR information


azurelinuxagent/common/event.py
```diff
@@ -581,24 +580,23 @@ def _trim_extension_event_parameters(event):
...
         Retuns a list of events that need to be sent to the telemetry pipeline and deletes the corresponding files
         from the events directory.
         """
         max_collect_errors_to_report = 5
-        event_list = TelemetryEventList()
         event_directory_full_path = os.path.join(conf.get_lib_dir(), EVENTS_DIRECTORY)
         event_files = os.listdir(event_directory_full_path)
-        unicode_error_count, unicode_errors = 0, []
-        collect_event_error_count, collect_event_errors = 0, []
+        unicode_error_count, unicode_errors = 0, set()
```
larohra, line 598: Changed the error catching to a set to ensure we catch unique errors rather than the same one

azurelinuxagent/common/event.py
```diff
@@ -631,26 +629,24 @@ def collect_events(self): # pylint: disable=R0914
                         else:
                             self._update_legacy_agent_event(event, event_file_creation_time)
 
-                    event_list.events.append(event)
+                    enqueue_event_func(event)
```
larohra, line 632: Changed the logic to more on-demand approach. Instead of using a list to capture all events and then send them, sending them out to wireserver as soon as they're captured. This on-demand mechanism would help us salvage some extra time if some event file takes longer to parse

azurelinuxagent/common/telemetryevent.py
```diff
@@ -82,33 +82,66 @@ def __eq__(self, other):
...
         self.providerId = providerId # pylint: disable=C0103
         self.parameters = DataContractList(TelemetryEventParam)
         self.file_type = ""
+        self._priority = priority
 
     # Checking if the particular param name is in the TelemetryEvent.
     def __contains__(self, param_name):
         return param_name in [param.name for param in self.parameters]
 
+    def __le__(self, other):
```
larohra, line 110: Introduced total_ordering for all events based on their priority. This helps with fetching the Events in the PriorityQueue in the correct order

azurelinuxagent/common/event.py
```diff
@@ -581,24 +580,23 @@ def _trim_extension_event_parameters(event):
...
+    def report_dropped_events_error(count, errors, operation_name, max_errors_to_report):
         err_msg_format = "DroppedEventsCount: {0}\nReasons (first {1} errors): {2}"
         if count > 0:
-            add_event(op=op,
+            add_event(op=operation_name,
                       message=err_msg_format.format(count, max_errors_to_report, ', '.join(errors)),
                       is_success=False)
 
-    def collect_events(self): # pylint: disable=R0914
+    def collect_events(self, enqueue_event_func): # pylint: disable=R0914
```
narrieta, line 590: Not sure if you looked at this linter warning. If you did, can you add a comment explaining why it is disabled? If you do this, could you also add a space before the #pylint comment (inline comments are supposed to be preceded by 2 spaces and pycharm is very noisy about this)

azurelinuxagent/common/event.py
narrieta, line 598: At this point we have separate logic to collect events for extensions and the agent. Are you planning on unifying those? It seems like there is logic (e.g. error handling) that should be shared between them.

narrieta: I have a few questions about the information we capture when we drop events; let's sync offline

azurelinuxagent/ga/monitor.py
```diff
@@ -172,14 +197,14 @@ class MonitorHandler(ThreadHandlerInterface): # pylint: disable=R0902
     def get_thread_name():
         return MonitorHandler._THREAD_NAME
 
-    def __init__(self):
+    def __init__(self, enqueue_event_func):
```
narrieta, line 200: the MonitorHandle shouldn't take the event function as parameter... the monitor is just a collection of arbitrary periodic operations and the details for each of those operations should be internal to them

azurelinuxagent/ga/telemetry_service.py
```diff
@@ -0,0 +1,122 @@
...
+        self.should_run = True
+        self._thread = None
+        self._should_process_events = threading.Event()
+        self._queue = PriorityQueue()
+
+    @staticmethod
+    def get_thread_name():
+        return TelemetryServiceHandler._THREAD_NAME
+
+    def run(self):
```
narrieta, line 54: I think the implementation of these methods is almost identical in all the thread classes. Consider moving the implementation to the base class

azurelinuxagent/ga/telemetry_service.py
```diff
@@ -0,0 +1,122 @@
...
+            self.join()
+
+    def join(self):
+        self._queue.join()
+        self._thread.join()
+
+    def stopped(self):
+        return not self.should_run
+
+    def enqueue_event_func(self, event):
```
narrieta, line 82: suggestion: remove "_func" from name

azurelinuxagent/ga/telemetry_service.py
```diff
@@ -0,0 +1,122 @@
...
+        # Add event to queue and set event
+        self._queue.put(event)
+
+        # Set the event if any enqueue happens (even if already set) to trigger sending those events
+        self._should_process_events.set()
+
+    def _process_telemetry_thread(self):
+        logger.info("Successfully started the {0} thread".format(self.get_thread_name()))
+        try:
+            # On demand wait, start processing as soon as there is any data available in the queue
```
narrieta, line 92: But we need to wait a little or we could end up sending a network request per event, wouldn't we?

azurelinuxagent/ga/telemetry_service.py
```diff
@@ -0,0 +1,122 @@
...
+            # In worst case, also keep checking every 5 mins to ensure that no data is being missed
+            while not self.stopped():
+                self._should_process_events.wait(timeout=TelemetryServiceHandler._MAX_TIMEOUT)
```
narrieta, line 95: Previously we had a bug where the agent would hang on service stop when it was doing a call to wait() on the extension's processes. Nor sure if Event would have the same problem? Could you trying issuing a service stop while this call is blocked waiting for the timeout?

azurelinuxagent/ga/telemetry_service.py
```diff
@@ -0,0 +1,122 @@
...
+
+    _THREAD_NAME = "TelemetryServiceHandler"
+    _MAX_TIMEOUT = datetime.timedelta(minutes=5).seconds
+
+    def __init__(self, protocol_util):
+        self._protocol = protocol_util.get_protocol()
+        self.should_run = True
+        self._thread = None
+        self._should_process_events = threading.Event()
+        self._queue = PriorityQueue()
```
narrieta, line 48: This should be a FIFO queue rather than a priority queue, right? we should preserver the order in which the events are created when pushing them to the wireserver.

azurelinuxagent/ga/telemetry_service.py
narrieta, line 95: Independently of service stop, the wait() affects stopped(). After calling stop(), we would be blocked until wait() times out

narrieta: Definitively watch out for that call to wait(). The documentation for Event doesn't mention anything, but the documentation for Queue.get() points out that this function is not interruptible (and, unfortunately, that is distro-dependent):

`Prior to 3.0 on POSIX systems, and for all versions on Windows, if block is true and timeout is None, this operation goes into an uninterruptible wait on an underlying lock. This means that no exceptions can occur, and in particular a SIGINT will not trigger a KeyboardInterrupt.
`

azurelinuxagent/ga/telemetry_service.py
```diff
@@ -0,0 +1,122 @@
...
+        self._queue.put(event)
+
+        # Set the event if any enqueue happens (even if already set) to trigger sending those events
+        self._should_process_events.set()
+
+    def _process_telemetry_thread(self):
+        logger.info("Successfully started the {0} thread".format(self.get_thread_name()))
+        try:
+            # On demand wait, start processing as soon as there is any data available in the queue
+            # In worst case, also keep checking every 5 mins to ensure that no data is being missed
```
narrieta, line 93: > to ensure that no data is being missed

There are some interactions where data can be missed. Likely they aren't happening in the current code, but the implementation of this class allows them. For example, the telemetry thread completes the call send_events_in_queue and is pre-empted before the check for stopped. Another thread does a call to enqueue, which puts the event and at this time another thread calls stop(). Now the telemetry thread does the check for stopped() and exits. The thread calling enqueue sets the event, but the telemetry thread already exited so the event would not be sent. Probably some sort of synchronization between stop and enqueue would be needed to prevent that scenatio.

narrieta, line 93: Regardless of the synchronization issue above, enqueue should raise if somebody tries to add an event when the telemetry thread has already been stopped.

azurelinuxagent/ga/telemetry_service.py
larohra: I made it a priority queue mainly to always prioritize agent events over extension events. When the extension telemetry pipeline is in use properly, each extension can send out upto 300 events per run. Given that each call to telemetry can take almost ~3secs and that we send approx ~25-35events per request, the thread can be blocked for quite a while if we have a lot of events. I figured it would be useful to always give ourselves the priority of using our own pipeline and keep sending the other requests on the side

narrieta: I would use a regular queue until the priority is defined. At this point the items would be sorted and since they do not have a priority who knows in what order they will be sorted and sent.

narrieta: When the priority is defined, seems like you will need two sets of priorities, with the set for the agent taking priority over the set for extensions. However, the priority of events within each of those sets should respect the order in which the events were generated,

narrieta: > since they do not have a priority

just noticed you defined an order... let me check that

narrieta: seems like it only differentiates between agent/other events... you also need to respect the timestamp

larohra: >However, the priority of events within each of those sets should respect the order in which the events were generated,

I agree about that. I was thinking of adding it but then figured it would be too much for this already big PR. I didn't do it mainly because we anyways fill the buffer up to the brim before sending out the requests and we group each request based on the ProviderID (the table it will end up in Kusto) so I figured for now this is good enough. I can add another timestamp parameter to it if you feel its a big deal though.

narrieta: also, this priority stuff may not send extensions events if the agent goes crazy and starts sending tons of events

azurelinuxagent/ga/telemetry_service.py
larohra: That's a fair point, I can set the event on `stop()` to ensure it stops blocking as soon as we want to stop the thread. We anyways wait for the queue to join so as soon as we finish executing everything in the queue, it would end the thread. Let me make this change

azurelinuxagent/ga/telemetry_service.py
```diff
@@ -0,0 +1,122 @@
...
+            finally:
+                self._queue.task_done()
+
+    def _send_events_in_queue(self):
+        # Process everything in Queue
+        if not self._queue.empty():
+            self._protocol.report_event(self._get_events_in_queue)
+
+        # Reset the event when done processing all events in queue
+        if self._should_process_events.is_set() and self._queue.empty():
```
narrieta, line 120: > and self._queue.empty()

Introduces the possibility of the telemetry thread keep sending events continuously. Previously add_event would cap the number of events to 1,000 and the monitor thread would periodically take whatever number of events were available at some point in time, send them and wait for 1 minute (or was it 5?). During that minute add_event would continue capping the number of events. Seems like now the telemetry thread can keep sending events all the time.

azurelinuxagent/ga/telemetry_service.py
narrieta: >I can add another timestamp parameter to it if you feel its a big deal though.

I think right now the code has the potential to never send old events if new ones keep coming up

larohra: That is a possibility agreed, but that would be a pain even if we have a FIFO queue (like whenever the monitor thread adds the events, we wont process anything but the agent events). Since right now we start sending as soon as we enqueue anything in the queue, I think we might be able to balance it out. 

The converse is also true though, if we get a very chatty extension our telemetry would be blocked. The best case I guess for such a scenario would be to have separate telemetry calls for both the threads separately (something like the current code) but that has its own problems I suppose.

azurelinuxagent/ga/telemetry_service.py
larohra: >Introduces the possibility of the telemetry thread keep sending events continuously.

But isint that a good thing? That is what this new thread is there for. `add_event` still caps events at 1000, I didnt change that logic. The major things that I did change were to - 
- Enqueue event as soon as we parse the file and create an object vs parsing all files and then sending out all events at once. The benefit this gives us is that now we dont have to wait to parse all files 
- Earlier since the same thread was doing parsing and sending, we would block the thread until the operation was complete and then wait another min before repeating the cycle. Now we parse the event files every minute and the telemetry thread keeps on sending it out always, ensuring that we capture more events and not miss out on anything. 

We can discuss more about this offline if you have more questions/doubts

azurelinuxagent/ga/telemetry_service.py
larohra: I agree that if the `self.should_run` property is False, we should raise and not let anyone enqueue events after that but I dont agree on raising when the thread has been stopped. Mainly because the thread can die due to some errors but we have code to restart it. The benefit of adding the Queue in the Handler object was that its independent of the lifecycle of the thread. So as soon as some other thread has an event ready, we should add it to the queue and the telemetry thread can start processing those as soon as its ready to do so. 

Even if the thread dies midway, our queue is not impacted and our data is not lost, which is an improvement from the previous design because incase the `collect_and_send_events` threw error earlier, all that unprocessed data was lost and telemetry was lost.

larohra: >Now the telemetry thread does the check for stopped() and exits.

But in the `stopped` function I'm waiting for the `queue.join()`. So if there is some data in the queue, we would not let the thread die until its processed. I totally agree on the part where we should raise and not let more events be added to the queue under certain conditions but apart from that I think the synchronization problem is taken care of.

azurelinuxagent/common/event.py
larohra: Sorry missed this one. Let me look it up

larohra: Sorry missed this one. Added a comment!

azurelinuxagent/common/event.py
larohra: both have somewhat different logics. The extension telemetry code maintains a dict of static error messages and their counts to inform Extension publishers for their breach of contract. This logic though only captures the first 5 errors thrown and are used for debugging from the agent. We can talk more about this offline

azurelinuxagent/ga/monitor.py
larohra: The main reason for doing this was trying to maintain less state. I do agree to your point but at the same time I figured it would be easier to pass out parameter rather than make the Queue a static variable (which is what this function is primary used for at the end of the day). 

1 way to mitigate this I can think of is to move CollectAndEnqueueEventsPeriodicOperation to the extension_telemetry thread. Basically have a single thread for collecting all kinds of events from disk - Agent and extensions. Does that sound like a reasonable approach?

azurelinuxagent/ga/telemetry_service.py
larohra: Fair point, let me send a different PR for it though. Dont want to club all these unrelated changes in this PR

azurelinuxagent/ga/telemetry_service.py
larohra: I was actually debating with myself on whether to add it or not and decided for it because I personally felt it made it obvious as to what this function pointer was doing (mainly because its passed around into different handlers). But I can remove that no issues!

azurelinuxagent/ga/telemetry_service.py
larohra: There will be a small chance of that, because by the time we get the request ready, another event would mostly be filled in the queue. Even if by the time our first request is ready and we dont have another event, a telemetry_event call to wireserver takes ~2-3secs which is ample amount of time to fill up the queue so that subsequent request calls would have enough data to go around. 

This was another reason to add a generator to yield events from the queue rather than saving all the events in a list and then sending them out one by one.

azurelinuxagent/ga/telemetry_service.py
larohra: Good point. Though I think if we get a SIGTERM we just `sys.exit` so I dont think any thread would matter at that point. 

```python
        if self.signal_handler is None:
            self.signal_handler = signal.signal(signal.SIGTERM, self.forward_signal)

    def forward_signal(self, signum, frame):
        if signum == signal.SIGTERM:
            self._shutdown()

        if self.child_process is None:
            return

 def _shutdown(self):
        self.running = False

        if not os.path.isfile(self._sentinel_file_path()):
            return
```

narrieta: If the wait is uninterruptible it won't respond to signals (other than KILL). We used to have a hang when trying to stop the agent service because there was a thread on wait() for extension's output.

azurelinuxagent/ga/monitor.py
narrieta: You could instantiate CollectAndEnqueueEventsPeriodicOperation in the main thread and expose a method in the MonitorHandle to add an item to the operations it monitors.

Independently of this, I mentioned before that having a single thread for telemetry seemed a better approach so I like your suggestion :)

azurelinuxagent/ga/telemetry_service.py
narrieta: ok

azurelinuxagent/ga/telemetry_service.py
larohra: Ahh interesting, didnt know that. Then do you think its better to do the `while()..sleep` approach here too? I was trying to avoid that because it wont be an on demand thing then but these other issues are a pain too

narrieta: Our bug was a hang while updating the agent when an extension's forked process was still running and we were waiting for its output (guess which extension it was :))

I learnt about uninterruptible waits while checking out how the cpu scheduler works (for the cgroups stuff) :)

The cgroups stuff made read how the cpu scheduler works, that is where I learnt about those waits :)

azurelinuxagent/ga/monitor.py
larohra: Sounds good, let me do that. Would definitely break a lot of tests lol

azurelinuxagent/ga/telemetry_service.py
larohra: That's fair, even small interrupts should be stopped, not just SIGTERMS. I'm guessing the thread.sleep() are interruptible waits and can be broken if the process wants to kill.

Though would something like this work, if we setup a signal trapper for each signal and stop() each thread specifically rather than just killing them - 

Example: 
```python
for i in [x for x in dir(signal) if x.startswith("SIG")]:
  try:
    signum = getattr(signal,i)
    signal.signal(signum, stop_all_threads())
  except (OSError, RuntimeError) as m: #OSError for Python3, RuntimeError for 2
    print ("Skipping {}".format(i))
```

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1121,33 +1138,32 @@ def report_event(self, event_list):
...
                 if len(buf[event.providerId] + event_str) >= MAX_EVENT_BUFFER_SIZE:
-                    self.send_encoded_event(event.providerId, buf[event.providerId])
+                    logger.verbose("No of events this request = {0}".format(events_per_request[event.providerId]))
+                    uni_err_count, err_count = _send_event(event.providerId)
+                    unicode_error_count += uni_err_count
+                    event_report_error_count += err_count
                     buf[event.providerId] = b''
-                    logger.verbose("No of events this request = {0}".format(events_per_request))
-                    events_per_request = 0
+                    events_per_request[event.providerId] = 0
```
kevinclark19a, line 1146: Would it be worth it to extract this functionality into a function? Though it isn't repeated in code, I still feel like giving it a name would help with readability.

azurelinuxagent/ga/monitor.py
```diff
@@ -108,6 +108,31 @@ def _operation_impl():
         logger.reset_periodic()
 
 
+class CollectAndEnqueueEventsPeriodicOperation(PeriodicOperation):
```
kevinclark19a, line 111: What's the relationship/difference between this class and the `ProcessExtensionTelemetry` class?

azurelinuxagent/ga/telemetry_service.py
kevinclark19a: I agree with @narrieta here. I'm not sure if a wait here is the answer, but I definitely think there needs to be some sort of limiter on this code path.

azurelinuxagent/common/telemetryevent.py
pgombar: nice!

azurelinuxagent/ga/monitor.py
pgombar: +1 for keeping the telemetry stuff in a single thread, would make it more clean and thread-safe!

azurelinuxagent/ga/telemetry_service.py
pgombar: Couldn't you simply modify your ordering to check both for `priority` and `time_generated` instead of just `priority` and that would solve the issue?

azurelinuxagent/ga/telemetry_service.py
pgombar: +1. Why did we abandon the batch sending approach? Wouldn't we want to keep that + eventually have QuickLog to send out important events on-demand?

azurelinuxagent/ga/monitor.py
larohra: After our last sync, we decided to keep 2 threads - One for collecting telemetry data (extension + agent) and another for sending data (telemetry service)

azurelinuxagent/ga/telemetry_service.py
larohra: So now I'm adding back a somewhat hybrid approach, where we batch for either 5secs or 30events, whichever happens first. This is to promote batching and isolation of the telemetry service both at the same time.

azurelinuxagent/ga/telemetry_service.py
larohra: After the last sync, we decided to remove the PriorityQueue altogether for now and just have a normal Queue approach. Incase later we feel that extension events are causing a lot of contention for Agent events and are blocking us heavily, we will add the PQ back with a priority + time_generated approach

azurelinuxagent/common/protocol/wire.py
larohra: I totally get what you mean, it does seem pretty convoluted. Let me see what I can do :)

azurelinuxagent/ga/monitor.py
larohra: This one is for the Agent telemetry events, the other one `ProcessExtensionTelemetry` is for a new feature where we allow extensions to send out their telemetry

azurelinuxagent/ga/telemetry_service.py
larohra: Done

azurelinuxagent/ga/telemetry_service.py
larohra: Done

azurelinuxagent/ga/telemetry_service.py
larohra: Done

azurelinuxagent/ga/telemetry_service.py
larohra: Done

azurelinuxagent/common/protocol/wire.py
larohra: Done

azurelinuxagent/ga/telemetry_service.py
larohra: Done

larohra: As per our discussion, reducing the max timeout for wait to 5 secs to ensure we dont hang for long

azurelinuxagent/ga/monitor.py
larohra: Done

azurelinuxagent/common/event.py
larohra: Kept both logics different as of now, but unified the error reporting for Reporting events and Collecting agent events

azurelinuxagent/common/telemetryevent.py
larohra: Removed the priority queue logic as of now, will introduce later if extension telemetry starts blocking agent events drastically

azurelinuxagent/ga/telemetry_service.py
larohra: Added a service stopped error to ensure we dont allow threads to add events if the telemtry service is stopped

azurelinuxagent/common/telemetryevent.py
```diff
@@ -83,51 +83,51 @@ def __eq__(self, other):
...
-    """
-    Class defining the priorities for telemetry events. Lower the number, higher the priority
-
-    Note: 0 is reserved for a feature like QuickLog in the Windows Agent (i.e. the ability to send out telemetry
-    instantly rather than waiting for a minute for the monitor thread to pick up the events)
-    """
-    AGENT_EVENT = 1 # Agent events always get the highest priority
-    EXTENSION_EVENT_NEW_PIPELINE = 2    # Prioritize extensions using the new dedicated pipeline over extensions hijacking the agent pipeline
-    EXTENSION_EVENT_OLD_PIPELINE = 3
+# class TelemetryEventPriorities(object): # pylint: disable=R0903
```
pgombar, line 86: Remove?

pgombar: Same for the rest of the code regarding prioritization that's commented out below.

larohra: Ohh shoot forgot to cleanup! Thanks for catching this, will clean this up

larohra: Done!

azurelinuxagent/ga/telemetry_service.py
```diff
@@ -38,14 +39,17 @@ class TelemetryServiceHandler(ThreadHandlerInterface):
...
+    _MIN_BATCH_WAIT_TIME = datetime.timedelta(seconds=5)
+
 
     def __init__(self, protocol_util):
         self._protocol = protocol_util.get_protocol()
         self.should_run = True
         self._thread = None
         self._should_process_events = threading.Event()
-        self._queue = PriorityQueue()
+        self._queue = Queue() #PriorityQueue()
```
pgombar, line 52: Remove comment?

azurelinuxagent/ga/telemetry_service.py
```diff
@@ -81,18 +85,27 @@ def join(self):
...
         return not self.should_run
 
-    def enqueue_event_func(self, event):
+    def enqueue_event(self, event):
         # Add event to queue and set event
-        self._queue.put(event)
+        if self.stopped():
+            raise ServiceStoppedError("{0} is stopped, not accepting anymore events".format(self.get_thread_name()))
+
+        # Queue.put() can block if the queue is full which can be an uninterruptible wait. Blocking for a max of
```
pgombar, line 93: Good to know!

pgombar: I will probably answer my own question here, but doesn't hurt to double check.

Why are we using no blocking for `get`, but blocking timeout for `put`? I assume it's because if `put` can't execute due to the queue being full, we expect it to be cleared out pretty frequently as we are sending out events. Is this true?

azurelinuxagent/ga/telemetry_service.py
```diff
@@ -38,14 +39,17 @@ class TelemetryServiceHandler(ThreadHandlerInterface):
     """
 
     _THREAD_NAME = "TelemetryServiceHandler"
-    _MAX_TIMEOUT = datetime.timedelta(minutes=5).seconds
+    _MAX_TIMEOUT = datetime.timedelta(seconds=5).seconds
+    _MIN_QUEUE_LIMIT = 30
```
pgombar, line 43: Minor suggestion: rename this for the usage, not the implementation, e.g. `_MIN_EVENT_LIMIT`.

Also, the naming `MIN` is confusing since you want it to be `< MIN`. Maybe `_MAX_EVENTS_TO_BATCH`?

tests/ga/test_collect_telemetry_events.py
```diff
@@ -527,4 +531,41 @@ def test_it_should_not_send_event_where_message_is_empty_and_report_event(self):
...
+
+    def test_it_should_not_delete_event_files_except_current_one_if_service_stopped_midway(self):
+        event_list = []
+        telemetry_handler = MagicMock(autospec=True)
+        telemetry_handler.stopped = MagicMock(return_value=False)
+        telemetry_handler.enqueue_event = MagicMock(side_effect=ServiceStoppedError("Telemetry service stopped"),
+                                                    wraps=event_list.append)
+        no_of_extensions = 3
+        # self._WELL_FORMED_FILES has 3 event files, i.e. total files for 3 extensions = 3 * 3 = 9
+        # But since we delete the file that we were processing last, expected count = 8
```
pgombar, line 556: Why do we only delete the last file? Is this test supposed to mock the service getting stopped as we are reading the last file? I'm having trouble seeing it in this test.

azurelinuxagent/common/event.py
```diff
@@ -282,6 +281,51 @@ def _log_event(name, op, message, duration, is_success=True): # pylint: disable=
         logger.info(_EVENT_MSG, name, op, message, duration)
 
 
+class EventDebugInfo(object):
```
narrieta, line 284: "EventDebugInfo" seems to generic a name. Consider renaming or adding a doc string describing what this class is meant for

azurelinuxagent/common/event.py
```diff
@@ -282,6 +281,51 @@ def _log_event(name, op, message, duration, is_success=True): # pylint: disable=
...
+                add_event(op=operation_name,
+                          message=err_msg_format.format(count, EventDebugInfo.MAX_ERRORS, ', '.join(errors)),
+                          is_success=False)
+
+        report_dropped_events_error(self.op_error_count, self.op_errors, self.op_errors_event)
+        report_dropped_events_error(self.unicode_error_count, self.unicode_errors, self.unicode_error_event)
+
+    @staticmethod
+    def _update_errors_and_get_count(error_count, errors, error):
+        error_count += 1
```
narrieta, line 316: what is the purpose of 'error_count'? should it be incremented only of the error was added? although the return value of the function doesn't seem to be used anyways... :)

narrieta: Or did you mean to use self.op_error_count? that one is initialized to 0 and never changes

azurelinuxagent/common/event.py
```diff
@@ -581,24 +625,22 @@ def _trim_extension_event_parameters(event):
...
+    def report_dropped_events_error(count, errors, operation_name, max_errors_to_report):
         err_msg_format = "DroppedEventsCount: {0}\nReasons (first {1} errors): {2}"
         if count > 0:
-            add_event(op=op,
+            add_event(op=operation_name,
                       message=err_msg_format.format(count, max_errors_to_report, ', '.join(errors)),
                       is_success=False)
 
-    def collect_events(self): # pylint: disable=R0914
+    # too-many-locals<R0914> Disabled: The number of local variables is OK
```
narrieta, line 635: we should just disable this warning globally :)

azurelinuxagent/common/event.py
```diff
@@ -631,25 +673,19 @@ def collect_events(self): # pylint: disable=R0914
...
-        EventLogger.report_dropped_events_error(unicode_error_count, unicode_errors,
-                                                WALAEventOperation.CollectEventUnicodeErrors,
-                                                max_collect_errors_to_report)
-
-        return event_list
+            except ServiceStoppedError as stopped_error:
+                logger.error(
+                    "Unable to enqueue events as service stopped: {0}, skipping events collection".format(
+                        ustr(stopped_error)))
+            except UnicodeError as uni_err:
```
narrieta, line 683: are we capturing enough information to identify the specific events that triggered these errors?

azurelinuxagent/common/event.py
```diff
@@ -775,8 +811,8 @@ def add_periodic(delta, name, op=WALAEventOperation.Unknown, is_success=True, du
                           message=message, log_event=log_event, force=force)
 
 
-def collect_events(reporter=__event_logger__):
-    return reporter.collect_events()
+def collect_events(enqueue_event, reporter=__event_logger__):
```
narrieta, line 814: suggestion: you may want to do a couple of renames here:

   collect_events doesn't collect events anymore, now it just applies a function to them... process_events?
   enqueue_event doesn't have to enqueue events, it is an arbitrary function that is applied to events... proceess_event_opeation?

narrieta (commented):
comments for event.py

azurelinuxagent/common/exception.py
```diff
@@ -227,6 +227,14 @@ def __init__(self, msg=None, inner=None):
         super(InvalidExtensionEventError, self).__init__(msg, inner)
 
 
+class ServiceStoppedError(AgentError):
+    """
+    Error thrown when trying to access a Servive which is stopped
```
narrieta, line 232: typo: Service

narrieta (commented):
comments for exception.py

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1079,20 +1077,28 @@ def send_encoded_event(self, provider_id, event_str, encoding='utf-8'):
             raise ProtocolError(
                 "Failed to send events:{0}".format(resp.status))
 
-    def report_event(self, event_list):
-        max_send_errors_to_report = 5
+    def report_event(self, get_events_in_queue):
```
narrieta, line 1080: consider renaming get_events_in_queue... event_iterator?

azurelinuxagent/ga/telemetry_service.py
larohra: Shoot missed this! Done :)

azurelinuxagent/ga/telemetry_service.py
larohra: For get, we're already using an `thread.Event` for synchronization, plus there are additional checks (like empty checks) in the code which take care of ensuring we dont try to fetch from the queue if its empty.

For put, as you mentioned, that is the expectation. And if the queue is still full, we just mark the service as stopped to ensure no new events are added either. 

Hope this clarifies your doubts

azurelinuxagent/ga/telemetry_service.py
larohra: Ahh makes sense. I named it `MIN` because this is the min number of events we will wait for, `MAX` seems counterintuitive. I'll rename it to something like `_MIN_EVENT_TO_BATCH` as it makes more sense given the use case.

tests/ga/test_collect_telemetry_events.py
larohra: This is specifically by design. The idea is basically that if we're processing an event file and the service is stopped midway, we dont want to process that file again to avoid event duplication, so we delete it. (Note: In the extension telemetry events, 1 file can contain multiple events). The tradeoff is that we would loose the remaining events in that particular file but avoid duplication

azurelinuxagent/common/event.py
```diff
@@ -282,6 +281,51 @@ def _log_event(name, op, message, duration, is_success=True): # pylint: disable=
         logger.info(_EVENT_MSG, name, op, message, duration)
 
 
+class EventDebugInfo(object):
+    MAX_ERRORS = 5
```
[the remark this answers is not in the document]
larohra: Good point, I thought I was using it externally but doesnt look like it. 
Done

azurelinuxagent/common/event.py
```diff
@@ -282,6 +281,51 @@ def _log_event(name, op, message, duration, is_success=True): # pylint: disable=
...
+    OP_REPORT = "Report"
+    OP_COLLECT = "Collect"
+
+    def __init__(self, operation=OP_REPORT):
+        self.unicode_error_count = 0
```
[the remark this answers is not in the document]
larohra: Done

azurelinuxagent/common/event.py
larohra: Done

azurelinuxagent/common/event.py
larohra: Haha I agree! Let me do it in another PR though just to keep both the issues isolated.

azurelinuxagent/common/event.py
larohra: Umm no, we're not capturing any data about the specific events that can cause the errors, we're mostly capturing the data on what failed.
For extension events though, the errors are logged based on per extension basis, so we atleast know which extension added the bad event (this is only for collecting, not sending telemetry)

azurelinuxagent/common/exception.py
larohra: Done

azurelinuxagent/common/event.py
larohra: Done

azurelinuxagent/ga/telemetry_service.py
larohra: Done

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1101,33 +1107,28 @@ def report_event(self, event_list):
...
-                events_per_request += 1
-            except UnicodeError as e: # pylint: disable=C0103
-                unicode_error_count += 1
-                if len(unicode_errors) < max_send_errors_to_report:
-                    unicode_errors.append(ustr(e))
-            except Exception as e: # pylint: disable=C0103
-                event_report_error_count += 1
-                if len(event_report_errors) < max_send_errors_to_report:
-                    event_report_errors.append(ustr(e))
+                events_per_request[event.providerId] += 1
```
narrieta, line 1120: why is buf initialized to b'' when we concatenate strings?

narrieta (commented):
comments for wire.py

azurelinuxagent/common/event.py
narrieta: For extension events though, the errors are logged based on per extension basis,

That would be for extensions using the new pipeline, but not for those using the current pipeline.

azurelinuxagent/common/telemetryevent.py
```diff
@@ -99,16 +99,12 @@ def is_extension_event(self):
         # "WALinuxAgent", it is an extension event.
         for param in self.parameters:
             if param.name == GuestAgentExtensionEventsSchema.Name:
-                return param.value != AGENT_NAME
+                if param.value != AGENT_NAME:
```
narrieta, line 102: wow. thanks for fixing this

azurelinuxagent/common/event.py
larohra: >That would be for extensions using the new pipeline, but not for those using the current pipeline.

Fair enough, though capturing data would be hard. Like for example if the file contains bad_json we wont be able to parse it, and the only metadata we can capture would be the name. But that name wont tell us enough (its something like 123456.tld). 

What meta-data do you think would be useful here? I can try to capture as much as I can.

azurelinuxagent/common/telemetryevent.py
larohra: ohh lol I think I added something here regarding that priority queue initially but removed it. Let me revert this change

azurelinuxagent/common/protocol/wire.py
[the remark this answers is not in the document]
larohra: Done

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1079,20 +1077,28 @@ def send_encoded_event(self, provider_id, event_str, encoding='utf-8'):
...
 
-    def report_event(self, event_list):
-        max_send_errors_to_report = 5
+    def report_event(self, get_events_in_queue):
         buf = {}
-        events_per_request = 0
-        unicode_error_count, unicode_errors = 0, []
-        event_report_error_count, event_report_errors = 0, []
+        debug_info = EventDebugInfo(operation=EventDebugInfo.OP_REPORT)
+        events_per_request = defaultdict(int)
```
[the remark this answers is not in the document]
larohra: Done

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1079,20 +1077,28 @@ def send_encoded_event(self, provider_id, event_str, encoding='utf-8'):
...
         # Group events by providerId
-        for event in event_list.events:
+        for event in get_events_in_queue():
             try:
                 if event.providerId not in buf:
                     buf[event.providerId] = b''
                 event_str = event_to_v1_encoded(event)
+
                 if len(event_str) >= MAX_EVENT_BUFFER_SIZE:
+                    # Ignore single events that are too large to send out
```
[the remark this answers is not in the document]
larohra: This is legacy code, its always been this case. But I do remember we truncate messages, cant find the code though for some reason, let me look around more

azurelinuxagent/common/protocol/wire.py
larohra: This change was added when I made the changes for the unicode stuff. Basically we encode the xml string that we need to send out and then save it in the buffer. Since we keep adding the encoded string to the buffer, the buffer had to be initialized with an encoded string (`b""`)

azurelinuxagent/common/telemetryevent.py
larohra: Done

azurelinuxagent/common/protocol/wire.py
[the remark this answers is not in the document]
larohra: Looks like we only trim stdout/stderr from the process output, no other trimming done for the events (I do trim messages for the new pipeline but couldn't find anything for the existing legacy events)

narrieta (commented):
Some additional comments

azurelinuxagent/ga/collect_telemetry_events.py
```diff
@@ -23,22 +23,26 @@
...
-from azurelinuxagent.ga.exthandlers import HANDLER_NAME_PATTERN
+from azurelinuxagent.common.interfaces import ThreadHandlerInterface
+from azurelinuxagent.ga.exthandlers import HANDLER_NAME_PATTERN, is_extension_telemetry_pipeline_enabled
 from azurelinuxagent.ga.periodic_operation import PeriodicOperation
 
 
-def get_extension_telemetry_handler(protocol_util):
-    return ExtensionTelemetryHandler(protocol_util)
+def get_telemetry_collector_handler(telemetry_service_handler):
+    return CollectTelemetryEventsHandler(telemetry_service_handler)
```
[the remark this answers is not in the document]
larohra: Sure, renamed

azurelinuxagent/ga/telemetry_service.py
```diff
@@ -0,0 +1,146 @@
...
+from azurelinuxagent.common.exception import ServiceStoppedError
+from azurelinuxagent.common.future import ustr, Queue, Full
+from azurelinuxagent.common.interfaces import ThreadHandlerInterface
+
+
+def get_telemetry_service_handler(protocol_util):
+    return TelemetryServiceHandler(protocol_util)
+
+
+class TelemetryServiceHandler(ThreadHandlerInterface):
```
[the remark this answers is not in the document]
larohra: Renamed to `SendTelemetryEventsHandler`

azurelinuxagent/ga/collect_telemetry_events.py
```diff
@@ -348,22 +370,52 @@ def add_common_params_to_extension_event(event, event_time):
         reporter.add_common_event_parameters(event, event_time)
 
 
-class ExtensionTelemetryHandler(ThreadHandlerInterface):
+class CollectAndEnqueueEventsPeriodicOperation(PeriodicOperation):
```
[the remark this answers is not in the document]
larohra: Logically that makes sense, but wont that make other things more complicated unnecessarily? (Like testing for example, calling a private class would not be ideal and would look dirty)

azurelinuxagent/ga/collect_telemetry_events.py
```diff
@@ -23,22 +23,26 @@
...
 from azurelinuxagent.ga.periodic_operation import PeriodicOperation
 
 
-def get_extension_telemetry_handler(protocol_util):
-    return ExtensionTelemetryHandler(protocol_util)
+def get_telemetry_collector_handler(telemetry_service_handler):
+    return CollectTelemetryEventsHandler(telemetry_service_handler)
 
+
+# too-few-public-methods<R0903> Disabled: This class is used as an Enum
```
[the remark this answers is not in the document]
larohra: From my understanding, a schema is basically an organization or structure for something. Keeping that in mind, this does look like a schema right? What name did you have in mind for it?

azurelinuxagent/ga/collect_telemetry_events.py
```diff
@@ -390,15 +442,24 @@ def stopped(self):
         return not self.should_run
 
     def daemon(self):
-        op = ProcessExtensionTelemetry(self.protocol_util) # pylint: disable=C0103
+        periodic_operations = [
+            CollectAndEnqueueEventsPeriodicOperation(self._telemetry_service_handler)
```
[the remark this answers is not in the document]
larohra: Renamed to `ProcessExtensionEventsPeriodicOperation`

azurelinuxagent/ga/telemetry_service.py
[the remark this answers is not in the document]
larohra: I didn't add it initially because this thread was just doing 1 single operation (which is not even a PeriodicOperation). On a deeper inspection, I realized I'm using certain things at a ThreadHandler level (like Queue, Event, etc) which might make the decoupling a bit more tedious. 

Given that the currently the thread is just doing one thing and purely from the perspective of not wanting to do extra work right now, I'm going to skip it but if any of you guys feel strongly about it let me know I'll change it :)

azurelinuxagent/ga/collect_telemetry_events.py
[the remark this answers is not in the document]
larohra: Renamed to `ProcessExtensionEventsPeriodicOperation`

azurelinuxagent/ga/collect_telemetry_events.py
```diff
@@ -69,44 +73,48 @@ class ProcessExtensionTelemetry(PeriodicOperation):
     _EXTENSION_EVENT_REQUIRED_FIELDS = [attr.lower() for attr in dir(ExtensionEventSchema) if
                                         not callable(getattr(ExtensionEventSchema, attr)) and not attr.startswith("__")]
 
-    def __init__(self, protocol_util):
+    def __init__(self, telemetry_service_handler):
```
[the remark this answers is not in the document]
larohra: Same reasoning as above

azurelinuxagent/ga/collect_telemetry_events.py
```diff
@@ -348,22 +370,52 @@ def add_common_params_to_extension_event(event, event_time):
...
+        """
+        Periodically send any events located in the events folder
+        """
+        try:
+            if self._telemetry_service_handler.stopped():
+                logger.warn("{0} service is not running, skipping iteration.".format(
+                    self._telemetry_service_handler.get_thread_name()))
+                return
+
+            process_events(self._telemetry_service_handler.enqueue_event)
```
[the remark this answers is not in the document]
larohra: Done

azurelinuxagent/ga/collect_telemetry_events.py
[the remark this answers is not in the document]
larohra: synced offline, not needed anymore

azurelinuxagent/ga/collect_telemetry_events.py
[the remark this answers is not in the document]
narrieta: The class should be public only if it is part of the interface of the module. It seems to me that this class is only part of the module's implementation.

If the class is not part of the interface, do not make it public for the benefit of the unit test. It is ok for the unit tests for A to reach (on a limited basis) into the private implementation of A. Some of our unit tests abuse this and the the unit tests for A use private implementation of D, when A uses B uses C uses D. Those we should avoid.

    Additionally this is also how the Monitor thread has its PeriodicOperations 

It shouldn't, shame on me :)

azurelinuxagent/ga/collect_telemetry_events.py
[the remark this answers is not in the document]
narrieta: It seems just like an enumeration of properties, with no organization or structure. PropertyNames? or just add a sample event as a comment to illustrate the structure (I assume it is just a flat object, with no nested properties)

azurelinuxagent/ga/telemetry_service.py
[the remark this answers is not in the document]
narrieta: When I first saw the use of SendTelemetryEventsHandler and CollectTelemetryEventsHandler from the main thread this threw me off. 

The main thread uses those handlers to manage (start/stop/query) the corresponding threads so I was surprised that one thread had a dependency on another. I had to dig more into the code to realize that CollectTelemetryEventsHandler doesn't really care about SendTelemetryEventsHandler's thread, but about the service that is running on that thread.

You can leave as is if you prefer so.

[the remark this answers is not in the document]
larohra: haha yeah its all decoupled. Is there any comment or anything I can add to make it more clearer? To avoid someone else having the same confusion you had basically

azurelinuxagent/ga/collect_telemetry_events.py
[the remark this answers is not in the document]
larohra: Fair enough, let me change the code and send an iteration :)

narrieta (commented):
a couple of comments

azurelinuxagent/ga/telemetry_service.py
```diff
@@ -0,0 +1,146 @@
...
+    _MIN_EVENTS_TO_BATCH = 30
+    _MIN_BATCH_WAIT_TIME = datetime.timedelta(seconds=5)
+
+
+    def __init__(self, protocol_util):
+        self._protocol = protocol_util.get_protocol()
+        self.should_run = True
+        self._thread = None
+        self._should_process_events = threading.Event()
+        self._queue = Queue()
```
narrieta, line 52: let's continue our offline discussion about using an unbounded queue (no size limit)

azurelinuxagent/ga/telemetry_service.py
```diff
@@ -0,0 +1,146 @@
...
+    return SendTelemetryEventsHandler(protocol_util)
+
+
+class SendTelemetryEventsHandler(ThreadHandlerInterface):
+    """
+    This Handler takes care of sending all telemetry out of the agent to Wireserver. It sends out data as soon as
+    there's any data available in the queue to send.
+    """
+
+    _THREAD_NAME = "SendTelemetryEventsHandler"
```
narrieta, line 41: since this goes to the log, maybe shorten it to "SendTelemetry"?

azurelinuxagent/ga/telemetry_service.py
narrieta: The batching addresses my concern. Thanks for adding it.

azurelinuxagent/ga/telemetry_service.py
```diff
@@ -0,0 +1,146 @@
...
+                self._should_process_events.wait(timeout=SendTelemetryEventsHandler._MAX_TIMEOUT)
+                self._send_events_in_queue()
+
+        except Exception as error:
+            err_msg = "An unknown error occurred in the {0} thread main loop, stopping thread. Error: {1}, Stack: {2}".format(
+                self.get_thread_name(), ustr(error), traceback.format_exc())
+            add_event(op=WALAEventOperation.UnhandledError, message=err_msg, is_success=False)
+
+    def _get_events_in_queue(self):
+        while not self._queue.empty():
```
narrieta, line 119: should this be checking for stopped() as well?

azurelinuxagent/ga/telemetry_service.py
pgombar: All clear now, thanks!

tests/ga/test_collect_telemetry_events.py
pgombar: Ok, now I understand the intent of the test better, but I'm still having issues understanding the implementation. Where are you stopping the service midway?

I see this, but it was added at the beginning:
`telemetry_handler.enqueue_event = MagicMock(side_effect=ServiceStoppedError("Telemetry service stopped"),
                                                    wraps=event_list.append)`

azurelinuxagent/common/utils/extensionprocessutil.py
```diff
@@ -103,7 +103,7 @@ def format_stdout_stderr(stdout, stderr):
     """
     Format stdout and stderr's output to make it suitable in telemetry.
     The goal is to maximize the amount of output given the constraints
-    of telemetry.
+    of telemetry. too-few-public-methods<R0903> Disabled: This class is used as an Enum
```
pgombar, line 106: Are you missing `# pylint: disable=` here?

azurelinuxagent/ga/telemetry_service.py
larohra: I specifically didnt add it here. The whole idea is that we should clear the queue even if the service is asked to stop to avoid data loss. We won't let anyone enqueue anymore events if `self.stopped` is True ensuring that the queue is not never ending.

tests/ga/test_collect_telemetry_events.py
larohra: Basically in this test I'm raising a `ServiceStoppedError` whenever someone tries to enqueue any event using the `telemetry_handler.enqueue_event` function. Since this function is only called midway into the collect_events service, this is how I'm ensuring that its called in the middle :)

azurelinuxagent/ga/collect_telemetry_events.py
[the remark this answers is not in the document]
larohra: Done

azurelinuxagent/ga/collect_telemetry_events.py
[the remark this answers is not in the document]
larohra: Added a sample in comments, hopefully that clears it up

azurelinuxagent/ga/telemetry_service.py
```diff
@@ -0,0 +1,146 @@
...
+        self._thread.setName(self.get_thread_name())
+        self._thread.start()
+
+    def stop(self):
+        """
+        Stop server communication and join the thread to main thread.
+        """
+        self.should_run = False
+        # Set the event to unblock the thread to ensure that the thread is not blocking shutdown.
+        self._should_process_events.set()
```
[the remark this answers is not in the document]
larohra: Yup that's correct. That's an open gap that we have currently. Added a Todo in our code capturing the gap (is a TFS task also needed?)

azurelinuxagent/ga/telemetry_service.py
```diff
@@ -0,0 +1,146 @@
...
+                # If the thread is requested to stop midway, we skip batching and send whatever we have in the queue.
+                logger.verbose("Waiting for events to batch. Queue size: {0}, Time elapsed: {1} secs",
+                               self._queue.qsize(), (datetime.datetime.utcnow() - start_time).seconds)
+                time.sleep(1)
+            self._protocol.report_event(self._get_events_in_queue)
+
+        # Reset the event when done processing all events in queue
+        if self._should_process_events.is_set() and self._queue.empty():
+            logger.verbose("Resetting the event")
+            self._should_process_events.clear()
```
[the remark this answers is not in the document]
larohra: Synced offline, changed the synchronisation from Event based to Queue based

azurelinuxagent/ga/telemetry_service.py
larohra: Synced offline, decided to leave it unbounded for now. Captured a Todo in our code for this gap.

azurelinuxagent/ga/telemetry_service.py
larohra: Just to keep it consistent and a bit short, renamed to "SendTelemetryHandler"

azurelinuxagent/common/utils/extensionprocessutil.py
larohra: I think this was a typo, removed

azurelinuxagent/ga/collect_telemetry_events.py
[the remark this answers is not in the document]
larohra: Done

larohra:
Green DCR run (except for 1 known setup failure) - https://tuxgold.corp.microsoft.com/job/AzLinux/job/DungeonCrawler/job/DungeonCrawler.larohra3/168/

azurelinuxagent/ga/send_telemetry_events.py
```diff
@@ -48,7 +48,15 @@ def __init__(self, protocol_util):
         self._protocol = protocol_util.get_protocol()
         self.should_run = True
         self._thread = None
-        self._should_process_events = threading.Event()
+        # self._should_process_events = threading.Event()
```
narrieta, line 51: nit: remove commented out code?

azurelinuxagent/ga/send_telemetry_events.py
```diff
@@ -92,55 +98,65 @@ def enqueue_event(self, event):
...
+        # limitations are set for the Queue, ensure to handle that correctly here.
         try:
             self._queue.put(event, timeout=SendTelemetryEventsHandler._MAX_TIMEOUT)
-        except Full as error:
+        except Exception as error:
             raise ServiceStoppedError(
-                "Queue full, stopping any more enqueuing until the next run. {0}".format(ustr(error)))
+                "Unable to enqueue due to: {0}, stopping any more enqueuing until the next run".format(ustr(error)))
+
+    def _wait_for_event_in_queue(self):
```
narrieta, line 110: Since None is so special, let's add a comment here describing when this function returns None

azurelinuxagent/ga/send_telemetry_events.py
larohra: Thanks for catching! Cleaned up!

azurelinuxagent/ga/send_telemetry_events.py
larohra: Done

narrieta (approved):

azurelinuxagent/ga/send_telemetry_events.py
```diff
@@ -48,11 +48,15 @@ def __init__(self, protocol_util):
         self._protocol = protocol_util.get_protocol()
         self.should_run = True
         self._thread = None
-        self._should_process_events = threading.Event()
+        # self._should_process_events = threading.Event()
```
pgombar: Remove?

pgombar: Nevermind, wasn't looking at your latest commit. This is outdated.

azurelinuxagent/ga/collect_telemetry_events.py
```diff
@@ -0,0 +1,585 @@
...
+                try:
+                    os.remove(os.path.join(event_dir_path, residue_file))
+                except Exception as error:
+                    # Only log the first error once per handler per run if unable to clean off residue files
+                    err = ustr(error) if err is None else err
+
+                if err is not None:
+                    logger.error("Failed to completely clear the {0} directory. Exception: {1}", event_dir_path, err)
+
+    def _get_captured_events_count(self, handler_name, event_file_path, captured_events_count,
```
pgombar, line 253: This method does much more than retrieve the count of captured events. I was finding it hard to figure out where we actually enqueue events, and the name does not help here. Consider renaming or breaking the function into smaller pieces of logic.

azurelinuxagent/ga/collect_telemetry_events.py
```diff
@@ -0,0 +1,585 @@
...
+    EventLevel = "EventLevel"
+    Message = "Message"
+    EventPid = "EventPid"
+    EventTid = "EventTid"
+    OperationId = "OperationId"
+
+
+class _ProcessExtensionEventsPeriodicOperation(PeriodicOperation):
+    """
+    Periodic operation for collecting and sending extension telemetry events to Wireserver.
```
pgombar, line 74: This doesn't really send the events to wireserver, it just enqueues them for the other thread to send them.

azurelinuxagent/ga/collect_telemetry_events.py
```diff
@@ -0,0 +1,585 @@
...
+            return
+
+        # Add the remaining params to the event
+        for param_name in replace_or_add_params:
+            event.parameters.append(TelemetryEventParam(param_name, replace_or_add_params[param_name]))
+
+
+class _CollectAndEnqueueEventsPeriodicOperation(PeriodicOperation):
+    """
+    Periodic operation to collect and send telemetry events located in the events folder
```
pgombar, line 387: Suggestion to clarify what "send" means -> to the common event queue, not to wireserver.

azurelinuxagent/ga/collect_telemetry_events.py
```diff
@@ -0,0 +1,585 @@
...
+        Periodically send any events located in the events folder
+        """
+        try:
+            if self._send_telemetry_events_handler.stopped():
+                logger.warn("{0} service is not running, skipping iteration.".format(
+                    self._send_telemetry_events_handler.get_thread_name()))
+                return
+            self.process_events()
+        except Exception as error:
+            err_msg = "Failure in collecting Agent events: {0}".format(ustr(error))
```
pgombar, line 410: These are not just agent events, right? While the extensions are not using the new pipeline, this will contain all events.

pgombar (commented):
Some comments and suggestions around improving the clarity of some parts of the code. Also synced offline on PEP8 style conventions for whitespace (less important).

azurelinuxagent/ga/collect_telemetry_events.py
larohra: I think I was being over-smart with the name. I was going for `captured_events` and trying to imply that it already enqueued lol. But I get your point, renamed :)

azurelinuxagent/ga/collect_telemetry_events.py
larohra: Modified the comment

azurelinuxagent/ga/collect_telemetry_events.py
larohra: Modified the comments

azurelinuxagent/ga/collect_telemetry_events.py
larohra: Yes true but they're going to be transitioned to the new pipeline soon. It would make the message too complicated if I wrote something like "Failure in Agent/legacy extension events using old pipeline" or something. Keeping it simple basically :)

narrieta (approved):

azurelinuxagent/ga/collect_telemetry_events.py
pgombar: Sure, but the message is misleading until ETP is enabled *and* until all extensions move to it. I'm concerned because this will be flagged as ERROR in the agent code and support/customers unfamiliar with the implementation might get the wrong idea from this message (like, what's happening to my extension's events?).

Today, this is the only avenue for events collection.

"Failure in collecting events using the legacy telemetry pipeline" captures the gist of it.

larohra: > legacy telemetry pipeline

This is an internal concept though, like for the agent telemetry its not legacy, its current. Its only legacy for extensions events still using this pipeline. 

To tackle this exact behavior, I also plan on changing the location of our `events` directory to ensure these legacy extensions are forced to move. But we cant do that without proper communication. I can just rename it to `Failure in collecting telemetry events` if that sounds better to you? The ETP pipeline specifies `Extension Events` so that way we can be sure its not the new one,

pgombar: Sounds good to me! You are right about the legacy comment, it is a technical detail.

larohra: Done

pgombar (approved):

larohra:
Green DCR run (except for the known agent-slb issue) - https://tuxgold.corp.microsoft.com/job/AzLinux/job/DungeonCrawler/job/DungeonCrawler.larohra3/169/

narrieta (approved):

PR #2022, opened 2020-09-24
Switch to run command

kevinclark19a:
## Description

Switched a set of `shellutil.run_get_output` and `shellutil.run` callers to the new `shellutil.run_command` function.

---

### PR information


azurelinuxagent/common/osutil/default.py
```diff
@@ -554,7 +554,12 @@ def set_selinux_context(self, path, con): # pylint: disable=R1710
             if not os.path.exists(path):
                 logger.error("Path does not exist: {0}".format(path))
                 return 1
-            return shellutil.run('chcon ' + con + ' ' + path)
+            
+            try:
+                shellutil.run_command(['chcon', con, path])
+            except shellutil.CommandError as cmd_err:
```
narrieta, line 560: shellutil logs errors by default... we would need to log here to match the original code

azurelinuxagent/common/osutil/default.py
```diff
@@ -630,9 +635,11 @@ def umount_dvd(self, chk_err=True, mount_point=None):
...
     def eject_dvd(self, chk_err=True):
         dvd = self.get_dvd_device()
-        retcode = shellutil.run("eject {0}".format(dvd))
-        if chk_err and retcode != 0:
-            raise OSUtilError("Failed to eject dvd: ret={0}".format(retcode))
+        try:
+            shellutil.run_command(["eject", dvd])
+        except shellutil.CommandError as cmd_err:
+            if chk_err:
+                raise OSUtilError("Failed to eject dvd: ret={0}".format(cmd_err.returncode))
```
narrieta, line 642: we should probably add the stdout/stderr of the command (the original code was logging errors)

azurelinuxagent/common/osutil/default.py
```diff
@@ -1112,15 +1126,17 @@ def set_dhcp_hostname(self, hostname):
...
+                shellutil.run_command(["ifup", ifname])
                 return
-            logger.warn("failed to restart {0}: return code {1}".format(ifname, return_code))
-            if attempt < retry_limit:
-                logger.info("retrying in {0} seconds".format(wait))
-                time.sleep(wait)
-            else:
-                logger.warn("exceeded restart retries")
+            except shellutil.CommandError as cmd_err:
+                logger.warn("failed to restart {0}: return code {1}".format(ifname, cmd_err.returncode))
```
narrieta, line 1134: we should add the stdout/stderr of the command

azurelinuxagent/common/utils/cryptutil.py
```diff
@@ -41,11 +41,11 @@ def gen_transport_cert(self, prv_file, crt_file):
...
-        cmd = ("{0} req -x509 -nodes -subj /CN=LinuxTransport -days 730 "
-               "-newkey rsa:2048 -keyout {1} "
-               "-out {2}").format(self.openssl_cmd, prv_file, crt_file)
-        rc = shellutil.run(cmd) # pylint: disable=C0103
-        if rc != 0:
+        cmd = [self.openssl_cmd, "req", "-x509", "-nodes", "-subj", "/CN=LinuxTransport", 
+            "-days", "730", "-newkey", "rsa:2048", "-keyout", prv_file, "-out", crt_file]
+        try:
+            shellutil.run_command(cmd)
+        except shellutil.CommandError:
```
narrieta, line 48: log stdout/stderr

azurelinuxagent/common/utils/cryptutil.py
```diff
@@ -79,18 +79,26 @@ def decrypt_p7m(self, p7m_file, trans_prv_file, trans_cert_file, pem_file):
...
+            first_cmd = [self.openssl_cmd, "cms", "-decrypt", "-in", p7m_file, "-inkey",
+                trans_prv_file, "-recip", trans_cert_file]
+            second_cmd = [self.openssl_cmd, "pkcs12", "-nodes", "-password", "pass:", 
+                "-out", pem_file]
+
+            first_proc = subprocess.Popen(first_cmd, stdout=subprocess.PIPE)
+            second_proc = subprocess.Popen(second_cmd, stdin=first_proc.stdout, stdout=subprocess.PIPE)
+            second_proc.wait()
+
+            if first_proc.returncode or second_proc.returncode:
```
narrieta, line 91: let's do an explicit comparison against 0

azurelinuxagent/common/utils/cryptutil.py
```diff
@@ -79,18 +79,26 @@ def decrypt_p7m(self, p7m_file, trans_prv_file, trans_cert_file, pem_file):
...
-            rc = shellutil.run(cmd) # pylint: disable=C0103
-            if rc != 0:
+            first_cmd = [self.openssl_cmd, "cms", "-decrypt", "-in", p7m_file, "-inkey",
+                trans_prv_file, "-recip", trans_cert_file]
+            second_cmd = [self.openssl_cmd, "pkcs12", "-nodes", "-password", "pass:", 
+                "-out", pem_file]
+
+            first_proc = subprocess.Popen(first_cmd, stdout=subprocess.PIPE)
+            second_proc = subprocess.Popen(second_cmd, stdin=first_proc.stdout, stdout=subprocess.PIPE)
+            second_proc.wait()
```
narrieta, line 89: we should use communicate() instead to get stdout/stderr and log it on failure

azurelinuxagent/common/utils/cryptutil.py
```diff
@@ -79,18 +79,26 @@ def decrypt_p7m(self, p7m_file, trans_prv_file, trans_cert_file, pem_file):
...
+            if first_proc.returncode or second_proc.returncode:
                 logger.error("Failed to decrypt {0}".format(p7m_file))
+            
 
     def crt_to_ssh(self, input_file, output_file):
-        shellutil.run("ssh-keygen -i -m PKCS8 -f {0} >> {1}".format(input_file,
-                                                                    output_file))
+        try:
+            with open(output_file, "a") as file_out:
+                output = shellutil.run_command(["ssh-keygen", "-i -m PKCS8 -f", input_file])
```
narrieta, line 98: I would feel more comfortable if we use Popen directly here (as we did in the previous function). run_command has some extra code to encode the command's output and, though I do not see any issue right now, I think it would be safer to use Popen.

azurelinuxagent/common/osutil/default.py
narrieta: There is small difference here in that the previous code would log commands errors as INFO if the return code was 1, and as ERROR otherwise.
Since anyways we are logging a WARN regardless of the return code, I think we are OK here (as long as we add stderr/stdout to the message)

narrieta (changes requested):
Added comments

narrieta (approved):

azurelinuxagent/common/utils/cryptutil.py
```diff
@@ -79,18 +86,36 @@ def decrypt_p7m(self, p7m_file, trans_prv_file, trans_cert_file, pem_file):
...
-            rc = shellutil.run(cmd) # pylint: disable=C0103
-            if rc != 0:
-                logger.error("Failed to decrypt {0}".format(p7m_file))
+            first_cmd = [self.openssl_cmd, "cms", "-decrypt", "-in", p7m_file, "-inkey",
+                trans_prv_file, "-recip", trans_cert_file]
+            second_cmd = [self.openssl_cmd, "pkcs12", "-nodes", "-password", "pass:", 
+                "-out", pem_file]
+
+            first_proc = subprocess.Popen(first_cmd, stdout=subprocess.PIPE)
+            second_proc = subprocess.Popen(second_cmd, stdin=first_proc.stdout, stdout=subprocess.PIPE)
```
larohra, line 95: Good way of solving the problem! I'm sure you must've already done this, but you confirmed this has no regressions right? And DCR passed too?

azurelinuxagent/common/utils/cryptutil.py
```diff
@@ -79,18 +86,36 @@ def decrypt_p7m(self, p7m_file, trans_prv_file, trans_cert_file, pem_file):
...
+            stdout, stderr = second_proc.communicate()
+
+            if first_proc.returncode != 0 or second_proc.returncode != 0:
+                msg = """Failed to decrypt {0}
+                [stdout]
+                {1}
+
+                [stderr]
+                {2}
+                """.format(p7m_file, stdout, stderr)
```
larohra, line 105: Make sure to encode the stdout/stderr before writing it to log (the run_command function always does that for you)

tests/common/osutil/test_default.py
```diff
@@ -56,15 +56,22 @@ def test_restart(self):
...
             # assert
             self.assertEqual(run_patch.call_count, retries)
-            self.assertEqual(run_patch.call_args_list[0][0][0], 'ifdown {0} && ifup {0}'.format(ifname))
+
+            cmd_queue = list(args[0] for (args, _) in run_patch.call_args_list)            
+            while cmd_queue:
+                self.assertEqual(cmd_queue.pop(0), ["ifdown", ifname])
+                # We don't expect the following command to be called because 'dummy' does
+                # not exist.
+                # self.assertEqual(cmd_queue.pop(0), ["ifup", ifname])
```
larohra, line 73: If you dont expect this, you should add an assert that it's actually not called rather than a comment :)

azurelinuxagent/common/osutil/default.py
```diff
@@ -1112,15 +1135,26 @@ def set_dhcp_hostname(self, hostname):
...
+            except shellutil.CommandError as cmd_err:
+                
+                msg = """failed to restart {0}: returncode={1}
+                [stdout]
+                {2}
+                
+                [stderr]
+                {3}
+                """.format(ifname, cmd_err.returncode, cmd_err.stdout, cmd_err.stderr)
+                logger.warn(msg)
```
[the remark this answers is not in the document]
pgombar: Isn't this logged in L1154?

tests/common/osutil/test_default.py
```diff
@@ -99,16 +106,15 @@ def test_mount_dvd_failure(self, _):
...
+            with patch.object(shellutil, 'run_command',
+                              side_effect=shellutil.CommandError("mount dvd", 1, "", msg)) as patch_run:
                 with patch.object(os, 'makedirs'):
                     try:
                         osutil.DefaultOSUtil().mount_dvd()
                         self.fail('OSUtilError was not raised')
                     except OSUtilError as ose:
                         self.assertTrue(msg in ustr(ose))
-                        self.assertTrue(patch_run.call_count == 6)
+                        self.assertEqual(patch_run.call_count, 5)
```
pgombar, line 117: Why did the number of calls change if we didn't change the retry number (6)?

pgombar (approved):
Minor comments, else LGTM

tests/common/osutil/test_default.py
kevinclark19a: The range has always been `[1..6)`, so I'm not quite sure how this test was passing before. Any ideas?

kevinclark19a: Ah, ok I figured it out. Before, we were mocking `run_get_output`, because that's what `self.mount` used before. There is a `run_get_output` call  in `mount_dvd` that we decided not to bother changing (as it uses a constant string), which was the 6th call before. Since our mock is now for `run_command`, that `run_get_output` call obviously doesn't get counted anymore, bringing the total number of calls from 6 (5 `run_get_output` calls in `self.mount` + 1 `run_get_output` in `mount_dvd` itself) to 5 (just the five `run_command` calls in `self.mount`).

pgombar: Thanks for checking! Makes sense now.

pgombar (approved):

azurelinuxagent/common/utils/cryptutil.py
```diff
@@ -96,6 +96,9 @@ def decrypt_p7m(self, p7m_file, trans_prv_file, trans_cert_file, pem_file):
             stdout, stderr = second_proc.communicate()
 
             if first_proc.returncode != 0 or second_proc.returncode != 0:
+                stdout = ustr(stdout, encoding='utf-8', errors="backslashreplace") if stdout else ""
```
larohra, line 99: NIT: How about making this function public and using this instead - 

```python
def _encode_command_output(output):
    return ustr(output, encoding='utf-8', errors="backslashreplace")
```

azurelinuxagent/common/osutil/default.py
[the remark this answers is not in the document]
larohra: Nope that's a different log. I was comparing the previous implementation with the new one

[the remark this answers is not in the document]
larohra: Basically this parameter for `expected_errors=[1]` in the previous implementation -

```python
 return_code=shellutil.run("ifdown {0} && ifup {0}".format(ifname), expected_errors=[1] if attempt < retries else [])
``` 
The `shellutil.run` function would've logged it as an info if the exit_code was 1 but now it would be logged as an error.

Not sure how big of a difference it makes, just noticed it so figured I'd point it out. I'm personally fine as is too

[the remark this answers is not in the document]
narrieta: @larohra I posted a similar comment here, not sure why it doesn't show up.

In any case, I added the 'expected_errors' parameter some time back as a stopgap to avoid logging an error when the error code is 1. This seems to be relatively common and was logged as an error due to the way the run() function is defined, and this was confusing people looking at the log. From the caller, any error code executing the command is clearly a warning.

The new implementation drops the INFO (which was there just because run logs all failures in the command it execurtes), but keeps the warning. This should be fine.

azurelinuxagent/common/utils/cryptutil.py
narrieta: @kevinclark19a DCR ok?

narrieta (approved):

azurelinuxagent/common/osutil/default.py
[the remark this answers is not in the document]
larohra: Ahh ok ok, makes sense. Thanks for the clarification

PR #2033, opened 2020-10-09
Switch to run command fixes

kevinclark19a:
## Description

Fixes an erroneously logged error caused by the switch to run command. 

---

### PR information


pgombar:
Can we get a DCR run to confirm?

azurelinuxagent/common/utils/cryptutil.py
```diff
@@ -92,6 +92,8 @@ def decrypt_p7m(self, p7m_file, trans_prv_file, trans_cert_file, pem_file):
                 "-out", pem_file]
 
             first_proc = subprocess.Popen(first_cmd, stdout=subprocess.PIPE)
+            first_proc.wait()
```
larohra, line 95: Does this have potential to hang indefinitely? Because as we saw `wait()` is uninterruptable

kevinclark19a: We're already using `communicate` a couple of lines down. Is this is a concern, we'd need to fix it there too.

As for this wait, I can replace it with a single `poll` and it works as well; `communicate` already waits for `first_proc` to complete (as well as `second_proc`). This call (or, equally, a `poll` call) just populates the `Popen` object's `returncode` member variable.

narrieta: I would not be too concerned about the wait since those commands complete pretty quickly. The issues we've had is when we wait on processes that can take a long time to complete or never complete.

narrieta, line 95: The wait has the potential to introduce a deadlock (different issue to the above comment): if the first command produces too much output the pipe may block until it is consumed, which will not happen until the next command is started, which will not happen until the wait completes.

narrieta: See https://docs.python.org/2/library/subprocess.html#replacing-shell-pipeline

The original code was missing the call to close()

narrieta:
Replaced by #2034

PR #2034, opened 2020-10-09
Adding missing call to close() and initializing returncode in pipe

narrieta:


azurelinuxagent/common/utils/cryptutil.py
```diff
@@ -93,7 +93,9 @@ def decrypt_p7m(self, p7m_file, trans_prv_file, trans_cert_file, pem_file):
 
             first_proc = subprocess.Popen(first_cmd, stdout=subprocess.PIPE)
             second_proc = subprocess.Popen(second_cmd, stdin=first_proc.stdout, stdout=subprocess.PIPE)
+            first_proc.stdout.close()  # see https://docs.python.org/2/library/subprocess.html#replacing-shell-pipeline
             stdout, stderr = second_proc.communicate()
+            first_proc.poll()  # ensure that returncode is initialized
```
larohra, line 98: poll wont block and set None if still running. 

>Popen.returncode
The child return code, set by poll() and wait() (and indirectly by communicate()). A None value indicates that the process hasn’t terminated yet.

Shouldn't we set to `wait` instead? Or just remove the `first_proc.returncode` check below? Like the 2nd command would anyways fail if the first doesnt succeed right?

narrieta: after communicate() completes we know the 2 processes have completed, i don't think it makes a difference so i tossed a coin and went for poll

i was also wondering whether we need the check for the return code since it was not there before

the pipe returns the value of the last command

```
$ (date; exit 1) 
Fri 09 Oct 2020 03:56:36 PM PDT
$ echo $?
1
$ (date; exit 1) | wc
      1       7      32
$ echo $?
0
$ 

```

so i think it may be good to add that check

narrieta: on second though, probably wait() makes the code more explicit, so let me do that

larohra: >after communicate() completes we know the 2 processes have completed

Is this true for all cases? Like 2nd process's communicate ensures the first one also succeeds? Not sure if there's any documentation about it

narrieta: Like 2nd process's communicate ensures the first one also succeeds? 

No, it does not ensure that it succeeds, it ensures that it completes. If communicate returned before the first process completes then that process could still produce output which would be a problem in the implementation of communicate

larohra: Ahh makes sense. I'll do a random experiment to see how it handles this scenario though, should be fun haha.

larohra (approved):
LGTM

pgombar (approved):

PR #2043, opened 2020-10-15
Echo log to /dev/console during provisioning

narrieta:


tests/common/test_version.py
```diff
@@ -123,6 +124,23 @@ def test_get_lis_version_should_return_a_string(self):
         lis_version = get_lis_version()
         self.assertIsInstance(lis_version, str)
 
+    def test_get_daemon_version_should_return_the_version_that_was_previously_set(self):
+        set_daemon_version("1.2.3.4")
```
kevinclark19a, line 128: This function's argument is named `flexible_version`. Should this be a `FlexibleVersion` object? If not, maybe the function argument should be renamed.

narrieta: thanks for pointing it out. i intended to make it a FlexibleVersion for symmetry with the getter. will update the code.

narrieta: done

azurelinuxagent/agent.py
```diff
@@ -140,6 +150,21 @@ def run_exthandlers(self, debug=False):
...
+
+        #
+        # Agents < 2.2.53 used to echo the log to the console. Since the extension handler could have been started by
+        # one of those daemons, output a message indicating that output to the console will stop, otherwise users
+        # may think that the agent died if they noticed that output to the console stops abruptly.
+        #
+        # Feel free to remove this code if telemetry shows there are no more agents <= 2.2.53 in the field.
+        #
+        if conf.get_logs_console() and get_daemon_version() < FlexibleVersion("2.2.53"):
+            self.__add_console_appender(logger.LogLevel.INFO)
```
larohra, line 162: Should this be added under the try just to be on the safer side?

azurelinuxagent/common/version.py
```diff
@@ -25,6 +25,35 @@
 from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
 from azurelinuxagent.common.future import ustr, get_linux_distribution
 
+__DAEMON_VERSION_ENV_VARIABLE = '_AZURE_GUEST_AGENT_DAEMON_VERSION_'
```
larohra, line 28: Why not put this under this class - 

`class ExtCommandEnvVariable(object):`

If position of that class is a problem, you can also move it from `exthandlers.py` to some file under `common/` maybe?

larohra (approved):
Some minor comments but everything else LGTM.
(Since you touched the provisioning code, might be a good idea to get this reviewed by the provisioning folks just to be on the safer side)

tests/common/test_logger.py
```diff
@@ -565,3 +562,32 @@ def test_stdout_appender(self, mock_sys_stdout):
 
         # Validating only test-error gets logged and not others.
         self.assertEqual(1, mock_sys_stdout.call_count)
+
+    def test_console_output_enabled_should_return_true_when_there_are_no_console_appenders(self):
```
narrieta, line 566: typo; should be "when_there_are_console_appenders"

kevinclark19a (dismissed):

azurelinuxagent/common/version.py
narrieta: those are mean for extensions, this one is not

i used underscores to mark variables not meant to be used by extensions

azurelinuxagent/agent.py
narrieta: if the appender was not added we do not want do remove it (nothing bad would happen in the current implementation, but still)

kevinclark19a (approved):

azurelinuxagent/agent.py
larohra: I think there's been some confusion, I meant moving the adding part under the try too. Just having a `logger.info` under try adds little value.

```python
 try:
                self.__add_console_appender(logger.LogLevel.INFO)
                logger.info(u"The agent will now check for updates and then will process extensions. Output to /dev/console will be suspended during those operations.")
            finally:
                logger.disable_console_output()
```

larohra (approved):

pgombar (approved):

PR #2046, opened 2020-10-20
Unstick HGAP channel as default

pgombar:
## Description

Today, our logic is to use the direct channel as default unless it fails. If it fails, we try HGAP and if that succeeds, we stick to it indefinitely.

This change introduces alternating between default channels for each scenario using `send_request_using_appropriate_channel`, which is downloading extension manifests and packages, downloading artifacts profile, and downloading agent manifests. Agent upgrade scenario is unchanged.

Initially, direct channel is the default channel.
Scenario 1: Direct succeeds. HGAP is not called, direct remains default channel.
Scenario 2: Direct fails, HGAP succeeds -> HGAP is set to default and subsequent requests use HGAP first. Only if HGAP fails, we try to use direct. If direct fails, HGAP remains default as the last known good.
Scenario 3: Direct fails, HGAP fails -> direct remains default as the last known good. Each subsequent request goes through direct first.

This PR fixes the following things:
1) Correct alternating between channels.
2) Fixes handling failed responses (callers of `_fetch_response` never accounted for a failed response and propagated the value thinking it's successful).

---

### PR information


azurelinuxagent/common/protocol/wire.py
```diff
@@ -913,49 +913,51 @@ def _call_hostplugin_with_container_check(self, host_func):
...
+            logger.periodic_info(logger.EVERY_HOUR, "[PERIODIC] Request failed using the direct channel, "
+                                                    "trying host plugin next. Error: {0}".format(ustr(e)))
 
-            if ret:
-                return ret
-        else:
-            logger.periodic_info(logger.EVERY_HALF_DAY, "[PERIODIC] Using host plugin as default channel.")
+        if ret and HostPluginProtocol.is_default_channel():
+            # Set direct channel as default if it succeeded.
+            logger.info("Switching default channel to WireServer.")
```
narrieta, line 942: not to wire server, but to direct channel

azurelinuxagent/common/protocol/wire.py
```diff
@@ -913,49 +913,51 @@ def _call_hostplugin_with_container_check(self, host_func):
...
-            logger.info("Setting host plugin as default channel from now on. "
-                        "Restart the agent to reset the default channel.")
-            HostPluginProtocol.set_default_channel(True)
+    def send_request_using_appropriate_channel(self, direct_func, host_func):
+        primary_ret = None
+        try:
+            primary_ret = self.__send_request_using_host_channel(host_func) if HostPluginProtocol.is_default_channel() \
+                else self.__send_request_using_direct_channel(direct_func)
+        except Exception:
+            pass
```
narrieta, line 953: if you swallow the exception we will lose all the debugging info that could help to determine what kind of error occurred

azurelinuxagent/common/protocol/wire.py
```diff
@@ -913,49 +913,51 @@ def _call_hostplugin_with_container_check(self, host_func):
...
-                    logger.periodic_info(logger.EVERY_HOUR, "[PERIODIC] Request failed using the direct channel, "
-                                                            "switching to host plugin.")
-            except (ResourceGoneError, InvalidContainerError) as e: # pylint: disable=C0103
+            # Different direct channel functions report failure in different ways: by returning None, False,
+            # or raising ResourceGone or InvalidContainer exceptions.
+            if not ret:
                 logger.periodic_info(logger.EVERY_HOUR, "[PERIODIC] Request failed using the direct channel, "
-                                                        "switching to host plugin. Error: {0}".format(ustr(e)))
+                                                        "trying host plugin next.")
+        except (ResourceGoneError, InvalidContainerError) as e:  # pylint: disable=C0103
```
narrieta, line 936: those are errors returned by the host ga plugin. why are we checking for those on the direct channel?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -913,49 +913,51 @@ def _call_hostplugin_with_container_check(self, host_func):
...
+            logger.periodic_info(logger.EVERY_HOUR, "[PERIODIC] Request failed using the direct channel, "
```
narrieta, line 937: using periodic messages was not a very good decision in the first place, but since now we may be switching the default channel multiple times they are even less justifiable

azurelinuxagent/common/protocol/wire.py
```diff
@@ -913,49 +913,51 @@ def _call_hostplugin_with_container_check(self, host_func):
...
+            primary_ret = self.__send_request_using_host_channel(host_func) if HostPluginProtocol.is_default_channel() \
+                else self.__send_request_using_direct_channel(direct_func)
+        except Exception:
+            pass
 
-        return ret
+        if primary_ret:
+            return primary_ret
+
+        secondary_ret = self.__send_request_using_direct_channel(direct_func) if HostPluginProtocol.is_default_channel() \
```
narrieta, line 958: we need to improve the clarity of this code... it is really confusing that we are using the direct channel if the default channel is the host ga plugin

this condition is relying in non-obvious side-effects of the previous calls

azurelinuxagent/common/protocol/wire.py
```diff
@@ -913,49 +913,51 @@ def _call_hostplugin_with_container_check(self, host_func):
...
+            else self.__send_request_using_host_channel(host_func)
+        return secondary_ret
```
narrieta, line 960: it is also confusing that we use try/except when using the primary channel, but not when using the secondary channel. when both the primary and secondary channels could be either direct or the host ga plugin

azurelinuxagent/common/protocol/wire.py
```diff
@@ -913,49 +913,51 @@ def _call_hostplugin_with_container_check(self, host_func):
...
+        return ret
 
-        if not HostPluginProtocol.is_default_channel():
-            logger.info("Setting host plugin as default channel from now on. "
-                        "Restart the agent to reset the default channel.")
-            HostPluginProtocol.set_default_channel(True)
+    def send_request_using_appropriate_channel(self, direct_func, host_func):
+        primary_ret = None
+        try:
+            primary_ret = self.__send_request_using_host_channel(host_func) if HostPluginProtocol.is_default_channel() \
```
jacwil, line 950: Would a verbose message indicating the channel used be too verbose and fill up the logs? Or is there already a log entry that does something similar when requests are successful?

azurelinuxagent/common/protocol/wire.py
larohra: +1

azurelinuxagent/common/protocol/wire.py
jacwil: I was afraid it would be too verbose. HostPlugin will have logs if guest agent used it so this comment can be resolved.

azurelinuxagent/common/protocol/wire.py
pgombar: They're the same... Do you mean not using "WireServer"? My intention was to be more informative for customers or support (to make it clearer that WireServer IP needs to be accessible).

azurelinuxagent/common/protocol/wire.py
pgombar: This is existing behaviour. I can't remember right now why direct channel also has this check, I can look at the history... but I am hesitant to change it.

azurelinuxagent/common/protocol/wire.py
pgombar: Why is that confusing? We want to retry if primary fails (with secondary), but if secondary fails, we give up and report failure.

azurelinuxagent/common/protocol/wire.py
pgombar: > it is really confusing that we are using the direct channel if the default channel is the host ga plugin

hence the name "secondary" :smile:

narrieta: huh?

azurelinuxagent/common/protocol/wire.py
narrieta: please look at the history... we are trying to improve this code

azurelinuxagent/common/protocol/wire.py
narrieta: isn't direct to the internet, rather than to the wire server?

azurelinuxagent/common/protocol/wire.py
narrieta: it may be a consequence of the code above... once that is clarified this may become clearer too

azurelinuxagent/common/protocol/wire.py
```diff
@@ -913,49 +913,51 @@ def _call_hostplugin_with_container_check(self, host_func):
...
+                else self.__send_request_using_direct_channel(direct_func)
+        except Exception:
+            pass
 
-        return ret
+        if primary_ret:
```
narrieta, line 955: let's do explicit comparison with None

azurelinuxagent/common/protocol/wire.py
narrieta: what i was trying to ask is that when HostPluginProtocol.is_default_channel() == True is supposed to indicate that we should use the host channel and this is using the direct channel

azurelinuxagent/common/protocol/wire.py
narrieta: depending on your investigation we may decide to leave as is with documentation, change it in this PR or change it later

azurelinuxagent/common/protocol/wire.py
pgombar: Sorry, you have a point. "Direct" means different things in different scenarios; sometimes it's a storage endpoint, sometimes it's wireserver. I'll just use "direct" then.

azurelinuxagent/common/protocol/wire.py
larohra: From what I remember `InvalidContainerError` was raised by HostGA when the containerId is out of sync. I think `ResourceGoneError` was also due to something with HostGA but its definitely better to confirm (maybe @jacwil  can also shed some light here)

azurelinuxagent/common/protocol/wire.py
```diff
@@ -913,49 +913,51 @@ def _call_hostplugin_with_container_check(self, host_func):
...
-            except (ResourceGoneError, InvalidContainerError) as e: # pylint: disable=C0103
+            # Different direct channel functions report failure in different ways: by returning None, False,
+            # or raising ResourceGone or InvalidContainer exceptions.
+            if not ret:
                 logger.periodic_info(logger.EVERY_HOUR, "[PERIODIC] Request failed using the direct channel, "
-                                                        "switching to host plugin. Error: {0}".format(ustr(e)))
+                                                        "trying host plugin next.")
+        except (ResourceGoneError, InvalidContainerError) as e:  # pylint: disable=C0103
+            logger.periodic_info(logger.EVERY_HOUR, "[PERIODIC] Request failed using the direct channel, "
+                                                    "trying host plugin next. Error: {0}".format(ustr(e)))
```
larohra, line 938: NIT: log message `trying host plugin next` would need to be changed based on whether it was first or second.

azurelinuxagent/common/protocol/wire.py
larohra: As far as I'm aware, direct is mostly always storage and not wireserver. The only things that we use wireserver directly are for - fetching goalstate and sending telemetry (and ofcourse HostGA but that's a separate port altogether)

azurelinuxagent/common/protocol/wire.py
larohra: +1

tests/protocol/test_wire.py
```diff
@@ -851,99 +851,139 @@ def http_put_handler(url, *_, **__):
...
-    def test_send_request_using_appropriate_channel_should_not_invoke_host_channel_when_direct_channel_succeeds(self):
-        with mock_wire_protocol(mockwiredata.DATA_FILE) as protocol:
-            protocol.client.get_host_plugin().set_default_channel(False)
-
-            def direct_func(*args): # pylint: disable=unused-argument
-                direct_func.counter += 1
-                return 42
+    @staticmethod
+    def _set_and_fail_helper_channel_functions(fail_direct=False, fail_host=False):
+        def direct_func(*args):  # pylint: disable=unused-argument
```
larohra, line 856: NIT: Unused `*args`

tests/protocol/test_wire.py
```diff
@@ -851,99 +851,139 @@ def http_put_handler(url, *_, **__):
...
+            direct_func.counter += 1
+            if direct_func.fail:
+                return None
+            return "direct"
 
-            def host_func(*args): # pylint: disable=useless-return,unused-argument
-                host_func.counter += 1
+        def host_func(*args):  # pylint: disable=unused-argument
```
larohra, line 862: NIT: Unused `*args`

tests/protocol/test_wire.py
```diff
@@ -851,99 +851,139 @@ def http_put_handler(url, *_, **__):
...
+            ret = protocol.client.send_request_using_appropriate_channel(direct_func, host_func)
+            self.assertEqual(None, ret)
+            self.assertEqual(1, direct_func.counter)
+            self.assertEqual(1, host_func.counter)
+            self.assertFalse(host.is_default_channel())
 
-            def host_func(*args): # pylint: disable=unused-argument
-                host_func.counter += 1
-                return 42
+            # If both channels keep failing, assert we are not changing the default, but keep trying both.
```
larohra, line 919: NIT: Why not combine the test cases from line 912-917 and 920-925?

tests/protocol/test_wire.py
```diff
@@ -851,99 +851,139 @@ def http_put_handler(url, *_, **__):
...
+            self.assertEqual(None, ret)
             self.assertEqual(1, direct_func.counter)
             self.assertEqual(1, host_func.counter)
-            self.assertEqual(True, host.is_default_channel())
+            self.assertTrue(host.is_default_channel())
 
-    def test_send_request_using_appropriate_channel_should_retry_the_host_channel_after_reloading_goal_state(self):
-        with mock_wire_protocol(mockwiredata.DATA_FILE) as protocol:
-            protocol.client.get_host_plugin().set_default_channel(False)
+            # If both channels keep failing, assert we are not changing the default, but keep trying both.
```
larohra, line 938: NIT: Same, why not combine the cases from line 933-936 and 940-944?

azurelinuxagent/common/protocol/wire.py
narrieta: @larohra yes, those are raised by HostGA plugin. It is odd that we using them for the direct channel.

Although those checks were not introduced by this PR, we should look at the code history to double-check we did not copy-paste those checks by mistake when refactoring this code previously. If they have been there from the beginning we should mark this code or open a task/issue to correct this later (hopefully soon)

azurelinuxagent/common/protocol/wire.py
pgombar: As discussed offline, I'll rewrite this method to be clearer.

azurelinuxagent/common/protocol/wire.py
pgombar: As discussed offline, I'll rewrite this method to be clearer.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -857,100 +857,77 @@ def check_wire_protocol_version(self):
...
-                                                    new_role_config_name, type(e).__name__)
-                add_periodic(delta=logger.EVERY_SIX_HOURS,
-                             name=AGENT_NAME,
-                             version=CURRENT_VERSION,
-                             op=WALAEventOperation.HostPlugin,
-                             is_success=False,
-                             message=msg,
-                             log_event=True)
-                raise
+                      "Exception type: {4}.".format(old_container_id, new_container_id, old_role_config_name, new_role_config_name, type(error).__name__)
```
narrieta, line 891: why do we log the exception type and not the message?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -628,20 +623,22 @@ def fetch_manifest(self, version_uris):
...
-            except Exception as e:  # pylint: disable=C0103
-                logger.warn("Exception when fetching manifest. Error: {0}".format(ustr(e)))
+            except Exception as error:
+                logger.warn("Failed to fetch manifest from {0}. Error: {1}", version.uri, ustr(error))
 
         raise ExtensionDownloadError("Failed to fetch manifest from all sources")
 
     def stream(self, uri, destination, headers=None, use_proxy=None):
-        success = False
+        success = None
```
narrieta, line 641: what does None mean in this context? Is the value returned by this function tri-state? (None/False/True)

azurelinuxagent/common/protocol/wire.py
```diff
@@ -695,10 +692,13 @@ def _fetch_response(self, uri, headers=None, use_proxy=None):
                 if host_plugin is not None:
                     host_plugin.report_fetch_health(uri, source='WireClient')
 
-        except (HttpError, ProtocolError, IOError) as e:  # pylint: disable=C0103
-            logger.verbose("Fetch failed from [{0}]: {1}", uri, e)
-            if isinstance(e, ResourceGoneError) or isinstance(e, InvalidContainerError):  # pylint: disable=R1701
+        except (HttpError, ProtocolError, IOError, ProtocolError) as error:
```
narrieta, line 695: ProtocolError is repeated

azurelinuxagent/common/protocol/wire.py
```diff
@@ -161,12 +161,7 @@ def download_ext_handler_pkg(self, uri, destination, headers=None, use_proxy=Tru
         # in the lambda.
         host_func = lambda: self._download_ext_handler_pkg_through_host(uri, destination)
 
-        try:
-            success = self.client.send_request_using_appropriate_channel(direct_func, host_func)
-        except Exception:
-            success = False
-
-        return success
+        return self.client.send_request_using_appropriate_channel(direct_func, host_func)
```
narrieta, line 164: Previously download_ext_handler_pkg was returning False on error... seems like now it can return None

azurelinuxagent/common/protocol/wire.py
narrieta: also, previously stream was returning False, now it returns None.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -857,100 +857,77 @@ def check_wire_protocol_version(self):
...
-                             is_success=False,
-                             message=msg,
-                             log_event=True)
-                raise
+                      "Exception type: {4}.".format(old_container_id, new_container_id, old_role_config_name, new_role_config_name, type(error).__name__)
+                add_event(AGENT_NAME, op=WALAEventOperation.HostPlugin, version=CURRENT_VERSION, is_success=False, message=msg, log_event=True)
+                return None
+        except Exception as error:
+            logger.info("Request failed using the host plugin channel. Error: {0}", ustr(error))
+            return None
```
narrieta, line 896: we also changing the way we are reporting errors here... previously we were raising and now we are returning None.

this changes the interface of upload_logs

azurelinuxagent/common/protocol/wire.py
pgombar: This change intentional, previously it was tri-state, now it's None/True.
Now all functions getting called in `send_request_using_appropriate_channel` return None on failure, not None or False.

azurelinuxagent/common/protocol/wire.py
pgombar: This change is intentional. Now all functions getting called in `send_request_using_appropriate_channel` return None on failure.

azurelinuxagent/common/protocol/wire.py
pgombar: Thanks, will remove.

azurelinuxagent/common/protocol/wire.py
pgombar: Upload logs doesn't use this function. It uses a method from `hostplugin.py`.

azurelinuxagent/common/protocol/wire.py
pgombar: This is old behaviour, not sure. Let me check

azurelinuxagent/common/protocol/wire.py
pgombar: These errors are specific to the host channel. We don't need to check for them on the direct channel, I updated the code.

azurelinuxagent/common/protocol/wire.py
pgombar: Discussed offline; reverted this change to keep interface consistent with current behaviour.

azurelinuxagent/common/protocol/wire.py
pgombar: Discussed offline; reverted this change to keep interface consistent with current behaviour.

azurelinuxagent/common/protocol/wire.py
pgombar: It actually uses this function, you were right. The code has been updated, the interface of upload_logs remains the same as before.

azurelinuxagent/common/protocol/wire.py
pgombar: Leaving as-is for now. We never logged the full exception, possibly due to privacy? Not sure, keeping this the same to reduce the scope of changes in this PR.

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -56,8 +56,8 @@
 MAXIMUM_PAGEBLOB_PAGE_SIZE = 4 * 1024 * 1024  # Max page size: 4MB
 
 
-class HostPluginProtocol(object):  # pylint: disable=R0902
-    _is_default_channel = False
+class HostPluginProtocol(object): # pylint: disable=R0902
```
kevinclark19a, line 59: This is removing a second space from before a comment.

Is this what we want? I know @narrieta had a giant PR a little while ago about _adding_ these in.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -908,49 +918,59 @@ def _call_hostplugin_with_container_check(self, host_func):
...
+            return ret
 
-        if not HostPluginProtocol.is_default_channel():
-            logger.info("Setting host plugin as default channel from now on. "
-                        "Restart the agent to reset the default channel.")
-            HostPluginProtocol.set_default_channel(True)
+        ret = secondary_channel()
+        if ret is not None:
+            HostPluginProtocol.is_default_channel = not HostPluginProtocol.is_default_channel
+            logger.info("Default channel changed to {0}.", "host" if HostPluginProtocol.is_default_channel else "direct")
```
larohra, line 973: How verbose can this be? Should we also send a Kusto event for this? We would have more data to check how frequently we switch channels that way

azurelinuxagent/common/protocol/wire.py
```diff
@@ -908,49 +918,59 @@ def _call_hostplugin_with_container_check(self, host_func):
                              log_event=True)
                 raise
 
+    def __send_request_using_host_channel(self, host_func):
+        """
+        Calls the host_func on host channel with retries for stale goal state and handles any exceptions, consistent with the caller for direct channel.
+        """
```
larohra, line 924: Can you also add the possible return values here too? Just to make it a bit more clearer without having to go in the internal functions

azurelinuxagent/common/protocol/wire.py
```diff
@@ -908,49 +918,59 @@ def _call_hostplugin_with_container_check(self, host_func):
...
 
-    def send_request_using_appropriate_channel(self, direct_func, host_func):
-        # A wrapper method for all function calls that send HTTP requests. The purpose of the method is to
-        # define which channel to use, direct or through the host plugin. For the host plugin channel,
-        # also implement a retry mechanism.
+    @staticmethod
+    def __send_request_using_direct_channel(direct_func):
+        """
+        Calls the direct_func on direct channel and handles any exceptions, consistent with the caller for host channel.
+        """
```
larohra, line 937: Same here, update the docstring to specify the possible return values

azurelinuxagent/common/protocol/wire.py
```diff
@@ -908,49 +918,59 @@ def _call_hostplugin_with_container_check(self, host_func):
...
 
-        # NOTE: direct_func and host_func are passed as lambdas. Be careful about capturing goal state data in them as
-        # they will not be refreshed even if a goal state refresh is called before retrying the host_func.
+    def send_request_using_appropriate_channel(self, direct_func, host_func):
+        """
+        Determines which communication channel to use. By default, the primary channel is direct, host channel is secondary.
+        We call the primary channel first and return on success. If primary fails, we try secondary. If secondary fails,
+        we return and *don't* switch the default channel. If secondary succeeds, we change the default channel.
+        This method doesn't raise since the calls to direct_func and host_func are already wrapped and handle any exceptions.
+        Possible return values are manifest, artifacts profile, boolean or None.
```
larohra, line 956: Correct me if I'm wrong, but this function will never return False right? 
If that understanding is correct, could you update the docstring to be more specific - 

>Possible return values are manifest, artifacts profile, **True** or None.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -162,7 +162,7 @@ def download_ext_handler_pkg(self, uri, destination, headers=None, use_proxy=Tru
         host_func = lambda: self._download_ext_handler_pkg_through_host(uri, destination)
 
         try:
-            success = self.client.send_request_using_appropriate_channel(direct_func, host_func)
+            success = self.client.send_request_using_appropriate_channel(direct_func, host_func) is not None
         except Exception:
             success = False
```
larohra, line 167: Additionally looks like this wont every throw now, we can remove this except check maybe as its redundant?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -908,49 +918,59 @@ def _call_hostplugin_with_container_check(self, host_func):
...
-                    logger.periodic_info(logger.EVERY_HOUR, "[PERIODIC] Request failed using the direct channel, "
-                                                            "switching to host plugin.")
-            except (ResourceGoneError, InvalidContainerError) as e:  # pylint: disable=C0103
-                logger.periodic_info(logger.EVERY_HOUR, "[PERIODIC] Request failed using the direct channel, "
-                                                        "switching to host plugin. Error: {0}".format(ustr(e)))
-
-            if ret:
-                return ret
+        if HostPluginProtocol.is_default_channel:
+            primary_channel, secondary_channel = host_channel, direct_channel
```
larohra, line 962: This makes it so clear to understand the channel! Thanks!

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1194,6 +1214,8 @@ def get_artifacts_profile(self):
 
             try:
                 profile = self.send_request_using_appropriate_channel(direct_func, host_func)
+                if profile is None:
+                    raise Exception("Failed to fetch artifacts profile from blob {0}".format(blob))
```
larohra, line 1218: This is a redundant raise, this will be caught in the next line and then eventually None would be returned. Why not just handle it like that to begin with (with proper logging ofcourse)

tests/protocol/test_wire.py
```diff
@@ -493,7 +493,7 @@ def http_get_handler(url, *_, **__):
             self.assertEqual(len(urls), 1, "Unexpected number of HTTP requests: [{0}]".format(urls))
             self.assertEqual(urls[0], extension_url, "The extension should have been downloaded over the direct channel")
             self.assertTrue(os.path.exists(target_file), 'The extension package was not downloaded')
-            self.assertEqual(HostPluginProtocol.is_default_channel(), False, "The host channel should not have been set as the default")
+            self.assertEqual(HostPluginProtocol.is_default_channel, False, "The host channel should not have been set as the default")
```
larohra, line 496: NIT: self.AssertFalse instead of self.assertEqual(False, val)

tests/protocol/test_wire.py
```diff
@@ -517,13 +517,13 @@ def http_get_handler(url, *_, **kwargs):
             self.assertEqual(urls[0], extension_url, "The first attempt should have been over the direct channel")
             self.assertTrue(self.is_host_plugin_extension_artifact_request(urls[1]), "The retry attempt should have been over the host channel")
             self.assertTrue(os.path.exists(target_file), 'The extension package was not downloaded')
-            self.assertEqual(HostPluginProtocol.is_default_channel(), True, "The host channel should have been set as the default")
+            self.assertEqual(HostPluginProtocol.is_default_channel, True, "The host channel should have been set as the default")
```
larohra, line 520: NIT: self.assertTrue

tests/protocol/test_wire.py
```diff
@@ -635,9 +633,9 @@ def http_get_handler(url, *_, **kwargs):
                 self.assertEqual(len(urls), 2, "Unexpected number of HTTP requests: [{0}]".format(urls))
                 self.assertEqual(urls[0], manifest_url, "The first attempt should have been over the direct channel")
                 self.assertTrue(self.is_host_plugin_extension_artifact_request(urls[1]), "The retry should have been over the host channel")
-                self.assertEqual(HostPluginProtocol.is_default_channel(), True, "The host should have been set as the default channel")
+                self.assertEqual(HostPluginProtocol.is_default_channel, True, "The host should have been set as the default channel")
```
larohra, line 636: NIT: self.assertTrue

tests/protocol/test_wire.py
```diff
@@ -674,9 +672,9 @@ def http_get_handler(url, *_, **kwargs):
                 self.assertTrue(self.is_host_plugin_extension_artifact_request(urls[1]), "The second attempt should have been over the host channel")
                 self.assertTrue(self.is_goal_state_request(urls[2]), "The host channel should have been refreshed the goal state")
                 self.assertTrue(self.is_host_plugin_extension_artifact_request(urls[3]), "The third attempt should have been over the host channel")
-                self.assertEqual(HostPluginProtocol.is_default_channel(), True, "The host should have been set as the default channel")
+                self.assertEqual(HostPluginProtocol.is_default_channel, True, "The host should have been set as the default channel")
```
larohra, line 675: NIT: self.assertTrue

tests/protocol/test_wire.py
```diff
@@ -722,14 +720,14 @@ def http_get_handler(url, *_, **__):  # pylint: disable=useless-return
...
-            HostPluginProtocol.set_default_channel(False)
+            HostPluginProtocol.is_default_channel = False
 
             return_value = protocol.client.get_artifacts_profile()
 
             self.assertIsInstance(return_value, InVMArtifactsProfile, 'The request did not return a valid artifacts profile: {0}'.format(return_value))
             urls = protocol.get_tracked_urls()
             self.assertEqual(len(urls), 1, "Unexpected HTTP requests: [{0}]".format(urls))
-            self.assertEqual(HostPluginProtocol.is_default_channel(), False)
+            self.assertEqual(HostPluginProtocol.is_default_channel, False)
```
larohra, line 730: NIT: self.assertFalse

tests/protocol/test_wire.py
```diff
@@ -754,9 +752,9 @@ def http_get_handler(url, *_, **kwargs):
                 self.assertEqual(len(urls), 2, "Invalid number of requests: [{0}]".format(urls))
                 self.assertTrue(self.is_in_vm_artifacts_profile_request(urls[0]), "The first request should have been over the direct channel")
                 self.assertTrue(self.is_host_plugin_extension_artifact_request(urls[1]), "The second request should have been over the host channel")
-                self.assertEqual(HostPluginProtocol.is_default_channel(), True, "The default channel should have changed to the host")
+                self.assertEqual(HostPluginProtocol.is_default_channel, True, "The default channel should have changed to the host")
```
larohra, line 755: NIT: self.assertTrue

tests/protocol/test_wire.py
```diff
@@ -792,9 +790,9 @@ def http_get_handler(url, *_, **kwargs):
                 self.assertTrue(self.is_host_plugin_extension_artifact_request(urls[1]), "The second request should have been over the host channel")
                 self.assertTrue(self.is_goal_state_request(urls[2]), "The goal state should have been refreshed before retrying the host channel")
                 self.assertTrue(self.is_host_plugin_extension_artifact_request(urls[3]), "The retry request should have been over the host channel")
-                self.assertEqual(HostPluginProtocol.is_default_channel(), True, "The default channel should have changed to the host")
+                self.assertEqual(HostPluginProtocol.is_default_channel, True, "The default channel should have changed to the host")
```
larohra, line 793: NIT: self.assertTrue

tests/protocol/test_wire.py
```diff
@@ -823,7 +821,7 @@ def http_get_handler(url, *_, **kwargs):
             self.assertTrue(self.is_host_plugin_extension_artifact_request(urls[1]), "The second request should have been over the host channel")
             self.assertTrue(self.is_goal_state_request(urls[2]), "The goal state should have been refreshed before retrying the host channel")
             self.assertTrue(self.is_host_plugin_extension_artifact_request(urls[3]), "The retry request should have been over the host channel")
-            self.assertEqual(HostPluginProtocol.is_default_channel(), False, "The default channel should not have changed")
+            self.assertEqual(HostPluginProtocol.is_default_channel, False, "The default channel should not have changed")
```
larohra, line 824: NIT: self.assertFalse

tests/protocol/test_wire.py
```diff
@@ -859,99 +857,119 @@ def http_put_handler(url, *_, **__):
...
+        with mock_wire_protocol(mockwiredata.DATA_FILE) as protocol:
+            # Scenario #1: Direct channel default
+            HostPluginProtocol.is_default_channel = False
 
-            def host_func(*args):  # pylint: disable=unused-argument
-                host_func.counter += 1
-                return 43
+            direct_func, host_func = self._set_and_fail_helper_channel_functions()
+            # Assert we're only calling the primary channel (direct) and that it succeeds.
+            for n in range(5):  # pylint: disable=invalid-name
```
larohra, line 889: NIT: Pylint error

tests/protocol/test_wire.py
```diff
@@ -859,99 +857,119 @@ def http_put_handler(url, *_, **__):
...
+            HostPluginProtocol.is_default_channel = True
+            direct_func, host_func = self._set_and_fail_helper_channel_functions()
 
-            # Assert we've only called the host channel function since it's the default channel
-            ret = protocol.client.send_request_using_appropriate_channel(direct_func, host_func)
-            self.assertEqual(43, ret)
-            self.assertEqual(0, direct_func.counter)
-            self.assertEqual(1, host_func.counter)
+            # Assert we're only calling the primary channel (host) and that it succeeds.
+            for n in range(5):  # pylint: disable=invalid-name
```
larohra, line 901: NIT: Pylint error

larohra (commented):
Some minor comments regarding renaming + code-beautification else LGTM

azurelinuxagent/common/protocol/hostplugin.py
pgombar: I think this is the result of GitHub's automatic merge. Thanks, I'll make sure to update this and any other instance.

azurelinuxagent/common/protocol/wire.py
pgombar: I intentionally put that raise so I can re-use the code that logs the warning and returns None, but I realized the error message in that case would look weird: "Exception retrieving artifacts profile: Failed to fetch artifacts profile from blob [blob]". I decided to go with your suggestion, so this is updated now.

azurelinuxagent/common/protocol/wire.py
pgombar: I'm so glad!! It took several iterations to get to this phase :smile:

azurelinuxagent/common/protocol/wire.py
pgombar: So this only happens if the primary fails **and** secondary succeeds. It will happen again only if that same channel we just set as default fails again **and** the other now succeeds.

I don't think this happens often, usually just one channel is down so the other takes over, in which case we wouldn't be hitting this successively. The only scenario where this might be verbose is if both channels are flaky **at the same time**, which I believe is a very small chance.

I like the idea of logging it to Kusto, if at least just for this release to give us a sense of how often it happens. Let me create a new operation for this.

azurelinuxagent/common/protocol/wire.py
pgombar: It can be False, `direct_func` or `host_func` can be calling `azurelinuxagent.common.protocol.wire.WireClient.stream` which returns False.

azurelinuxagent/common/protocol/wire.py
pgombar: Sure, that's a good point.

azurelinuxagent/common/protocol/wire.py
pgombar: Done.

azurelinuxagent/common/protocol/wire.py
pgombar: I had initially changed this and the caller of this method (since its try/catch is also redundant), but Norberto and I agreedto keep this as-is now so we contain the changes to the level of `send_request_using_appropriate_channel` and methods below. This way, the review is relatively simple without having to go up the call stack.

There's still a lot of improvements to be done with these internal functions (unifying their interfaces for one), so your comment can be addressed in a future round of changes.

azurelinuxagent/common/protocol/wire.py
larohra: True, but for both functions (`__send_request_using_direct_channel` and `__send_request_using_host_channel`) you have this check which ensures it can never be False right? (Not sure if I'm missing something here)  

```python
            if ret is None or not ret:
                logger.periodic_info(logger.EVERY_HOUR, "[PERIODIC] Request failed using the direct channel.")
                return None

            if ret is None or not ret:
                raise Exception("Request failed using the host channel.")
```

azurelinuxagent/common/protocol/wire.py
larohra: Umm I get the point of making less changes but in this particular instance the code path (line 166) will never be hit (because the underlying function can never raise after these changes). Given that, wouldn't it make sense to just remove that? Why bother with having more dead code to eventually deal with later when we can just do it right now :) 

>This way, the review is relatively simple without having to go up the call stack.

True this, I wish I had read this before but since I already went deep I figured I'd add the comment haha. 
I'll leave it upto you in that case :)

azurelinuxagent/common/protocol/wire.py
```diff
@@ -934,6 +937,9 @@ def __send_request_using_host_channel(self, host_func):
     def __send_request_using_direct_channel(direct_func):
         """
         Calls the direct_func on direct channel and handles any exceptions, consistent with the caller for host channel.
+        At the time of writing, direct_func internally calls either:
+        1) WireClient.stream which returns a boolean, or
```
larohra, line 941: Based on line 948, this can never be False

```
if ret is None or not ret:
                logger.periodic_info(logger.EVERY_HOUR, "[PERIODIC] Request failed using the direct channel.")	                logger.periodic_info(logger.EVERY_HOUR, "[PERIODIC] Request failed using the direct channel.")
                return None	                return None
```

Would make sense to rename it to something like `Either returns True or None`

azurelinuxagent/common/protocol/wire.py
```diff
@@ -921,6 +921,9 @@ def _call_hostplugin_with_container_check(self, host_func):
     def __send_request_using_host_channel(self, host_func):
         """
         Calls the host_func on host channel with retries for stale goal state and handles any exceptions, consistent with the caller for direct channel.
+        At the time of writing, host_func internally calls either:
+        1) WireClient.stream which returns a boolean, or
```
larohra, line 925: Similar comment for this

azurelinuxagent/common/protocol/wire.py
```diff
@@ -970,8 +976,9 @@ def send_request_using_appropriate_channel(self, direct_func, host_func):
         ret = secondary_channel()
         if ret is not None:
             HostPluginProtocol.is_default_channel = not HostPluginProtocol.is_default_channel
-            logger.info("Default channel changed to {0}.", "host" if HostPluginProtocol.is_default_channel else "direct")
-
+            message = "Default channel changed to {0}.".format("host" if HostPluginProtocol.is_default_channel else "direct")
```
larohra, line 979: Being very nitpicky here, but would it make more sense to say "HostGA" instead of "host"?
(either works for me!)

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1215,9 +1222,10 @@ def get_artifacts_profile(self):
             try:
                 profile = self.send_request_using_appropriate_channel(direct_func, host_func)
                 if profile is None:
-                    raise Exception("Failed to fetch artifacts profile from blob {0}".format(blob))
+                    logger.warn("Failed to fetch artifacts profile from blob {0}", blob)
+                    return None
             except Exception as e:  # pylint: disable=C0103
```
larohra, line 1227: NIT: Pylint error

azurelinuxagent/common/protocol/wire.py
pgombar: The docstring refers to return values of the internal function (`direct_func`) not `__send_request_using_direct_channel`. I'll clarify and update the docstring.

azurelinuxagent/common/protocol/wire.py
pgombar: Sure, HostGA is more specific.

azurelinuxagent/common/protocol/wire.py
larohra: Since this is a wrapper over those calls, shouldn't this docstring specify what this wrapper is returning rather than what the internal functions can return?

pgombar: I added what the wrapper is returning now. I wanted to include what the wrapped functions are returning too since it's not obvious (we just pass them as lambdas, there's no quick look-up). The docstring should contain everything now!

larohra (approved):

azurelinuxagent/common/protocol/wire.py
```diff
@@ -858,47 +859,56 @@ def check_wire_protocol_version(self):
...
-                    add_periodic(delta=logger.EVERY_SIX_HOURS,
-                                 name=AGENT_NAME,
-                                 version=CURRENT_VERSION,
-                                 op=WALAEventOperation.HostPlugin,
-                                 is_success=True,
-                                 message=msg,
-                                 log_event=True)
 
-            except (ResourceGoneError, InvalidContainerError) as e:  # pylint: disable=C0103
+                if ret is None or not ret:
```
kevinclark19a, line 892: The `ret is None` clause is redundant, no? `not ret` already covers the `ret == None` case.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -908,50 +918,69 @@ def _call_hostplugin_with_container_check(self, host_func):
...
+        2) WireClient.fetch which returns None or a HTTP response.
+        This method returns either None (failure case where direct_func returned None or False), True or an HTTP response.
+        """
+        ret = None
+        try:
+            ret = direct_func()
 
-        # By default, the direct channel is the default channel. If that is the case, try getting a response
-        # through that channel. On failure, fall back to the host plugin channel.
+            if ret is None or not ret:
```
kevinclark19a, line 950: Same comment as above.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -858,47 +859,56 @@ def check_wire_protocol_version(self):
...
-        ret = None
+        """
+        Calls host_func on host channel and accounts for stale resource (ResourceGoneError or InvalidContainerError).
+        If stale, it refreshes the goal state and retries host_func.
+        This method can throw, so the callers need to handle that.
+        """
         try:
             ret = host_func()
-        except (ResourceGoneError, InvalidContainerError) as e:  # pylint: disable=C0103
+            if ret is None or not ret:
```
kevinclark19a, line 869: Oh, didn't notice this one until I had already commented the comment below.

Same thing, though.

azurelinuxagent/common/protocol/wire.py
pgombar: This is intentionally explicit to make the code more readable and signal clearly that `ret` can be both `None` or False.

azurelinuxagent/common/protocol/wire.py
pgombar: Same reply.

azurelinuxagent/common/protocol/wire.py
pgombar: Same.

azurelinuxagent/common/protocol/wire.py
kevinclark19a: I don't think this is a blocker for the PR, but it might be better to have, e.g.,
```python
if ret in (None, False):
    ...
```

Not necessary, but a bit nicer way to make the code explicit.

This also has the benefit (however small) of ignoring `ret == [], {}` for example.

kevinclark19a (dismissed):
LGTM. I don't think it's necessary to do the changes I suggested, but I'll reapprove if you want to make the changes.

azurelinuxagent/common/protocol/wire.py
pgombar: This does look cleaner, let me make that change. Thanks!

larohra (approved):

kevinclark19a (approved):

PR #2051, opened 2020-10-23
Fix timestamp for goal state archive

pgombar:
## Description

Today, we archive the last processed goal state when a new goal state comes in. However, we erroneously use the *new* goal state timestamp for that archive instead of when we received the *last* processed goal state. This PR fixed the archive timestamp.

---

### PR information

### DIFF
--- a/azurelinuxagent/common/protocol/goal_state.py
+++ b/azurelinuxagent/common/protocol/goal_state.py
@@ -20,6 +20,7 @@ import json
 import os
 import re
 import time
+from datetime import datetime
 
 import azurelinuxagent.common.conf as conf
 import azurelinuxagent.common.logger as logger
@@ -141,6 +142,7 @@ class GoalState(object):  # pylint: disable=R0902
             raise ProtocolError(msg="Error fetching goal state", inner=exception)
         finally:
             logger.info('Fetch goal state completed')
+            self.fetch_timestamp = datetime.utcnow()
 
     @staticmethod
     def fetch_goal_state(wire_client):
--- a/azurelinuxagent/common/protocol/goal_state.py
+++ b/azurelinuxagent/common/protocol/goal_state.py
@@ -20,19 +20,18 @@ import json
 import os
 import re
 import time
-from datetime import datetime
 
 import azurelinuxagent.common.conf as conf
 import azurelinuxagent.common.logger as logger
 from azurelinuxagent.common.AgentGlobals import AgentGlobals
 from azurelinuxagent.common.datacontract import set_properties
 from azurelinuxagent.common.event import add_event, WALAEventOperation
+from azurelinuxagent.common.exception import IncompleteGoalStateError
 from azurelinuxagent.common.exception import ProtocolError
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.protocol.restapi import Cert, CertList, Extension, ExtHandler, ExtHandlerList, \
     ExtHandlerVersionUri, RemoteAccessUser, RemoteAccessUsersList, \
     VMAgentManifest, VMAgentManifestList, VMAgentManifestUri
-from azurelinuxagent.common.exception import IncompleteGoalStateError
 from azurelinuxagent.common.utils import fileutil
 from azurelinuxagent.common.utils.cryptutil import CryptUtil
 from azurelinuxagent.common.utils.textutil import parse_doc, findall, find, findtext, getattrib, gettext
@@ -142,7 +141,6 @@ class GoalState(object):  # pylint: disable=R0902
             raise ProtocolError(msg="Error fetching goal state", inner=exception)
         finally:
             logger.info('Fetch goal state completed')
-            self.fetch_timestamp = datetime.utcnow()
 
     @staticmethod
     def fetch_goal_state(wire_client):
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -16,14 +16,13 @@
 #
 # Requires Python 2.6+ and Openssl 1.0+
 
-import datetime
 import json
 import os
 import random
 import time
 import traceback
 import xml.sax.saxutils as saxutils
-from datetime import datetime # pylint: disable=ungrouped-imports
+from datetime import datetime  # pylint: disable=ungrouped-imports
 
 import azurelinuxagent.common.conf as conf
 import azurelinuxagent.common.logger as logger
@@ -542,6 +541,8 @@ class WireClient(object): # pylint: disable=R0904
         self._endpoint = endpoint
         self._goal_state = None
         self._last_try_update_goal_state_failed = False
+        self._previous_goal_state_timestamp = None
+        self._current_goal_state_timestamp = None
         self._host_plugin = None
         self.status_blob = StatusBlob(self)
         self.goal_state_flusher = StateFlusher(conf.get_lib_dir())
@@ -758,11 +759,18 @@ class WireClient(object): # pylint: disable=R0904
             self._host_plugin.update_role_config_name(role_config_name)
 
     def _save_goal_state(self):
-        try:
-            self.goal_state_flusher.flush(datetime.utcnow())
+        self._current_goal_state_timestamp = datetime.utcnow()
 
+        # If this is the first goal state, the timestamp is current
+        if not self._previous_goal_state_timestamp:
+            self._previous_goal_state_timestamp = self._current_goal_state_timestamp
+
+        try:
+            self.goal_state_flusher.flush(self._previous_goal_state_timestamp)
         except Exception as e: # pylint: disable=C0103
             logger.warn("Failed to save the previous goal state to the history folder: {0}", ustr(e))
+        finally:
+            self._previous_goal_state_timestamp = self._current_goal_state_timestamp
 
         try:
             local_file = os.path.join(conf.get_lib_dir(), INCARNATION_FILE_NAME)
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -22,7 +22,6 @@ import random
 import time
 import traceback
 import xml.sax.saxutils as saxutils
-from datetime import datetime  # pylint: disable=ungrouped-imports
 
 import azurelinuxagent.common.conf as conf
 import azurelinuxagent.common.logger as logger
@@ -720,9 +719,8 @@ class WireClient(object): # pylint: disable=R0904
                 new_goal_state = GoalState.fetch_full_goal_state_if_incarnation_different_than(self, self._goal_state.incarnation)
 
             if new_goal_state is not None:
-                previous_goal_state_fetch_timestamp = self._goal_state.fetch_timestamp if self._goal_state else datetime.utcnow()
                 self._goal_state = new_goal_state
-                self._save_goal_state(previous_goal_state_fetch_timestamp)
+                self._save_goal_state()
                 self._update_host_plugin(new_goal_state.container_id, new_goal_state.role_config_name)
 
         except Exception as exception:
@@ -757,9 +755,9 @@ class WireClient(object): # pylint: disable=R0904
             self._host_plugin.update_container_id(container_id)
             self._host_plugin.update_role_config_name(role_config_name)
 
-    def _save_goal_state(self, previous_goal_state_fetch_timestamp):
+    def _save_goal_state(self):
         try:
-            self.goal_state_flusher.flush(previous_goal_state_fetch_timestamp)
+            self.goal_state_flusher.flush()
         except Exception as e: # pylint: disable=C0103
             logger.warn("Failed to save the previous goal state to the history folder: {0}", ustr(e))
 
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -541,8 +541,6 @@ class WireClient(object): # pylint: disable=R0904
         self._endpoint = endpoint
         self._goal_state = None
         self._last_try_update_goal_state_failed = False
-        self._previous_goal_state_timestamp = None
-        self._current_goal_state_timestamp = None
         self._host_plugin = None
         self.status_blob = StatusBlob(self)
         self.goal_state_flusher = StateFlusher(conf.get_lib_dir())
@@ -722,8 +720,9 @@ class WireClient(object): # pylint: disable=R0904
                 new_goal_state = GoalState.fetch_full_goal_state_if_incarnation_different_than(self, self._goal_state.incarnation)
 
             if new_goal_state is not None:
+                previous_goal_state_fetch_timestamp = self._goal_state.fetch_timestamp if self._goal_state else datetime.utcnow()
                 self._goal_state = new_goal_state
-                self._save_goal_state()
+                self._save_goal_state(previous_goal_state_fetch_timestamp)
                 self._update_host_plugin(new_goal_state.container_id, new_goal_state.role_config_name)
 
         except Exception as exception:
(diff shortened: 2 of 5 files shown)


azurelinuxagent/common/protocol/wire.py
```diff
@@ -542,6 +541,8 @@ def __init__(self, endpoint):
         self._endpoint = endpoint
         self._goal_state = None
         self._last_try_update_goal_state_failed = False
+        self._previous_goal_state_timestamp = None
```
narrieta, line 544: the timestamp should be in the goal state itself

we want to detach the goal state from the protocol (wire client) so we should not add more dependencies on the properties of the goal state here.

pgombar: Ah, good point, let me take that approach instead.

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -141,6 +142,7 @@ def __init__(self, wire_client, full_goal_state=False, base_incarnation=None):
             raise ProtocolError(msg="Error fetching goal state", inner=exception)
         finally:
             logger.info('Fetch goal state completed')
+            self.fetch_timestamp = datetime.utcnow()
```
narrieta, line 145: can we capture the time at which we fetch the actual goal state (in case there are delays for the other requests)

azurelinuxagent/common/protocol/wire.py
```diff
@@ -721,8 +720,9 @@ def update_goal_state(self, forced=False):
                 new_goal_state = GoalState.fetch_full_goal_state_if_incarnation_different_than(self, self._goal_state.incarnation)
 
             if new_goal_state is not None:
+                previous_goal_state_fetch_timestamp = self._goal_state.fetch_timestamp if self._goal_state else datetime.utcnow()
```
narrieta, line 723: "previous" doesn't quite fit with the naming of the other variables. At this point we have self._goal_state and new_goal_state, but we don't have a "previous" goal state.

i'd leave this function unchanged and simply do 
`
self.goal_state_flusher.flush(self._goal_state.fetch_timestamp)
`
below

pgombar: I initially did just that, but `self._goal_state` is `None` the first time we fetch a goal state.

azurelinuxagent/common/protocol/goal_state.py
pgombar: Sure, let me change that.

azurelinuxagent/common/protocol/wire.py
narrieta: If self._goal_state is None then saving the current files on disk using datetime.utcnow() would be incorrect. 

You probably want to check at the timestamp on GoalState.*.xml. Using that timestamp would actually work for all the cases since we save the file when we fetch the goal state. If you prefer to be more explicit than relying in the timestamp you could extend the XML with an extra property for the timestamp.

pgombar: One problem with checking the timestamp on GoalState.*.xml file -- `os.path.getmtime` returns the epoch time, but in the *local* timezone, and it's a mess to figure out what the timezone is and how to convert to UTC across different python versions.

I wanted to avoid the second approach since extending the XML would add extra computation (currently, we only parse the XML in the beginning and work with the string representation which is finally saved to disk), but I think I will take it.

pgombar: I decided to go with the timestamp of GoalState.*.xml since the code was getting to complicated and open for bugs in the latter approach with modifying the XML. This is because we internally only parse the XML once, at the beginning, and then just work with the string representation of it in memory and I didn't want to 1) add a hardcoded XML node with the timestamp as a string before saving the file, nor 2) add the XML node to the XML and re-parse it just in order to store it as a text file.

kevinclark19a (dismissed):

azurelinuxagent/common/utils/archive.py
```diff
@@ -41,71 +41,89 @@
...
 
-        d = os.path.join(self._source, ARCHIVE_DIRECTORY_NAME) # pylint: disable=C0103
-        if not os.path.exists(d):
+        directory = os.path.join(self._source, ARCHIVE_DIRECTORY_NAME)
+        if not os.path.exists(directory):
             try:
-                fileutil.mkdir(d)
-            except OSError as e: # pylint: disable=C0103
+                fileutil.mkdir(directory)
+            except OSError as e:  # pylint: disable=C0103
```
larohra, line 65: NIT: Missed linter

azurelinuxagent/common/utils/archive.py
```diff
@@ -41,71 +41,89 @@
...
                 if e.errno != errno.EEXIST:
                     logger.error("{0} : {1}", self._source, e.strerror)
 
-    def flush(self, timestamp):
+    def flush(self):
         files = self._get_files_to_archive()
-        if len(files) == 0: # pylint: disable=len-as-condition
+        if len(files) == 0:  # pylint: disable=len-as-condition
```
larohra, line 71: NIT: Missed linter

azurelinuxagent/common/utils/archive.py
```diff
@@ -41,71 +41,89 @@
...
+            self._archive(files, goal_state_timestamp)
         else:
             self._purge(files)
 
     def history_dir(self, timestamp):
         return os.path.join(self._source, ARCHIVE_DIRECTORY_NAME, timestamp.isoformat())
 
+    @staticmethod
+    def _get_latest_timestamp(files):
+        # Get the most recently modified GoalState.*.xml (if there are more than one) and use that timestamp for the archive name.
```
larohra, line 85: >(if there are more than one)

Is this a possibility? I thought we deleted all the old ones?

azurelinuxagent/common/utils/archive.py
```diff
@@ -41,71 +41,89 @@
...
+        latest_timestamp_ms = None
+        for current_file in files:
+            match = GOAL_STATE_PATTERN.match(current_file)
```
larohra, line 88: Just brainstorming here, how about getting the incarnation of the goalstate (goalstate.incarnation) and checking if that `GoalState.<incarnation>.xml` file already exists and using the timestamp of that rather than all files that match that regex?

azurelinuxagent/common/utils/archive.py
```diff
@@ -41,71 +41,89 @@
...
+            if not match:
+                continue
+
+            creation_time_ms = os.path.getmtime(current_file)
```
larohra, line 92: NIT: rename to `modified_time_ms` instead. Creation time suggests when the file was created but we're not checking that.

azurelinuxagent/common/utils/archive.py
```diff
@@ -41,71 +41,89 @@
...
-        d = self.history_dir(timestamp) # pylint: disable=C0103
+        directory = self.history_dir(timestamp)
 
         try:
-            fileutil.mkdir(d, mode=0o700)
+            fileutil.mkdir(directory, mode=0o700)
             return True
-        except IOError as e: # pylint: disable=C0103
-            logger.error("{0} : {1}".format(d, e.strerror))
+        except IOError as e:  # pylint: disable=C0103
```
larohra, line 125: NIT: Missed linter error

azurelinuxagent/common/utils/archive.py
```diff
@@ -148,32 +166,32 @@ def __ge__(self, other):
 
 
 class StateZip(State):
-    def __init__(self, path, timestamp): # pylint: disable=W0235
-        super(StateZip,self).__init__(path, timestamp)
+    def __init__(self, path, timestamp):  # pylint: disable=W0235
```
larohra, line 169: NIT: Missed linter error

azurelinuxagent/common/utils/archive.py
```diff
@@ -148,32 +166,32 @@ def __ge__(self, other):
...
+        super(StateZip, self).__init__(path, timestamp)
 
     def delete(self):
         os.remove(self._path)
 
 
 class StateDirectory(State):
-    def __init__(self, path, timestamp): # pylint: disable=W0235
+    def __init__(self, path, timestamp):  # pylint: disable=W0235
```
larohra, line 177: NIT: Missed linter error

azurelinuxagent/common/utils/archive.py
```diff
@@ -184,7 +202,7 @@ def __init__(self, lib_dir):
         if not os.path.isdir(self._source):
             try:
                 fileutil.mkdir(self._source, mode=0o700)
-            except IOError as e: # pylint: disable=C0103
+            except IOError as e:  # pylint: disable=C0103
```
larohra, line 205: NIT: Missed linter error

azurelinuxagent/common/utils/archive.py
```diff
@@ -41,71 +41,89 @@
...
 CACHE_PATTERNS = [
-    re.compile("^(.*)\.(\d+)\.(agentsManifest)$", re.IGNORECASE), # pylint: disable=W1401
-    re.compile("^(.*)\.(\d+)\.(manifest\.xml)$", re.IGNORECASE), # pylint: disable=W1401
-    re.compile("^(.*)\.(\d+)\.(xml)$", re.IGNORECASE) # pylint: disable=W1401
+    re.compile(r"^(.*)\.(\d+)\.(agentsManifest)$", re.IGNORECASE),
+    re.compile(r"^(.*)\.(\d+)\.(manifest\.xml)$", re.IGNORECASE),
+    re.compile(r"^(.*)\.(\d+)\.(xml)$", re.IGNORECASE)
 ]
 
+GOAL_STATE_PATTERN = re.compile(r"^(.*)GoalState\.(\d+)\.xml$", re.IGNORECASE)
```
narrieta, line 49: we do not want the "(.*)" before "GoalState", do we?

narrieta, line 49: Should be private to the module

pgombar: This regex is compared to the file's full path, e.g. `/var/lib/waagent/GoalState.N.xml` (or a random tmp directory while testing).

azurelinuxagent/common/utils/archive.py
pgombar: The problem is `self._goal_state` can be None and relying on the highest incarnation number on file is not deterministic either, since in the VM migrate scenario we can start getting incarnation 0 even though the previous one was N.

azurelinuxagent/common/utils/archive.py
pgombar: We do move all old ones in `save_goal_state`, but I was thinking of corner cases like ADE locking the filesystem or the agent dying/getting shut down before flushing the previous goal state, etc.

azurelinuxagent/common/utils/archive.py
pgombar: Good point, thanks.

azurelinuxagent/common/utils/archive.py
narrieta: if that is the case then try this instead:

`r"^(.*)/GoalState\.(\d+)\.xml$"`

azurelinuxagent/common/utils/archive.py
```diff
@@ -36,76 +36,94 @@
...
         else:
             self._purge(files)
 
     def history_dir(self, timestamp):
-        return os.path.join(self._source, ARCHIVE_DIRECTORY_NAME, timestamp.isoformat())
+        return os.path.join(self._source, _ARCHIVE_DIRECTORY_NAME, timestamp.isoformat())
+
+    @staticmethod
+    def _get_latest_timestamp(files):
+        # Get the most recently modified GoalState.*.xml (if there are more than one) and use that timestamp for the archive name.
```
narrieta, line 85: let's turn this comment into a docstring and specify when the function returns None

azurelinuxagent/common/utils/archive.py
```diff
@@ -36,76 +36,94 @@
...
+        latest_timestamp_ms = None
+        for current_file in files:
+            match = _GOAL_STATE_PATTERN.match(current_file)
+            if not match:
+                continue
+
+            modification_time_ms = os.path.getmtime(current_file)
+            if not latest_timestamp_ms or latest_timestamp_ms < modification_time_ms:
```
narrieta, line 93: is `not latest_timestamp_ms` comparing against zero or None -- let's be explicit

pgombar: I'm comparing against `None` since that is the initial state of the variable. If the variable is uninitialized, I set it to the retrieved timestamp. Why is that incorrect?

azurelinuxagent/common/utils/archive.py
pgombar: If `goal_state_timestamp` is None, that means there is no goal state file. This would typically happen when the first goal state comes in.

Actually, the code would return if there are no files:
```
files = self._get_files_to_archive()
        if not files:
            return
```
This ensures the `goal_state_timestamp` can never be None since there is a goal state file on disk. I can remove that check for timestamp.

azurelinuxagent/common/utils/archive.py
narrieta: i was probably thrown off by the current check. please do a explicit comparison against None, will help clarify the intention of the code

pgombar: Ok, will do that.

azurelinuxagent/common/utils/archive.py
pgombar: Ok, will do that.

azurelinuxagent/common/utils/archive.py
pgombar: Updated the docstring, let me know if it's clearer now.

azurelinuxagent/common/utils/archive.py
larohra: From the code, we set the `self._goal_state` right before calling the save function 

```python
            if new_goal_state is not None:
                self._goal_state = new_goal_state
                self._save_goal_state()
```

Essentially all I'm saying is on service restart, if we re-run the same goalstate the incarnation would be the same. So we would already have the `GoalState.<incarnation>.xml` file present on filesystem. Wouldn't it be better to just check the timestamp for that file? That would also take care of the other comment I had earlier where there might be a chance of having multiple `Goalstate.*.xml` files?

pgombar: Which <incarnation> are you referring to? `self.goal_state` can be None, and current goal state is the one we just fetched so that file is not saved to disk yet.

larohra: let's sync offline about this, it would just be faster that way

azurelinuxagent/common/utils/archive.py
```diff
@@ -48,10 +48,10 @@
 
 _GOAL_STATE_PATTERN = re.compile(r"^(.*)/GoalState\.(\d+)\.xml$", re.IGNORECASE)
 
-# 2018-04-06T08:21:37.142697
-# 2018-04-06T08:21:37.142697.zip
-_ARCHIVE_PATTERNS_DIRECTORY = re.compile(r"^\d{4}\-\d{2}\-\d{2}T\d{2}:\d{2}:\d{2}\.\d+$")
-_ARCHIVE_PATTERNS_ZIP = re.compile(r"^\d{4}\-\d{2}\-\d{2}T\d{2}:\d{2}:\d{2}\.\d+\.zip$")
+# 2018-04-06T08:21:37.142697_incarnation_N
+# 2018-04-06T08:21:37.142697_incarnation_N.zip
+_ARCHIVE_PATTERNS_DIRECTORY = re.compile(r"^\d{4}\-\d{2}\-\d{2}T\d{2}:\d{2}:\d{2}\.\d+_incarnation_(\d+)$")
```
larohra, line 53: For the transitionary phase, do we need to care about the archives created by the older agents? Or can we ignore those for now? Maybe you can make the regex conditional to look for archives that have `_incarnation_(\d+)` and also at the ones that dont have it (the older ones)

tests/utils/test_archive.py
```diff
@@ -32,19 +32,28 @@ def tearDown(self):
...
     @property
     def history_dir(self):
         return os.path.join(self.tmp_dir, 'history')
 
+    @staticmethod
+    def _parse_archive_name(name):
+        # Name can be a directory or a zip
+        # '0000-00-00T00:00:00.000000_incarnation_0'
+        # '0000-00-00T00:00:00.000000_incarnation_0.zip'
+        timestamp_str, _, incarnation_ext = name.split("_")
```
larohra, line 53: Easier way would be just to use `_incarnation_` to split instead of `_` :)

larohra (commented):
some small comments

azurelinuxagent/common/utils/archive.py
pgombar: Good point, we only purge every 24 hours so there can definitely be some leftovers.
Let me make the incarnation addition optional.

tests/utils/test_archive.py
pgombar: D'oh. Thanks! :smile:

larohra (dismissed):

azurelinuxagent/common/utils/archive.py
```diff
@@ -36,76 +36,105 @@
...
+    def flush(self):
         files = self._get_files_to_archive()
-        if len(files) == 0: # pylint: disable=len-as-condition
+        if not files:
             return
 
-        if self._mkdir(timestamp):
-            self._archive(files, timestamp)
+        archive_name = self._get_archive_name(files)
+        if archive_name and self._mkdir(archive_name):
```
narrieta, line 76: let's do an explicit comparison against None

narrieta, line 76: if for any reason there is no goal state file we should not purge the rest of the files

larohra (approved):

narrieta (approved):

PR #2059, opened 2020-11-02
Fix depends on errors

larohra:
## Description
This PR takes care of a couple of things - 
- For depends-on feature, we first check the handler status and then start polling for the extension status in order to fail fast if the handler already failed for some reason.
- Improved error handling around the depends-on scenario. Currently if any errors are thrown in the `wait_for_handler_successful_completion` function, we can go on an infinite error reporting loop until a new goalstate comes in. This can blast Kusto, fill up logs and hide the original issue. 
- Modified the logic for fetching SeqNo from File-System and added plans to eventually deprecate that function (we should only rely on seqNo from the GoalState rather than the filesystem)

TFS bugs - 
- [Bug 7101306](https://msazure.visualstudio.com/One/_workitems/edit/7101306): Wrongfully waiting for dependent extension status if extension installation failed
- [Bug 7101366](https://msazure.visualstudio.com/One/_workitems/edit/7101366): Wrongfully waiting for 90 mins even though the extension timedout after 15mins

Issue # 

---

### PR information


azurelinuxagent/ga/exthandlers.py
```diff
@@ -1070,26 +1092,47 @@ def update_with_install(self, uninstall_exit_code=None):
...
+            logger.verbose("Error fetching sequence number from config files: {0}".format(ustr(error)))
+
         return seq_no
 
     def get_status_file_path(self, extension=None):
         path = None
-        seq_no = self._get_largest_seq_no()
+        # Todo: Remove check on filesystem for fetching sequence number (legacy behaviour).
+        # We should technically only fetch the sequence number from GoalState and not rely on the filesystem at all,
+        # But since we still have Kusto data from the operation below (~0.000065% VMs are still reporting
```
larohra, line 1132: `~0.000065% VMs `  - This number is still pretty low, so if you guys agree I dont mind sending a separate PR to remove this function altogether (it would need a separate PR to handle all the callers and UTs properly and would also keep the state clean)

tests/ga/test_extension.py
```diff
@@ -2709,6 +2744,23 @@ def test_collect_ext_status(self, mock_lib_dir, *args):
         self.assertEqual(sub_status.message, None)
         self.assertEqual(sub_status.status, ValidHandlerStatus.success)
 
+    @patch("azurelinuxagent.common.conf.get_lib_dir")
+    def test_it_should_collect_ext_status_even_when_config_dir_deleted(self, mock_lib_dir, *args):
```
larohra, line 2748: This important case was failing in the older code where due to missing proper error handling in the  `_get_last_modified_seq_no_from_config_files` function (previously `_get_largest_seq_no`), we could not even collect the status properly. Added a basic test case for it too

azurelinuxagent/ga/exthandlers.py
```diff
@@ -391,48 +391,70 @@ def handle_ext_handlers(self, etag=None):
...
+                      message=error_message,
+                      log_event=True)
+            return False
+
+        try:
+            handler_i = ExtHandlerInstance(ext_handler, self.protocol)
+
+            # First check if HandlerStatus was a success, if the Handler failed, no need to poll for extension status
+            handler_status = handler_i.get_handler_status()
+            if not handler_status:
```
narrieta, line 421: when we are uninstalling the extension by the time we get here the file will already have been deleted. we should not flag that as an error

narrieta (changes requested):

azurelinuxagent/ga/exthandlers.py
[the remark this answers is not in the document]
larohra: Done

azurelinuxagent/ga/exthandlers.py
```diff
@@ -374,11 +373,16 @@ def _extension_processing_allowed(self):
         return True
 
     def handle_ext_handlers(self, etag=None):
-        if self.ext_handlers.extHandlers is None or \
-                len(self.ext_handlers.extHandlers) == 0: # pylint: disable=len-as-condition
+        if not self.ext_handlers.extHandlers:
             logger.verbose("No extension handler config found")
             return
 
+        if self.last_etag == etag:
```
larohra, line 380: Moving the check one level up to ensure we dont check for dependsOn status unless the incarnation changes

azurelinuxagent/ga/exthandlers.py
```diff
@@ -391,61 +395,77 @@ def handle_ext_handlers(self, etag=None):
...
+            handler_i = ExtHandlerInstance(ext_handler, self.protocol)
+
+            # First check if HandlerStatus was a success, if the Handler failed, no need to poll for extension status
+            handler_status = handler_i.get_handler_status()
+            if handler_status is None:
+                msg = "No HandlerStatus available for Handler: {0}, marking the extension as failed".format(
+                    ext_handler.name)
+                return _report_error_event_and_return_false(msg)
+
+            if handler_status.code != ExtensionErrorCodes.PluginSuccess:
```
pgombar, line 430: Here you are checking `handler_status.code` (`ExtensionErrorCodes`), but below you are evaluating against `status` (`ValidHandlerStatus`). Why the discrepancy? Can we be consistent?

tests/ga/test_extension.py
pgombar: Thanks for fixing this!

azurelinuxagent/ga/exthandlers.py
larohra: So the `handler_status` is written by the agent, and we use `ExtensionErrorCodes` to specify the state of the handler. The status on the other hand is written by the extension in the status file and they use String representations of `ValidHandlerStatus` to specify the status of the execution of the extension script. That's the primary reason why you see a discrepancy. `ExtensionErrorCodes` are more verbose (multiple options) while the `ValidHandlerStatus` is not (lesser options). Since both were designed for different features, its hard to make it consistent.

pgombar: So why not check `handler_status != ValidHandlerStatus.success` here?

larohra: `handler_status` is an object of `ExtHandlerStatus()` class here. But the check below is a basic string comparison (`status_from_status_file == "success"`) (these terminologies can be super confusing lol)

pgombar: Sorry, what I meant was check `handler_status.status != ValidHandlerStatus.success`. I don't see why we need to check against the code first, then later against status if we can do status in both cases.

larohra: The `handler_status.status` is either `Ready` or `NotReady`. We need to report this back to CRP so we cant use this to check the `ValidHandlerStatus.success`. 

Additionally I personally like the code (int) check more since its more verbose and elaborate. `ValidHandlerStatus` is something that the extensions report so that's limited too. The `code` on the other hand is set by the agent, so it can be set to something more specific. Like for now, say we want to ignore the `HeartBeat` failures reported in the `ExtHandlerStatus` then we can just ignore that code here. It gives us more flexibility here if anything.

At the end of the day, they're just 2 different constructs with similar naming which is probably what's throwing you off.

azurelinuxagent/ga/exthandlers.py
narrieta: Move it one more level up and do it along with ` if self._extension_processing_allowed():`?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -374,11 +373,16 @@ def _extension_processing_allowed(self):
...
-                len(self.ext_handlers.extHandlers) == 0: # pylint: disable=len-as-condition
+        if not self.ext_handlers.extHandlers:
             logger.verbose("No extension handler config found")
             return
 
+        if self.last_etag == etag:
+            # Skip processing if GoalState incarnation did not change
+            return
+
+        logger.info("New incarnation detected, starting processing of GoalState with incarnation: {0}".format(etag))
```
narrieta, line 384: Let's rephrase this message and make it look symmetric to

 ` logger.info('ProcessGoalState completed [incarnation {0}; {1} ms]',`

azurelinuxagent/ga/exthandlers.py
```diff
@@ -391,61 +395,77 @@ def handle_ext_handlers(self, etag=None):
...
             dep_level = ext_handler.sort_key()
-            if dep_level >= 0 and dep_level < max_dep_level: # pylint: disable=R1716
+            if 0 <= dep_level < max_dep_level:
                 if not self.wait_for_handler_successful_completion(ext_handler, wait_until):
                     logger.warn("An extension failed or timed out, will skip processing the rest of the extensions")
                     break
 
     def wait_for_handler_successful_completion(self, ext_handler, wait_until):
-        '''
+        """
```
narrieta, line 404: let's remove "successful" from `wait_for_handler_successful_completion` since we wait for completion, either success or failure

narrieta (commented):
a few minor comments

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1076,26 +1094,47 @@ def update_with_install(self, uninstall_exit_code=None):
...
-                        if curr_seq_no > seq_no:
+                        curr_modified_time = os.path.getmtime(item_path)
+                        if curr_modified_time > largest_modified_time:
                             seq_no = curr_seq_no
+                            largest_modified_time = curr_modified_time
                 except (ValueError, IndexError, TypeError):
                     self.logger.verbose("Failed to parse file name: {0}", item)
                     continue
+        except Exception as error:
+            logger.verbose("Error fetching sequence number from config files: {0}".format(ustr(error)))
```
narrieta, line 1126: seq_no could have change at line 1120 so we need to reset it to -1 here

azurelinuxagent/ga/exthandlers.py
larohra: Changed the logic, not relying on HandlerStatus anymore

azurelinuxagent/ga/exthandlers.py
larohra: Good point, done

azurelinuxagent/ga/exthandlers.py
larohra: Good point, added the change

azurelinuxagent/ga/exthandlers.py
larohra: Done

azurelinuxagent/ga/exthandlers.py
larohra: Done

azurelinuxagent/ga/exthandlers.py
larohra: Changed the logic to not use handler status now and return True or False if handler was successfully executed

azurelinuxagent/ga/exthandlers.py

pgombar (approved):

azurelinuxagent/ga/exthandlers.py
```diff
@@ -371,11 +371,14 @@ def _extension_processing_allowed(self):
                 logger.info("Extension handling is on hold")
                 return False
 
+        if self.last_etag == etag:
```
narrieta, line 374: Let's move this check to the if statement at line 281 or think of a better name for '_extension_processing_allowed' to reflect better what it is doing.

larohra: I think the name is pretty apt. The name `_extension_processing_allowed` implies that we only proceed if its allowed and the incarnation check is one of those checks that we use to determine if its allowed.
How about a docstring? 
Or maybe renaming it to `_extension_processing_allowed_if_incarnation_changed` maybe?

narrieta: no, i don't think the name is pretty apt... the check for the incarnation seems hidden in this function. _extension_processing_allowed_if_incarnation_changed doesn't seem apt either. I'd move the check one level up

narrieta (approved):

pgombar (approved):

PR #2050, opened 2020-10-22
Switch to run command changes, + provisioning changes that need to be reverted.

kevinclark19a:
## Description

---

### PR information
- [ ] The title of the PR is clear and informative.
- [ ] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).
- [ ] Except for special cases involving multiple contributors, the PR is started from a fork of the main repository, not a branch.
- [ ] If applicable, the PR references the bug/issue that it fixes in the description.
- [ ] New Unit tests were added for the changes made and Travis.CI is passing.

### Quality of Code and Contribution Guidelines
- [ ] I have read the [contribution guidelines](https://github.com/Azure/WALinuxAgent/blob/master/.github/CONTRIBUTING.md).

PR #2069, opened 2020-11-05
Fixed bug causing "MAC verified OK" message

kevinclark19a:
## Description

A bug in older versions of openssl was causing the message "MAC verified OK" to be sent out on stderr during the processing of every goal state in the agent. Before PR #2050, the agent would suppress the stderr output. This PR adds that suppression back.

---

### PR information


azurelinuxagent/common/utils/cryptutil.py
```diff
@@ -88,7 +88,8 @@ def decrypt_p7m(self, p7m_file, trans_prv_file, trans_cert_file, pem_file):
 
             first_proc = subprocess.Popen(first_cmd, stdout=subprocess.PIPE)
 
-            second_proc = subprocess.Popen(second_cmd, stdin=first_proc.stdout, stdout=subprocess.PIPE)
+            second_proc = subprocess.Popen(second_cmd, stdin=first_proc.stdout,
+                stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
larohra, line 92: A weird question. The default value is `stderr=None`. If it was None, how were we getting the weird output message in the console? Any idea?

larohra (approved):
1 small question else LGTM

azurelinuxagent/common/utils/cryptutil.py
kevinclark19a: My understanding is that `None` defaults to passing through the output to the agent process.

larohra: Ahh you're right - 

>stdout=None means that subprocess prints to whatever place your script prints. stdout=PIPE means that subprocess' stdout is redirected to a pipe that you should read e.g., using process.communicate() to read all at once or using process.stdout object to read via a file/iterator interfaces

pgombar (approved):

PR #2074, opened 2020-11-09
Multiply CPU percentage by number of processors

narrieta:
Changed the CPU usage to multiply by the number of CPUs. This matches the input that systemd takes when setting the CPU quota.

Also, removed some dead code.

### DIFF
--- a/azurelinuxagent/common/cgroup.py
+++ b/azurelinuxagent/common/cgroup.py
@@ -131,27 +131,6 @@ class CGroup(object):
                                  ' Internal error: {1}'.format(self.path, ustr(e)))
         return False
 
-    def get_tracked_processes(self):
-        """
-        :return: List of Str (Pids). Will return an empty string if we couldn't fetch any tracked processes.
-        """
-        procs = []
-        try:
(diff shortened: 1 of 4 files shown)


azurelinuxagent/common/cgroup.py
```diff
@@ -131,27 +131,6 @@ def is_active(self):
                                  ' Internal error: {1}'.format(self.path, ustr(e)))
         return False
 
-    def get_tracked_processes(self):
```
narrieta, line 134: dead code

azurelinuxagent/common/cgroupstelemetry.py
```diff
@@ -110,13 +110,6 @@ def poll_all_tracked():
 
         return metrics
 
-    @staticmethod
```
narrieta, line 113: Dead code

larohra (dismissed):

pgombar (dismissed):

pgombar (approved):

larohra (approved):

PR #2081, opened 2020-11-18
Parse InVMGoalStateMetaData from Extension Config

larohra:
## Description
ExtensionConfig file contains extra metadata that's helpful for debugging. This PR parses the new things and logs activity Id and Correlation Ids for an easier debug experience

Issue # 

---

### PR information


pgombar (dismissed):

kevinclark19a (approved):

azurelinuxagent/ga/exthandlers.py
```diff
@@ -290,10 +290,13 @@ def run(self):
 
             if self._extension_processing_allowed() and self._incarnation_changed(etag):
                 activity_id, correlation_id = self.get_activity_and_correlation_id()
-                logger.info(
-                    "ProcessGoalState started [incarnation {0}; Activity Id: {1}; Correlation Id: {2}]".format(etag,
-                                                                                                               activity_id,
-                                                                                                               correlation_id))
+
+                log_msgs = ["Incarnation: {0}".format(etag)]
+                if activity_id is not None:
```
pgombar, line 295: You added this handling because the feature to add activityid/correlationid is not released everywhere yet.
What if, when this is released, either value is None due to a bug in the xml or in our parsing logic? Wouldn't we want to log it to surface the issue? What is the harm of always logging the ids regardless of their values? They will be None for the time being, but this will change as that feature is released.

larohra: Hmm good point. I added this because there's no ETA on the release in CRP and I didnt want the log messages to look dirty. 

This is how it would look if those values are None - 
`ProcessGoalState completed [Incarnation: 1; 10 ms; Activity Id: ; Correlation Id: ]`

This looks very dirty. 
I can either make them say 'NA' or something or just not add them at all. 

I do like your point though. Let me replace the empty spaces with NA or something to make it more apparent

pgombar: Yeah, basically handling a non-value would do it. Thanks!

pgombar (dismissed):

azurelinuxagent/ga/exthandlers.py
```diff
@@ -275,6 +275,12 @@ def _incarnation_changed(self, etag):
         # Skip processing if GoalState incarnation did not change
         return self.last_etag != etag
 
+    def get_activity_and_correlation_id(self):
+
+        def check_empty(value): return value if value else "NA"
```
narrieta, line 280: let's get in the habit of making explicit comparisons against None ("if value is not None")

azurelinuxagent/ga/update.py
```diff
@@ -362,20 +362,21 @@ def run(self, debug=False):  # pylint: disable=R0912
                     if last_etag != exthandlers_handler.last_etag:
                         self._ensure_readonly_files()
                         duration = elapsed_milliseconds(utc_start)
-                        logger.info('ProcessGoalState completed [incarnation {0}; {1} ms]',
-                                    exthandlers_handler.last_etag,
-                                    duration)
+                        activity_id, correlation_id = exthandlers_handler.get_activity_and_correlation_id()
+                        msg = 'ProcessGoalState completed [Incarnation: {0}; {1} ms; Activity Id: {2}; Correlation Id: {3}]'.format(
```
narrieta, line 366: Let's also log it when we fetch the goal state (you can remove it from here if you don't want to log it twice)

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -290,6 +290,7 @@ def __init__(self, xml_text):
         self.xml_text = xml_text
         self.ext_handlers = ExtHandlerList()
         self.vmagent_manifests = VMAgentManifestList()
+        self.in_vm_metadata = InVMGoalStateMetaData()
```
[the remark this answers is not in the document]
larohra: Done

azurelinuxagent/common/protocol/restapi.py
```diff
@@ -125,6 +126,25 @@ def sort_key(self):
         return level
 
 
+# too-few-public-methods<R0903> Disabled: More public methods are not needed here.
```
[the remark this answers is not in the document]
larohra: Yeah I think we did, let me send a separate PR for this

azurelinuxagent/common/protocol/restapi.py
```diff
@@ -125,6 +126,25 @@ def sort_key(self):
...
+class InVMGoalStateMetaData(DataContract):  # pylint: disable=R0903
+    """
+    Object for parsing the GoalState MetaData received from CRP
+    Eg: <InVMGoalStateMetaData inSvdSeqNo="2" createdOnTicks="637405409304121230" activityId="555e551c-600e-4fb4-90ba-8ab8ec28eccc" correlationId="400de90b-522e-491f-9d89-ec944661f531" />
+    """
+    def __init__(self):
+        self.in_svd_seq_no = None
+        self.created_on_ticks = None
```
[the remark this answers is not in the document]
larohra: I think so but let me confirm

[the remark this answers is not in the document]
larohra: I saw this comment in the CRP code - 
>/// We need to keep the value "CreatedOnTicks" in the serialized SVD because the Windows Guest Agent
        /// Depends on it. We should be able to remove this when FastTrack V2 is world wide, as this will no
        /// longer be relevant then.

Since this is destined to be removed from the GoalState eventually, I'm going to remove the `created_on_ticks` property altogether from the code. We dont need it anyways right now

azurelinuxagent/ga/exthandlers.py
larohra: Done (and will keep in mind from next time)

azurelinuxagent/ga/update.py
larohra: I already have that - 

>logger.info(
                    "ProcessGoalState started [Incarnation: {0}; Activity Id: {1}; Correlation Id: {2}]".format(etag,
                                                                                                                activity_id,
                                                                                                                correlation_id))

Is that not what you mean?

azurelinuxagent/common/protocol/restapi.py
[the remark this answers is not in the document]
larohra: Removed

azurelinuxagent/common/protocol/restapi.py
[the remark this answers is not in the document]
narrieta: ok

azurelinuxagent/ga/exthandlers.py
narrieta: Great! This is a good example of why we need to be explicit. When I read this code I did not realize that empty string was a valid value. Doing 'if value not in (None, '')' this instead would clarify the intention of this code.

Although scripting (and other) languages provide this kind of flexibility, as the size of production code grows, it is better to do explicit checks, rather than debug obscure bugs. In this specific case being explicit vs being the implicit may not make a difference, but in other cases it does. Also, as the code evolves an implicit check that was initially OK may end up being incorrect.

`Though I'll keep in mind to always have a None check if there is only None to be compared against (and not empty strings or empty arrays [])
`

Yes, when checking against None please also always be explicit. PEP 8 touches briefly on this:

`
Also, beware of writing if x when you really mean if x is not None -- e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!`

azurelinuxagent/ga/update.py
narrieta: cool, i missed that one

narrieta:
@larohra why was created_on_ticks removed?

larohra:
> @larohra why was created_on_ticks removed?

As per @jscalev and the comment in the CRP code - 
```
/// We need to keep the value "CreatedOnTicks" in the serialized SVD because the Windows Guest Agent
/// Depends on it. We should be able to remove this when FastTrack V2 is world wide, as this will no
/// longer be relevant then.
```
[Code link](https://msazure.visualstudio.com/One/_git/Compute-CPlat-Core?path=%2Fsrc%2FCRP%2FDev%2FComputeResourceProvider%2FCore.Shared%2FInVMArtifactsProfileManagement%2FModels%2FInVMArtifactsProfileDataModel.cs&version=GBmaster&line=43&lineEnd=44&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)

Looks like the `createdOnTicks` is going away once FT v2 comes in. If that's the case, I figured we shouldn't even add it as the Linux Agent doesnt really need it.

azurelinuxagent/ga/exthandlers.py
larohra: Makes sense, I'll send an update

larohra: Done

azurelinuxagent/common/protocol/restapi.py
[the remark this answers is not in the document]
larohra: TFS link - https://msazure.visualstudio.com/One/_workitems/edit/8910632

narrieta (dismissed):

narrieta (approved):

pgombar (approved):

PR #2082, opened 2020-11-19
Support Python 3.9 and add it to Travis CI

pgombar:
## Description

We need to address two breaking changes in Python 3.9:
1) [bpo-38916](https://bugs.python.org/issue38916): array.array: Remove tostring() and fromstring() methods. They were aliases to tobytes() and frombytes(), deprecated since Python 3.2.
2) [bpo-39377](https://bugs.python.org/issue39377): Removed encoding option from json.loads(). It has been deprecated since Python 3.1.

Python 3.9 changelog can be found [here](https://docs.python.org/3/whatsnew/changelog.html#changelog).

---

### PR information

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,12 @@ matrix:
           SETUPOPTS=""
           PYLINTOPTS="--rcfile=ci/3.6.pylintrc"
           PYLINTFILES="azurelinuxagent setup.py makepkg.py tests"
+    - python: 3.9
+      env:
+        - >-
+          NOSEOPTS="--verbose --with-timer" SETUPOPTS=""
+          PYLINTOPTS="--rcfile=ci/3.6.pylintrc"
+          PYLINTFILES="azurelinuxagent setup.py makepkg.py tests"
 
 install:
   - pip install -r requirements.txt
--- a/azurelinuxagent/common/future.py
+++ b/azurelinuxagent/common/future.py
@@ -133,7 +133,7 @@ def get_openwrt_platform():
     return result
 
 
-def array_to_string_or_bytes(buff):
+def array_to_bytes(buff):
     # Python 3.9 removed the tostring() method on arrays, the new alias is tobytes()
     if sys.version_info[0] == 2:
         return buff.tostring()
--- a/azurelinuxagent/common/future.py
+++ b/azurelinuxagent/common/future.py
@@ -131,3 +131,14 @@ def get_openwrt_platform():
                 if product_matches.group(1) == "OpenWrt":
                     result[0] = "openwrt"
     return result
+
+
+def array_to_string_or_bytes(buffer):
+    # Python 3.9 removed the tostring() method on arrays, the new alias is tobytes()
+    if sys.version_info[0] == 2:
+        return buffer.tostring()
+
+    if sys.version_info[0] == 3 and sys.version_info[1] <= 8:
+        return buffer.tostring()
+
+    return buffer.tobytes()
--- a/azurelinuxagent/common/future.py
+++ b/azurelinuxagent/common/future.py
@@ -133,12 +133,12 @@ def get_openwrt_platform():
     return result
 
 
-def array_to_string_or_bytes(buffer):
+def array_to_string_or_bytes(buff):
     # Python 3.9 removed the tostring() method on arrays, the new alias is tobytes()
     if sys.version_info[0] == 2:
-        return buffer.tostring()
+        return buff.tostring()
 
     if sys.version_info[0] == 3 and sys.version_info[1] <= 8:
-        return buffer.tostring()
+        return buff.tostring()
 
-    return buffer.tobytes()
+    return buff.tobytes()
--- a/azurelinuxagent/common/osutil/bigip.py
+++ b/azurelinuxagent/common/osutil/bigip.py
@@ -24,6 +24,8 @@ import socket
 import struct
 import time
 
+from azurelinuxagent.common.future import array_to_string_or_bytes
+
 try:
     # WAAgent > 2.1.3
     import azurelinuxagent.common.logger as logger
@@ -287,12 +289,8 @@ class BigIpOSUtil(DefaultOSUtil):
         if retsize == (expected * struct_size):
             logger.warn(('SIOCGIFCONF returned more than {0} up '
                          'network interfaces.'), expected)
-        try:
-            # Python 3.9 removed the tostring() method on arrays, tobytes() is the new alias
-            sock = buff.tostring()
-        except AttributeError:
-            sock = buff.tobytes()  # pylint: disable=no-member
 
+        sock = array_to_string_or_bytes(buff)
         for i in range(0, struct_size * expected, struct_size):
             iface = self._format_single_interface_name(sock, i)
 
--- a/azurelinuxagent/common/osutil/bigip.py
+++ b/azurelinuxagent/common/osutil/bigip.py
@@ -24,7 +24,7 @@ import socket
 import struct
 import time
 
-from azurelinuxagent.common.future import array_to_string_or_bytes
+from azurelinuxagent.common.future import array_to_bytes
 
 try:
     # WAAgent > 2.1.3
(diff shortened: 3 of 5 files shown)


kevinclark19a (approved):

azurelinuxagent/common/osutil/bigip.py
```diff
@@ -287,7 +287,12 @@ def get_first_if(self):
         if retsize == (expected * struct_size):
             logger.warn(('SIOCGIFCONF returned more than {0} up '
                          'network interfaces.'), expected)
-        sock = buff.tostring()
+        try:
+            # Python 3.9 removed the tostring() method on arrays, tobytes() is the new alias
+            sock = buff.tostring()
+        except AttributeError:
```
larohra, line 293: Instead of try-except, wouldn't it be better to have a check for python version instead? (To ensure we dont catch another error as part of this)

azurelinuxagent/common/osutil/default.py
```diff
@@ -798,7 +798,11 @@ def _get_all_interfaces(self):
             logger.warn(('SIOCGIFCONF returned more than {0} up '
                          'network interfaces.'), expected)
 
-        ifconf_buff = buff.tostring()
+        try:
+            # Python 3.9 removed the tostring() method on arrays, tobytes() is the new alias
+            ifconf_buff = buff.tostring()
+        except AttributeError:
```
larohra, line 804: Same comment here

.travis.yml
```diff
@@ -51,6 +51,12 @@ matrix:
           SETUPOPTS=""
           PYLINTOPTS="--rcfile=ci/3.6.pylintrc"
           PYLINTFILES="azurelinuxagent setup.py makepkg.py tests"
+    - python: 3.9
```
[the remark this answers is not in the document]
pgombar: I just checked, we haven't been running coverage at all since Kevin added the helper scripts. I'm in favour of removing it completely as I never found much use for it... thoughts?

azurelinuxagent/common/osutil/bigip.py
```diff
@@ -287,7 +287,12 @@ def get_first_if(self):
...
+            sock = buff.tobytes()  # pylint: disable=no-member
```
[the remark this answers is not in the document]
pgombar: I had to add this since pylint for Python 2.7 was complaining about the method not existing.

azurelinuxagent/common/osutil/default.py
```diff
@@ -798,7 +798,11 @@ def _get_all_interfaces(self):
...
+            ifconf_buff = buff.tobytes()  # pylint: disable=no-member
```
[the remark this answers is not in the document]
pgombar: I had to add this since pylint for Python 2.7 was complaining about the method not existing.

azurelinuxagent/common/osutil/bigip.py
pgombar: I think AttributeError is already pretty specific:

> Raised when an attribute reference (see Attribute references) or assignment fails. 

From [here](https://docs.python.org/3/library/exceptions.html#AttributeError).
For example, we wouldn't catch TypeErrors so I don't think we are catching anything we're not supposed to here.

.travis.yml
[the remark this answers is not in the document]
larohra: haha wow! Now that you mention it, I do agree haven't seen that either in a while lol. 
I dont mind either tbh, I personally dont check it a lot but maybe the external contributors do? Not sure

azurelinuxagent/common/osutil/bigip.py
larohra: But I think the main thing we're trying to catch here is just the issues raised by py3.9 right? I mainly dont want some other error to be masked by this check (not sure though the feasibility of another error being raised as an AttributeError tbh). It also makes the code more apparent on what we're trying to achieve rather than inferencing it behind some other check.

azurelinuxagent/common/osutil/bigip.py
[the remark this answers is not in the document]
larohra: Ahh ok. Can you add a small message on top of this line? That's our standard process for the cases where pylint disable is necessary. 

Something like this - 
```python
# too-many-instance-attributes<R0902> Disabled: class complexity is OK
    # invalid-name<C0103> Disabled: class is private, so name starts with __
    class __Impl(object):  # pylint: disable=R0902,C0103
```

azurelinuxagent/common/osutil/default.py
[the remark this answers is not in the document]
larohra: same as above

azurelinuxagent/common/osutil/bigip.py
narrieta: +1: should be version specific, plus it should be somehow abstracted so that we do not have version checks peppered throughout the code

.travis.yml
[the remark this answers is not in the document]
narrieta: i am inclined to remove it

azurelinuxagent/common/osutil/bigip.py
kevinclark19a: I agree; should there be an alias in our `future.py`?

pgombar: Ok, adding a helper method to `future.py`

.travis.yml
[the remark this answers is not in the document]
pgombar: I won't make any changes to `.travis.yml` outside Py 3.9 since we are replacing it with GitHub Actions.

azurelinuxagent/common/future.py
```diff
@@ -130,4 +130,15 @@ def get_openwrt_platform():
             elif product_matches:
                 if product_matches.group(1) == "OpenWrt":
                     result[0] = "openwrt"
-    return result
\ No newline at end of file
+    return result
+
+
+def array_to_string_or_bytes(buff):
```
larohra, line 136: NIT: Wouldn't this make more sense in the `textutil` module? 
Basically have a check for like `is_py_39` in `future.py` and then this function in `textutil`?

A minor suggestion but I'm good with this too :)

larohra (dismissed):
1 minor suggestion else LGTM

azurelinuxagent/common/future.py
pgombar: The way I saw it was that this is a breaking change for future Py versions so it would make sense to have it in `future.py`. That module already contains other breaking changes like the move from `platform` to `distro`. But now that you mention `textutil`, I see we have some version-specific code there too, so that also makes sense.

Anyways, if somebody else feels strongly about this I can make the change, but I don't mind it as it is now.

kevinclark19a (approved):

azurelinuxagent/common/future.py
narrieta: Instead of introducing a new function, we should prepare the code for how it would look in the future, something similar to what we do with httpclient

```

if sys.version_info[0] == 3:	if sys.version_info[0] == 3:
    import http.client as httpclient
elif sys.version_info[0] == 2:	elif sys.version_info[0] == 2:
    import httplib as httpclient

```

pgombar: The problem with that is that we are not importing anything. The API on the `array` data structure changed from `[array].tostring()` to `[array].tobytes()`. Instead of extending the array class and defining our own API depending on the Python version, I found this solution much simpler.

If you have a better idea how to do this, please let me know.

narrieta: if python < 3.9 : tobytes = tostring

narrieta: hold on, this is a class member, 1 sec

narrieta: in what class is tostring() defined?

pgombar: `class array.array`
https://docs.python.org/3/library/array.html#array.array.tobytes

narrieta: ok, since it is a builtin, it won't be possible. I would suggest name the function simply array_to_bytes

pgombar: Yep, this was my conclusion too. Ok, will do the rename.

narrieta (approved):

larohra (approved):

PR #2086, opened 2020-11-23
Add support for redirection to shellutil.run_command

narrieta:
Extended shellutil.run_command to support input/output redirection.

Added shellutil.run_pipe to execute command pipes.

### DIFF
--- a/azurelinuxagent/common/logcollector.py
+++ b/azurelinuxagent/common/logcollector.py
@@ -115,7 +115,7 @@ class LogCollector(object):  # pylint: disable=R0903
             return ustr(output, encoding="utf-8", errors="backslashreplace")
 
         try:
-            process = subprocess.Popen(command, stdin=subprocess.PIPE, stdout=stdout, stderr=subprocess.PIPE, shell=False)
+            process = subprocess.Popen(command, stdout=stdout, stderr=subprocess.PIPE, shell=False)
             stdout, stderr = process.communicate()
             return_code = process.returncode
         except Exception as e:  # pylint: disable=C0103
--- a/azurelinuxagent/common/utils/cryptutil.py
+++ b/azurelinuxagent/common/utils/cryptutil.py
@@ -155,7 +155,7 @@ class CryptUtil(object):
         try:
             decoded = base64.b64decode(encrypted_password)
             args = DECRYPT_SECRET_CMD.format(self.openssl_cmd, private_key).split(' ')
-            output = shellutil.run_command(args, input=decoded, stderr=subprocess.STDOUT, encode_output=False)
+            output = shellutil.run_command(args, input=decoded, stderr=subprocess.STDOUT, encode_input=False, encode_output=False)
             return output.decode('utf-16')
         except shellutil.CommandError as command_error:
             raise subprocess.CalledProcessError(command_error.returncode, "openssl cms -decrypt", output=command_error.stdout)
--- a/azurelinuxagent/common/utils/cryptutil.py
+++ b/azurelinuxagent/common/utils/cryptutil.py
@@ -155,7 +155,7 @@ class CryptUtil(object):
         try:
             decoded = base64.b64decode(encrypted_password)
             args = DECRYPT_SECRET_CMD.format(self.openssl_cmd, private_key).split(' ')
-            output = shellutil.run_command(args, input=decoded, stderr=subprocess.STDOUT)
+            output = shellutil.run_command(args, input=decoded, stderr=subprocess.STDOUT, encode_output=False)
             return output.decode('utf-16')
         except shellutil.CommandError as command_error:
             raise subprocess.CalledProcessError(command_error.returncode, "openssl cms -decrypt", output=command_error.stdout)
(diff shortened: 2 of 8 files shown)


azurelinuxagent/common/logcollector.py
```diff
@@ -115,7 +115,7 @@ def _encode_command_output(output):
             return ustr(output, encoding="utf-8", errors="backslashreplace")
 
         try:
-            process = subprocess.Popen(command, stdin=subprocess.PIPE, stdout=stdout, stderr=subprocess.PIPE, shell=False)
```
narrieta, line 118: This was not needed. I noticed it while reviewing the usages of Popen.

tests/ga/test_extension.py
```diff
@@ -2168,8 +2168,10 @@ def test_ext_sequence_no_should_be_set_from_within_extension(self, *args):
                 printenv | grep ConfigSequenceNumber
                 """
 
-        base_dir = os.path.join(conf.get_lib_dir(), 'OSTCExtensions.ExampleHandlerLinux-1.0.0', test_file_name)
-        self.create_script(test_file_name, test_file, base_dir)
```
narrieta, line 2172: AgentTestCase had drifted from its original implementation and some of the parameters were confusing. I did some simplification of the method's  signature.

azurelinuxagent/common/utils/cryptutil.py
```diff
@@ -81,46 +81,22 @@ def decrypt_p7m(self, p7m_file, trans_prv_file, trans_cert_file, pem_file):
...
-                    logger.error(msg)
-                
-            except Exception as exception:
-                msg = u"Exception on Command: [{0}]. exception={1}" \
-                    .format(cmd, exception)
-
-
-
+                shellutil.run_command(cmd, stdout=file_out, log_error=True)
+            except shellutil.CommandError:
```
pgombar, line 98: Previously, we were catching all exception types here. Would it be safer to keep it that way?

narrieta: The original code read  like this:

```
    def crt_to_ssh(self, input_file, output_file):
        shellutil.run("ssh-keygen -i -m PKCS8 -f {0} >> {1}".format(input_file,
                                                                    output_file))
```
The try/except was introduced in one of the iterations moving to run_command.

tests/utils/test_shell_util.py
```diff
@@ -127,92 +130,237 @@ def test_it_should_log_unexpected_errors_as_errors(self):
...
+        # The test pipe splits the output of "ls" in lines and count the number of lines.
+        #
+        # Sample output of "ls":
+        #     drwxrwxr-x 13 nam nam 4096 Nov 13 16:54 .
+        #
+        pipe = [["ls", "-ld", "."], ["sed", "s/ /\\n/g"], ["wc", "-l"]]
+
+        output = shellutil.run_pipe(pipe)
+
+        self.assertEqual(int(output), 9)
```
larohra, line 180: Will this always be 9? Like we might add some file later which will break this test right? 
It might be a better idea to create a temp directory and add some random pre-determined files there and ls in that to ensure any changes to this directory wont affect the test

larohra (commented):
1 minor comment about a test else LGTM

tests/utils/test_shell_util.py
narrieta: I am using a pre-determined file (directory, actually): "." so this wouldn't change

tests/tools.py
```diff
@@ -421,32 +421,20 @@ def _create_files(tmp_dir, prefix, suffix, count, with_sleep=0):
...
-        :return:
+        Creates an executable script with the given contents. If file ends with ".py", it creates a Python3 script,
+        otherwise it creates a bash script.
         """
-        if not file_path:
-            file_path = os.path.join(self.tmp_dir, file_name)
-
-        directory = os.path.dirname(file_path)
-        if not os.path.exists(directory):
-            os.mkdir(directory)
```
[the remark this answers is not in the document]
narrieta: I initially created this method as a private member of a test I was writing. The 'file_path' and create directory code were added later, when the method was used from other tests.

I don't think it should be be the responsibility of this method to create the directory (for example, somebody may have made a typo in a path and we may end up creating the directory in the wrong directory).

tests/utils/test_shell_util.py
larohra: Ohh got it! I had to run the command to see the output lol.

larohra (approved):

tests/tools.py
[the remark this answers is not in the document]
larohra: Fair enough. My point was basically its good to have it in a common space than duplicates of the same code in multiple places. 
Either way works for me though honestly

tests/utils/test_shell_util.py
narrieta: Oh, I see, I think know where you are coming from. I am using the -d option, so this does not list the contents of ".", but instead it lists the info for "." itself. This always produces only 1 line of output so this won't change.

pgombar: I had the same thought process and I had to run the command to figure out what it was doing :smile:

azurelinuxagent/common/utils/cryptutil.py
pgombar: I see, thanks for confirming.

pgombar (approved):

PR #2087, opened 2020-11-24
Parse ext runtime settings

larohra:
## Description
This is the first PR to handle the MultiConfig changes in the Guest Agent. This PR deals with parsing the multi-config related changes to the ExtensionConfig and ensures we're backward compatible.
Another major change in this PR is to report errors (if any) if the ExtensionConfig is malformed and fail fast rather than waiting for CRP to timeout eventually. This has been added because - 
- Since this feature is new in both CRP and Guest Agent. This will help us Identify any gaps in the config from CRP side too
- Improve customer experience by failing fast rather than waiting for 90 mins and timing out eventually

Issue # 

---

### PR information


azurelinuxagent/common/exception.py
```diff
@@ -83,21 +83,18 @@ def __init__(self, msg=None, inner=None, code=-1, exit_code=-1):
...
-        super(ExtensionUpdateError, self).__init__(msg, inner, code)
+    # When failed to update an extension
+    pass
 
 
 class ExtensionDownloadError(ExtensionError):
-    """
-    When failed to download and setup an extension
-    """
+    # When failed to download and setup an extension
```
larohra, line 91: Fixed 2 linter issues here - W0235 (useless-super-delegation) and W0107 (unnecessary-pass). 
The unnecessary-pass raises when we only have a pass statement with a docstring.

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -331,78 +338,211 @@ def __init__(self, xml_text):
...
-    def _parse_plugin(plugin):
-        ext_handler = ExtHandler()
-        ext_handler.name = getattrib(plugin, "name")
-        ext_handler.properties.version = getattrib(plugin, "version")
-        ext_handler.properties.state = getattrib(plugin, "state")
+    def _parse_plugin(ext_handler, plugin):
+
+        def _raise_config_error_if_none(attr_name, value):
+            if value is None:
+                raise ExtensionConfigError("{0} is None for ExtensionConfig, failing Extension.".format(attr_name))
```
larohra, line 345: These values are expected to be there for every Plugin. If they're not there then we might (or not even) report wrong status while CRP would be polling for something else. Failing here to ensure we fail fast

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -331,78 +338,211 @@ def __init__(self, xml_text):
...
         if plugin_settings is None:
             return
 
-        name = ext_handler.name
+        handler_name = ext_handler.name
         version = ext_handler.properties.version
 
-        to_lower = lambda str_to_change: str_to_change.lower() if str_to_change is not None else None
-        ext_handler_plugin_settings = [x for x in plugin_settings if to_lower(getattrib(x, "name")) == to_lower(name)]
+        def to_lower(str_to_change): return str_to_change.lower() if str_to_change is not None else None
```
larohra, line 367: PyCharm was complaining about this. Apparently PEP doesn't like using named-lambda functions, they are only supposed to be used as temporary functions (no-name)

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -331,78 +338,211 @@ def __init__(self, xml_text):
...
             return
 
         settings = [x for x in ext_handler_plugin_settings if getattrib(x, "version") == version]
         if len(settings) != len(ext_handler_plugin_settings):
-            msg = "ExtHandler PluginSettings Version Mismatch! Expected PluginSettings version: {0} for Handler: " \
-                  "{1} but found versions: ({2})".format(version, name, ', '.join(
-                set([getattrib(x, "version") for x in ext_handler_plugin_settings]))) 
+            msg = "ExtHandler PluginSettings Version Mismatch! Expected PluginSettings version: {0} for Handler: {1} but found versions: ({2})".format(
+                version, handler_name, ', '.join(set([getattrib(x, "version") for x in ext_handler_plugin_settings])))
             add_event(AGENT_NAME, op=WALAEventOperation.PluginSettingsVersionMismatch, message=msg, log_event=False,
```
larohra, line 376: Still keeping this event as there are monitors based on this operation

azurelinuxagent/common/protocol/restapi.py
```diff
@@ -85,13 +85,15 @@ def __init__(self,  # pylint: disable=R0913
...
-                 dependencyLevel=0):
+                 dependencyLevel=0,
+                 state="enabled"):
         self.name = name
         self.sequenceNumber = sequenceNumber  # pylint: disable=C0103
         self.publicSettings = publicSettings  # pylint: disable=C0103
         self.protectedSettings = protectedSettings  # pylint: disable=C0103
         self.certificateThumbprint = certificateThumbprint  # pylint: disable=C0103
         self.dependencyLevel = dependencyLevel  # pylint: disable=C0103
+        self.state = state
```
larohra, line 96: For MultiConfig, each settings also has a state (as compared to before where the state was only available in Plugin level)

azurelinuxagent/common/protocol/restapi.py
```diff
@@ -111,6 +113,15 @@ def __init__(self, name=None):
         self.name = name
         self.properties = ExtHandlerProperties()
         self.versionUris = DataContractList(ExtHandlerVersionUri)  # pylint: disable=C0103
+        self.__invalid_handler_reason = None
+
+    @property
+    def is_invalid_with_reason(self):
```
larohra, line 119: Introduced a new property to check if the ExtHandler is invalid for any reason. 
The benefit of adding this here is - 
- If extA is invalid but extB is valid, we would still execute both extA and extB and report faulty handler for only extA. 

(Note: Open to suggestions on the name of the property. Couldn't figure out a better name for what its trying to accomplish)

azurelinuxagent/ga/exthandlers.py
```diff
@@ -501,31 +506,37 @@ def handle_ext_handler(self, ext_handler, etag):
 
             return True
 
+        except ExtensionConfigError as error:
+            # Catch and report Invalid ExtensionConfig errors here to fail fast rather than timing out after 90 min
+            self.__handle_and_report_ext_handler_errors(ext_handler_i, error,
+                                                        report_op=WALAEventOperation.InvalidExtensionConfig,
+                                                        message="[Retryable Error] {0}".format(ustr(error)))
```
larohra, line 513: Generally config errors are retryable because most of the times CRP generates a new config which fixes the issue. 
This is purely based on my experiences with CRP, I unfortunately do not have any data to prove this. 

I mainly added the prefix `[Retryable Error]` here to inform the customer that a retry might fix this issue. But since I dont have any data to prove it, I dont mind removing this

larohra:
Green DCR (with known agent-slb setup error) : https://tuxgold.corp.microsoft.com/job/AzLinux/job/DungeonCrawler/job/DungeonCrawler.larohra2/612/

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -331,78 +338,211 @@ def __init__(self, xml_text):
...
+            msg = "Multiple plugin settings found for the same handler: {0} and version: {1} (Expected: 1; Available: {2})".format(
+                handler_name, version, len(settings))
+            raise ExtensionConfigError(msg)
+
+        plugin_settings_node = settings[0]
+        runtime_settings_nodes = findall(plugin_settings_node, "RuntimeSettings")
+        extension_runtime_settings_nodes = findall(plugin_settings_node, "ExtensionRuntimeSettings")
+
+        if runtime_settings_nodes and extension_runtime_settings_nodes:
+            # There can only be a single RuntimeSettings node or multiple ExtensionRuntimeSettings nodes per Plugin
```
pgombar, line 390: I know this comes from upstream, but the names are just so confusing.

What is the difference then between RuntimeSettings and ExtensionRuntimeSettings? The latter is per extension version, not per extension, so there can be multiple per extension?

larohra: So a quick recap on the new terminology - 

- Handler - The main Plugin/handler specified by the publishers (Eg: Microsoft.OSTC.CustomScript). 
- Extension - The runtime settings associated with a Handler (1.settings, 2.settings, etc). 
**Note:** Prior to multi-config, we used Handler and Extension interchangeably because we only supported 1 handler per extension. But moving on they would mean different things so keep an eye out for those.
- RuntimeSettings - This is the name of the node that we get from CRP which only supports **1 extension per handler** (this is the legacy behavior)
- ExtensionRuntimeSettings - This is the name of the new node that we get from CRP when the handler supports Multi-config (i.e. **multiple extensions per handler**)
- Plugin - This is the name of the node that we get from CRP that specifies the Handler. There should be 1 Plugin node per Handler

**Note:** Each plugin can have _either a single RuntimeSettings node or multiple ExtensionRuntimeSettings node, not both_.


I hope this brief overview helps! If you're still confused we can sync offline and discuss more about it :)

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -331,78 +338,211 @@ def __init__(self, xml_text):
...
+
+        if runtime_settings_nodes:
+            if len(runtime_settings_nodes) > 1:
+                msg = "Multiple RuntimeSettings found for the same handler: {0} and version: {1} (Expected: 1; Available: {2})".format(
+                    handler_name, version, len(runtime_settings_nodes))
+                raise ExtensionConfigError(msg)
+            # Only Runtime settings available, parse that
+            ExtensionsConfig.__parse_runtime_settings(plugin_settings_node, runtime_settings_nodes[0], handler_name,
+                                                      ext_handler)
+        else:
```
leisu001, line 403: May need to handle this case: both runtime_settings_nodes and extension_runtime_settings_nodes are null.

larohra: Good point. That is still a valid case right? Like no settings available but its still a new goalstate?

leisu001: yes, for winGA this case is possible.

leisu001: BGInfo for example

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -319,9 +320,15 @@ def __init__(self, xml_text):
         plugin_settings = findall(plugin_settings_list, "Plugin")
 
         for plugin in plugins:
-            ext_handler = ExtensionsConfig._parse_plugin(plugin)
+
+            ext_handler = ExtHandler()
```
narrieta, line 324: Let's consider failing the entire goal state if there is an error in any of the extensions.

My concern is that by trying to recover from errors in CRP, skip the incorrect extensions and execute the rest,  we may end up doing the wrong thing.

azurelinuxagent/common/protocol/goal_state.py
larohra: Ok makes sense, I'll add this case too. Thanks for the catch!

larohra: I missed it earlier, but we already handle this case in Line 370 :)
Though I still added the condition here too, just to make the code more explicit of what its doing rather than having some implicit dependencies

azurelinuxagent/common/exception.py
kevinclark19a: Just a note: You don't technically need a `pass` statement when you have a docstring on an empty class. I.e. you could add the docstring back and remove the `pass` statement and end up with the same functionality (though the docstring will of course be added as a `__doc__` attribute, which is actually probably preferred). I believe that's what pylint is complaining about when you have the pass statement; as long as the docstring is there, the `pass` is unneccessary.

See [this stack overflow answer](https://stackoverflow.com/a/49754261).

azurelinuxagent/common/protocol/goal_state.py
kevinclark19a: @narrieta could you expand on what you mean by this? Do you mean that the user might have extensions that make sense together, but not necessarily without all of them? Or is it more a concern about the goal state potentially being read incorrectly if some of the extensions are incorrect?

tests/ga/test_extension.py
```diff
@@ -548,6 +548,48 @@ def _assert_handler_status_and_manifest_download_count(protocol, test_data, mani
...
+
+            with patch('azurelinuxagent.common.event.add_event') as patch_add_event:
+                exthandlers_handler.run()
+                self._assert_handler_status(protocol.report_vm_status, "NotReady", 0, "1.0.0")
+
+                invalid_config_errors = [kw for _, kw in patch_add_event.call_args_list if
+                                         kw['op'] == WALAEventOperation.InvalidExtensionConfig]
+                self.assertEqual(1, len(invalid_config_errors), "Error not logged and reported to Kusto")
+
+    def test_it_should_process_valid_extensions_if_present(self, mock_get, mock_crypt_util, *args):
```
kevinclark19a, line 570: Is this behavior what @narrieta is talking about up above (in `.../goal_state.py`)? I.e. he wanted a single invalid extension to fail the entire goal state?

azurelinuxagent/common/exception.py
larohra: Ohh very cool! I didnt know about this! I'll do this instead than an empty pass if we do end up removing the super calls! Thanks!

azurelinuxagent/common/protocol/goal_state.py
larohra: Synced offline about it, and we decided to **only** fail the extension with the bad config rather than the whole GoalState. 

 >Do you mean that the user might have extensions that make sense together, but not necessarily without all of them?

The concern was about extensions that have external dependencies where they specifically set their ARM template having our sequential execution in mind. I can sync you up more about this on a call if you want :)

tests/ga/test_extension.py
larohra: No not really, this test ensures that even if some extension's config is bad, we still process other extensions. The ask was to fail all extensions instead. 

We're not going with that behavior though, we're keeping this behavior

azurelinuxagent/common/exception.py
kevinclark19a: Yep! I actually learned about this when I got the same `unnecessary-pass` error in a prior PR, lol.

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -331,78 +338,220 @@ def __init__(self, xml_text):
...
+                  }
+                }
+              ]
+            }
+        </RuntimeSettings>
+        </Plugin>
+        """
+        depends_on_nodes = findall(plugin_settings_node, "DependsOn")
+        if len(depends_on_nodes) > 1:
+            msg = "Extension Handler can only have a single dependsOn node for Single config extensions. Found: {0}".format(
```
narrieta, line 451: is this accurate? (we can sync offline)

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -331,78 +338,220 @@ def __init__(self, xml_text):
...
+            if not ExtensionsConfig.__is_not_empty(extension_name):
+                raise ExtensionConfigError("No Name not specified for DependsOn object in ExtensionRuntimeSettings for MultiConfig!")
+
+            dependency_level = ExtensionsConfig.__get_dependency_level_from_node(depends_on_node, extension_name)
+            dependency_levels[extension_name] = dependency_level
+
+        for extension_runtime_setting_node in extension_runtime_settings_nodes:
+            # Name and State will only be set for ExtensionRuntimeSettings for Multi-Config
+            extension_name = getattrib(extension_runtime_setting_node, "name")
+            if not ExtensionsConfig.__is_not_empty(extension_name):
```
narrieta, line 524: similar question to above: getattrib always return None if no match? why the additional check against "" and []?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -381,7 +381,8 @@ def _extension_processing_allowed(self):
 
     def handle_ext_handlers(self, etag=None):
         if not self.ext_handlers.extHandlers:
-            logger.verbose("No extension handler config found")
+            # This is a valid case, some extensions dont have any settings. Just log and exit processing
```
narrieta, line 384: not sure I get the comment. this was checking whether there were any extensions to execute... what settings does the comment refer to? not to the settings passed to the extension, does it?

narrieta (commented):
some minor comments in the parsing code

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -319,9 +320,15 @@ def __init__(self, xml_text):
         plugin_settings = findall(plugin_settings_list, "Plugin")
 
         for plugin in plugins:
-            ext_handler = ExtensionsConfig._parse_plugin(plugin)
+
```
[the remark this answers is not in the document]
larohra: Done

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -331,78 +338,220 @@ def __init__(self, xml_text):
...
-        ext_handler.properties.version = getattrib(plugin, "version")
-        ext_handler.properties.state = getattrib(plugin, "state")
+    def _parse_plugin(ext_handler, plugin):
+
+        def _raise_config_error_if_none(attr_name, value):
+            if value is None:
+                raise ExtensionConfigError("{0} is None for ExtensionConfig, failing Extension.".format(attr_name))
+            return value
+
+        ext_handler.name = _raise_config_error_if_none("Handler Name", getattrib(plugin, "name"))
```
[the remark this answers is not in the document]
larohra: Good idea, made the change with context

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -331,78 +338,220 @@ def __init__(self, xml_text):
...
+    def __is_not_empty(obj):
+        return obj not in (None, "", [])
 
     @staticmethod
     def _parse_plugin_settings(ext_handler, plugin_settings):
         if plugin_settings is None:
             return
 
-        name = ext_handler.name
+        handler_name = ext_handler.name
```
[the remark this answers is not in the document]
larohra: Done

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -331,78 +338,220 @@ def __init__(self, xml_text):
...
         version = ext_handler.properties.version
 
-        to_lower = lambda str_to_change: str_to_change.lower() if str_to_change is not None else None
-        ext_handler_plugin_settings = [x for x in plugin_settings if to_lower(getattrib(x, "name")) == to_lower(name)]
+        def to_lower(str_to_change): return str_to_change.lower() if str_to_change is not None else None
+
+        ext_handler_plugin_settings = [x for x in plugin_settings if to_lower(getattrib(x, "name")) == to_lower(handler_name)]
```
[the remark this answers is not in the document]
larohra: That is how it was before. This change was added when we got the bug where the version of `Plugin` and `PluginSettings.Plugin` was different. 

We basically use the check for `name` to confirm if there even are any settings for that particular plugin or not. I would prefer to leave this as is

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -331,78 +338,220 @@ def __init__(self, xml_text):
...
+        if len(settings) > 1:
+            msg = "Multiple plugin settings found for the same handler: {0} and version: {1} (Expected: 1; Available: {2})".format(
+                handler_name, version, len(settings))
+            raise ExtensionConfigError(msg)
+
+        plugin_settings_node = settings[0]
+        runtime_settings_nodes = findall(plugin_settings_node, "RuntimeSettings")
+        extension_runtime_settings_nodes = findall(plugin_settings_node, "ExtensionRuntimeSettings")
+
+        if ExtensionsConfig.__is_not_empty(runtime_settings_nodes) and ExtensionsConfig.__is_not_empty(
```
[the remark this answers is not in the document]
larohra: The `ExtensionsConfig.__is_not_empty` is just a standard utility function to check for empty values. instead of putting the empty checks everywhere (like multiple places with `not None` or `not in (None, "")`, I just created a simple utility and use that everywhere to make it easier to understand.

azurelinuxagent/common/protocol/goal_state.py
larohra: Added some data to the classes we create for these. Let me know if that is enough or more data/info is required

azurelinuxagent/common/protocol/goal_state.py
larohra: That is my understanding yes. I can confirm from CRP code

azurelinuxagent/common/protocol/goal_state.py
larohra: Answered above

azurelinuxagent/common/protocol/goal_state.py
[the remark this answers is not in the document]
narrieta: ok

azurelinuxagent/common/protocol/goal_state.py
narrieta: never mind - This is DependsOn, not DependsOnExtension

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -329,79 +322,318 @@ def __init__(self, xml_text):
...
+                msg = "Multiple RuntimeSettings found for the same handler: {0} and version: {1} (Expected: 1; Available: {2})".format(
+                    handler_name, version, len(runtime_settings_nodes))
+                raise ExtensionConfigError(msg)
+            # Only Runtime settings available, parse that
+            ExtensionsConfig.__parse_runtime_settings(plugin_settings_node, runtime_settings_nodes[0], handler_name,
+                                                      ext_handler)
+        elif ExtensionsConfig.__is_not_empty(extension_runtime_settings_nodes):
+            # Parse the ExtensionRuntime settings for the given extension
+            ExtensionsConfig.__parse_extension_runtime_settings(plugin_settings_node, extension_runtime_settings_nodes,
+                                                                ext_handler)
```
pgombar, line 494: Do we need to check for (and fail) when neither settings are set?

pgombar: Nevermind, I just noticed you return early in that case.

azurelinuxagent/ga/exthandlers.py
pgombar: Since the time you wrote this comment, did you perhaps get any clarity on this? Is it retryable?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -521,31 +526,37 @@ def handle_ext_handler(self, ext_handler, etag):
...
 
-    def handle_ext_handler_error(self, ext_handler_i, e, code=-1, report_telemetry_event=True):  # pylint: disable=C0103
-        msg = ustr(e)
+    @staticmethod
+    def handle_ext_handler_error(ext_handler_i, error, code=-1, report_telemetry_event=True):
+        msg = ustr(error)
         ext_handler_i.set_handler_status(message=msg, code=code)
 
         if report_telemetry_event:
             ext_handler_i.report_event(message=msg, is_success=False, log_event=True)
```
pgombar, line 554: At first I didn't see the difference between this method and `__handle_and_report_ext_handler_errors`. After all, they both set the handler status and report the error. The only difference I see is in the operation being reported to telemetry? Is there a way to combine these two methods since they're so similar?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -521,31 +526,37 @@ def handle_ext_handler(self, ext_handler, etag):
...
             self.handle_ext_handler_error(ext_handler_i, error, error.code)
         except Exception as error:
             self.handle_ext_handler_error(ext_handler_i, error)
 
         return False
 
-    def handle_ext_handler_error(self, ext_handler_i, e, code=-1, report_telemetry_event=True):  # pylint: disable=C0103
-        msg = ustr(e)
+    @staticmethod
+    def handle_ext_handler_error(ext_handler_i, error, code=-1, report_telemetry_event=True):
```
pgombar, line 549: At first I didn't see the difference between this method and __handle_and_report_ext_handler_errors. After all, they both set the handler status and report the error. The only difference I see is in the operation being reported to telemetry? Is there a way to combine these two methods since they're so similar?

pgombar (dismissed):
Left one suggestion to merge two similar methods, otherwise LGTM!

azurelinuxagent/ga/exthandlers.py
larohra: Unfortunately no, just to be on the safer side, I'm going to modify the message to say something like "Ran into some error, please retry the operation with modified settings to try to mitigate it" or something similar

azurelinuxagent/common/protocol/goal_state.py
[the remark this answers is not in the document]
larohra: Added explicit checks instead

azurelinuxagent/ga/exthandlers.py
larohra: Sorry wrong comment, got confused with something here, removed it.

azurelinuxagent/ga/exthandlers.py
larohra: I thought of the same thing and tried doing it too, unfortunately a lot of tests are mocking internals of those functions and its just not worth changing those :)

narrieta (approved):

pgombar (approved):

PR #2097, opened 2020-12-08
Adding Github Actions Merge Validation

kevinclark19a:
## Description

Added a github actions workflow to enable unit test merge validation.

This PR is part of a move away from Travis unit test merge validation and to the recommended CD, CI pipeline.

---

### PR information


.github/workflows/ci_pr.yml
```diff
@@ -0,0 +1,140 @@
...
+        sudo env "PATH=$PATH" python -m pip install --upgrade pip
+        sudo env "PATH=$PATH" pip install -r requirements.txt
+        sudo env "PATH=$PATH" pip install -r test-requirements.txt
+
+    - name: Run pylint
+      run: |
+        pylint $PYLINTOPTS --jobs=0 $PYLINTFILES
+
+    - name: Test with nosetests
+      if: always()
```
larohra, line 110: What does this check mean? If it runs always then why have this check?

.github/workflows/ci_pr.yml
```diff
@@ -0,0 +1,140 @@
...
+      run: |
+        ./ci/nosetests.sh || EXIT_CODE=$?
+        exit $EXIT_CODE
+
+    - name: Run Coverage
```
larohra, line 115: How to read and parse this data? Like from the `checks` page it wasnt very clear how to read the data. 
Is there a way we can add it to the PR page (similar to how we had it for travis)?

.github/workflows/ci_pr.yml
Aymendje: Basically even if the previous task failed, this one should always run.

.github/workflows/ci_pr.yml
Aymendje: It generates a codecov report.
Also will post the test results at the end of the checks page.

.github/workflows/ci_pr.yml
```diff
@@ -0,0 +1,140 @@
...
+          echo Failed coverage check. Found $COVERAGE_PERCENT%, expected minimum of $MIN_COVERAGE_GOAL%;
+          EXIT_CODE=1;
+        fi
+        exit $EXIT_CODE
+
+    - name: Upload Coverage
+      if: ${{ github.event_name == 'pull_request' }} && matrix.python-version == 3.8
+      uses: codecov/codecov-action@v1
+      with:
+        file: ./coverage.xml
```
Aymendje, line 140: It is really hard to see which test failed.
Can we add one of the commands to post the result output ?
```
    - name: Publish Unit Test Results
      uses: EnricoMi/publish-unit-test-result-action@v1.6
      if: matrix.python-version == 3.8 && always()
      with:
        github_token: ${{ secrets.GITHUB_TOKEN }}
        files: nosetests.*.xml
```

.github/workflows/ci_pr.yml
larohra: Ahh interesting. Isint that what the `fail-fast: false` is there for? Also then wouldnt we need this check for each task?

.github/workflows/ci_pr.yml
larohra: I saw a link and clicked on it but it showed nothing. Can you post the link just incase I picked the wrong one?

>but we should add another task for publish unit tests results

That would be helpful yes :)

.github/workflows/ci_pr.yml
larohra: >if: matrix.python-version == 3.8 && always()

Why only post for 3.8? Or will it post for all runs?

Additionally, publishing data should be a separate job (both UTs and CodeCov)

.github/workflows/ci_pr.yml
larohra: This is the link I tried - 

> -> View reports at https://codecov.io/github/Azure/WALinuxAgent/commit/39e5f5e2e3ccd2294044a705624f245ec6240204

.github/workflows/ci_pr.yml
kevinclark19a: `always()` here allows this step to run regardless of failure in prior steps. This allows our pylint to run even when the unit tests fail.

Do you think an inline comment here explaining what `always()` does would be useful? Or do you think it would be better just to link the docs to github actions at, e.g., the top of the yml file?
@pgombar, @narrieta, also interested in your viewpoints.

Aymendje: nop, the fail-fast is to not fail all element in the matrix if one element fails
ex : if python 2.6 fails, and fail-fast is enabled, all other python will get automatically cancelled

But to be honest, @kevinclark19a  why do we have it here ?
If the setup part fails, the Unit-Test should not run, since the whole test will fail anyways

Usually always() is used for reporting Unit-test or code coverage failure, so even if they fail, we can still report the results

.github/workflows/ci_pr.yml
Aymendje: Just scroll down a bit, you will see
image

Then click on details
ex : https://github.com/Azure/WALinuxAgent/pull/2097/checks?check_run_id=1518915394

larohra: But these are details for the whole run. I meant specifically for codecov (the link I pasted above is from the details of Pyton 3.8 Unit Tests)

.github/workflows/ci_pr.yml
kevinclark19a: > Additionally, publishing data should be a separate job (both UTs and CodeCov)

It would be much harder to publish data in a separate job because different jobs don't share a file system; is there a reason why we can't just have a separate step of the same job? That way, we could run scripts on the same VM, and access any files generated by prior steps (e.g., the `nosetest` run)

larohra: Ohh nevermind then, just a suggestion mainly for clarity nothing else. You can keep it as is :)

.github/workflows/ci_pr.yml
kevinclark19a: The idea was that it is still useful to check unit test correctness, even if the linter found some errors, or vice versa. It's mainly a debuggability thing, in my opinion.

.github/workflows/ci_pr.yml
Aymendje: there was no code change here, so codecov didnt show a diff I think

.github/workflows/ci_pr.yml
kevinclark19a: Yeah, a separate job would certainly be more readable! I don't think there's a good way of talking across jobs, though, short of, e.g., uploading to blob storage as an intermediary, which seems excessive.

Aymendje: These kind of job can only run once, so we need to upload the logs separately, then download them back into that task
it is doable, but usually if UT fails, they should fail everywhere so this is fine for now.
If we see issues were UT fails in specific version of python, we can try to implement multiple reports.

.github/workflows/ci_pr.yml
larohra: Hmm interesting, @kevinclark19a can you give this a test run? Like change some random code and ensure the codecov is being hit? Else might as well remove this check (because we dont have it now either)

.github/workflows/ci_pr.yml
larohra: >If we see issues were UT fails in specific version of python, we can try to implement multiple reports.

Unfortunately that is exactly what we see lol (like most of the times tests pass in py2 and fail in py3 and vice versa). Might as well publish tests for all runs? If it becomes too verbose or dirty maybe we can remove it then?

.github/workflows/ci_pr.yml
larohra: Hmm makes sense. Then should we also add `always()` for the pylint too?

.github/workflows/ci_pr.yml
Aymendje: If you connect with your github account, and go in files, you can see the test coverage
By default, codecov opens on the diff page.

.github/workflows/ci_pr.yml
Aymendje: The linter would not run in this case.
The linter have its own task, and the task is not marked as `always()`
My point here is if the setup task fails, (the one installing python / pip checking the version), we cannot know what version of python is currently running the test.

.github/workflows/ci_pr.yml
Aymendje: Its not too hard
@kevinclark19a  you can follow documentation here : https://github.com/EnricoMi/publish-unit-test-result-action#use-with-matrix-strategy
Basically after every test in the matrix, you call `actions/upload-artifact@v2` with file `nosetests.*.xml` and then you create a new job with task `actions/download-artifact@v2` and `EnricoMi/publish-unit-test-result-action@v1.6` to publish

.github/workflows/ci_pr.yml
larohra: nvm the codecov report showed up finally (it probably takes some time to reflect in the portal)
 
image

Aymendje: I think it was waiting for Travis to run (it appeared at the same time Travis finished) that's weird

.github/workflows/ci_pr.yml
kevinclark19a: If setup fails, both the linter and the unit tests will fail. Nosetests are after pylint, so the `always()` on nosetests is only to allow nosetests to run even if pylint has run and failed.

I think that it would be semantically better to only run nosetests if setup has succeeded but pylint has failed (or, obviously, if everything prior has succeeded), but I don't know of any simple mechanism to allow that. The `always()` on nosetests seemed like the best option, then: if setup fails, we would have pylint skipped and nosetests fail (and the failure should happen pretty quickly, as, e.g. `python` wouldn't be in the env), and that doesn't change the results of the workflow; if, however, setup succeeds but pylint fails, nosetests will still be able to run and provide info on unit test correctness. 

In that second case, even if the workflow will fail (because pylint failed), it's nice to have the info on what tests passed, if any.

Essentially, the later of the two (pylint or nosetests) needs an `always()`, but not the earlier.

kevinclark19a: Additionally, even if the unit tests _don't_ fail in the first case, the only thing we lose is ~2 minutes of compute. The workflow will still fail because the setup already failed.

Aymendje: You are correct
My bad, not sure what I was looking at, but I thought the pylint was under nosetest...
(also, maybe use `success() || failure()`, since `always()` run even when job gets  cancelled)

.github/workflows/ci_pr.yml
```diff
@@ -0,0 +1,140 @@
...
+    - uses: actions/checkout@v2
+
+    - name: Install dependencies
+      run: |
+        sudo env "PATH=$PATH" pip install -r requirements.txt
+        sudo env "PATH=$PATH" pip install -r test-requirements.txt
+    
+    - name: Test with nosetests
+      run: |
+        ./ci/nosetests.sh || EXIT_CODE=$?
```
larohra, line 44: You havent defined exit code before, and if `./ci/nosetests.sh `  succeeds then `EXIT_CODE=$?` this command wont be run. 

For when you finally exit `exit $EXIT_CODE`, would this automatically set to `0` even if nosetest command succeeds?

kevinclark19a: My understanding is that `$EXIT_CODE` is replaced with the empty string, and that `exit` with no arguments does in fact exit with a code of 0. However, I think it would be better to explicitly set `EXIT_CODE` before hand, if only for the sake of readability.

.github/workflows/ci_pr.yml
larohra: >Do you think an inline comment here explaining what always() does would be useful?

Yeah for sure, more doc is good in general :)

>then: if setup fails, we would have pylint skipped and nosetests fail (and the failure should happen pretty quickly, as, e.g. python wouldn't be in the env)

Are you sure about it? Like wont the container already have some python pre-installed in it? Like my concern here is that if python-setup fails and we use `always()` then the tests might run with the pre-installed python version which is something that we dont want either.
Maybe try doing something like @Aymendje mentioned above, a combination of `success() || failure()` or something?

.github/workflows/ci_pr.yml
larohra: hmm weird, we're gonna remove travis anyways so even if this is a thing, its a fleeting pain lol

.github/workflows/ci_pr.yml
larohra: ping on this (@kevinclark19a to enable this and see how it looks in the UI page)

ci/nosetests.sh
```diff
@@ -0,0 +1,21 @@
...
+
+set -u
+
+EXIT_CODE=0
+
+echo "========================================="
+echo "nosetests -a '!requires_sudo' output"
+echo "========================================="
+nosetests -a '!requires_sudo' tests $(echo "${NOSEOPTS/__uuid__/$(uuidgen)}") || EXIT_CODE=$(($EXIT_CODE || $?))
+echo EXIT_CODE nosetests = $EXIT_CODE
```
larohra, line 11: NIT: `echo EXIT_CODE **no_sudo** nosetests = $EXIT_CODE`

ci/nosetests.sh
```diff
@@ -0,0 +1,21 @@
...
+sudo mv .coverage coverage.$(uuidgen).no_requires_sudo.data
+
+echo "========================================="
+echo "nosetests -a 'requires_sudo' output"
+echo "========================================="
+sudo env "PATH=$PATH" nosetests -a 'requires_sudo' tests $(echo "${NOSEOPTS/__uuid__/$(uuidgen)}") || EXIT_CODE=$(($EXIT_CODE || $?))
+echo EXIT_CODE sudo nosetest = $EXIT_CODE
```
larohra, line 18: NIT: `echo EXIT_CODE **with_sudo** nosetests = $EXIT_CODE`

.github/workflows/ci_pr.yml
```diff
@@ -0,0 +1,140 @@
...
+        sudo env "PATH=$PATH" coverage xml
+        COVERAGE_PERCENT=$(coverage report | tail -n 1 | rev | cut -d ' ' -f 1 | cut -c 2- | rev)
+        EXIT_CODE=1
+        echo Codecoverage : $COVERAGE_PERCENT%
+        if [ "$COVERAGE_PERCENT" -eq "$COVERAGE_PERCENT" ] && [ "$COVERAGE_PERCENT" -gt "$MIN_COVERAGE_GOAL" ]; then 
+          echo Pass coverage check.
+          EXIT_CODE=0;
+        else 
+          echo Failed coverage check. Found $COVERAGE_PERCENT%, expected minimum of $MIN_COVERAGE_GOAL%;
+          EXIT_CODE=1;
```
larohra, line 132: So far we've not had `codecov` as a required feature, with this though we would enforce min 60% right? I wonder if we should make this an enforcing thing or not. Maybe just publish the coverage data but dont fail if coverage < 60%?

larohra (commented):
Added some small comments

.github/workflows/ci_pr.yml
kevinclark19a: > Like my concern here is that if python-setup fails and we use always() then the tests might run with the pre-installed python version which is something that we dont want either.

From my runs debugging this workflow, the failure is typically `nosetest: command not found`. But even if the wrong version of python runs the unit tests, is there any sort of impact? The workflow will still fail, because we're presupposing that the setup step failed, and we'll be able to clearly see that the right python wasn't installed. As mentioned above, my understanding is that we only lose out of ~2 min compute...

> Maybe try doing something like @Aymendje mentioned above, a combination of success() || failure() or something?

I think this would have the same issue that you're mentioning, as a failure in setup would cause `failure()` to return true, and the unit tests would run. I mean, I definitely think it's preferable so that unit tests don't get run when a job is cancelled, like @Aymendje mentioned, but I don't think it's applicable to this particular issue you're raising.

.github/workflows/ci_pr.yml
Aymendje: Yes, or simply
`./ci/nosetests.sh; EXIT_CODE=$?`

.github/workflows/ci_pr.yml
Aymendje: The previous travis script had code coverage silently enforced to 60%
If the coverage (I think total coverage, not PR coverage...) dropped under 60%, the PR would have failed.

.github/workflows/ci_pr.yml
kevinclark19a: Quick question: could we use this method for code cov as well? I think that would also benefit from being a separate job. Or, at the very least, a part of the new, "Publish Unit Test Results" job, instead of the "Run Unit Tests" job.

.github/workflows/ci_pr.yml
pgombar: I think we should explicitly set `EXIT_CODE` to nosetest's exit code instead of relying on exit with no arguments defaulting to 0.

.github/workflows/ci_pr.yml
```diff
@@ -0,0 +1,140 @@
...
+    
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+
+          - python-version: 2.7
+            PYLINTOPTS: "--rcfile=ci/2.7.pylintrc"
+            NOSEOPTS: "--verbose --with-timer"
+            PYLINTFILES: "azurelinuxagent setup.py makepkg.py tests"
```
pgombar, line 57: nit: Is it possible to extract these flags into variables so we don't repeat their values for each python job?

ci/nosetests.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -u
+
+EXIT_CODE=0
+
+echo "========================================="
+echo "nosetests -a '!requires_sudo' output"
+echo "========================================="
+nosetests -a '!requires_sudo' tests $(echo "${NOSEOPTS/__uuid__/$(uuidgen)}") || EXIT_CODE=$(($EXIT_CODE || $?))
```
pgombar, line 10: I'm comparing against the nosetests script we have today... what is this addition for and why? `$(echo "${NOSEOPTS/__uuid__/$(uuidgen)}")`

pgombar: specifially the mention of "uuid" is throwing me off. if you could explain where this comes from that would be great, thanks!

.github/workflows/ci_pr.yml
pgombar: My thoughts:

> Or do you think it would be better just to link the docs to github actions at, e.g., the top of the yml file?

I think we should have this regardless of having comments elsewhere inline in the file.

Regarding failing or continuing, I think we should treat **tests** separate from **linter**. E.g., tests can fail, but I still want the linter to run, or vice-versa.

However, if the setup step fails, then I agree it doesn't make sense to proceed with tests or linter.

.github/workflows/ci_pr.yml
pgombar: The PR wouldn't have failed. The coverage step would fail, but we didn't treat it as a required check for merging.

I think we should have coverage as an optional check, at least in the beginning.

.github/workflows/ci_pr.yml
pgombar: For v1 of GHA, I don't think we should be over-complicating with coverage. Let's have it as it is today, only for py3.8 (and not required for merging) and get GHA working first.

pgombar (commented):
Left my comments. Let's focus on having the same capabilities as our existing CI. We can easily add features to it after we get GHA working.

.github/workflows/ci_pr.yml
kevinclark19a: What precisely are you referring to? Like using a default value that gets overridden, e.g. `--rcfile=ci/3.6.pylintrc` as a default that can be overridden for python2.7 and 3.4?

ci/nosetests.sh
Aymendje: There is a bug with the current script you have today, where the output of `requires_sudo` gets overwritten by `!requires_sudo`
So we need to give it a unique name instead.

NOSEOPTS contains this part `--xunit-file=nosetests.__uuid__.xml` and we perform a regex to replace `__uuid__` with the output of `$(uuidgen)`.

uuidgen is a bin that comes with debian that generates a unique id every time

.github/workflows/ci_pr.yml
kevinclark19a: If we make code cov a separate job, that would be easy to configure via the repo's settings page.

Aymendje: @pgombar It woudnt have failed because there was a bug in the previous script :
 NOSEOPTS were not getting passed to any run, including python 3.8

If the argument were passed correctly, python3.8 would run coverage validation during unit test, and would fail the unit test part if its sub 60%

pgombar: No, I'm talking about the time when we were actually running codecov, before the bug. The codecov check fails, but the PR doesn't fail because we set the codecov hook as optional in the settings for develop branch.

.github/workflows/ci_pr.yml
Aymendje: Also, we should add `fail_ci_if_error: false` here to not fail the check if codecov returns < 60%

.github/workflows/ci_pr.yml
```diff
@@ -0,0 +1,140 @@
...
+    strategy:
+      fail-fast: false
+    env:
+      NOSEOPTS: "--verbose"
+
+    steps:
+
+    - name: Install Python 2.6
+      run: |
+        sudo add-apt-repository ppa:deadsnakes/ppa -y
```
narrieta, line 24: can we do 'set -euo pipefail' on those scripts (is this bash?)

narrieta, line 24: It is not clear to me who publishes this repo. There are multiple references in blog posts, so probably it is safe enough for unit tests, but we should avoid it if possible (unless we know the publisher and its reputation). couldn't we just create a python venv, upload it to storage and install it here?

larohra: If going that route it might be better to try to get and publish a docker container image with py2.6 preinstalled. Creating a python 2.6 venv might or might not work in the test container, totally depends on the availability of the same set of libraries as the VM/container where the venv was created

narrieta: Checking whether we can create a venv should not be too difficult. This is running on ubuntu 16; if it is not possible to compile 2.6 from source there, we could try to create a VM in the closest distro to ubuntu 16 that we can find that uses 2.6 and create the venv there.

If this is too difficult we can do it on a later PR.

larohra: Removed ETP

kevinclark19a: I think `runs-on: ubuntu-16.04` is actually using bash. It's hard to find concrete info on it online, but `command not found` errors for 18.04 (during testing of this PR) show that `bash` is being used for `ubuntu-18.04` run commands... And I think plain `sh` is being used for 14.04. I can't remember, however, what 16.04 is using, but I think it was bash as well.

I can obviously just test it via another push. I am hesitant to use anything other than purely POSIX-compliant commands, however, as I can't find any documentation guaranteeing that, e.g., `bash` is used.

narrieta: BTW, Travis uses this venv on ubuntu 14, we can try to check whether there are any licensing that would prevent us from using it. https://storage.googleapis.com/travis-ci-language-archives/python/binaries/ubuntu/14.04/x86_64/python-2.6.tar.bz2

kevinclark19a: Definitely. I think Travis actually does pretty much this for their python2.6 environment, and it's obviously (at least, imo) a much cleaner solution. Do we think this is a high enough priority to stall the PR until we have the venv built and uploaded? I can't image it would take _too_ long, but the current code is already working, and upgrading to a blobbed venv should be a simple matter of tweaking these couple of shell lines once the venev is uploaded.

narrieta: yes, we can do it on a subsequent pr

.github/workflows/ci_pr.yml
kevinclark19a: Marking this as resolved: replaced the `always()` with `steps.install-dependencies.outcome == 'success'`, which should achieve the wanted behavior of only ignoring linter failures when running the tests, and not setup failures.

.github/workflows/ci_pr.yml
kevinclark19a: I can sync offline with anyone who wants more info on this, but it turns out that there is some unforeseen, additional complexity introduced via the methods mentioned above. For now, I'm going to stick to implementing only the behavior that we had prior, on Travis CI. However, I've created a branch for exploring these methods. Hopefully, this is something we can return to on a later revision of the pipeline!

.github/workflows/ci_pr.yml
kevinclark19a: Fixed! Resolving this thread now.

.github/workflows/ci_pr.yml
kevinclark19a: I've kept the codecov upload, but removed the check for >60% coverage. That way, we can get the coverage info from the link, but it doesn't fail the workflow, and we can still merge a PR regardless.

This is going back to the stated purpose of this PR, which was to replicate the functionality we had with Travis CI in Github Actions.

.github/workflows/ci_pr.yml
pgombar: That's what I was thinking, but I see we do the exact same thing in Travis, so you can ignore for now

.github/workflows/ci_pr.yml
kevinclark19a: Follow up: ubuntu-16.04 does in fact use bash for any `run` actions.

pgombar: We should mirror what we do for Travis at least:
`set -u`

ci/nosetests.sh
```diff
@@ -0,0 +1,21 @@
...
+set -u
+
+EXIT_CODE=0
+
+echo "========================================="
+echo "nosetests -a '!requires_sudo' output"
+echo "========================================="
+nosetests -a '!requires_sudo' tests $NOSEOPTS || EXIT_CODE=$(($EXIT_CODE || $?))
+echo EXIT_CODE no_sudo nosetests = $EXIT_CODE
+sudo mv .coverage coverage.$(uuidgen).no_sudo.data
```
pgombar, line 12: Do you still need these uuids?

ci/nosetests.sh
```diff
@@ -0,0 +1,21 @@
...
+
+echo "========================================="
+echo "nosetests -a 'requires_sudo' output"
+echo "========================================="
+sudo env "PATH=$PATH" nosetests -a 'requires_sudo' tests $NOSEOPTS || EXIT_CODE=$(($EXIT_CODE || $?))
+echo EXIT_CODE with_sudo nosetests = $EXIT_CODE
+sudo mv .coverage coverage.$(uuidgen).with_sudo.data
```
pgombar, line 19: Do you still need these uuids?

ci/nosetests.sh
kevinclark19a: Yes, to combine the .coverage files for codecov.

.github/workflows/ci_pr.yml
Aymendje: We can also build python and run it
https://github.com/pyenv/pyenv does this very well, and there is no licensing about it. Also its clean and pyenv is well known and supported by the community.
(I tried it and it worked on Ubuntu 14-04, and took around 5 extra minutes to build)

pgombar (dismissed):
LGTM, can't wait to have proper CI again :smile:

ci/nosetests.sh
pgombar: I see, thanks

.github/workflows/ci_pr.yml
larohra: >https://github.com/pyenv/pyenv does this very well

Correct me if I'm wrong, but this doesnt build for you right? Just makes it easier to manager already installed python versions on the box?

larohra: But on that note, would this `run` step fail if any command in this shell fail without `set -euo pipefail`? If not, then we should add it

.github/workflows/ci_pr.yml
```diff
@@ -0,0 +1,122 @@
...
+        echo "Installed python2.6"
+        python -V
+        curl https://bootstrap.pypa.io/2.6/get-pip.py -o get-pip.py
+        sudo env "PATH=$PATH" python get-pip.py
+
+    - uses: actions/checkout@v2
+
+    - name: Install dependencies
+      run: |
+        sudo env "PATH=$PATH" pip install -r requirements.txt
```
larohra, line 39: NIT: Add `set -euo pipefail` to all steps where using `run` to run shell scripts

.github/workflows/ci_pr.yml
```diff
@@ -0,0 +1,122 @@
...
+        sudo env "PATH=$PATH" python -m pip install --upgrade pip
+        sudo env "PATH=$PATH" pip install -r requirements.txt
+        sudo env "PATH=$PATH" pip install -r test-requirements.txt
+
+    - name: Run pylint
+      run: |
+        pylint $PYLINTOPTS --jobs=0 $PYLINTFILES
+
+    - name: Test with nosetests
+      if: steps.install-dependencies.outcome == 'success'
```
larohra, line 104: Thanks for adding this! Just confirming, but did you verify this works? Like forcing pylint to fail and ensuring the nosetests are still running?

.github/workflows/ci_pr.yml
```diff
@@ -0,0 +1,122 @@
...
+      if: matrix.python-version == 3.9
+      run: |
+        echo looking for coverage files :
+        ls -alh | grep -i coverage
+        sudo env "PATH=$PATH" coverage combine coverage.*.data
+        sudo env "PATH=$PATH" coverage xml
+        sudo env "PATH=$PATH" coverage report
+
+    - name: Upload Coverage
+      if: ${{ github.event_name == 'pull_request' }} && matrix.python-version == 3.9
```
larohra, line 119: qq, why do we only upload coverage results on PRs? Why not just let upload it everytime the GHA runs?

kevinclark19a: Ah ok. I had thought this was the behavior with travis, but I just checked and it seems we did in fact upload on every run. Is that something we need in this PR?

.github/workflows/ci_pr.yml
kevinclark19a: I'll run an edited branch and upload a screenshot of the results. One second!

.github/workflows/ci_pr.yml
larohra: Yeah why not, should be a small change from your side :)

.github/workflows/ci_pr.yml
kevinclark19a: As is, it doesn't look like nosetests will still run if pylint fails.

Editing the conditional to `success() || (failure() && steps.install-dependencies.output == 'success'` results in the following, though;

image
image

larohra: Awesome! Thanks for checking and fixing :)

.github/workflows/ci_pr.yml
kevinclark19a: any failed command in a step do in fact fail that entire step!

pgombar (approved):

larohra (approved):

.github/workflows/ci_pr.yml
Aymendje: It builds for you
`pyenv install 2.6.7` will fetch the source for 2.6.7, patch it, and build it
`pyenv global 2.6.7` to set it as system version

PR #2098, opened 2020-12-08
Add callbacks to run_command

narrieta:
Added callbacks to run/run_get_output/run_command/run_pipe for commands started and completed.

These will be used to monitor the commands started by the agent.

### DIFF
--- a/azurelinuxagent/common/utils/shellutil.py
+++ b/azurelinuxagent/common/utils/shellutil.py
@@ -97,26 +97,31 @@ def run_get_output(cmd, chk_err=True, log_cmd=True, expected_errors=None):
         expected_errors = []
     if log_cmd:
         logger.verbose(u"Command: [{0}]", cmd)
-    try:
-        output = subprocess.check_output(cmd, stderr=subprocess.STDOUT, shell=True)
-        output = __encode_command_output(output)
-    except subprocess.CalledProcessError as e:  # pylint: disable=C0103
-        output = __encode_command_output(e.output)
 
-        if chk_err:
-            msg = u"Command: [{0}], " \
-                  u"return code: [{1}], " \
-                  u"result: [{2}]".format(cmd, e.returncode, output)
-            if e.returncode in expected_errors:
-                logger.info(msg)
-            else:
-                logger.error(msg)
-        return e.returncode, output
-    except Exception as e:  # pylint: disable=C0103
+    try:
+        process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True)
+        _invoke_on_command_started_callback(process.pid)
+
+        output, _ = process.communicate()
+        _invoke_on_command_completed_callback(process.pid)
+
+        output = __encode_command_output(output)
+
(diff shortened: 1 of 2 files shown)


tests/utils/test_shell_util.py
```diff
@@ -360,6 +364,134 @@ def test_run_pipe_should_return_a_bytes_object_when_encode_output_is_false(self)
...
+        try:
+            child_processes = []
+            parent_processes = []
+
+            started_commands = []
+            completed_commands = []
+
+            # W0108: Lambda may not be necessary (unnecessary-lambda) - The use of lambda is appropriate
+            shellutil.set_on_command_started_callback(lambda pid: started_commands.append(pid))  # pylint:disable=W0108
+            shellutil.set_on_command_completed_callback(lambda pid: completed_commands.append(pid))  # pylint:disable=W0108
```
kevinclark19a, line 392: My understanding of `W0108` is that, instead of `lambda pid: started_commands.append(pid)`, you can just do `started_commands.append`.

See, e.g., [this StackOverflow post](https://stackoverflow.com/a/56446811).

narrieta: Yes, either way is correct functionally. For me, it is clearer to say in the code that I am passing a callable that takes a pid and appends it to the array, it is more explicit that just passing append.

kevinclark19a: I personally agree about the readability. I think it would be beneficial to modify the comment to point out that benefit, but I don't think it's really that big of a deal. Your discretion!

kevinclark19a (approved):
LGTM!

tests/utils/test_shell_util.py
narrieta: I would advocate for disabling all the style rules in Pylint; I find that most of the time they just get in the way.

larohra: +1 to that, if not disable all then atleast disable most (like too many branches, not enough instances, etc etc)

larohra (approved):

azurelinuxagent/common/utils/shellutil.py
```diff
@@ -325,4 +337,50 @@ def quote(word_list):
...
+#
+# Only 1 callback is maintained for each of those notifications. Setting a callback overrides the previous value.
+# Set the callback to None to stop the notifications
+#
+# C0103: Constant name "foo" doesn't conform to UPPER_CASE naming style (invalid-name) -- Disabled: These are not constants
+_on_command_started_callback = None  # pylint:disable=C0103
+_on_command_completed_callback = None  # pylint:disable=C0103
+
+
+def set_on_command_started_callback(callback):
```
pgombar, line 357: This is not used anywhere? Maybe I'm missing something, is this only the first PR of multiple that involves callback changes?

If I'm seeing this correctly, `_invoke_on_command_started_callback` does nothing since the `_on_command_started_callback` will always be None, correct?

narrieta: It will be used to check the processes in the agent cgroup

PR #2106, opened 2020-12-11
Fixing two unit tests incompatible with Github Actions

kevinclark19a:
## Description

Two dhcp unit tests relied on the fact that Travis CI (and, indeed, most dev boxes) did not run on Azure Cloud VMs.

Github Actions, however, runs workflows on azure VMs, so changes were needed so that the failing unit tests were more or less platform agnostic.

Essentially, both `...test_dhcp.TestDHCP.test_wireserver_rote_not_exists` and `...test_dhcp.testDHCP.test_dhcp_skip_cache` assumed that the current system's route table (the contents of the `/proc/net/route` file on most linux distros, including the versions of ubuntu used for our CD/CI) did not contain wireserver's information. That information is, of course, present on systems virtualized on azure cloud, so that the services that need to talk to wireserver can do so.

The fix in this PR is to mock the contents of that file to an environment independent state.

---

### PR information


tests/common/dhcp/test_dhcp.py
```diff
@@ -65,7 +65,18 @@ def test_wireserver_route_not_exists(self):
         self.assertTrue(dhcp_handler.gateway is None)
 
         # execute
-        self.assertFalse(dhcp_handler.wireserver_route_exists)
+        routing_table = "\
```
larohra, line 68: NIT: Since this is being used in multiple places, why not make it a small helper function for this class? (Just trying to avoid code-duplication as much as possible)

larohra (dismissed):
Small improvement comment, but everything else LGTM

tests/common/dhcp/test_dhcp.py
kevinclark19a: Hmm, that makes sense. Do you think it would be easiest to make two member variables? One with wireserver's info, and one without?

larohra: You can do that, or maybe make the string without the wireserver info a template and add the line with the wireserver info if a flag is set? Totally your call, whatever you feel more comfortable with and you think is the easiest way to do :)

tests/common/dhcp/test_dhcp.py
```diff
@@ -23,6 +23,11 @@
 
 class TestDHCP(AgentTestCase):
 
+    DEFAULT_ROUTING_TABLE = "\
+            Iface	Destination	Gateway 	Flags	RefCnt	Use	Metric	Mask		MTU	Window	IRTT \n\
```
pgombar, line 27: Why do you need `\n`? It wasn't there when the string was inline.

pgombar: Nevermind, just saw it :)

tests/common/dhcp/test_dhcp.py
```diff
@@ -37,19 +42,12 @@ def test_wireserver_route_exists(self):
...
-        routing_table = "\
-            Iface	Destination	Gateway 	Flags	RefCnt	Use	Metric	" \
-                        "Mask		MTU	Window	IRTT \n\
-            eth0	00000000	10813FA8	0003	0	    0	5	" \
-                        "00000000	0	0	0   \n\
-            eth0	00345B0A	00000000	0001	0	    0	5	" \
-                        "00000000	0	0	0   \n\
-            lo	    00000000	01345B0A	0003	0	    0	1	" \
-                        "00FCFFFF	0	0	0   \n"
+        routing_table = TestDHCP.DEFAULT_ROUTING_TABLE + \
```
pgombar, line 45: Could you rename so it's clear what this is for, something like `routing_table_with_wireserver_route`. Thanks!

pgombar (commented):
Minor suggestion, else LGTM!

tests/common/dhcp/test_dhcp.py
kevinclark19a: sure, that makes sense.

pgombar (approved):

pgombar (approved):

narrieta (approved):

PR #2107, opened 2020-12-11
Fix race condition in unit test

narrieta:
There was a race condition between the test command creating its file and the invocation of the callback.


narrieta:
the changes in the cgroups code should not have been included in this PR; will post a new one

PR #2108, opened 2020-12-11
Fix race condition in unit test

narrieta:
There was a race condition between the child process creating its file and the callback being invoked. Fixed by waiting for the callbacks explicitly.

### DIFF
--- a/tests/utils/test_shell_util.py
+++ b/tests/utils/test_shell_util.py
@@ -366,16 +366,16 @@ exit({0})
 
     # R0912: Too many branches (13/12) (too-many-branches) -- Disabled: Branches are sequential
     def test_run_command_run_pipe_run_get_output_should_notify_when_commands_start_and_complete(self):  # pylint:disable=R0912
-        # The children processes run this script, which creates a file (this signals that process has started) and then sleeps for a long time
+        # The children processes run this script, which creates a file with the PIDs of the script and its parent and then sleeps for a long time
         child_script = os.path.join(self.tmp_dir, "should_notify_when_commands_start_and_complete.py")
         AgentTestCase.create_script(child_script, """
 import os
 import sys
 import time
 
-with open(sys.argv[1], "w") as signal_file:
-    signal_file.write("{0} {1}".format(os.getpid(), os.getppid()))
-time.sleep(60)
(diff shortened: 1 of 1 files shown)


tests/utils/test_shell_util.py
```diff
@@ -490,6 +480,15 @@ def _get_command_line(pid):
             pass
         return "[PID: {0}] UNKNOWN".format(pid)
 
+    @staticmethod
+    def _wait_for(predicate):
+        start_time = datetime.datetime.now()
+        while RunCommandTestCase._to_seconds(datetime.datetime.now() - start_time) < 30:
```
pgombar, line 486: So this one test could take up to 30 seconds? Any reason for waiting that long?
Granted, this should only happen if the threads were unable to start for some reason, which is not a normal case.

larohra (approved):

tests/utils/test_shell_util.py
narrieta: No, the test takes 0.05-0,0.08 s (my local machine). 30 s is the timeout if things go wrong

pgombar: Yes, I was wondering if that timeout could be shorter so we fail faster, but I realized this probably won't happen often.

pgombar (approved):

tests/utils/test_shell_util.py
narrieta: Oh, I think I misunderstood your point. 30 s is probably too generous. I am trying to avoid the test being noisy if, for example, one is running the unit test in a laptop while using a chat app that is taking most of the cpu.

pgombar: Sounds like a bad chat app :smile:

narrieta: And probably we should add a verification to the entire suite to let us know if the run takes more than some given time. A bad mock, a regression in prod/test code may go unnoticed until somebody stumbles into a slow test run and realizes something is not correct.

kevinclark19a: I can look into the Github Actions docs; I feel like this sort of thing should be easy to automate as a separate, not required job.

PR #2109, opened 2020-12-13
Use an array to keep track of running commands

narrieta:
Using an array instead of callbacks is simpler.

Also, fixing an unrelated pylint warning that I noticed while browsing the code.


tests/utils/test_shell_util.py
```diff
@@ -426,49 +419,54 @@ def invoke(command):
...
+            # once the processes complete, their PIDs should go away
+            running_commands = [[]]
+
+            def all_commands_completed():
+                running_commands[0] = shellutil.get_running_commands()
+                return len(running_commands[0]) == 0
+
+            if not self._wait_for(all_commands_completed):
+                self.fail("shellutil.get_running_commands() should return empty after the commands complete. Got: {0}".format(
+                    self._format_pids(running_commands[0])))
```
pgombar, line 460: Would it be useful for debugging to output the process names along with PIDs?

tests/utils/test_shell_util.py
```diff
@@ -426,49 +419,54 @@ def invoke(command):
...
+
+                def all_commands_running():
+                    running_commands[0] = shellutil.get_running_commands()
+                    return len(running_commands[0]) >= len(commands_to_execute) + 1  # +1 because run_pipe starts 2 commands
 
-                # wait for the command_started callbacks and verify them
-                if not self._wait_for(lambda: len(started_commands) == len(commands_to_execute) + 1):   # +1 because run_pipe starts 2 commands
-                    raise Exception("The child processes did not complete within the allowed timeout")
+                if not self._wait_for(all_commands_running):
+                    self.fail("shellutil.get_running_commands() did not report the expected number of commands after the allowed timeout. Got: {0}".format(
```
pgombar, line 434: Can we add the expected number to the error message?

tests/utils/test_shell_util.py
narrieta: _format_pids does that

tests/utils/test_shell_util.py
```diff
@@ -426,49 +419,54 @@ def invoke(command):
...
 
-                # the first item to in the PIDs we fetched corresponds run_get_output, which invokes the command using the
-                # shell, so in that case we need to use the parent's pid (which was the process we started)
-                started_commands_expected = parent_processes[0:1] + child_processes[1:]
+                # the first item to in the PIDs we fetched corresponds to run_get_output, which invokes the command using the
+                # shell, so in that case we need to use the parent's pid (i.e. the shell that we started)
+                started_commands = parent_processes[0:1] + child_processes[1:]
+
+                # wait for all the commands to start
+                running_commands = [[]]
```
larohra, line 427: Out of curiosity, why are we using a 2D array here?

tests/utils/test_shell_util.py
pgombar: You're right, based on the name I assumed it just formats the PIDs, not the names too.

larohra (approved):

tests/utils/test_shell_util.py
narrieta: all_commands_running() needs to change the value and would not be able to do it if it was scalar

tests/utils/test_shell_util.py
narrieta: i'll report the expected commands

narrieta: done

pgombar (approved):

tests/utils/test_shell_util.py
larohra: Ahh got it. I thought single arrays were modifiable outside their scope but apparently not, thanks for the clarification!

larohra (approved):

PR #2116, opened 2020-12-16
Make IPoIB interface online

longlimsft:
Some Infiniband applications require IPoIB interface to be online for
address resolution to work. This patch makes the interface online during
provisioning. The return code is ignored as the link status doesn't need
to be checked and it doesn't affect its IP address.

## Description

Issue # 

---

### PR information


narrieta:
> The return code is ignored as the link status doesn't need
> to be checked and it doesn't affect its IP address.

If the intention is to ignore the return code, then this call should be wrapped by try/except CommandError:

```
                        ip_command = ["ip", "link", "set", nic, "up"]
                        shellutil.run_command(ip_command)
```

longlimsft:
> > The return code is ignored as the link status doesn't need
> > to be checked and it doesn't affect its IP address.
> 
> If the intention is to ignore the return code, then this call should be wrapped by try/except CommandError:
> 

This command always return 0 regardless of link status changes. The only time it will error out is when "nic" is an invalid name, in which case there is no need to execute the following command in the same try block.

Since it's in the same try block as the following command , there is no need to use another try block.

> ```
>                         ip_command = ["ip", "link", "set", nic, "up"]
>                         shellutil.run_command(ip_command)
> ```

narrieta:
@longlimsft ok; will run automation

larohra (commented):
Code looks good, running automation

larohra (approved):
Automation ok. LGTM

PR #2117, opened 2020-12-17
Use PIDs instead of command lines to identify processes in the agent's cgroup

narrieta:
Using command lines to identify those processes was unreliable and hard to maintain. PIDs should address those issues.


azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -225,29 +237,70 @@ def __impl():
...
+
+            unexpected = []
+            for process in agent_cgroup:
+                if process in (daemon, extension_handler):
+                    continue
+                # check if the process is a command started by the agent or a descendant of one of those commands
+                current = process
+                while current != 0 and current not in agent_commands:
+                    current = self._get_parent(current)
+                if current == 0:
```
pgombar, line 267: Isn't `current` supposed to be a PID? Why compare against 0?

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -225,29 +237,70 @@ def __impl():
...
+            agent_commands.update(shellutil.get_running_commands())
+
+            unexpected = []
+            for process in agent_cgroup:
+                if process in (daemon, extension_handler):
+                    continue
+                # check if the process is a command started by the agent or a descendant of one of those commands
+                current = process
+                while current != 0 and current not in agent_commands:
+                    current = self._get_parent(current)
```
pgombar, line 266: Why not add `current` but its parent?

tests/common/test_cgroupconfigurator.py
```diff
@@ -17,16 +17,21 @@
...
 from azurelinuxagent.common.cgroup import CGroup
-from azurelinuxagent.common.cgroupconfigurator import CGroupConfigurator
+from azurelinuxagent.common.cgroupconfigurator import CGroupConfigurator, UnexpectedProcessesInCGroupException
 from azurelinuxagent.common.cgroupstelemetry import CGroupsTelemetry
 from azurelinuxagent.common.exception import CGroupsException
+from azurelinuxagent.common.utils import shellutil
 from tests.common.mock_cgroup_commands import mock_cgroup_commands
 from tests.tools import AgentTestCase, patch, mock_sleep
-
+from tests.utils.miscelaneous_tools import format_processes, wait_for
```
pgombar, line 34: Typo in module name, miscelaneous > miscellaneous

azurelinuxagent/common/cgroupconfigurator.py
narrieta: 0 is the PID of the scheduler; all processes descent from it

azurelinuxagent/common/cgroupconfigurator.py
narrieta: sorry, not sure i understand the question

tests/common/test_cgroupconfigurator.py
narrieta: thanks, will fix it

azurelinuxagent/common/cgroupconfigurator.py
pgombar: Didn't realize you are climbing up the process tree until you reach either the agent or the scheduler (parent process of all). Makes sense now!

tests/common/test_cgroupconfigurator.py
narrieta: done

pgombar (approved):

PR #2119, opened 2020-12-18
Devuan support

peter9370:
## Description

Support for the devuan linux distribution (debian without systemd)

Issue # 

The changes are currently in the branch devuan_support: the latest changes on the Azure/WALinuxAgent develop branch were merged into local develop a few days ago, and the changes in devuan_support were rebased onto the latest state of the local develop branch.

Changes/additions:
==================

1) Added azurelinuxagent/common/extralib/debian_recheck.py to circumvent the problem that currently in devuan, the python module platform.linux_distribution() reports the distro as "debian". An instance of the object DebianRecheck is created, and the distribution information as currently known is passed to it. The init method
within the object collects distribution information from files in /etc and /var/lib/apt/lists, and saves it in the self.sourcedata
structure. If anything doesn't look right or unexpected behaviour results, fail safe - reset the items in self.sourcedata to the information originally given.

The intention in writing the DebianRecheck code was to use to use python core elements only, and to work out the distribution from information in files only: that way, it was hoped to avoid the instability in the python platform modules, and also to permit changes to DebianRecheck without affecting anything else. If or when the instability in the python platform modules is resolved so that devuan is correctly detected in all cases, debian_recheck.py will become redundant and can be removed.

2) Modified azurelinuxagent/common/future.py - if the distro is reported as debian, use DebianRecheck to re-check it.

3) Added azurelinuxagent/common/osutil/devuan.py - class DevuanOSUtil containing methods for devuan distros (start and stop using openrc init scripts instead of systemd)

4) Modified azurelinuxagent/common/osutil/factory.py - return DevuanOSUtil if debian is reported as distro and if DebianRecheck affirms that it's actually devuan

5) Modified azurelinuxagent/common/version.py:

- added "devuan" to the list of extra distros supported;
- unsure about value for AGENT_VERSION - tried 2.2.53.1 but this seems to break tests. Eventually retained 2.2.53 as in current version on develop branch.

6) Added config/devuan/waagent.conf

7) Added init/devuan/walinuxagent - init.d start/stop script for walinuxagent (NB this is different from the one currently used in debian packages: in the latter, DAEMON is set to /usr/sbin/waagent - with this, the system can't stop the agent daemon successfully. DAEMON needs to be set to /usr/bin/python3, with /usr/sbin/waagent as the first argument to it.)

8) Modified setup.py:

- added sections to install non-systemd init files
- added entry_points - without this, /usr/sbin/waagent doesn't get generated during the debian build process

9) Modified tests/common/osutil/test_factory.py - added method to test operation when devuan is given as the distro. (At the moment, the DebianRecheck() code is not being adequately tested under travis, as the ubuntu host environment is causing the code to fail-safe every time, so devuan is never reported. 

10) Modified various tests to add fixes and workarounds. This was done because the tests were failing when run as part of attempts to build .deb packages under the devuan v2.1 environment. They may or may not be relevant in other
environments.

- tests/common/test_cgroupconfigurator.py - added code to skip tests in CGroupConfiguratorSystemdTestCase if systemd isn't present. This is probably controversial - maybe for each skipped test, there should be an additional test to prove that the functionality works when systemd isn't present?

- tests/common/test_event.py. This test was failing because the build was being run as root: 2 valid and 3 invalid events should have been seen, and one of the invalid ones is based on a file being set to be unreadable. But if the build is run as root, the file can still be read, so the event is collected as valid.  Added code to check if we're running as root, and if so, increment/decrement the expected valid and invalid event counts.

- tests/ga/test_monitor.py - skip the test test_it_should_report_processes_that_do_not_belong_to_the_agent_cgroup if
systemd is not present 

- tests/ga/test_update.py - when run under devuan 2.1, test_run_latest_defaults_to_current was failing:
apparently the value from sys.argv[0] was expected to be a list, whereas it was actually a string containg three space-separated values. Added code to turn it into a list before incorporating into the result. This probably needs revisiting: the problem occurred whilst running the tests under devuan, and might not occur in other environments:
probably, sys.argv[0] should be tested before being modified.

- tests/utils/test_rest_util.py - test_get_http_proxy_configuration_requires_host and test_get_no_proxy_default were failing. Think this was due to environment variables being set - presumably by previous tests. Tried adding code to clean up the environment before running the test - this seemed to fix the problem.

11) It was realised that the Travis builds effectively only test that DebianRecheck fails safe (it's told that the distro is "debian", but the environment is Ubuntu). To address this and provide more detailed testing, tests/common/extralib/test_debian_recheck.py was provided. This uses files collected from the following environments: debian stretch (v9),
debian buster (v10), devuan ascii (v2.1) and devuan beowulf (v3.0), and placed in tests/data/devuan_support. The tests create an instance of DebianRecheck, then substitute paths to the test data files for the ones found in the live environment. The code in debian_recheck.py was modified to facilitate this.
---

### PR information
- [x ] The title of the PR is clear and informative.
- [x ] Except for special cases involving multiple contributors, the PR is started from a fork of the main repository, not a branch.
- [] If applicable, the PR references the bug/issue that it fixes in the description.


narrieta:
@peter9370 thank you for your contribution. we'll check it out.

peter9370:
Hi Norberto,

Thank you. It's the first time that I've contributed to a github
project, so I've probably missed something - if so, let me know what's
wrong and/or missing, and I'll be more than happy to fix it if I can.

Best regards,
Peter

peter9370:
Hi,

Apologies - I missed the notification that I needed to sign the CLA before this could proceed. I'm now discussing this with my employer, and hope that I'll be able to sign the agreement in the very near future.

Best regards,
Peter

peter9370:
@narrieta please could you ask the assigned code reviewers that they have not yet reviewed my proposed additions and changes. If there is something which is preventing the code review from being done, please let me know, and I will fix it. 

Thanks.

narrieta:
@peter9370 Thanks for your contribution, we are in the middle of a release. but will start checking this PR within a week or so

narrieta:
@peter9370 - sorry for the delay, i'm starting this review now should be able to complete it in a few days

peter9370:
@narrieta thanks for that - I'll wait to hear further. If I can do anything to assist, let me know.

peter9370:
@narrieta Hi Norbert - do you know if there was any further progress in reviewing my changes?  If there's anything I need to do now before the review can proceed, let me know.

PR #2125, opened 2020-12-28
Suppress Convention and Refactor pylint rules globally

narrieta:


ci/2.7.pylintrc
```diff
@@ -7,19 +7,12 @@
 # bad-open-mode<W1501>: (hi-pri) Python supports: r, w, a[, x] modes with b, +, and U (only with r) options.
 # bad-super-call<E1003>: (hi-pri) Used when another argument than the current class is given as first argument of the super builtin.
 # bare-except<W0702>: (hi-pri) Used when an except clause doesn't specify exceptions type to catch.
-# blacklisted-name<C0102>: (needs review) Used when the name is listed in the black list (unauthorized names).
 # cell-var-from-loop<W0640>: (hi-pri) A variable used in a closure is defined in a loop. This will result in all closures using the same value for the closed-over variable.
-# cyclic-import<R0401>: (hi-pri) Used when a cyclic import between two or more modules is detected.
```
larohra, line 12: We should not remove this error, this doesn't look cosmetic

ci/2.7.pylintrc
```diff
@@ -7,19 +7,12 @@
...
 # expression-not-assigned<W0106>: (hi-pri) Used when an expression that is not a function call is assigned to nothing.
-# duplicate-code<R0801>: (hi-pri) Indicates that a set of similar lines has been detected among multiple file. (also disabled globally)
 # function-redefined<E0102>: (hi-pri) Used when a function / class / method is redefined.
 # global-statement<W0603>: Used when you use the "global" statement to update a global variable.
 # import-error<E0401>: (hi-pri) Used when pylint has been unable to import a module.
-# inconsistent-return-statements<R1710>: (hi-pri) if any return statement returns an expression, any return statements where no value is returned should explicitly state this as return None, and an explicit return statement should be present at the end of the function
```
larohra, line 18: Maybe this too?

ci/2.7.pylintrc
```diff
@@ -7,19 +7,12 @@
...
-# invalid-name<C0103>: (needs review) Used when the name doesn't conform to naming rules associated to its type (constant, variable, class...).
 # keyword-arg-before-vararg<W1113>: (hi-pri) When defining a keyword argument before variable positional arguments, one can end up in having multiple values passed for the aforementioned parameter in case the method is called with keyword arguments.
-# len-as-condition<C1801>: (hi-pri) Used when Pylint detects that len(sequence) is being used without explicit comparison inside a condition to determine if a sequence is empty.
-# no-else-return<R1705>: (hi-pri) Used in order to highlight an unnecessary block of code following an if containing a return statement.
```
larohra, line 22: maybe this too?

ci/3.6.pylintrc
```diff
@@ -8,26 +8,14 @@
...
-# invalid-name<C0103>: (needs review) Used when the name doesn't conform to naming rules associated to its type (constant, variable, class...).
 # keyword-arg-before-vararg<W1113>: (hi-pri) When defining a keyword argument before variable positional arguments, one can end up in having multiple values passed for the aforementioned parameter in case the method is called with keyword arguments.
-# len-as-condition<C1801>: (hi-pri) Used when Pylint detects that len(sequence) is being used without explicit comparison inside a condition to determine if a sequence is empty.
 # lost-exception<W0150>: (needs review) Used when a break or a return statement is found inside the finally clause of a try...finally block: the exceptions raised in the try clause will be silently swallowed instead of being re-raised.
-# no-else-break<R1723>: (hi-pri) Used in order to highlight an unnecessary block of code following an if containing a break statement.
```
larohra, line 27: maybe these too?

ci/2.7.pylintrc
narrieta: There were no instances of this suppression, this comment was out of date.

ci/2.7.pylintrc
narrieta: Instead of suppressing all R warning, I changed the settings to suppress only the warnings the agent is disabling. I added this one back.

ci/2.7.pylintrc
narrieta: those no-else-* warnings are more stylistic than a real issue

ci/3.6.pylintrc
narrieta: those no-else-* warnings are more stylistic than a real issue

ci/2.7.pylintrc
larohra: >There were no instances of this suppression

Wont it still be good to have this check incase we unknowingky add a cyclic dependency in the future?

ci/2.7.pylintrc
larohra: Awesome thanks! If we're not blocking this then we should keep this comment too

ci/2.7.pylintrc
larohra: Keep this comment if we're not ignoring this error

narrieta: The original intention of those comments was to document the text description of the message, but they are already out of sync with the code and are likely to get more out of sync and we make additional changes. I was inclined to remove them altogether, but the rcfile for 2.7 has some infor wrt python versions that may be useful.

ci/2.7.pylintrc
narrieta: The original intention of those comments was to document the text description of the message, but they are already out of sync with the code and are likely to get more out of sync as we make additional changes. I was inclined to remove them altogether, but the rcfile for 2.7 has some infor wrt python versions that may be useful.

ci/2.7.pylintrc
narrieta: BTW, the description can be obtained from the command line using "pylint --list-msgs"

larohra: Hmm interesting, I thought the idea behind adding the descriptions here was to have a one stop shop for all the messages we were ignoring in our code. 

Out of curiosity, what exactly do you mean by out of sync? Do we have some pylint warnings which are there in the code but not in our .rc file and vice versa?

narrieta: >I thought the idea behind adding the descriptions here was to have a one stop shop for all the messages we were ignoring in our code.
Exactly, but they are already out of sync. The one stop shop can instead be  "pylint --list-msgs"

> what exactly do you mean by out of sync? 
Not all the suppressions in the actual code are in this list. This list was also meant to be used for a rough prioritization of messages that need to be looked at, but with this change we are suppressing globally all C and many R, so the rest of the suppressions in the code do need to be looked at and addressed.

larohra: >The one stop shop can instead be "pylint --list-msgs"

Agreed to this, always good to follow the standards of the tools we're using rather than keeping our own copy of it.

 >Not all the suppressions in the actual code are in this list.

Hmm interesting, I guess we missed some down the line. Anywho, I'm good with removing all these comments then (Y)

narrieta: OK, I just pushed another commit removing all the comments

ci/2.7.pylintrc
```diff
@@ -1,50 +1,6 @@
 # python2.7 uses pylint 1.9.5, whose docs can be found here: http://pylint.pycqa.org/en/1.9/technical_reference/features.html#messages
 # python3.4 uses pylint 2.3.1, whose docs can be found here: http://pylint.pycqa.org/en/pylint-2.3.1/technical_reference/features.html
 
-# The source code suppresses the following error types common between python2.7 and python3.4:
```
larohra, line 4: Thanks for removing this. Can you add a small comment on how to view the description (basically adding the command `pylint --list-msgs`)

azurelinuxagent/common/cgroupapi.py
```diff
@@ -368,13 +366,13 @@ def start_extension_command(self, extension_name, command, timeout, shell, cwd,
...
                     self._systemd_run_commands.remove(process.pid)
-        except ExtensionError as e:  # pylint: disable=C0103
+        except ExtensionError as e:
             # The extension didn't terminate successfully. Determine whether it was due to systemd errors or
             # extension errors.
             systemd_failure = self._is_systemd_failure(scope, stderr)
             process_output = read_output(stdout, stderr)
 
-            if not systemd_failure:  # pylint: disable=R1720
+            if not systemd_failure:
```
kevinclark19a, line 375: I'm not sure that `R1720=no-else-raise` is the same type of warning as the others here.

This one is referring to the `else` below this `if`, right? It might be clearer to simply remove the `else`.

azurelinuxagent/common/conf.py
```diff
@@ -55,7 +55,7 @@ def get(self, key, default_val):
 
     def get_switch(self, key, default_val):
         val = self.values.get(key)
-        if val is not None and val.lower() == 'y':  # pylint: disable=R1705
+        if val is not None and val.lower() == 'y':
```
kevinclark19a, line 58: Similar comment to the one I left above, though this time I wonder if this warning is just a symptom of this code being unnecessarily verbose/repetitive. 

Not sure it's worth trying to refactor this to a potentially cleaner structure (especially with a lack of native switch statements in python), but we can always just turn the `elif` into an `if`. I think that would make it a bit more readable, to me at least.

Thoughts?

kevinclark19a: I think the normal python replacement for a switch statement is to use a dictionary (a `defaultdict` probably better matches this case if that is easily imported across all our supported versions of python), if we want to go that route.

azurelinuxagent/common/datacontract.py
```diff
@@ -45,7 +45,7 @@ def validate_param(name, val, expected_type):
 
 
 def set_properties(name, obj, data):
-    if isinstance(obj, DataContract):  # pylint: disable=R1705
+    if isinstance(obj, DataContract):
```
kevinclark19a, line 48: I think leaving the `else` below in this case is actually better for readability, so anyone reading this code doesn't have to track down every branch below to make sure it returns. So I think I'm good with this change.

azurelinuxagent/common/datacontract.py
```diff
@@ -69,7 +69,7 @@ def set_properties(name, obj, data):
 
 
 def get_properties(obj):
-    if isinstance(obj, DataContract):  # pylint: disable=R1705
+    if isinstance(obj, DataContract):
```
kevinclark19a, line 72: Same as the other comment on this file.

azurelinuxagent/common/dhcp.py
```diff
@@ -67,7 +67,7 @@ def wait_for_network(self):
         Wait for network stack to be initialized.
         """
         ipv4 = self.osutil.get_ip4_addr()
-        while ipv4 == '' or ipv4 == '0.0.0.0':  # pylint: disable=R1714
+        while ipv4 == '' or ipv4 == '0.0.0.0':
```
kevinclark19a, line 70: I personally prefer using the `in` that this warning (`R1714=consider-using-in`) suggests, e.g. `ipv4 in ('', '0.0.0.0')`, but I don't think it really matters here. I'm fine with it as is, but figured I should mention it because I've been pointing out the `R17..` suppressions.

azurelinuxagent/common/dhcp.py
```diff
@@ -257,7 +257,7 @@ def parse_route(response, option, i, length, bytes_recv):  # pylint: disable=W06
 
 
 def parse_ip_addr(response, option, i, length, bytes_recv):
-    if i + 5 < bytes_recv:  # pylint: disable=R1705
+    if i + 5 < bytes_recv:
```
kevinclark19a, line 260: Consider removing the `else` below; I personally would find it more readable.

azurelinuxagent/common/cgroupapi.py
narrieta: This PR is not attempting to fix any warnings, it is simply suppressing warnings we do not care about globally instead of using inline suppressions.

Yes, the code can be written without the else, do feel free to address those in a separate PR.

azurelinuxagent/common/conf.py
narrieta: Same as my answer above.

azurelinuxagent/common/dhcp.py
narrieta: Feel free to fix those on a separate PR

azurelinuxagent/common/dhcp.py
narrieta: same as above

azurelinuxagent/common/cgroupapi.py
kevinclark19a: As per the offline discussion, I think it's fine to leave these as is! My thoughts are that it would be a nice to have, but I don't think a PR should be blocked for not conforming to this style.

azurelinuxagent/pa/rdma/suse.py
```diff
@@ -25,7 +25,7 @@
 
 class SUSERDMAHandler(RDMAHandler):
 
-    def install_driver(self):  # pylint: disable=R0912,R1710
+    def install_driver(self):  # pylint: disable=R1710
```
kevinclark19a, line 28: As a note to any potential future reviewers, the remaining warning suppression `R1710=inconsistent-return-statements` is _not_ style related. It marks this function as having `return` and `return True` statements in the same function.

azurelinuxagent/pa/rdma/ubuntu.py
```diff
@@ -79,7 +79,7 @@ def install_driver(self):  # pylint: disable=R0912,R0911
         status,output = shellutil.run_get_output('apt-cache show --no-all-versions linux-azure')
         if status != 0:
             return
-        r = re.search('Version: (\S+)', output)  # pylint: disable=W1401,C0103
+        r = re.search('Version: (\S+)', output)  # pylint: disable=W1401
```
kevinclark19a, line 82: Ah, this `W1401=anomalous-backslash-in-string` is I think a missed warning suppression; we had a PR removing this warning from the comment block in `ci/*.pylintrc`. With the move to `pylint --list-msgs` this is actually not a big deal, so I don't think we need to take any action (i.e. creating a ticket to fix it) besides our usual linter hunting.

Thoughts?

ci/2.7.pylintrc
```diff
@@ -1,113 +1,38 @@
...
-# no-else-raise<R1720>: (hi-pri) Used in order to highlight an unnecessary block of code following an if containing a raise statement. 
-# simplifiable-if-expression<R1719>: Used when an if expression can be replaced with 'bool(test)'.
-# subprocess-popen-preexec-fn<W1509>: (hi-pri) The preexec_fn parameter is not safe to use in the presence of threads in your application. The child process could deadlock before exec is called.
-# useless-return<R1711>: (hi-pri) Emitted when a single "return" or "return None" statement is found at the end of function or method definition.
-
-
 [MESSAGES CONTROL]
 
-disable=bad-continuation, # Buggy, **REMOVED in pylint-2.6.0**
+disable=C, # (C) convention, for programming standard violation
```
kevinclark19a, line 6: This replaces the specific, removed warnings below, right?

ci/2.7.pylintrc
```diff
@@ -1,113 +1,38 @@
...
-    multiple-statements, # Used when more than on statement are found on the same line.
-    no-self-use, # Used when a method doesn’t use its bound instance, and so could be written as a function.
     old-style-class, # Used when a class is defined that does not inherit from another class and does not inherit explicitly from “object”. (2.7 only)
-    superfluous-parens, # Used when a single item in parentheses follows an if, for, or other keyword.
-    too-many-statements, # Used when a function or method has too many statements. You should then split it in smaller functions / methods.
-    trailing-newlines, # Used when there are trailing blank lines in a file.
-    trailing-whitespace, # Used when there is whitespace between the end of a line and the newline.
-    useless-object-inheritance, # Used when a class inherit from object, which under python3 is implicit (3+ only)
-    duplicate-code, # Indicates that a set of similar lines has been detected among multiple file. TODO: this is on our "fix" list, and is thus left last.
     fixme, # Used when a warning note as FIXME or TODO is detected
```
kevinclark19a, line 38: NIT: trailing comma on this line. There might be some issues with parsing the rcfile here if we ever add another section after this one.

narrieta: Yes, I noticed that extra comma and decided to leave it alone since it may be handy if adding more suppressions.

ci/2.7.pylintrc
narrieta: It removes all the Convenition warnings:
>$ pylint --list-msgs | grep -E 'C[[:digit:]]{4}'
:blacklisted-name (C0102): *Black listed name "%s"*
:invalid-name (C0103): *%s name "%s" doesn't conform to %s*
:empty-docstring (C0112): *Empty %s docstring*
:unneeded-not (C0113): *Consider changing "%s" to "%s"*
:missing-module-docstring (C0114): *Missing module docstring*
:missing-class-docstring (C0115): *Missing class docstring*
:missing-function-docstring (C0116): *Missing function or method docstring*
:singleton-comparison (C0121): *Comparison to %s should be %s*
:misplaced-comparison-constant (C0122): *Comparison should be %s*
:unidiomatic-typecheck (C0123): *Using type() instead of isinstance() for a typecheck.*
:non-ascii-name (C0144): *%s name "%s" contains a non-ASCII unicode character*
:consider-using-enumerate (C0200): *Consider using enumerate instead of iterating with range and len*
:consider-iterating-dictionary (C0201): *Consider iterating the dictionary directly instead of calling .keys()*
:bad-classmethod-argument (C0202): *Class method %s should have %s as first argument*
:bad-mcs-method-argument (C0203): *Metaclass method %s should have %s as first argument*
:bad-mcs-classmethod-argument (C0204): *Metaclass class method %s should have %s as first argument*
:single-string-used-for-slots (C0205): *Class __slots__ should be a non-string iterable*
:line-too-long (C0301): *Line too long (%s/%s)*
:too-many-lines (C0302): *Too many lines in module (%s/%s)*
:trailing-whitespace (C0303): *Trailing whitespace*
:missing-final-newline (C0304): *Final newline missing*
:trailing-newlines (C0305): *Trailing newlines*
:multiple-statements (C0321): *More than one statement on a single line*
:superfluous-parens (C0325): *Unnecessary parens after %r keyword*
:mixed-line-endings (C0327): *Mixed line endings LF and CRLF*
:unexpected-line-ending-format (C0328): *Unexpected line ending format. There is '%s' while it should be '%s'.*
:wrong-spelling-in-comment (C0401): *Wrong spelling of a word '%s' in a comment:*
:wrong-spelling-in-docstring (C0402): *Wrong spelling of a word '%s' in a docstring:*
:invalid-characters-in-docstring (C0403): *Invalid characters %r in a docstring*
:multiple-imports (C0410): *Multiple imports on one line (%s)*
:wrong-import-order (C0411): *%s should be placed before %s*
:ungrouped-imports (C0412): *Imports from package %s are not grouped*
:wrong-import-position (C0413): *Import "%s" should be placed at the top of the module*
:useless-import-alias (C0414): *Import alias does not rename original package*
:import-outside-toplevel (C0415): *Import outside toplevel (%s)*
:len-as-condition (C1801): *Do not use `len(SEQUENCE)` without comparison to determine if a sequence is empty*

kevinclark19a: Also, does this grouping include any warnings besides the ones removed below (and, of course, the warnings removed in this PR)?

narrieta: (of all the C warnings, the agent was hitting those:
>$ function pylint-rules {
>     pylint --list-msgs | sed -E '/^:([[:alpha:]-]{2,}).*\(C[[:digit:]]{4}\).*/!d;s//\1/' | sort -u
> }
$ function pylint-agent {
>     find azurelinuxagent tests -name '*.py' -exec sed -E '/^.*# pylint:\s?disable=([[:alpha:]-]{2,}).*$/!d;s//\1/' {} \; | sort -u
> }
$ 
$ comm -12 <(pylint-rules) <(pylint-agent)
blacklisted-name
invalid-name
len-as-condition
too-many-lines
ungrouped-imports
unidiomatic-typecheck

)

kevinclark19a (approved):

larohra (approved):

PR #2127, opened 2021-01-06
Disable resource enforcement when the agent's cgroup includes other processes

narrieta:
* Disable resource enforcement when the agent's cgroup includes other processes
* Move the reporting logic from the monitor thread to cgroup configurator, since this check will be done from other places, too
* Always create a periodic task for resource monitoring/enforcement, even if the feature is initially disabled (we will start enabling/disabling the feature dynamically)

### DIFF
--- a/azurelinuxagent/common/cgroupconfigurator.py
+++ b/azurelinuxagent/common/cgroupconfigurator.py
@@ -30,22 +30,11 @@ from azurelinuxagent.common.utils.extensionprocessutil import handle_process_com
 from azurelinuxagent.common.event import add_event, WALAEventOperation
 
 
-class UnexpectedProcessesInCGroupException(CGroupsException):
-    """
-    Raised by CGroupConfigurator.check_processes_in_agent_cgroup() when the agent's cgroup includes processes
-    that should not belong to it.
-    The 'unexpected' property is a list of the processes (strings) that should not belong to the agent's cgroup
-    """
-    def __init__(self, unexpected):
-        super(UnexpectedProcessesInCGroupException, self).__init__("Unexpected processes in agent's cgroup")
-        self.unexpected = unexpected
-
-
 class CGroupConfigurator(object):
     """
     This class implements the high-level operations on CGroups (e.g. initialization, creation, etc)
 
-    NOTE: with the exception of start_extension_command and check_processes_in_agent_cgroup, none of the methods in this class
+    NOTE: with the exception of start_extension_command, none of the methods in this class
     raise exceptions (cgroup operations should not block extensions)
     """
     class __Impl(object):
@@ -56,8 +45,8 @@ class CGroupConfigurator(object):
             self._cgroups_api = None
             self._agent_cpu_cgroup_path = None
             self._agent_memory_cgroup_path = None
-            self._get_processes_in_agent_cgroup_last_error = None
-            self._get_processes_in_agent_cgroup_error_count = 0
+            self._check_processes_in_agent_cgroup_last_error = None
+            self._check_processes_in_agent_cgroup_error_count = 0
 
         def initialize(self):
             try:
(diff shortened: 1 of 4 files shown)


azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -248,37 +237,52 @@ def check_processes_in_agent_cgroup(self):
...
+                            break
+                if unexpected:
+                    unexpected = self._format_processes(unexpected)
+                    unexpected.sort()
+                    log_message("The agent's cgroup includes unexpected processes; disabling CPU enforcement. Unexpected: {0}".format(unexpected))
+                    self.disable()
+                    return False
+            except Exception as exception:
+                log_message("Failed to check the processes in the agent's cgroup: {0}".format(ustr(exception)))
+            return True
```
pgombar, line 285: This will return True if there was an exception to check the processes, is that expected?

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -248,37 +237,52 @@ def check_processes_in_agent_cgroup(self):
...
-                    if len(unexpected) >= 5:  # collect just a small sample
-                        break
-            if unexpected:
-                raise UnexpectedProcessesInCGroupException(unexpected=self._format_processes(unexpected))
+            def log_message(message):
+                # Report only a small sample of errors
+                if message != self._check_processes_in_agent_cgroup_last_error and self._check_processes_in_agent_cgroup_error_count < 5:
+                    self._check_processes_in_agent_cgroup_error_count += 1
+                    self._check_processes_in_agent_cgroup_last_error = message
+                    logger.info(message)
```
pgombar, line 249: Why not warning?

azurelinuxagent/common/cgroupconfigurator.py
narrieta: Yes, it is expected

azurelinuxagent/common/cgroupconfigurator.py
narrieta: In previous releases people were concerned about cgroup warnings... once this is stable (a few releases) I'll change them to warn.

pgombar: Makes sense, thanks.

pgombar (dismissed):

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -248,37 +237,52 @@ def check_processes_in_agent_cgroup(self):
...
+            try:
+                daemon = os.getppid()
+                extension_handler = os.getpid()
+                agent_commands = set()
+                agent_commands.update(shellutil.get_running_commands())
+                systemd_run_commands = set()
+                systemd_run_commands.update(self._cgroups_api.get_systemd_run_commands())
+                agent_cgroup = CGroupsApi.get_processes_in_cgroup(self._agent_cpu_cgroup_path)
+                # get the running commands again in case new commands were started while we were fetching the processes in the cgroup;
+                agent_commands.update(shellutil.get_running_commands())
```
larohra, line 261: Why not just have it in a single place (Ln 261) instead of 2 places (Ln 256 and Ln 261)?

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -248,37 +237,52 @@ def check_processes_in_agent_cgroup(self):
...
+                    current = process
+                    while current != 0 and current not in agent_commands:
+                        current = self._get_parent(current)
+                    if current == 0:
+                        unexpected.append(process)
+                        if len(unexpected) >= 5:  # collect just a small sample
+                            break
+                if unexpected:
+                    unexpected = self._format_processes(unexpected)
+                    unexpected.sort()
```
larohra, line 279: Any particular reason for this? Or just because it looks clean?

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -248,37 +237,52 @@ def check_processes_in_agent_cgroup(self):
...
+                    log_message("The agent's cgroup includes unexpected processes; disabling CPU enforcement. Unexpected: {0}".format(unexpected))
+                    self.disable()
+                    return False
+            except Exception as exception:
+                log_message("Failed to check the processes in the agent's cgroup: {0}".format(ustr(exception)))
```
larohra, line 284: Shouldn't we return False here too?

larohra (commented):
Majorly looks good, just a few comments

azurelinuxagent/common/cgroupconfigurator.py
narrieta: it's try to address this scenario:

` # get the running commands again in case new commands were started while we were fetching the processes in the cgroup`

azurelinuxagent/common/cgroupconfigurator.py
narrieta: So that the comparison with the previous value is consistent --- i'll add a comment here since I moved the code doing the comparison elsewhere

azurelinuxagent/common/cgroupconfigurator.py
narrieta: At this point the return value is not being used. I am considering removing the return value altogether and also disable cgroups if I can't perform the check, but I need to think a little more on that. Will be coming on a new PR

kevinclark19a (approved):

azurelinuxagent/common/cgroupconfigurator.py
larohra: Then why even keep keep it in the ln 256? Like Isint the call in 261 enough?

azurelinuxagent/common/cgroupconfigurator.py
larohra: Fair enough

azurelinuxagent/common/cgroupconfigurator.py
narrieta: I see, we need to capture both processes that started or completed. I'll update the comment; thanks for pointing this out.

larohra (approved):

kevinclark19a (approved):

PR #2128, opened 2021-01-07
Disable CPU enforcement if systemd-run fails

narrieta:
When systemd-run fails we invoke extensions directly so they stay in the agent's cgroup. At that point we need to disable resource enforcement or the limits for the agent would apply to the extension.


azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -328,34 +328,20 @@ def start_extension_command(self, extension_name, command, timeout, shell, cwd,
...
-                                                                          cwd=cwd, 
-                                                                          env=env, 
-                                                                          stdout=stdout, 
-                                                                          stderr=stderr, 
-                                                                          error_code=error_code) 
-
-            return process_output
+            if self.enabled():
+                try:
+                    return self._cgroups_api.start_extension_command(extension_name, command, timeout, shell=shell, cwd=cwd, env=env, stdout=stdout, stderr=stderr, error_code=error_code)
```
narrieta, line 333: Previously CGroupsApi would do the retry. I moved the retry logic here and nor the CGroupsConfigurator does the retry after disabling cgroups.

azurelinuxagent/common/event.py
```diff
@@ -72,6 +72,7 @@ class WALAEventOperation:  # pylint: disable=no-init
     ArtifactsProfileBlob = "ArtifactsProfileBlob"
     CGroupsCleanUp = "CGroupsCleanUp"
     CGroupsDebug = "CGroupsDebug"
+    CGroupsDisable = "CGroupsDisable"
```
narrieta, line 75: I need to collect info of when we disable cgroups. Added this new event.

tests/common/test_cgroupapi.py
```diff
@@ -260,217 +256,6 @@ def test_start_extension_command_should_use_systemd_to_execute_the_command(self,
                 self.assertEqual(1, len(extension_calls), "The extension should have been invoked exactly once")
                 self.assertIn("systemd-run --unit=Microsoft.Compute.TestExtension_1.2.3", extension_calls[0], "The extension should have been invoked using systemd")
 
-    @patch('time.sleep', side_effect=lambda _: mock_sleep())
```
narrieta, line 263: Moved the retry tests to cgroupconfigurator

tests/common/test_cgroupconfigurator.py
```diff
@@ -296,6 +299,233 @@ def mock_popen(command_arg, *args, **kwargs):
...
+                        extension_name="Microsoft.Compute.TestExtension-1.2.3",
+                        command=command,
+                        timeout=300,
+                        shell=True,
+                        cwd=self.tmp_dir,
+                        env={},
+                        stdout=output_file,
+                        stderr=output_file)
+
+                    self.assertFalse(configurator.enabled(), "Cgroups should have been disabled")
```
narrieta, line 331: These tests come from cgroup api with an extra check to verify cgroups were disabled

azurelinuxagent/common/cgroupapi.py
```diff
@@ -356,64 +364,28 @@ def start_extension_command(self, extension_name, command, timeout, shell, cwd,
...
                 # There was an extension error; it either timed out or returned a non-zero exit code. Re-raise the error
                 raise
+
+            # There was an issue with systemd-run. We need to log it and retry the extension without systemd.
+            process_output = read_output(stdout, stderr)
+            # Reset the stdout and stderr
+            stdout.truncate(0)
+            stderr.truncate(0)
+
+            if isinstance(e, ExtensionOperationError):
```
kevinclark19a, line 381: Can this block be broken up into multiple `except` clauses? I see an inheritance tree for `ExtensionError`s; is the `self._is_systemd_failure` check on line 371 not related to the `ExtensionError` inheritance tree? i.e., can a `systemd_failure` be reported under any `ExtensionError` subtype?

narrieta: _is_systemd_failure does not look at the exception, it is unrelated to SystemdRunError

ExtensionError indicates an error on the extension, while SystemdRunError indicates an error on systemd-run; they are unrelated.

pgombar (dismissed):

azurelinuxagent/common/cgroupapi.py
```diff
@@ -356,64 +364,28 @@ def start_extension_command(self, extension_name, command, timeout, shell, cwd,
...
+                err_msg = 'Systemd process exited with code %s and output %s' % (e.exit_code, process_output)  # pylint: disable=no-member
```
larohra, line 382: NIT: Unhandled Pylint error

azurelinuxagent/common/event.py
```diff
@@ -72,6 +72,7 @@ class WALAEventOperation:  # pylint: disable=no-init
     ArtifactsProfileBlob = "ArtifactsProfileBlob"
     CGroupsCleanUp = "CGroupsCleanUp"
     CGroupsDebug = "CGroupsDebug"
+    CGroupsDisabled = "CGroupsDisable"
```
larohra, line 75: NIT: Shouldn't it be `"CGroupsDisable**d**"` instead?

tests/common/test_cgroupconfigurator.py
```diff
@@ -296,6 +299,233 @@ def mock_popen(command_arg, *args, **kwargs):
...
+
+                    args, kwargs = mock_add_event.call_args
+                    self.assertIn("Failed to start extension Microsoft.Compute.TestExtension-1.2.3 using systemd-run.",
+                                  kwargs['message'])
+                    self.assertIn("Failed to find executable syntax_error: No such file or directory",
+                                  kwargs['message'])
+                    self.assertEqual(False, kwargs['is_success'])
+                    self.assertEqual('CGroupsDisable', kwargs['op'])
```
larohra, line 339: NIT: Why not use `WALAEventOperation.CGroupsDisabled` instead? Would make life easier if we have to change the variables later

tests/common/test_cgroupconfigurator.py
```diff
@@ -296,6 +299,233 @@ def mock_popen(command_arg, *args, **kwargs):
...
+                        stderr=output_file)
+
+                    self.assertFalse(configurator.enabled(), "Cgroups should have been disabled")
+
+                    args, kwargs = mock_add_event.call_args
+                    self.assertIn("Failed to start extension Microsoft.Compute.TestExtension-1.2.3 using systemd-run.",
+                                  kwargs['message'])
+                    self.assertIn("Failed to find executable syntax_error: No such file or directory",
+                                  kwargs['message'])
+                    self.assertEqual(False, kwargs['is_success'])
```
larohra, line 338: NIT: Use `self.assertFalse` instead

larohra (commented):
Some minor cosmetic comments else LGTM

azurelinuxagent/common/event.py
narrieta: You may need to refresh... a second commit fixed this typo

larohra: I dont see it?  Did you push your changes yet haha?

narrieta: ah, you mean the string value, I'll fix that

azurelinuxagent/common/cgroupapi.py
narrieta: this code was moved here from elsewhere... I'll add a comment

tests/common/test_cgroupconfigurator.py
narrieta: These tests were moved here from somewhere else. I also noticed some small improvements that could be done but I decided not to invest time on that at this point.

tests/common/test_cgroupconfigurator.py
narrieta: Will do it since I need to fix the typo in the event name anyway

larohra (approved):

pgombar (approved):

PR #2130, opened 2021-01-08
Redact protected settings in goal state debug files

narrieta:


azurelinuxagent/common/protocol/wire.py
```diff
@@ -607,11 +607,13 @@ def fetch_cache(self, local_file):
         except IOError as e:
             raise ProtocolError("Failed to read cache: {0}".format(e))
 
-    def save_cache(self, local_file, data):
+    @staticmethod
+    def _save_cache(data, file_name):
         try:
-            fileutil.write_file(local_file, data)
+            file_path = os.path.join(conf.get_lib_dir(), file_name)
```
narrieta, line 613: the caller used to set the path; cleaner if we do it here

larohra (approved):

pgombar (approved):

PR #2138, opened 2021-01-19
Fix logrotate config

kevinclark19a:
## Description

The `waagent-extn.logrotate` file added in PR#1740 conflicts with some linux extensions (e.g. LAD and OMS) that provide `logrotate` configuration for log files that they own.

The fix for this issue is to limit that file to only the files created (and therefore owned) by the agent.

---

### PR information


config/waagent-extn.logrotate
```diff
@@ -1,4 +1,4 @@
-/var/log/azure/*/*.log {
+/var/log/azure/*/CommandExecution.log {
```
narrieta, line 1: will this produce a conflict if an extension creates a rule for /var/log/azure/<extension>/*?

kevinclark19a: Yes, but don't think there's a way around that short of moving the location of the `CommandExecution.log` file, assuming that we still want it `logrotate`d. I've verified that the original extensions we conflicted with do not have that behavior, and I don't think it's viable to test every possible extension. The best bet (in my opinion) is to fix the conflicting configuration we already have with OMS/LAD and keep an eye on the situation in case any new conflicts pop up.

narrieta: Monitoring conflicting extensions is not feasible. To address the issue raised by OMS/LAD we can simply remove the rule altogether.

kevinclark19a: We've had issues raised for the `CommandExecution.log` file growing too large, so it seems like it would require logrotate. I'm not sure if I see any alternative to some form of this rule.

narrieta: If there is a conflict then we are preventing logs from being rotated so then end result may be the same (using too much disk).

narrieta: As far as the size issue, some things to consider:
* The way the code is written, anything logged in ExtHandlerInstance will be echoed to this file. That is not appropriate, the majority of these messages correspond to the operation of the agent, not of the extension. Cleaning this up would reduce the file size significantly.
* Some extensions don't seem to have this file. Not sure if we have a bug or if extensions are deleting it (I checked custom script and run command)
* I would not mind removing the log altogether, since everything here is already in the agent's log
* If publishers really want this file they could rotate it
* If they can't rotate it and we really want to keep this file we could add special code for this log to truncate it to a small size

kevinclark19a:
C

PR #2147, opened 2021-01-25
Add a binary file for setting up firewall rules on the VM

larohra:
## Description
This is the 1st PR for the change to add firewall rules to the VM on system reboot (since iptables are not persisted on reboots).

This PR deals with - 
- Creating a new binary file for the sole purpose of setting up waagent-network rules
- Separating out the bare minimum needed for setting up the firewall rules

Issue # 

---

### PR information


azurelinuxagent/common/utils/networkutil.py
```diff
@@ -93,3 +96,49 @@ def __str__(self):
...
+    def _add_wait(wait, command):
+        """
+        If 'wait' is True, adds the wait option (-w) to the given iptables command line
+        """
+        if wait:
+            command.insert(1, "-w")
+        return command
+
+    @staticmethod
+    def get_iptables_accept_command(wait, command, destination, owner_uid):
```
larohra, line 113: These 2 commands are all we need to setup the rules on system reboot so just created a new Util just for them

bin/waagent_network_setup.py
```diff
@@ -0,0 +1,70 @@
...
+    waagent.service start. The sole purpose of this file is to setup firewall rules for the VM.
+    
+    The parameters for this file would be set in the unit file for the service, which would be created dynamically the 
+    first time waagent.service starts. On every service restart, we would monitor the Wireserver IP and reset it in the 
+    unit file if it changes.
+    
+    This service would run before the network would be setup, so we need to maintain state of the previous Wireserver IP
+    to ensure we cover all attack vectors to the Wire IP.
+    """
+    _setup_firewall_rules()
```
larohra, line 69: The idea is to create a new service file which would call this binary to setup the network rules on startup. Adding another parameter to our `/sbin/waagent` binary was being a bit painful because that way we would've needed to separate out all dependencies on the VM (like file access, network access, etc) because we plan to run this service way high up in the boot order where these bare minimum dependencies would not be available.

Setting up the new service would come in subsequent PRs

narrieta (approved):

PR #2148, opened 2021-01-25
Command execution log fix

kevinclark19a:
## Description

Added some custom log management to the `CommandExecution.log` file after the `logrotate` config was removed in a prior PR.

Also, has a small fix for a unit test that sometimes failed on python2 envs.

---

### PR information


azurelinuxagent/ga/exthandlers.py
```diff
@@ -783,6 +783,19 @@ def __init__(self, ext_handler, protocol):
             self.logger.error(u"Failed to create extension log dir: {0}", e)
 
         log_file = os.path.join(self.get_log_dir(), "CommandExecution.log")
+
+        try:
+            if os.stat(log_file).st_size > execution_log_max_size:
+                with open(log_file, "rb") as existing_log_file_buffer:
+                    existing_log_file_buffer.seek(-1 * execution_log_max_size, 2)
+                    _ = existing_log_file_buffer.readline()
```
larohra, line 791: Out of curiosity, why is this needed?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -783,6 +783,19 @@ def __init__(self, ext_handler, protocol):
...
+
+                    fileutil.write_file("".join([log_file, ".tmp"]), existing_log_file_buffer.read(), asbin=True)
+                    os.rename("".join([log_file, ".tmp"]), log_file)
+
+        except (IOError, OSError):
+            pass
```
larohra, line 797: Incase of an error should we consider just deleting the file rather than keep on adding to it?

larohra (commented):
some very minor comments

azurelinuxagent/ga/exthandlers.py
kevinclark19a: The purpose of this except line was to catch a `FileNotFound` error thrown by `os.stat`; `log_file` is not necessarily already there, and "better ask forgiveness than permission" is considered pythonic.

There're more file operations within this try block, though, so it might be worth handling those separately...

azurelinuxagent/ga/exthandlers.py
kevinclark19a: Are you referring specifically to L#791 itself?
If so,  it's just an easy way to get the log file at a size less than the max, but on a line ending.

larohra: Ahh very cool, interesting hack haha

azurelinuxagent/ga/exthandlers.py
larohra: Why not just do a blanket catch and delete file if any random error? If considering the `FileNotFound` error also, deleting a file that's not there is just going to be a no-op

larohra: to be clear I only mean delete the file if the size > MAX_SIZE and we throw some error when trying to truncate it

narrieta: ` "better ask forgiveness than permission" is considered pythonic.
`
Sure, but let's ask for forgiveness explicitly, catch FileNotFound and handle it appropriately

azurelinuxagent/ga/exthandlers.py
narrieta, line 791: what was the conclusion wrt utf-8 encoding and truncating a char that has been encoded using more than 1 byte?

narrieta, line 791: readline won't have issues handling a binary file?

azurelinuxagent/ga/exthandlers.py
kevinclark19a: Unfortunately, I believe FileNotFound is python3 only. I originally had that as the exception name.

azurelinuxagent/ga/exthandlers.py
kevinclark19a: It seems we are ascii encoding after all-- @larohra pointed out that the comment I thought referred to encoding as utf-8 was actually talking about the python2 unicode object; later on, we encode that object with ascii. However, `seek` and `readline` should both be utf-8 aware anyway.
Truncating a multi-byte character halfway through would cause file corruption, and is therefore something we really want to avoid. As far as I can tell, this is mainly a problem if `read` (not `readline`) is called on a utf-8 encoded file opened in binary mode.

kevinclark19a: No, as far as I can tell (via testing and posts), it is even utf-8 aware.

azurelinuxagent/ga/exthandlers.py
narrieta: > I'm leaning towards writing a function in future.py that determines if an IOError is from a file not found exception or not.

Sounds like a good idea

azurelinuxagent/ga/exthandlers.py
narrieta: ok

narrieta: ok

azurelinuxagent/ga/exthandlers.py
```diff
@@ -783,6 +783,19 @@ def __init__(self, ext_handler, protocol):
...
 
         log_file = os.path.join(self.get_log_dir(), "CommandExecution.log")
+
+        try:
+            if os.stat(log_file).st_size > execution_log_max_size:
+                with open(log_file, "rb") as existing_log_file_buffer:
+                    existing_log_file_buffer.seek(-1 * execution_log_max_size, 2)
+                    _ = existing_log_file_buffer.readline()
+
+                    fileutil.write_file("".join([log_file, ".tmp"]), existing_log_file_buffer.read(), asbin=True)
```
narrieta, line 793: We shouldn't read the whole file to memory. At this point you can probably use shutil.copyfileobj to copy the rest of the file

kevinclark19a: ah, that works. To be clear, we are only reading the max file size of 10MB into memory at this point, but I think a dedicated copy function is obviously better.

azurelinuxagent/common/future.py
```diff
@@ -128,6 +128,17 @@ def get_openwrt_platform():
...
+def is_file_not_found_error(exception):
+    # Python 2 uses IOError(errno=2)
+    if sys.version_info[0] == 2:
+        return isinstance(exception, IOError) and exception.errno == 2
+    
+    # Python 3 has its own type for this error.
+    elif sys.version_info[0] == 3:
+        return isinstance(exception, FileNotFoundError)
+    
+    else:
```
narrieta, line 145: not sure if we should have this condition here. the module already protects against unknow versions. having a duplicate may introduce issues if one check is updated but not the other

narrieta (commented):
i posted one minor comment( wrt check for py version), otherwise LGTM

narrieta (dismissed):

azurelinuxagent/ga/exthandlers.py
```diff
@@ -768,6 +768,42 @@ def report_ext_handler_status(self, vm_status, ext_handler):
...
+            if is_file_not_found_error(e):
+                # If CommandExecution.log does not exist, it's not noteworthy;
+                # this just means that no extension with self.ext_handler.name is
+                # installed.
+                return
+            
+            logger.error("Exception occurred while attempting to truncate CommandExecution.log "
+                "for extension {0}. Exception is: {1}", self.ext_handler.name, e)
+
+            for f in (filename, filename + ".tmp"):
```
larohra, line 796: Small NIT: Might be worthwhile to make this into a small utility function in fileutil.py so that we can use it in other places too 

Something like - 

> fileutil.rmfile(ignore-errors=True)

I'm good either way though

larohra (dismissed):

narrieta (approved):

PR #2145, opened 2021-01-25
Move agent to azure slice

narrieta:
This PR moves the agent to the azure slice. The slice hierarchy is described in CGroupConfigurator.__ensure_azure_slices_exist.

It also includes some minor changes in telemetry ti distinguish between info and warning messages.

PR #2154, opened 2021-02-01
Setup persistent firewall rules on service restart

larohra:
## Description

This PR ensures that the firewall rules for the agent are persisted over system restarts. This PR does the following - 
- Ensures Firewall rules are persisted over restarts by either using `firewalld.service` or creating a new unit file for doing the same
- If firewalld is available on the box, prioritize that because it already has provisions for making the rules persist over reboots.
- If not, create a new custom agent-network-setup.service which would always setup firewall rules on reboots before network of the VM is even established.

Issue # 

---

### PR information


azurelinuxagent/common/persist_firewall_rules.py
```diff
@@ -0,0 +1,219 @@
...
+
+[Install]
+WantedBy=network.target
+"""
+
+    __OVERRIDE_CONTENT = """
+# This drop-in unit file was created by the Azure VM Agent.
+# Do not edit.
+[Service]
+Environment=
```
larohra, line 58: This line resets the `Environment` content, and the next line overwrites it.

azurelinuxagent/common/osutil/clearlinux.py
```diff
@@ -42,6 +42,14 @@ def __init__(self):
         self.agent_conf_file_path = '/usr/share/defaults/waagent/waagent.conf'
         self.jit_enabled = True
 
+    @staticmethod
+    def get_systemd_unit_file_install_path():
+        return "/usr/lib/systemd/system"
+
+    @staticmethod
+    def get_agent_bin_path():
+        return "/usr/bin"
```
narrieta, line 51: we need to ensure those values are consistent with the values in setup.py (maybe setup should use osutil???)

narrieta: done already, cool

setup.py
```diff
@@ -62,7 +62,7 @@ def set_sysv_files(data_files, dest="/etc/rc.d/init.d", src=None):
     data_files.append((dest, src))
 
 
-def set_systemd_files(data_files, dest="/lib/systemd/system", src=None):
+def set_systemd_files(data_files, dest, src=None):
```
narrieta, line 65: Can you remove the 2 slice files on line 70? #2145 moves them to the correct place, but our PRs will be merged to different releases

larohra: Would it throw an error if we let it be for this release? Like I would prefer if it be bundled with your main PR but if its susceptible to errors then I can remove it now too

azurelinuxagent/common/persist_firewall_rules.py
```diff
@@ -0,0 +1,219 @@
...
+          - Else
+              - Dynamically create the unit file and setup the IPs correctly
+              - Check if /sbin/walinuxagent_setup_service.py exists
+              - If not, then copy it over
+          - Create a drop in conf file in /etc/systemd/system/waagent-network-setup.d/override.conf and override the ExecStart with the new values
+            - Do this everytime on service restart
+        """
+        osutil = get_osutil()
+        self._network_setup_service_name = self._AGENT_NETWORK_SETUP_NAME_FORMAT.format(osutil.get_service_name())
+        self._is_systemd = CGroupsApi.is_systemd()
```
narrieta, line 81: is_systemd probably fits better in osutil (it's weird that the firewall rules have a dependency on cgroups, for example). I'll move it there in a future PR.

azurelinuxagent/common/persist_firewall_rules.py
```diff
@@ -0,0 +1,219 @@
...
+[Unit]
+Description=Setup network rules for WALinuxAgent 
+Before=network-pre.target
+Wants=network-pre.target
+DefaultDependencies=no
+ConditionPathExists={0}
+
+[Service]
+Type=oneshot
+Environment="DST_IP={1}" "UID={2}" "WAIT={3}"
```
narrieta, line 46: Isn't the call to format() on line 195 passing 5 arguments?

Consider moving the definition of this constant closer to its usage; it's hard to jump back and from here to the format() call.

azurelinuxagent/common/persist_firewall_rules.py
narrieta: Maybe add the comment in the unit file itself?

azurelinuxagent/common/persist_firewall_rules.py
larohra: Yup agreed. I can do it but I dont have enough context on how to determine if systemd is present or not in multiple distros (or is the path /run/systemd/system standard for all?)

azurelinuxagent/common/persist_firewall_rules.py
```diff
@@ -0,0 +1,219 @@
...
+
+    _AGENT_NETWORK_SETUP_BIN_FILE = "waagent_network_setup.py"
+    _AGENT_NETWORK_SETUP_NAME_FORMAT = "{0}-network-setup.service"
+    _DROP_IN_ENV_FILE_NAME = "10-environment.conf"
+
+    _FIREWALLD_RUNNING_CMD = ["firewall-cmd", "--state"]
+
+    def __init__(self, dst_ip, uid):
+        """
+        If not, then create the new service file and copy it over to the /lib/systemd/system/ if doesnt exist
```
narrieta, line 70: "If not," incomplete sentence?

azurelinuxagent/common/persist_firewall_rules.py
larohra: Yup this constant takes in 5 args - 
```python
ConditionPathExists={0}
Environment="DST_IP={1}" "UID={2}" "WAIT={3}"
ExecStart={4}
```

>Consider moving the definition of this constant closer to its usage; it's hard to jump back and from here to the format() call.

I mainly try to keep all constants at the top of the class definition to keep a clear separation between the methods and constants. If its a big pain I can move it down I suppose

azurelinuxagent/common/persist_firewall_rules.py
larohra: Sure good point, let me do it

azurelinuxagent/common/persist_firewall_rules.py
larohra: Ohh shoot sorry forgot to re-write this comment. Will do!

azurelinuxagent/common/persist_firewall_rules.py
narrieta: That is the path that systemd itself checks. Didn't check whether it changes across systemd versions.

But leaving aside the implemtation, osutil would be a better place for it. I'll make the change as part of the work I am doing in cgroups.

azurelinuxagent/common/persist_firewall_rules.py
```diff
@@ -0,0 +1,219 @@
...
+        try:
+            return shellutil.run_command(firewalld_state).rstrip() == "running"
+        except Exception as error:
+            logger.info("{0} command returned error/not running: {1}".format(' '.join(firewalld_state), ustr(error)))
+        return False
+
+    def setup(self):
+        if self._is_firewall_service_running():
+            logger.info("Firewalld.service present on the VM, setting up permanent rules on the VM")
+            # Incase of a failure, this would throw. In such a case, we don't need to try to setup our custom service
```
narrieta, line 101: typo: "Incase"

azurelinuxagent/common/persist_firewall_rules.py
```diff
@@ -0,0 +1,219 @@
...
+    def _setup_network_setup_service(self):
+        if self.__verify_network_setup_service_enabled():
+            logger.info("Service: {0} already enabled. No change needed.".format(self._network_setup_service_name))
+        else:
+            # Ensure waagent_network_setup.py is available in the desired spot
+            self.__set_network_setup_bin_file()
+            # Create unit file with default values
+            self.__set_service_unit_file()
+
+        # Even if service is enabled, we need to overwrite the drop-in file with the current IP incase it changed.
```
narrieta, line 159: typpo: "incase"

azurelinuxagent/common/persist_firewall_rules.py
```diff
@@ -0,0 +1,219 @@
...
+        # Check if firewall-cmd can connect to the daemon
+        # https://docs.fedoraproject.org/en-US/Fedora/19/html/Security_Guide/sec-Check_if_firewalld_is_running.html
+        # Eg:    firewall-cmd --state
+        #           >   running
+        firewalld_state = PersistFirewallRulesHandler._FIREWALLD_RUNNING_CMD
+        try:
+            return shellutil.run_command(firewalld_state).rstrip() == "running"
+        except Exception as error:
+            logger.info("{0} command returned error/not running: {1}".format(' '.join(firewalld_state), ustr(error)))
+        return False
```
narrieta, line 96: On success returns a string (stdout) and on failure a boolean. Let's make it consistent

azurelinuxagent/common/persist_firewall_rules.py
```diff
@@ -0,0 +1,219 @@
...
+        try:
+            return shellutil.run_command(cmd).rstrip() == "enabled"
+        except CommandError as error:
+            msg = "{0} not enabled. Command: {1}, ExitCode: {2}\nStdout: {3}\nStderr: {4}".format(
+                self._network_setup_service_name, ' '.join(cmd), error.returncode, error.stdout, error.stderr)
+        except Exception as error:
+            msg = "Ran into error, {0} not enabled. Error: {1}".format(self._network_setup_service_name, ustr(error))
+
+        logger.info(msg)
+        return False
```
narrieta, line 148: inconsistent return value (string vs boolean)

azurelinuxagent/common/persist_firewall_rules.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def __set_network_setup_bin_file(self):
+        if os.path.exists(self._agent_network_setup_bin_file):
+            logger.verbose("{0} file exists in the expected place".format(self._agent_network_setup_bin_file))
+            return
+
+        logger.warn("Network file: {0} not available in the expected path. Copying it over".format(
+            self._agent_network_setup_bin_file))
+        try:
+            parent, _ = os.path.split(os.path.abspath(__file__))
```
narrieta, line 171: what is the value of __file__ at this point wouldn't this code be executing from the agent's egg?

setup.py
narrieta: we would create unit files that are not used by anybody (and they would be in a system directory)

tests/ga/test_update.py
```diff
@@ -1621,6 +1622,25 @@ def test_it_should_recreate_handler_env_on_service_startup(self):
...
+                executed_commands.append(cmd)
+                cmd = ["echo", "running"]
+            return original_popen(cmd, *args, **kwargs)
+
+        with self._get_update_handler(iterations) as (update_handler, _):
+            with patch("azurelinuxagent.common.utils.shellutil.subprocess.Popen", side_effect=_mock_popen):
+                update_handler.run(debug=True)
+
+        # Firewall-cmd should only be called 3 times - 1st to check if running, 2nd & 3rd for the QueryPassThrough cmd
+        self.assertEqual(3, len(executed_commands), "The number of times firwall-cmd should be called is only 3")
```
narrieta, line 1641: maybe check the arguments as well?

tests/common/test_persist_firewall_rules.py
```diff
@@ -0,0 +1,297 @@
...
+        self.__assert_firewall_cmd_running_called(validate_command_called=True)
+        self.__assert_firewall_called(cmd=FirewallCmdDirectCommands.QueryPassThrough, validate_command_called=False)
+        self.__assert_firewall_called(cmd=FirewallCmdDirectCommands.PassThrough, validate_command_called=False)
+
+    def test_it_should_always_replace_only_drop_in_file_if_using_custom_network_service(self):
+        test_str = 'Environment="DST_IP={0}" "UID={1}" "WAIT={2}"'
+        self.__replace_popen_cmd = TestPersistFirewallRulesHandler.__mock_network_setup_service_disabled
+        with self._get_persist_firewall_rules_handler() as handler:
+            self.assertFalse(os.path.exists(self._drop_in_file), "Drop in file should not be there")
+            self.assertFalse(os.path.exists(self._network_service_unit_file), "Unit file not present")
```
narrieta, line 188: The message ("Unit file not present") seems to imply that the file should be present, but we are asserting it is not

narrieta (commented):
Added some minor comments, but LGTM

setup.py
larohra: sounds good, let me remove them then

azurelinuxagent/common/persist_firewall_rules.py
larohra: Fixed

azurelinuxagent/common/persist_firewall_rules.py
larohra: Fixed

azurelinuxagent/common/persist_firewall_rules.py
larohra: Same, its a boolean - 

>return **shellutil.run_command(cmd).rstrip() == "enabled"**

azurelinuxagent/common/persist_firewall_rules.py
larohra: `__file__` points to the current file being executed (in this case the `persist_firewall_rules.py` file). I did a manual test using an egg and it maintains the directory structure

Eg: 
>/var/lib/waagent/WALinuxAgent-2.2.48.1/bin/agent/test.egg/azurelinuxagent/agent.py

I needed the path to the current directory to successfully copy over the new network-setup.py file to the correct `bin` directory of the repo (until atleast the daemon is updated and it ensures that all VMs come up with the new file already present there)

tests/ga/test_update.py
larohra: Good point, let me add that

tests/common/test_persist_firewall_rules.py
larohra: Ohh yes good catch! Let me fix the message.

azurelinuxagent/common/persist_firewall_rules.py
narrieta: thanks, i missed the comparison

azurelinuxagent/common/persist_firewall_rules.py
narrieta: Does shutil.copyfile reach into the egg?

azurelinuxagent/agent.py
```diff
@@ -190,6 +209,17 @@ def collect_logs(self, is_full_mode):
             print("Detailed log output can be found at {0}".format(OUTPUT_RESULTS_FILE_PATH))
             sys.exit(1)
 
+    @staticmethod
+    def setup_firewall(firewall_metadata):
+
+        print("Setting up firewall for the WALinux Agent with args: {0}".format(firewall_metadata))
```
narrieta, line 215: where will the output be captured?

larohra: This would be printed to console. When it would be run from the agent-network-setup.service file, it would be captured by the systemctl logs (journalctl). I've added a function to capture the agent-network-setup.service logs and send it out to Kusto (check the `__log_network_setup_service_logs` function in `persist_firewall_rules.py`)

azurelinuxagent/common/persist_firewall_rules.py
larohra: Changed the approach, this is not relevant anymore

azurelinuxagent/common/persist_firewall_rules.py
```diff
@@ -0,0 +1,235 @@
...
+            # In case of a failure, this would throw. In such a case, we don't need to try to setup our custom service
+            # because on system reboot, all iptable rules are reset by firewalld.service so it would be a no-op.
+            self._setup_permanent_firewalld_rules()
+            return
+
+        logger.info(
+            "Firewalld service not running/unavailable, trying to set up {0}".format(self._network_setup_service_name))
+
+        if not CGroupsApi.is_systemd():
+            raise Exception("Systemd not enabled, unable to set {0}".format(self._network_setup_service_name))
```
narrieta, line 108: Let's rephrase ""Systemd not enabled" to "Did not detect systemd" or something similar

azurelinuxagent/common/persist_firewall_rules.py
```diff
@@ -0,0 +1,235 @@
...
+
+        self._setup_network_setup_service()
+
+    def __verify_firewall_rules_enabled(self):
+        # Check if firewall-rules have already been enabled
+        # This function would also return False if the dest-ip is changed. So no need to check separately for that
+        try:
+            AddFirewallRules.check_firewalld_rule_applied(self._dst_ip, self._uid)
```
narrieta, line 116: How does this check work? raises both if the rules are not enabled or it if fails for some other reason?

azurelinuxagent/agent.py
narrieta: ok

narrieta (commented):
A couple of minor comments, but LGTM

narrieta:
There are a couple of INFO messages in the log that look like errors and may be misleading. Can we rephrase them?

```
>>>>>>>>>>>> 2021-02-05T15:06:39.031001Z INFO ExtHandler ExtHandler firewall-cmd --state command returned error/not running: [Errno 2] No such file or directory: 'firewall-cmd': 'firewall-cmd'**
2021-02-05T15:06:39.031472Z INFO ExtHandler ExtHandler Firewalld service not running/unavailable, trying to set up walinuxagent-network-setup.service
>>>>>>>>>>>2021-02-05T15:06:39.130154Z INFO ExtHandler ExtHandler walinuxagent-network-setup.service not enabled. Command: systemctl is-enabled walinuxagent-network-setup.service, ExitCode: 1
Stdout:
Stderr: Failed to get unit file state for walinuxagent-network-setup.service: No such file or directory

2021-02-05T15:06:39.371742Z INFO ExtHandler ExtHandler Successfully added and enabled the walinuxagent-network-setup.service
2021-02-05T15:06:39.372280Z INFO ExtHandler ExtHandler Drop-in file /lib/systemd/system/walinuxagent-network-setup.service.d/10-environment.conf successfully updated
```

azurelinuxagent/common/persist_firewall_rules.py
```diff
@@ -0,0 +1,235 @@
...
+    __OVERRIDE_CONTENT = """
+# This drop-in unit file was created by the Azure VM Agent.
+# Do not edit.
+[Service]
+# The first line clears the old data, and the 2nd line overwrites it.
+Environment=
+Environment="EGG={egg_path}" "DST_IP={wire_ip}" "UID={user_id}" "WAIT={wait}"
+"""
+
+    _AGENT_NETWORK_SETUP_NAME_FORMAT = "{0}-network-setup.service"
```
narrieta, line 61: "-" is the path separator for cgroups, we should probably use "_" in the file name

azurelinuxagent/common/persist_firewall_rules.py
```diff
@@ -0,0 +1,235 @@
...
+# Do not edit.
+[Unit]
+Description=Setup network rules for WALinuxAgent 
+Before=network-pre.target
+Wants=network-pre.target
+DefaultDependencies=no
+
+[Service]
+Type=oneshot
+Environment="EGG={egg_path}" "DST_IP={wire_ip}" "UID={user_id}" "WAIT={wait}"
```
narrieta, line 44: we should probably not parameterize the uid

azurelinuxagent/common/persist_firewall_rules.py
```diff
@@ -0,0 +1,235 @@
...
+            logger.info("Service: {0} already enabled. No change needed.".format(self._network_setup_service_name))
+            self.__log_network_setup_service_logs()
+
+        else:
+            # Create unit file with default values
+            self.__set_service_unit_file()
+            logger.info("Successfully added and enabled the {0}".format(self._network_setup_service_name))
+
+        # Even if service is enabled, we need to overwrite the drop-in file with the current IP in case it changed.
+        # This is to handle the case where WireIP can change midway on service restarts.
```
narrieta, line 159: Good point. These parameters also need to be overwritten to update the egg's path during agent auto-update. We should probably add that to the comment as well.

azurelinuxagent/common/persist_firewall_rules.py
```diff
@@ -0,0 +1,235 @@
...
+                                                                                            error.stdout, error.stderr)
+            logger.warn(msg)
+        except Exception as error:
+            msg = "Ran into unexpected error when getting logs for {0} service. Error: {1}".format(
+                self._network_setup_service_name, ustr(error))
+            logger.warn(msg)
+
+        add_event(
+            op=WALAEventOperation.PersistFirewallRules,
+            is_success=cmd_success,
```
narrieta, line 233: but this is the status of the journalctl command; we should get the status of the service

azurelinuxagent/ga/update.py
```diff
@@ -829,6 +831,24 @@ def _ensure_extension_telemetry_state_configured_properly(protocol):
         except Exception as e:
             logger.warn("Error when trying to delete existing Extension events directory. Error: {0}".format(ustr(e)))
 
+    @staticmethod
+    def _ensure_firewall_rules_persisted(dst_ip):
+        is_success = False
+        try:
+            PersistFirewallRulesHandler(dst_ip=dst_ip, uid=os.getuid()).setup()
+            msg = "Persistent firewall rules added successfully"
```
narrieta, line 839: can we rephrase this message and the one at 843... they give the impression that the rules were added, when they were actually just setup... maybe add "setup" somewhere in the message?

azurelinuxagent/common/persist_firewall_rules.py
larohra: Done

azurelinuxagent/common/persist_firewall_rules.py
larohra: If the service is working properly, it would fail if the firewall rules for the specified data (IP, UID) are not applied. 

The check to see if `firewalld.service` is enabled/running on the box is different, its done before this check - `_is_firewall_service_running()`

If the `firewalld.service` is not working properly for external reasons, then it would fail too but that would also mostly fail the initial check, which would make us go the other route and setup our custom service.

I hope that answers your question, if not we can sync offline about this

azurelinuxagent/common/persist_firewall_rules.py
larohra: I was initially using `_` but then after doing a simple ls under the systemd directories on the VM realised that all the service names have a `-` instead of an `_` in them. I kept it as a `-` just to be more consistent with the other linux services

azurelinuxagent/common/persist_firewall_rules.py
larohra: Ahh good point, let me add that too.

azurelinuxagent/common/persist_firewall_rules.py
larohra: This is actually the logs of the service (the same logs that you see when we do systemctl status <service>). It would get the verbose logs of the service but would probably not get the status of it (active, failed, etc). I was initially considering that but then I figured it would be more helpful to get the complete logs than the status. Do you think status makes more sense?

azurelinuxagent/ga/update.py
larohra: Technically they were added (but not applied :) ) 
But yeah fair point, rephrased the message

azurelinuxagent/common/persist_firewall_rules.py
larohra: which part? I feel the parameterization helps a lot, makes it a bit more flexible to work with

azurelinuxagent/common/persist_firewall_rules.py
larohra: Btw this is an example of how journalctl logs look like - 

```log
2021-02-05T06:57:44.194001Z INFO ExtHandler ExtHandler Logs from the walinuxagent-network-setup.service since system boot: -- 
Logs begin at Fri 2021-02-05 06:57:36 UTC, end at Fri 2021-02-05 06:57:44 UTC. --
Feb 05 06:57:38 edpwz5814j python3[441]: Setting up firewall for the WALinux Agent with args: {'wait': '-w', 'uid': '0', 'dst_ip': '168.63.129.16'}
Feb 05 06:57:38 edpwz5814j systemd[1]: Started Setup network rules for WALinuxAgent.
```

azurelinuxagent/common/persist_firewall_rules.py
narrieta: ok

azurelinuxagent/common/persist_firewall_rules.py
narrieta: we are logging as verbose, so we won't see the messages under normal runs. seems that we would like to know about some of those, e.g. "If the service is working properly, it would fail if the firewall rules for the specified data (IP, UID) are not applied."

azurelinuxagent/common/persist_firewall_rules.py
narrieta: I meant the is_success we send in telemetry. Probably we do not care about failures executing journalctl, but we do care if the firewall rules were not created

larohra: Ahh good point, let me also query the service to see if it was in a failed state or not and send the is-success based on that

azurelinuxagent/common/persist_firewall_rules.py
larohra: In such a case it would still print a generic message like - "Firewall rules not added yet, adding them now using firewalld.service" or "Service: {0} not enabled. Adding it now". 

If you want more data then make it verbose. Initially I added the metadata of these commands as info too (stdout/stderr) but that definitely has a potential of being a bit confusing to the untrained eye so figured I'd make it verbose. Do you have a simpler alternative in mind?

narrieta (approved):

PR #2155, opened 2021-02-02
Refresh host ga plugin in periodic task

narrieta:
Refreshing the goal state for the Host GA Plugin doesn't need to be done in the main method of the monitor thread. Moved it to the periodic task for the plugin.

### DIFF
--- a/azurelinuxagent/ga/monitor.py
+++ b/azurelinuxagent/ga/monitor.py
@@ -144,6 +144,8 @@ class SendHostPluginHeartbeatOperation(PeriodicOperation):
         try:
             host_plugin = self.protocol.client.get_host_plugin()
             host_plugin.ensure_initialized()
+            self.protocol.update_host_plugin_from_goal_state()
+
             is_currently_healthy = host_plugin.get_health()
 
             if is_currently_healthy:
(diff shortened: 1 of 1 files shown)


azurelinuxagent/ga/monitor.py
```diff
@@ -144,6 +144,8 @@ def _operation_impl(self):
         try:
             host_plugin = self.protocol.client.get_host_plugin()
             host_plugin.ensure_initialized()
+            self.protocol.update_host_plugin_from_goal_state()
```
larohra, line 147: Shouldn't this be before Line145?

narrieta: get_host_plugin instantiates (and caches) the plugin; the update needs to happen after that

if the plugin is not instantiated the update is a no-op, btw

narrieta: (get_host_plugin also  retrieves the goal state, so the update is not needed during instantiation; given that the order of get/update doesn't really matter, but the code reads better as "get the plugin, then update it")

larohra: Ahh ok. My main concern was basically updating a different reference than what we were using but I think that would not be a problem here since Python passes it by reference

larohra (approved):

azurelinuxagent/ga/monitor.py
narrieta: I see. The host_plugin is an instance member of the protocol (the protocol client). we are using the same protocol instance so we're good.

PR #2156, opened 2021-02-03
Add Swtich to disable NetworkConfigurationChanges

ZhidongPeng:
## Description

Issue # 

---

### PR information


tests/ga/test_monitor.py
```diff
@@ -58,28 +58,34 @@ def periodic_operation_run(self):
         invoked_operations = []
 
         with _mock_wire_protocol():
-            with patch("azurelinuxagent.ga.monitor.MonitorHandler.stopped", side_effect=[False, True]):
+            with patch("azurelinuxagent.ga.monitor.MonitorHandler.stopped", side_effect=[False, True, False, True]):
```
narrieta, line 61: I think this should be [ False, False, True ] (i.e. do 2 iterations of the main loop and then stop)

tests/ga/test_monitor.py
```diff
@@ -58,28 +58,34 @@ def periodic_operation_run(self):
...
+                                expected_operations = [
+                                    'poll resource usage',
+                                    'report network errors',
+                                    'reset periodic log messages',
+                                    'send_host_plugin_heartbeat',
+                                    'send_imds_heartbeat'
+                                ]
 
-                        self.assertEqual(invoked_operations, expected_operations, "The monitor thread did not invoke the expected operations")
+                                if monitor_network_changes():
```
narrieta, line 82: Invoking the mock here doesn't feel quite right, I'd suggest using the loop variable instead (network_changes)

narrieta (commented):
2 small comments in the test code, otherwise LGTM

tests/ga/test_monitor.py
ZhidongPeng: yes, we need run it with 2 iterations, while it requires both **run and join** at each iteration. so it has to be True to allow the monitor thread only run once per iteration. if I set to False, False, True, the invoked_operations get doubled and it only runs for the first iteration.

tests/ga/test_monitor.py
ZhidongPeng: make sense, I updated it.

tests/ga/test_monitor.py
narrieta: yup, you are correct

narrieta (approved):

PR #2159, opened 2021-02-05
Suse12 hostname fix

kevinclark19a:
## Description

Potential fix for suse12 DCR test failure on PR #1951.

---

### PR information

### DIFF
--- a/azurelinuxagent/common/future.py
+++ b/azurelinuxagent/common/future.py
@@ -137,7 +137,7 @@ def is_file_not_found_error(exception):
 
     if sys.version_info[0] == 2:
         # Python 2 uses OSError(errno=2)
-        return isinstance(exception, OSError) and exception.errno == 2
+        return isinstance(exception, (IOError, OSError)) and exception.errno == 2
     elif sys.version_info[0] == 3:
         return isinstance(exception, FileNotFoundError)
     
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -42,7 +42,7 @@ import azurelinuxagent.common.utils.shellutil as shellutil
 import azurelinuxagent.common.utils.textutil as textutil
 
 from azurelinuxagent.common.exception import OSUtilError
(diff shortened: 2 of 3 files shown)


azurelinuxagent/common/future.py
```diff
@@ -137,7 +137,7 @@ def is_file_not_found_error(exception):
 
     if sys.version_info[0] == 2:
         # Python 2 uses OSError(errno=2)
-        return isinstance(exception, OSError) and exception.errno == 2
+        return isinstance(exception, (IOError, OSError)) and exception.errno == 2
```
kevinclark19a, line 140: Some weirdness here; `IOError` was originally used, but that fails on our end-to-end testing for, e.g., centos 7.8 images. `OSError` seemed to fix this failure, until the change in `default.py`, using this function.

The python documentation mentions that the error should be `IOError` for python 2, but we need both here.

azurelinuxagent/common/osutil/default.py
```diff
@@ -1287,16 +1287,17 @@ def set_hostname_record(self, hostname):
...
-        record = fileutil.read_file(hostname_record)
-        return record
+            return fileutil.read_file(hostname_record)
+        except (IOError, OSError) as e:
+            if is_file_not_found_error(e):
+                logger.info('Hostname record does not exist.')
+            else:
+                logger.error("Exception reading hostname record: {0}", e)
+            
+            return None
```
kevinclark19a, line 1300: This is the main part of the fix; `ga.env:EnvHandler.run()` uses this function to populate `self.hostname` (L#99 for this branch), but this function only stores the hostname in the filesystem, it doesn't call `publish_hostname`. 

The extension handler only ever calls `publish_hostname` in `ga.env:EnvHandler._monitor_hostname_changes()`, gated behind a check for any hostname changes. Returning the hostname here therefore means that the agent never calls `publish_hostname` in scenarios where `cloud-init` is used.

azurelinuxagent/common/future.py
narrieta: could you add a unit test that illustrates this?

PR #2162, opened 2021-02-07
Refactoring the monitor and env threads to use PeriodicOperations

narrieta:


azurelinuxagent/ga/collect_logs.py
```diff
@@ -85,10 +84,7 @@ def __init__(self):
         self.event_thread = None
         self.should_run = True
         self.last_state = None
-
```
narrieta, line 88: There was no need to use PeriodicOperations in the log collector thread. This was probably copy and pasted from the Monitor thread.

azurelinuxagent/ga/collect_telemetry_events.py
```diff
@@ -68,7 +68,7 @@ class ExtensionEventSchema(object):
     OperationId = "OperationId"
 
 
-class _ProcessExtensionEventsPeriodicOperation(PeriodicOperation):
```
narrieta, line 71: I removed the "PeriodicOperation" and "Operation" suffixes across all the usages of PeriodicOperation to improve readability

azurelinuxagent/ga/periodic_operation.py
```diff
@@ -33,9 +33,8 @@ class PeriodicOperation(object):
     # To prevent flooding the log with error messages we report failures at most every hour
     _LOG_WARNING_PERIOD = datetime.timedelta(minutes=60)
 
-    def __init__(self, name, operation, period):
-        self._name = name
-        self._operation = operation
+    def __init__(self, period):
```
narrieta, line 36: Now that all the code is refactored to use subclasses of PeriodicOperation, there is no need of a string to identify the operation (we can use the type) nor the need to pass the function to execute as parameter (this is now done using inheritance)

larohra (approved):
Thanks for the refactor!

azurelinuxagent/ga/collect_logs.py
narrieta: The thread blocks waiting for the log collector. If we did this on one of the other threads we may end up blocking all the operations in that thread while the log collector runs. With a little more of extra code we could just start the collector and not block on it, similarly to what we do when we start extensions. I think the code was written like this to make it a little simpler. We can always improve on it later.

PR #2163, opened 2021-02-08
Refactor mock_cgroup_commands

narrieta:
Split mock_cgroup_commands into MockEnvironment (which is generic) and mock_cgroup_environment (which mocks commands and files specific to cgroups)


tests/common/mock_cgroup_commands.py
```diff
@@ -1,206 +0,0 @@
-# -*- coding: utf-8 -*-
```
narrieta, line 1: This file was renamed to mock_cgroup_environment.py. In the _MOCKED* arrays stay the same in the new file and the mocking logic is now moved to mock_environment.py

tests/common/mock_environment.py
```diff
@@ -0,0 +1,149 @@
...
+import os
+import re
+import shutil
+import subprocess
+
+from azurelinuxagent.common.utils import fileutil
+from tests.tools import patch, patch_builtin
+
+
+class MockEnvironment:
```
narrieta, line 27: As new functionality was added, the mocking logic was getting too long to be defined in a single function so I decided to make this a proper class instead.

tests/common/mock_environment.py
```diff
@@ -0,0 +1,149 @@
...
+        self._original_open = open
+
+        self.patchers = [
+            patch_builtin("open", side_effect=self._mock_open),
+            patch("subprocess.Popen", side_effect=self._mock_popen),
+            patch("os.path.exists", side_effect=self._mock_path_exists),
+            patch("azurelinuxagent.common.utils.fileutil.mkdir", side_effect=self._mock_mkdir)
+        ]
+
+    def __enter__(self):
```
narrieta, line 76: Previously the contextmanager contract was implemented using the contextlib.contextmanager decorator; now it is explicitly implemented using __enter__ and __exit__.

tests/common/mock_cgroup_environment.py
```diff
@@ -0,0 +1,106 @@
+# -*- coding: utf-8 -*-
```
narrieta, line 1: This file was renamed from mock_cgroup_environment.py. The _MOCKED* arrays stay the same in the new file and the mocking logic is now moved to mock_environment.py

larohra (approved):
Awesome thanks for adding this functionality, I'm sure it'll be super helpful in the future!

PR #2173, opened 2021-02-10
Monitor agent's throttle time

narrieta:


azurelinuxagent/common/cgroup.py
```diff
@@ -38,39 +38,25 @@ class MetricsCounter(object):
...
 re_user_system_times = re.compile(r'user (\d+)\nsystem (\d+)\n')
 
 
-class CGroupContollers(object):
-    CPU = "cpu"
-    MEMORY = "memory"
-
-
 class CGroup(object):
-    @staticmethod
```
narrieta, line 52: This create method was weird: the caller needs to pass the type of the cgroup. Probably it was done to keep track of the controller (for use when formatting to string), but that would not be needed since the path includes it.

azurelinuxagent/common/cgroup.py
```diff
@@ -38,39 +38,25 @@ class MetricsCounter(object):
...
         :param: name: Name of the CGroup
         :param: cgroup_path: Path of the controller
-        :param: controller_type:
         :return:
         """
         self.name = name
         self.path = cgroup_path
-        self.controller = controller_type
+
+    def __str__(self):
```
narrieta, line 58: moved this method to the base class and simplified the format

azurelinuxagent/common/cgroup.py
```diff
@@ -161,24 +145,39 @@ def _get_cpu_ticks(self, allow_no_such_file_or_directory_error=False):
...
+                raise CGroupsException("The contents of {0} are invalid: {1}".format(self._get_cgroup_file('cpuacct.stat'), cpuacct_stat))
             cpu_ticks = int(match.groups()[0]) + int(match.groups()[1])
 
         return cpu_ticks
 
+    def _get_throttled_time(self):
+        try:
+            with open(os.path.join(self.path, 'cpu.stat')) as cpu_stat:
+                for line in cpu_stat:
+                    match = re.match(r'throttled_time\s+(\d+)', line)
```
larohra, line 170: Could you add a sample string of how the date looks? Also a command maybe on how to view it on a VM? THis is primarily for debugging purposes

azurelinuxagent/common/cgroup.py
```diff
@@ -161,24 +145,39 @@ def _get_cpu_ticks(self, allow_no_such_file_or_directory_error=False):
...
+                    if match is not None:
+                        return int(match.groups()[0])
+                raise Exception("Cannot find throttled_time")
+        except (IOError, OSError) as e:
+            if e.errno == errno.ENOENT:
+                return 0
```
larohra, line 176: I see that you're returning 0 if FileNotFound. Wouldn't this be misconstrued as the agent was not throttled at all?

azurelinuxagent/common/cgroup.py
```diff
@@ -213,26 +214,29 @@ def get_cpu_usage(self):
...
+        Computes the throttled time (in nanoseconds) since the last call to this function.
+        NOTE: initialize_cpu_usage() must be invoked before calling this function
+        """
+        if not self._cpu_usage_initialized():
+            raise CGroupsException("initialize_cpu_usage() must be invoked before the first call to get_throttled_time()")
+
+        self._previous_throttled_time = self._current_throttled_time
+        self._current_throttled_time = self._get_throttled_time()
+
+        return  self._current_throttled_time - self._previous_throttled_time
```
larohra, line 228: Is this always guaranteed to be incrementing? Like at current would always be > previous? 
Incase of filenotfound, current would be 0. Is there a possibility that this file can be deleted somehow externally?

azurelinuxagent/common/cgroup.py
```diff
@@ -139,19 +125,17 @@ def get_tracked_metrics(self):
 
 
 class CpuCgroup(CGroup):
-    def __init__(self, name, cgroup_path):
-        super(CpuCgroup, self).__init__(name, cgroup_path, CGroupContollers.CPU)
+    def __init__(self, name, cgroup_path, track_throttled_time=False):
```
larohra, line 128: Out of curiosity, when will `track_throttled_time` be False? Like what's the use case of that?

larohra (commented):
some minor comments else LGTM

azurelinuxagent/common/cgroup.py
narrieta: yup, will do

azurelinuxagent/common/cgroup.py
narrieta: yes, it is incremental from the creation of the cgroup

azurelinuxagent/common/cgroup.py
narrieta: the file wouldn't exist if the cpu controller wasn't active, and then there would not be any throttling

azurelinuxagent/common/cgroup.py
narrieta: i will hook this up to the enable/disable feature... if quota is disabled then there is no need to send this metric

narrieta: (if quota is disabled we don't need this metric, but we would still be monitoring cpu usage)

azurelinuxagent/common/cgroup.py
narrieta: oh, just noted your question about deletion

cgroups it's a "pseudo" file system. can't delete the file (there might be a way to do it, but straight rm doesn't do it):

root@nam-u18:/var/lib/waagent# rm /sys/fs/cgroup/cpuacct/azure.slice/cpu.stat 
rm: cannot remove '/sys/fs/cgroup/cpuacct/azure.slice/cpu.stat': Operation not permitted

azurelinuxagent/common/cgroup.py
narrieta: done

azurelinuxagent/common/cgroup.py
narrieta: also, this won't happen on the agent, but when an extension completes the cgroup and associated files go away after a short time. we should probably detect that case and just don't send telemetry (will require some kind of communication within the extension handler and the monitor thread, though)

larohra: Ahh makes sense

azurelinuxagent/common/cgroup.py
larohra: woah! That is interesting!

azurelinuxagent/common/cgroup.py
larohra: Yeah makes sense, without quota the agent wouldn't be throttled at all right? Or can it still be? Any idea?

larohra (approved):

azurelinuxagent/common/cgroup.py
narrieta: maybe we can just let the not found error bubble up to the monitor thread and check there if the cgroup still exists and cleanup there. we already have some cleanup logic that i wanted to hook up to cgroup notifications... i'll make a note, all of that would be for extensions

azurelinuxagent/common/cgroup.py
narrieta: the quota/throttling are handled by the same controller: cpu. the cpu usage is done by a different controller: cpuacct
there can be cpu usage monitoring without quota, but throttling would be done only if a quota is set

larohra: I got that. My main question was if there is any possibility of the agent being throttled even if there is no quota? Its more a systemd question than anything to do with this PR

narrieta: i see... the scheduler tries to allocate cpu for all processes, but it also uses other factors to  decide who gets the cpu (e.g. the nice level) so I guess that could also be seen as a form of "throttling"

PR #2176, opened 2021-02-11
Set CPU quota only if cgroups are enabled

narrieta:


azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -260,26 +263,22 @@ def __setup_unit_files(self):
...
-            systemd_unit_file_install_path = systemd.get_unit_file_install_path()
-            agent_unit_file_dropin_path = os.path.join(systemd_unit_file_install_path, "{0}.d".format(self._cgroups_api.get_agent_unit_name()))
-            azure_slice = os.path.join(systemd_unit_file_install_path, _AZURE_SLICE)
-            vmextensions_slice = os.path.join(systemd_unit_file_install_path, _VMEXTENSIONS_SLICE)
-            agent_unit_file = os.path.join(systemd_unit_file_install_path, self._cgroups_api.get_agent_unit_name())
-            agent_drop_in_file_slice = os.path.join(agent_unit_file_dropin_path, _AGENT_DROP_IN_FILE_SLICE)
-            agent_drop_in_file_cpu_accounting = os.path.join(agent_unit_file_dropin_path, _AGENT_DROP_IN_FILE_CPU_ACCOUNTING)
-            agent_drop_in_file_cpu_quota = os.path.join(agent_unit_file_dropin_path, _AGENT_DROP_IN_FILE_CPU_QUOTA)
-            agent_unit_name = self._cgroups_api.get_agent_unit_name()
+            unit_file_install_path = systemd.get_unit_file_install_path()
```
narrieta, line 275: we don't create _AGENT_DROP_IN_FILE_CPU_ACCOUNTING here anymore

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -113,21 +115,22 @@ def initialize(self):
...
+                self.__setup_azure_slice()
 
                 cpu_controller_root, memory_controller_root = self.__get_cgroup_controllers()
                 self._agent_cpu_cgroup_path, self._agent_memory_cgroup_path = self.__get_agent_cgroups(agent_slice, cpu_controller_root, memory_controller_root)
 
                 if self._agent_cpu_cgroup_path is not None:
                     self.__log_cgroup_info("Agent CPU cgroup: {0}", self._agent_cpu_cgroup_path)
-                    CGroupsTelemetry.track_cgroup(CpuCgroup(agent_unit_name, self._agent_cpu_cgroup_path, track_throttled_time=True))
                     self._cgroups_enabled = True
+                    track_throttled_time = self.set_cpu_quota(_AGENT_CPU_QUOTA)
```
narrieta, line 132: now we create the quota unit file here, after we know cgroups are enabled

azurelinuxagent/common/cgroupapi.py
```diff
@@ -228,11 +227,6 @@ def get_cgroup2_controllers():
                 return mount_point, controllers
         return None, None
 
-    def get_agent_unit_name(self):
```
narrieta, line 231: i refactored more functions from cgroupapy to the systemd module

tests/common/mock_environment.py
```diff
@@ -20,10 +20,19 @@
 import shutil
 import subprocess
 
+from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.utils import fileutil
 from tests.tools import patch, patch_builtin
 
 
+class MockCommand:
```
narrieta, line 28: I am adding negative tests, so need failures when mocking commands.

I created a new class for mock commands (it used to be tuple <command>,<stdout>) to include the return value and stderr

azurelinuxagent/common/cgroupconfigurator.py
larohra: Aren't accounting and Quota related? Or is Accounting more for Monitoring and Quota for Enforcement?

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -392,19 +380,48 @@ def __get_agent_cgroups(self, agent_slice, cpu_controller_root, memory_controlle
...
         def disable(self):
             self._cgroups_enabled = False
             CGroupsTelemetry.reset()
 
+        def set_cpu_quota(self, quota):
+            """
+            Sets the agent's CPU quota to the given percentage (100% == 1 CPU).
+
+            This is done using a dropin file in the default dropin directory; any local overrides on the VM will take precedence
+            over this setting.
```
larohra, line 397: Add the return type of the function in the doc? (i.e. True if success else False)

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -392,19 +380,48 @@ def __get_agent_cgroups(self, agent_slice, cpu_controller_root, memory_controlle
...
+            """
+            logger.info("Setting agent's CPUQuota to {0}%", quota)
+            return self._set_cpu_quota(quota)
+
+        def reset_cpu_quota(self):
+            """
+            Removes any CPUQuota on the agent
```
larohra, line 404: Same

tests/common/mock_command.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python3
+import os
+import sys
+
+if len(sys.argv) != 4:
```
larohra, line 5: This check would imply <4 is also a good number. Maybe change the condition to `len(sys.argv) < 4:` instead?

larohra (commented):
Minor comments else LGTM

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -62,12 +62,14 @@
...
 _AGENT_DROP_IN_FILE_CPU_QUOTA = "12-CPUQuota.conf"
-_AGENT_DROP_IN_FILE_CPU_QUOTA_CONTENTS = """
+_AGENT_DROP_IN_FILE_CPU_QUOTA_CONTENTS_FORMAT = """
 # This drop-in unit file was created by the Azure VM Agent.
 # Do not edit.
 [Service]
-CPUQuota=5%
+CPUQuota={0}%
 """
+_AGENT_CPU_QUOTA = 5
```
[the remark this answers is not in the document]
narrieta: yes, before shipping. for testing i want to try 5%

azurelinuxagent/common/cgroupconfigurator.py
narrieta: they are different controllers (cpuacct vs cpu). quota is handled by the cpu controller. usage stats by cpuacct

cpu requires cpuacct (but not the other way around)

azurelinuxagent/common/cgroupconfigurator.py
narrieta: will do

tests/common/mock_command.py
narrieta: it requires exactly 4 (stdout, return value & stderr + the command itself)

azurelinuxagent/common/cgroupconfigurator.py
narrieta: updated

azurelinuxagent/common/cgroupconfigurator.py
narrieta: updated

tests/common/mock_command.py
narrieta: btw since the script is not meant to be used manually (at least not frequently) i did not put any effort on setting default values for the parameters

azurelinuxagent/common/cgroupconfigurator.py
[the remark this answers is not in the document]
narrieta: should have mentioned that i have a task to add these settings to the conf file. so eventually this will come from the conf file and the tests can change it as they need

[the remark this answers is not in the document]
larohra: Ahh ok fair enough

tests/common/mock_command.py
larohra: Ohh ok makes sense

larohra (approved):

PR #2177, opened 2021-02-13
Disable cgroups when the agent Throttled Time exceeds threshold

narrieta:


azurelinuxagent/common/cgroup.py
```diff
@@ -117,23 +118,22 @@ def is_active(self):
...
-    def get_tracked_metrics(self):
+    def get_tracked_metrics(self, **_):
         """
         Retrieves the current value of the metrics tracked for this cgroup and returns them as an array
         """
         raise NotImplementedError()
 
 
 class CpuCgroup(CGroup):
-    def __init__(self, name, cgroup_path, track_throttled_time=False):
```
narrieta, line 128: There is no need to maintain this flag for each instance of the CpuCgroup currently tracked need to use the same value. Moved the flag to CGroupsConfigurator.

azurelinuxagent/common/cgroup.py
```diff
@@ -230,7 +229,7 @@ def get_cpu_usage(self):
 
     def get_throttled_time(self):
         """
-        Computes the throttled time (in nanoseconds) since the last call to this function.
+        Computes the throttled time (in seconds) since the last call to this function.
```
narrieta, line 232: seconds are more readable than nanoseconds

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -69,6 +69,19 @@
 CPUQuota={0}%
 """
 _AGENT_CPU_QUOTA = 5
+_AGENT_THROTTLED_TIME_THRESHOLD = 120  # 2 minutes
+
+
```
narrieta, line 74: The CGroupsConfigurator instance is shared by the extension handler and monitor threads (monitor invokes the checks on cgroups). As I add more logic to those checks, I need to be more strict with the internal state of the configurator. I made several methods static.

These 2 loggin methods used to be instance methods; making them static would require prefixing each invocation with "CGroupsConfigurator._Impl", which hurts readability so I made them global (but internal) to the module.

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -78,16 +91,14 @@ class CGroupConfigurator(object):
     NOTE: with the exception of start_extension_command, none of the methods in this class
     raise exceptions (cgroup operations should not block extensions)
     """
-    class __Impl(object):
+    class _Impl(object):
```
narrieta, line 94: I need access to some of the non-public members in the unit tests so removed 1 underscore from the name to avoid name mangling.

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -377,69 +381,97 @@ def __get_agent_cgroups(self, agent_slice, cpu_controller_root, memory_controlle
...
             commands used to start extensions on their own cgroup).
             Other processes started by the agent (e.g. extensions) and processes not started by the agent (e.g. services installed by extensions) are reported
             as unexpected, since they should belong to their own cgroup.
-            """
-            if not self.enabled():
-                return True
 
-            def log_message(message):
-                # Report only a small sample of errors
-                if message != self._check_processes_in_agent_cgroup_last_error and self._check_processes_in_agent_cgroup_error_count < 5:
```
narrieta, line 438: We were keeping track of previous errors to avoid being too verbose when logging/sending telemetry. Since now we disable cgroups when this check fails, there is no need to keep track of them.

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -493,6 +520,13 @@ def get_command_line(pid):
 
             return [get_command_line(pid) for pid in pid_list]
 
+        @staticmethod
+        def _check_agent_throttled_time(cgroup_metrics):
```
narrieta, line 524: This is just a bare-bones implementation. As i work on the end-to-end tests I'll adjust the logic accordingly

azurelinuxagent/common/resourceusage.py
```diff
@@ -1,140 +0,0 @@
-# Copyright 2019 Microsoft Corporation
```
narrieta, line 1: Ununsed module

tests/common/mock_environment.py
```diff
@@ -32,6 +32,9 @@ def __init__(self, command, stdout='', return_value=0, stderr=''):
         self.return_value = return_value
         self.stderr = stderr
 
+    def __str__(self):
```
narrieta, line 35: Makes debugging easier

tests/common/test_cgroupconfigurator.py
```diff
@@ -48,7 +48,7 @@ def tearDownClass(cls):
         AgentTestCase.tearDownClass()
 
     @contextlib.contextmanager
-    def _get_cgroup_configurator(self, initialize=True, mock_commands=None):
+    def _get_cgroup_configurator(self, initialize=True, enable=True, mock_commands=None):
```
narrieta, line 51: Added an option to call initialize without enabling cgroups

tests/common/test_cgroupconfigurator.py
```diff
@@ -365,46 +358,34 @@ def mock_popen(command, *args, **kwargs):
...
         with self._get_cgroup_configurator() as configurator:
-            pass  # release the mocks used to create the test CGroupConfigurator so that they do not conflict the mock Popen below
-
-        # Systemd has its own internal timeout which is shorter than what we define for extension operation timeout.
-        # When systemd times out, it will write a message to stderr and exit with exit code 1.
-        # In that case, we will internally recognize the failure due to the non-zero exit code, not as a timeout.
-        original_popen = subprocess.Popen
-        systemd_timeout_command = "echo 'Failed to start transient scope unit: Connection timed out' >&2 && exit 1"
-
-        def mock_popen(command, *args, **kwargs):
```
narrieta, line 376: Changed the test to use add_command instead of creating its own mock

tests/common/test_cgroupconfigurator.py
```diff
@@ -656,41 +637,69 @@ def get_completed_process():
...
-            def get_telemetry_event_messages():
-                return [kwargs["message"] for (_, kwargs) in add_event_patcher.call_args_list if "The agent's cgroup includes unexpected processes" in kwargs["message"]]
-
             agent_processes = [os.getppid(), os.getpid()] + agent_command_processes + [start_extension.systemd_run_pid]
             other_processes = [1, get_completed_process()] + extension_processes
 
             with patch("azurelinuxagent.common.cgroupconfigurator.CGroupsApi.get_processes_in_cgroup", return_value=agent_processes + other_processes):
-                with patch("azurelinuxagent.common.cgroupconfigurator.add_event") as add_event_patcher:
-                    cgroup_configurator = CGroupConfigurator.get_instance()
+                with self.assertRaises(CGroupsException) as context_manager:
```
narrieta, line 644: Now _check_processes_in_agent_cgroup raises an exception and the caller sends the telemetry

tests/common/test_cgroupstelemetry.py
```diff
@@ -196,12 +195,10 @@ def test_telemetry_polling_with_changing_cgroups_state(self, patch_get_statm, pa
         current_cpu = 30
         current_memory = 209715200
         current_max_memory = 471859200
-        current_proc_statm = 20000000
```
narrieta, line 199: These were defined in the module I removed.

tests/ga/test_extension.py
```diff
@@ -346,24 +346,10 @@ def test_migration_ignores_tree_remove_errors(self, shutil_mock):  # pylint: dis
         return
 
 
-class ExtensionTestCase(AgentTestCase):
```
narrieta, line 349: These tests were calling CGroupConfigurator.disable(). There is no need, since the configurator is disabled by default.

tests/ga/test_exthandlers.py
```diff
@@ -336,15 +335,7 @@ def setUp(self):
         self.mock_sleep = patch("time.sleep", lambda *_: mock_sleep(0.01))
         self.mock_sleep.start()
 
-        self.cgroups_enabled = CGroupConfigurator.get_instance().enabled()
```
narrieta, line 339: No need to call disable

larohra (approved):
1 minor comment else LGTM

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -377,69 +381,97 @@ def __get_agent_cgroups(self, agent_slice, cpu_controller_root, memory_controlle
...
             """
-            Removes any CPUQuota on the agent and returns True on success, False otherwise
+            Removes any CPUQuota on the agent
+
+            NOTE: This resets the quota on the agent's default dropin file; any local overrides on the VM will take precedence
+            over this setting.
             """
             logger.info("Resetting agent's CPUQuota")
-            return self._set_cpu_quota('')  # setting an empty value resets to the default (infinity)
+            if CGroupConfigurator._Impl.__try_set_cpu_quota(''):  # setting an empty value resets to the default (infinity)
```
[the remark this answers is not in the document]
narrieta: Good catch, let me fix that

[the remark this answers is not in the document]
narrieta: Fixed. Updated the unit tests to actually check the quota

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -409,8 +409,9 @@ def __set_cpu_quota(quota):
             NOTE: This is done using a dropin file in the default dropin directory; any local overrides on the VM will take precedence
             over this setting.
             """
-            logger.info("Setting agent's CPUQuota to {0}%", quota)
-            if CGroupConfigurator._Impl.__try_set_cpu_quota(quota):
+            quota_percentage = "{0}%".format(quota)
+            logger.info("Setting agent's CPUQuota to {0}", quota_percentage)
```
larohra, line 413: `_log_cgroup_info` instead maybe (for the Kusto log)?

larohra (dismissed):

azurelinuxagent/common/cgroupconfigurator.py
narrieta: sounds good; i also did _log_cgroup_warning when we fail to change the quota

larohra (approved):

PR #2188, opened 2021-03-01
Additional locations

kevinclark19a:
## Description

This PR adds support for an `additionalLocations` node in the extension conf xml schema.

---

### PR information


azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -398,7 +398,12 @@ def _parse_plugin(ext_handler, plugin):
         Sample config:
 
         <Plugins>
-              <Plugin name="Microsoft.CPlat.Core.NullSeqB" version="2.0.1" location="https://zrdfepirv2cbn04prdstr01a.blob.core.windows.net/f72653efd9e349ed9842c8b99e4c1712/Microsoft.CPlat.Core_NullSeqB_useast2euap_manifest.xml" state="enabled" autoUpgrade="false" failoverlocation="https://zrdfepirv2cbz06prdstr01a.blob.core.windows.net/f72653efd9e349ed9842c8b99e4c1712/Microsoft.CPlat.Core_NullSeqB_useast2euap_manifest.xml" runAsStartupTask="false" isJson="true" useExactVersion="true" />
+              <Plugin name="Microsoft.CPlat.Core.NullSeqB" version="2.0.1" location="https://zrdfepirv2cbn04prdstr01a.blob.core.windows.net/f72653efd9e349ed9842c8b99e4c1712/Microsoft.CPlat.Core_NullSeqB_useast2euap_manifest.xml" state="enabled" autoUpgrade="false" failoverlocation="https://zrdfepirv2cbz06prdstr01a.blob.core.windows.net/f72653efd9e349ed9842c8b99e4c1712/Microsoft.CPlat.Core_NullSeqB_useast2euap_manifest.xml" runAsStartupTask="false" isJson="true" useExactVersion="true">
+                <additionalLocation>
```
larohra, line 402: Can you add a bit more context about this please? Like what is this used for, why is it required? When and when not will it be provided etc etc? 
Thanks!

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -398,7 +398,12 @@ def _parse_plugin(ext_handler, plugin):
...
+                    <additionalLocation>http://www.publisher.type.failover1.com/</additionalLocation>
+                    <additionalLocation>http://www.publisher.type.failover2.com/</additionalLocation>
```
narrieta, line 404: let's use a more realistic sample (similar to the samples above: https://zrdfepirv2cbz06prdstr01a.blob.core.windows.net/f72653efd9e349ed9842c8b99e4c1712/Microsoft.CPlat.Core_NullSeqB_useast2euap_manifest.xml) to give more context of what those uris mean

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -421,9 +426,19 @@ def _log_error_if_none(attr_name, value):
         if ext_handler.properties.state in (None, ""):
             raise ExtensionConfigError("Received empty Extensions.Plugins.Plugin.state, failing Handler")
 
-        location = getattrib(plugin, "location")
-        failover_location = getattrib(plugin, "failoverlocation")
-        for uri in [location, failover_location]:
+        def getattrib_wrapped_in_list(node, attr_name):
+            attr = getattrib(node, attr_name)
+            return [attr] if attr else []
```
narrieta, line 431: is "if attr" a comparison against None? can we make it explicit?

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -421,9 +426,19 @@ def _log_error_if_none(attr_name, value):
...
+
+        location = getattrib_wrapped_in_list(plugin, "location")
+        failover_location = getattrib_wrapped_in_list(plugin, "failoverlocation")
+
+        additional_location_node = find(plugin, "additionalLocations")
+        nodes_list = findall(additional_location_node, "additionalLocation") or []
```
narrieta, line 437: findall already returns an empty list, i don't think we need "or []"

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -421,9 +426,19 @@ def _log_error_if_none(attr_name, value):
...
+        additional_locations = [gettext(node) for node in nodes_list]
+
+        locations = location + additional_locations + failover_location
+        for uri in locations:
```
narrieta, line 441: let's set a limit on this... (this is wrt the internal conversation we've been having on network failures and delays)

azurelinuxagent/common/protocol/goal_state.py
kevinclark19a: In the case of `node` being non-None but not containing an attribute named `attr_name`, getattrib returns an empty string (via `common/utils/textutil.py#91`).

Still though, I agree with the idea of making this explicit: `if attr not in (None, "")`.

azurelinuxagent/common/protocol/goal_state.py
kevinclark19a: Limit is added, in common/protocol/wire.py

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -432,9 +443,19 @@ def _log_error_if_none(attr_name, value):
...
-        failover_location = getattrib(plugin, "failoverlocation")
-        for uri in [location, failover_location]:
+        def getattrib_wrapped_in_list(node, attr_name):
+            attr = getattrib(node, attr_name)
+            return [attr] if attr not in (None, "") else []
+
+        location = getattrib_wrapped_in_list(plugin, "location")
+        failover_location = getattrib_wrapped_in_list(plugin, "failoverlocation")
+
+        additional_location_node = find(plugin, "additionalLocations")
```
narrieta, line 453: `In regions with this feature enabled,
        CRP will provide any extra links in the format above. If no extra links
        are provided, the subnode will not exist.
`
if the subnode does not exit, find will return None. Let's add a check for that

azurelinuxagent/common/protocol/goal_state.py
narrieta: thanks

azurelinuxagent/common/protocol/wire.py
```diff
@@ -661,7 +662,13 @@ def fetch_manifest(self, version_uris):
         version_uris_shuffled = version_uris
         random.shuffle(version_uris_shuffled)
 
+        start_time = datetime.now()
         for version in version_uris_shuffled:
+            
+            if datetime.now() - start_time > timedelta(minutes=5):
```
larohra, line 668: Can you add a comment here suggesting why we're doing this

azurelinuxagent/common/protocol/wire.py
```diff
@@ -661,7 +662,13 @@ def fetch_manifest(self, version_uris):
...
+                logger.warn("Agent timed-out while fetching extension manifests.")
+                break
```
larohra, line 670: raise instead of break

tests/ga/test_extension.py
```diff
@@ -3093,6 +3093,58 @@ def test_collect_ext_status_parse_ext_status_exceptions(self, mock_lib_dir, *arg
...
+        AgentTestCase.setUp(self)
+        self.mock_sleep = patch("time.sleep", lambda *_: mock_sleep(0.0001))
+        self.mock_sleep.start()
+        self.test_data = DATA_FILE_EXT_ADDITIONAL_LOCATIONS.copy()
+
+    def tearDown(self):
+        self.mock_sleep.stop()
+        AgentTestCase.tearDown(self)
+
+    def test_additional_locations_node_is_consumed(self):
```
larohra, line 3108: Can you add a test case for ensuring we adhere to the timeout? Thanks!

azurelinuxagent/common/protocol/wire.py
```diff
@@ -661,7 +662,13 @@ def fetch_manifest(self, version_uris):
         version_uris_shuffled = version_uris
         random.shuffle(version_uris_shuffled)
 
+        start_time = datetime.now()
         for version in version_uris_shuffled:
+            
+            if datetime.now() - start_time > timedelta(minutes=5):
+                logger.warn("Agent timed-out while fetching extension manifests.")
```
narrieta, line 669: the current error message is "Failed to fetch manifest from all sources". Let's improve that for this specific failure to also mention that we timed out after 5 min (maybe including the number of uri that were tried)

azurelinuxagent/common/protocol/goal_state.py
kevinclark19a: `findall` actually handles this on testutil.py#45:
```python
if root is None:
    return []
```

I suppose we could inline the variable (so that, in the future, changes don't use `additional_location_node` thinking it is non-None), but current functionality isn't effected by this (when I tested this PR in the wrong region, the node was missing and things worked as expected).

azurelinuxagent/common/protocol/wire.py
kevinclark19a: You mean `timeout=5` for fetch manifest itself, right?

larohra: Actually yeah parameterizing it might even be better. I just meant adding a global variable but your idea is better haha!

azurelinuxagent/common/protocol/goal_state.py
narrieta: Sure, the current code handles None, but the fact that this variable can be None is not clear from the code, and as this code changes bugs may be introduced.

Being more explicit would help

```
additional_location_node = find(plugin, "additionalLocations")
if additional_location_node is not None:
   nodes_list = findall(additional_location_node, "additionalLocation")
   additional_locations = [gettext(node) for node in nodes_list]
   locations = location + additional_locations + failover_location
```

narrieta (approved):

PR #2199, opened 2021-03-05
Delete CGroups.Enabled flag; add debug info

narrieta:
I added CGroups.Enabled a few iterations ago, but after the first round of tests I don't see a good motivation for keeping it so I am removing it.

Also, I am adding some small changes to help end-to-end tests. Some of those changes are activated using new "Debug" flags in waagent.conf.


azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -75,13 +74,13 @@
 
 def _log_cgroup_info(format_string, *args):
     message = format_string.format(*args)
-    logger.info(message)
+    logger.info("[CGI]" + message)
```
narrieta, line 77: Adding [CGI] and [CGW] makes easier to parse the log from tests.

azurelinuxagent/common/conf.py
```diff
@@ -150,7 +154,12 @@ def load_conf_from_file(conf_file_path, conf=__conf__):
     "HttpProxy.Port": None,
     "ResourceDisk.SwapSizeMB": 0,
     "Autoupdate.Frequency": 3600,
-    "Logs.CollectPeriod": 3600
+    "Logs.CollectPeriod": 3600,
+    #
+    # "Debug" options are experimental and may be removed in later
+    # versions of the Agent.
+    #
+    "Debug.CgroupCheckPeriod": 300,
```
larohra, line 162: Nice! Small thought, could you group all the `Debug` options together? Makes it easier to read and categorize.

larohra (approved):
1 minor comment else LGTM

azurelinuxagent/common/conf.py
narrieta: The current code groups them by type so I would need to create 3 parallel arrays (one for each type) and add code to traverse those arrays to several functions.

Hopefully that comment help finding the Debug options easier.

PR #2200, opened 2021-03-05
Use a sample agent zip for update tests

narrieta:
Now the version of the test agent zip does not need to match the current agent versions.

The tests use a base zip (WALinuxAgent-0.0.0.0.zip) and make a copy using the current agent version.


tests/ga/test_update.py
```diff
@@ -79,43 +79,66 @@
...
+
+    @classmethod
+    def tearDownClass(cls):
+        AgentTestCase.tearDownClass()
+        shutil.rmtree(UpdateTestCase._agent_zip_dir)
 
-def get_agent_name():
-    return os.path.basename(get_agent_path())
+    @staticmethod
+    def _get_agent_pkgs(in_dir=None):
```
narrieta, line 110: I made those methods class members to allow me to clean up on tearDownClass

kevinclark19a (approved):

PR #2205, opened 2021-03-12
logger.warn when http fetch with exception happened.

ZhidongPeng:
## Description

logger.warn when http fetch with exception happened.

---

### PR information


kevinclark19a (approved):

kevinclark19a (dismissed):

narrieta (dismissed):

azurelinuxagent/common/protocol/wire.py
```diff
@@ -714,7 +714,7 @@ def _fetch_response(self, uri, headers=None, use_proxy=None):
                     host_plugin.report_fetch_health(uri, source='WireClient')
 
         except (HttpError, ProtocolError, IOError) as error:
-            logger.verbose("Fetch failed from [{0}]: {1}", uri, error)
+            logger.warn("Fetch failed from [{0}]: {1}", uri, error)
```
larohra, line 717: Wont this be too verbose? What do the Agent-NSG scenario logs look like with your changes? I'm afraid if its too verbose it would cause a lot of red-herrings

ZhidongPeng: I tested it with NSG and disabled the HostGAPlugin, not that  verbose.

- 6 for VMAgent manifest file download. it per walinuxagent start or hourly self update check
- 1 for vmsettings blob download.  it try download vmSettins blob every 6 minutes
- 2 for my VMAccess extension manifest file download. it is per  new goalstate per extension handler

larohra: Yup but all those failures are for the direct download right? And additionally even after those errors we do succeed with HostGA. Essentially all I'm trying to say the scenario succeeded even if a certain part of it failed. Given that, wouldn't it be cleaner to keep it as is? Like the errors should be more like "Direct download failed" rather than a failure for each link, and I think we already have that if I'm not wrong right?

larohra: synced offline, resolving this for now.

larohra (approved):

kevinclark19a (approved):

kevinclark19a (approved):

azurelinuxagent/common/utils/restutil.py
```diff
@@ -172,6 +172,19 @@ def _parse_url(url):
         secure = True
     return o.hostname, o.port, secure, rel_uri
 
+def _trim_url_parameters(url):
```
larohra, line 175: Since this is a util function, it doesn't need to be a protected function. Can rename it to `trim_url_parameters` instead

larohra (approved):

PR #2206, opened 2021-03-16
Report and cleanup handlers only if GS supported

larohra:
## Description
Report and cleanup extension handlers only if last GoalState was supported. 
Incase it was unsupported, we should report the state of the extensions in the VM over what the new GS was asking us for and skip cleanup too.

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -360,9 +360,7 @@ class ExtHandlersHandler(object):
                 self.__gs_aggregate_status = GoalStateAggregateStatus(status=GoalStateStatus.Failed, seq_no=etag,
                                                                       code=GoalStateAggregateStatusCodes.GoalStateUnsupportedRequiredFeatures,
                                                                       message=msg)
-                add_event(AGENT_NAME,
-                          version=CURRENT_VERSION,
-                          op=WALAEventOperation.GoalStateUnsupportedFeatures,
+                add_event(op=WALAEventOperation.GoalStateUnsupportedFeatures,
                           is_success=False,
                           message=msg,
                           log_event=False)
@@ -377,9 +375,7 @@ class ExtHandlersHandler(object):
                                                                   code=GoalStateAggregateStatusCodes.GoalStateUnknownFailure,
                                                                   message=msg)
             logger.warn(msg)
-            add_event(AGENT_NAME,
-                      version=CURRENT_VERSION,
-                      op=WALAEventOperation.ExtensionProcessing,
+            add_event(op=WALAEventOperation.ExtensionProcessing,
                       is_success=False,
                       message=msg,
                       log_event=False)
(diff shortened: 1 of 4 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -728,17 +733,30 @@ def report_ext_handlers_status(self):
...
-                        op=WALAEventOperation.ExtensionProcessing,
-                        is_success=False,
-                        message=ustr(error))
+                    handler_instance = ExtHandlersHandler.get_ext_handler_instance_from_path(name=item,
+                                                                                             path=path,
+                                                                                             protocol=self.protocol)
+                    if handler_instance is not None:
+                        handlers_to_report.append(handler_instance.ext_handler)
+                except Exception:
+                    continue
```
narrieta, line 749: do we need to log the error here?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -296,6 +296,11 @@ def _incarnation_changed(self, etag):
         # Skip processing if GoalState incarnation did not change
         return self.last_etag != etag
 
+    def __last_gs_unsupported(self):
```
narrieta, line 299: The code will work as is, but the return value of this function is valid only after the aggregate status has been initialized. Consider raising an exception if it is called before that.

Also, the fact that there are unsupported features is more a property of the goal state itself, not of the aggregate status or the ExtHandlersHandler. We can leave as is an do that on V3, if you prefer so.

narrieta (dismissed):
LGTM; couple of comments

azurelinuxagent/ga/exthandlers.py
larohra: no not really needed, this would throw in very rare cases where we're unable to parse the path and convert it to an ExtHandler object, not to mention that this code would be called every 6 secs so logging would fill up the logs really fast. I think its fine as is, worst case we won't report the buggy handler.

azurelinuxagent/ga/exthandlers.py
larohra: >but the return value of this function is valid only after the aggregate status has been initialized.

This would always be initialized, we're initializing a default value in `__init__`

```python
self.__gs_aggregate_status = GoalStateAggregateStatus(status=GoalStateStatus.Initialize, seq_no="-1",
                                                              code=GoalStateAggregateStatusCodes.Success,
                                                              message="Initializing new GoalState")
```

>Also, the fact that there are unsupported features is more a property of the goal state itself, not of the aggregate status or the ExtHandlersHandler.

I totally agree with this, I was contemplating making it a property of the GS but decided against it as it would need extra code changes which would not add a lot of value. I agree with making it part of V3, we can just design it accordingly.

narrieta: It is initialized, but the value doesn't have an actual valid meaning, since the computation for unsupported features is done later. At this point the best one can say is 'unknown'
Ok to do in V3

azurelinuxagent/ga/exthandlers.py
narrieta: If this can be produced by a buggy handler then I think we should definitively report it (addressing the verbosity issue)

larohra: Sure, added a change to report this only on incarnation change.

narrieta (approved):

PR #2210, opened 2021-03-23
Fix for invalid gs createdOnTicks values

kevinclark19a:
## Description

CRP sometimes sends an extension config file with the `createdOnTicks` attribute of the `InVmGoalStateMetaData` node set to `0`. This isn't a fatal error, as this info is just debug information for linux as of now. This PR adds logic to ignore invalid data like this and process the goal state anyway.

---

### PR information

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -278,15 +278,29 @@ class ExtHandlersHandler(object):
         :return: Tuple (activity_id, correlation_id, gs_created_timestamp) or "NA" for any property that's not available
         """
 
-        def check_empty(value): return value if value not in (None, "") else "NA"
+        def parse_value(parse_fn, value):
+
+            try:
+                if value not in (None, ""):
+                    return parse_fn(value)
+            except Exception as e:
+                # A failure here isn't a fatal error, because the info we're
+                # trying to retrieve is debug only on linux.
+                error_msg = u"Couldn't parse debug metadata value: {0}".format(e)
+                logger.verbose(error_msg)
+            
+            return "NA"
+        
+        fmt_time = lambda time: time.strftime(logger.Logger.LogTimeFormatInUTC)
+        identity = lambda value: value
 
         in_vm_gs_metadata = self.protocol.get_in_vm_gs_metadata()
-        gs_creation_time = check_empty(in_vm_gs_metadata.created_on_ticks)
-        gs_creation_time = gs_creation_time.strftime(
-            logger.Logger.LogTimeFormatInUTC) if gs_creation_time != "NA" else gs_creation_time
 
-        return check_empty(in_vm_gs_metadata.activity_id), check_empty(
-            in_vm_gs_metadata.correlation_id), gs_creation_time
+        gs_creation_time = parse_value(fmt_time, in_vm_gs_metadata.created_on_ticks)
+        activity_id = parse_value(identity, in_vm_gs_metadata.activity_id)
+        correlation_id = parse_value(identity, in_vm_gs_metadata.correlation_id)
+
+        return activity_id, correlation_id, gs_creation_time
 
     def run(self):
 
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -278,7 +278,7 @@ class ExtHandlersHandler(object):
         :return: Tuple (activity_id, correlation_id, gs_created_timestamp) or "NA" for any property that's not available
         """
 
-        def parse_value(parse_fn, value):
+        def format_value(parse_fn, value):
 
             try:
                 if value not in (None, ""):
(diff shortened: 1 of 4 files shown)


tests/ga/test_extension.py
```diff
@@ -1337,7 +1337,7 @@ def test_last_etag_on_extension_processing(self, *args):
             self.assertEqual(etag, exthandlers_handler.last_etag,
                              "Last etag and etag should be same if extension processing is enabled")
 
-    def test_it_should_parse_in_vm_metadata_properly(self, mock_get, mock_crypt, *args):
+    def test_it_should_parse_valid_in_vm_metadata_properly(self, mock_get, mock_crypt, *args):
```
larohra, line 1340: Thanks for the verbose testing!

azurelinuxagent/ga/exthandlers.py
```diff
@@ -298,11 +298,28 @@ def run(self):
...
+                        "Activity Id: {0}".format(activity_id),
+                        "Correlation Id: {0}".format(correlation_id),
+                        "GS Creation Time: {0}".format(gs_creation_time)
+                    ])
+                    logger.info("ProcessGoalState started [{0}]", debug_info)
+                except Exception as e:
+                    # A failure here isn't a fatal error, because the info we're
+                    # trying to retrieve is debug only on linux. We still need to 
+                    # process the goal state.
+                    error_msg = u"ProcessGoalState started, but failed to retrieve gs debug metadata: '{0}'"\
```
larohra, line 315: I think this is a bit too much debug data for something small (i.e. debug data). If this data is not there or we're unable to parse it properly, we can just skip it altogether and just log a basic message. At max, just log it once as a warn or something per goalstate. No need for Kusto logging here either.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -298,11 +298,28 @@ def run(self):
             self.log_report = True
 
             if self._extension_processing_allowed() and self._incarnation_changed(etag):
-                activity_id, correlation_id, gs_creation_time = self.get_goal_state_debug_metadata()
-
-                logger.info(
-                    "ProcessGoalState started [Incarnation: {0}; Activity Id: {1}; Correlation Id: {2}; GS Creation Time: {3}]".format(
-                        etag, activity_id, correlation_id, gs_creation_time))
+                try:
+                    activity_id, correlation_id, gs_creation_time = self.get_goal_state_debug_metadata()
```
larohra, line 302: A small suggestion, try parsing the data and if unavailable then return `NA` for the ones that we were unable to parse. Maybe move the try-catch inside the `get_goal_state_debug_metadata` function to make this main loop cleaner?

tests/ga/test_extension.py
```diff
@@ -1349,6 +1349,8 @@ def test_it_should_parse_in_vm_metadata_properly(self, mock_get, mock_crypt, *ar
         self.assertEqual(correlation_id, "400de90b-522e-491f-9d89-ec944661f531", "Incorrect correlation Id")
         self.assertEqual(gs_creation_time, '2020-11-09T17:48:50.412125Z', "Incorrect GS Creation time")
 
+    def test_it_should_parse_missing_in_vm_metadata_properly(self, mock_get, mock_crypt, *args):
```
larohra, line 1352: The name suggests that the function would parse metadata even if its missing. Maybe rename to `test_it_should_process_goal_state_even_if_metadata_missing` and merge the two tests in one?

kevinclark19a: I would like to keep the tests separate because they're all using different data files, but I agree about the rename

azurelinuxagent/ga/exthandlers.py
```diff
@@ -278,15 +278,29 @@ def get_goal_state_debug_metadata(self):
...
+            try:
+                if value not in (None, ""):
+                    return parse_fn(value)
+            except Exception as e:
+                # A failure here isn't a fatal error, because the info we're
+                # trying to retrieve is debug only on linux.
+                error_msg = u"Couldn't parse debug metadata value: {0}".format(e)
+                logger.verbose(error_msg)
+            
+            return "NA"
```
narrieta, line 292: Rather than returning for None, empty string or unparsable time we should just leave the value unchanged so that we know what we are getting from CRP. Displaying it as a date/time was just meant to make reading the log easier. If we get "0", we should just log "0".

azurelinuxagent/ga/exthandlers.py
```diff
@@ -278,15 +278,30 @@ def get_goal_state_debug_metadata(self):
...
+
+            try:
+                if value not in (None, ""):
+                    return parse_fn(value)
+            except Exception as e:
+                # A failure here isn't a fatal error, because the info we're
+                # trying to retrieve is debug only on linux.
+                error_msg = u"Couldn't parse debug metadata value: {0}".format(e)
+                logger.verbose(error_msg)
+                return value
```
larohra, line 291: the type of the `in_vm_gs_metadata.created_on_ticks` is datetime so just returning this wont work. If you want to return the string value of what we get from CRP, we need to modify this - `self.created_on_ticks = __ticks_to_datetime(getattrib(in_vm_metadata_node, "createdOnTicks"))`

narrieta (approved):

larohra (approved):

PR #2211, opened 2021-03-23
Logrotate version bugfix

kevinclark19a:
## Description

Some versions of `logrotate` respond to `--version` on `stderr` instead of `stdout`. This PR pipes `stderr` output to `stdout`, and additionally makes the version parsing more robust.

---

### PR information

### DIFF
--- a/azurelinuxagent/common/version.py
+++ b/azurelinuxagent/common/version.py
@@ -18,12 +18,13 @@
 import os
 import re
 import platform
+import subprocess
 import sys
 
 import azurelinuxagent.common.conf as conf
 import azurelinuxagent.common.utils.shellutil as shellutil
 from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
-from azurelinuxagent.common.future import ustr, get_linux_distribution
+from azurelinuxagent.common.future import ustr, get_linux_distribution, is_file_not_found_error
 
 __DAEMON_VERSION_ENV_VARIABLE = '_AZURE_GUEST_AGENT_DAEMON_VERSION_'
 """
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/version.py
```diff
@@ -183,15 +184,22 @@ def get_lis_version():
...
+        cmd_output = shellutil.run_command(["logrotate", "--version"], stderr=subprocess.STDOUT)
+        match = re.search(regex, cmd_output)
+        if match:
+            return match.group("version")
+
+    except Exception as e:
+        if is_file_not_found_error(e):
+            # run_command actually throws a FileNotFound if there
+            # is no executable corresponding to the command given.
+            return COMMAND_ABSENT
```
narrieta, line 200: "Absent" and "Failed" don't seem to be very useful. Why do we even attempt to parse instead of just logging what the command returns?

kevinclark19a: In the case that logrotate is not present on the system, there is no output, and a file not found error is thrown. I suppose the regex matching doesn't belong in the try clause... But with no output, we have to return something. 

Additionally, certain logrotate versions have very verbose outputs, e.g.:

```bash
logrotate 3.14.0

    Default mail command:       /usr/bin/mail
    Default compress command:   /bin/gzip
    Default uncompress command: /bin/gunzip
    Default compress extension: .gz
    Default state file path:    /var/lib/logrotate/status
    ACL support:                yes
    SELinux support:            yes
```

We don't want to log all of this in the OSInfo message, right?

narrieta: If the output changes per distro then attempting to parse may also have issues. Let's remove this info from the log and come back to it on V3

narrieta (approved):

PR #2216, opened 2021-03-26
Report previous extension data on Unsupported GS 

larohra:
## Description
Report previous extension data if the last GS was Unsupported
(Earlier we were only reporting HandlerStatus and not extension status)

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -47,7 +47,7 @@ from azurelinuxagent.common.exception import ExtensionDownloadError, ExtensionEr
     ExtensionOperationError, ExtensionUpdateError, ProtocolError, ProtocolNotFoundError, ExtensionConfigError, GoalStateAggregateStatusCodes
 from azurelinuxagent.common.future import ustr, is_file_not_found_error
 from azurelinuxagent.common.protocol.restapi import ExtensionStatus, ExtensionSubStatus, ExtHandler, ExtHandlerStatus, \
-    VMStatus, GoalStateAggregateStatus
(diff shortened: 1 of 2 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -741,7 +741,11 @@ def report_ext_handlers_status(self, incarnation_changed=False):
                                                                                              path=path,
                                                                                              protocol=self.protocol)
                     if handler_instance is not None:
-                        handlers_to_report.append(handler_instance.ext_handler)
+                        ext_handler = handler_instance.ext_handler
+                        # We need to add one extension for the agent to report its status (Single config only expects
+                        # a single extension per Handler). Without this no extensions would be reported for this Handler
+                        ext_handler.properties.extensions.append(Extension(name=ext_handler.name))
```
narrieta, line 747: do we need to initialize other properties other than the name. e.g. normally crp would check the sequence number... even if the "unsupported gs" code path in crp doesn't check it today, it might be a good idea to initialize it.

larohra: The sequence number is fetched later in the code by looking at the last updated settings file (look at this function `def get_status_file_path`).

This code path is very convoluted, we fetch extensions in one place, read the status in a different place and then report it in a completely different place. I can walk you through the code for more clarity and how its all propagated down.

I wanted to modify it (I actually even am a lot of it for MultiConfig) but at the same time not touching a lot of it because of v3. Let me know if you want a deeper sync about it

narrieta (approved):

PR #2242, opened 2021-05-05
Use venv for Python 2.6 unit tests

narrieta:


.github/workflows/ci_pr.yml
```diff
@@ -22,26 +22,14 @@ jobs:
...
-        sudo apt-get update -y
-        sudo apt-get install python2.6 python2.6-dev -y
-        sudo ln -sf /usr/bin/python2.6 /usr/bin/python
-        PATH="/usr/bin/${PATH:+:${PATH}}"
-        echo $PATH
-        echo "Installed python2.6"
-        python -V
-        curl "https://bootstrap.pypa.io/pip/2.6/get-pip.py" -o get-pip.py
-        sudo env "PATH=$PATH" python get-pip.py
+        curl https://dcrdata.blob.core.windows.net/python/python-2.6.tar.bz2 -o python-2.6.tar.bz2
```
larohra, line 25: Does this blob ever expire? Can you add comments with links on how to access the blob and/or update it if necessary. Thanks for the fix though, much needed!

larohra (approved):

.github/workflows/ci_pr.yml
larohra: Also if possible, could you also add the steps on how to update the venv if we ever add any dependencies (very unlikely though still good to have)

narrieta: no, it does not expire (i did not use a sas token since i did not want it to expire)

PR #2257, opened 2021-05-20
Updated _read_status_file to include a fragment of status file in the exception

dhivyaganesan:
## Description

Issue # 

---

### PR information


azurelinuxagent/ga/exthandlers.py
```diff
@@ -2133,35 +2133,35 @@ def get_env_file(self):
...
+    @staticmethod
+    def _read_status_file(ext_status_file):
+        err_count = 0
+        while True:
+            try:
+                return ExtHandlerInstance._read_and_parse_json_status_file(ext_status_file)
+            except Exception as e:
+                err_count += 1
+                if err_count >= _NUM_OF_STATUS_FILE_RETRIES:
+                    raise e
```
larohra, line 2145: NIT: Use `raise` instead of `raise e`

Explanation: 
 `raise e` will create a new Exception object, in this case the stack trace would begin from this line.  
On the contrary, just simply `raise` will actually re-raise the original error and the stack trace would be from the place where the original error is coming from. 

More info on this post - https://stackoverflow.com/a/45986945

tests/ga/test_extension.py
```diff
@@ -1585,14 +1603,52 @@ def mock_popen(cmd, *args, **kwargs):
...
 
                 # The Handler Status for the dependent extension should be NotReady as it was not executed at all
                 # And since it was not executed, it should not report any extension status either
                 self._assert_handler_status(protocol.report_vm_status, "NotReady", 0, "1.0.0",
                                             expected_msg="Dependent Extension OSTCExtensions.OtherExampleHandlerLinux did not reach a terminal state within the allowed timeout. Last status was {0}".format(
                                                 ValidHandlerStatus.warning))
 
+    def test_it_should_include_part_of_status_in_ext_handler_message(self, mock_http_get, mock_crypt_util, *args):
+        """
+        Testing depends-on scenario when the status file is invalid,
```
larohra, line 1617: NIT: Modify the comment for this specific test

tests/ga/test_extension.py
```diff
@@ -1585,14 +1603,52 @@ def mock_popen(cmd, *args, **kwargs):
...
+                invalid_json_path = os.path.join(data_dir, "ext", "sample-status-invalid-json-format.json")
+
+                if os.path.exists(status_path):
+                    invalid_json = fileutil.read_file(invalid_json_path)
+                    fileutil.write_file(status_path,invalid_json)
+
+            return original_popen(["echo", "Yes"], *args, **kwargs)
+
+        with patch('azurelinuxagent.common.cgroupapi.subprocess.Popen', side_effect=mock_popen):
+            with patch('azurelinuxagent.ga.exthandlers._DEFAULT_EXT_TIMEOUT_MINUTES', 0.001):
```
larohra, line 1640: This patch is not relevant since its not a depends-on case, not needed here.

tests/ga/test_extension.py
```diff
@@ -3335,26 +3435,25 @@ def manifest_location_handler(url, **kwargs):
...
 
+        manifest_location_handler.num_times_called = 0
 
         with mock_wire_protocol(self.test_data, http_get_handler=manifest_location_handler) as protocol:
             ext_handlers, _ = protocol.get_ext_handlers()
 
             with self.assertRaises(ExtensionDownloadError):
                 protocol.client.fetch_manifest(ext_handlers.extHandlers[0].versionUris,
-                    timeout_in_minutes=0, timeout_in_ms=200)
+                                               timeout_in_minutes=0, timeout_in_ms=200)
```
larohra, line 3456: Thanks for doing these cosmetic changes! I hope they're at-par with the pep-8 standards!

azurelinuxagent/ga/exthandlers.py
dhivyaganesan: Thank you for the suggestion , Fixed it

tests/ga/test_extension.py
dhivyaganesan: changed it.

tests/ga/test_extension.py
dhivyaganesan: Right, Sorry I missed it . Removed it now.

tests/ga/test_extension.py
dhivyaganesan: This created issues with pylint checks , have to undo most of it unfortunately :)

azurelinuxagent/ga/exthandlers.py
```diff
@@ -2133,35 +2133,35 @@ def get_env_file(self):
...
-                failed_to_read = True
-                raised_exception = e
-            except (ValueError, TypeError) as e:
-                failed_to_parse_json = True
-                raised_exception = e
-            time.sleep(_STATUS_FILE_RETRY_DELAY)
-
-        if failed_to_read:
-            raise ExtensionStatusError(msg=ustr(raised_exception), inner=raised_exception,
+        try:
```
narrieta, line 2153: we don't need the two above initializations (pycharm is complaining about them)

```
data_str = None
data = None

```

tests/data/ext/sample-status-invalid-json-format.json
```diff
@@ -0,0 +1,36 @@
+[
```
narrieta, line 1: could you add a comment explaining what makes this file invalid?

you could, for example, add a dummy property to the json:

> "_comment": "This is an invalid status file, it's missing a brace at line XX"

tests/ga/test_extension.py
```diff
@@ -3127,6 +3158,29 @@ def test_collect_ext_status(self, mock_lib_dir, *args):
         self.assertEqual(sub_status.message, None)
         self.assertEqual(sub_status.status, ValidHandlerStatus.success)
 
+    @patch("azurelinuxagent.common.conf.get_lib_dir")
+    def test_collect_ext_status_for_invalid_json(self, mock_lib_dir, *args):
+        """
+        This test validates that collect_ext_status correctly picks up the status file (sample-status-invalid-json-format.json)
+        and then since the Json cannot be parsed correctly it extension status message should include 2000 bytes of status file
+        and the line number in which it failed to parse
```
narrieta, line 3166: can you add the same comment as above? "The uniqueMachineId tag comes from status file"

narrieta (commented):
LGTM, a few minor comments.

tests/ga/test_extension.py
```diff
@@ -3127,6 +3158,29 @@ def test_collect_ext_status(self, mock_lib_dir, *args):
...
+        ext_status = ext_handler_i.collect_ext_status(extension)
+
+        self.assertEqual(ext_status.code, ExtensionErrorCodes.PluginSettingsStatusInvalid)
+        self.assertEqual(ext_status.configurationAppliedTime, None)
+        self.assertEqual(ext_status.operation, None)
+        self.assertEqual(ext_status.sequenceNumber, 0)
+        self.assertRegex(ext_status.message, r".*The status reported by the extension {0}-{1}\(Sequence number {2}\), "
+                                             r"was in an incorrect format and the agent could not parse it correctly."
+                                             r" Failed due to.*".
+                         format("TestHandler", "1.0.0", 0))
```
larohra, line 3179: NIT: Since these are string constants, why not just use them as is in the string above rather than using `format`?

i.e. instead of 
`r".*The status reported by the extension {0}-{1}\(Sequence number {2}\), ".format("TestHandler", "1.0.0", 0))`

use 
`r".*The status reported by the extension TestHandler-1.0.0\(Sequence number 0\), "`

larohra (approved):

narrieta (approved):

PR #2269, opened 2021-06-10
 Merge Release 2.3.0.2 into master

narrieta:


larohra (dismissed):

azurelinuxagent/common/version.py
```diff
@@ -196,7 +196,9 @@ def has_logrotate():
 
 AGENT_NAME = "WALinuxAgent"
 AGENT_LONG_NAME = "Azure Linux Agent"
-AGENT_VERSION = '2.2.54.2'
+# Setting the version to 9.9.9.9 to ensure DCR always uses this version and never auto-updates.
```
larohra, line 199: Shouldn't we remove this comment on master? I'm good either way personally just asking more for my understanding.

larohra (approved):

azurelinuxagent/common/version.py
narrieta: I had the same thought about removing it, but I think we should rather edit it... will do that on develop

dhivyaganesan (approved):

PR #2270, opened 2021-06-10
Save waagent_status to history folder and add additional details to the status file

dhivyaganesan:
## Description

- Updated waagent_status.json to have incarnation number included
- When incarnation changes waagent_status will be included in the history folder of previous incarnation
- Included more details to waagent_status

---

### PR information


azurelinuxagent/ga/exthandlers.py
```diff
@@ -969,38 +968,53 @@ def report_ext_handlers_status(self, incarnation_changed=False):
...
+            "extension_supported_features": [name for name, _ in
+                                             get_agent_supported_features_list_for_extensions().items()]
         }
 
-        # Convert VMStatus class to Dict.
-        data = get_properties(vm_status)
+        if vm_status is not None:
+            # Convert VMStatus class to Dict.
+            data = get_properties(vm_status)
+            goal_state_status = data['vmAgent']['vm_artifacts_aggregate_status']['goal_state_aggregate_status']
```
larohra, line 987: This will throw if "vm_artifacts_aggregate_status" key is not avaialable in data['vmAgent'].

azurelinuxagent/ga/exthandlers.py
```diff
@@ -969,38 +968,53 @@ def report_ext_handlers_status(self, incarnation_changed=False):
...
+
+            agent_details["daemon_version"] = data['vmAgent']['version']
+            agent_details["distro_details"] = "{0}:{1}".format(data['vmAgent']['osname'], data['vmAgent']['osversion'])
+            agent_details["agent_status"] = data['vmAgent']['status']
```
larohra, line 991: All of these would throw if the 'vmAgent' key is not available in `data`. Please check for the availability of those keys to ensure we dont throw

azurelinuxagent/ga/exthandlers.py
```diff
@@ -969,38 +968,53 @@ def report_ext_handlers_status(self, incarnation_changed=False):
...
+            agent_details["agent_message"] = data['vmAgent']['message']
+            agent_details["agent_hostname"] = data['vmAgent']['hostname']
+
+            goal_state_aggregate_status = {
+                "processed_time": time.strftime("%Y-%m-%dT%H:%M:%SZ", goal_state_status['_GoalStateAggregateStatus__utc_timestamp']),  # the timestamp is slightly different
```
larohra, line 996: You're accessing a private member of the `GoalStateAggregateStatus` class. Consider using this property instead - 

```python
@property
    def processed_time(self):
        return self.__utc_timestamp
```

azurelinuxagent/ga/exthandlers.py
```diff
@@ -969,38 +968,53 @@ def report_ext_handlers_status(self, incarnation_changed=False):
...
         }
 
-        # Convert VMStatus class to Dict.
-        data = get_properties(vm_status)
+        if vm_status is not None:
+            # Convert VMStatus class to Dict.
+            data = get_properties(vm_status)
+            goal_state_status = data['vmAgent']['vm_artifacts_aggregate_status']['goal_state_aggregate_status']
+
+            agent_details["daemon_version"] = data['vmAgent']['version']
```
larohra, line 989: Are you sure the `data['vmAgent']['version']` is the `daemon_version`? Also Where's the `goal_state_version`, I cant see it anywhere. Thanks!

azurelinuxagent/ga/exthandlers.py
```diff
@@ -969,38 +968,53 @@ def report_ext_handlers_status(self, incarnation_changed=False):
...
+    def write_ext_handlers_status_to_info_file(self, vm_status):
+
+        status_path = os.path.join(conf.get_lib_dir(), AGENT_STATUS_FILE.format(self.last_etag))
 
         agent_details = {
             "agent_name": AGENT_NAME,
-            "current_version": str(CURRENT_VERSION),
             "goal_state_version": str(GOAL_STATE_AGENT_VERSION),
-            "distro_details": "{0}:{1}".format(DISTRO_NAME, DISTRO_VERSION),
-            "last_successful_status_upload_time": time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime()),
```
larohra, line 981: I also dont see an alternative for this in the final fields, what did you replace it with?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -978,33 +978,40 @@ def write_ext_handlers_status_to_info_file(self, vm_status):
...
 
-            agent_details["daemon_version"] = data['vmAgent']['version']
-            agent_details["distro_details"] = "{0}:{1}".format(data['vmAgent']['osname'], data['vmAgent']['osversion'])
-            agent_details["agent_status"] = data['vmAgent']['status']
-            agent_details["agent_message"] = data['vmAgent']['message']
-            agent_details["agent_hostname"] = data['vmAgent']['hostname']
+            # .get handles the keyError and accepts a default value as 2nd parameter. If no second parameter is
+            # provided it uses None. Using {'default': None} in the case of 'vmAgent' or any other key is not available
+
+            agent_details["daemon_version"] = data.get('vmAgent', {'default': None}).get('version')
```
larohra, line 992: In case 'vmAgent' is not there, the default would return None. 
If the value is None, then 'None'.get('version') would throw too right?

Why not have an if condition at the top and go like - 

```python
if 'vmAgent' in data:
    # Fetch other stuff
```

And maybe do this for all nested properties?

dhivyaganesan: I found . get to be a good way to handle nested dict. In case of VmAgent not available it will consider vmAgent = {'default': None} , which is a valid JSON and will not throw KeyError. I.e it will try to get 'version' from {'default': None} and since 'version' is not available, it will return None but not Key Error.

The problem with nested Ifs is for every property we have to surround it by an If , but felt like .get handles it more elegantly.

azurelinuxagent/ga/exthandlers.py
dhivyaganesan: Good point, Since in the original code 'vm_status' is converted to a dictionary called 'data' I could no longer accesses class properties of GoalStateAggregateStatus. Is there any other good way to do it ?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -978,33 +978,40 @@ def write_ext_handlers_status_to_info_file(self, vm_status):
             "python_version": "Python: {0}.{1}.{2}".format(PY_VERSION_MAJOR, PY_VERSION_MINOR, PY_VERSION_MICRO),
             "crp_supported_features": [name for name, _ in get_agent_supported_features_list_for_crp().items()],
             "extension_supported_features": [name for name, _ in
-                                             get_agent_supported_features_list_for_extensions().items()]
+                                             get_agent_supported_features_list_for_extensions().items()],
+            "last_successful_status_upload_time": time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime())
```
larohra, line 982: Do we not log this property in the status blob? It would be preferable to get this data directly from the blob as compared to getting it from the `time` module.

From the [publshers doc](https://microsoft.sharepoint.com/:w:/t/ComputeVM/EforWBxWd9dGrdcuW732cBkBYS5xlb0UV1LzjtXb2CISUw?e=azeir0) sec 2.4.2 (Status reporting), I do see we have a property where we log the time - 

`"timestampUTC": "<current utc time>",`

We should use this property

azurelinuxagent/ga/exthandlers.py
dhivyaganesan: I was assuming the "processed_time" we collect under "goal_state_aggregate_status " . But thinking about it , if goal_state_aggregate_status is not available then we will miss it. so added it back at line no 982

azurelinuxagent/ga/exthandlers.py
dhivyaganesan: I would like to sync offline about the daemon_version. goal_state_version is available in line 977

azurelinuxagent/ga/exthandlers.py
```diff
@@ -969,38 +968,60 @@ def report_ext_handlers_status(self, incarnation_changed=False):
...
+                                                               data.get('vmAgent', {'default': None}).get('osversion'))
+            agent_details["agent_status"] = data.get('vmAgent', {'default': None}).get('status')
+            agent_details["agent_message"] = data.get('vmAgent', {'default': None}).get('message')
+            agent_details["agent_hostname"] = data.get('vmAgent', {'default': None}).get('hostname')
+
+            goal_state_status = data.get('vmAgent', {'default': None}).get('vm_artifacts_aggregate_status',
+                                                {'default': None}).get('goal_state_aggregate_status', {'default': None})
+            goal_state_aggregate_status = {
+                "processed_time": time.strftime("%Y-%m-%dT%H:%M:%SZ",
+                                                goal_state_status.get('_GoalStateAggregateStatus__utc_timestamp')),
```
nagworld9, line 1003: Not sure what happens if we get None here. Quick test gave me this error, could you please double check and handle if this is true.
```
$python main.py
  File "main.py", line 6, in <module>
    print(time.strftime("%Y-%m-%dT%H:%M:%SZ", None))
TypeError: argument must be 9-item sequence, not None
```

dhivyaganesan: Good catch @nagworld9 , Addressed.

azurelinuxagent/ga/exthandlers.py
dhivyaganesan: Addressed

azurelinuxagent/ga/exthandlers.py
dhivyaganesan: Addressed

azurelinuxagent/ga/exthandlers.py
dhivyaganesan: Addressed

azurelinuxagent/common/utils/archive.py
```diff
@@ -43,7 +43,8 @@
 _CACHE_PATTERNS = [
     re.compile(r"^(.*)\.(\d+)\.(agentsManifest)$", re.IGNORECASE),
     re.compile(r"^(.*)\.(\d+)\.(manifest\.xml)$", re.IGNORECASE),
-    re.compile(r"^(.*)\.(\d+)\.(xml)$", re.IGNORECASE)
+    re.compile(r"^(.*)\.(\d+)\.(xml)$", re.IGNORECASE),
```
narrieta, line 46: what files is this one trying to capture? thanks

azurelinuxagent/common/utils/archive.py
```diff
@@ -43,7 +43,8 @@
...
+    re.compile(r"^(.*)waagent_status\.(\d+)\.json$")
```
narrieta, line 47: we don't need "(.*)" before "waagent_status"

azurelinuxagent/ga/exthandlers.py
```diff
@@ -979,39 +978,42 @@ def report_ext_handlers_status(self, incarnation_changed=False):
...
                       message=msg)
 
-    @staticmethod
-    def write_ext_handlers_status_to_info_file(vm_status):
-        status_path = os.path.join(conf.get_lib_dir(), AGENT_STATUS_FILE)
-
-        agent_details = {
-            "agent_name": AGENT_NAME,
-            "current_version": str(CURRENT_VERSION),
-            "goal_state_version": str(GOAL_STATE_AGENT_VERSION),
```
narrieta, line 989: did we remove goal_state_version?

azurelinuxagent/ga/exthandlers.py
narrieta: did you add last_successful_status_upload_time back?

tests/ga/test_extension.py
```diff
@@ -605,6 +605,50 @@ def test_ext_handler(self, *args):
 
         self._assert_no_handler_status(protocol.report_vm_status)
 
+    def test_it_should_zip_waagent_status_when_incarnation_changes(self, *args):
+        """
+        This test checks when the incarnation changes the waagnet_status file for the previous incarnation
```
narrieta, line 610: typo: "waagnet"

tests/ga/test_extension.py
```diff
@@ -605,6 +605,50 @@ def test_ext_handler(self, *args):
...
+        """
+        temp_files = [
+            'ExtensionsConfig.1.xml',
+            'GoalState.1.xml',
+            'OSTCExtensions.ExampleHandlerLinux.1.manifest.xml',
+            'waagent_status.1.json'
+        ]
+
+        test_data = mockwiredata.WireProtocolData(mockwiredata.DATA_FILE)
+        exthandlers_handler, protocol = self._create_mock(test_data, *args)  # pylint: disable=no-value-for-parameter
```
narrieta, line 621: new code should not have pylint suppressions unless there is no way to avoid them (and it that case, the reason should be documented)

tests/ga/test_extension.py
```diff
@@ -605,6 +605,50 @@ def test_ext_handler(self, *args):
...
+        # Updating incarnation to 2 , hence the history folder should have waaagent_status.1.json added under
+        # incarnation 1
+        test_data.set_incarnation(2)
+        protocol.update_goal_state()
+
+        test_subject = StateArchiver(self.tmp_dir)
+        test_subject.archive()
+
+        timestamp_zips = os.listdir(os.path.join(self.tmp_dir, "history"))
+        self.assertEqual(1, len(os.listdir(os.path.join(self.tmp_dir, "history"))))
```
narrieta, line 635: should we use timestamp_zips instead of os.listdir(os.path.join(self.tmp_dir, "history")?

narrieta: Please add an appropriuate message  to the assertion for the case when it fails

tests/ga/test_extension.py
```diff
@@ -605,6 +605,50 @@ def test_ext_handler(self, *args):
...
+        exthandlers_handler.report_ext_handlers_status()
+
+        # Updating incarnation to 3 , hence the history folder should have 2 zips files corresponding to incarnation
+        # 1 and 2
+        test_data.set_incarnation(3)
+        protocol.update_goal_state()
+        test_subject.archive()
+        self.assertEqual(2, len(os.listdir(os.path.join(self.tmp_dir, "history"))))
+
+        exthandlers_handler.run()
```
narrieta, line 650: is this call to run() needed?

tests/ga/test_extension.py
```diff
@@ -3411,6 +3398,111 @@ def manifest_location_handler(url, **kwargs):
...
+                    return MockHttpResponse(status=500)
+                protocol.aggregate_status = json.loads(args[0])
+                return MockHttpResponse(status=201)
+
+            protocol.aggregate_status = None
+            protocol.set_http_handlers(http_put_handler=mock_http_put)
+            exthandlers_handler = get_exthandlers_handler(protocol)
+            yield exthandlers_handler
+
+    @patch('time.gmtime', MagicMock(return_value=time.gmtime(0)))
```
narrieta, line 3425: this test should be with the other tests for status... any reason to use a new test class for it?

tests/ga/test_extension.py
```diff
@@ -3411,6 +3398,111 @@ def manifest_location_handler(url, **kwargs):
...
+class TestExtensionWithMockHTTP(AgentTestCase):
+
+    def setUp(self):
+        AgentTestCase.setUp(self)
+
+    def tearDown(self):
+        AgentTestCase.tearDown(self)
+
+    @contextlib.contextmanager
+    def _setup_test_env(self, test_data):
```
narrieta, line 3410: in this case the context manager is not buying you anything, I think. You could do the setup in the test itself (or are you planning to re-use the context manager for additional tests?)

azurelinuxagent/ga/exthandlers.py
dhivyaganesan: It now comes from the status_blob_data as "timestampUTC". The new data follows the same structure of the status_blob minus 'formattedMessage' and 'substatus' from each handlerStatus

azurelinuxagent/common/utils/archive.py
dhivyaganesan: This pattern was already present as a part of previous code. Digging further its matches files such as 'GoalState.0.xml', 'Microsoft.Azure.Extensions.CustomScript.0.xml' .These files will be added to the list of files that needs to be archived to history

azurelinuxagent/ga/exthandlers.py
dhivyaganesan: Goal state version now comes from the status_blob_data 
  "aggregateStatus": {
    "guestAgentStatus": {
      "version": "9.9.9.9", // Goal state version
      "status": "Ready",
      "formattedMessage": {
        "lang": "en-US",
        "message": "Guest Agent is running"
      }
    }

tests/ga/test_extension.py
dhivyaganesan: Probably not, good point , addressed

tests/ga/test_extension.py
dhivyaganesan: Addressed in next commit

tests/ga/test_extension.py
dhivyaganesan: Addressed in the next commit

tests/ga/test_extension.py
dhivyaganesan: Addressed in the next commit

azurelinuxagent/common/utils/archive.py
dhivyaganesan: Addressed in the next commit

tests/ga/test_extension.py
dhivyaganesan: Addressed it

azurelinuxagent/common/utils/archive.py
narrieta: ok

azurelinuxagent/ga/exthandlers.py
```diff
@@ -979,39 +978,42 @@ def report_ext_handlers_status(self, incarnation_changed=False):
...
-        fileutil.write_file(status_path, agent_details_json)
+    def write_ext_handlers_status_to_info_file(self, vm_status):
+        status_path = os.path.join(conf.get_lib_dir(), AGENT_STATUS_FILE.format(self.protocol.get_incarnation()))
+        status_blob_data = self.protocol.get_status_blob_data()
+        data = dict()
+        if status_blob_data is not None:
+            data = json.loads(status_blob_data)
+
+        # Populating the fields that does not come from vm_status or status_blob_data
+        data["agentName"] = AGENT_NAME
```
larohra, line 989: Any reason for changing the naming convention from underscore to Camel casing? 

Eg: `agent_name` to `agentName`

azurelinuxagent/ga/exthandlers.py
narrieta: ok

tests/ga/test_extension.py
```diff
@@ -3411,6 +3354,150 @@ def manifest_location_handler(url, **kwargs):
                 protocol.client.fetch_manifest(ext_handlers.extHandlers[0].versionUris,
                     timeout_in_minutes=0, timeout_in_ms=200)
 
+# New test cases should be added here.This class uses mock_wire_protocol
```
narrieta, line 3357: Thanks for the updates

azurelinuxagent/ga/exthandlers.py
dhivyaganesan: The new logic uses the details directly from the status blob and poping out the fields that are not needed. The data in the blob is in Camel casing and for consistently reason I changed the fields that are manually populated in Camel casing as well.

larohra: Cool thanks!

larohra (approved):

azurelinuxagent/ga/exthandlers.py
```diff
@@ -67,7 +66,7 @@
 HANDLER_COMPLETE_NAME_PATTERN = re.compile(_HANDLER_PATTERN + r'$', re.IGNORECASE)
 HANDLER_PKG_EXT = ".zip"
 
-AGENT_STATUS_FILE = "waagent_status.json"
+AGENT_STATUS_FILE = "waagent_status.{0}.json"
```
kevinclark19a, line 69: NIT: not necessary, but might consider naming the variable to include that this is a format string (i.e. can't be used before `.format`)

kevinclark19a (approved):
LGTM; small comment, but feel free to merge without it.

azurelinuxagent/ga/exthandlers.py
dhivyaganesan: Something like "AGENT_STATUS_FILE_FORMAT" ?

PR #2272, opened 2021-06-11
Support sles 15 sp2 distro

nagworld9:
## Description

Issue # [TASK 8742449](https://msazure.visualstudio.com/One/_sprints/taskboard/VM%20Agent/One/Azure/CY2021H1%20-%20Cobalt/06%20-%20Jun?workitem=8742449)

Add support for SLES 15 distro

[DCR run ](https://tuxgold.corp.microsoft.com/job/AzLinux/job/DungeonCrawler/job/DungeonCrawler.nnandigam/35/)

---

### PR information


narrieta (approved):

setup.py
```diff
@@ -165,6 +165,13 @@ def get_data_files(name, version, fullname):  # pylint: disable=R0912
         else:
             # sles 12+ and openSUSE 13.2+ use systemd
             set_systemd_files(data_files, dest=systemd_dir_path)
+    elif name == 'sles': # sles 15+ distro named as sles
```
larohra, line 168: have you confirmed this comment? - 

># sles 15+ distro named as sles

I mean what does suse 13,14,etc do? Do we even have those distros? It would be wise to check all this out before committing it to code.

nagworld9: we don't have 13.x, 14.x versions and also, [azure ](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/endorsed-distros)not endorsed those distros.

kevinclark19a (dismissed):

bin/version2/waagent
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python3
+#
```
narrieta, line 2: i'd suggest naming the new directory 'py3' or something like that instead of 'v2'

setup.py
```diff
@@ -43,6 +43,10 @@ def set_bin_files(data_files, dest, src=None):
         src = ["bin/waagent", "bin/waagent2.0"]
     data_files.append((dest, src))
 
+def set_bin_version2_files(data_files, dest, src=None):
```
narrieta, line 46: we don't need this new function, do we? we can use set_bin_files

nagworld9: Idea for separate function is we only set new binary file if that distro has py3 requirement rest we use set_bin_files.

nagworld9: Keeping in one place required extra logic and need to know distro type

narrieta: the only thing we are changing is the default value... am i missing something? instead of creating a new function with a different default value you could just call the original function with the different value.

narrieta: and we also need to copy waagent2.0

nagworld9: That's is good point but my initial idea is to keep this new path as default value in other function since this is constant path(not change for other distros). This would avoid passing every time. I'm happy to change to use original function.

The current setup is first we are setting old files(waagent & waagent2.0) then overriding the waagent with new file.

narrieta: I think having 2 functions that do the same and only change the default value is very odd, but you can leave as is if you feel strongly about it.
This, though, we should change:

> The current setup is first we are setting old files(waagent & waagent2.0) then overriding the waagent with new file.

so you saying for SLES first we call set_bin_files and then set_bin_version2_files? that is confusing... I guess that is why I missed that waagent2.0 was being copied. Other people reading the code may even think that it is a bug (calling both functions)

nagworld9: Not only for SLES.. The setup was codded in way that we set binary files first irrespective of what distro it is.. If we want to call only one time then I would have to remove that and include all 10 if block statements..

nagworld9: calling outside https://github.com/Azure/WALinuxAgent/blob/f59375f211951ffe17aea1ff6c1949c7c8d16319/setup.py#L103

narrieta: I see. I looked at the rest of the code and it could be better abstracted, but for now calling set_bin_files with the correct files for each of the distros (just ad we call set_conf_files) would be clearer

nagworld9: cool I will update that

nagworld9: addressed

bin/version2/waagent
nagworld9: addressed

narrieta (approved):

kevinclark19a (approved):

PR #2279, opened 2021-06-23
onboard ubuntu20

nagworld9:
## onboard ubuntu20

Issue #[ TASK 8691973](https://msazure.visualstudio.com/One/_sprints/taskboard/VM%20Agent/One/Azure/CY2021H1%20-%20Cobalt/06%20-%20Jun?workitem=8691973)

---

### PR information

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -135,20 +135,23 @@ def get_data_files(name, version, fullname):  # pylint: disable=R0912
         set_systemd_files(data_files, dest=systemd_dir_path,
                           src=["init/mariner/waagent.service"])
     elif name == 'ubuntu':
-        set_bin_files(data_files, dest=agent_bin_path)
         set_conf_files(data_files, src=["config/ubuntu/waagent.conf"])
         set_logrotate_files(data_files)
         set_udev_files(data_files)
         if version.startswith("12") or version.startswith("14"):
             # Ubuntu12.04/14.04 - uses upstart
+            if version.startswith("12"):
+                set_bin_files(data_files, dest=agent_bin_path)
+            else:
+                set_bin_files(data_files, dest=agent_bin_path,
+                              src=["bin/py3/waagent", "bin/waagent2.0"])
             set_files(data_files, dest="/etc/init",
                       src=["init/ubuntu/walinuxagent.conf"])
             set_files(data_files, dest='/etc/default',
                       src=['init/ubuntu/walinuxagent'])
-        elif fullname == 'Snappy Ubuntu Core':
-            set_files(data_files, dest="<TODO>",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -135,20 +135,23 @@ def get_data_files(name, version, fullname):  # pylint: disable=R0912
         set_systemd_files(data_files, dest=systemd_dir_path,
                           src=["init/mariner/waagent.service"])
     elif name == 'ubuntu':
-        set_bin_files(data_files, dest=agent_bin_path)
         set_conf_files(data_files, src=["config/ubuntu/waagent.conf"])
         set_logrotate_files(data_files)
         set_udev_files(data_files)
         if version.startswith("12") or version.startswith("14"):
             # Ubuntu12.04/14.04 - uses upstart
+            if version.startswith("12"):
```
narrieta, line 143: I'm thinking we should probably not change anything for ubuntu 14, just in case

The issue with Ubuntu 20 is that "python" is gone, but it should be there on 14

setup.py
```diff
@@ -188,7 +191,8 @@ def get_data_files(name, version, fullname):  # pylint: disable=R0912
         set_conf_files(data_files, src=["config/openbsd/waagent.conf"])
         set_openbsd_rc_files(data_files)
     elif name == 'debian':
-        set_bin_files(data_files, dest=agent_bin_path)
+        set_bin_files(data_files, dest=agent_bin_path,
+                      src=["bin/py3/waagent", "bin/waagent2.0"])
```
narrieta, line 195: same for debian, we should probably leave it alone... or was this needed?

setup.py
nagworld9: In the DCR, packer using py3 template from ubuntu 14 and debian too.  So we decided to have setup handle this for all distros that need it and remove this from our DCR provisioners.

nagworld9: Anyway we should have python until 20. So old agent file with python shebang still works. Are you asking only for 14 or until 20?

narrieta: I was asking for Ubuntu 14 only. So the DCR provisioner for ubuntu 14 was changing the shebang?

nagworld9: Yes we are.. Same template from ubuntu 14

setup.py
nagworld9: same like ubuntu.. all debian distros using py3 packer template

setup.py
narrieta: k

setup.py
narrieta: k

narrieta (approved):

kevinclark19a (approved):

PR #2289, opened 2021-06-30
Logcollector cgroups

kevinclark19a:
## Description

The log collector has the potential to be a heavy operation, and thus should be in its own cgroup. This PR kills the process if it is not in the expected cgroup.

---

### PR information

### DIFF
--- a/azurelinuxagent/agent.py
+++ b/azurelinuxagent/agent.py
@@ -208,8 +208,8 @@ class Agent(object):
 
             cpu_cgroup_path, memory_cgroup_path = SystemdCgroupsApi.get_process_cgroup_relative_paths("self")
 
-            cpu_regex_match = cgroup_path_regex.fullmatch(cpu_cgroup_path)
-            memory_regex_match = cgroup_path_regex.fullmatch(memory_cgroup_path)
+            cpu_regex_match = cgroup_path_regex.match(cpu_cgroup_path)
+            memory_regex_match = cgroup_path_regex.match(memory_cgroup_path)
             
             cpu_slice, cpu_unit = cpu_regex_match.group("slice", "unit") if cpu_regex_match else (None, None)
             memory_slice, memory_unit = memory_regex_match.group("slice", "unit") if memory_regex_match else (None, None)
--- a/azurelinuxagent/agent.py
+++ b/azurelinuxagent/agent.py
@@ -29,7 +29,7 @@ import subprocess
 import sys
 import threading
 import traceback
-from azurelinuxagent.common import logcollector
+from azurelinuxagent.common import cgroupconfigurator, logcollector
 from azurelinuxagent.common.cgroupapi import SystemdCgroupsApi
 
 import azurelinuxagent.common.conf as conf
@@ -215,7 +215,7 @@ class Agent(object):
                 
                 slice_group, unit_group = regex_match.group("slice", "unit")
 
-                slice_matches = (slice_group == logcollector.CGROUPS_SLICE)
+                slice_matches = (slice_group == cgroupconfigurator.AZURE_SLICE)
                 unit_matches = (unit_group == logcollector.CGROUPS_UNIT)
 
                 return slice_matches, unit_matches
--- a/azurelinuxagent/agent.py
+++ b/azurelinuxagent/agent.py
@@ -204,22 +204,38 @@ class Agent(object):
 
         # Check the cgroups unit
         if CollectLogsHandler.should_validate_cgroups():
-            cgroup_path_regex = re.compile(r'^(?P<slice>.*)/(?P<unit>.*)$')
+
+            def validate_cgroup_path(path):
+                if path is None:
+                    return False, False
+
+                regex_match = re.match(r'^(?P<slice>.*)/(?P<unit>.*)$', path)
+                if regex_match is None:
+                    return False, False
+                
+                slice, unit = regex_match.group("slice", "unit")
+
+                slice_matches = (slice == logcollector.CGROUPS_SLICE)
+                unit_matches = (unit == logcollector.CGROUPS_UNIT)
+
+                return slice_matches, unit_matches
 
             cpu_cgroup_path, memory_cgroup_path = SystemdCgroupsApi.get_process_cgroup_relative_paths("self")
 
-            cpu_regex_match = cgroup_path_regex.match(cpu_cgroup_path)
-            memory_regex_match = cgroup_path_regex.match(memory_cgroup_path)
-            
-            cpu_slice, cpu_unit = cpu_regex_match.group("slice", "unit") if cpu_regex_match else (None, None)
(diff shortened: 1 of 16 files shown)


narrieta:
@kevinclark19a could you update the title to be more descriptive? thanks

azurelinuxagent/ga/collect_logs.py
```diff
@@ -73,11 +73,27 @@ class CollectLogsHandler(ThreadHandlerInterface):
     """
 
     _THREAD_NAME = "CollectLogsHandler"
+    __CGROUPS_FLAG_ENV_VARIABLE = "_AZURE_GUEST_AGENT_LOG_COLLECTOR_MONITOR_CGROUPS_"
```
narrieta, line 76: Is the environment variable used to distinguish between calls from the agent and calls from the command line? If yes, a command line option on the log collector (e.g. --check-cgroup <foo>) may be more appropriate.

azurelinuxagent/agent.py
```diff
@@ -199,6 +202,26 @@ def collect_logs(self, is_full_mode):
         else:
             print("Running log collector mode normal")
 
+        # Check the cgroups unit
+        if CollectLogsHandler.should_validate_cgroups():
+            cgroup_path_regex = re.compile(r'^(?P<slice>.*)/(?P<unit>.*)$')
+
+            cpu_cgroup_path, memory_cgroup_path = SystemdCgroupsApi.get_process_cgroup_relative_paths("self")
```
narrieta, line 209: Note that get_process_cgroup_relative_paths can return None if, for example, the CPU and/or memory controllers are not enabled. You need to protect your code against this condition.

azurelinuxagent/agent.py
```diff
@@ -199,6 +202,26 @@ def collect_logs(self, is_full_mode):
...
+
+            cpu_regex_match = cgroup_path_regex.match(cpu_cgroup_path)
+            memory_regex_match = cgroup_path_regex.match(memory_cgroup_path)
+            
+            cpu_slice, cpu_unit = cpu_regex_match.group("slice", "unit") if cpu_regex_match else (None, None)
+            memory_slice, memory_unit = memory_regex_match.group("slice", "unit") if memory_regex_match else (None, None)
+
+            if cpu_slice != logcollector.CGROUPS_SLICE or memory_slice != logcollector.CGROUPS_SLICE:
+                print("Log Collector not in the expected slice, skipping execution.")
```
narrieta, line 218: since this condition can be triggered both by cpu or memory, we shoud add which condition failed to the message

azurelinuxagent/common/logcollector.py
```diff
@@ -43,6 +43,11 @@
 OUTPUT_RESULTS_FILE_PATH = os.path.join(_LOG_COLLECTOR_DIR, "results.txt")
 COMPRESSED_ARCHIVE_PATH = os.path.join(_LOG_COLLECTOR_DIR, "logs.zip")
 
+CGROUPS_SLICE = "azure.slice"
```
narrieta, line 46: you should use _AZURE_SLICE (from cgroupconfigurator.py) instead of redefining this slice

narrieta: (or, I have another comment in your PR) pass the slice name as a command-line parameter

azurelinuxagent/ga/collect_logs.py
```diff
@@ -143,13 +162,15 @@ def _get_resource_limits():
...
+    def _collect_logs(self):
         logger.info("Starting log collection...")
 
         # Invoke the command line tool in the agent to collect logs, with resource limits on CPU and memory (RAM).
-        scope_name = "collect-logs-{0}.scope".format(ustr(int(time.time() * 1000000)))
-        systemd_cmd = ["systemd-run", "--unit={0}".format(scope_name), "--scope"]
+
+        systemd_cmd = [
+            "systemd-run", "--unit={0}".format(logcollector.CGROUPS_UNIT),
+            "--slice={0}".format(logcollector.CGROUPS_SLICE),  "--scope"
```
narrieta, line 172: Instead of using a transient scope, I'd use a permanent slice. This has the advantage of allowing the user to override any of our settings with something that is more appropriate to their environment (you can see how the agent uses slices, or we can sync offline)

azurelinuxagent/ga/collect_logs.py
```diff
@@ -180,14 +201,20 @@ def exec_command(output_file):
                 return True
             except Exception as e:
                 duration = elapsed_milliseconds(start_time)
+                err_msg = ustr(e)
 
                 if isinstance(e, CommandError):
-                    exception_message = ustr("[stderr] %s", e.stderr)  # pylint: disable=no-member
-                else:
-                    exception_message = ustr(e)
+                    # pylint has limited (i.e. no) awareness of control flow w.r.t. typing. we disable=no-member
```
narrieta, line 207: great comment, thanks!

azurelinuxagent/ga/collect_logs.py
kevinclark19a: I'm a bit concerned with the complexity of the `azurelinuxagent.agent.parse_args` method... it doesn't quite seem tenable to keep adding flags in that manner. I do know there is an `argparse` in python3.2+, and a [PyPI package of the same name](https://pypi.org/project/argparse/) for earlier versions.

I'm not sure if it is worth switching over to this, and I'm equally unsure if my concern for adding to this `parse_args` method is warranted. Any thoughts?

((By the way, I believe this same method (of setting an environment variable) is used to pass the daemon version.))

azurelinuxagent/agent.py
kevinclark19a: the newest commit should handle this case. Let me know! thanks.

narrieta (approved):

azurelinuxagent/ga/collect_logs.py
```diff
@@ -143,13 +162,15 @@ def _get_resource_limits():
...
         logger.info("Starting log collection...")
 
         # Invoke the command line tool in the agent to collect logs, with resource limits on CPU and memory (RAM).
-        scope_name = "collect-logs-{0}.scope".format(ustr(int(time.time() * 1000000)))
-        systemd_cmd = ["systemd-run", "--unit={0}".format(scope_name), "--scope"]
+
+        systemd_cmd = [
+            "systemd-run", "--unit={0}".format(logcollector.CGROUPS_UNIT),
+            "--slice={0}".format(cgroupconfigurator.AZURE_SLICE)
+        ]
```
nagworld9, line 173: Is logcollector demon managed under azure.slice group. May be comment like [this ](https://github.com/Azure/WALinuxAgent/blob/b268011333a50e07db385824ffb963f975513890/azurelinuxagent/common/cgroupconfigurator.py#L254)helps

nagworld9 (approved):
LGTM.

PR #2294, opened 2021-07-02
Added support for vmSettings' ETag

narrieta:
Use the ETag of the vmSettings API; update extensions goal state only if the ETag changed.

### DIFF
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -35,6 +35,7 @@ from azurelinuxagent.common.event import add_event, WALAEventOperation, report_e
 from azurelinuxagent.common.exception import ProtocolNotFoundError, \
     ResourceGoneError, ExtensionDownloadError, InvalidContainerError, ProtocolError, HttpError
 from azurelinuxagent.common.future import httpclient, bytebuffer, ustr
+from azurelinuxagent.common.protocol.extensions_goal_state import ExtensionsGoalState
 from azurelinuxagent.common.protocol.goal_state import GoalState, TRANSPORT_CERT_FILE_NAME, TRANSPORT_PRV_FILE_NAME
 from azurelinuxagent.common.protocol.hostplugin import HostPluginProtocol
 from azurelinuxagent.common.protocol.restapi import DataContract, ExtHandlerPackage, \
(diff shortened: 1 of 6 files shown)


narrieta:
@jscalev  FYI

azurelinuxagent/common/protocol/wire.py
```diff
@@ -666,7 +670,7 @@ def fetch_manifest(self, version_uris, timeout_in_minutes=5, timeout_in_ms=0):
                 logger.verbose('The specified manifest URL is empty, ignored.')
                 continue
 
-            direct_func = lambda: self.fetch(version.uri, max_retry=1)  # pylint: disable=W0640
+            direct_func = lambda: self.fetch(version.uri, max_retry=1)[0]  # pylint: disable=W0640
             # NOTE: the host_func may be called after refreshing the goal state, be careful about any goal state data
             # in the lambda.
             host_func = lambda: self.fetch_manifest_through_host(version.uri)  # pylint: disable=W0640
```
kevinclark19a, line 676: There is a pitfall here that might be a good idea to document. `direct_func` and `host_func` both reference `version.uri`, but `version` is a mutating variable declared in a loop. Thus, after the loop, all the lambdas will use the last `version.uri` in `version_uris_shuffled`. See [the pylint page](http://pylint.pycqa.org/en/stable/technical_reference/features.html) for more info (just `Ctrl+F` for `W0640`). I believe this isn't an issue right now because we only access these lambdas from inside the loop, but any future code that somehow tries to export a particular `direct_func`/`host_func` would run into a pretty hard to debug error. Might be worth a code comment.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -829,6 +848,16 @@ def _set_host_plugin(self, new_host_plugin):
             logger.warn("Setting empty Host Plugin object!")
         self._host_plugin = new_host_plugin
 
+    def get_etag(self):
+        if self._extensions_goal_state is None:
+            return "0000000000"
```
kevinclark19a, line 853: The formatting choice here is a bit odd  to me. What's the reasoning for this number of `0`'s? 
Also, would `"0"` not work?

narrieta: Functionally, it doesn't release matter. The longer string is easier to stop in the debugger.

kevinclark19a (approved):
Left a single unresolved comment above about documenting a pitfall, but everything looks good to me, overall.

azurelinuxagent/common/protocol/wire.py
narrieta: We create a new closure in each iteration of the loop, so this is not an issue. I can add a comment explaining the suppression on a subsequent PR.

dhivyaganesan (approved):

PR #2295, opened 2021-07-06
Enable Periodic Log Collection in systemd distros

kevinclark19a:
## Description

Set the default configuration to enable periodic log collection.

---

### PR information


azurelinuxagent/ga/collect_logs.py
```diff
@@ -169,18 +169,12 @@ def _collect_logs(self):
 
         systemd_cmd = [
             "systemd-run", "--unit={0}".format(logcollector.CGROUPS_UNIT),
-            "--slice={0}".format(cgroupconfigurator.AZURE_SLICE)
+            "--slice={0}".format(cgroupconfigurator.LOGCOLLECTOR_SLICE), "--scope"
```
nagworld9, line 172: Just to make sure, are we managing logcollector under `azure.slice/azure-walinuxagent-periodic_logcollector.slice/collect-logs.scope` but what I read is systemd-run creates the scope under the system slice by default as per our implementation for extension slice
https://github.com/Azure/WALinuxAgent/blob/develop/azurelinuxagent/common/cgroupapi.py#L258. 
May be either defining slice name in contents or something like creating file under azure.slice should do the trick.

azurelinuxagent/agent.py
```diff
@@ -205,25 +205,36 @@ def collect_logs(self, is_full_mode):
...
+                unit_matches = (unit_group == expected_unit)
 
                 return slice_matches, unit_matches
 
             cpu_cgroup_path, memory_cgroup_path = SystemdCgroupsApi.get_process_cgroup_relative_paths("self")
 
-            cpu_slice_matches, cpu_unit_matches = validate_cgroup_path(cpu_cgroup_path)
-            memory_slice_matches, memory_unit_matches = validate_cgroup_path(memory_cgroup_path)
+            cpu_slice_matches, cpu_unit_matches = validate_cgroup_path(cpu_cgroup_path,
+                cgroupconfigurator.LOGCOLLECTOR_SLICE, None)
```
nagworld9, line 235: Any reason why `None` here?

narrieta: related to @nagworld9's question: why is there an asymmetry between the cpu & memory paths? (i.e. None vs CGROUPS_UNIT)

azurelinuxagent/agent.py
```diff
@@ -205,25 +205,36 @@ def collect_logs(self, is_full_mode):
         # Check the cgroups unit
         if CollectLogsHandler.should_validate_cgroups():
 
-            def validate_cgroup_path(path):
+            def validate_cgroup_path(path, expected_slice, expected_unit):
                 if path is None:
                     return False, False
 
-                regex_match = re.match(r'^(?P<slice>.*)/(?P<unit>.*)$', path)
+                # '(.*/)?': process slice may be nested in other, heirarchical slices.
```
narrieta, line 212: small typo: "heirarchical"

azurelinuxagent/agent.py
```diff
@@ -205,25 +205,36 @@ def collect_logs(self, is_full_mode):
...
+                # '[^\s./]*': process slice can't contain spaces, periods, or slashes.
```
narrieta, line 213: thanks for adding those comments. could you also add the slices we are expecting? i don't quite get why (.*/) is optional (in r'(.*/)?(?P<slice>[^\s./]*.slice)')

azurelinuxagent/agent.py
```diff
@@ -205,25 +205,36 @@ def collect_logs(self, is_full_mode):
...
                 return slice_matches, unit_matches
 
             cpu_cgroup_path, memory_cgroup_path = SystemdCgroupsApi.get_process_cgroup_relative_paths("self")
 
-            cpu_slice_matches, cpu_unit_matches = validate_cgroup_path(cpu_cgroup_path)
-            memory_slice_matches, memory_unit_matches = validate_cgroup_path(memory_cgroup_path)
+            cpu_slice_matches, cpu_unit_matches = validate_cgroup_path(cpu_cgroup_path,
+                cgroupconfigurator.LOGCOLLECTOR_SLICE, None)
+            memory_slice_matches, memory_unit_matches = validate_cgroup_path(memory_cgroup_path,
+                cgroupconfigurator.LOGCOLLECTOR_SLICE, logcollector.CGROUPS_UNIT)
```
narrieta, line 237: why using print? is this leftover debug code or should those messages go to the agent's log?

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -290,6 +305,12 @@ def __setup_azure_slice():
             if not os.path.exists(vmextensions_slice):
                 files_to_create.append((vmextensions_slice, _VMEXTENSIONS_SLICE_CONTENTS))
 
+            if not os.path.exists(logcollector_slice):
+                from azurelinuxagent.ga.collect_logs import CollectLogsHandler
```
narrieta, line 309: can we move this import to the top of the file?

azurelinuxagent/agent.py
kevinclark19a: Maybe the attempt was misguided, but the intent was to make this regex general to any cgroup slice/unit path. I.e. it's possible that our process starts up in `system.slice/tmp.scope` or something, in which case `.*/` (non-optional) would capture `system.slice`, and our regex groups would be off. I think the main benefit would be that we could more easily log expected vs. actual, but the code doesn't actually do that, so...

As for the happy path, I had thought we only  had to worry about `azure.slice/azure-walinuxagent.slice/azure-walinuxagent-periodic_logcollector.slice/collect-logs.scope`  (as that is what I expected from systemd as I understand it), but actually running the code on ubuntus 16, 18, and 20 show this seems to depend w.r.t. systemd version...

So far, I've seen:

ubuntu 20 (systemd v245):
* cpu path: `azure.slice/azure-walinuxagent.slice/azure-walinuxagent-periodic_logcollector.slice` (note the lack of the `*.scope`)
* memory path: `azure.slice/azure-walinuxagent.slice/azure-walinuxagent-periodic_logcollector.slice/collect-logs.scope`

ubuntu 18 (systemd v237, I think):
* cpu path: `azure.slice/azure-walinuxagent.slice/azure-walinuxagent-periodic_logcollector.slice/collect-logs.scope`
* memory path: `azure.slice/azure-walinuxagent.slice/azure-walinuxagent-periodic_logcollector.slice/collect-logs.scope`

azurelinuxagent/agent.py
kevinclark19a: `print`s get routed, undecorated, to the log file when invoked by the periodic task and (obviously) printed out to `stdout` on manual invocation. (so the second)

azurelinuxagent/common/cgroupconfigurator.py
kevinclark19a: I was worried about a cyclical import (as we import `cgroupconfigurator` into `ga.collect_logs`); I haven't actually tried it yet, though, so I can do so and update here.

azurelinuxagent/agent.py
kevinclark19a: @nagworld9, this is a reflection of `systemd v245`'s behavior on my ubuntu 20.04 dev VM; the path within the process cgroup file is `azure.slice/azure-walinuxagent.slice/azure-walinuxagent-periodic_logcollector.slice` (note the missing unit name, `collect-logs.scope`). It isn't universal though, unfortunately, as I don't believe this asymmetry exists on ubuntu 18.04 (off the top of my head, `systemd v237`). 

@nagworld9 & @narrieta: I need to do more research here as I'm not quite sure if this is just a configuration error, but barring that, we might need to abstract this check into `osutil` or something.

azurelinuxagent/ga/collect_logs.py
kevinclark19a: As per our offline sync, `systemd-run` is unfortunately a bit wild with how it generates the hierarchical slices: we're passing `azure-walinuxagent-periodic_logcollector.slice` as the slice name here, and `systemd-run` translates that to `azure.slice/azure-walinuxagent.slice/azure-walinuxagent-periodic_logcollector.slice` by using the `-`s in the original as delimiters.

azurelinuxagent/common/cgroupconfigurator.py
kevinclark19a: Update: this indeed does cause a cyclical import, so I'm inclined to leave it as is.

azurelinuxagent/agent.py
```diff
@@ -205,36 +205,49 @@ def collect_logs(self, is_full_mode):
...
                 if path is None:
-                    return False, False
+                    return False
 
-                regex_match = re.match(r'^(?P<slice>.*)/(?P<unit>.*)$', path)
+                expected_slice, expected_unit = cgroupconfigurator.LOGCOLLECTOR_SLICE, logcollector.CGROUPS_UNIT
+
+                # '(.*/)?': process slice may be nested in other, hierarchical slices.
+                # '[^\s./]*': process slice can't contain spaces, periods, or slashes.
+                slice_regex = r'(.*/)?(?P<slice>[^\s./]*.slice)'
```
narrieta, line 216: I don't think we need such a generic regex for this check. Would it be enough to check that collector is within cgroupconfigurator.LOGCOLLECTOR_SLICE?

azurelinuxagent/agent.py
```diff
@@ -205,36 +205,49 @@ def collect_logs(self, is_full_mode):
...
+                regex_match = re.match(path_regex, path)
                 if regex_match is None:
-                    return False, False
+                    return False
                 
                 slice_group, unit_group = regex_match.group("slice", "unit")
 
-                slice_matches = (slice_group == cgroupconfigurator.AZURE_SLICE)
-                unit_matches = (unit_group == logcollector.CGROUPS_UNIT)
+                if unit_group != expected_unit:
```
narrieta, line 229: If we just check against cgroupconfigurator.LOGCOLLECTOR_SLICE I think this check can also be omitted

narrieta: Or maybe I am misunderstanding the reason for these 2 checks? The first check silently returns False, while this emits telemetry.

narrieta: Note that the existing code for this function is more complicated because the collector was not on its own slice. Seems to me that most of this function can be replaced with a simple check for cgroupconfigurator.LOGCOLLECTOR_SLICE

kevinclark19a: If the process fails the first check, it should exit with INVALID_CGROUP_ERROR. The idea behind this check was to get telemetry for when we run into the abnormal `/proc/self/cgroup` content.

narrieta: what is the first check doing?

narrieta: never mind, misread the code... this is a helper function...

but i still think this is the case:

Note that the existing code for this function is more complicated because the collector was not on its own slice. Seems to me that most of this function can be replaced with a simple check for cgroupconfigurator.LOGCOLLECTOR_SLICE

azurelinuxagent/common/cgroupconfigurator.py
narrieta: OK, but coming to think about it, the CGroupConfigurator should not have a dependency on the CollectLogsHandler.

Seems like you just need the value of a few constants? Maybe move those to this module, or move them to a module that both CGroupConfigurator and CollectLogsHandler depend on?

narrieta (dismissed):

narrieta (approved):

PR #2297, opened 2021-07-07
Use If-None-Match for ETag header

narrieta:
The header in the request should have been "If-None-Match" instead of 'ETag' (it is 'ETag' only in the response.)

Also, per Rohit's recommendation, do not send the "If-None-Match" header altogether, instead of sending a dummy etag (all zeroes).


narrieta:
FYI @jscalev

narrieta:
closing for now, i need a few more changes

PR #2298, opened 2021-07-09
Do not create placeholder status file for AKS extensions

larohra:
## Description

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -735,7 +735,8 @@ class ExtHandlersHandler(object):
         # failures back to CRP. If a placeholder for an extension already exists with Transitioning status, we would
         # not override it, hence we only create a placeholder for enable/disable commands but the extensions have the
         # data to create their own if needed.
-        ext_handler_i.create_placeholder_status_file(extension)
+        if ext_handler_i.should_create_default_placeholder(extension):
+            ext_handler_i.create_placeholder_status_file(extension)
         self.__handle_extension(ext_handler_i, extension, uninstall_exit_code)
 
     @staticmethod
@@ -1397,6 +1398,17 @@ class ExtHandlerInstance(object):
 
         # Save HandlerEnvironment.json
         self.create_handler_env()
+
+    def should_create_default_placeholder(self, extension=None):
+        """
+        There's a bug in the AKS extension where they dont update the status file if it exists.
+        This violates the contract we have with extensions. Until they fix their extension,
+        we're going to skip creating a placeholder for them to ensure they dont have any downtime.
+        For all other extensions, we should create a
+        """
+
+        ignore_extension_regex = "Microsoft\\.AKS\\.Compute\\.AKS\\S*"
+        return re.match(ignore_extension_regex, self.get_extension_full_name(extension)) is None
 
     def create_placeholder_status_file(self, extension=None, status=ValidHandlerStatus.transitioning, code=0,
                                        operation="Enabling Extension", message="Install/Enable is in progress."):
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -1407,7 +1407,7 @@ class ExtHandlerInstance(object):
         For all other extensions, we should create a
         """
 
-        ignore_extension_regex = "Microsoft\\.AKS\\.Compute\\.AKS\\S*"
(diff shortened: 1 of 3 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -1398,6 +1399,17 @@ def initialize(self):
...
 
+    def should_create_default_placeholder(self, extension=None):
+        """
+        There's a bug in the AKS extension where they dont update the status file if it exists.
+        This violates the contract we have with extensions. Until they fix their extension,
+        we're going to skip creating a placeholder for them to ensure they dont have any downtime.
+        For all other extensions, we should create a
+        """
+
+        ignore_extension_regex = "Microsoft\\.AKS\\.Compute\\.AKS\\S*"
```
narrieta, line 1410: minor: using a raw string may make the regex slighly more readable: r"Microsoft\.AKS\.Compute\.AKS\S*"

azurelinuxagent/ga/exthandlers.py
```diff
@@ -735,7 +735,8 @@ def handle_enable(self, ext_handler_i, extension):
         # failures back to CRP. If a placeholder for an extension already exists with Transitioning status, we would
         # not override it, hence we only create a placeholder for enable/disable commands but the extensions have the
         # data to create their own if needed.
-        ext_handler_i.create_placeholder_status_file(extension)
+        if ext_handler_i.should_create_default_placeholder(extension):
+            ext_handler_i.create_placeholder_status_file(extension)
```
narrieta, line 739: create_placerholder_status_file also has some logic as to whether the file should be created or not... would be better to have those conditions in the same place as the AKS condition

if status_path is not None and not os.path.exists(status_path):

tests/data/wire/ext_conf_aks_extension.xml
```diff
@@ -0,0 +1,69 @@
...
+                {
+                "handlerSettings": {
+                "publicSettings": {"message": "Enabling AKSNode"}
+                }
+                }
+                ]
+                }
+            </RuntimeSettings>
+        </Plugin>
+        <Plugin name="Microsoft.AKS.Compute.AKS-Engine.Linux.Billing" version="1.0.0">
```
narrieta, line 51: cool

narrieta (approved):
minor comments, lgtm

azurelinuxagent/ga/exthandlers.py
larohra: Done

azurelinuxagent/ga/exthandlers.py
larohra: The `create_placeholder_file` function is used in multiple places where we need to create a placeholder for error reporting purposes. Its mostly used only for multi-config as of now, all the other places have the multi-config only check in place. 

This was the only place where it was being used for both single and multi-config extensions. That's the reason I put the check here and not inside that function

tests/ga/test_extension.py
```diff
@@ -1664,6 +1664,56 @@ def mock_popen(cmd, *args, **kwargs):
                                             expected_msg="Dependent Extension OSTCExtensions.OtherExampleHandlerLinux did not reach a terminal state within the allowed timeout. Last status was {0}".format(
                                                 ValidHandlerStatus.warning))
 
+    def test_it_should_not_create_placeholder_for_aks_extension(self, mock_http_get, mock_crypt_util, *args):
```
dhivyaganesan, line 1667: Just a FYI, As a part of new PR which recently got merged all new test cases should be a part of the new TestExtension class. It was created to use the newer 'mock_wire_protocol' . Since this is a release branch this shouldn't be an issue.

larohra: Thanks for the update (and also for creating that new class). 
Since this is on the release branch, it still doesn't have that new class in here so I just added it to the old class. I'll make sure to add new tests to the new class moving forward though :)

dhivyaganesan (approved):

azurelinuxagent/ga/exthandlers.py
narrieta: seems like different paths do different validations? i'd consider simplifying that unless those validations are very specific to one path (which I do not think should be the case for the ask check)

narrieta (approved):

azurelinuxagent/ga/exthandlers.py
larohra: synced offline, gonna keep as is for now.

PR #2304, opened 2021-07-14
Use If-None-Match for ETag header

narrieta:
* Fixes the name of the header: If-None-Match instead of ETag header.
* Update goal state only if the response includes an etag


narrieta:
FYI: @jscalev

larohra (approved):

kevinclark19a (approved):

PR #2323, opened 2021-08-11
Update HostGAplugin headers before fetching vmSettings

narrieta:
The vmSettings were out of sync with ExtensionsConfig because the vmSettings API requires the latest RoleConfigName so the headers need to be updated before issuing the request.

With this change the retry loop is no longer needed so I am removing it as well.


kevinclark19a (approved):

azurelinuxagent/common/protocol/wire.py
```diff
@@ -789,58 +789,50 @@ def update_goal_state(self, forced=False):
...
+            response_etag = None
+            for h in response_headers:
+                if h[0].lower() == 'etag':
+                    response_etag = h[1]
                     break
 
+            # TODO: Remove this log statement when invoking this method on each iteration of the goal state loop (currently it is invoked only on a new goal state)
+            logger.info("Fetched extensions goal state [correlation ID: {0} eTag: {1}]", correlation_id, response_etag)
+
+            logger.info("Fetched new extensions goal state [correlation ID: {0} eTag: {1}]", correlation_id, response_etag)
```
larohra, line 833: Any reason for logging the same data twice? (Line 831 and 833)?

narrieta: 831 is no longer needed now that the retry loop is gone, let me remove it

kevinclark19a (approved):

larohra (approved):

kevinclark19a (approved):

larohra (approved):

PR #2332, opened 2021-08-18
Implement InitialGoalStatePeriod parameter + improvements in logging goal state processing

narrieta:
Implementation of Extensions.InitialGoalStatePeriod in /etc/waagent.conf.

In order to implement this parameter, I need to keep track of the convergence of the first goal state so I took this as an opportunity to improve the logging of the status of the goal state,

### DIFF
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -56,7 +56,7 @@ from azurelinuxagent.ga.collect_logs import get_collect_logs_handler, is_log_col
 from azurelinuxagent.ga.env import get_env_handler
 from azurelinuxagent.ga.collect_telemetry_events import get_collect_telemetry_events_handler
 
-from azurelinuxagent.ga.exthandlers import HandlerManifest, ExtHandlersHandler, list_agent_lib_directory
+from azurelinuxagent.ga.exthandlers import HandlerManifest, ExtHandlersHandler, list_agent_lib_directory, ValidHandlerStatus
 from azurelinuxagent.ga.monitor import get_monitor_handler
 
 from azurelinuxagent.ga.send_telemetry_events import get_send_telemetry_events_handler
@@ -72,12 +72,17 @@ CHILD_POLL_INTERVAL = 60
 
 MAX_FAILURE = 3  # Max failure allowed for agent before blacklisted
 
-GOAL_STATE_INTERVAL_DISABLED = 5 * 60
+GOAL_STATE_PERIOD_EXTENSIONS_DISABLED = 5 * 60
 
 ORPHAN_POLL_INTERVAL = 3
 ORPHAN_WAIT_INTERVAL = 15 * 60
 
 AGENT_SENTINEL_FILE = "current_version"
+
+# This file marks that the first goal state (after provisioning) has been completed, either because it converged or because we received another goal
+# state before it converged. The contents will be an instance of ExtensionsSummary. If the file does not exist then we have not finished processing
+# the goal state.
+INITIAL_GOAL_STATE_FILE = "initial_goal_state"
 
 READONLY_FILE_GLOBS = [
     "*.crt",
@@ -86,6 +91,29 @@ READONLY_FILE_GLOBS = [
     "*.prv",
     "ovf-env.xml"
 ]
+
+
+class ExtensionsSummary(object):
+    """
+    The extensions summary is a list of (extension name, extension status) tuples for the current goal state; it is
+    used to report changes in the status of extensions and to keep track of when the goal state converges (i.e. when
+    all extensions in the goal state reach a terminal state: success or error.)
+    The summary is computed from the VmStatus reported to blob storage.
+    """
+    def __init__(self, vm_status=None):
+        if vm_status is None:
+            self.summary = []
+            self.converged = True
+        else:
+            self.summary = [(h.extension_status.name, h.extension_status.status) for h in vm_status.vmAgent.extensionHandlers]
+            self.summary.sort(key=lambda s: s[0])  # sort by extension name to make comparisons easier
+            self.converged = all(status in (ValidHandlerStatus.success, ValidHandlerStatus.error) for _, status in self.summary)
+
+    def __eq__(self, other):
+        return self.summary == other.summary
+
+    def __str__(self):
+        return ustr(self.summary)
 
 
 def get_update_handler():
@@ -118,6 +146,18 @@ class UpdateHandler(object):
         self._last_try_update_goal_state_failed = False
 
         self.last_incarnation = None
+
+        self._extensions_summary = ExtensionsSummary()
+
+        self._is_initial_goal_state = not os.path.exists(self._initial_goal_state_file_path())
+
+        if not conf.get_extensions_enabled():
+            self._goal_state_period = GOAL_STATE_PERIOD_EXTENSIONS_DISABLED
+        else:
+            if self._is_initial_goal_state:
+                self._goal_state_period = conf.get_initial_goal_state_period()
+            else:
+                self._goal_state_period = conf.get_goal_state_period()
 
     def run_latest(self, child_args=None):
         """
@@ -316,14 +356,14 @@ class UpdateHandler(object):
             for thread_handler in all_thread_handlers:
                 thread_handler.run()
 
-            goal_state_interval = conf.get_goal_state_period() if conf.get_extensions_enabled() else GOAL_STATE_INTERVAL_DISABLED
+            logger.info("Goal State Period: {0} sec. This indicates how often the agent checks for new goal states and reports status.", self._goal_state_period)
 
             while self.is_running:
                 self._check_daemon_running(debug)
                 self._check_threads_running(all_thread_handlers)
                 self._process_goal_state(protocol, exthandlers_handler, remote_access_handler)
                 self._send_heartbeat_telemetry(protocol)
-                time.sleep(goal_state_interval)
+                time.sleep(self._goal_state_period)
 
         except ExitException as exitException:
             logger.info(exitException.reason)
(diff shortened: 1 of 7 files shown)


azurelinuxagent/ga/update.py
```diff
@@ -119,6 +147,18 @@ def __init__(self):
 
         self.last_incarnation = None
 
+        self._extensions_summary = ExtensionsSummary()
+
+        self._is_initial_goal_state = not os.path.exists(self._initial_goal_state_file_path())
+
+        if not conf.get_extensions_enabled():
```
narrieta, line 154: This initialization used to happen just before the main loop, and the period was a local variable. Now the period is an instance member and is initialized here (plus it takes into account the InitialGoalStatePeriod)

tests/common/test_cgroupconfigurator.py
```diff
@@ -189,15 +189,6 @@ def test_setup_extension_slice_should_create_unit_files(self):
...
             self.assertTrue(os.path.exists(extension_slice_unit_file), "{0} was not created".format(extension_slice_unit_file))
 
-    def test_remove_extension_slice_should_remove_unit_files(self):
-        with self._get_cgroup_configurator() as configurator:
-            # get the paths to the mocked files
-            extension_slice_unit_file = configurator.mocks.get_mapped_path(UnitFilePaths.extensionslice)
-
-            configurator.remove_extension_slice(extension_name="Microsoft.CPlat.Extension")
-
-            self.assertFalse(os.path.exists(extension_slice_unit_file), "{0} should not be present".format(extension_slice_unit_file))
```
kevinclark19a, line 199: why remove this test? is it unneeded now due to a change in behavior?

azurelinuxagent/ga/update.py
```diff
@@ -464,19 +535,22 @@ def _emit_restart_event(self):
     def _emit_changes_in_default_configuration():
         try:
             def log_event(msg):
-                logger.info(msg)
+                logger.info("******** {0} ********", msg)
```
narrieta, line 538: we had an incident where the GoalStatePeriod was changed, but we did not notice the log statement... I tried to do this message more visible by adding asterisks here and an explanatory note a few lines below

tests/common/test_cgroupconfigurator.py
```diff
@@ -189,15 +189,6 @@ def test_setup_extension_slice_should_create_unit_files(self):
 
             self.assertTrue(os.path.exists(extension_slice_unit_file), "{0} was not created".format(extension_slice_unit_file))
 
-    def test_remove_extension_slice_should_remove_unit_files(self):
```
narrieta, line 192: I need to undo this (i.e. re-add this test)

tests/ga/test_update.py
```diff
@@ -1501,13 +1501,21 @@ def test_update_happens_when_extensions_disabled(self, _):
...
-    def test_interval_changes_when_extensions_disabled(self, _):
+    def test_interval_should_be_default_when_extensions_enabled(self):
         """
         When extension processing is disabled, the goal state interval should be larger.
         """
-        self.update_handler._upgrade_available = Mock(return_value=False)
-        self._test_run(invocations=1, calls=1, sleep_interval=(300,))
+        with patch('azurelinuxagent.common.conf.get_extensions_enabled', return_value=True):
+            update_handler = get_update_handler()
+            self.assertEqual(6, update_handler._goal_state_period, "Incorrect goal state period when extensions are enabled")
```
kevinclark19a, line 1510: Could this reference a constant variable instead of a literal? perhaps just call the `conf` module itself?

tests/ga/test_update.py
```diff
@@ -1501,13 +1501,21 @@ def test_update_happens_when_extensions_disabled(self, _):
...
+
+    def test_interval_changes_when_extensions_disabled(self):
+        """
+        When extension processing is disabled, the goal state interval should be larger.
+        """
+        with patch('azurelinuxagent.common.conf.get_extensions_enabled', return_value=False):
+            update_handler = get_update_handler()
+            self.assertEqual(300, update_handler._goal_state_period, "Incorrect goal state period when extensions are disabled")
```
kevinclark19a, line 1518: same here as above.

tests/ga/test_update.py
narrieta: this test should actually be checking that we are using the default value (which happens to be 6), and the one below should verify that we are using the constant defined in update.py, but I am not planning on changing those existing tests as part of this work item. i noticed an important test case and added it here, not planning any additional improvements

narrieta: (I need to update the comment describing the new test, though, so I'll do that now

narrieta: (updated comment)

azurelinuxagent/ga/update.py
```diff
@@ -392,17 +432,48 @@ def _process_goal_state(self, protocol, exthandlers_handler, remote_access_handl
...
+                add_event(op=WALAEventOperation.GoalState, message=message)
+                if self._is_initial_goal_state:
+                    self._on_initial_goal_state_completed(self._extensions_summary)
+
+    def _on_initial_goal_state_completed(self, extensions_summary):
+        fileutil.write_file(self._initial_goal_state_file_path(), ustr(extensions_summary))
+        if conf.get_extensions_enabled():
+            previous = self._goal_state_period
+            self._goal_state_period = conf.get_goal_state_period()
+            if self._goal_state_period != previous:
```
nagworld9, line 473: why this check matters, regardless we need to switch right?

kevinclark19a (approved):

azurelinuxagent/ga/update.py
narrieta: we want to log this only 1 time (the switch always happens, one line above, though from the second goal state forward we are "switching" to the same value)

azurelinuxagent/ga/update.py
```diff
@@ -88,6 +93,29 @@
...
+    all extensions in the goal state reach a terminal state: success or error.)
+    The summary is computed from the VmStatus reported to blob storage.
+    """
+    def __init__(self, vm_status=None):
+        if vm_status is None:
+            self.summary = []
+            self.converged = True
+        else:
+            self.summary = [(h.extension_status.name, h.extension_status.status) for h in vm_status.vmAgent.extensionHandlers]
+            self.summary.sort(key=lambda s: s[0])  # sort by extension name to make comparisons easier
```
larohra, line 109: This will throw if the name is None. 
It generally shouldn't be None but would be safer to have a check

azurelinuxagent/ga/update.py
narrieta: alternatively, the code could have been written as 

```
        if conf.get_extensions_enabled() and conf.get_goal_state_period() != conf.get_initial_goal_state_period():
            self._goal_state_period = conf.get_goal_state_period()
            logger.info("Initial goal state completed, switched the goal state period to {0}", self._goal_state_period)

```

azurelinuxagent/ga/update.py
narrieta: how can the extension name be None?

azurelinuxagent/ga/update.py
nagworld9: This bring me general question, Ideal world it should not be but what if both first goal state period and goal state period values are same then do we want to go through this entire new flow?

narrieta: Not sure I follow you question... in the vast majority of VMs the 2 periods will be the same. On HPC machines they will be different. What is your concern?

nagworld9: I thought something else. Makes sense now. NM,

narrieta: ok, i'll rewrite the code as i pointed out above, which seems a little clearer

azurelinuxagent/ga/update.py
narrieta: i guess my point is that name is not special, why check for name not to be None, but not check status or vmAgent, or extensionHandlers?

rather, we should ensure we handle exceptions correctly if at some point a bug is introduced that makes any of those fields to be undefined. i realized the new UpdateHandler._report_status should handle errors in a similar way to ExtHandlersHandler.report_ext_handlers_status; I'll add that.

azurelinuxagent/ga/update.py
```diff
@@ -392,17 +436,58 @@ def _process_goal_state(self, protocol, exthandlers_handler, remote_access_handl
...
             if incarnation != self.last_incarnation:
                 remote_access_handler.run()
         finally:
             self.last_incarnation = incarnation
 
+    def _report_status(self, exthandlers_handler, incarnation_changed):
+        # report_ext_handlers_status does its own error handling and returns None if an error occurred
+        vm_status = exthandlers_handler.report_ext_handlers_status(incarnation_changed=incarnation_changed)
+        if vm_status is None:
+            return
```
larohra, line 461: Should we modify the self._extension_summary in this case? or maybe log it once per incarnation or something?

azurelinuxagent/ga/update.py
```diff
@@ -464,19 +549,22 @@ def _emit_restart_event(self):
...
-            def log_if_int_changed_from_default(name, current):
+            def log_if_int_changed_from_default(name, current, message=""):
                 default = conf.get_int_default_value(name)
                 if default != current:
-                    log_event("{0} changed from its default: {1}. New value: {2}".format(name, default, current))
+                    log_event("{0} changed from its default: {1}. New value: {2}. {3}".format(name, default, current, message))
 
             def log_if_op_disabled(name, value):
                 if not value:
                     log_event("{0} is set to False, not processing the operation".format(name))
```
larohra, line 562: Maybe even add a message for these ops too? (EnableFirewall and Extensions.Enabled)?

larohra (approved):
some minor comments else LGTM

azurelinuxagent/ga/update.py
narrieta: we want to log it every time something changes, and changes can happen on every call to exthandlers_handler.report_ext_handlers_status

note that this logging is more for debugging the goal state, it is not related to the initial goal state logic (though it build on the same infrastructure)

kevinclark19a (approved):

PR #2340, opened 2021-08-24
Save the original subprocess.Popen before mocking it

narrieta:
The firewalll tests need  the value of subprocess.Popen before other tests mock it (in case those tests run in parallel)

### DIFF
--- a/tests/ga/test_update.py
+++ b/tests/ga/test_update.py
@@ -19,6 +19,8 @@ import zipfile
 from datetime import datetime, timedelta
 from threading import currentThread
 
+_ORIGINAL_POPEN = subprocess.Popen
+
 from mock import PropertyMock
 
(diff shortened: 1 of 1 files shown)


tests/ga/test_update.py
```diff
@@ -1105,7 +1107,7 @@ def _test_run_latest(self, mock_child=None, mock_time=None, child_args=None):
         if mock_time is None:
             mock_time = TimeMock()
 
-        with patch('subprocess.Popen', return_value=mock_child) as mock_popen:
+        with patch('azurelinuxagent.ga.update.subprocess.Popen', return_value=mock_child) as mock_popen:
```
narrieta, line 1110: this doesn't change the behavior of the tests, but is more explicit as to what we intent to mock

PR #2342, opened 2021-08-27
Refactor update_goal_state(); always query vmSettings; update vmSettings test data

narrieta:
I did some refactoring on the logic to update the goal state. Before fast track it was a simple check on incarnation. Fast track requires additional logic; I'll complete that logic on a separate PR, for now I am moving the current logic from GoalState to WireClient.

In addition, with this PR now we query the vmSettings on each iteration of the main loop.

Lastly, I updated the test data files for vmSettings to include data for Multi-Config Extensions.

### DIFF
--- a/azurelinuxagent/common/protocol/goal_state.py
+++ b/azurelinuxagent/common/protocol/goal_state.py
@@ -49,19 +49,12 @@ _NUM_GS_FETCH_RETRIES = 6
 
 
 class GoalState(object):
-    def __init__(self, wire_client, full_goal_state=False, base_incarnation=None):
+    def __init__(self, wire_client):
         """
         Fetches the goal state using the given wire client.
 
-        By default it fetches only the goal state itself; to fetch the entire goal state (that includes all the
-        nested components, such as the extension config) use the 'full_goal_state' parameter.
-
-        If 'base_incarnation' is given, it fetches the full goal state if the new incarnation is different than
-        the given value, otherwise it fetches only the goal state itself.
-
-        For better code readability, use the static fetch_* methods below instead of instantiating GoalState
-        directly.
-
+        __init__ fetches only the goal state itself, not including inner properties such as ExtensionsConfig; to fetch the entire goal state
+        use the fetch_full_goal_state().
         """
         uri = GOAL_STATE_URI.format(wire_client.get_endpoint())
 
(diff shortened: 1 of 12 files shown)


azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -78,91 +71,59 @@ def __init__(self, wire_client, full_goal_state=False, base_incarnation=None):
...
+            xml_text = wire_client.fetch_config(self._shared_conf_uri, wire_client.get_header())
             self.shared_conf = SharedConfig(xml_text)
 
-            uri = findtext(xml_doc, "Certificates")
-            if uri is None:
-                self.certs = None
-            else:
-                xml_text = wire_client.fetch_config(uri, wire_client.get_header_for_cert())
+            if self._certs_uri is not None:
+                xml_text = wire_client.fetch_config(self._certs_uri, wire_client.get_header_for_cert())
```
kevinclark19a, line 109: w.r.t. my comment above, do we know why the old code had a guard on the `fetch_config` call specifically for Certificates? was it just that we had actually experienced a value of `None` for Certificates in the wild, and not yet for the others, or is there some fundamental difference between Certificates and the other uris, like Certificates being explicitly optional?

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -78,91 +71,59 @@ def __init__(self, wire_client, full_goal_state=False, base_incarnation=None):
...
             raise ProtocolError(msg="Error fetching goal state", inner=exception)
 
+    def fetch_full_goal_state(self, wire_client):
         try:
-            logger.info('Fetching new goal state [incarnation {0} ({1})]', self.incarnation, reason)
+            logger.info('Fetching goal state [incarnation {0}]', self.incarnation)
 
-            uri = findtext(xml_doc, "HostingEnvironmentConfig")
-            xml_text = wire_client.fetch_config(uri, wire_client.get_header())
+            xml_text = wire_client.fetch_config(self._hosting_env_uri, wire_client.get_header())
```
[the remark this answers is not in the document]
narrieta: The refactoring does not affect the operations we do, I just moved the code to a different place. We will still fetch the full goal state on service start, incarnation change or explicit force == True.

At first glance it seems like the code would raise an exception on null uri, but I do not want to change any error handling in this code, at  least not as part of the work for fast track,

azurelinuxagent/common/protocol/goal_state.py
narrieta: The exiting code assumes some of those fields are mandatory and some optional. All this is underspecified at best and not spec'ed most of the time. I do not want to change any error handling in this code, at least not as part of the work for fast track,

kevinclark19a (approved):
LGTM

azurelinuxagent/common/protocol/wire.py
```diff
@@ -774,36 +771,45 @@ def update_host_plugin_from_goal_state(self):
...
+                    updated = True
+
+            if updated:
                 self._save_goal_state()
 
         except Exception as exception:
             raise ProtocolError("Error fetching goal state: {0}".format(ustr(exception)))
 
-    def update_extensions_goal_state(self):
+    def _update_extensions_goal_state(self, force_update):
```
larohra, line 808: Sometimes I read this as "update_extension_config" and get confused. Is there a more explicit name that we can give this? Something like "fetch_extension_config_via_fast_track" or something?

tests/data/hostgaplugin/ext_conf.xml
```diff
@@ -1,130 +1,81 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Extensions version="1.0.0.0" goalStateIncarnation="1"><GuestAgentExtension xmlns:i="http://www.w3.org/2001/XMLSchema-instance">
+<Extensions version="1.0.0.0" goalStateIncarnation="10"><GuestAgentExtension xmlns:i="http://www.w3.org/2001/XMLSchema-instance">
```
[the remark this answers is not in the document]
larohra: Thanks for updating this. Just FYI for future tests, we also have a whole suite of MultiConfig test files for testing under - tests/data/wire/multi-config

larohra (approved):
some minor comments else LGTM

azurelinuxagent/common/protocol/wire.py
narrieta: Yup, this should become clearer on the next (or next+1) PR. In terms of queries, we need to get the vmSettings from hostgaplugin and/or the ExtensionsConfig from wireserver. We need a common representation for both, that will be named ExtensionsGoalState, which will be updated by update_extensions_config, which underneath will invoke fetch_vm_settings and fetch_extensions_config.

PR #2350, opened 2021-09-02
Retrieve full goal state explicitly

narrieta:
UpdateHandler.run() always fetches a full goal state, since it makes the first call to update_goal_state in the extension handler process. I made the call to update_goal_state  to explicitly request a full goal state.

A couple of unit tests failed since the make multiple calls to run() using the same instance of UpdateHandler, so they were never retrieving the full goal state after the first call. Reusing the same instance of the UpdateHandler would never happen in the actual agent.

### DIFF
--- a/tests/ga/test_update.py
+++ b/tests/ga/test_update.py
@@ -43,7 +43,7 @@ from azurelinuxagent.common.version import AGENT_PKG_GLOB, AGENT_DIR_GLOB, AGENT
 from azurelinuxagent.ga.exthandlers import ExtHandlersHandler, ExtHandlerInstance, HandlerEnvironment, ValidHandlerStatus
 from azurelinuxagent.ga.update import GuestAgent, GuestAgentError, MAX_FAILURE, AGENT_MANIFEST_FILE, \
     get_update_handler, ORPHAN_POLL_INTERVAL, AGENT_PARTITION_FILE, AGENT_ERROR_FILE, ORPHAN_WAIT_INTERVAL, \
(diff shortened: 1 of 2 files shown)


kevinclark19a (approved):

tests/ga/test_update.py
```diff
@@ -1597,6 +1600,15 @@ def test_it_should_recreate_handler_env_on_service_startup(self):
...
                           "{0} not found in HandlerEnv file".format(HandlerEnvironment.eventsFolder))
 
+    def _add_write_permission_to_goal_state_files(self):
+        # UpdateHandler.run() marks some of the files from the goal state as read-only. Those files are overwritten when
+        # a new goal state is fetched. This is not a problem for the agent, since it  runs as root, but tests need
+        # to make those files writtable before fetching a new goal state. Note that UpdateHandler.run() fetches a new
+        # goal state, so tests that make multiple calls to that method need to call this function in-between calls.
+        for gb in READONLY_FILE_GLOBS:
+            for path in glob.iglob(os.path.join(conf.get_lib_dir(), gb)):
+                fileutil.chmod(path, stat.S_IRUSR | stat.S_IWUSR)
```
larohra, line 1610: Out of curiosity, if the tests are not running as root, how are we able to change their permissions? Can the same user change permissions for a read-only file if non-root?

larohra (approved):

tests/ga/test_update.py
narrieta: Yup, setting the read bit does not affect chmod, so all ok with the tests

PR #2358, opened 2021-09-17
Use handler status if extension status is None when computing the Ext…

narrieta:
If an extension doesn't have runtimeSettings, it's status won't have an extension status. In that case, we use the handler status instead.

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -87,7 +87,19 @@ _STATUS_FILE_RETRY_DELAY = 2  # seconds
 # This is the default sequence number we use when there are no settings available for Handlers
 _DEFAULT_SEQ_NO = "0"
 
+
+class HandlerStatus(object):
+    """
+    Statuses for Handlers
+    """
+    ready = "Ready"
+    not_ready = "NotReady"
+
+
 class ValidHandlerStatus(object):
+    """
+    Statuses for Extensions
+    """
     transitioning = "transitioning"
     warning = "warning"
     error = "error"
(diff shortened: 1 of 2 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -87,7 +87,19 @@
...
+
+class HandlerStatus(object):
+    """
+    Statuses for Handlers
+    """
+    ready = "Ready"
+    not_ready = "NotReady"
+
+
 class ValidHandlerStatus(object):
```
larohra, line 99: NIT: `HandlerStatus` and `ValidHandlerStatus` sounds very confusing. Maybe change the name of this class to `ExtensionStatus` or `ExtensionRuntimeStatus` or something to make it more clearer?

azurelinuxagent/ga/update.py
```diff
@@ -105,9 +105,10 @@ def __init__(self, vm_status=None):
             self.summary = []
             self.converged = True
         else:
-            self.summary = [(h.extension_status.name, h.extension_status.status) for h in vm_status.vmAgent.extensionHandlers]
+            # take the name and status of the extension if is it not None, else use the handler's
+            self.summary = [(o.name, o.status) for o in map(lambda h: h.extension_status if h.extension_status is not None else h, vm_status.vmAgent.extensionHandlers)]
             self.summary.sort(key=lambda s: s[0])  # sort by extension name to make comparisons easier
-            self.converged = all(status in (ValidHandlerStatus.success, ValidHandlerStatus.error) for _, status in self.summary)
+            self.converged = all(status in (ValidHandlerStatus.success, ValidHandlerStatus.error, HandlerStatus.ready) for _, status in self.summary)
```
larohra, line 111: When looking at HandlerStatus, both Ready and NotReady imply that the Handler is `Converged`. They're just used to identify if the handler failed or not, but it still means that it converged. You should add `HandlerStatus.not_ready` here too

azurelinuxagent/ga/exthandlers.py
narrieta: Yes, but I plan to do that on a separate PR on develop. For now, i'm keeping it minimalistic for the release branch.

azurelinuxagent/ga/update.py
narrieta: yes, good point

larohra (approved):

dhivyaganesan (approved):

Issue #1673, opened 2019-10-15
"OS.EnableFirewall=y" is blocking DNS queries if a switch to TCP is required


malachma:
SHORT DESCRIPTION
------------------------
The current implementation of the following security rule is to restrictive
```
iptables -L -t security
target     prot opt source               destination
ACCEPT     tcp  --  anywhere             168.63.129.16        owner UID match root
DROP       tcp  --  anywhere             168.63.129.16        ctstate INVALID,NEW
```

it does block any DNS query traffic if a switch from UDP to TCP is necessary if a large payload has to to be returned to client initiating the DNS reolving process.

HOW TO REPRODUCE
-------------------------
Log on to a Linux Azure VM and run the following query
--> dig  aerserv-bc-us-east.bidswitch.net @168.63.129.16
which will result in the following response
```
;; Truncated, retrying in TCP mode.

; <<>> DiG 9.9.9-P1 <<>> aerserv-bc-us-east.bidswitch.net @168.63.129.16
;; global options: +cmd
;; connection timed out; no servers could be reached
```
The reason for this behaviour is the rule which allows TCP connections against 168.63.129.16
only for the root user!

Run the same query again but this time as root
#dig  aerserv-bc-us-east.bidswitch.net @168.63.129.16
```
;; Truncated, retrying in TCP mode.

; <<>> DiG 9.9.9-P1 <<>> aerserv-bc-us-east.bidswitch.net @168.63.129.16
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 51459
;; flags: qr rd ra; QUERY: 1, ANSWER: 133, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1280
;; QUESTION SECTION:
;aerserv-bc-us-east.bidswitch.net. IN   A

;; ANSWER SECTION:
aerserv-bc-us-east.bidswitch.net. 119 IN CNAME  bidcast-bcserver-gce-sc.bidswitch.net.
bidcast-bcserver-gce-sc.bidswitch.net. 119 IN CNAME bidcast-bcserver-gce-sc-multifo.bidswitch.net.
bidcast-bcserver-gce-sc-multifo.bidswitch.net. 59 IN A 35.207.51.125
bidcast-bcserver-gce-sc-multifo.bidswitch.net. 59 IN A 35.211.78.105
[...]

PR #2368, opened 2021-09-29
Update firewall rules

dhivyaganesan:
## Description

This is the fix for the bug https://github.com/Azure/WALinuxAgent/issues/1673

This is the new rule that is added is 
iptables -t security -I OUTPUT -d 168.63.129.16 -p tcp --destination-port 53 -j ACCEPT -w

which will allow DNS TCP access to a non root user as well.

Issue # 

---

### PR information


azurelinuxagent/common/utils/networkutil.py
```diff
@@ -149,6 +150,13 @@ def __get_common_accept_command_params(wait, command, destination, uid):
         cmd.extend(["owner", "--uid-owner", str(uid), "-j", "ACCEPT"])
         return AddFirewallRules._add_wait(wait, cmd)
 
+    @staticmethod
+    def __get_accept_command_params_nonroot_tcp(wait, command, destination):
+        cmd = AddFirewallRules.__get_common_command_params(command, destination)
+        cmd.pop() #poped the last parameter -m since it is not needed in this rule
```
larohra, line 156: Can you paste some doc for the `-m` parameter? Or explain a bit what it does and why its not needed here. 
Thanks!

azurelinuxagent/common/osutil/default.py
```diff
@@ -79,6 +79,10 @@ def _get_firewall_accept_command(wait, command, destination, owner_uid):
     return AddFirewallRules.get_iptables_accept_command(wait, command, destination, owner_uid)
 
 
+def _get_firewall_accept_command_nonroot_tcp(wait, command, destination):
+    return AddFirewallRules.get_iptables_accept_command_nonroot_tcp(wait, command, destination)
```
larohra, line 83: Can you post a comment on how the iptable command would look for this here please?

azurelinuxagent/common/osutil/default.py
```diff
@@ -109,6 +113,10 @@ def _get_firewall_delete_conntrack_drop_command(wait, destination):
                      ["iptables", "-t", "security", "-D", "OUTPUT", "-d", destination, "-p", "tcp", "-m", "conntrack",
                       "--ctstate", "INVALID,NEW", "-j", "DROP"])
 
+def _get_firewall_delete_accept_nonroot_tcp_command(wait, destination):
+    return _add_wait(wait,
```
larohra, line 117: Can this be parameterized similar to `AddFirewallRules.get_iptables_accept_command_nonroot_tcp(wait, command, destination)`

azurelinuxagent/common/osutil/default.py
```diff
@@ -267,11 +276,22 @@ def enable_firewall(self, dst_ip, uid):
...
             try:
                 drop_rule = _get_firewall_drop_command(wait, "-C", dst_ip)
                 shellutil.run_command(drop_rule)
+                try:
+                    accept_rule = _get_firewall_accept_command_nonroot_tcp(wait, "-C", dst_ip)
+                    shellutil.run_command(accept_rule)
+                except CommandError as e:
+                    logger.info("Iptable rule to allow tcp request to wireserver is not found. Adding it")
+                    try:
+                        accept_rule = _get_firewall_accept_command_nonroot_tcp(wait, "-I", dst_ip)
```
larohra, line 289: This part looks very clunky, could you modularize it to make it a bit cleaner? Thanks!

larohra: Also, I see you added a new parameter `__INSERT_COMMAND` in networkutil.py file, can you use that here too please (might need to make that constant public to use it in this module)

azurelinuxagent/common/osutil/default.py
```diff
@@ -267,11 +276,22 @@ def enable_firewall(self, dst_ip, uid):
 
         try:
             wait = self.get_firewall_will_wait()
-
             # If the DROP rule exists, make no changes
             try:
                 drop_rule = _get_firewall_drop_command(wait, "-C", dst_ip)
                 shellutil.run_command(drop_rule)
+                try:
+                    accept_rule = _get_firewall_accept_command_nonroot_tcp(wait, "-C", dst_ip)
```
larohra, line 284: On the same note, can you also create a new parameter for `-C` parameter and add a small comment on what it does (checks if the rule exists in the chain).

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -120,6 +120,7 @@ class AddFirewallRules(object):
     """
 
     __ACCEPT_COMMAND = "-A"
+    __INSERT_COMMAND = "-I"
```
larohra, line 123: Can you add a small comment here on the difference between `-A` and `-I`

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -175,6 +189,13 @@ def get_firewalld_accept_command(command, destination, uid, wait=""):
                                                                 uid))
         return cmd
 
+    @staticmethod
+    def get_firewalld_accept_command_nonroot_tcp(command, destination, wait=""):
```
larohra, line 193: This is not being used anywhere currently, is that part of the 2nd PR where you'll figure out where/how to add it?

tests/ga/test_update.py
```diff
@@ -1648,8 +1648,8 @@ def _mock_popen(cmd, *args, **kwargs):
                 with patch('azurelinuxagent.common.conf.enable_firewall', return_value=True):
                     update_handler.run(debug=True)
 
-        # Firewall-cmd should only be called 3 times - 1st to check if running, 2nd & 3rd for the QueryPassThrough cmd
-        self.assertEqual(3, len(executed_commands), "The number of times firwall-cmd should be called is only 3")
+        # Firewall-cmd should only be called 4 times - 1st to check if running, 2nd & 3rd & 4th for the QueryPassThrough cmd
+        self.assertEqual(3, len(executed_commands), "The number of times firwall-cmd should be called is only 4")
```
larohra, line 1652: This test is not doing anything as of now, can you please revert the comments here

tests/common/osutil/test_default.py
```diff
@@ -777,7 +781,7 @@ def test_enable_firewall_should_not_set_firewall_if_the_drop_rule_exists(self):
             success = osutil.DefaultOSUtil().enable_firewall(dst_ip=mock_iptables.destination, uid=mock_iptables.uid)
 
             self.assertTrue(success, "Enabling the firewall was not successful")
-            self.assertEqual(len(mock_iptables.command_calls), 1, "Incorrect number of calls to iptables: [{0}]". format(mock_iptables.command_calls))
+            self.assertEqual(len(mock_iptables.command_calls), 2, "Incorrect number of calls to iptables: [{0}]". format(mock_iptables.command_calls))
```
larohra, line 784: Can you please also add a testcase for the new scenario that you've introduced, where you check if the TCP rule exists or not and add just that one.

larohra (changes requested):

azurelinuxagent/common/utils/networkutil.py
narrieta: this should remove the -m parameter explicitly instead of relying on it always being the last command

azurelinuxagent/common/osutil/default.py
```diff
@@ -267,11 +276,22 @@ def enable_firewall(self, dst_ip, uid):
 
         try:
             wait = self.get_firewall_will_wait()
-
             # If the DROP rule exists, make no changes
             try:
                 drop_rule = _get_firewall_drop_command(wait, "-C", dst_ip)
                 shellutil.run_command(drop_rule)
+                try:
```
narrieta, line 283: This should not be done here... this code is just checking whether the firewall rule is already setup (by checking for the drop rule). As the comment @ line 279 points out, if the firewall is already setup, this code should make no changes.

azurelinuxagent/common/utils/networkutil.py
dhivyaganesan: You are right , firewalld changes are also included now

azurelinuxagent/common/utils/networkutil.py
dhivyaganesan: Addressed

azurelinuxagent/common/osutil/default.py
dhivyaganesan: I have written it the way other delete rules have been coded . Do you think all of them needs to be parameterized ?

azurelinuxagent/common/osutil/default.py
dhivyaganesan: Done

azurelinuxagent/common/osutil/default.py
dhivyaganesan: After having a discussion with Norberto , this logic moved to Update.py

azurelinuxagent/common/osutil/default.py
dhivyaganesan: This logic moved under Update.py and addressed your comment there.

tests/common/osutil/test_default.py
dhivyaganesan: Added few test cases under test_update.py

azurelinuxagent/common/osutil/default.py
dhivyaganesan: Moved the logic to add the new rule under Update.py

tests/ga/test_update.py
dhivyaganesan: Fixed the test according to new logic

azurelinuxagent/common/utils/networkutil.py
dhivyaganesan: According to https://unix.stackexchange.com/questions/241759/what-is-the-purpose-of-m-match-in-an-iptables-rule/241790, The -m or --match option is used to enable one or more extended packet matching modules with the given name and there is not need to specify -m tcp when -p tcp is included in the rule. 
In our case
 iptables -t security -I OUTPUT 1 -d 168.63.129.16 -m tcp -p tcp  --destination-port 53 -j ACCEPT -w and
 iptables -t security -I OUTPUT 1 -d 168.63.129.16 -p tcp --destination-port 53 -j ACCEPT -w both works the same way (verified it on a VM). I went with excluding the -m option since that was the suggestion given in the bug
https://github.com/Azure/WALinuxAgent/issues/1673

dhivyaganesan: Removed -m parameter explicitly, rather than using pop

azurelinuxagent/ga/update.py
```diff
@@ -980,6 +985,34 @@ def _ensure_firewall_rules_persisted(dst_ip):
             message=msg,
             log_event=False)
 
+    @staticmethod
+    def _add_nonroot_tcp_firewallrule_if_not_enabled(dst_ip):
+
+        update_handler = get_update_handler()
```
narrieta, line 991: This instance of UpdateHandler is not needed.

Also, with the current code it is not a good idea to create multiple instances of the UpdateHandler since there is some shared static state

azurelinuxagent/ga/update.py
```diff
@@ -980,6 +985,34 @@ def _ensure_firewall_rules_persisted(dst_ip):
...
+                    logger.info("Firewall rule to allow non root users to do tcp request to wireserver unavailable . Setting it now.")
+                    accept_non_root = _get_firewall_accept_command_nonroot_tcp("-w", "-I", dst_ip)
+                    shellutil.run_command(accept_non_root)
+                    logger.info("Succesfully added firewall rule to allow non root users to do tcp request to wireserver ")
+                except Exception as e:
+                    msg = "Unable to set the nonroot tcp access firewall rule:{0}".format(ustr(e))
+                    logger.error(msg)
+
+    @staticmethod
+    def _check_if_iptable_rule_is_available(command):
```
narrieta, line 1010: we need to rename this function: it has nothing to do with iptables, it just executes an arbitrary command

azurelinuxagent/ga/update.py
```diff
@@ -980,6 +985,34 @@ def _ensure_firewall_rules_persisted(dst_ip):
...
+        if not update_handler._check_if_iptable_rule_is_available(drop_rule):
+            return
+        else:
+            accept_non_root = _get_firewall_accept_command_nonroot_tcp("-w", "-C", dst_ip)
+            if not update_handler._check_if_iptable_rule_is_available(accept_non_root):
+                try:
+                    logger.info("Firewall rule to allow non root users to do tcp request to wireserver unavailable . Setting it now.")
+                    accept_non_root = _get_firewall_accept_command_nonroot_tcp("-w", "-I", dst_ip)
+                    shellutil.run_command(accept_non_root)
+                    logger.info("Succesfully added firewall rule to allow non root users to do tcp request to wireserver ")
```
narrieta, line 1004: "to do tcp request to wireserver" -> "to do DNS TCP requests to wireserver"

azurelinuxagent/ga/update.py
```diff
@@ -980,6 +985,34 @@ def _ensure_firewall_rules_persisted(dst_ip):
...
+
+        # "-C" checks if the iptable rule is available in the chain
+        drop_rule = _get_firewall_drop_command("-w", "-C", dst_ip)
+        if not update_handler._check_if_iptable_rule_is_available(drop_rule):
+            return
+        else:
+            accept_non_root = _get_firewall_accept_command_nonroot_tcp("-w", "-C", dst_ip)
+            if not update_handler._check_if_iptable_rule_is_available(accept_non_root):
+                try:
+                    logger.info("Firewall rule to allow non root users to do tcp request to wireserver unavailable . Setting it now.")
```
narrieta, line 1001: "to do tcp request to wireserver unavailable" -> "to do DNS TCP requests to wireserver does not exist"

azurelinuxagent/ga/update.py
larohra: Additionally, just noticed this is only used by the `_add_nonroot_tcp_firewallrule_if_not_enabled` function. Maybe consider making it an inline function inside that function since its localized?

tests/ga/test_update.py
```diff
@@ -1648,13 +1651,83 @@ def _mock_popen(cmd, *args, **kwargs):
                 with patch('azurelinuxagent.common.conf.enable_firewall', return_value=True):
                     update_handler.run(debug=True)
 
-        # Firewall-cmd should only be called 3 times - 1st to check if running, 2nd & 3rd for the QueryPassThrough cmd
-        self.assertEqual(3, len(executed_commands), "The number of times firwall-cmd should be called is only 3")
+        # Firewall-cmd should only be called 4 times - 1st to check if running, 2nd & 3rd & 4th for the QueryPassThrough cmd
```
larohra, line 1654: NIT: Rewrite as - 

> 2nd, 3rd & 4th for the QueryPassThrough cmd

azurelinuxagent/common/osutil/default.py
```diff
@@ -104,18 +105,15 @@ def _get_firewall_delete_conntrack_accept_command(wait, destination):
 
 
 def _get_firewall_delete_owner_accept_command(wait, destination, owner_uid):
-    return _add_wait(wait, ["iptables", "-t", "security", "-D", "OUTPUT", "-d", destination, "-p", "tcp", "-m", "owner",
-                            "--uid-owner", str(owner_uid), "-j", "ACCEPT"])
+    return _add_wait(wait, AddFirewallRules.get_iptables_accept_command("", "-D", destination, owner_uid))
```
larohra, line 108: The function `get_iptables_accept_command(wait, command, destination, owner_uid)` takes in `wait` as the first parameter. You can simply use this call - 

`return AddFirewallRules.get_iptables_accept_command(wait, "-D", destination, owner_uid)`

azurelinuxagent/common/osutil/default.py
```diff
@@ -104,18 +105,15 @@ def _get_firewall_delete_conntrack_accept_command(wait, destination):
...
 
 
 def _get_firewall_delete_conntrack_drop_command(wait, destination):
-    return _add_wait(wait,
-                     ["iptables", "-t", "security", "-D", "OUTPUT", "-d", destination, "-p", "tcp", "-m", "conntrack",
-                      "--ctstate", "INVALID,NEW", "-j", "DROP"])
+    return _add_wait(wait, AddFirewallRules.get_iptables_drop_command("", "-D", destination))
```
larohra, line 112: Same here

azurelinuxagent/common/osutil/default.py
larohra: Additionally, could you parameterize `-D` just like you did `-I` and `-A` please (maybe even make it an Enum potentially?)

azurelinuxagent/common/osutil/default.py
```diff
@@ -104,18 +105,15 @@ def _get_firewall_delete_conntrack_accept_command(wait, destination):
...
 
-def _get_firewall_delete_accept_nonroot_tcp_command(wait, destination):
-    return _add_wait(wait,
-                     ["iptables", "-t", "security", "-D", "OUTPUT", "-d", destination, "-p", "tcp", "--destination-port", "53", "-j", "ACCEPT"])
+
+def _get_firewall_delete_accept_dns_tcp_command(wait, destination):
+    return _add_wait(wait, AddFirewallRules.get_iptables_accept_dns_tcp_request_command("", "-D", destination))
```
larohra, line 116: Same here

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -119,12 +119,12 @@ class AddFirewallRules(object):
     Please make sure to not log anything in any function this class.
     """
 
-    #-A adds the rule to the eno of the ip table chain
+   # __ACCEPT_COMMAND adds the rule to the end of the iptable chain
```
larohra, line 122: PEP 8 violation here - 

`PEP 8: E114 indentation is not a multiple of four (comment)`

azurelinuxagent/common/osutil/default.py
dhivyaganesan: Addressed

azurelinuxagent/ga/update.py
dhivyaganesan: Addressed

azurelinuxagent/ga/update.py
dhivyaganesan: Addressed

azurelinuxagent/ga/update.py
```diff
@@ -124,6 +127,7 @@ def get_update_handler():
     return UpdateHandler()
 
 
+
```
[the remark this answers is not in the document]
dhivyaganesan: Addressed

azurelinuxagent/ga/update.py
```diff
@@ -980,6 +985,34 @@ def _ensure_firewall_rules_persisted(dst_ip):
             message=msg,
             log_event=False)
 
+    @staticmethod
+    def _add_nonroot_tcp_firewallrule_if_not_enabled(dst_ip):
```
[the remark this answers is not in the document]
dhivyaganesan: Renamed it to "_add_dns_tcp_firewall_rule_if_not_enabled"

tests/ga/test_update.py
dhivyaganesan: Addressed

azurelinuxagent/ga/update.py
```diff
@@ -980,6 +985,34 @@ def _ensure_firewall_rules_persisted(dst_ip):
...
+
+        update_handler = get_update_handler()
+
+        # "-C" checks if the iptable rule is available in the chain
+        drop_rule = _get_firewall_drop_command("-w", "-C", dst_ip)
+        if not update_handler._check_if_iptable_rule_is_available(drop_rule):
```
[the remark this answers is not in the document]
dhivyaganesan: The logic behind here is that we first check the DROP Command, if DROP doesn't exist then none of the rule is set. The code returns here as the environment thread takes care of setting up all three rules ( success = self._osutil.enable_firewall(dst_ip=self._protocol.get_endpoint(), uid=os.getuid())

If the drop rule exists then we check if the new rule exists , if not we add it here. 

If we don't check the DROP command first , for a new VM just the new rule will get added before the other rules are in place which is not the expected behavior.

azurelinuxagent/common/osutil/default.py
dhivyaganesan: Addressed

azurelinuxagent/common/osutil/default.py
dhivyaganesan: Addressed

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -119,8 +119,14 @@ class AddFirewallRules(object):
     Please make sure to not log anything in any function this class.
     """
 
+    #-A adds the rule to the eno of the ip table chain
```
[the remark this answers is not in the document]
dhivyaganesan: Addressed

dhivyaganesan:
Latest DCR run for the all the changes: https://tuxgold.corp.microsoft.com/job/AzLinux/job/DungeonCrawler/job/AgentBVTs.dhivya/64/

azurelinuxagent/ga/update.py
```diff
@@ -982,6 +986,34 @@ def _ensure_firewall_rules_persisted(dst_ip):
...
+        # Helper to execute a run command, returns True if no exception else returns False
+        def _execute_run_command(command):
+            try:
+                shellutil.run_command(command)
+                return True
+            except Exception:
+                return False
+
+        # "-C" checks if the iptable rule is available in the chain. It throws an exception if the ip table rule doesnt exist
+        drop_rule = _get_firewall_drop_command("-w", "-C", dst_ip)
```
larohra, line 1001: There's an existing function to check weather to add wait or not based on the iptable version (check `wait = self.get_firewall_will_wait()` function). Please use that instead of adding `-w` directly

larohra: Additionally, can you add a parameter of `-C` too (similar to `-A-`, `-D`, `-I`) and use that instead of using it directly

azurelinuxagent/ga/update.py
```diff
@@ -982,6 +986,34 @@ def _ensure_firewall_rules_persisted(dst_ip):
...
+        if not _execute_run_command(drop_rule):
+            return
+        else:
+            accept_non_root = _get_firewall_accept_dns_tcp_request_command("-w", "-C", dst_ip)
+            if not _execute_run_command(accept_non_root):
+                try:
+                    logger.info("Firewall rule to allow DNS TCP request to wireserver for a non root user unavailable . Setting it now.")
+                    accept_non_root = _get_firewall_accept_dns_tcp_request_command("-w", "-I", dst_ip)
```
larohra, line 1009: NIT: Use the existing variable of `-I` instead of using it directly

azurelinuxagent/ga/update.py
```diff
@@ -982,6 +986,34 @@ def _ensure_firewall_rules_persisted(dst_ip):
...
+                    shellutil.run_command(accept_non_root)
+                    logger.info("Succesfully added firewall rule to allow non root users to do a DNS TCP request to wireserver ")
+                except Exception as e:
+                    msg = "Unable to set the nonroot tcp access firewall rule:{0}".format(ustr(e))
+                    logger.error(msg)
+
+
```
larohra, line 1016: NIT: Extra line here. Throwing a PEP violation warning - `PEP 8: E303 too many blank lines (3)`

azurelinuxagent/ga/update.py
```diff
@@ -982,6 +986,34 @@ def _ensure_firewall_rules_persisted(dst_ip):
...
+            except Exception:
+                return False
+
+        # "-C" checks if the iptable rule is available in the chain. It throws an exception if the ip table rule doesnt exist
+        drop_rule = _get_firewall_drop_command("-w", "-C", dst_ip)
+        if not _execute_run_command(drop_rule):
+            return
+        else:
+            accept_non_root = _get_firewall_accept_dns_tcp_request_command("-w", "-C", dst_ip)
+            if not _execute_run_command(accept_non_root):
```
larohra, line 1006: Can you add a log statement stating `Not adding DNS firewall rule since it already exists` or something. Thanks!

azurelinuxagent/common/osutil/default.py
larohra: > Additionally, could you parameterize `-D` just like you did `-I` and `-A` please (maybe even make it an Enum potentially?)

NIT on this comment

azurelinuxagent/ga/update.py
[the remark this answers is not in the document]
dhivyaganesan: Addressed in the next commit

azurelinuxagent/common/osutil/default.py
dhivyaganesan: I parameterized all the commands , am I missing any NIT comment ?

azurelinuxagent/ga/update.py
dhivyaganesan: Addressed in the next commit

azurelinuxagent/ga/update.py
dhivyaganesan: Addressed in the next commit

azurelinuxagent/ga/update.py
```diff
@@ -982,6 +986,34 @@ def _ensure_firewall_rules_persisted(dst_ip):
...
+            try:
+                shellutil.run_command(command)
+                return True
+            except Exception:
+                return False
+
+        # "-C" checks if the iptable rule is available in the chain. It throws an exception if the ip table rule doesnt exist
+        drop_rule = _get_firewall_drop_command("-w", "-C", dst_ip)
+        if not _execute_run_command(drop_rule):
+            return
```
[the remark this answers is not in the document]
dhivyaganesan: Addressed in the next commit

PR #2370, opened 2021-09-30
Report status even if goal state cannot be processed

kevinclark19a:
## Description

The idea behind this PR is to report status via a cached goal state if a new one fails to be retrieved.

There are three cases to watch out for here:

* The very first goal state fails, such that there is no cached goal state (in `wire_client._goal_state`)
* Any goal state after the first fails, such that there is a cached goal state
* A goal state doesn't fail, aka the happy path.

The first isn't an issue, as it's an exit condition (i.e. it happens outside of any interior `try: except` and ends the process). The second and third should be handled here.

---

### PR information


azurelinuxagent/ga/update.py
```diff
@@ -423,37 +423,40 @@ def _process_goal_state(self, exthandlers_handler, remote_access_handler):
...
-            if incarnation != self.last_incarnation:  # TODO: This check should be based in the etag for the extensions goal state
-                if not self._extensions_summary.converged:
-                    message = "A new goal state was received, but not all the extensions in the previous goal state have completed: {0}".format(self._extensions_summary)
-                    logger.warn(message)
-                    add_event(op=WALAEventOperation.GoalState, message=message, is_success=False, log_event=False)
-                    if self._is_initial_goal_state:
-                        self._on_initial_goal_state_completed(self._extensions_summary)
-                self._extensions_summary = ExtensionsSummary()
-                exthandlers_handler.run()
+        else:
```
larohra, line 430: NIT: This else is redundant, the previous if condition will return if we ever go in that block

tests/ga/test_update.py
```diff
@@ -2110,6 +2110,60 @@ class ReportStatusTestCase(AgentTestCase):
...
+        return AgentTestCase.setUp(self)
+
+    def tearDown(self):
+        
+        for p in self.patches:
+            p.stop()
+
+        return AgentTestCase.tearDown(self)
+
+    def test_upload_status_blob_even_on_failed_goal_state(self):
```
larohra, line 2132: NIT: Maybe rename to `test_upload_status_blob_even_on_failed_goal_state_fetch` to capture what exactly is the failing case

tests/ga/test_update.py
```diff
@@ -2110,6 +2110,60 @@ class ReportStatusTestCase(AgentTestCase):
...
+
+        @contextlib.contextmanager
+        def mock_update_handler():
+            def goal_state_handler(url, **kwargs): # pylint: disable=unused-argument
```
larohra, line 2136: NIT: remove pylint supression

tests/ga/test_update.py
```diff
@@ -2110,6 +2110,60 @@ class ReportStatusTestCase(AgentTestCase):
...
+                    update_handler._report_status = Mock()
+
+                    is_running_patch.start()
+                    yield update_handler
+            finally:
+                is_running_patch.stop()
+
+        with mock_update_handler() as update_handler:
+            update_handler.run(debug=True)
+            update_handler._report_status.assert_called_once()
```
larohra, line 2164: Thanks for adding this test case. Can you also add another test case where you verify that the status being reported is indeed the last cached one.

tests/ga/test_update.py
kevinclark19a: I believe this function's signature has to coincide with `http_request` which takes in info not needed by this function. I'll add a comment mentioning why this supression is necessary.

tests/ga/test_update.py
kevinclark19a: I'm pushing a commit that checks the contents of the status blob; my idea is to validate that the extension mentioned in the extension config is also mentioned in the status blob. Let me know if you don't think this is adequate-- not quite sure of a better solution than this.

tests/ga/test_update.py
```diff
@@ -2110,6 +2110,141 @@ class ReportStatusTestCase(AgentTestCase):
...
+                original_report_status = update_handler._report_status
+                update_handler._report_status = Mock(side_effect=original_report_status)
+
+                is_running_patch.start()
+                yield update_handler
+        finally:
+            is_running_patch.stop()
+
+    @staticmethod
+    def _goal_state_handler(url, **kwargs): # pylint: disable=unused-argument
```
larohra, line 2176: if `kwargs` is not being used, maybe change the name to `**_` or something?

tests/ga/test_update.py
```diff
@@ -2110,6 +2110,141 @@ class ReportStatusTestCase(AgentTestCase):
...
+        and thus needs **kwargs. We don't need the information, though, so we tell pylint 
+        we aren't planning on using the argument.
+
+        Relies on this function to have the property return_vals, so it'll need to be
+        set before being called
+        """
+        if not HttpRequestPredicates.is_goal_state_request(url):
+            return None
+        try:
+            return ReportStatusTestCase._goal_state_handler.return_vals.pop()
```
larohra, line 2188: From the new tests that are using this function, all of them are pretty much returning None. Why not keep it simple and remove the `return_vals` property and just return `None` in all cases?

tests/ga/test_update.py
larohra: Additionally, can you rename the function to something more specific? Like say `_fail_get_goal_state`?

tests/ga/test_update.py
```diff
@@ -2110,6 +2110,141 @@ class ReportStatusTestCase(AgentTestCase):
...
+        """
+        Note: this function's signature has to coincide with that of http_get_request,
+        and thus needs **kwargs. We don't need the information, though, so we tell pylint 
+        we aren't planning on using the argument.
+
+        Relies on this function to have the property return_vals, so it'll need to be
+        set before being called
+        """
+        if not HttpRequestPredicates.is_goal_state_request(url):
+            return None
```
larohra, line 2186: If its not a GoalState request, it should return success right?

tests/ga/test_update.py
```diff
@@ -2110,6 +2110,141 @@ class ReportStatusTestCase(AgentTestCase):
...
+                    actual_val = ext_handler_statuses[0].get(key)
+
+                    msg = "Extension information '{0}' did not match the provided extension.".format(key)
+                    self.assertEqual(actual_val, expected_val, msg)
+
+        finally:
+            # clean up the static variable
+            del ReportStatusTestCase._goal_state_handler.return_vals
+
+
```
larohra, line 2247: NIT: Extra line

tests/protocol/mockwiredata.py
```diff
@@ -286,15 +291,17 @@ def mock_http_post(self, url, *args, **kwargs):  # pylint: disable=unused-argume
         resp.read = Mock(return_value=content.encode("utf-8"))
         return resp
 
-    def mock_http_put(self, url, *args, **kwargs):  # pylint: disable=unused-argument
-        content = None
+    def mock_http_put(self, url, data, **kwargs):  # pylint: disable=unused-argument
```
larohra, line 294: NIT: pylint error.
If the variable is not being used, please rename it to `**_` or something

larohra (changes requested):

tests/ga/test_update.py
kevinclark19a: Yes, I can make it more explicit, but that is what this line is already doing. The `mock_wire_protocol` context manager basically queries the appropriate provided handler (i.e., this function is called on GET requests) and calls the default handler if the provided handler returns `None`.

Essentially, the way to get the mock protocol to return the file data is to return `None`.

tests/ga/test_update.py
```diff
@@ -2110,6 +2110,141 @@ class ReportStatusTestCase(AgentTestCase):
...
+        mock_wire_data_file = kwargs.get("mock_wire_data_file", mockwiredata.DATA_FILE_STATUS_BLOB)
+
+        try:
+            with mock_wire_protocol(mock_wire_data_file, mock_http_get, mock_http_post, mock_http_put) as protocol:
+                update_handler = get_update_handler()
+                
+                update_handler.protocol_util.get_protocol = Mock(return_value=protocol)
+
+                original_report_status = update_handler._report_status
+                update_handler._report_status = Mock(side_effect=original_report_status)
```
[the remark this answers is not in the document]
kevinclark19a: I think you're right that this is unneeded. We shouldn't be relying on private methods in the first place, I think.

I'll push a change shortly.

tests/ga/test_update.py
kevinclark19a: I think the main issue here is that we need to save state between function calls, so the static property would I think be needed regardless. I.e. replacing this line with `return None` would just require us to use a static property `count` or something that we decrement each time.

That's obviously do-able, but I'm not sure it really gains much. I could be missing something simple, though, did you have a specific idea on how to do this?

tests/ga/test_update.py
larohra: > Essentially, the way to get the mock protocol to return the file data is to return None.

Are you sure about this? From what I remember returning a `None` is an error. I mean that is exactly what you're doing too right when you set this mock - 

`ReportStatusTestCase._goal_state_handler.return_vals = [ None, None ]`

tests/ga/test_update.py
larohra: I dont get it sorry. From the code usage, you're only returning `None` as the return_vals, I didnt see any other return value at all anywhere. Am I missing something here?

tests/ga/test_update.py
kevinclark19a: Yes, the code in question related to this is `tests/protocol/mocks.py#85-103`:

```python
if handler is not None:
    if method == 'GET':
        return_value = handler(url, **kwargs)
    else:
        return_value = handler(url, data, **kwargs)
    if return_value is not None:
        tracked_urls.append(url)
        if isinstance(return_value, Exception):
            raise return_value
        return return_value
        
# if the request was not handled try to use the mock wireserver data
if _USE_MOCK_WIRE_DATA_RE.match(url) is not None:
    if method == 'GET':
        return protocol.mock_wire_data.mock_http_get(url, **kwargs)
    if method == 'POST':
        return protocol.mock_wire_data.mock_http_post(url, data, **kwargs)
    if method == 'PUT':
        return protocol.mock_wire_data.mock_http_put(url, data, **kwargs)
```

And this is documented in `mock_wire_protocol`'s docstring (last line in the quote is most important here):

```
The caller can also provide handler functions for specific HTTP methods using the http_*_handler arguments. The return value of the handler
function is interpreted similarly to the "return_value" argument of patch(): if it is an exception the exception is raised or, if it is
any object other than None, the value is returned by the mock. If the handler function returns None the call is handled using the mock
wireserver data or passed to the original to restutil.http_request.
```

What `ReportStatusTestCase._goal_state_handler.return_vals = [ None, None ]` is doing is determining how many times a goal state request succeeds before failing with a `HttpError` (via raisining an `indexError` when `...return_vals.pop()` is called on an empty array). I can add to the comment right above that line, mentioning that returning `None` has a special meaning to `mock_wire_protocol` that we inherit from `patch()`.

tests/ga/test_update.py
kevinclark19a: Yes, I think this function is not clear enough and I'll need to either do some refactoring or add more comments. The "`other return value`" is actually `raise HttpError()`.

The idea behind this method is to return a list of goal states as requested, in order, and then fail with an `HttpError`, mimicking the scenario this PR is supposed to fix. In both of the tests below, I've written it to just return the same goal state either two or three times before failing with `HttpError`, but as written, this concievably supports a different goal state each time.

tests/ga/test_update.py
```diff
@@ -2164,19 +2163,16 @@ def _mock_update_handler(iterations=1, **kwargs):
...
-                update_handler._report_status = Mock(side_effect=original_report_status)
-
                 is_running_patch.start()
                 yield update_handler
         finally:
             is_running_patch.stop()
 
     @staticmethod
-    def _goal_state_handler(url, **kwargs): # pylint: disable=unused-argument
+    def _fail_goal_state_fetch(url, **_): # pylint: disable=unused-argument
```
larohra, line 2172: you dont need the pylint warning suppression if using `**_` variable naming convention, please remove it here

tests/ga/test_update.py
larohra: Ahh I see now, it makes sense thank you. A comment would be super helpful here, I did not get this behavior at all at first glance, I thought that you were implementing some sort of a patch behavior where each `None` represented the value that will be returned when the code tries to do a GET.

tests/ga/test_update.py
```diff
@@ -2195,28 +2191,28 @@ def test_upload_vm_status_even_on_failed_goal_state_fetch(self):
...
                 mock_protocol = update_handler.protocol_util.get_protocol()
                 self.assertEqual(mock_protocol.mock_wire_data.call_counts['/StatusBlob'], 1,
                     "Expected a single status blob to be uploaded")
         finally:
             # clean up the static variable
-            del ReportStatusTestCase._goal_state_handler.return_vals
+            del ReportStatusTestCase._fail_goal_state_fetch.return_vals
     
-    def test_update_status_for_cached_goal_state_on_failed_fetch(self):
+    def test_upload_status_for_cached_goal_state_on_failed_fetch(self):
```
larohra, line 2206: I missed this earlier, but we generally follow this naming convention for our tests - `test_it_should_*`. Please rename this to the same pattern

tests/ga/test_update.py
```diff
@@ -2110,6 +2110,136 @@ class ReportStatusTestCase(AgentTestCase):
...
+        set before being called
+        """
+        if not HttpRequestPredicates.is_goal_state_request(url):
+            return None
+        try:
+            return ReportStatusTestCase._fail_goal_state_fetch.return_vals.pop()
+        except IndexError:
+            raise HttpError()
+
+    def test_upload_vm_status_even_on_failed_goal_state_fetch(self):
```
larohra, line 2188: NIT: Rename function name

tests/protocol/mockwiredata.py
```diff
@@ -187,7 +192,7 @@ def reload(self):
         if in_vm_artifacts_profile_file is not None:
             self.in_vm_artifacts_profile = load_data(in_vm_artifacts_profile_file)
 
-    def mock_http_get(self, url, *args, **kwargs):  # pylint: disable=unused-argument
+    def mock_http_get(self, url, *_, **kwargs):  # pylint: disable=unused-argument
```
larohra, line 195: NIT: remove unnecessary suppression

tests/protocol/mockwiredata.py
```diff
@@ -271,7 +276,7 @@ def mock_http_get(self, url, *args, **kwargs):  # pylint: disable=unused-argumen
         resp.getheaders = Mock(return_value=response_headers)
         return resp
 
-    def mock_http_post(self, url, *args, **kwargs):  # pylint: disable=unused-argument
+    def mock_http_post(self, url, *_, **__):  # pylint: disable=unused-argument
```
larohra, line 279: NIT: remove unnecessary suppression

tests/protocol/mockwiredata.py
```diff
@@ -286,15 +291,17 @@ def mock_http_post(self, url, *args, **kwargs):  # pylint: disable=unused-argume
         resp.read = Mock(return_value=content.encode("utf-8"))
         return resp
 
-    def mock_http_put(self, url, *args, **kwargs):  # pylint: disable=unused-argument
-        content = None
+    def mock_http_put(self, url, data, **_):  # pylint: disable=unused-argument
```
larohra, line 294: NIT: remove unnecessary suppression

PR #2382, opened 2021-10-18
Parse statusUploadBlob in vmSettings

narrieta:


kevinclark19a (dismissed):

azurelinuxagent/common/protocol/extensions_goal_state.py
```diff
@@ -91,8 +101,20 @@ def report_difference(attribute):
...
+            report_difference("_status_upload_blob_type")
+        if from_extensions_config.get_required_features() != from_vm_settings.get_required_features():
+            report_difference("_required_features")
+
+    def _do_common_validations(self):
+        """
+        Does validations common to vmSettings and ExtensionsConfig
+        """
+        if self._status_upload_blob_type not in ["BlockBlob", "PageBlob"]:
+            logger.info("Status Blob type '{0}' is ot valid, assuming BlockBlob", self._status_upload_blob)
```
dhivyaganesan, line 116: Small typo , 'not'

dhivyaganesan (dismissed):

azurelinuxagent/common/protocol/extensions_goal_state.py
```diff
@@ -91,8 +101,20 @@ def report_difference(attribute):
...
+        if from_extensions_config.get_status_upload_blob_type() != from_vm_settings.get_status_upload_blob_type():
+            report_difference("_status_upload_blob_type")
+        if from_extensions_config.get_required_features() != from_vm_settings.get_required_features():
+            report_difference("_required_features")
+
+    def _do_common_validations(self):
+        """
+        Does validations common to vmSettings and ExtensionsConfig
+        """
+        if self._status_upload_blob_type not in ["BlockBlob", "PageBlob"]:
```
larohra, line 115: NIT: Maybe make these values an Enum?

azurelinuxagent/common/protocol/extensions_goal_state.py
```diff
@@ -105,7 +127,14 @@ def __init__(self, incarnation, xml_text):
         super(_ExtensionsGoalStateFromExtensionsConfig, self).__init__()
         self._id = incarnation
         self._text = xml_text
+        self._parse_extensions_config(xml_text)
 
+        try:
+            self._do_common_validations()
+        except Exception as e:
+            raise ExtensionsConfigError("Error parsing ExtensionsConfig (incarnation: {0}): {1}\n{2}".format(incarnation, format_exception(e), xml_text))
```
larohra, line 135: Should we trim the xml text? Incase its very big? Is there a chance `xml_text` can contain secrets? We should redact it in that case too

azurelinuxagent/common/protocol/extensions_goal_state.py
```diff
@@ -128,15 +157,17 @@ def __init__(self, incarnation, xml_text):
...
 
         status_upload_node = find(xml_doc, "StatusUploadBlob")
-        self.status_upload_blob_type = getattrib(status_upload_node, "statusBlobType")
-        logger.verbose("Extension config shows status blob type as [{0}]", self.status_upload_blob_type)
+        self._status_upload_blob_type = getattrib(status_upload_node, "statusBlobType")
+        logger.verbose("Extension config shows status blob type as [{0}]", self._status_upload_blob_type)
 
         self.in_vm_gs_metadata.parse_node(find(xml_doc, "InVMGoalStateMetaData"))
 
+        self._do_common_validations()
```
larohra, line 169: Any reason for validating in 2 different places? (Line 133 and 169)?

azurelinuxagent/common/protocol/extensions_goal_state.py
```diff
@@ -507,22 +538,106 @@ def __init__(self, etag, json_text):
...
+        vm_settings = _CaseFoldedDict.from_dict(json.loads(json_text))
+        self._parse_status_upload_blob(vm_settings)
         self._parse_required_features(vm_settings)
         # TODO: Parse all atttributes
 
+    def _parse_status_upload_blob(self, vm_settings):
+        status_upload_blob = vm_settings.get("statusUploadBlob")
+        if status_upload_blob is None:
+            raise Exception("Missing statusUploadBlob")
+        self._status_upload_blob = status_upload_blob["value"]
```
larohra, line 555: We should also check if `value` is present in the `status_upload_blob` and raise appropriate `Exception` if not (else a `KeyError` might be raised)

azurelinuxagent/common/protocol/extensions_goal_state.py
```diff
@@ -507,22 +538,106 @@ def __init__(self, etag, json_text):
...
+        if self._status_upload_blob is None:
+            raise Exception("Missing statusUploadBlob.value")
+        self._status_upload_blob_type = status_upload_blob["statusBlobType"]
```
larohra, line 558: same comment here

azurelinuxagent/common/protocol/extensions_goal_state.py
```diff
@@ -507,22 +538,106 @@ def __init__(self, etag, json_text):
...
+            self._required_features.extend([feature["name"] for feature in required_features])
 
     def get_redacted_text(self):
         return re.sub(r'("protectedSettings"\s*:\s*)"[^"]+"', r'\1"*** REDACTED ***"', self._text)
+
+
+#
+# TODO: The current implementation of the vmSettings API uses inconsistent cases on the names of the json items it returns.
+#       To work around that, we use _CaseFoldedDict to query those json items in a case-insensitive matter, Do not use
+#       _CaseFoldedDict for other purposes. Remove it once the vmSettings API is updated.
```
larohra, line 576: We might never be able to remove this given the long tail nature of the HostGA deployments and also the fact that some hosts are never updated/not updated frequently. Just a thought to keep in mind when considering removing this.

larohra (commented):
Some minor comments else LGTM

azurelinuxagent/common/protocol/extensions_goal_state.py
narrieta: I want to be really, really sure the text is valid...

No, seriously, the second call is a leftover, will remove it. Thanks

azurelinuxagent/common/protocol/extensions_goal_state.py
narrieta: A future version of the API will include a version number and we'll send it over telemetry. That'll help us drive the removal/update of many of those TODO comments I've been adding.

PR #2403, opened 2021-11-01
Control agent auto-upgrade

larohra:
## Description
The agent currently auto-upgrades within 1 hour of a new agent release. This has the potential of rolling out a bad build region-wide instantly. In order to mitigate this, we're now going to space out the releases (4hrs for hotfix release and 24hours for normal releases)

Issue # 

---

### PR information


azurelinuxagent/ga/update.py
```diff
@@ -986,6 +1009,55 @@ def _ensure_firewall_rules_persisted(dst_ip):
...
+        if next_hotfix_time > now and next_normal_time > now:
+            return
+
+        # Update the last upgrade check time even if no new agent is available for upgrade
+        self._last_hotfix_upgrade_time = now if next_hotfix_time <= now else self._last_hotfix_upgrade_time
+        self._last_normal_upgrade_time = now if next_normal_time <= now else self._last_normal_upgrade_time
+
+        available_agent = self.get_latest_agent()
+        if available_agent is None:
+            logger.verbose("No agent upgrade discovered")
```
dhivyaganesan, line 1035: Guess it is beneficial to include current agent version here in the logs

larohra: We already have that data in the OSInfo log that we emit at the beginning of service, I particularly don't see any advantage of adding it here too. Plus this is a verbose log, it wont be visible in most of the cases to begin with.

nagworld9:
I'm just curious how actual agent upgrade happening after raise ExitException(exit agent process)

nagworld9 (dismissed):

larohra:
> I'm just curious how actual agent upgrade happening after raise ExitException(exit agent process)

We raise the `ExitException` from the `ext-handler` process after downloading the latest updated version. When that process is dead, the `daemon` process tries to restart the `ext-handler` from the highest version available on filesystem. You can check the `run_latest` function in `update.py` file for the exact code on how this works.

azurelinuxagent/ga/update.py
```diff
@@ -373,6 +385,9 @@ def run(self, debug=False):
                 time.sleep(self._goal_state_period)
 
         except ExitException as exitException:
+            if exitException.add_event:
+                add_event(AGENT_NAME, op=WALAEventOperation.AgentUpgrade, version=CURRENT_VERSION, is_success=True,
```
kevinclark19a, line 389: it isn't clear to me what guarantees that an exception with the `add_event` flag set would come from an `AgentUpgrade` operation, besides just that each and every `raise ExitException` has `add_event` set if and only if it comes from an `AgentUpgrade` operation.

Could we either: 
* make this code more general, i.e. capture _all_ of the event information in the exception object
* make the name of the flag more specific, i.e. change it from `add_event` to something like `caused_by_upgrade` or something

Just seems like there's currently a mismatch in specificity here.

larohra: Sorry for the confusion, the expectation is not that the ExitException has to come from an AgentUpgrade scenario to send an event, its just that we should not send an event for a very specific case (this one `Agent {0} is an orphan -- exiting`) and add an event for all the other cases :)
Hope that clears it up!

azurelinuxagent/common/exception.py
```diff
@@ -26,9 +26,10 @@ class ExitException(BaseException):
     """
     Used to exit the agent's process
     """
-    def __init__(self, reason):
+    def __init__(self, reason, add_event=True):
         super(ExitException, self).__init__()
         self.reason = reason
+        self.add_event = add_event
```
narrieta, line 32: it's odd to add this flag here... couldn't you use the reason instead, or derive a new exception?

azurelinuxagent/ga/update.py
```diff
@@ -132,7 +140,11 @@ def __init__(self):
         self.protocol_util = get_protocol_util()
 
         self._is_running = True
+
+        # Member variables to keep track of the Agent AutoUpgrade
         self.last_attempt_time = None
+        self._last_hotfix_upgrade_time = None
```
narrieta, line 146: could we initialize these 3 members to something other than None? maybe a time in the past, if that make sense

azurelinuxagent/ga/update.py
```diff
@@ -430,13 +445,16 @@ def _process_goal_state(self, exthandlers_handler, remote_access_handler):
             self._report_status(exthandlers_handler, incarnation_changed=False)
             return
 
-        if self._upgrade_available(protocol):
+        if self._check_and_download_agent_if_upgrade_available(protocol):
             available_agent = self.get_latest_agent()
+            # Legacy behavior: The current agent can become unavailable and needs to be reverted.
```
narrieta, line 450: not quite "the current agent".  This would exit if none of the versions downloaded are available, right?

azurelinuxagent/ga/update.py
```diff
@@ -986,6 +1009,55 @@ def _ensure_firewall_rules_persisted(dst_ip):
             message=msg,
             log_event=False)
 
+    def __upgrade_agent_if_permitted(self):
```
narrieta, line 1012: "permitted" is not the right term... that makes me think that this is related to AutoUpdate being enabled or not, when it is actually related to the period of the update

narrieta: I think keeping __upgrade_agent_if_permitted and _check_and_download_agent_if_upgrade_available as separate operations makes the logic harder to understand. Wouldn't it be simpler to do the download and upgrade within the same operation?

Currently we have that issue where a bad error.json raises an exception. by calling get_latest_agent() within __upgrade_agent_if_permitted() I think that condition would prevent the update from happening.

azurelinuxagent/common/exception.py
larohra: There was just one instance where it didnt make sense to add an event hence just added this flag. I can rewrite it as a new Exception or something no issues

azurelinuxagent/ga/update.py
larohra: The legacy code had a None check so I just kept it as is. I can modify that part, shouldn't be a big change

azurelinuxagent/ga/update.py
larohra: The function `self.get_latest_agent()` will work as you've described above, i.e. it would return None if all of the downloaded versions are blacklisted. 

But, we would not get into this code path unless the `_check_and_download_agent_if_upgrade_available` returns True, and it will only do so if either a new higher version is actually available (and not blacklisted) or the current version becomes in-eligible (blacklisted) - 

```python
# Remove all blacklisted agents here, so we dont consider blacklisted agents for our computation
self._filter_blacklisted_agents()

# Return True if current agent is no longer available or an
            # agent with a higher version number is available
            return not self._is_version_eligible(base_version) \
                   or (len(self.agents) > 0 and self.agents[0].version > base_version)
```

So this code path will only exit when the current agent indeed does become unavailable. 
Hope that clears it up, the code in this area is a bit convoluted. Hopefully when I get to removing the blacklisting logic, we can make it a bit clearer and easier to understand.

azurelinuxagent/ga/update.py
larohra: > "permitted" is not the right term... that makes me think that this is related to AutoUpdate being enabled or not, when it is actually related to the period of the update

Fair, I'll rename the function to `__upgrade_agent_if_time_elapsed`

> I think keeping __upgrade_agent_if_permitted and _check_and_download_agent_if_upgrade_available as separate operations makes the logic harder to understand. Wouldn't it be simpler to do the download and upgrade within the same operation?

I purposefully kept it separate, like I dont want the upgrading to be dependent on the download because that would add atleast an overhead of 1 hour (i.e. we'll start our counter after we download which is every one hour). The idea behind keeping it separate is to kind of trickle the upgrades as and when they come because they would be dependent on the service start time. I can explain more over a call if you would prefer that.

> Currently we have that issue where a bad error.json raises an exception. by calling get_latest_agent() within __upgrade_agent_if_permitted() I think that condition would prevent the update from happening.

Good point, though I feel that bug will block the updates even now. I think we can handle that bug separately as part of the task to phase out blacklisting logic.

azurelinuxagent/common/exception.py
larohra: Done

narrieta: thanks

azurelinuxagent/ga/update.py
narrieta: Yes, I noticed the legacy code initializes to None, and that is not a good practice in general. I'm ok leaving it as is, but let's try not to follow those patterns in new code (unless the cost is excessive)

azurelinuxagent/ga/update.py
narrieta: Sorry, I meant to say the new comment is a little misleading. If there is no latest agent the code reverts to the daemon installed in the image (I think there is a log statement pointing this out)

azurelinuxagent/ga/update.py
narrieta: It's OK as is, I think the implementation can be further simplified but you have a fair point that this is just temporary code while the Agent versioning get implemented, so no issue here.

narrieta (approved):

azurelinuxagent/ga/update.py
larohra: Makes sense, I'll update this if I touch this code for GAVersioning

azurelinuxagent/ga/update.py
larohra: Ahh I see, that's a good point. I'll update it if I have to push another update, else will handle it in the upcoming PRs

nagworld9 (approved):

PR #2413, opened 2021-11-16
Add check for HostGAPlugin version

narrieta:
Some of the older versions of the HostGAPlugin do not include the status blob; added check to skip it when comparing against ExtensionsConfig.

Also added telemetry for the HostGAPlugin  version.

### DIFF
--- a/azurelinuxagent/common/protocol/extensions_goal_state.py
+++ b/azurelinuxagent/common/protocol/extensions_goal_state.py
@@ -21,6 +21,7 @@ import azurelinuxagent.common.logger as logger
 from azurelinuxagent.common.AgentGlobals import AgentGlobals
 from azurelinuxagent.common.exception import AgentError
 from azurelinuxagent.common.utils import textutil
+from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
 from azurelinuxagent.common.protocol.restapi import ExtHandlerList, VMAgentManifestList
 
 
@@ -108,8 +109,10 @@ class ExtensionsGoalState(object):
         compare_attribute("activity_id")
         compare_attribute("correlation_id")
         compare_attribute("created_on_timestamp")
-        compare_attribute("status_upload_blob")
-        compare_attribute("status_upload_blob_type")
+        # The status blob was added after version 112
+        if from_vm_settings.host_ga_plugin_version > FlexibleVersion("1.0.8.112"):
+            compare_attribute("status_upload_blob")
+            compare_attribute("status_upload_blob_type")
         compare_attribute("required_features")
         compare_attribute("on_hold")
 
--- a/azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
+++ b/azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
@@ -21,6 +21,7 @@ import sys
 
 from azurelinuxagent.common.exception import VmSettingsError
 from azurelinuxagent.common.protocol.extensions_goal_state import ExtensionsGoalState
+from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
 from azurelinuxagent.common.utils.textutil import format_exception
 
 
@@ -28,9 +29,9 @@ class ExtensionsGoalStateFromVmSettings(ExtensionsGoalState):
     def __init__(self, etag, json_text):
         super(ExtensionsGoalStateFromVmSettings, self).__init__()
         self._id = etag
-        self._host_ga_plugin_version = None
-        self._schema_version = None
         self._text = json_text
+        self._host_ga_plugin_version = FlexibleVersion()
+        self._schema_version = FlexibleVersion()
         self._activity_id = None
         self._correlation_id = None
         self._created_on_timestamp = None
@@ -48,6 +49,14 @@ class ExtensionsGoalStateFromVmSettings(ExtensionsGoalState):
     @property
     def id(self):
         return self._id
+
+    @property
+    def host_ga_plugin_version(self):
+        return self._host_ga_plugin_version
+
+    @property
+    def schema_version(self):
+        return self._schema_version
 
     @property
     def activity_id(self):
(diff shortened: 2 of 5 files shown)


azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
```diff
@@ -21,16 +21,17 @@
...
 
 class ExtensionsGoalStateFromVmSettings(ExtensionsGoalState):
     def __init__(self, etag, json_text):
         super(ExtensionsGoalStateFromVmSettings, self).__init__()
         self._id = etag
-        self._host_ga_plugin_version = None
-        self._schema_version = None
         self._text = json_text
+        self._host_ga_plugin_version = FlexibleVersion()
+        self._schema_version = FlexibleVersion()
```
dhivyaganesan, line 34: Can you please help me understand what FlexibleVersion() without any parameter passed returns and what will be the default value of _host_ga_plugin_version?
Had a quick read about FlexibleVersion(version.Version) and looks like it expects a version.

dhivyaganesan (dismissed):
Appreciate if you could help me understand FlexibleVersion() , otherwise looks good to me.

azurelinuxagent/common/protocol/extensions_goal_state.py
```diff
@@ -108,8 +109,10 @@ def compare_attribute(attribute):
         compare_attribute("activity_id")
         compare_attribute("correlation_id")
         compare_attribute("created_on_timestamp")
-        compare_attribute("status_upload_blob")
-        compare_attribute("status_upload_blob_type")
+        # The status blob was added after version 112
```
larohra, line 112: I'm guessing this check is temporary until this feature is stabilized? If so, could you please add a comment about it too. Thanks!

larohra: On that note, do the latest version of HostGA contain all the attributes that we're parsing? Or will we have to have a similar check like this for other versions too in the future?

larohra: 1 more question, what's the min version where the HostGA version is available? Is it .112 or greater than that?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1332,9 +1332,17 @@ def get_header_for_cert(self):
...
             goal_state = GoalState(self)
-            self._set_host_plugin(HostPluginProtocol(self.get_endpoint(),
-                                                     goal_state.container_id,
-                                                     goal_state.role_config_name))
+            self._set_host_plugin(HostPluginProtocol(self.get_endpoint(), goal_state.container_id, goal_state.role_config_name))
+            try:
+                etag, vm_settings = self._fetch_vm_settings(None)
+                extensions_goal_state = ExtensionsGoalStateFactory.create_from_vm_settings(etag, vm_settings)
+                message = "HostGAPlugin version: {0}".format(extensions_goal_state.host_ga_plugin_version)
+                logger.warn(message)
```
larohra, line 1340: NIT: This should be an info

tests/protocol/test_extensions_goal_state.py
```diff
@@ -28,6 +29,32 @@ def test_property(name, value):
...
+            from_extensions_config = protocol.client.get_extensions_goal_state()
+            from_vm_settings = protocol.client._extensions_goal_state_from_vm_settings
+
+            # we set a the status upload value to a dummy value to verify whether the compare() method is checking it or not
+            from_vm_settings._status_upload_blob = "A DUMMY VALUE FOR THE STATUS UPLOAD BLOB"
+
+            #
+            # 112 and below should not check the status blob
+            #
+            from_vm_settings._host_ga_plugin_version = FlexibleVersion("1.0.8.112")
```
larohra, line 43: Maybe worth adding a check where we dont get any HostGA version at all too?

azurelinuxagent/common/protocol/extensions_goal_state.py
narrieta: * I'm guessing this check is temporary until this feature is stabilized? If so, could you please add a comment about it too. Thanks!
It will stay there for a while, probably it can be removed once we get telemetry indicating that there are no old HostGAPlugins in the wild. For the moment this check is there just to avoid generating noise. No harm if this is not removed so I'll leave it as is

* On that note, do the latest version of HostGA contain all the attributes that we're parsing? Or will we have to have a similar check like this for other versions too in the future?

We will have additional checks on the version. The purpose of this PR is mainly  to set the infrastructure needed to perform those checks.

* 1 more question, what's the min version where the HostGA version is available? Is it .112 or greater than that?

Not sure of the exact version, but Flexible version defaults to 0.0.0.0 so version 0 indicates the version field is not there

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
narrieta: self.version is intialized to empty and each property defaults to 0, as in

    def major(self):
        return self.version[0] if len(self.version) > 0 else 0

azurelinuxagent/common/protocol/wire.py
narrieta: thanks; fixed

tests/protocol/test_extensions_goal_state.py
narrieta: didn't consider it interesting, since flexible version inits to 0

tests/data/hostgaplugin/vm_settings.json
```diff
@@ -1,4 +1,6 @@
 {
+    "hostGAPluginVersion":"1.0.8.115",
+    "vmSettingsSchemaVersion":"0.0",
```
kevinclark19a, line 3: Just to verify, these fields have always been present in real VmSettings objects, right?

kevinclark19a (dismissed):

larohra (approved):

azurelinuxagent/common/protocol/extensions_goal_state.py
larohra: > We will have additional checks on the version. The purpose of this PR is mainly to set the infrastructure needed to perform those checks.

I am personally not a big fan of putting checks on versions because they can be never ending. But at the same time I'm not sure of a better way to do it either so approved the PR

tests/data/hostgaplugin/vm_settings.json
narrieta: They were added post 112, not sure when. We default to 0.0.0.0 if not present

dhivyaganesan (approved):

PR #2418, opened 2021-11-18
Parse onHold in vmSettings

narrieta:
While doing this I also initialized the FlexibleVersions explicitly. Although the default FlexibleVersion is equivalent to version 0.0.0.0, it displays kinda funny. Explicit initialization displays slightly better.
```

>>> f = FlexibleVersion()
>>> f0 = FlexibleVersion('0.0.0.0')
>>> f1 = FlexibleVersion('1.0.0.0')
>>> f.major, f.minor
(0, 0)
>>> f > f1
False
>>> f == f0
True
>>> f
FlexibleVersion ('', '.', ('alpha', 'beta', 'rc'))
>>> f0
FlexibleVersion ('0.0.0.0', '.', ('alpha', 'beta', 'rc'))
```

### DIFF
--- a/azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
+++ b/azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
@@ -30,8 +30,8 @@ class ExtensionsGoalStateFromVmSettings(ExtensionsGoalState):
         super(ExtensionsGoalStateFromVmSettings, self).__init__()
         self._id = etag
         self._text = json_text
-        self._host_ga_plugin_version = FlexibleVersion()
-        self._schema_version = FlexibleVersion()
+        self._host_ga_plugin_version = FlexibleVersion('0.0.0.0')
+        self._schema_version = FlexibleVersion('0.0.0.0')
         self._activity_id = None
         self._correlation_id = None
         self._created_on_timestamp = None
@@ -100,15 +100,21 @@ class ExtensionsGoalStateFromVmSettings(ExtensionsGoalState):
         # TODO: Parse all atttributes
 
     def _parse_simple_attributes(self, vm_settings):
-        host_ga_plugin_version = vm_settings.get("hostGAPluginVersion")
-        if host_ga_plugin_version is not None:
-            self._host_ga_plugin_version = FlexibleVersion(host_ga_plugin_version)
-        schema_version = vm_settings.get("vmSettingsSchemaVersion")
-        if schema_version is not None:
-            self._schema_version = FlexibleVersion(schema_version)
         self._activity_id = self._string_to_id(vm_settings.get("activityId"))
         self._correlation_id = self._string_to_id(vm_settings.get("correlationId"))
         self._created_on_timestamp = self._ticks_to_utc_timestamp(vm_settings.get("extensionsLastModifiedTickCount"))
+
+        host_ga_plugin_version = vm_settings.get("hostGAPluginVersion")
+        if host_ga_plugin_version is not None:
+            self._host_ga_plugin_version = FlexibleVersion(host_ga_plugin_version)
+
+        schema_version = vm_settings.get("vmSettingsSchemaVersion")
+        if schema_version is not None:
+            self._schema_version = FlexibleVersion(schema_version)
+
+        on_hold = vm_settings.get("onHold")
+        if on_hold is not None:
+            self._on_hold = on_hold
 
     def _parse_status_upload_blob(self, vm_settings):
         status_upload_blob = vm_settings.get("statusUploadBlob")
--- a/tests/protocol/mocks.py
+++ b/tests/protocol/mocks.py
@@ -21,16 +21,12 @@ from azurelinuxagent.common.utils import restutil
 from tests.tools import patch
 from tests.protocol import mockwiredata
 
-# regex used to determine whether to use the mock wireserver data
-_USE_MOCK_WIRE_DATA_RE = re.compile(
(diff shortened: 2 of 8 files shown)


tests/protocol/mocks.py
```diff
@@ -101,15 +97,17 @@ def http_request(method, url, data, **kwargs):
                 return return_value
 
         # if the request was not handled try to use the mock wireserver data
-        if _USE_MOCK_WIRE_DATA_RE.match(url) is not None:
```
narrieta, line 104: When I designed the mock protocol I wanted requests that would return mock data to be explicit (they had to start with http://mock-goal-state (or https)). As I've been adding more data files this becomes cumbersome. Furthermore, the underlying implementation of the mocks in mockwiredata.py always try to use mock data.

While working on this PR I forgot to mark my mock data with "mock-goal-state" and all the new unit tests were ok, but some existing tests were failing. Turns out that the test request was failing on the direct call and was falling back to the HostGAPlugin, changing the default channel in the process.

Some of the HostGAPlugin were assuming the initial state of the default channel, but that is incorrect, since it is actually a static variable that can be changed as a side effect of other test.

I fixed the HostGAPlugin tests, but I also dropped the enforcement of "mock-goal-state" in the test data. Not only the test data is easier to write, but now the behaviour of this mock is consistent with the behaviour of the mock HostGAPlugin (both of them always try to use mock data)

kevinclark19a (approved):

tests/protocol/mocks.py
```diff
@@ -101,15 +97,17 @@ def http_request(method, url, data, **kwargs):
...
+        try:
             if method == 'GET':
                 return protocol.mock_wire_data.mock_http_get(url, **kwargs)
             if method == 'POST':
                 return protocol.mock_wire_data.mock_http_post(url, data, **kwargs)
             if method == 'PUT':
                 return protocol.mock_wire_data.mock_http_put(url, data, **kwargs)
+        except NotImplementedError:
```
dhivyaganesan, line 107: Is NotImplementedError a temporary error we raise for development purpose ? Just want to make sure why we are not doing anything with it after catching.

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
```diff
@@ -100,15 +100,21 @@ def _parse_vm_settings(self, json_text):
...
         if host_ga_plugin_version is not None:
             self._host_ga_plugin_version = FlexibleVersion(host_ga_plugin_version)
+
         schema_version = vm_settings.get("vmSettingsSchemaVersion")
         if schema_version is not None:
             self._schema_version = FlexibleVersion(schema_version)
-        self._activity_id = self._string_to_id(vm_settings.get("activityId"))
-        self._correlation_id = self._string_to_id(vm_settings.get("correlationId"))
-        self._created_on_timestamp = self._ticks_to_utc_timestamp(vm_settings.get("extensionsLastModifiedTickCount"))
+
```
dhivyaganesan, line 114: Will it be a good idea to add an comment about what "onHold" is as you have explained in the chat ? - "if overprovisioning is enabled, when creating a scale set, a few extra machines are created and they are put "on hold" in case some of the other machines fail"

I understand we can always refer to other documentation

dhivyaganesan (approved):

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
narrieta: Thanks for the suggestion, but explaining that in the code at this point is just too much. Do we need to explain what "onHold" is? what the WireServer is? What Fabric is?, etc. I'd suggest asking in the internal chat or checking the internal documentation for that level of detail. Do feel free to ask in PRs (or the internal chat) when something is not clear to you.

tests/protocol/mocks.py
narrieta: Yes, good comment. I am not totally happy with the choice of NotImplemented, since that exception sort of gives the impression that this is temporary, as you point out. I used it as meaning that support for the specific "url" given as parameter is not supported by the mock_http_* functions (you can see when this exception is raised in these functions). Initially I considered a TypeError. but that is currently used to indicated a bad request (e.g incorrect headers). In the cases where NotImplemented is raised for a specific request, it is totally possible to add support for that request in the mocks, even if right now they are not supported, so I went for that error.

As far as why we are not doing anything in the except clause, that is explained in a comment a couple of lines below: if the mock_http_* functions don't handle the request we need to continue and fail the request or pass it to the original function we are mocking, depending on the flag passed as argument to this function.

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
dhivyaganesan: Yes agreed

tests/protocol/mocks.py
dhivyaganesan: Makes sense. Thanks for explaining.

tests/protocol/test_hostplugin.py
```diff
@@ -244,6 +244,8 @@ def test_default_channel(self, patch_put, patch_upload, _):
         as part of status upload.
         """
         with self.create_mock_protocol() as wire_protocol:
+            wire.HostPluginProtocol.is_default_channel = False
```
larohra, line 247: would it be worthwhile to add this as a mock inside `create_mock_protocol` since we want the state to be clean pretty much after every run?

larohra (approved):

tests/protocol/test_hostplugin.py
narrieta: probably, but that requires additional changes in those tests, and probably many of the tests in this module have similar dependencies.

i fixed only the ones that broke

PR #2422, opened 2021-11-23
Fix failing cases in DCR v2

larohra:
## Description
- We were not fetching all the reports in case of a VMSS, fixed that
- Some runs used to fail due to an extra quote (`'`) in the variable names. This is a known issue in bash scripts in Azure Pipelines. Fixed that too

Issue # 

---

### PR information

### DIFF
--- a/dcr/scenarios/ext-seq-multiple-dependencies/run.py
+++ b/dcr/scenarios/ext-seq-multiple-dependencies/run.py
@@ -1,3 +1,5 @@
+import socket
+
 from dcr.scenario_utils.check_waagent_log import check_waagent_log_for_errors
 from dcr.scenario_utils.test_orchestrator import TestFuncObj, TestOrchestrator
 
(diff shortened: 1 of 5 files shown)


dcr/scenarios/ext-seq-multiple-dependencies/run.py
```diff
@@ -9,5 +11,5 @@
 
     test_orchestrator = TestOrchestrator("ExtSeqDependency-VM", tests=tests)
     test_orchestrator.run_tests()
-    test_orchestrator.generate_report_on_vm("test-result-ext-seq-vm.xml")
+    test_orchestrator.generate_report_on_vm(f"test-result-ext-seq-vm-{socket.gethostname()}.xml")
```
dhivyaganesan, line 14: Will socket.gethostname() always return a hostname ? Just wondering if this will break if the connection is lost

larohra: This script is run inside the test VM. This call is equivalent to calling `hostname` command on the terminal. There's no network connection required as such. If we're unable to SSH into the VM at all to execute this script, that's a bigger problem then which is handled differently

dhivyaganesan (approved):

narrieta (approved):

PR #2425, opened 2021-11-29
Remove agent blacklisting logic

larohra:
## Description
Older versions of the agent used to blacklist new agent upgrades if they ever failed to auto-upgrade. Removing that logic now

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -1065,7 +1065,7 @@ class UpdateHandler(object):
         for path in glob.glob(os.path.join(conf.get_lib_dir(), AGENT_DIR_GLOB, AGENT_ERROR_FILE)):
             try:
                 os.remove(path)
-            except:
+            except Exception:
                 # Make a best case effort to delete the the error.json file
                 pass
 
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -49,9 +49,9 @@ from azurelinuxagent.common.osutil import get_osutil, systemd
 from azurelinuxagent.common.protocol.util import get_protocol_util
 from azurelinuxagent.common.protocol.hostplugin import HostPluginProtocol
 from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
-from azurelinuxagent.common.version import AGENT_NAME, AGENT_VERSION, AGENT_DIR_PATTERN, CURRENT_AGENT,\
+from azurelinuxagent.common.version import AGENT_NAME, AGENT_VERSION, AGENT_DIR_PATTERN, CURRENT_AGENT, \
     CURRENT_VERSION, DISTRO_NAME, DISTRO_VERSION, is_current_agent_installed, get_lis_version, \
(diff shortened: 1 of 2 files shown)


azurelinuxagent/ga/update.py
```diff
@@ -275,15 +275,6 @@ def run_latest(self, child_args=None):
                     is_success=False,
                     message=msg)
 
-            if ret is not None and ret > 0:
```
narrieta, line 278: should we keep the log statement?

azurelinuxagent/ga/update.py
```diff
@@ -1069,6 +1056,19 @@ def __upgrade_agent_if_permitted(self):
...
+    def _ensure_no_error_json_files():
+        """
+        Older versions of the agent used to maintain an error.json file to blacklist agents if they ever failed to start.
+        We've since deprecated that behavior. Delete any error.json files that might still exist to keep a clean state.
+        """
+        for path in glob.glob(os.path.join(conf.get_lib_dir(), AGENT_DIR_GLOB, AGENT_ERROR_FILE)):
+            try:
+                os.remove(path)
+            except Exception:
+                # Make a best case effort to delete the the error.json file
```
narrieta, line 1069: log a message?

azurelinuxagent/ga/update.py
```diff
@@ -1132,47 +1127,20 @@ def get_agent_cmd(self):
...
-        try:
-            if not os.path.isdir(self.get_agent_dir()):
-                os.makedirs(self.get_agent_dir())
-            self.error.mark_failure(is_fatal=is_fatal)
-            self.error.save()
-            if self.error.is_blacklisted:
-                logger.warn(u"Agent {0} is permanently blacklisted", self.name)
-        except Exception as e:
-            logger.warn(u"Agent {0} failed recording error state: {1}", self.name, ustr(e))
+        return os.path.isfile(self.get_agent_manifest_path())
```
narrieta, line 1138: what does "available" mean? how is it different to "downloaded"?

azurelinuxagent/ga/update.py
narrieta: (actually, i would not even bother cleaning up those files, but if we consider this important enough then we should log the error)

azurelinuxagent/ga/update.py
larohra: it was a repeat of the one above, the only thing relevant here was the blacklisting logic in line 286 but since its not relevant anymore, I removed it

azurelinuxagent/ga/update.py
larohra: The only reason I added this is to make sure the daemon doesn't blacklist wrongfully due to some uncleaned state. 
And yes, I'll add a log

azurelinuxagent/ga/update.py
larohra: earlier available meant downloaded and not blacklisted and downloaded meant the availability of the manifest file inside the agent directory. Since blacklisted is not relevant anymore, removed `is_downloaded` and using `is_available` everywhere.

larohra:
Closing the PR as we decided to keep this feature for now

PR #2429, opened 2021-11-30
Add DNS TCP Firewall rule - Simplified logic

dhivyaganesan:
Modified the way to add the New rule to be more readable

The new rule : iptables -t security -I OUTPUT -d 168.63.129.16 -p tcp --destination-port 53 -j ACCEPT -w

Issue # 

---

### PR information


azurelinuxagent/common/utils/networkutil.py
```diff
@@ -155,6 +185,18 @@ def __get_common_drop_command_params(wait, command, destination):
...
 
+    # This rule allows DNS TCP request to wireserver ip for non root users
+    @staticmethod
+    def get_iptables_accept_dns_tcp_request_command(wait, command, destination):
+        cmd = AddFirewallRules.__get_iptables_base_command(wait)
+        if command == AddFirewallRules.get_insert_command():
+            cmd.extend(['-t', 'security', '-I', 'OUTPUT', '-d', destination, '-p', 'tcp', '--destination-port', '53', '-j', 'ACCEPT'])
+        if command == AddFirewallRules.get_check_command():
+            cmd.extend(['-t', 'security', '-C', 'OUTPUT', '-d', destination, '-p', 'tcp', '--destination-port', '53', '-j', 'ACCEPT'])
+        if command == AddFirewallRules.get_delete_command():
```
dhivyaganesan, line 196: Though we are not using this if block so far , as a part of next iteration I have plans to move the delete command we generate in default.py in here , so this block will be used

dhivyaganesan:
https://tuxgold.corp.microsoft.com/job/AzLinux/job/DungeonCrawler/job/AgentBVTs.dhivya/166/

This run is yellow because of unable to resolve unable to resolve host error for ubuntu 16. Will update it once I get a all green .

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -171,15 +213,21 @@ def get_iptables_drop_command(wait, command, destination):
...
         cmd.extend(
-            AddFirewallRules.__get_common_accept_command_params(wait, AddFirewallRules.__ACCEPT_COMMAND, destination,
+            AddFirewallRules.__get_common_accept_command_params(wait, AddFirewallRules.get_append_command(), destination,
                                                                 uid))
         return cmd
 
+    @staticmethod
+    def get_firewalld_accept_dns_tcp_command(command, destination):
+        cmd = AddFirewallRules.__get_firewalld_base_command(command)
+        cmd.extend(['-t', 'security', '-I', 'OUTPUT', '-d', destination, '-p', 'tcp', '--destination-port', '53', '-j', 'ACCEPT'])
```
larohra, line 223: I am not a big fan of the rule duplication for iptables and firewalld. I would prefer if the common code/rules can come from one single function to avoid code duplication as much as possible since it makes modifying it easier in the future if needed. 

Is there anyway that duplication can be avoided here too?

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -155,6 +185,18 @@ def __get_common_drop_command_params(wait, command, destination):
         cmd.extend(["conntrack", "--ctstate", "INVALID,NEW", "-j", "DROP"])
         return AddFirewallRules._add_wait(wait, cmd)
 
+    # This rule allows DNS TCP request to wireserver ip for non root users
+    @staticmethod
+    def get_iptables_accept_dns_tcp_request_command(wait, command, destination):
```
larohra, line 190: From the code it seems the expectation of this function is to act as a common place for fetching iptable rules for different commands, not just append/accept. Could you please modify the name here to reflect that? Maybe make it a bit more generic for iptables?

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -200,21 +248,27 @@ def __execute_cmd(cmd):
 
     @staticmethod
     def add_iptables_rules(wait, dst_ip, uid):
-
+        # The order in which the below rules are added matters for the ip table rules to work as expected
         AddFirewallRules.__raise_if_empty(dst_ip, "Destination IP")
         AddFirewallRules.__raise_if_empty(uid, "User ID")
 
-        accept_rule = AddFirewallRules.get_iptables_accept_command(wait, AddFirewallRules.__ACCEPT_COMMAND, dst_ip, uid)
+        accept_rule_dns_tcp_rule = AddFirewallRules.get_iptables_accept_dns_tcp_request_command(wait, AddFirewallRules.get_insert_command(),dst_ip)
```
larohra, line 255: NIT: 
> PEP 8: E231 missing whitespace after ','

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -200,21 +248,27 @@ def __execute_cmd(cmd):
...
+        AddFirewallRules.__execute_cmd(accept_rule_dns_tcp_rule)
+
+        accept_rule = AddFirewallRules.get_iptables_accept_command(wait, AddFirewallRules.get_append_command(), dst_ip, uid)
         AddFirewallRules.__execute_cmd(accept_rule)
 
-        drop_rule = AddFirewallRules.get_iptables_drop_command(wait, AddFirewallRules.__ACCEPT_COMMAND, dst_ip)
+        drop_rule = AddFirewallRules.get_iptables_drop_command(wait, AddFirewallRules.get_append_command(), dst_ip)
```
larohra, line 261: These names were a bit confusion, could you rename all of them to `get_iptables_drop_rule` instead of `get_iptables_drop_command`. Similarly for the others too (`get_iptables_accept_rule` and `get_iptables_accept_dns_tcp_request_rule`)

azurelinuxagent/common/utils/networkutil.py
larohra: If you're adding this rule at the beginning of the chain, it doesn't really need Insert right? Append would work too no?

azurelinuxagent/common/osutil/default.py
```diff
@@ -99,6 +102,11 @@ def _get_firewall_delete_conntrack_accept_command(wait, destination):
                       "--ctstate", "INVALID,NEW", "-j", "ACCEPT"])
 
 
+def _get_firewall_delete_accept_dns_tcp_command(wait, destination):
```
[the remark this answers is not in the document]
narrieta: @larohra - do you mean calling get_iptables_accept_dns_tcp_request_command from here? I think the current implementation is clearer.

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -119,7 +119,35 @@ class AddFirewallRules(object):
...
+    __INSERT_COMMAND = "-I"
+
+    # -D deletes the specific rule in the iptable chain
+    __DELETE_COMMAND = "-D"
+
+    # -C checks if a specific rule exists
+    __CHECK_COMMAND = "-C"
+
+    @staticmethod
+    def get_append_command():
```
narrieta, line 137: these getters seem an overkill (probably i am missing something) why not use the constants you defined above directly?

also, they don't seem to be used outside this class, so they should be private

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -155,6 +185,18 @@ def __get_common_drop_command_params(wait, command, destination):
         cmd.extend(["conntrack", "--ctstate", "INVALID,NEW", "-j", "DROP"])
         return AddFirewallRules._add_wait(wait, cmd)
 
+    # This rule allows DNS TCP request to wireserver ip for non root users
+    @staticmethod
+    def get_iptables_accept_dns_tcp_request_command(wait, command, destination):
+        cmd = AddFirewallRules.__get_iptables_base_command(wait)
+        if command == AddFirewallRules.get_insert_command():
+            cmd.extend(['-t', 'security', '-I', 'OUTPUT', '-d', destination, '-p', 'tcp', '--destination-port', '53', '-j', 'ACCEPT'])
+        if command == AddFirewallRules.get_check_command():
```
narrieta, line 194: Please use + instead of extend()... right now this works because __get_iptables_base_command always returns a new instance of the array, but if that changes this code would keep extending the same array.

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -200,21 +248,27 @@ def __execute_cmd(cmd):
...
-
+        # The order in which the below rules are added matters for the ip table rules to work as expected
         AddFirewallRules.__raise_if_empty(dst_ip, "Destination IP")
         AddFirewallRules.__raise_if_empty(uid, "User ID")
 
-        accept_rule = AddFirewallRules.get_iptables_accept_command(wait, AddFirewallRules.__ACCEPT_COMMAND, dst_ip, uid)
+        accept_rule_dns_tcp_rule = AddFirewallRules.get_iptables_accept_dns_tcp_request_command(wait, AddFirewallRules.get_insert_command(),dst_ip)
+        AddFirewallRules.__execute_cmd(accept_rule_dns_tcp_rule)
+
+        accept_rule = AddFirewallRules.get_iptables_accept_command(wait, AddFirewallRules.get_append_command(), dst_ip, uid)
```
narrieta, line 258: can we be consistent and either append or insert all the rules?

azurelinuxagent/common/osutil/default.py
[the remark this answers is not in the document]
larohra: yes (with better name for the function ofcourse), but yeah that's the idea. From the implementation, it looks like the only thing different is the command we pass to it (`-A` vs `-D`). If that is indeed the case, I would rather have one single function for it rather than copies of the same thing to avoid code duplication primarily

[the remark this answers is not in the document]
narrieta: OK, we can try to reuse code, but the previous abstraction was an overkill and made the code really hard to verify.  I am not concerned about having some duplication for just 4 commands. The resulting code should be fairly simple and easy to read. I think a few array constants and array concatenation would do the trick, but @dhivyaganesan can decide that.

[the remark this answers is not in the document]
larohra: I totally agree with you, array constants and concatenations would solve my comments too :)

azurelinuxagent/common/utils/networkutil.py
dhivyaganesan: There is no issue using append everywhere when it comes to iptable rules.However,there is an issue when it comes to the Distros that persist the firewall rules using firewalld

For example in RHEL , since it persist the rules we have to use insert in order to maintain the order. Therefore an insert for the new rule in case of persist firewall scenario using firewalld is a must since the older agent has already set the older ACCEPT and DROP rule

I considered using Insert in both iptable rule and firewalld . I that case insert needs an index to be provided for rule number 2 and 3

Example: 
['iptables', '-t', 'security', '-A', 'OUTPUT', **'2'** '-d', '168.63.129.16', '-p', 'tcp', '-m', 'owner', '--uid-owner', '0', '-j', 'ACCEPT', '-w']

But since the other commands such as DELETE and CHECK need not require an index , the function will have an additional parameter that will be used only for insert . I felt it is a bit clumsy.
 
Therefore I settled with using append for all iptable rules and insert for the new rule for firewalld.

I can change it to insert everywhere and write logic to use the index only in case of insert if that is suggested.

azurelinuxagent/common/utils/networkutil.py
dhivyaganesan: Sure, Addressed this in the next commit

azurelinuxagent/common/utils/networkutil.py
dhivyaganesan: Since we are using the insert , check and delete command in update ,test_update and default files I used getters .So I thought of using it everywhere for consistency.  Any other suggestions to avoid them ?

azurelinuxagent/common/utils/networkutil.py
dhivyaganesan: Agreed, When it comes to iptable rules it doesn't matter. But for firewalld since its persistent we have to use insert. Earlier I used insert to be consistent with both iptable and firewalld . Now changed it to append for iptable

azurelinuxagent/common/utils/networkutil.py
dhivyaganesan: I have modified the name of function for just this rule as "get_firewall_accept_dns_tcp_rule" everywhere. Based on the input parameter it could be generated with "-A" or "-D" or "-I" or "-C"

in the next PR I will rename other rules as "get_firewall_root_accept_rule" ( or something better) and "get_firewall_drop_rule"

azurelinuxagent/common/utils/networkutil.py
dhivyaganesan: I have modified the logic a bit now with one optional parameter for firewalld . For firewalld we need to specify if its a "--query-passthrough " or "--passthrough" . if it is not specified then I generate the iptable command. 

Example:  

firewall-cmd --permanent --direct **--query-passthrough** ipv4 _-t security _-I OUTPUT -d 168.63.129.16 -p tcp --destination-port 53 -j ACCEPT__

firewall-cmd --permanent --direct **--passthrough** ipv4 _-t security _-I OUTPUT -d 168.63.129.16 -p tcp --destination-port 53 -j ACCEPT__

iptables -w _-t security -I OUTPUT -d 168.63.129.16 -p tcp --destination-port 53 -j ACCEPT_

Please let me know you thoughts on this @narrieta @larohra

azurelinuxagent/common/utils/networkutil.py
dhivyaganesan: I have modified the name as ""get_firewall_accept_dns_tcp_rule" consistently. For next PR after I change the implementation of other two rules I will modified the names of the to use "rule" everywhere

azurelinuxagent/common/osutil/default.py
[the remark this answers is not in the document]
dhivyaganesan: I haven't moved the delete rule from default to networkutil yet. I will do it when I modify all the rules and move everything to networkutil. This required a remane in many test files and I feel its easier to do in one sitting for all rules after the logic change is implemented

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -155,6 +185,18 @@ def __get_common_drop_command_params(wait, command, destination):
         cmd.extend(["conntrack", "--ctstate", "INVALID,NEW", "-j", "DROP"])
         return AddFirewallRules._add_wait(wait, cmd)
 
+    # This rule allows DNS TCP request to wireserver ip for non root users
```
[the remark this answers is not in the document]
dhivyaganesan: Addressed

azurelinuxagent/common/osutil/default.py
```diff
@@ -74,6 +74,9 @@ def _add_wait(wait, command):
 def _get_iptables_version_command():
     return ["iptables", "--version"]
 
+def _get_firewall_accept_dns_tcp_command(wait, command, destination):
```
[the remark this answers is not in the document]
dhivyaganesan: Addressed

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -131,7 +159,9 @@ def _add_wait(wait, command):
         return command
 
     @staticmethod
-    def __get_iptables_base_command():
+    def __get_iptables_base_command(wait=""):
```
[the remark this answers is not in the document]
dhivyaganesan: Addressed

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -155,6 +185,18 @@ def __get_common_drop_command_params(wait, command, destination):
         cmd.extend(["conntrack", "--ctstate", "INVALID,NEW", "-j", "DROP"])
         return AddFirewallRules._add_wait(wait, cmd)
 
+    # This rule allows DNS TCP request to wireserver ip for non root users
+    @staticmethod
+    def get_iptables_accept_dns_tcp_request_command(wait, command, destination):
+        cmd = AddFirewallRules.__get_iptables_base_command(wait)
+        if command == AddFirewallRules.get_insert_command():
```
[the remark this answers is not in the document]
dhivyaganesan: Agreed and addressed

dhivyaganesan:
https://tuxgold.corp.microsoft.com/job/AzLinux/job/DungeonCrawler/job/AgentBVTs.dhivya/184/ - Green run for the latest changes

azurelinuxagent/common/osutil/default.py
```diff
@@ -71,40 +71,47 @@ def _add_wait(wait, command):
     return command
 
 
-def _get_iptables_version_command():
+def get_iptables_version_command():
     return ["iptables", "--version"]
 
+def get_firewall_accept_dns_tcp_rule(wait, command, destination):
```
larohra, line 77: Why do we need this function at all? Why can't we directly just use `AddFirewallRules.get_firewall_accept_dns_tcp_rule` instead?

azurelinuxagent/common/osutil/default.py
```diff
@@ -71,40 +71,47 @@ def _add_wait(wait, command):
...
+    return AddFirewallRules.get_firewall_accept_dns_tcp_rule(wait, command, destination)
 
-def _get_firewall_accept_command(wait, command, destination, owner_uid):
+
+def get_firewall_accept_command(wait, command, destination, owner_uid):
     return AddFirewallRules.get_iptables_accept_command(wait, command, destination, owner_uid)
```
larohra, line 82: Same for this

azurelinuxagent/common/utils/networkutil.py
larohra: We can just make these variables as public and use them directly instead of using a getter :)

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -100,14 +100,16 @@ def __str__(self):
 
 
 class FirewallCmdDirectCommands(object):
-    # firewall-cmd --direct --passthrough ipv4 -t security -A OUTPUT -d 1.2.3.5 -p tcp -m owner --uid-owner 999 -j ACCEPT
+    # firewall-cmd --direct --permanent --passthrough ipv4 -t security -A OUTPUT -d 1.2.3.5 -p tcp -m owner --uid-owner 999 -j ACCEPT
```
larohra, line 103: Thanks for adding the `--permanent` flag

azurelinuxagent/ga/update.py
```diff
@@ -1017,6 +1023,49 @@ def _ensure_firewall_rules_persisted(dst_ip):
...
+
+        try:
+            wait = self.osutil.get_firewall_will_wait()
+
+            # "-C" checks if the iptable rule is available in the chain. It throws an exception with return code 1 if the ip table rule doesnt exist
+            drop_rule = get_firewall_drop_command(wait, AddFirewallRules.CHECK_COMMAND, dst_ip)
+            if not _execute_run_command(drop_rule):
+                # DROP command doesn't exist indicates then none of the firewall rules are set yet
+                # exiting here as the environment thread will set up all firewall rules
+                return
```
larohra, line 1047: Maybe add an info log here to be clear why we didnt do something, would be helpful while debugging

azurelinuxagent/ga/update.py
```diff
@@ -1017,6 +1023,49 @@ def _ensure_firewall_rules_persisted(dst_ip):
...
+                accept_tcp_rule = get_accept_tcp_rule(wait, AddFirewallRules.CHECK_COMMAND, dst_ip)
+                if not _execute_run_command(accept_tcp_rule):
+                    try:
+                        logger.info(
+                            "Firewall rule to allow DNS TCP request to wireserver for a non root user unavailable . Setting it now.")
+                        accept_tcp_rule = get_accept_tcp_rule(wait, AddFirewallRules.INSERT_COMMAND, dst_ip)
+                        shellutil.run_command(accept_tcp_rule)
+                        logger.info(
+                            "Succesfully added firewall rule to allow non root users to do a DNS TCP request to wireserver ")
+                    except Exception as e:
```
larohra, line 1059: NIT: This should be a check for `CommandError` and not `Exception` as the only thing that can potentially throw here is `shellutil.run_command`

azurelinuxagent/ga/update.py
```diff
@@ -1017,6 +1023,49 @@ def _ensure_firewall_rules_persisted(dst_ip):
...
+                return
+            else:
+                # DROP rule exists in the ip table chain. Hence checking if the DNS TCP to wireserver rule exists. If not we add it.
+                accept_tcp_rule = get_accept_tcp_rule(wait, AddFirewallRules.CHECK_COMMAND, dst_ip)
+                if not _execute_run_command(accept_tcp_rule):
+                    try:
+                        logger.info(
+                            "Firewall rule to allow DNS TCP request to wireserver for a non root user unavailable . Setting it now.")
+                        accept_tcp_rule = get_accept_tcp_rule(wait, AddFirewallRules.INSERT_COMMAND, dst_ip)
+                        shellutil.run_command(accept_tcp_rule)
```
larohra, line 1056: Note: this will not log the error by default. If you want the function to log the error please use `log_error=True`

> if 'log_error'
        is True, it also logs details about the error.

If you don't want the function to log the error, then add more details in the except catch below as to why the command failed. We need as much logging details as possible

tests/common/osutil/test_default.py
```diff
@@ -657,16 +658,19 @@ def set_command(command, output='', exit_code=0):
...
-        set_command(osutil._get_firewall_drop_command(wait, "-C", destination))
-        set_command(osutil._get_firewall_drop_command(wait, "-A", destination))
-        set_command(osutil._get_firewall_accept_command(wait, "-A", destination, uid))
+        version_command = set_command(osutil.get_iptables_version_command(), output=str(version))
+        list_command = set_command(osutil.get_firewall_list_command(wait), output="Mock Output")
+        set_command(osutil.get_firewall_packets_command(wait))
+        set_command(osutil.get_firewall_drop_command(wait, AddFirewallRules.CHECK_COMMAND, destination))
+        set_command(osutil.get_firewall_drop_command(wait, AddFirewallRules.APPEND_COMMAND, destination))
+        set_command(osutil.get_firewall_accept_command(wait, AddFirewallRules.APPEND_COMMAND, destination, uid))
+        set_command(osutil.get_accept_tcp_rule("-w" if wait else "", AddFirewallRules.APPEND_COMMAND, destination))
```
larohra, line 667: This check `"-w" if wait else ""` will be relevant for all the functions passing `wait`, maybe add this to all?

tests/common/osutil/test_default.py
```diff
@@ -736,19 +740,22 @@ def test_enable_firewall_should_set_up_the_firewall(self):
...
 
             self.assertTrue(success, "Enabling the firewall was not successful")
-            self.assertEqual(len(mock_iptables.command_calls), 3, "Incorrect number of calls to iptables: [{0}]". format(mock_iptables.command_calls))
+            self.assertEqual(len(mock_iptables.command_calls), 4, "Incorrect number of calls to iptables: [{0}]". format(mock_iptables.command_calls))
             self.assertEqual(mock_iptables.command_calls[0], drop_check_command, "The first command should check the drop rule")
-            self.assertEqual(mock_iptables.command_calls[1], accept_command, "The second command should add the accept rule")
-            self.assertEqual(mock_iptables.command_calls[2], drop_add_command, "The third command should add the drop rule")
+            self.assertEqual(mock_iptables.command_calls[1], accept_tcp_append_rule,
+                            "The second command should add the dns tcp accept rule")
+            self.assertEqual(mock_iptables.command_calls[2], accept_command, "The second command should add the accept rule")
```
larohra, line 757: Can you modify the failure messages here please - 

```suggestion
            self.assertEqual(mock_iptables.command_calls[2], accept_command, "The third command should add the accept rule")
```

tests/common/osutil/test_default.py
```diff
@@ -736,19 +740,22 @@ def test_enable_firewall_should_set_up_the_firewall(self):
...
+            self.assertEqual(mock_iptables.command_calls[3], drop_add_command, "The third command should add the drop rule")
```
larohra, line 758:
```suggestion
            self.assertEqual(mock_iptables.command_calls[3], drop_add_command, "The fourth command should add the drop rule")
```

tests/common/osutil/test_default.py
```diff
@@ -757,12 +764,12 @@ def test_enable_firewall_should_not_use_wait_when_iptables_does_not_support_it(s
...
         with TestOSUtil._mock_iptables(version=osutil._IPTABLES_LOCKING_VERSION - 1) as mock_iptables:
             # fail the rule check to force enable of the firewall
-            mock_iptables.set_command(osutil._get_firewall_drop_command(mock_iptables.wait, "-C", mock_iptables.destination), exit_code=1)
+            mock_iptables.set_command(osutil.get_firewall_drop_command(mock_iptables.wait, AddFirewallRules.CHECK_COMMAND, mock_iptables.destination), exit_code=1)
 
             success = osutil.DefaultOSUtil().enable_firewall(dst_ip=mock_iptables.destination, uid=mock_iptables.uid)
 
             self.assertTrue(success, "Enabling the firewall was not successful")
-            self.assertEqual(len(mock_iptables.command_calls), 3, "Incorrect number of calls to iptables: [{0}]". format(mock_iptables.command_calls))
+            self.assertEqual(len(mock_iptables.command_calls), 4, "Incorrect number of calls to iptables: [{0}]". format(mock_iptables.command_calls))
```
larohra, line 772: Can you add a comment on the relevance of 4 here please

tests/common/osutil/test_default.py
```diff
@@ -845,13 +856,17 @@ def mock_popen(command, *args, **kwargs):
...
+                delete_conntrack_accept_command = TestOSUtil._command_to_string(osutil.get_firewall_delete_conntrack_accept_command(mock_iptables.wait, mock_iptables.destination))
+                delete_accept_tcp_rule = TestOSUtil._command_to_string(
+                    osutil.get_delete_accept_tcp_rule(mock_iptables.wait, mock_iptables.destination))
+                delete_owner_accept_command = TestOSUtil._command_to_string(osutil.get_firewall_delete_owner_accept_command(mock_iptables.wait, mock_iptables.destination, mock_iptables.uid))
+                delete_conntrack_drop_command = TestOSUtil._command_to_string(osutil.get_firewall_delete_conntrack_drop_command(mock_iptables.wait, mock_iptables.destination))
 
                 self.assertTrue(success, "Removing the firewall should have succeeded")
-                self.assertEqual(len(delete_commands), 3, "Expected 3 delete commands: [{0}]".format(delete_commands))
+                self.assertEqual(len(delete_commands), 4, "Expected 3 delete commands: [{0}]".format(delete_commands))
```
larohra, line 866:
```suggestion
                self.assertEqual(len(delete_commands), 4, "Expected 4 delete commands: [{0}]".format(delete_commands))
```

tests/common/osutil/test_default.py
```diff
@@ -845,13 +856,17 @@ def mock_popen(command, *args, **kwargs):
...
                 # delete rules < 2.2.26
+                self.assertIn(delete_accept_tcp_rule, delete_commands, "The delete dns tcp accept command was not executed")
+                self.assertEqual(delete_commands[delete_accept_tcp_rule], 2, "The delete dns tcp accept command should have been executed twice")
```
larohra, line 869: > "The delete dns tcp accept command should have been executed twice"

Why? Can you add a comment please?

tests/common/osutil/test_default.py
larohra: can you also add a comment for the relevance of `4`?

tests/ga/test_update.py
```diff
@@ -1670,15 +1673,85 @@ def _mock_popen(cmd, *args, **kwargs):
...
+        self.assertEqual(4, len(executed_commands),
+                         "The number of times firewall-cmd should be called is only 4; Executed firewall commands: {0}; All popen calls: {1}".format(
                              executed_commands, mock_popen.call_args_list))
         self.assertEqual(PersistFirewallRulesHandler._FIREWALLD_RUNNING_CMD, executed_commands.pop(0),
                          "First command should be to check if firewalld is running")
         self.assertTrue([FirewallCmdDirectCommands.QueryPassThrough in cmd for cmd in executed_commands],
                         "The remaining commands should only be for querying the firewall commands")
 
+    def test_it_should_set_dns_tcp_iptable_if_drop_available_accept_unavailable(self):
+        osutil._enable_firewall = True
```
larohra, line 1686: Please refrain from setting a static variable during test execution. This has the potential of breaking other unrelated test cases. The test cases should be self-contained, the mocks we use, the properties we set should ONLY be relevant inside the test, not outside. 

Please change this to use a mock instead, something like - 

`with patch.object(osutil, '_enable_firewall', True):`

tests/common/osutil/test_default.py
```diff
@@ -657,16 +658,19 @@ def set_command(command, output='', exit_code=0):
...
-        set_command(osutil._get_firewall_drop_command(wait, "-A", destination))
-        set_command(osutil._get_firewall_accept_command(wait, "-A", destination, uid))
+        version_command = set_command(osutil.get_iptables_version_command(), output=str(version))
+        list_command = set_command(osutil.get_firewall_list_command(wait), output="Mock Output")
+        set_command(osutil.get_firewall_packets_command(wait))
+        set_command(osutil.get_firewall_drop_command(wait, AddFirewallRules.CHECK_COMMAND, destination))
+        set_command(osutil.get_firewall_drop_command(wait, AddFirewallRules.APPEND_COMMAND, destination))
+        set_command(osutil.get_firewall_accept_command(wait, AddFirewallRules.APPEND_COMMAND, destination, uid))
+        set_command(osutil.get_accept_tcp_rule("-w" if wait else "", AddFirewallRules.APPEND_COMMAND, destination))
+        set_command(osutil.get_accept_tcp_rule("-w" if wait else "", AddFirewallRules.INSERT_COMMAND, destination))
```
larohra, line 668: Why dont you also add the `CHECK_COMMAND` here too? 

i.e. - 
> set_command(osutil.get_accept_tcp_rule("-w" if wait else "", AddFirewallRules.CHECK_COMMAND, destination))

tests/ga/test_update.py
```diff
@@ -1670,15 +1673,85 @@ def _mock_popen(cmd, *args, **kwargs):
...
+                mock_iptables.set_command(osutil.get_firewall_drop_command(mock_iptables.wait, AddFirewallRules.CHECK_COMMAND, mock_iptables.destination), exit_code=0)
+                # non root tcp iptable rule is absent
+                mock_iptables.set_command(osutil.get_accept_tcp_rule(mock_iptables.wait, AddFirewallRules.CHECK_COMMAND, mock_iptables.destination), exit_code=1)
+                update_handler.run(debug=True)
+
+                drop_check_command = TestOSUtil._command_to_string(osutil.get_firewall_drop_command(mock_iptables.wait, AddFirewallRules.CHECK_COMMAND, mock_iptables.destination))
+                accept_tcp_check_rule = TestOSUtil._command_to_string(osutil.get_accept_tcp_rule(mock_iptables.wait, AddFirewallRules.CHECK_COMMAND, mock_iptables.destination))
+                accept_tcp_insert_rule = TestOSUtil._command_to_string(osutil.get_accept_tcp_rule(mock_iptables.wait, AddFirewallRules.INSERT_COMMAND, mock_iptables.destination))
+
+                get_firewall_packets = TestOSUtil._command_to_string(osutil.get_firewall_packets_command(mock_iptables.wait))
```
larohra, line 1700: This is an irrelevant command for your use case. You can't keep adding all iptable calls here if we ever add anything in the future. Instead, you should filter out the `mock_iptables.command_calls` array and get only the commands you added and want to verify. Additionally, in this specific use case, the order matters too so make sure of that.

If any doubts of what I'm proposing, please ping me offline and we can get on a quick call.

tests/ga/test_update.py
```diff
@@ -1670,15 +1673,85 @@ def _mock_popen(cmd, *args, **kwargs):
...
+                                 "The first command should check the drop rule")
+                self.assertEqual(mock_iptables.command_calls[1], accept_tcp_check_rule,
+                                 "The second command should check the accept rule")
+                self.assertEqual(mock_iptables.command_calls[2], accept_tcp_insert_rule,
+                                 "The third command should add the accept rule")
+                self.assertEqual(mock_iptables.command_calls[3], get_firewall_packets,
+                                 "The fourth command should be for listing the firewall rules")
+
+    def test_it_should_not_set_dns_tcp_iptable_if_drop_unavailable(self):
+        osutil._enable_firewall = True
```
larohra, line 1713: Same as above

tests/ga/test_update.py
```diff
@@ -1670,15 +1673,85 @@ def _mock_popen(cmd, *args, **kwargs):
...
+
+        with TestOSUtil._mock_iptables() as mock_iptables:
+            with _get_update_handler(test_data=DATA_FILE) as (update_handler, _):
+                #drop rule is not available
```
larohra, line 1717: NIT: Pep 8 warning - 
> PEP 8: E265 block comment should start with '# '

(on the side note, please enable PEP8 inspection in pycharm :) )

tests/ga/test_update.py
```diff
@@ -1670,15 +1673,85 @@ def _mock_popen(cmd, *args, **kwargs):
...
+                mock_iptables.set_command(osutil.get_firewall_drop_command(mock_iptables.wait, AddFirewallRules.CHECK_COMMAND, mock_iptables.destination), exit_code=1)
+
+                update_handler.run(debug=True)
+
+                drop_check_command = TestOSUtil._command_to_string(osutil.get_firewall_drop_command(mock_iptables.wait, AddFirewallRules.CHECK_COMMAND, mock_iptables.destination))
+                get_firewall_packets_command = TestOSUtil._command_to_string(osutil.get_firewall_packets_command(mock_iptables.wait))
```
larohra, line 1723: same comment as above regarding this

tests/ga/test_update.py
```diff
@@ -1670,15 +1673,85 @@ def _mock_popen(cmd, *args, **kwargs):
...
+
+                self.assertEqual(len(mock_iptables.command_calls), 2,"Incorrect number of calls to iptables: [{0}]".format(mock_iptables.command_calls))
+                self.assertEqual(mock_iptables.command_calls[0], drop_check_command,
+                                 "The first command should check the drop rule")
+                self.assertEqual(mock_iptables.command_calls[1], get_firewall_packets_command,
+                                 "The fourth command should be for listing the firewall rules")
+
+    def test_it_should_not_set_dns_tcp_iptable_if_drop_and_accept_available(self):
+        osutil._enable_firewall = True
```
larohra, line 1732: same as above

tests/ga/test_update.py
```diff
@@ -1670,15 +1673,85 @@ def _mock_popen(cmd, *args, **kwargs):
...
+            with _get_update_handler(test_data=DATA_FILE) as (update_handler, _):
+                #drop rule is available
+                mock_iptables.set_command(osutil.get_firewall_drop_command(mock_iptables.wait, AddFirewallRules.CHECK_COMMAND, mock_iptables.destination), exit_code=0)
+                # non root tcp iptable rule is available
+                mock_iptables.set_command(osutil.get_accept_tcp_rule(mock_iptables.wait, AddFirewallRules.CHECK_COMMAND, mock_iptables.destination), exit_code=0)
+
+                update_handler.run(debug=True)
+
+                drop_check_command = TestOSUtil._command_to_string(osutil.get_firewall_drop_command(mock_iptables.wait, AddFirewallRules.CHECK_COMMAND, mock_iptables.destination))
+                get_firewall_packets_command = TestOSUtil._command_to_string(osutil.get_firewall_packets_command(mock_iptables.wait))
```
larohra, line 1744: same as above

larohra (changes requested):

tests/ga/test_update.py
dhivyaganesan: Addressed

tests/ga/test_update.py
dhivyaganesan: Addressed

tests/ga/test_update.py
dhivyaganesan: Addressed

tests/ga/test_update.py
dhivyaganesan: Addressed. Hopefully I did not miss anything more this time :)

tests/ga/test_update.py
dhivyaganesan: Addressed

tests/ga/test_update.py
dhivyaganesan: Yes I agree, I now filter the command_calls for only relevant commands. Please let me know your thoughts. In this way the order is maintained.

tests/common/osutil/test_default.py
dhivyaganesan: Addressed

tests/ga/test_update.py
dhivyaganesan: Thanks for this suggestion , I made the changes and understood the importance of not setting a static variable in test. I found other instance where _enable_firewall was set in Test in test_default, updated that as well.

tests/common/osutil/test_default.py
dhivyaganesan: I think the comment in the test which was already included explains why all the rules will be executed twice(not just the new rule). Do you still want to make it more explicit ?

                    # The agent invokes the delete commands continuously until they return 1 to indicate the rules has been removed
                    # The mock returns 0 (success) the first time it is invoked and 1 (rule does not exist) thereafter

tests/common/osutil/test_default.py
dhivyaganesan: Addressed

tests/common/osutil/test_default.py
dhivyaganesan: Addressed

tests/common/osutil/test_default.py
dhivyaganesan: Addressed

tests/common/osutil/test_default.py
dhivyaganesan: Addressed

tests/common/osutil/test_default.py
dhivyaganesan: Addressed

azurelinuxagent/ga/update.py
dhivyaganesan: I have modified the error to have more details . Now it will log something like this .

`Unable to set the non root tcp access firewall rule :['iptables', '-t', 'security', '-C', 'OUTPUT', '-d', '168.63.129.16', '-p', 'tcp', '-m', 'conntrack', '--ctstate', 'INVALID,NEW', '-j', 'DROP', '-w'].Run command execution for iptables -t security -C OUTPUT -d 168.63.129.16 -p tcp -m conntrack --ctstate INVALID,NEW -j DROP -w failed with error:iptables: Bad rule (does a matching rule exist in that chain?).
.Return Code:1`

azurelinuxagent/ga/update.py
dhivyaganesan: Addressed

azurelinuxagent/ga/update.py
dhivyaganesan: Yes, added a log statement

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -200,21 +227,30 @@ def __execute_cmd(cmd):
...
 
     @staticmethod
     def __execute_firewalld_commands(command, dst_ip, uid):
         AddFirewallRules.__raise_if_empty(dst_ip, "Destination IP")
         AddFirewallRules.__raise_if_empty(uid, "User ID")
 
+        # Firwalld.service fails if we set `-w` in the iptables command, so not adding it at all for firewalld commands
+        wait = False
+
+        accept_tcp_rule = AddFirewallRules.get_accept_tcp_rule(wait, AddFirewallRules.INSERT_COMMAND, dst_ip, firewalld_command=command)
```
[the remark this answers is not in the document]
dhivyaganesan: Yes changed in the next commit

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -200,21 +227,30 @@ def __execute_cmd(cmd):
 
     @staticmethod
     def add_iptables_rules(wait, dst_ip, uid):
-
+        # The order in which the below rules are added matters for the ip table rules to work as expected
         AddFirewallRules.__raise_if_empty(dst_ip, "Destination IP")
         AddFirewallRules.__raise_if_empty(uid, "User ID")
 
-        accept_rule = AddFirewallRules.get_iptables_accept_command(wait, AddFirewallRules.__ACCEPT_COMMAND, dst_ip, uid)
+        accept_tcp_rule = AddFirewallRules.get_accept_tcp_rule(wait, AddFirewallRules.APPEND_COMMAND, dst_ip, firewalld_command="")
```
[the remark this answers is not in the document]
dhivyaganesan: Addressed in the next commit

azurelinuxagent/common/osutil/default.py
```diff
@@ -71,42 +71,50 @@ def _add_wait(wait, command):
...
 
-def _get_firewall_packets_command(wait):
+def get_firewall_packets_command(wait):
     return _add_wait(wait, ["iptables", "-t", "security", "-L", "OUTPUT", "--zero", "OUTPUT", "-nxv"])
 
 
 # Precisely delete the rules created by the agent.
 # this rule was used <= 2.2.25.  This rule helped to validate our change, and determine impact.
-def _get_firewall_delete_conntrack_accept_command(wait, destination):
+def get_firewall_delete_conntrack_accept_command(wait, destination):
```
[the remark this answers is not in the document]
dhivyaganesan: Noted , will keep it in mind

azurelinuxagent/ga/update.py
```diff
@@ -1017,6 +1023,49 @@ def _ensure_firewall_rules_persisted(dst_ip):
...
+
+            # "-C" checks if the iptable rule is available in the chain. It throws an exception with return code 1 if the ip table rule doesnt exist
+            drop_rule = get_firewall_drop_command(wait, AddFirewallRules.get_check_command(), dst_ip)
+            if not _execute_run_command(drop_rule):
+                # DROP command doesn't exist indicates then none of the firewall rules are set yet
+                # exiting here as the environment thread will set up all firewall rules
+                return
+            else:
+                # DROP rule exists in the ip table chain. Hence checking if the DNS TCP to wireserver rule exists. If not we add it.
+                accept_non_root = get_firewall_accept_dns_tcp_rule(wait, AddFirewallRules.get_check_command(), dst_ip)
```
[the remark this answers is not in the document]
dhivyaganesan: Addressed

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -200,21 +243,30 @@ def __execute_cmd(cmd):
...
 
     @staticmethod
     def __execute_firewalld_commands(command, dst_ip, uid):
         AddFirewallRules.__raise_if_empty(dst_ip, "Destination IP")
         AddFirewallRules.__raise_if_empty(uid, "User ID")
 
+        # Wait is set to false for firewalld commands since it fails
+        wait = False
+
+        accept_rule_dns_tcp_rule = AddFirewallRules.get_firewall_accept_dns_tcp_rule(wait, AddFirewallRules.get_insert_command(), dst_ip, firewalldcommand=command)
```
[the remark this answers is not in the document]
dhivyaganesan: Addressed

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -200,21 +243,30 @@ def __execute_cmd(cmd):
...
+        drop_rule = AddFirewallRules.get_iptables_drop_command(wait, AddFirewallRules.get_append_command(), dst_ip)
         AddFirewallRules.__execute_cmd(drop_rule)
 
     @staticmethod
     def __execute_firewalld_commands(command, dst_ip, uid):
         AddFirewallRules.__raise_if_empty(dst_ip, "Destination IP")
         AddFirewallRules.__raise_if_empty(uid, "User ID")
 
+        # Wait is set to false for firewalld commands since it fails
+        wait = False
```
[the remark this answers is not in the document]
dhivyaganesan: Addressed in the next commit and made wait as optional parameter

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -155,6 +187,17 @@ def __get_common_drop_command_params(wait, command, destination):
         cmd.extend(["conntrack", "--ctstate", "INVALID,NEW", "-j", "DROP"])
         return AddFirewallRules._add_wait(wait, cmd)
 
+    @staticmethod
+    def get_firewall_accept_dns_tcp_rule(wait, command, destination, firewalldcommand=False):
```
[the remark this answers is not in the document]
dhivyaganesan: Addressed in the next commit

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -131,7 +161,9 @@ def _add_wait(wait, command):
         return command
 
     @staticmethod
-    def __get_iptables_base_command():
+    def __get_iptables_base_command(wait=False):
```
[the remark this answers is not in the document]
dhivyaganesan: Addressed in the next commit

azurelinuxagent/common/osutil/default.py
dhivyaganesan: As discussed offline all the methods in default.py will be moved to networkutil

azurelinuxagent/common/osutil/default.py
dhivyaganesan: As discussed offline all the methods in default.py will be moved to networkutil

dhivyaganesan:
https://tuxgold.corp.microsoft.com/job/AzLinux/job/DungeonCrawler/job/AgentBVTs.dhivya/186/
https://tuxgold.corp.microsoft.com/job/AzLinux/job/DungeonCrawler/job/AgentBVTs.dhivya/185/

DCR run for latest changes

@larohra  Thanks for the detailed review. Please have a look again when you get a chance .

tests/common/osutil/test_default.py
larohra: Instead of having a copy of the same statement, maybe make it a variable and pass that to each function? 

```python
wait_cmd = "-w" if wait else ""
set_command(osutil.get_firewall_packets_command(wait_cmd))
        set_command(osutil.get_firewall_drop_command(wait_cmd, AddFirewallRules.CHECK_COMMAND, destination))
```

azurelinuxagent/common/osutil/default.py
```diff
@@ -291,7 +300,7 @@ def enable_firewall(self, dst_ip, uid):
             logger.info("Successfully added Azure fabric firewall rules")
```
larohra, line 300: I can't add a comment there, but can you modify the comment on line 293 to say - `# Otherwise, append all rules`

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -200,21 +227,29 @@ def __execute_cmd(cmd):
...
-        drop_rule = AddFirewallRules.get_iptables_drop_command(wait, AddFirewallRules.__ACCEPT_COMMAND, dst_ip)
+        drop_rule = AddFirewallRules.get_iptables_drop_command(wait, AddFirewallRules.APPEND_COMMAND, dst_ip)
         AddFirewallRules.__execute_cmd(drop_rule)
 
     @staticmethod
     def __execute_firewalld_commands(command, dst_ip, uid):
         AddFirewallRules.__raise_if_empty(dst_ip, "Destination IP")
         AddFirewallRules.__raise_if_empty(uid, "User ID")
 
+        # Firwalld.service fails if we set `-w` in the iptables command, so not adding it at all for firewalld commands
```
larohra, line 248:
```suggestion
        # Firewalld.service fails if we set `-w` in the iptables command, so not adding it at all for firewalld commands
```
```suggestion
        # Firwalld.service fails if we set `-w` in the iptables command, so not adding it at all for firewalld commands
```

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -224,6 +259,8 @@ def __execute_firewalld_commands(command, dst_ip, uid):
     @staticmethod
     def add_firewalld_rules(dst_ip, uid):
         # Firwalld.service fails if we set `-w` in the iptables command, so not adding it at all for firewalld commands
```
larohra, line 261:
```suggestion
        # Firewalld.service fails if we set `-w` in the iptables command, so not adding it at all for firewalld commands
```
```suggestion
        # Firwalld.service fails if we set `-w` in the iptables command, so not adding it at all for firewalld commands
```

azurelinuxagent/ga/update.py
```diff
@@ -1017,6 +1023,52 @@ def _ensure_firewall_rules_persisted(dst_ip):
             message=msg,
             log_event=False)
 
+    def _add_accept_tcp_firewall_rule_if_not_enabled(self, dst_ip):
+        # Helper to execute a run command, returns True if no exception
```
larohra, line 1027: Can you move this comment inside the  `_execute_run_command` function please. It gave me the false sense that `_add_accept_tcp_firewall_rule_if_not_enabled` was doing this.

tests/ga/test_update.py
```diff
@@ -1680,15 +1683,85 @@ def _mock_popen(cmd, *args, **kwargs):
...
+                with patch.object(osutil, '_enable_firewall', True):
+                    # drop rule is not available
+                    mock_iptables.set_command(osutil.get_firewall_drop_command(mock_iptables.wait, AddFirewallRules.CHECK_COMMAND, mock_iptables.destination), exit_code=1)
+
+                    update_handler.run(debug=True)
+
+                    drop_check_command = TestOSUtil._command_to_string(osutil.get_firewall_drop_command(mock_iptables.wait, AddFirewallRules.CHECK_COMMAND, mock_iptables.destination))
+
+                    # Filtering the mock iptable command calls with only the once related to this test.
+                    filtered_mock_iptable_calls = [cmd for cmd in mock_iptables.command_calls if cmd in [drop_check_command]]
```
larohra, line 1735: This will only filter out the drop calls but if the APPEND/CHECK calls for accept_dns rules are there, this test would still succeed. Along with checking if the DROP rule was called, also check if the others were not called to have a more complete test.

tests/ga/test_update.py
```diff
@@ -1680,15 +1683,85 @@ def _mock_popen(cmd, *args, **kwargs):
...
+                    # non root tcp iptable rule is available
+                    mock_iptables.set_command(osutil.get_accept_tcp_rule(mock_iptables.wait, AddFirewallRules.CHECK_COMMAND, mock_iptables.destination), exit_code=0)
+
+                    update_handler.run(debug=True)
+
+                    drop_check_command = TestOSUtil._command_to_string(osutil.get_firewall_drop_command(mock_iptables.wait, AddFirewallRules.CHECK_COMMAND, mock_iptables.destination))
+                    accept_tcp_check_rule = TestOSUtil._command_to_string(osutil.get_accept_tcp_rule(mock_iptables.wait, AddFirewallRules.CHECK_COMMAND, mock_iptables.destination))
+
+                    # Filtering the mock iptable command calls with only the once related to this test.
+                    filtered_mock_iptable_calls = [cmd for cmd in mock_iptables.command_calls if cmd in [drop_check_command, accept_tcp_check_rule]]
```
larohra, line 1757: Same comment here, if the accept_tcp_append_rule is called, this filter will ignore it and pass the test irrespective. Please fix this too

larohra (changes requested):
Thanks for the changes, please fix the requested changes. Thanks!

azurelinuxagent/ga/update.py
dhivyaganesan: Addressed

azurelinuxagent/common/osutil/default.py
dhivyaganesan: Addressed

azurelinuxagent/common/utils/networkutil.py
dhivyaganesan: Addressed

azurelinuxagent/ga/update.py
dhivyaganesan: Addressed

tests/ga/test_update.py
dhivyaganesan: I agree with the suggestion of adding all the applicable rule to 'filtered_mock_iptable_calls` and I addressed it. However, since we are testing the length of `filtered_mock_iptable_calls` in each test and then checking the order and the particular rule that is called, I don't think it's necessary to check if a rule is not called. If in case a rule is called unintended the length check will fail, if not checks where we test the order will fail. Thoughts on this ?

tests/ga/test_update.py
dhivyaganesan: Same as above. Please share your thoughts

dhivyaganesan:
Green run for the above changes https://tuxgold.corp.microsoft.com/job/AzLinux/job/DungeonCrawler/job/AgentBVTs.dhivya/189/

tests/ga/test_update.py
larohra: /?how will the length check work here? Basically consider a buggy scenario where we execute 4 firewall commands. When this code is executed (`                    filtered_mock_iptable_calls = [cmd for cmd in mock_iptables.command_calls if cmd in [drop_check_command]]
`), the value of `filtered_mock_iptable_calls` will only contain one value because you're only checking the existence of `drop_check_command`, it wont contain the other 3 firewall calls. In this case, the `len(filtered_mock_iptable_calls)` will always be 1 irrespective of how many firewall calls there are. Is that understanding wrong? Am I missing something else here?

tests/ga/test_update.py
larohra: Thanks!

tests/ga/test_update.py
larohra: same as above

tests/common/osutil/test_default.py
larohra: missed that, thanks for pointing me to it

tests/common/osutil/test_default.py
```diff
@@ -657,16 +658,20 @@ def set_command(command, output='', exit_code=0):
...
         uid = 42
 
-        version_command = set_command(osutil._get_iptables_version_command(), output=str(version))
-        list_command = set_command(osutil._get_firewall_list_command(wait), output="Mock Output")
-        set_command(osutil._get_firewall_packets_command(wait))
-        set_command(osutil._get_firewall_drop_command(wait, "-C", destination))
-        set_command(osutil._get_firewall_drop_command(wait, "-A", destination))
-        set_command(osutil._get_firewall_accept_command(wait, "-A", destination, uid))
+        version_command = set_command(osutil.get_iptables_version_command(), output=str(version))
+        list_command = set_command(osutil.get_firewall_list_command("-w" if wait else ""), output="Mock Output")
```
larohra, line 662: Not sure where my comment got lost, but since you're using the `"-w" if wait else ""` in every line, can you make it a variable to keep the code clean? Thanks!

larohra (approved):

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -223,7 +258,9 @@ def __execute_firewalld_commands(command, dst_ip, uid):
 
     @staticmethod
     def add_firewalld_rules(dst_ip, uid):
-        # Firwalld.service fails if we set `-w` in the iptables command, so not adding it at all for firewalld commands
+        # Firewalld.service fails if we set `-w` in the iptables command, so not adding it at all for firewalld commands
+        # Firewalld.service with the "--permanent --passthrough" parameter ensures that a firewall rule is set only once even if command is executed multiple times
```
nagworld9, line 262: Good comment. I'm having similar question like same rule applying multiple times.

azurelinuxagent/ga/update.py
```diff
@@ -1017,6 +1023,53 @@ def _ensure_firewall_rules_persisted(dst_ip):
...
+
+        try:
+            wait = self.osutil.get_firewall_will_wait()
+
+            # "-C" checks if the iptable rule is available in the chain. It throws an exception with return code 1 if the ip table rule doesnt exist
+            drop_rule = get_firewall_drop_command(wait, AddFirewallRules.CHECK_COMMAND, dst_ip)
+            if not _execute_run_command(drop_rule):
+                # DROP command doesn't exist indicates then none of the firewall rules are set yet
+                # exiting here as the environment thread will set up all firewall rules
+                logger.info("DROP rule is not available which implies no firewall rules are set yet. Environment thread will set it up.")
```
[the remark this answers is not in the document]
dhivyaganesan: Yes, I see your point. This block of code is based on similar logic of how we set up the firewall rule on the env thread (Based on if block rule exists or not) . There where instance in the past where one rule was missed being set. To handle such cases we have a separate task to check all the rules periodically. That implementation should take care of duplicate rules being set as well.

tests/common/osutil/test_default.py
dhivyaganesan: Addressed

larohra (approved):

azurelinuxagent/ga/update.py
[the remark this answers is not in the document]
nagworld9: okay. I'm good as long as we are addressing it.

nagworld9 (approved):

PR #2430, opened 2021-12-01
Parse and compare extensions in vmSettings

narrieta:
Extension settings will come in a separate PR


azurelinuxagent/common/protocol/extensions_goal_state.py
```diff
@@ -82,7 +81,7 @@ def agent_manifests(self):
         raise NotImplementedError()
 
     @property
-    def ext_handlers(self):
+    def extensions(self):
```
narrieta, line 84: Renamed ExtHandler to Extension, which a more appropriate term.

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
```diff
@@ -232,8 +236,96 @@ def _parse_agent_manifests(self, vm_settings):
...
+        #         ...
+        #     ]
+        #     ...
+        # }
+        extension_goal_states = vm_settings.get("extensionGoalStates")
+        if extension_goal_states is not None:
+            if not isinstance(extension_goal_states, list):
+                raise Exception("extension_goal_states should be an array")
+            for extension_gs in extension_goal_states:
+                extension = Extension()
```
nagworld9, line 306: NIT: Looks like at this level of parsing we are calling these are `ext_handler` that's how it's defined in other places. Can you make it consistent?

larohra: We should change the names from `ext_handler` to `extension` in the other file rather than changing it here, that would make the intension more clearer

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
```diff
@@ -232,8 +236,96 @@ def _parse_agent_manifests(self, vm_settings):
...
+
+                extension.name = extension_gs['name']
+                extension.version = extension_gs['version']
+                extension.state = extension_gs['state']
+                if extension.state not in ExtensionRequestedState.All:
+                    raise Exception('Invalid extension state: {0} ({1})'.format(extension.state, extension.name))
+                is_multi_config = extension_gs.get('isMultiConfig')
```
larohra, line 313: Maybe you can modify the `CaseFoldedDict.get()` function to accept a default parameter to avoid using `None` checks in all these places

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1051,7 +1053,7 @@ def supports_multi_config(self):
 
     @property
     def extensions(self):
```
larohra, line 1055: Maybe change this property name to `extension_settings` to avoid confusion

tests/protocol/test_extensions_goal_state.py
```diff
@@ -34,6 +34,10 @@ def test_property(name, value):
             agent_manifests[0].family = 'MOCK_FAMILY'
             test_property("agent_manifests", agent_manifests)
 
+            extensions = copy.copy(from_vm_settings.extensions)
```
larohra, line 37: Any reason for not using deepcopy here? Maybe keep them similar for consistency?

tests/ga/test_extension.py
```diff
@@ -2503,20 +2497,19 @@ def _set_dependency_levels(self, dependency_levels, exthandlers_handler):
...
 
             handler = handler_map[handler_name]
-            for ext in handler.properties.extensions:
+            for ext in handler.settings:
                 ext.dependencyLevel = level
 
-        exthandlers_handler.ext_handlers.extHandlers = []
-        for handler in all_handlers:
-            exthandlers_handler.ext_handlers.extHandlers.append(handler)
+        exthandlers_handler.protocol.client.get_extensions_goal_state()._extensions *= 0
```
larohra, line 2511: Out of curiosity, what does this do - `*= 0`? Does this initialize a new array with the value 0? Does simply using ` = []` not work?

larohra (approved):
Thank you for refactoring the code, it definitely makes the code look simpler and cleaner. That being said, I do feel we also need to change all the variable names that were using the changed attributes of the code (for eg `extension` instead of `ext_handler` and so on) to make it more complete. Without it, the code would look very confusing for someone who's not familiar with it. I totally understand that it can be a tedious task so I wont ask you to do it all on your own. We can work out a way to refactor those bit by bit on subsequent PRs, I can help out on those. Or if you want I can modify your PR with those refactoring too if that sounds like a better approach. Either way that's my only comment. Else everything LGTM.

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
narrieta: The CaseFoldedDict will go away eventually. but in any case I feel the explicit check for None is better for those optional properties. Creating custom dictionaries with default values for the different pieces of the goal state would obscure the code, I think.

narrieta:
> Thank you for refactoring the code, it definitely makes the code look simpler and cleaner. That being said, I do feel we also need to change all the variable names that were using the changed attributes of the code (for eg `extension` instead of `ext_handler` and so on) to make it more complete. Without it, the code would look very confusing for someone who's not familiar with it. I totally understand that it can be a tedious task so I wont ask you to do it all on your own. We can work out a way to refactor those bit by bit on subsequent PRs, I can help out on those. Or if you want I can modify your PR with those refactoring too if that sounds like a better approach. Either way that's my only comment. Else everything LGTM.

@larohra - Yes, additional rename are needed. Each rename potentially touches a lot of code and needs to be tested individually (and carefully). I may do a second pass of renames once I code complete the Fast Track feature. I'll focus on those sections that are the most confusing.

azurelinuxagent/common/protocol/extensions_goal_state.py
narrieta: In the new terminology an extension is an Extension and the settings are ExtensionSettings. Actually, I wanted to use ExtensionGoalState for the former, but today we use ExtHandler both for the extension goal state and, for example, extensions installed on disk, so that requires more refactoring.

As far as the additinal renaming, I agree, see my overall comment on renames.

azurelinuxagent/common/protocol/extensions_goal_state_from_extensions_config.py
```diff
@@ -167,13 +166,13 @@ def agent_manifests(self):
...
-    def ext_handlers(self):
-        return self._ext_handlers
+    def extensions(self):
+        return self._extensions
 
     def get_redacted_text(self):
         text = self._text
-        for ext_handler in self._ext_handlers.extHandlers:
-            for extension in ext_handler.properties.extensions:
+        for ext_handler in self._extensions:
```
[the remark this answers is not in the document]
narrieta: I posted an overall comment wrt additional renames.

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
narrieta: I posted an overall comment wrt additional renames.

tests/protocol/test_extensions_goal_state.py
narrieta: should actually be a deep copy, thanks for pointing it out. right now this works because this is the last test in this method.

i am already working on the next change to parse the extension settings so I'll leave this PR as it is and do the change as part of the code I am working on.

tests/ga/test_extension.py
narrieta: clears the array (python2 does not have a list.clear() method).

i want to clear the same instance, rather than replacing it with a new empty instance

azurelinuxagent/ga/exthandlers.py
```diff
@@ -297,14 +298,15 @@ def run(self):
         etag, activity_id, correlation_id, gs_creation_time = None, None, None, None
 
         try:
+            extensions_goal_state = self.protocol.get_extensions_goal_state()
+
             # self.ext_handlers needs to be initialized first, since status reporting depends on it
-            self.ext_handlers, etag = self.protocol.get_ext_handlers()
+            self.ext_handlers = extensions_goal_state.extensions
```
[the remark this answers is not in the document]
narrieta: I posted an overall comment wrt additional renames.

nagworld9 (approved):

PR #2432, opened 2021-12-03
Parse and compare extension settings

narrieta:


azurelinuxagent/common/protocol/extensions_goal_state.py
```diff
@@ -97,29 +98,68 @@ def compare(from_extensions_config, from_vm_settings):
...
-        compare_attribute("created_on_timestamp")
-        # The status blob was added after version 112
-        if from_vm_settings.host_ga_plugin_version > FlexibleVersion("1.0.8.112"):
-            compare_attribute("status_upload_blob")
-            compare_attribute("status_upload_blob_type")
-        compare_attribute("required_features")
-        compare_attribute("on_hold")
-        compare_attribute("agent_manifests")
-        compare_attribute("extensions")
+        context = []  # used to keep track of the attribute that is being compared
```
narrieta, line 101: Initially I implemented this method to try to catch errors in the HostGAPlugin and it was pretty simple, since I do not expect the HostGAPlugin to return incorrect values deep in the goal state's object model.

However, while parsing the extension settings from vmSettings, I realized this method is extremely useful to compare against the current parsing of extensionsConfig,

This new implementation compares every single attribute of the goal state's object model and adds extra context information to debug differences in the parsing.

(while writing this code I found some subtle details wrt representing numeric attributes a strings or ints, empty collections or None, etc)

azurelinuxagent/common/protocol/extensions_goal_state_from_extensions_config.py
```diff
@@ -521,7 +521,7 @@ def __parse_and_add_extension_settings(settings_node, name, ext_handler, depends
             # There is no "extension name" for single Handler Settings. Use HandlerName for those
             ext.name = name
             ext.state = state
-            ext.sequenceNumber = seq_no
+            ext.sequenceNumber = int(seq_no)
```
narrieta, line 524: the current code represents the sequence number as a string. I decided to change that to an int, which seems more natural. Unit tests and End-to-End automation reported no errors, so this change seems safe. If you have concerns about it, please post them here and I can make the new code parse this as a string instead (if it makes sense)

azurelinuxagent/common/protocol/restapi.py
```diff
@@ -71,12 +71,6 @@ def __init__(self, family, version=None):
         self.version = "0.0.0.0" if version is None else version
         self.uris = []
 
-    def __eq__(self, other):
```
narrieta, line 74: This was moved to ExtensionsGoalState.compare, which allows me to add extra debug information to make it easier to follow up any differences in the goal states.

azurelinuxagent/common/protocol/restapi.py
```diff
@@ -177,17 +177,6 @@ def dependency_level_sort_key(self):
             level = (0 - level) - 1
         return level
 
-    def __eq__(self, other):
```
narrieta, line 180: This was moved to ExtensionsGoalState.compare, which allows me to add extra debug information to make it easier to follow up any differences in the goal states.

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
```diff
@@ -325,7 +335,52 @@ def _parse_extensions(self, vm_settings):
...
+
+                #
+                # Dependency level
+                #
+                depends_on = extension_gs.get("dependsOn")
+                if depends_on is not None:
+                    if not isinstance(depends_on, list):
+                        raise Exception('dependsOn should be an array')
+                    if extension.supports_multi_config:
+                        # TODO: Parse dependencies for multi-config extensions
```
narrieta, line 378: I will implement this on a separate PR. I could implement this code right now using specs and the reference implementation of the parsing of extensionsConfig, but the latter is not tested end-to-end.

I want to write an end-to-end test first, and then implement this code based on actual data harvested from that end-to-end test.

tests/protocol/test_extensions_goal_state.py
```diff
@@ -5,45 +5,65 @@
...
-
-            agent_manifests = copy.deepcopy(from_vm_settings.agent_manifests)
-            agent_manifests[0].family = 'MOCK_FAMILY'
-            test_property("agent_manifests", agent_manifests)
-
-            extensions = copy.copy(from_vm_settings.extensions)
-            extensions[0].name = 'MOCK_NAME'
-            test_property("extensions", extensions)
+            def assert_compare_raises(setup_copy, failing_attribute):
+                from_vm_settings_copy = copy.deepcopy(from_vm_settings)
```
narrieta, line 30: @larohra : this now implements the deepcopy you pointed out in #2430.

Again, thanks for pointing that out!

tests/protocol/test_wire.py
```diff
@@ -1180,42 +1180,38 @@ def test_update_goal_state_should_not_persist_the_protected_settings(self):
                     "Could not find the expected number of redacted settings. Expected {0}.\n{1}".format(len(protected_settings), extensions_config))
 
     def test_update_goal_state_should_save_goal_state(self):
-        # Enable FastTrack to also save the vmSettings
-        with patch("azurelinuxagent.common.protocol.wire.conf.get_enable_fast_track", return_value=True):
```
narrieta, line 1184: Fast Track is now enabled by default, so this mock is no longer needed

azurelinuxagent/common/protocol/extensions_goal_state.py
```diff
@@ -97,29 +98,68 @@ def compare(from_extensions_config, from_vm_settings):
...
+
+                if first_value != second_value:
+                    raise Exception("[{0}] != [{1}] (Attribute: {2})".format(first_value, second_value, ".".join(context)))
+            finally:
+                context.pop()
+
+        try:
+            compare_goal_states(from_extensions_config, from_vm_settings)
+        except Exception as exception:
+            raise GoalStateMismatchError("Mismatch in Goal States [Incarnation {0}] != [Etag: {1}]: {2}".format(from_extensions_config.id, from_vm_settings.id, ustr(exception)))
```
larohra, line 162: So this would still raise as soon as it detects the first mismatch right? I can't understand the use of `context` in the new changes, could you add a comment or a sample or something of how the debug messages would look incase of a mismatch?

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
```diff
@@ -258,7 +258,17 @@ def _parse_extensions(self, vm_settings):
         #                     "protectedSettings": "MIIBsAYJKoZIhvcNAQcDoIIBoTCCAZ0CAQAxggFpMIIBZQIBADBNMDkxNzA1BgoJkiaJk/IsZAEZFidXaW5kb3dzIEF6dXJlIENSUCBDZXJ0aWZpY2F0ZSBHZW5lcmF0b3ICEFpB/HKM/7evRk+DBz754wUwDQYJKoZIhvcNAQEBBQAEggEADPJwniDeIUXzxNrZCloitFdscQ59Bz1dj9DLBREAiM8jmxM0LLicTJDUv272Qm/4ZQgdqpFYBFjGab/9MX+Ih2x47FkVY1woBkckMaC/QOFv84gbboeQCmJYZC/rZJdh8rCMS+CEPq3uH1PVrvtSdZ9uxnaJ+E4exTPPviIiLIPtqWafNlzdbBt8HZjYaVw+SSe+CGzD2pAQeNttq3Rt/6NjCzrjG8ufKwvRoqnrInMs4x6nnN5/xvobKIBSv4/726usfk8Ug+9Q6Benvfpmre2+1M5PnGTfq78cO3o6mI3cPoBUjp5M0iJjAMGeMt81tyHkimZrEZm6pLa4NQMOEjArBgkqhkiG9w0BBwEwFAYIKoZIhvcNAwcECC5nVaiJaWt+gAhgeYvxUOYHXw==",
         #                     "publicSettings": "{\"GCS_AUTO_CONFIG\":true}"
         #                 }
-        #             ]
+        #             ],
+        #             "dependsOn": [
+        #                 {
+        #                     "DependsOnExtension": [
+        #                         {
+        #                             "handler": "Microsoft.Azure.Security.Monitoring.AzureSecurityLinuxAgent"
```
larohra, line 266: Out of curiosity, is this example from an actual DCR run?

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
```diff
@@ -325,7 +335,52 @@ def _parse_extensions(self, vm_settings):
...
                     extension.manifest_uris.extend(additional_locations)
 
-                # TODO: Add settings
+                #
+                # Settings
+                #
+                settings_list = extension_gs.get('settings')
+                if settings_list is not None:
+                    if not isinstance(settings_list, list):
+                        raise Exception("'settings' should be an array (extension: {0})".format(extension.name))
```
larohra, line 344: I was wondering if we should raise these errors differently (like `VMSettingConfigError` or something) to have a logical separation of errors due to actual issues in the contract vs other unhandled exceptions. We don't need it now, but maybe later on we this separation would be helpful in some cases (like we have currently for `ExtensionConfigErrors` for detecting if last GS was unsupported or not

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
```diff
@@ -325,7 +335,52 @@ def _parse_extensions(self, vm_settings):
...
+                        raise Exception("Single-config extension includes multiple settings (extension: {0})".format(extension.name))
+
+                    for s in settings_list:
+                        settings = ExtensionSettings()
+                        public_settings = s.get('publicSettings')
+                        settings.publicSettings = {} if public_settings is None else json.loads(public_settings)
+                        # Note that protectedSettings and protectedSettingsCertThumbprint can be None (which would be serialized to the extension's settings file as null)
+                        settings.protectedSettings = s.get('protectedSettings')
+                        thumbprint = s.get('protectedSettingsCertThumbprint')
+                        if thumbprint is None and settings.protectedSettings is not None:
```
larohra, line 355: we should probably check for the opposite case too, i.e. - 

`if thumbprint is not None and settings.protectedSettings is None`
```suggestion
                        (if thumbprint is not None and settings.protectedSettings is None) or (if thumbprint is None and settings.protectedSettings is not None):
```

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
```diff
@@ -325,7 +335,52 @@ def _parse_extensions(self, vm_settings):
...
+                depends_on = extension_gs.get("dependsOn")
+                if depends_on is not None:
+                    if not isinstance(depends_on, list):
+                        raise Exception('dependsOn should be an array')
+                    if extension.supports_multi_config:
+                        # TODO: Parse dependencies for multi-config extensions
+                        pass
+                    else:
+                        if len(depends_on) != 1:
+                            raise Exception('dependsOn should be an array with exactly one item for single-config extensions')
```
larohra, line 382: I dont get this condition. There can be extensions with multiple dependencies, eg: `A -> B, C, D`. That would fail this condition check right?

azurelinuxagent/common/protocol/extensions_goal_state.py
narrieta: the unit tests have samples (they check the error message). For example, when comparing the state in an instance of extension settings, the context would be something similar to "extensions[0].settings[0].state"

yes, this raises on the first diff. this is intentional.

narrieta: (the full message would be something like "[Incarnation 1] != [Etag: 1234]: [enable] != [bad-state] (Attribute: extensions[0].settings[0].state"

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
narrieta: kind of --- copied it from the multiple-dependencies scenario, changed the name of the extension and inserted it here.

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
narrieta: All those exceptions are converted to a VmSettingsError by ExtensionsGoalStateFromVmSettings.__init__. We use these classes for bad goal states:
```

class ExtensionsGoalStateError(ExtensionError):
    """
    Error raised when the ExtensionsGoalState is malformed
    """


class ExtensionsConfigError(ExtensionsGoalStateError):
    """
    Error raised when the ExtensionsConfig is malformed
    """


class VmSettingsError(ExtensionsGoalStateError):
    """
    Error raised when the VmSettings are malformed
    """

```

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
narrieta: protected settings without a cert would definitively be an error since there would be no way to decrypt the settings

not sure if CRP would ever send a thumbprint without protected settings, so i decided to be cautious and did not add that check

narrieta: (btw the current parsing of extensions config does not do any checks and simply sets the properties to None)

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
larohra: I saw that, let me rephrase, I meant separating out the errors where we were expecting something but it turned out to be something different vs something else that went wrong. Though I don't think there would be a big bucket of the latter but still I was wondering if having something like that would be useful.

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
narrieta: let me double check that on an actual run... i thought multiple dependencies would be expressed in the DependsOnExtension array

```
            "dependsOn": [
                {
                    "DependsOnExtension": [
                        {
                            "handler": "Microsoft.Azure.Security.Monitoring.AzureSecurityLinuxAgent"
                        }
                    ],
                    "dependencyLevel": 1
                }
            ]

```

azurelinuxagent/common/protocol/extensions_goal_state.py
larohra: > the unit tests have samples (they check the error message). For example, when comparing the state in an instance of extension settings, the context would be something similar to "extensions[0].settings[0].state"

Ahh right, thanks for pointing it out!

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
larohra: > not sure if CRP would ever send a thumbprint without protected settings, so i decided to be cautious and did not add that check

Fair enough, nothing would go wrong either I suppose if we have a cert but no one asking for it haha

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
narrieta: Not sure if that would be useful, I guess we could do it. __init__ also outputs the call stack so that would help debug issues other than parsing an incorrect goal state. Do you feel we need more than that?

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
narrieta: yes, that seems to be the case (they are in DependsOnExtension)

            "dependsOn": [
                {
                    "DependsOnExtension": [
                        {
                            "handler": "Microsoft.CPlat.Core.RunCommandLinux"
                        },
                        {
                            "handler": "Microsoft.Azure.Extensions.Edp.GATestExtGo"
                        }
                    ],
                    "dependencyLevel": 2
                }
            ]

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
larohra: I think its fine for now, just wanted to brainstorm I guess a thought that came to mind

kevinclark19a (approved):

larohra (approved):

PR #2436, opened 2021-12-13
Changed default Logs.Collect value from False/n to True/y

kevinclark19a:
## Description

Logs.Collect controls the task that runs the log collector in the background. This PR enables this behavior by default and updates the readme to reflect the new defaults.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,7 +195,7 @@ ResourceDisk.EnableSwap=n
 ResourceDisk.EnableSwapEncryption=n
 ResourceDisk.SwapSizeMB=0
 Logs.Verbose=n
-Logs.Collect=n
+Logs.Collect=y
 Logs.CollectPeriod=3600
 OS.AllowHTTP=n
 OS.RootDeviceScsiTimeout=300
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -461,7 +461,7 @@ leverages the system logrotate functionality to rotate logs.
 #### __Logs.Collect__
 
 _Type: Boolean_  
-_Default: n_
+_Default: y_
 
 If set, agent logs will be periodically collected and uploaded to a secure location for improved supportability.
```
nagworld9, line 466: If I remembered correctly log collector enabled only when other two conditions met along with this. Is it worthwhile to mention here otherwise it confuse the customers, right?

narrieta: +1

kevinclark19a: I think this is a valid point; I'll submit a rewrite, let me know what you think.

narrieta (approved):

README.md
nagworld9: Looks good.

nagworld9 (approved):

PR #2437, opened 2021-12-14
Add support for writing updateStatus to the status blob if requested version in GS

larohra:
## Description
This PR adds support for writing updateStatus to the status blob if requested version in GS. If not, we don't report anything for the updateStatus.

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/common/protocol/restapi.py
+++ b/azurelinuxagent/common/protocol/restapi.py
@@ -72,15 +72,15 @@ class VMAgentManifest(object):
     def __init__(self, family, version=None):
         self.family = family
         # This is the Requested version as specified by the Goal State, it defaults to 0.0.0.0 if not specified in GS
-        self.version = VERSION_0 if version is None else version
+        self.requested_version_string = VERSION_0 if version is None else version
         self.uris = []
 
     @property
     def requested_version(self):
-        return FlexibleVersion(self.version)
+        return FlexibleVersion(self.requested_version_string)
 
     @property
-    def is_requested_version_available(self):
+    def is_requested_version_specified(self):
         """
         If we don't get any requested_version from the GS, we default it to 0.0.0.0.
         This property identifies if a requested Version was passed in the GS or not.
--- a/azurelinuxagent/common/protocol/restapi.py
+++ b/azurelinuxagent/common/protocol/restapi.py
@@ -22,8 +22,12 @@ import time
 
 from azurelinuxagent.common.datacontract import DataContract, DataContractList
 from azurelinuxagent.common.future import ustr
+from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
 from azurelinuxagent.common.utils.textutil import getattrib
 from azurelinuxagent.common.version import DISTRO_VERSION, DISTRO_NAME, CURRENT_VERSION
+
+
+VERSION_0 = "0.0.0.0"
 
 
 class VMInfo(DataContract):
@@ -68,8 +72,20 @@ class VMAgentManifest(object):
     def __init__(self, family, version=None):
         self.family = family
         # This is the Requested version as specified by the Goal State, it defaults to 0.0.0.0 if not specified in GS
-        self.version = "0.0.0.0" if version is None else version
+        self.version = VERSION_0 if version is None else version
         self.uris = []
+
+    @property
+    def requested_version(self):
+        return FlexibleVersion(self.version)
+
+    @property
+    def is_requested_version_available(self):
+        """
+        If we don't get any requested_version from the GS, we default it to 0.0.0.0.
+        This property identifies if a requested Version was passed in the GS or not.
+        """
+        return self.requested_version > FlexibleVersion(VERSION_0)
 
     def __repr__(self):
         return self.__str__()
@@ -267,7 +283,7 @@ class ExtHandlerStatus(DataContract):
 
 
 class VMAgentStatus(DataContract):
-    def __init__(self, status=None, message=None, gs_aggregate_status=None):
+    def __init__(self, status=None, message=None, gs_aggregate_status=None, update_status=None):
         self.status = status
         self.message = message
         self.hostname = socket.gethostname()
@@ -276,11 +292,13 @@ class VMAgentStatus(DataContract):
         self.osversion = DISTRO_VERSION
         self.extensionHandlers = DataContractList(ExtHandlerStatus)
         self.vm_artifacts_aggregate_status = VMArtifactsAggregateStatus(gs_aggregate_status)
+        self.update_status = update_status
 
 
 class VMStatus(DataContract):
-    def __init__(self, status, message, gs_aggregate_status=None):
-        self.vmAgent = VMAgentStatus(status=status, message=message, gs_aggregate_status=gs_aggregate_status)
+    def __init__(self, status, message, gs_aggregate_status=None, vm_agent_update_status=None):
+        self.vmAgent = VMAgentStatus(status=status, message=message, gs_aggregate_status=gs_aggregate_status,
+                                     update_status=vm_agent_update_status)
 
 
 class GoalStateAggregateStatus(DataContract):
@@ -311,3 +329,18 @@ class RemoteAccessUser(DataContract):
 class RemoteAccessUsersList(DataContract):
     def __init__(self):
         self.users = DataContractList(RemoteAccessUser)
+
+
+class VMAgentUpdateStatuses(object):
+    Success = ustr("Success")
+    Transitioning = ustr("Transitioning")
+    Error = ustr("Error")
+    Unknown = ustr("Unknown")
+
+
+class VMAgentUpdateStatus(object):
+    def __init__(self, expected_version, status=VMAgentUpdateStatuses.Success, message="", code=0):
+        self.expected_version = expected_version
+        self.status = status
+        self.message = message
+        self.code = code
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -292,7 +292,21 @@ def ga_status_to_v1(ga_status):
         "status": ga_status.status,
         "formattedMessage": __get_formatted_msg_for_status_reporting(ga_status.message)
     }
+
+    if ga_status.update_status is not None:
+        v1_ga_status["updateStatus"] = get_ga_update_status_to_v1(ga_status.update_status)
+
     return v1_ga_status
+
+
+def get_ga_update_status_to_v1(update_status):
+    v1_ga_update_status = {
+        "expectedVersion": update_status.expected_version,
+        "status": update_status.status,
+        "code": update_status.code,
+        "formattedMessage": __get_formatted_msg_for_status_reporting(update_status.message)
+    }
+    return v1_ga_update_status
 
 
 def ext_substatus_to_v1(sub_status_list):
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -889,7 +889,7 @@ class ExtHandlersHandler(object):
 
         return handlers_to_report
 
-    def report_ext_handlers_status(self, incarnation_changed=False, vm_agent_upgrade_status=None):
+    def report_ext_handlers_status(self, incarnation_changed=False, vm_agent_update_status=None):
         """
         Go through handler_state dir, collect and report status.
         Returns the status it reported, or None if an error occurred.
@@ -897,7 +897,7 @@ class ExtHandlersHandler(object):
         try:
             vm_status = VMStatus(status="Ready", message="Guest Agent is running",
                                  gs_aggregate_status=self.__gs_aggregate_status,
-                                 vm_agent_update_status=vm_agent_upgrade_status)
+                                 vm_agent_update_status=vm_agent_update_status)
             handlers_to_report = []
 
             # In case of Unsupported error, report the status of the handlers in the VM
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -889,15 +889,15 @@ class ExtHandlersHandler(object):
 
         return handlers_to_report
 
-    def report_ext_handlers_status(self, incarnation_changed=False):
+    def report_ext_handlers_status(self, incarnation_changed=False, vm_agent_upgrade_status=None):
         """
         Go through handler_state dir, collect and report status.
         Returns the status it reported, or None if an error occurred.
         """
         try:
             vm_status = VMStatus(status="Ready", message="Guest Agent is running",
-                                 gs_aggregate_status=self.__gs_aggregate_status)
-
+                                 gs_aggregate_status=self.__gs_aggregate_status,
+                                 vm_agent_update_status=vm_agent_upgrade_status)
             handlers_to_report = []
 
             # In case of Unsupported error, report the status of the handlers in the VM
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -507,7 +507,7 @@ class UpdateHandler(object):
                 if incarnation_changed:
                     logger.info("Unable to report update status as no matching manifest found for family: {0}".format(
                         conf.get_autoupdate_gafamily()))
-                return
+                return None
 
             if manifest.is_requested_version_specified:
                 if CURRENT_VERSION == manifest.requested_version:
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -46,6 +46,7 @@ from azurelinuxagent.common.event import add_event, initialize_event_logger_vmin
 from azurelinuxagent.common.exception import ResourceGoneError, UpdateError, ExitException, AgentUpgradeExitException
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.osutil import get_osutil, systemd
+from azurelinuxagent.common.protocol.restapi import VMAgentUpdateStatus, VMAgentUpdateStatuses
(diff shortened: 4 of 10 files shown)


azurelinuxagent/common/protocol/restapi.py
```diff
@@ -68,9 +72,21 @@ class VMAgentManifest(object):
     def __init__(self, family, version=None):
         self.family = family
         # This is the Requested version as specified by the Goal State, it defaults to 0.0.0.0 if not specified in GS
```
kevinclark19a, line 74: Should this comment be moved up to where `VERSION_0` is defined?

larohra: `VERSION_0` is just a constant depicting the `0.0.0.0` value. This comment is specific to the logic specified here so I feel its in the apt place.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -889,15 +889,15 @@ def __get_handlers_on_file_system(self, incarnation_changed):
...
         """
         Go through handler_state dir, collect and report status.
         Returns the status it reported, or None if an error occurred.
         """
         try:
             vm_status = VMStatus(status="Ready", message="Guest Agent is running",
-                                 gs_aggregate_status=self.__gs_aggregate_status)
-
+                                 gs_aggregate_status=self.__gs_aggregate_status,
+                                 vm_agent_update_status=vm_agent_upgrade_status)
```
kevinclark19a, line 900: NIT: these two have slightly different names (`...agent_update...` vs `...agent_upgrade...`), might consider picking just a single one.

larohra: good catch! Let me fix it.

azurelinuxagent/common/protocol/restapi.py
```diff
@@ -68,9 +72,21 @@ class VMAgentManifest(object):
...
-        self.version = "0.0.0.0" if version is None else version
+        self.version = VERSION_0 if version is None else version
         self.uris = []
 
+    @property
+    def requested_version(self):
```
narrieta, line 79: I think having a requested_version and a version properties is confusing... maybe the latter should be called requested_version_string?

azurelinuxagent/common/protocol/restapi.py
```diff
@@ -68,9 +72,21 @@ class VMAgentManifest(object):
...
+        return FlexibleVersion(self.version)
+
+    @property
+    def is_requested_version_available(self):
```
narrieta, line 83: "available" may not be the correct term... "specified"?

azurelinuxagent/common/protocol/restapi.py
```diff
@@ -68,9 +72,21 @@ class VMAgentManifest(object):
...
+        """
+        If we don't get any requested_version from the GS, we default it to 0.0.0.0.
```
narrieta, line 85: doesn't the spec mention that the requested version is always in the GS?

azurelinuxagent/ga/update.py
```diff
@@ -487,9 +488,44 @@ def _process_goal_state(self, exthandlers_handler, remote_access_handler):
...
+        """
+        This function gets the VMAgent upgrade status as per the last GoalState.
+        Returns: None if the last GS does not ask for requested version else VMAgentUpdateStatus
+        """
+        upgrade_status = None
+
+        try:
+            agent_manifests, _ = protocol.get_vmagent_manifests()
+            # Expectation here is that there will only be one manifest per family passed down from CRP.
+            # If there are multiple, we pick the first one
```
narrieta, line 502: what would multiple statuses indicate? an error? do we need to log that?

tests/ga/test_update.py
```diff
@@ -1740,6 +1741,66 @@ def test_it_should_recreate_extension_event_directories_for_existing_extensions_
...
 
+    def test_it_should_report_update_status_in_status_blob(self):
+        with _get_update_handler(iterations=1) as (update_handler, protocol):
+            with patch.object(conf, "get_autoupdate_gafamily", return_value="Prod"):
+                with patch("azurelinuxagent.common.logger.warn") as patch_warn:
+
+                    protocol.aggregate_status = None
+                    protocol.incarnation = 1
+
+                    def mock_http_put(url, *args, **_):
```
kevinclark19a, line 1752: this feels like it could break if different types of PUT calls are added to the agent; there are only two choices below and one of them is assumed (i.e. not explicitly checked for, it's just in  the equivalent of an `else` clause.

I think this could be more or less fixed by making the second scenario explicit (via an if statement) and then do a `return None` to ignore anything else.

tests/ga/test_update.py
```diff
@@ -1740,6 +1741,66 @@ def test_it_should_recreate_extension_event_directories_for_existing_extensions_
...
+                        self.assertTrue(update_handler.exit_mock.called, "The process should have exited")
+                        exit_args, _ = update_handler.exit_mock.call_args
+                        self.assertEqual(exit_args[0], 0,
+                                         "Exit code should be 0; List of all warnings logged by the agent: {0}".format(
+                                             patch_warn.call_args_list))
+
+                    protocol.set_http_handlers(http_put_handler=mock_http_put)
+
+                    # Case 1: No requested version in GS; updateStatus should not be reported
+                    update_handler.run(debug=True)
```
kevinclark19a, line 1774: A bit confused by this.. Why not use the `update_goal_state_and_run_handler` method from above?

tests/ga/test_update.py
```diff
@@ -1740,6 +1741,66 @@ def test_it_should_recreate_extension_event_directories_for_existing_extensions_
...
+                    self.assertFalse("updateStatus" in protocol.aggregate_status['aggregateStatus']['guestAgentStatus'],
+                                     "updateStatus should not be reported if not asked in GS")
+
+                    # Case 2: Requested version in GS != Current Version; updateStatus should not be error
```
kevinclark19a, line 1778: this comment seems backwards... Case 2 is expecting an error, right?
```python
self.assertEqual(VMAgentUpdateStatuses.Error, update_status['status'], "Status should be an error")
```

azurelinuxagent/ga/update.py
```diff
@@ -487,9 +488,44 @@ def _process_goal_state(self, exthandlers_handler, remote_access_handler):
...
+                err_msg = "[This error will only be logged once per incarnation] " \
+                          "Ran into error when trying to fetch updateStatus for the agent, skipping reporting update satus. Error: {0}".format(
+                           textutil.format_exception(error))
+                logger.warn(err_msg)
+                add_event(op=WALAEventOperation.AgentUpgrade, is_success=False, message=err_msg, log_event=False)
+
+        return upgrade_status
+
     def _report_status(self, exthandlers_handler, incarnation_changed):
+        vm_agent_upgrade_status = self.__get_vmagent_upgrade_status(exthandlers_handler.protocol, incarnation_changed)
```
narrieta, line 525: should this be disabled in production until the feature is stable?

azurelinuxagent/ga/update.py
```diff
@@ -487,9 +488,44 @@ def _process_goal_state(self, exthandlers_handler, remote_access_handler):
...
+            manifest = next(m for m in agent_manifests if m.family == conf.get_autoupdate_gafamily())
+            if manifest.is_requested_version_available:
+                if CURRENT_VERSION == manifest.requested_version:
+                    status = VMAgentUpdateStatuses.Success
+                    code = 0
+                else:
+                    status = VMAgentUpdateStatuses.Error
+                    code = -1
+                upgrade_status = VMAgentUpdateStatus(expected_version=manifest.version, status=status, code=code)
+        except StopIteration:
```
narrieta, line 512: could we have this except clause enclose only the actual iteration? if at some point another iteration is added to this code this could lead to problems

azurelinuxagent/ga/update.py
```diff
@@ -487,9 +488,44 @@ def _process_goal_state(self, exthandlers_handler, remote_access_handler):
...
+            logger.verbose("No matching manifest found for family: {0}".format(conf.get_autoupdate_gafamily()))
```
narrieta, line 513: should we log this one also once per goal state?

tests/ga/test_update.py
```diff
@@ -1740,6 +1741,66 @@ def test_it_should_recreate_extension_event_directories_for_existing_extensions_
...
+                    protocol.mock_wire_data.set_extension_config("wire/ext_conf_requested_version.xml")
+                    update_goal_state_and_run_handler()
+                    self.assertTrue("updateStatus" in protocol.aggregate_status['aggregateStatus']['guestAgentStatus'],
+                                    "updateStatus should be in status blob. Warns: {0}".format(patch_warn.call_args_list))
+                    update_status = protocol.aggregate_status['aggregateStatus']['guestAgentStatus']["updateStatus"]
+                    self.assertEqual(VMAgentUpdateStatuses.Error, update_status['status'], "Status should be an error")
+                    self.assertEqual(update_status['expectedVersion'], "9.9.9.10", "incorrect version reported")
+                    self.assertEqual(update_status['code'], -1, "incorrect code reported")
+
+                    # Case 3: Requested version in GS == Current Version; updateStatus should not be Success
```
kevinclark19a, line 1788: Same as above, isn't the code below expecting success?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -292,9 +292,23 @@ def ga_status_to_v1(ga_status):
...
         "formattedMessage": __get_formatted_msg_for_status_reporting(ga_status.message)
     }
+
+    if ga_status.update_status is not None:
+        v1_ga_status["updateStatus"] = get_ga_update_status_to_v1(ga_status.update_status)
+
     return v1_ga_status
 
 
+def get_ga_update_status_to_v1(update_status):
```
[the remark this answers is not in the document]
kevinclark19a: +1

azurelinuxagent/ga/update.py
```diff
@@ -487,9 +488,44 @@ def _process_goal_state(self, exthandlers_handler, remote_access_handler):
...
+
+        try:
+            agent_manifests, _ = protocol.get_vmagent_manifests()
+            # Expectation here is that there will only be one manifest per family passed down from CRP.
+            # If there are multiple, we pick the first one
+            manifest = next(m for m in agent_manifests if m.family == conf.get_autoupdate_gafamily())
+            if manifest.is_requested_version_available:
+                if CURRENT_VERSION == manifest.requested_version:
+                    status = VMAgentUpdateStatuses.Success
+                    code = 0
```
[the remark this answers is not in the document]
larohra: Its used for logging in CRP. Windows agent primarily sends different HR Codes for different cases which gets logged in CRP. Here's what Aymen said about it - 

> Usually it would be an error code(mainly for windows)
Its only for logging
CRP will only cares about one thing : "status"

azurelinuxagent/ga/update.py
larohra: Done

azurelinuxagent/ga/update.py
larohra: Done

azurelinuxagent/ga/update.py
larohra: This would only be reported if CRP includes a requested version in GS. This does no harm to the agent or CRP since its just adding to the status blob. Even if by a bug CRP sends us a requested version in GS before the feature is stable and we end up adding the `updateStatus` in the status blob, it wont do us any harm. It'll just be ignored.

Did you have any other use case in mind?

tests/ga/test_update.py
larohra: That's a good point, we should indeed make it explicit. That being said, unfortunately, as per the current code, its very hard to differentiate if a PUT call is for status or not because we directly upload to the storage blob uri we get from GS - 

> if self.status_blob.upload(extensions_goal_state.status_upload_blob):

The only other differentiating factor we have is a PUT call. One way to deal with it would be to change all of our test files and add a unique string in the `<StatusUploadBlob>` object to specify that its a status blob and then filter it out based on those, but for the time being, since we dont have any other PUT calls, I'd rather keep it as is to avoid the extra trouble :)

tests/ga/test_update.py
larohra: No particular reason honestly, for the subsequent calls we needed these functions - 

> protocol.mock_wire_data.set_incarnation(protocol.incarnation)
                        self._add_write_permission_to_goal_state_files()
                        update_handler.set_iterations(1)

But for the first one it wasn't needed so skipped it. But since you pointed it out, it is beneficial to call the `update_goal_state_and_run_handler` here too since it has an extra assert so added it here

tests/ga/test_update.py
larohra: Good catch! I think I got confused between the wording of case 2 and 3. Fixed it thanks!

tests/ga/test_update.py
larohra: Fixed

tests/ga/test_update.py
kevinclark19a: Could you just create a new GS file to use for this test and edit it such that the storage blob uri matches some regex (or just straight up is some string) and filter off that?

That would still be a bit fragile (i.e. someone changing the specific file you used might break things) but I feel like it's a good compromise between leaving it as is and doing the extra work of manipulating the status blob string itself.

kevinclark19a (approved):

tests/ga/test_update.py
larohra: After thinking about this again, since its being used in multiple places, I'd rather fix this in a separate PR in all those usages.

azurelinuxagent/common/protocol/wire.py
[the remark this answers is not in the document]
narrieta: thanks; i guess we can leave it as is

azurelinuxagent/ga/update.py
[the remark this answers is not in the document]
narrieta: if crp does not care about it, why populate it at all? or we need some documentation about what 0 and -1 mean

azurelinuxagent/ga/update.py
narrieta: maybe change the comment to say that during validation we already checked there is only 1 item so we just pick up the first one?

azurelinuxagent/ga/update.py
narrieta: where do we ensure that we report this only if the GS includes a requested version?

azurelinuxagent/ga/update.py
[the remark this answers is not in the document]
larohra: CRP logs it. I think the intention is to have a clear message on what went wrong in the CRP logs itself. 

>  why populate it at all?

It's primarily needed for Windows where they're going to return different HR codes based on the scenario.   

> or we need some documentation about what 0 and -1 mean

Since we dont have a specific need for it, I just stuck to the basic exit code usages that we generally use in the code (0 for success and -1 for failure). Let me know if you feel the need to have a comment about it and I can add it.

azurelinuxagent/ga/update.py
larohra: The check for this is in Line 512 of update.py - 

```python
if manifest.is_requested_version_specified:
                if CURRENT_VERSION == manifest.requested_version:
                    status = VMAgentUpdateStatuses.Success
                    code = 0
                else:
                    status = VMAgentUpdateStatuses.Error
                    code = -1
                update_status = VMAgentUpdateStatus(expected_version=manifest.requested_version_string, status=status,
                                                    code=code)
```

In case the `manifest.is_requested_version_specified` condition is not satisfied, we return None. 

When we're preparing the status blob, we check if the `vm_status.vmAgent.update_status` is not None and if so, include the status in the blob else skip it (line 296 in wire.py) - 

```py
if ga_status.update_status is not None:
        v1_ga_status["updateStatus"] = get_ga_update_status_to_v1(ga_status.update_status)
```

Hope this helps!

azurelinuxagent/ga/update.py
larohra: sure let me rephrase

larohra: Done

azurelinuxagent/ga/update.py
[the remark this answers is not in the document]
narrieta: in what instances do we use -1 for failure?

azurelinuxagent/ga/update.py
narrieta: I still feel this should be completely disabled in production until the feature is stable. If at some point there is some issue in the feature and we have some agents pushing this status, who knows, that could be an issue.

azurelinuxagent/ga/update.py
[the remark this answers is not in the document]
larohra: You're right, the normal exit codes are positive non-0s. I'll change it to `1`

azurelinuxagent/ga/update.py
larohra: I don't see any big downside to it honestly. It's being controlled by CRP (the availability of `requested_version`) and we would upload status only if asked. On top of it, this status is only meant for RSM so it should be pretty invisible to the customers. 

On the flipside though, if I think about it, one case that comes to mind on when it can break is if CRP messes up and kind of couples the updateStatus with the extension status and fails extensions due to a failed upgrade. I guess we can go with a flagged approach and enable it by default when we have more confidence in the feature e2e. I'll also open a task to add a DCR scenario for it with the flag enabled and another for enabling it globally in the future once its ready.

azurelinuxagent/ga/update.py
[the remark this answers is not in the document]
narrieta: thanks

narrieta (approved):

kevinclark19a (approved):

PR #2438, opened 2021-12-15
Update README.md

mabicca:
Adding note about installing from source not being supported.

## Description

Issue # 

---

### PR information

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,9 @@ You can view more installation options by running:
 
 The agent's log file is kept at `/var/log/waagent.log`.
 
+## NOTE
+Installing the waagent from source **is not supported**
+
 ## Upgrade
 
 Upgrading via your distribution's package repository is strongly preferred.


README.md
```diff
@@ -106,6 +106,9 @@ You can view more installation options by running:
 
 The agent's log file is kept at `/var/log/waagent.log`.
 
+## NOTE
+Installing the waagent from source **is not supported**
+
```
narrieta, line 111: I would propose something similar to "Installing from source is meant to be used in advanced scenarios, such as creating a new VM image, and support is very limited (open an issue here and we may be able to provide guidance). Do not install from source while troubleshooting issues with the Agent."

mabicca: What about we revamp that section to specifically talk about installing from source when creating a new, customized image and explaining that installing from source is not supported on Azure Endorsed Distributions? At least that would be clear in terms of the support in those scenarios where we should leverage the distribution package management.

narrieta: ok, can you post how it would read?

mabicca: Sure, could we add a note saying:

Installing the waagent from the source is to be used only for new virtual image creation (custom images) and the [support](https://docs.microsoft.com/en-us/troubleshoot/azure/cloud-services/support-linux-open-source-technology) is very limited for those scenarios. 

For [Endorsed Linux distributions on Azure](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/endorsed-distros) images, the only supported method of installation is using the distribution's own packaging mechanisms such as apt, yum or zypper.

mabicca: @narrieta let me know if this is good to update.

mabicca: @narrieta I know it's been a while, but not sure if we were able to update the README about this?

narrieta: @mabicca sorry for the delay. Submitted #2941

narrieta:
Replaced by #2941

PR #2442, opened 2021-12-17
Rename ext_handler to extension

narrieta:
Updated extensions_goal_state_from_extensions_config.py to use the term "Extension" instead of "Extension Handler"

### DIFF
--- a/azurelinuxagent/common/protocol/extensions_goal_state_from_extensions_config.py
+++ b/azurelinuxagent/common/protocol/extensions_goal_state_from_extensions_config.py
@@ -229,17 +229,17 @@ class ExtensionsGoalStateFromExtensionsConfig(ExtensionsGoalState):
         plugin_settings = findall(plugin_settings_list, "Plugin")
 
         for plugin in plugins:
-            ext_handler = Extension()
+            extension = Extension()
             try:
-                ExtensionsGoalStateFromExtensionsConfig._parse_plugin(ext_handler, plugin)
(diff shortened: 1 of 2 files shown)


nagworld9:
@narrieta Did we do same rename in VMSettings or we already calling as `Extension`

azurelinuxagent/common/protocol/extensions_goal_state_from_extensions_config.py
```diff
@@ -527,7 +527,7 @@ def __parse_and_add_extension_settings(settings_node, name, ext_handler, depends
             ext.dependencyLevel = depends_on_level
             thumbprint = handler_settings.get("protectedSettingsCertThumbprint")
             ext.certificateThumbprint = thumbprint
-            ext_handler.settings.append(ext)
+            extension.settings.append(ext)
```
nagworld9, line 530: looks like we are using `ext` for settings, can we rename that too (If it's small effort) to avoid confusion.

narrieta: fixed

larohra (approved):

azurelinuxagent/common/protocol/extensions_goal_state_from_extensions_config.py
nagworld9: Thanks for the change.

nagworld9 (approved):

PR #2449, opened 2021-12-21
Blacklisting only on Agent Updates

larohra:
## Description
Currently we blacklist the agents if the ext-handler process exits with any error (this is the primary stakeholder in blacklisting logic, other scenarios are very rare). This PR changes the model to blacklist only on auto-update. 

This is a pre-req to GAVersioning changes as that depends heavily on having a stable blacklisting logic.

Issue # 

---

### PR information


azurelinuxagent/ga/update.py
```diff
@@ -138,8 +139,13 @@ def get_update_handler():
     return UpdateHandler()
 
 
+def get_agent_global_update_signal_file():
+    return os.path.join(conf.get_lib_dir(), AGENT_UPDATE_SIGNAL_FILE)
+
+
 class UpdateHandler(object):
     TELEMETRY_HEARTBEAT_PERIOD = timedelta(minutes=30)
+    CLEAN_UPDATE_SIGNAL_PERIOD = timedelta(minutes=15)
```
larohra, line 148: We've had the logic in place in the daemon code where we mark a version as successful only if it's been running correctly for over 15 mins. Keeping the same time check here to mark an update successful

azurelinuxagent/ga/update.py
```diff
@@ -184,6 +191,12 @@ def __init__(self):
             else:
                 self._goal_state_period = conf.get_goal_state_period()
 
+    @property
+    def process_agent_update_signal_file(self):
```
larohra, line 195: This check is mainly for those cases where we're unable to delete the signal file due to OS issues. In case there are any such cases, we won't blacklist any agents at all

azurelinuxagent/ga/update.py
```diff
@@ -1170,23 +1203,86 @@ def __upgrade_agent_if_permitted(self):
 
         if (upgrade_type == AgentUpgradeType.Hotfix and next_hotfix_time <= now) or (
                 upgrade_type == AgentUpgradeType.Normal and next_normal_time <= now):
+            self._set_agent_update_signal_file()
```
larohra, line 1206: Set the signal file during agent update

azurelinuxagent/ga/update.py
```diff
@@ -1170,23 +1203,86 @@ def __upgrade_agent_if_permitted(self):
...
+            except Exception as err_:
+                # Make best case effort to delete it, if can't then ignore
+                logger.warn("[Retry {0}/5] Unable to delete signal file {1}: {2}".format(
+                    self._try_count_for_removing_update_signal_file + 1, signal_file, err_))
+                self._try_count_for_removing_update_signal_file += 1
+
+        # Since the Agent Update Signal file exists, we still haven't verified the stability of the current version.
+        update_time = None
+        max_try = 3
+        for retry in range(1, max_try+1):
```
narrieta, line 1253: what kind of errors is the retry logic trying to address? I don't quite get the new for retries

azurelinuxagent/ga/update.py
narrieta, line 148: is the name CLEAN_UPDATE_SIGNAL_PERIOD accurate? i think we are cleaning up in the main loop

azurelinuxagent/ga/update.py
```diff
@@ -70,6 +70,7 @@
 AGENT_ERROR_FILE = "error.json"  # File name for agent error record
 AGENT_MANIFEST_FILE = "HandlerManifest.json"
 AGENT_PARTITION_FILE = "partition"
+AGENT_UPDATE_SIGNAL_FILE = "AgentUpgradeSignal"
```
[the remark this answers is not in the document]
larohra: Agreed on the renaming in general, though some I think some things are good as is. For example:   `AgentUpdating.json` doesn't make sense since this file is not a json file, its just a simple file which contains a timestamp in it so I'll keep it as  `AgentUpgradeSignal`

azurelinuxagent/ga/update.py
larohra: yup we are. This variable is just to signify the period of time we'd wait before cleaning up the signal file. Does this rename sound better `REMOVE_UPDATE_SIGNAL_PERIOD`?

azurelinuxagent/ga/update.py
larohra: Any os level issues where we're unable to read from the file. It won't throw `FileNotFound` error because we've already verified it but there's still chances for `OSError` or even `UnicodeError` (the `fileutil.read_file` module also decodes the string before sending it back). 

This is a basic catch all to try to read the file. I was considering reading the creation time but apparently there's no supported in-box way to do it - https://stackoverflow.com/a/1408281. There's some external modules (like [crtime](https://pypi.org/project/crtime/) but they require manual installation due to which I decided against it

azurelinuxagent/ga/update.py
```diff
@@ -944,15 +968,24 @@ def _check_and_download_agent_if_upgrade_available(self, protocol, base_version=
...
+                [GuestAgent(pkg=pkg, host=host, process_agent_update_signal_file=self.process_agent_update_signal_file)
+                 for pkg in pkg_list.versions])
+
+            # When we're checking for new versions for auto-update, the agent blacklists the errors if anything goes wrong.
+            # We should remove the agents which only have the errors blacklisted from directory to keep the state clean.
+            # Keeping this in mind, we should filter out the agents that are not blacklisted but have the errors blacklisted.
 
+            # This line keeps only the agents that we should keep on the file.
+            # The self._purge_agents() function will purge out the agents we need to delete as per the comments above
+            self.agents = [agent for agent in self.agents if agent.is_blacklisted or not agent.is_error_blacklisted]
```
nagworld9, line 981: I don't understand this check. I assume this is to filter non blacklisted agents. `agent.is_blacklisted` is combination of `is_error_blacklisted` plus exist signal file. is ` not agent.is_blacklisted` check not enough here? which means either agent.is_error_blacklisted is not true or signal file not present. Don't we call a valid agent if this is the case?

PR #2451, opened 2021-12-28
Use vmSettings for extension processing

narrieta:
Now we use vmSettings for extension processing. We still fallback to extensionsConfig if the HostGAPlugin does not support Fast Track, if there are errors fetching the vmSettings, or if they are different to extensionsConfig.

### DIFF
--- a/azurelinuxagent/common/protocol/extensions_goal_state.py
+++ b/azurelinuxagent/common/protocol/extensions_goal_state.py
@@ -22,7 +22,6 @@ from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.AgentGlobals import AgentGlobals
 from azurelinuxagent.common.exception import AgentError
 from azurelinuxagent.common.utils import textutil
-from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
 
 
 class GoalStateMismatchError(AgentError):
@@ -104,10 +103,8 @@ class ExtensionsGoalState(object):
             compare_attributes(first, second, "activity_id")
             compare_attributes(first, second, "correlation_id")
             compare_attributes(first, second, "created_on_timestamp")
-            # The status blob was added after version 112
-            if from_vm_settings.host_ga_plugin_version > FlexibleVersion("1.0.8.112"):
-                compare_attributes(first, second, "status_upload_blob")
-                compare_attributes(first, second, "status_upload_blob_type")
+            compare_attributes(first, second, "status_upload_blob")
+            compare_attributes(first, second, "status_upload_blob_type")
             compare_attributes(first, second, "required_features")
             compare_attributes(first, second, "on_hold")
             compare_array(first.agent_manifests, second.agent_manifests, compare_agent_manifests, "agent_manifests")
--- a/azurelinuxagent/common/protocol/extensions_goal_state_from_extensions_config.py
+++ b/azurelinuxagent/common/protocol/extensions_goal_state_from_extensions_config.py
@@ -33,6 +33,7 @@ class ExtensionsGoalStateFromExtensionsConfig(ExtensionsGoalState):
     def __init__(self, incarnation, xml_text, wire_client):
         super(ExtensionsGoalStateFromExtensionsConfig, self).__init__()
         self._id = incarnation
+        self._incarnation = incarnation
         self._text = xml_text
         self._status_upload_blob = None
         self._status_upload_blob_type = None
@@ -130,6 +131,10 @@ class ExtensionsGoalStateFromExtensionsConfig(ExtensionsGoalState):
     @property
     def id(self):
         return self._id
+
+    @property
+    def incarnation(self):
+        return self._incarnation
 
     @property
     def activity_id(self):
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -969,14 +969,9 @@ class WireClient(object):
 
     def get_extensions_goal_state(self):
         if self._extensions_goal_state is None:
-            raise ProtocolError("Trying to fetch ExtensioalState before initialization!")
+            raise ProtocolError("Trying to fetch ExtensionsGoalState before initialization!")
 
-        try:
-            ExtensionsGoalState.compare(self._goal_state.extensions_config,  self._extensions_goal_state)
-        except Exception as e:
-            logger.info("{0}", textutil.format_exception(e))
-
-        return self._extensions_goal_state  # self._goal_state.extensions_config  # self._extensions_goal_state
+        return self._extensions_goal_state
 
     def get_ext_manifest(self, ext_handler):
         if self._goal_state is None:
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -804,7 +804,10 @@ class WireClient(object):
                 goal_state_updated = True
 
             vm_settings_goal_state_updated = False
-            if conf.get_enable_fast_track():
+            if not conf.get_enable_fast_track():
+                # if Fast Track is not enabled then use extensionsConfig
+                self._extensions_goal_state = self._goal_state.extensions_config
+            else:
                 try:
                     vm_settings_goal_state = self._fetch_vm_settings_goal_state(force_update=force_update)
                     if vm_settings_goal_state is None:  # if vmSettings are None, Fast Track is not supported; use extensionsConfig
@@ -877,7 +880,6 @@ class WireClient(object):
             else:  # since the vmSettings were updated, the response must include an etag
                 raise ProtocolError("The vmSettings do no include an Etag [correlation ID: {0}]".format(correlation_id))
 
-            logger.info("Fetched new vmSettings [correlation ID: {0} New eTag: {1}]", correlation_id, response_etag)
             response_content = self.decode_config(response.read())
             vm_settings = ExtensionsGoalStateFactory.create_from_vm_settings(response_etag, response_content)
 
@@ -891,6 +893,8 @@ class WireClient(object):
             # Don't support HostGAPlugin versions older than 115
             if vm_settings.host_ga_plugin_version < FlexibleVersion("1.0.8.115"):
                 return None
+
+            logger.info("Fetched new vmSettings [correlation ID: {0} New eTag: {1}]", correlation_id, vm_settings.etag)
 
             return vm_settings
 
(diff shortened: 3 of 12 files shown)


azurelinuxagent/common/protocol/extensions_goal_state.py
```diff
@@ -104,10 +103,8 @@ def compare_goal_states(first, second):
             compare_attributes(first, second, "activity_id")
             compare_attributes(first, second, "correlation_id")
             compare_attributes(first, second, "created_on_timestamp")
-            # The status blob was added after version 112
```
narrieta, line 107: now there is a global check for the version so this one is no longer needed

azurelinuxagent/common/protocol/extensions_goal_state_from_extensions_config.py
```diff
@@ -131,6 +132,10 @@ def fetch_through_host():
     def id(self):
         return self._id
 
+    @property
```
narrieta, line 135: initially i created an abstraction over incarnation/etag (the "id" property), but there are some places where it is important to distinguish between them so now i am exposing them as separate properties

azurelinuxagent/ga/exthandlers.py
```diff
@@ -300,13 +300,13 @@ def run(self):
         try:
             extensions_goal_state = self.protocol.get_extensions_goal_state()
 
-            # self.ext_handlers needs to be initialized first, since status reporting depends on it
+            # self.ext_handlers and etag need to be initialized first, since status reporting depends on them
             self.ext_handlers = extensions_goal_state.extensions
+            etag = self.protocol.client.get_goal_state().incarnation
```
narrieta, line 305: despite the name of this local, this needs to be the incarnation number for status reporting to work - I introduced a regression a couple of PRs ago and was using the 'id' property; now i am fixing this to use the incarnation explicitly

there are a lot renames/refactoring needed in this module, i am planning on doing those gradually

azurelinuxagent/common/protocol/extensions_goal_state_from_extensions_config.py
larohra: in this case, shouldn't you rename the `id` property as `etag` to make it clearer?

larohra: ohh nvm just saw that both still have the `id` property.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -781,76 +782,67 @@ def update_goal_state(self, force_update=False):
...
-                    self._extensions_goal_state = ExtensionsGoalStateFactory.create_empty()
-                else:
-                    xml_text = self.fetch_config(self._goal_state.extensions_config_uri, self.get_header())
-                    self._extensions_goal_state = ExtensionsGoalStateFactory.create_from_extensions_config(goal_state.incarnation, xml_text, self)
-                updated = True
-
-            if conf.get_enable_fast_track():
-                # Since currently we query vmSettings only to exercise the HostGAPlugin, errors in this section of the code should not
-                # prevent the agent from processing the goal state; we simply report a summary of those errors.
+                goal_state_updated = True
```
larohra, line 804: maybe rename this to `fabric_goal_state_updated` for more clarity?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -893,9 +880,23 @@ def get_vm_settings():
...
             response_content = self.decode_config(response.read())
-            return response_etag, response_content
+            vm_settings = ExtensionsGoalStateFactory.create_from_vm_settings(response_etag, response_content)
+
+            # log the HostGAPlugin version when we fetch the vmSettings for the first time
+            is_first_vm_settings = self._vm_settings_goal_state is None
+            if is_first_vm_settings:
+                message = "HostGAPlugin version: {0}".format(vm_settings.host_ga_plugin_version)
+                logger.info(message)
+                add_event(op=WALAEventOperation.HostPlugin, message=message, is_success=True)
```
larohra, line 891: If the HGAP version < .115, we will keep emitting this event on every call to `update_goal_state`. We should only log this once irrespective

azurelinuxagent/common/protocol/wire.py
```diff
@@ -969,7 +970,13 @@ def get_certs(self):
...
         if self._extensions_goal_state is None:
             raise ProtocolError("Trying to fetch ExtensioalState before initialization!")
-        return self._extensions_goal_state
+
+        try:
+            ExtensionsGoalState.compare(self._goal_state.extensions_config,  self._extensions_goal_state)
+        except Exception as e:
+            logger.info("{0}", textutil.format_exception(e))
+
+        return self._extensions_goal_state  # self._goal_state.extensions_config  # self._extensions_goal_state
```
larohra, line 979: Forgot to remove the comments?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -969,7 +970,13 @@ def get_certs(self):
     def get_extensions_goal_state(self):
         if self._extensions_goal_state is None:
             raise ProtocolError("Trying to fetch ExtensioalState before initialization!")
-        return self._extensions_goal_state
+
+        try:
+            ExtensionsGoalState.compare(self._goal_state.extensions_config,  self._extensions_goal_state)
```
larohra, line 975: Any particular reason why comparing here? I thought we were only comparing when we get a new GS

azurelinuxagent/common/protocol/wire.py
narrieta: the 2 goal states are named _goal_state and _vm_settings_goal_state, the update variables are parallel to those names (goal_state_updated and vm_settings_goal_state updated)

renaming self._goal_state to self.fabric_goal_state would not be appropriate, since in the next iteration there will be only 1 goal state (which can come from fabric or fast track) and self._vm_settings_goal_state will go away)

azurelinuxagent/common/protocol/wire.py
narrieta: Yes, good point. I have another change in the pipeline where if we do not support vmsettings we will stop trying to use it for several hours. I'll add the check for minimum version there, too.

azurelinuxagent/common/protocol/wire.py
narrieta: oh, sorry, i forgot to remove this debugging code (there was an issue reporting status). I'll remove this and the comments below, thanks for pointing it out.

azurelinuxagent/common/protocol/wire.py
narrieta: yup, removing

azurelinuxagent/common/protocol/wire.py
```diff
@@ -781,76 +782,67 @@ def update_goal_state(self, force_update=False):
...
         try:
-            updated = False
-
             goal_state = GoalState(self)
 
             # Always update the hostgaplugin, since the agent may issue requests to it even if there are no other changes
             # in the goal state (e.g. report status)
             self._update_host_plugin(goal_state.container_id, goal_state.role_config_name)
 
+            # Fetch the full goal state if needed
```
dhivyaganesan, line 791: Not sure if this comment is  appropriate here. It threw me a little off which trying to understand the code. Guess the right place is at line 795 after the if statement and before fetch_full_goal_state?

azurelinuxagent/common/protocol/extensions_goal_state_from_extensions_config.py
dhivyaganesan: Not sure if I am missing something , do we still need 'id' since we can always use 'incarnation' in case of extension_config and etag in case of vm settings ?

Also curious to know why it is important to distinguish , guess it is because of the regression you mentioned but not sure what the issue exactly is.  personally, I like the idea of abstraction since its easier to follow

tests/protocol/test_wire.py
```diff
@@ -1313,10 +1314,89 @@ def http_get_handler(url, *_, **__):
...
                 self.assertEqual(3, len(messages), "Expected additional errors to be reported in the next period (got: {0})".format(messages))
 
+    def test_it_should_use_vm_settings_by_default(self):
+        with mock_wire_protocol(mockwiredata.DATA_FILE_VM_SETTINGS) as protocol:
+            extensions_goal_state = protocol.get_extensions_goal_state()
+            self.assertTrue(
+                isinstance(extensions_goal_state, ExtensionsGoalStateFromVmSettings),
+                'The extensions goal state should have been created from the vmSettings (got: {0})'.format(type(extensions_goal_state)))
+
+    def _assert_is_extensions_goal_state_from_extesions_config(self, extensions_goal_state):
```
dhivyaganesan, line 1324: Nit: typo : extesions

azurelinuxagent/common/protocol/wire.py
narrieta: i can remove the comment altogether if you find it confusing, let me know

azurelinuxagent/common/protocol/extensions_goal_state_from_extensions_config.py
narrieta: 'id' is used when we just need a way to distinguish from one goal state to another (e.g. when saving to disk for debugging purposes). 'incarnation' and 'etag' are used when we actually need those values (e.g. report status needs the incarnation)

though at this point 'id' corresponds to 'incarnation' or 'etag', once the 2 goal states get merged into 1 i'm planning on making the 'id' the concatenation of both values

tests/protocol/test_wire.py
narrieta: thanks; fixed

azurelinuxagent/common/protocol/extensions_goal_state_from_extensions_config.py
dhivyaganesan: Thanks for the explanation.

dhivyaganesan (approved):

larohra (approved):

azurelinuxagent/common/protocol/wire.py
larohra: Cool cool. So this will be taken care in the next PR?

azurelinuxagent/common/protocol/wire.py
larohra: sounds good.

azurelinuxagent/common/protocol/wire.py
narrieta: Yes, I need some refactoring to get accurate telemetry and need to touch the error handling code first. That is the reason I postponed the handling of not supported. Version check is just a variation of not supported.

PR #2454, opened 2021-12-31
Add telemetry for vmSettings

narrieta:
Adding telemetry for interesting errors in vmSettings (see _VmSettingsErrorReporter in wire.py for a description of the counters we keep track of)

Also, if the vmSettings API is not supported stop using it for 6 hours.

### DIFF
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -818,7 +818,7 @@ class WireClient(object):
                             self._extensions_goal_state = vm_settings_goal_state
                             vm_settings_goal_state_updated = True
 
-                    if goal_state_updated or vm_settings_goal_state_updated:
+                    if goal_state_updated or vm_settings_goal_state_updated and self._vm_settings_goal_state is not None:
                         # compare() raises a GoalStateMismatchError if the goal states don't match
                         ExtensionsGoalState.compare(self._goal_state.extensions_config, self._vm_settings_goal_state)
                 except Exception as error:
@@ -1529,7 +1529,7 @@ class _VmSettingsErrorReporter(object):
                 }
                 message = json.dumps(summary)
                 logger.info("[VmSettingsSummary] {0}", message)
-                add_event(WALAEventOperation.VmSettingsSummary, message=message, is_success=False, log_event=False)
+                add_event(op=WALAEventOperation.VmSettingsSummary, message=message, is_success=False, log_event=False)
             self._error_count = 0
             self._request_count = 0
             self._error_count = 0
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -907,7 +907,7 @@ class WireClient(object):
                     response_etag = h[1]
                     break
             else:  # since the vmSettings were updated, the response must include an etag
-                message = format_message("The vmSettings response does no include an Etag header")
+                message = format_message("The vmSettings response does not include an Etag header")
                 self._vm_settings_error_reporter.report_error(message)
                 raise ProtocolError(message)
 
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -579,6 +579,9 @@ class WireClient(object):
         self._extensions_goal_state = None  # The goal state to use for extensions; can be an ExtensionsGoalStateFromVmSettings or ExtensionsGoalStateFromExtensionsConfig
         self._vm_settings_goal_state = None  # Cached value of the most recent ExtensionsGoalStateFromVmSettings
         self._host_plugin = None
+        self._host_plugin_version = FlexibleVersion("0.0.0.0")  # Version 0 means "unknown"
+        self._host_plugin_supports_vm_settings = False
+        self._host_plugin_supports_vm_settings_next_check = datetime.now()
         self.status_blob = StatusBlob(self)
         self.goal_state_flusher = StateFlusher(conf.get_lib_dir())
         self._vm_settings_error_reporter = _VmSettingsErrorReporter()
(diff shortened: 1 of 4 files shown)


azurelinuxagent/common/utils/restutil.py
```diff
@@ -354,7 +354,14 @@ def http_request(method,
                  max_retry=None,
                  retry_codes=None,
                  retry_delay=DELAY_IN_SECONDS,
-                 redact_data=False):
+                 redact_data=False,
+                 return_raw_response=False):
```
narrieta, line 358: The error handling done by http_request mixes logic for several APIs, and makes hard (or impossible) to handle specific errors for new APIs (in this case vmSettings). Our mock infrastructure is based on this method, though, so I decided to add a new parameter to bypass all the error handling here.

Eventually this method needs to be rewritten/replaced to do better error handling.

tests/protocol/test_wire.py
```diff
@@ -1256,24 +1254,22 @@ def assert_no_exception(test_case, test_function, expected_error):
...
-
-                    def http_get_vm_settings(_method, _host, _relative_url, **_):
+                def http_get_handler(url, *_, **__):
+                    if self.is_host_plugin_vm_settings_request(url):
                         if isinstance(mock_response, Exception):
                             raise mock_response
                         return mock_response
+                    return None
 
-                    with patch("azurelinuxagent.common.utils.restutil._http_request", side_effect=http_get_vm_settings):
```
narrieta, line 1267: I was mocking the internal method to bypass the error handling done by the public method, but that is no longer needed.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -871,40 +888,53 @@ def get_vm_settings():
...
+                self._vm_settings_error_reporter.report_error(message)
+                raise ProtocolError(message)
 
             response_content = self.decode_config(response.read())
             vm_settings = ExtensionsGoalStateFactory.create_from_vm_settings(response_etag, response_content)
 
-            # log the HostGAPlugin version when we fetch the vmSettings for the first time
-            is_first_vm_settings = self._vm_settings_goal_state is None
-            if is_first_vm_settings:
+            # log the HostGAPlugin version
```
narrieta, line 913: @larohra this fixes the issue you pointed out in my previous PR (always logging this when vmSettings is not supported)

azurelinuxagent/common/protocol/wire.py
```diff
@@ -871,40 +888,53 @@ def get_vm_settings():
...
             return vm_settings
 
         except ProtocolError:
             raise
-        except HttpError as http_error:
-            raise Exception("{0} [correlation ID: {1} eTag: {2}]".format(ustr(http_error), correlation_id, etag))
         except Exception as exception:
-            error = textutil.format_exception(exception)  # since this is a generic error, we format the exception to include the traceback
-            raise Exception("Error fetching vmSettings [correlation ID: {0} eTag: {1}]: {2}".format(correlation_id, etag, error))
+            if isinstance(exception, IOError) and "timed out" in ustr(exception):
```
kevinclark19a, line 931: I don't like this string comparision but I couldn't find a version agnostic way of testing for this. Looks like the classing heirarchy is different for 2 and 3. I tried checking `errno`, but that was `None` on my python 3.8 install...

Any other possibilities? Have you looked deeply into this yet?

narrieta: don't like it either, but i found no way other than this check

the existing code in restapi.py handles IOError. telemetry shows that the exception is indeed an IOError and that the message is "timed out"

tests/protocol/test_wire.py
```diff
@@ -1392,13 +1389,82 @@ def test_it_should_compare_goal_states_when_extensions_config_change(self):
...
+
+            mock_response = httpclient.HTTPException()
+            protocol.client.update_goal_state()
+            protocol.client.update_goal_state()
+
+            # force the summary by resetting its period and calling update_goal_state
+            with patch("azurelinuxagent.common.protocol.wire.add_event") as add_event:
+                mock_response = None  # stop producing errors
+                protocol.client._vm_settings_error_reporter._next_period = datetime.now()
+                protocol.client.update_goal_state()
```
kevinclark19a, line 1422: This can't be replaced by calling the private `_reset` function on the error reporter, can it? That would have the benefit of avoiding any potential side effects of the `update_goal_state` call and removing the need for the `+ 2` down on the definition for `expected['requests']`.

kevinclark19a (dismissed):

tests/protocol/test_wire.py
narrieta: no, _reset sets the counters to 0; this actually output the summary

azurelinuxagent/common/protocol/wire.py
```diff
@@ -871,40 +888,53 @@ def get_vm_settings():
...
                     error_description = "[Internal error in HostGAPlugin] {0}".format(error_description)
+
+                if 400 <= response.status <= 499:
+                    self._vm_settings_error_reporter.report_error(error_description, _VmSettingsError.ClientError)
+                elif 500 <= response.status <= 599:
+                    self._vm_settings_error_reporter.report_error(error_description, _VmSettingsError.ServerError)
+                else:
+                    self._vm_settings_error_reporter.report_error(error_description)
+
                 raise ProtocolError("GET vmSettings [correlation ID: {0}]: {1} ({2})".format(correlation_id, response.status, error_description))
```
nagworld9, line 899: Protocol error had extra error message like correlation Id and status which we are not reporting back in L#842. So can you change the error description L#890 to include extra details.

narrieta: They are added in the exception message @ L#899

nagworld9: yeah, my understanding is that we don't report if it's coming from ProtocolError exception. So, we lose that information.

`# _fetch_vm_settings_goal_state() does its own detailed error reporting and raises ProtocolError; do not report those`

narrieta: got it, let me fix that

narrieta: fixed (added correlation id and etag to all messages) (the status is already added by restutil.read_response_error)

azurelinuxagent/common/protocol/wire.py
kevinclark19a: from what I could find, the exception is a `socket.timeout` error, but from what I could tell, there's differences between the definition of that class in 2 vs. 3.

narrieta: According to telemetry it's an IOError (derived from OSError after 3.3, I think). These are the errors we have got for the last week

[HttpError] [HTTP Retry] GET http://168.63.129.16:32526/vmSettings -- Status Code 502
[HttpError] [HTTP Retry] GET http://168.63.129.16:32526/vmSettings -- Status Code 500
[HttpError] [HTTP Retry] GET http://168.63.129.16:32526/vmSettings -- Status Code 403
[HttpError] [HTTP Failed] GET http://168.63.129.16:32526/vmSettings -- IOError timed out
[HttpError] [HTTP Failed] GET http://168.63.129.16:32526/vmSettings -- IOError [Errno 113] No route to host
[HttpError] [HTTP Failed] GET http://168.63.129.16:32526/vmSettings -- IOError [Errno 111] Connection refused
[HttpError] [HTTP Failed] GET http://168.63.129.16:32526/vmSettings -- IOError [Errno 110] Connection timed out
[HttpError] [HTTP Failed] GET http://168.63.129.16:32526/vmSettings -- IOError [Errno 104] Connection reset by peer
[HttpError] [HTTP Failed] GET http://168.63.129.16:32526/vmSettings -- IOError [Errno 101] Network is unreachable

tests/protocol/test_wire.py
```diff
@@ -1271,7 +1271,7 @@ def http_get_handler(url, *_, **__):
         test_error_in_http_request("Internal error in the HostGAPlugin", MockHttpResponse(httpclient.BAD_GATEWAY), "[Internal error in HostGAPlugin] [HTTP Failed] [502: None]")
         test_error_in_http_request("Arbitrary error in the request (BAD_REQUEST)", MockHttpResponse(httpclient.BAD_REQUEST), "[HTTP Failed] [400: None]")
         test_error_in_http_request("Generic error in the request", Exception("GENERIC REQUEST ERROR"), "GENERIC REQUEST ERROR")
-        test_error_in_http_request("Response headers with no Etag", MockHttpResponse(200, b""), "The vmSettings do no include an Etag")
+        test_error_in_http_request("Response headers with no Etag", MockHttpResponse(200, b""), "The vmSettings response does no include an Etag header")
```
kevinclark19a, line 1274: NIT: is `The vmSettings response does no include` supposed to be `...does not include`?

kevinclark19a (approved):

kevinclark19a (approved):

nagworld9 (approved):

PR #2468, opened 2022-01-11
Improve reporting for Depends on errors for Single Config extensions

kevinclark19a:
## Description

In the current code base, SC extensions report no extension status if any of their dependencies fail. This makes for unhelpful customer facing error messages. This PR  addresses this by explicitly reporting failed dependencies as a reason for not running.

---

### PR information

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -241,7 +241,6 @@ class ExtHandlerState(object):
     Installed = "Installed"
     Enabled = "Enabled"
     FailedUpgrade = "FailedUpgrade"
-    DidNotRun = "DidNotRun"
 
 
 class GoalStateStatus(object):
@@ -514,11 +513,6 @@ class ExtHandlersHandler(object):
                 # For SC extensions, overwrite the HandlerStatus with the relevant message
                 else:
                     handler_i.set_handler_status(message=depends_on_err_msg, code=-1)
-                    # We need to convey state to the extension status reporting logic. It will be looking for a status file
-                    # that should have been created by the extension itself, but it cannot be run in this case. Additionally,
-                    # the agent cannot create the status file itself as some extensions will not run when their status file
-                    # already exists, per a legacy bug in the agent.
-                    handler_i.set_handler_state(ExtHandlerState.DidNotRun)
 
                 continue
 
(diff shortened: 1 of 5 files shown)


tests/ga/test_multi_config_extension.py
```diff
@@ -1129,12 +1129,22 @@ def __assert_invalid_status_scenario(self, protocol, fail_code, mc_status="NotRe
...
+        sc_independent_handler = self._assert_and_get_handler_status(aggregate_status=protocol.aggregate_status,
+                                                                    handler_name="Microsoft.Azure.Geneva.GenevaMonitoring",
+                                                                    handler_version="1.1.0", status="NotReady", message=err_msg)
+        for handler in sc_independent_handler:
+            extension_status_msg = get_keypath_or_throw(handler, "runtimeSettingsStatus/settingsStatus/status/formattedMessage/message",
+                msg="SC extension '{0}' did not report extension status even though it is independent.".format(handler['handlerName']))
+            
+            # TODO: This might be a bug; independent extensions probably should not fail when 
+            # other extensions do, but that's what is happening here.
+            self.assertEqual(err_msg, extension_status_msg, "Expected the independent extension's status to match {0}, but got '{1}' instead.")
```
kevinclark19a, line 1147: The exisiting behavior of this test treated dependent SC extensions and independent SC extensions the same, but that doesn't seem right. Left this `TODO` here and opened a bug. When that (potential) bug is fixed, this test will fail and should be updated to reflect `sc_independent_handler` succeeding.

azurelinuxagent/ga/exthandlers.py
```diff
@@ -241,6 +241,7 @@ class ExtHandlerState(object):
     Installed = "Installed"
     Enabled = "Enabled"
     FailedUpgrade = "FailedUpgrade"
+    DidNotRun = "DidNotRun"
```
larohra, line 244: I do not like the idea of creating a new state for this case. It seems like an overkill for what we're trying to achieve since creating and maintaining a new state requires dealing with the filesystem which comes with its own issues. Additionally, I don't think it fits the needs completely either. 

For eg, you're setting the HandlerState for Single config extension to `DidNotRun` but not for MultiConfig extensions. This mismatch makes the code more prone to errors.

larohra: Maybe save this state in-memory and use it when its about to be reported? For example modifying the `extHandler` object because that's common for both the extension processing and reporting and can potentially be used for this case?

PR #2471, opened 2022-01-13
Don't test for agent log in between tests ETP

larohra:
## Description
Currently we were checking for errors in agent log mid-test in ETP. Refactored that to only check at the end and check for relevant errors mid-test

Issue # 

---

### PR information

### DIFF
--- a/dcr/scenario_utils/check_waagent_log.py
+++ b/dcr/scenario_utils/check_waagent_log.py
@@ -92,21 +92,6 @@ def check_waagent_log_for_errors(waagent_log=AGENT_LOG_FILE, ignore=None):
         {
             'message': r"WARNING EnvHandler ExtHandler Dhcp client is not running."
         },
-        # This happens in CENTOS and RHEL when waagent attempt to format and mount the error while cloud init is already doing it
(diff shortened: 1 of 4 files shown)


dcr/scenario_utils/check_waagent_log.py
```diff
@@ -93,6 +92,21 @@ def check_waagent_log_for_errors(waagent_log=AGENT_LOG_FILE, ignore=None):
         {
             'message': r"WARNING EnvHandler ExtHandler Dhcp client is not running."
         },
+        # This happens in CENTOS and RHEL when waagent attempt to format and mount the error while cloud init is already doing it
```
narrieta, line 95: do we get these errors at all in DCR V2?

larohra: haven't seen it honestly. I just copied the complete ignore list from v1 and copied it in v2.

narrieta: we wouldn't expect it, right? sounds like a provisioning error. i'd suggest removing it here and we can always add it back if there a reason to do so

larohra: makes sense, let me remove it

narrieta (approved):

kevinclark19a (approved):

PR #2473, opened 2022-01-14
Set agent version to 2.7.0.1

narrieta:


nagworld9 (approved):

azurelinuxagent/common/version.py
```diff
@@ -209,7 +209,7 @@ def has_logrotate():
 #
 #            When doing a release, be sure to use the actual agent version.
 #
-AGENT_VERSION = '2.7.0.0'
+AGENT_VERSION = '2.7.0.1'
```
larohra, line 212: Since we haven't released the 2.7.0.0 version yet, we technically dont need the version update.

narrieta: It's part of our processes. We started the 2.7.0.0 release, found an issue, fixed it and now we are incrementing the version.

PR #2475, opened 2022-01-18
Set vmSettings as primary channel for extensions goal state

narrieta:
Now  vmSettings is the primary channel for extensions goal state. There is not more fallback to extensionsConfig nor comparison of goal states.

ExtensionConfig is used only when Fast Track is not enable or is not supported.

All goal states are still going thru Fabric.

### DIFF
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -789,7 +789,9 @@ class WireClient(object):
             # (via the self._fetch_vm_settings_goal_state method).
             #
             # We always need at least 2 queries: one to the WireServer (to check for incarnation changes) and one to the HostGAPlugin
-            # (to check for extension updates).
+            # (to check for extension updates). Note that vmSettings are not a full goal state; they include only the extension information
+            # (minus certificates). The check on incarnation (which is also not included in the vmSettings) is needed to check for changes
+            # in, for example, the remote users for JIT access.
             #
             # We start by fetching the goal state from the WireServer. The response to this initial query will include the incarnation,
             # container ID, role config, and URLs to the rest of the goal state (certificates, remote users, extensions config, etc). We
@@ -827,7 +829,7 @@ class WireClient(object):
                 goal_state_updated = True
 
             #
-            # And, lastly, we fall back to extensionsConfig if Fast Track is disabled or not supported
+            # And, lastly, we use extensionsConfig if we don't have the vmSettings (Fast Track may be disabled or not supported).
             #
             if vm_settings_goal_state is not None:
                 self._extensions_goal_state = vm_settings_goal_state
@@ -850,7 +852,7 @@ class WireClient(object):
         Queries the vmSettings from the HostGAPlugin and returns an (ExtensionsGoalStateFromVmSettings, bool) tuple with the vmSettings and
         a boolean indicating if they are an updated (True) or a cached value (False).
 
-        Raises TypeError if the HostGAPlugin does not support the vmSettings API, or ProtocolError if the request fails for any other reason
+        Raises VmSettingsNotSupported if the HostGAPlugin does not support the vmSettings API, or ProtocolError if the request fails for any other reason
         (e.g. not supported, time out, server error).
         """
         def raise_not_supported(reset_state=False):
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -785,39 +785,34 @@ class WireClient(object):
         """
         try:
             #
-            # The entire goal state needs to be retrieved from the WireServer (via the GoalState class), and the HostGAPlugin
+            # The goal state needs to be retrieved using both the WireServer (via the GoalState class) and the HostGAPlugin
             # (via the self._fetch_vm_settings_goal_state method).
             #
-            # We fetch it in 3 parts:
+            # We always need at least 2 queries: one to the WireServer (to check for incarnation changes) and one to the HostGAPlugin
+            # (to check for extension updates).
             #
-            # 1) The "main" goal state from the WireServer, which includes the incarnation, container ID, role config, and URLs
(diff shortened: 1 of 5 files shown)


azurelinuxagent/common/protocol/wire.py
```diff
@@ -780,45 +779,40 @@ def update_host_plugin_from_goal_state(self):
...
         Updates the goal state if the incarnation or etag changed or if 'force_update' is True
         """
         try:
             #
-            # The entire goal state needs to be retrieved from the WireServer (via the GoalState class), and the HostGAPlugin
+            # The goal state needs to be retrieved using both the WireServer (via the GoalState class) and the HostGAPlugin
             # (via the self._fetch_vm_settings_goal_state method).
             #
-            # We fetch it in 3 parts:
+            # We always need at least 2 queries: one to the WireServer (to check for incarnation changes) and one to the HostGAPlugin
```
larohra, line 791: > one to the WireServer (to check for incarnation changes)

I was under the impression that this was going to be abstracted by the HostGA Plugin? Is the vmSettings only going to be an abstraction for the `ExtensionConfig` part of the GoalState and not the complete goal state?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -875,7 +850,8 @@ def _fetch_vm_settings_goal_state(self, force_update):
         Queries the vmSettings from the HostGAPlugin and returns an (ExtensionsGoalStateFromVmSettings, bool) tuple with the vmSettings and
         a boolean indicating if they are an updated (True) or a cached value (False).
 
-        Raises ProtocolError if the request fails for any reason (e.g. not supported, time out, server error)
+        Raises TypeError if the HostGAPlugin does not support the vmSettings API, or ProtocolError if the request fails for any other reason
```
larohra, line 853: NIT: 
```suggestion
        Raises VmSettingsNotSupported if the HostGAPlugin does not support the vmSettings API, or ProtocolError if the request fails for any other reason
```

azurelinuxagent/common/protocol/wire.py
```diff
@@ -780,45 +779,40 @@ def update_host_plugin_from_goal_state(self):
...
             if conf.get_enable_fast_track():
                 try:
                     vm_settings_goal_state, vm_settings_goal_state_updated = self._fetch_vm_settings_goal_state(force_update=force_update)
-
-                except Exception as error:
-                    # _fetch_vm_settings_goal_state() does its own detailed error reporting and raises ProtocolError; do not report those
-                    if not isinstance(error, ProtocolError):
-                        self._vm_settings_error_reporter.report_error(format_exception(error))
-                self._vm_settings_error_reporter.report_summary()
+                except VmSettingsNotSupported:
```
larohra, line 810: In case of a `ProtocolError`, we won't even try to fetch the `ExtensionsConfig` from Wireserver. Is that the expected behavior? Shouldn't we try to fetch ExtensionsConfig from Wireserver if we can't fetch VMSettings for whatever reason?

tests/protocol/test_wire.py
```diff
@@ -1242,46 +1241,13 @@ def http_get_vm_settings(_method, _host, _relative_url, **kwargs):
...
-                with mock_wire_protocol(mockwiredata.DATA_FILE_VM_SETTINGS) as protocol:
-                    protocol.client.update_goal_state()
 
-        assert_no_exception("Goal state mismatch", fail_compare, "TEST COMPARE FAILED")
+    @staticmethod
+    def _update_goal_state_ignoring_errors(protocol):
+        try:
+            protocol.client.update_goal_state()
+        except ProtocolError:
+            pass
```
larohra, line 1250: How are we getting so many errors that we need to ignore? Isint this a mocked environment, we shouldn't be getting any errors right?

azurelinuxagent/common/protocol/wire.py
narrieta: The vmsettings don't include the incarnation number, that comes from the wireserver. Even if Fast Track is enabled and supported we also need to check for other changes in the goal state (e.g JIT access would change the remote users). Will add a little more text to this comment

azurelinuxagent/common/protocol/wire.py
narrieta: This change makes VmSettings *the* way to get the extensions goal state. Errors on vmSettings need to be handled the same as  any other errors in the WireServer goal state.

larohra: Gotcha, basically in the scenario that there is FT data in the vmSettings, then falling back to Wireserver would be wrong too. 

Thanks for explaining!

azurelinuxagent/common/protocol/wire.py
larohra: vmSettings would still catch the JIT user scenario right? Since it is fetching the goal states from both channels and comparing and giving us the latest one right?

Also, does HostGA plan to move the complete goal state to vmSettings (certs, roleConfig, etc) or is it going to be limited to the ExtensionsConfig?

larohra (approved):

azurelinuxagent/common/protocol/wire.py
narrieta: vmsettings is not a complete goal state as of today, so it would not include remote users

afaik, vmsettings will not go beyond extensions (i.e. no remote users, for example). as of today, the certs are still missing, but i believe they will be added in a (as yet unspecified) future version

azurelinuxagent/common/protocol/wire.py
narrieta: `Gotcha, basically in the scenario that there is FT data in the vmSettings, then falling back to Wireserver would be wrong too.
`

Yes, exactly. Not for this specific PR because everything is Fabric still, but well get FastTrack as soon as I implement the handshake with CRP (coming 2 or 3 PRs from this one)

nagworld9 (approved):

larohra (approved):

nagworld9 (approved):

PR #2479, opened 2022-01-20
Upgrade/Downgrade agent if Requested Version available

larohra:
## Description
This PR contains the next set of changes for GA Versioning. Certain key points of this PR - 
- The agent wont accept downgrades for requested versions < current daemon version
- Blacklist the Agent on downgrade
- AutoUpdate as soon as a requested version is available rather than waiting for the wait times (4 hrs/24 hrs)
- Remove dead code that is not relevant anymore
- A bit of a re-write of the auto-update logic (with emphasis on making sure the current scenario is not broken)

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/common/version.py
+++ b/azurelinuxagent/common/version.py
@@ -298,6 +298,3 @@ def set_goal_state_agent():
 
 GOAL_STATE_AGENT_VERSION = set_goal_state_agent()
 
-
-def is_current_agent_installed():
-    return CURRENT_AGENT == AGENT_LONG_VERSION
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -457,7 +457,7 @@ class UpdateHandler(object):
 
     def _process_goal_state(self, exthandlers_handler, remote_access_handler):
 
-        def log_update_time():
+        def log_next_update_time():
             next_normal_time, next_hotfix_time = self.__get_next_upgrade_times()
             upgrade_type = self.__get_agent_upgrade_type(available_agent)
             next_time = next_hotfix_time if upgrade_type == AgentUpgradeType.Hotfix else next_normal_time
@@ -468,6 +468,32 @@ class UpdateHandler(object):
                       message=message_, log_event=False)
             logger.info(message_)
 
+        def handle_updates_for_requested_version():
+            if requested_version == CURRENT_VERSION:
+                logger.info("The requested version is running as the current version: {0}".format(requested_version))
+                return
+            if requested_version < CURRENT_VERSION:
+                prefix = "downgrade"
+                # In case of a downgrade, we blacklist the current agent to avoid starting it back up ever again
+                # (the expectation here being that if RSM is asking us to a downgrade, there's a good reason for not wanting the current version).
+                try:
+                    # We should always have an agent directory for the CURRENT_VERSION
+                    # (unless the CURRENT_VERSION == daemon version, but since we don't support downgrading
+                    # below daemon version, we will never reach this code path if that's the scenario)
+                    current_agent = next(agent for agent in self.agents if agent.version == CURRENT_VERSION)
+                    current_agent.mark_failure(is_fatal=True)
+                except StopIteration:
+                    logger.warn(
+                        "Could not find a matching agent with current version {0} to blacklist, skipping it".format(
+                            CURRENT_VERSION))
+            else:
+                # In case of an upgrade, we don't need to blacklist anything as the daemon will automatically
+                # start the next available highest version which would be the requested version
+                prefix = "upgrade"
+            raise AgentUpgradeExitException(
+                "Agent {0} discovered, moving to the requested version: {1} -- exiting".format(
+                    prefix, requested_version))
+
         protocol = exthandlers_handler.protocol
         if not self._try_update_goal_state(protocol):
             self._heartbeat_update_goal_state_error_count += 1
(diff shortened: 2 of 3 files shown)


azurelinuxagent/common/version.py
```diff
@@ -298,6 +298,3 @@ def set_goal_state_agent():
 
 GOAL_STATE_AGENT_VERSION = set_goal_state_agent()
 
-
-def is_current_agent_installed():
```
larohra, line 302: This code logic was not being used anymore, so removed it

azurelinuxagent/ga/update.py
```diff
@@ -815,17 +853,6 @@ def _is_orphaned(self):
 
         return fileutil.read_file(conf.get_agent_pid_file_path()) != ustr(parent_pid)
 
-    def _is_version_eligible(self, version):
```
larohra, line 818: This function was only being used in 1 place (during auto-update) and due to the condition below, it was always True (`if version == CURRENT_VERSION and is_current_agent_installed(): return True`). Since it was not adding much value to the code, removed it

tests/ga/test_update.py
```diff
@@ -1100,34 +1101,6 @@ def test_is_orphaned_returns_true_if_parent_does_not_exist(self):
         with patch('os.getppid', return_value=42):
             self.assertTrue(self.update_handler._is_orphaned)
 
-    def test_is_version_available(self):
```
larohra, line 1103: Removed the test cases for the functionality that was removed

azurelinuxagent/ga/update.py
```diff
@@ -459,29 +456,64 @@ def __goal_state_updated(self, incarnation):
...
+                except StopIteration:
+                    logger.warn(
+                        "Could not find a matching agent with current version {0} to blacklist, skipping it".format(
+                            CURRENT_VERSION))
+            else:
+                # In case of an upgrade, we don't need to blacklist anything as the daemon will automatically
+                # start the next available highest version which would be the requested version
+                prefix = "upgrade"
+            raise AgentUpgradeExitException(
+                "Agent {0} discovered, moving to the requested version: {1} -- exiting".format(
```
narrieta, line 492: can we rephrase to something like "Exiting current process to upgrade/downgrade to the request Agent version XYZ"?

azurelinuxagent/ga/update.py
```diff
@@ -923,59 +967,78 @@ def _check_and_download_agent_if_upgrade_available(self, protocol, base_version=
         if not conf.get_autoupdate_enabled():
             return False
 
-        def report_error(msg_, version=CURRENT_VERSION):
+        def report_error(msg_, version_=CURRENT_VERSION, op=WALAEventOperation.Download):
```
narrieta, line 970: Let's not default this WALAEventOperation.Download, but make it an explicit parameter. There are some usages where the default is being used to report issues not related do download

narrieta (commented):
submitting a few comments for now... still reviewing the code

azurelinuxagent/ga/update.py
```diff
@@ -459,29 +456,64 @@ def __goal_state_updated(self, incarnation):
         return incarnation != self.last_incarnation
 
     def _process_goal_state(self, exthandlers_handler, remote_access_handler):
+
```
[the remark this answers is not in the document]
larohra: yup agreed, its getting too cluttered, will refactor it.

azurelinuxagent/ga/update.py
```diff
@@ -459,29 +456,64 @@ def __goal_state_updated(self, incarnation):
...
+                prefix = "downgrade"
+                # In case of a downgrade, we blacklist the current agent to avoid starting it back up ever again
+                # (the expectation here being that if RSM is asking us to a downgrade,
+                # there's a good reason for not wanting the current version).
+                try:
+                    # We should always have an agent directory for the CURRENT_VERSION
+                    # (unless the CURRENT_VERSION == daemon version, but since we don't support downgrading
+                    # below daemon version, we will never reach this code path if that's the scenario)
+                    current_agent = next(agent for agent in self.agents if agent.version == CURRENT_VERSION)
+                    current_agent.mark_failure(is_fatal=True)
```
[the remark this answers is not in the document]
larohra: Done (though I did mention the word `blacklisting` because I feel its pretty self explanatory (and makes it easier to explain) even to folks outside our dev group who read the logs for debugging.

azurelinuxagent/ga/update.py
```diff
@@ -459,29 +456,64 @@ def __goal_state_updated(self, incarnation):
...
+                except StopIteration:
+                    logger.warn(
```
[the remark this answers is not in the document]
larohra: >  i think that would happen only if the current version is running as the daemon

This scenario is already handled separately. If that is the case, we would not hit this code path. That is the primary reason for this warning, as that situation should not happen ever

azurelinuxagent/ga/update.py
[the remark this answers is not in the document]
larohra: I totally agree with the sentiment here, the problem is that those helper functions are being used in multiple places which are not necessarily being called during the agent auto-update (for example: one of the calls is in the `get_latest_agent_greater_than_daemon` function which is also called in the `run_latest` function (called by the daemon) to determine which version to start the exthandler process as. 

Let me do a deep dive of all the places this is called and see if I can reduce any calls without breaking the code too much

larohra: Which usage exactly did you see where we're reporting it as a download failure but its not one? This helper function is mainly to log those errors inside the `_download_agent_if_upgrade_available` function and instead of one call to it, all other calls for this function are `Download` errors, that's why I defaulted to that operation to make the code cleaner

azurelinuxagent/ga/update.py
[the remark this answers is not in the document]
larohra: Done

azurelinuxagent/ga/update.py
larohra: done

azurelinuxagent/ga/update.py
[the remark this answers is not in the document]
larohra: As it turns out, wasn't that big of a change. Removed the use of that flag from helper functions

narrieta (approved):

kevinclark19a (approved):

PR #2480, opened 2022-01-20
Add check for all Firewall rules periodically

nagworld9:
## Description

Issue # 

---

### PR information


azurelinuxagent/common/osutil/default.py
```diff
@@ -86,9 +86,11 @@ def get_firewall_accept_command(wait, command, destination, owner_uid):
...
     return AddFirewallRules.get_iptables_drop_command(wait, command, destination)
 
-
-def get_firewall_list_command(wait):
-    return _add_wait(wait, ["iptables", "-t", "security", "-L", "-nxv"])
+# Verbose output add extra details like packets and bytes.
+def get_firewall_list_command(wait, verbose=True):
+    if verbose:
+        return _add_wait(wait, ["iptables", "-t", "security", "-L", "-nxv"])
+    return _add_wait(wait, ["iptables", "-t", "security", "-L", "-nx"])
```
nagworld9, line 93: Added this to avoid verbose output for comparison purpose while logging

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -156,18 +156,20 @@ def __get_firewalld_base_command(command):
         return ["firewall-cmd", "--permanent", "--direct", command, "ipv4"]
 
     @staticmethod
-    def __get_common_command_params(command, destination):
+    def __get_common_command_params(command, destination, pos):
+        if command == AddFirewallRules.INSERT_COMMAND:
+            return ["-t", "security", command, "OUTPUT", pos, "-d", destination, "-p", "tcp", "-m"]
```
nagworld9, line 161: Insert command has position option

azurelinuxagent/ga/env.py
```diff
@@ -131,7 +132,14 @@ def _operation(self):
             self._osutil.remove_legacy_firewall_rule(dst_ip=self._protocol.get_endpoint())
             self._try_remove_legacy_firewall_rule = True
 
+        before_enable_firewall_list = self._osutil.get_firewall_list(verbose=False)
         success = self._osutil.enable_firewall(dst_ip=self._protocol.get_endpoint(), uid=os.getuid())
+        after_enable_firewall_list = self._osutil.get_firewall_list(verbose=False)
+
+        if self._log_initial_firewall_list or before_enable_firewall_list != after_enable_firewall_list:
```
nagworld9, line 139: Since this is periodic operation logging very first time or if firewall list got changed after enable.

azurelinuxagent/common/osutil/default.py
```diff
@@ -86,9 +86,11 @@ def get_firewall_accept_command(wait, command, destination, owner_uid):
 def get_firewall_drop_command(wait, command, destination):
     return AddFirewallRules.get_iptables_drop_command(wait, command, destination)
 
-
-def get_firewall_list_command(wait):
-    return _add_wait(wait, ["iptables", "-t", "security", "-L", "-nxv"])
+# Verbose output add extra details like packets and bytes.
+def get_firewall_list_command(wait, verbose=True):
```
larohra, line 90: NIT: PEP warning 
> PEP 8: E302 expected 2 blank lines, found 1

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -156,18 +156,20 @@ def __get_firewalld_base_command(command):
...
         return ["-t", "security", command, "OUTPUT", "-d", destination, "-p", "tcp", "-m"]
 
     @staticmethod
-    def __get_common_accept_command_params(wait, command, destination, uid):
-        cmd = AddFirewallRules.__get_common_command_params(command, destination)
+    def __get_common_accept_command_params(wait, command, destination, uid, pos=""):
+        cmd = AddFirewallRules.__get_common_command_params(command, destination, pos)
```
larohra, line 166: There's just way too many redirects for the `pos` column here which makes it really hard to follow (and think about cases if it can cause issues somewhere). Is there any way we can refactor this code to make it a bit more easier to understand? Maybe consider refactoring the code the same way `get_accept_tcp_rule` function is written? 

Additionally, since the `pos` is only needed for the `INSERT` commands, maybe consider creating a new function for it rather than hooking it up in the existing commands.

(I'm asking this just for clarity's sake because the current code is really hard to follow, not saying anything is wrong with the logic you applied here)

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -225,20 +227,42 @@ def __execute_cmd(cmd):
                                                                                             ustr(error.stderr))
             raise Exception(msg)
 
+    @staticmethod
+    def __execute_check_command(cmd):
+        # Helper to execute a check command, returns True if no exception
```
larohra, line 232: The first line for the comment seems irrelevant here, the 2nd line below kind of captures the gist of what the function is doing.

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -185,13 +187,13 @@ def get_accept_tcp_rule(command, destination, firewalld_command="", wait=""):
     @staticmethod
     def get_iptables_accept_command(wait, command, destination, owner_uid):
         cmd = AddFirewallRules.__get_iptables_base_command()
-        cmd.extend(AddFirewallRules.__get_common_accept_command_params(wait, command, destination, owner_uid))
+        cmd.extend(AddFirewallRules.__get_common_accept_command_params(wait, command, destination, owner_uid, pos="2"))
```
larohra, line 190: I have a question on how the position works for `iptables`. What if say the DNS rule and the DROP rule exist, so in this case if we add the ACCEPT command at `pos: 2`, will it be added at the right order? Or does it need the order of the rules to be in a specific order? 

Can there be a scenario where the rules are in a messed up order? Say the DROP rule is 1st and then the TCP rule is 3rd? What happens in that case? Will these code changes be able to handle that scenario?

azurelinuxagent/ga/env.py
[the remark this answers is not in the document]
nagworld9: The `success` here not necessarily mean we changed something in firewall every time. It's combination of check plus update the firewall rules if not exist. Consider the case (most of the time) all the rules present and enable firewall success. I don't want to list firewall in those cases. It's just too much given 30secs periodic time. So, I decided to compare before and after firewall list to log. Hope this answered your questions

azurelinuxagent/common/osutil/default.py
nagworld9: Updated

azurelinuxagent/common/osutil/default.py
```diff
@@ -314,6 +301,15 @@ def enable_firewall(self, dst_ip, uid):
                         "{0}".format(ustr(e)))
             return False
 
+    def get_firewall_list(self, verbose=True):
```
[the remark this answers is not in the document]
nagworld9: updated

azurelinuxagent/ga/env.py
```diff
@@ -131,7 +132,14 @@ def _operation(self):
             self._osutil.remove_legacy_firewall_rule(dst_ip=self._protocol.get_endpoint())
             self._try_remove_legacy_firewall_rule = True
 
+        before_enable_firewall_list = self._osutil.get_firewall_list(verbose=False)
```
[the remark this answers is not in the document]
nagworld9: updated

azurelinuxagent/common/utils/networkutil.py
nagworld9: removed

kevinclark19a (approved):

azurelinuxagent/common/persist_firewall_rules.py
```diff
@@ -149,6 +149,8 @@ def _setup_permanent_firewalld_rules(self):
             return
 
         logger.info("Firewall rules not added yet, adding them now using firewalld.service")
+        # Remove first if partial list present
+        AddFirewallRules.remove_firewalld_rules(self._dst_ip, self._uid)
```
larohra, line 153: Thanks for adding this here, make sure to add a test for this too in DCR (along with the normal iptable tests)

azurelinuxagent/ga/env.py
```diff
@@ -118,6 +118,8 @@ def __init__(self, osutil, protocol):
         self._osutil = osutil
         self._protocol = protocol
         self._try_remove_legacy_firewall_rule = False
+        self._log_initial_firewall_list = True
+        self._wait = self._osutil.get_firewall_will_wait()
```
larohra, line 122: This function has the potential of raising, and if it raises, the whole environment thread will not be started. Could you add more error handling here to ensure that never happens? Thanks!

larohra (changes requested):

azurelinuxagent/common/osutil/default.py
```diff
@@ -80,15 +80,18 @@ def get_accept_tcp_rule(wait, command, destination):
 
 
 def get_firewall_accept_command(wait, command, destination, owner_uid):
-    return AddFirewallRules.get_iptables_accept_command(wait, command, destination, owner_uid)
+    return AddFirewallRules.get_accept_command(command, destination, owner_uid, wait=wait)
```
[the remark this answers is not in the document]
nagworld9: Man, this being used in lot of places and especially in all test cases. I don't know the reason behind this. I'm afraid to change this might break test cases or need lot of changes.

azurelinuxagent/common/osutil/default.py
```diff
@@ -268,43 +271,29 @@ def remove_legacy_firewall_rule(self, dst_ip):
             logger.info(
                 "Unable to remove legacy firewall rule, won't try removing it again. Error: {0}".format(ustr(error)))
 
-    def enable_firewall(self, dst_ip, uid):
+    def enable_firewall(self, dst_ip, uid, wait):
```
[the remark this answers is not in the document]
nagworld9: I missed this part, thanks for pointing out.

azurelinuxagent/common/osutil/default.py
```diff
@@ -268,43 +271,29 @@ def remove_legacy_firewall_rule(self, dst_ip):
...
-            # If the DROP rule exists, make no changes
+            # check every iptable rule and delete others if any rule is missing
+            #   and append every iptable rule to the end of the chain.
             try:
-                drop_rule = get_firewall_drop_command(wait, AddFirewallRules.CHECK_COMMAND, dst_ip)
-                shellutil.run_command(drop_rule)
-                logger.verbose("Firewall appears established")
-                return True
+                if not AddFirewallRules.verify_iptables_rules_exist(wait, dst_ip, uid):
+                    self.remove_firewall(dst_ip, uid)
```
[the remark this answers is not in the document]
nagworld9: It does return code 1 if rule doesn't exist which was handled in `_delete_rule` function. 

`Additionally, I just checked, the function remove_firewall returns a boolean indicating whether it was able to delete the rules or not. If we're unable to delete the rules, are you just planning to add new rules irrespective?`
That's good point, it will mess up the rules and might have duplicate rules. I will update as If delete unsuccessful, we don't add rules.

azurelinuxagent/ga/env.py
nagworld9: look like It's not good place to initialize here. If some reason first time failed, we are not retrying every run. Additionally, if this raises, we shouldn't enable the firewall and updating the global flag `_enable_firewall` to false. Any Idea how we can achieve this in env thread.

azurelinuxagent/ga/env.py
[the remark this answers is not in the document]
nagworld9: Good point, we don't need this anymore with new changes and log whenever we add rules.

azurelinuxagent/common/osutil/default.py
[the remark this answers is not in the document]
larohra: Don't be afraid of change (especially for tests, you will need to get your hands dirtier sooner or later haha). It shouldn't be too much work :)

azurelinuxagent/common/osutil/default.py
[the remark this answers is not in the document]
larohra: > I will update as If delete unsuccessful, we don't add rules.

The con of this would be that the agent would not be able to access wireserver, and even worse, wireserver might open up to non-root until we try again (probably in the next 30 secs). What would be the con of adding the same rules multiple times? Would it break firewall rules?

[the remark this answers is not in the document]
larohra: Maybe consider adding a retry for deleting rules and if we're unable to delete it then just add the rules irrespective?

azurelinuxagent/ga/env.py
larohra: you can keep the original implementation, like initialize the wait inside the Operation function maybe?

azurelinuxagent/common/osutil/default.py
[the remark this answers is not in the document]
nagworld9: `What would be the con of adding the same rules multiple times? Would it break firewall rules?`

It won't' break. Looking at the cons side I think you are right, will add the rules irrespective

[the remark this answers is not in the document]
larohra: Make sure to test it out (that firewall rules are not being affected) :)

azurelinuxagent/ga/env.py
nagworld9: we may not be able to update the global flag `_enable_firewall ` from `default.py `file if it's inside Operation function, can we?

larohra: I think it should be, but just to keep things simple, I dont mind keeping it as is. I honestly don't like the design of using a global flag and then to stop trying to set firewalls if we fail even once, technically we should be retrying every hour or so until we can get it in just to make sure the firewalls are up, but I guess that's a bigger change than your bug so I'm good with as is.

azurelinuxagent/common/osutil/default.py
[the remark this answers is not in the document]
nagworld9: challenge accepted, lol. I have addressed it now :)

azurelinuxagent/common/osutil/default.py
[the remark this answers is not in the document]
nagworld9: not required since not adding this parameter

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -234,10 +238,10 @@ def add_iptables_rules(wait, dst_ip, uid):
         accept_tcp_rule = AddFirewallRules.get_accept_tcp_rule(AddFirewallRules.APPEND_COMMAND, dst_ip, wait=wait)
         AddFirewallRules.__execute_cmd(accept_tcp_rule)
 
-        accept_rule = AddFirewallRules.get_iptables_accept_command(wait, AddFirewallRules.APPEND_COMMAND, dst_ip, uid)
+        accept_rule = AddFirewallRules.get_accept_command(AddFirewallRules.APPEND_COMMAND, dst_ip, uid, wait=wait)
```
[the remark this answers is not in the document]
nagworld9: Updated

azurelinuxagent/common/osutil/default.py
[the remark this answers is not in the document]
nagworld9: Actually, someone thought about this before and added retry on delete rule. Kudos to them

azurelinuxagent/ga/env.py
nagworld9: Now I don't see a need for wait call to be outside of osutil functions, I reverted to original implementation

azurelinuxagent/common/osutil/default.py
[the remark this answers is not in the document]
larohra: I just noticed `remove_firewall` does its own call for `wait = self.get_firewall_will_wait()`. Maybe you can remove this one since we're anyways fetching the one in `enable_firewall`?

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -226,43 +210,62 @@ def __execute_cmd(cmd):
...
+        AddFirewallRules.__execute_firewall_commands(dst_ip, uid, firewalld_command=FirewallCmdDirectCommands.PassThrough)
 
     @staticmethod
     def check_firewalld_rule_applied(dst_ip, uid):
-        AddFirewallRules.__execute_firewalld_commands(FirewallCmdDirectCommands.QueryPassThrough, dst_ip, uid)
+        AddFirewallRules.__execute_firewall_commands(dst_ip, uid, firewalld_command=FirewallCmdDirectCommands.QueryPassThrough)
+
+    @staticmethod
+    def remove_firewalld_rules(dst_ip, uid):
+        AddFirewallRules.__execute_firewall_commands(dst_ip, uid, firewalld_command=FirewallCmdDirectCommands.RemovePassThrough)
```
larohra, line 271: Not sure if you already did this check, but did you verify how `firewalld` handles it if we try to `remove-passthrough` a command that does not exist? Depending on if it exits with a non-0 exit code, we might need to handle it properly in the `persist_firewall_rules.py @ ln: 53`

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -179,34 +167,30 @@ def get_accept_tcp_rule(command, destination, firewalld_command="", wait=""):
...
+                     '-j', 'ACCEPT']
         return cmd
 
     @staticmethod
-    def get_iptables_accept_command(wait, command, destination, owner_uid):
-        cmd = AddFirewallRules.__get_iptables_base_command()
-        cmd.extend(AddFirewallRules.__get_common_accept_command_params(wait, command, destination, owner_uid))
-        return cmd
+    def get_accept_command(command, destination, owner_uid, firewalld_command="", wait=""):
+        if firewalld_command != "":
```
larohra, line 176: Thanks for all the refactoring and the effort into making this code cleaner. Just 1 more request, I noticed this `if-else` condition is common for all 3 of our rules, maybe consider making this a common function too?

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -226,43 +210,62 @@ def __execute_cmd(cmd):
...
 
     @staticmethod
-    def __execute_firewalld_commands(command, dst_ip, uid):
+    def __execute_firewall_commands(dst_ip, uid, command=APPEND_COMMAND, firewalld_command="", wait=""):
+        # The order in which the below rules are added matters for the ip table rules to work as expected
+
         AddFirewallRules.__raise_if_empty(dst_ip, "Destination IP")
         AddFirewallRules.__raise_if_empty(uid, "User ID")
 
         # Firewalld.service fails if we set `-w` in the iptables command, so not adding it at all for firewalld commands
```
larohra, line 241: This comment is not relevant here anymore, it should be added in the place we fetch the `firewalld` base command (which is divided in the 3 `get_*_rules`

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -179,34 +167,30 @@ def get_accept_tcp_rule(command, destination, firewalld_command="", wait=""):
...
+        cmd = cmd + ['-t', 'security', command, 'OUTPUT', '-d', destination, '-p', 'tcp', '--destination-port', '53',
+                     '-j', 'ACCEPT']
         return cmd
 
     @staticmethod
-    def get_iptables_accept_command(wait, command, destination, owner_uid):
-        cmd = AddFirewallRules.__get_iptables_base_command()
-        cmd.extend(AddFirewallRules.__get_common_accept_command_params(wait, command, destination, owner_uid))
-        return cmd
+    def get_accept_command(command, destination, owner_uid, firewalld_command="", wait=""):
```
larohra, line 175: NIT: `get_accept_command` does not make complete sense here, maybe rename it to `get_wire_root_accept_rule`?
```suggestion
    def get_wire_root_accept_rule(command, destination, owner_uid, firewalld_command="", wait=""):
```

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -179,34 +167,30 @@ def get_accept_tcp_rule(command, destination, firewalld_command="", wait=""):
...
         return cmd
 
     @staticmethod
-    def get_firewalld_accept_command(command, destination, uid, wait=""):
-        cmd = AddFirewallRules.__get_firewalld_base_command(command)
-        cmd.extend(
-            AddFirewallRules.__get_common_accept_command_params(wait, AddFirewallRules.APPEND_COMMAND, destination,
-                                                                uid))
-        return cmd
+    def get_drop_command(command, destination, firewalld_command="", wait=""):
```
larohra, line 186: Similarly here

```suggestion
    def get_wire_non_root_drop_rule(command, destination, firewalld_command="", wait=""):
```

azurelinuxagent/ga/env.py
```diff
@@ -133,6 +134,10 @@ def _operation(self):
 
         success = self._osutil.enable_firewall(dst_ip=self._protocol.get_endpoint(), uid=os.getuid())
 
+        if self._log_initial_firewall_list:
```
larohra, line 137: Do we really need this anymore? The only time where `enable_firewall` rule will not run on service start is if the rules are already present, and in that case, do we need to list the rules? Isin't it enough to just list them whenever the `enable_firewall` actually sets the rules?

tests/common/osutil/test_default.py
```diff
@@ -737,59 +738,97 @@ def test_enable_firewall_should_set_up_the_firewall(self):
...
 
                 success = osutil.DefaultOSUtil().enable_firewall(dst_ip=mock_iptables.destination, uid=mock_iptables.uid)
 
                 self.assertTrue(success, "Enabling the firewall was not successful")
-                # Exactly 4 calls have to be made.
-                # First is the DROP rule check which was mocked to fail, Second, third and forth
-                # to Append the three IPTable rules
-                self.assertEqual(len(mock_iptables.command_calls), 4,
+                # Exactly 8 calls have to be made.
+                # First check rule, delete 4 rules.
```
larohra, line 805: How come delete 4 rules? Shouldn't it be 3? 

Additionally, NIT on the full stop - 

```suggestion
                # First check rule, delete 4 rules,
```

tests/common/test_persist_firewall_rules.py
```diff
@@ -215,6 +221,7 @@ def test_it_should_skip_setup_if_agent_network_setup_service_already_enabled_and
             self.__assert_systemctl_reloaded(validate_command_called=False)
             self.__assert_firewall_cmd_running_called(validate_command_called=True)
             self.__assert_firewall_called(cmd=FirewallCmdDirectCommands.QueryPassThrough, validate_command_called=False)
+            self.__assert_firewall_called(cmd=FirewallCmdDirectCommands.RemovePassThrough, validate_command_called=False)
```
larohra, line 224: Can you add a case where you handle the scenarios on how the agent should react if the `remove_firewall` fails, for both iptables and firewalld. Thanks.

larohra (commented):
some more comments

azurelinuxagent/ga/env.py
nagworld9: Actually, two cases one is service start and second if network setup service already enabled the firewall (Happy path scenario all the rules present all the time). In those cases, if we don't log these, while debugging livesites it's hard to know what rules already setup.

tests/common/osutil/test_default.py
nagworld9: we had legacy rule that was used <=2.2.25

azurelinuxagent/common/utils/networkutil.py
nagworld9: It's always return success if rule exist or not. If we think it may return other command errors or error codes, then it's true for `passthrough `or `query-passthrough`. We may need address for all.

azurelinuxagent/ga/env.py
nagworld9: I would like to send telemetry whenever the `enable_firewall` actually sets the rules. One is to exercise new changes and second is to understand how frequently we are readding these rules and also, figure out how bad the containers hitting missing rules and etc.

larohra: > if we don't log these, while debugging livesites it's hard to know what rules already setup.

If that is the goal, then wont it make more sense to log it periodically rather than just doing it once on service start?

> I would like to send telemetry whenever the enable_firewall actually sets the rules.

Correct me if I'm wrong, but I dont think you're doing that right now in the current code, all I see is logging, no telemetry. And speaking of which, could you add it there lol.

> One is to exercise new changes and second is to understand how frequently we are readding these rules 

Totally with you on the 2nd one but not sure I follow you on the first one, can you elaborate a bit more please

azurelinuxagent/common/utils/networkutil.py
larohra: >  If we think it may return other command errors or error codes

It's not a guessing game, we should verify from the code and the VMs on how the service actually reacts to it :)

> then it's true for passthrough or query-passthrough

Not sure I follow you there. Both those commands have a different use case altogether. `passthrough` just adds iptable rules using firewalld. `query-passthrough` checks if a rule is there or not, if not it exits with a non-0 code and raises and we have exception handling around that code to handle those scenarios. 

Similarly, I want to verify 2 things if we try removing an iptable rule from firewalld service using `remove-passthrough` - 
- Do the rules get removed (adding a test here would help if you can)
- If we try to remove a rule that does not exist, does it fail with a non-0 exit code or not. Because if it does indeed fail with a non-0 exit code, the implementation that you've added will not handle that case

larohra: Might also be worthwhile to add a test case for this case too (in case `remove_firewalld_rules` fails and raises an exception because you haven't added any handling there)

azurelinuxagent/ga/env.py
nagworld9: `If that is the goal, then wont it make more sense to log it periodically rather than just doing it once on service start?`

I feel adding periodically is too much, you end of seeing this in entire log. In general, on agent start you would expect to see firewall rules setup. So, in this you would never see in log after restart. I feel that's really confuses.

` Correct me if I'm wrong, but I dont think you're doing that right now in the current code, all I see is logging, no telemetry. And speaking of which, could you add it there lol.`
I'm not. I'm gonna add now.

`Totally with you on the 2nd one but not sure I follow you on the first one, can you elaborate a bit more please`
I mean to check the code path hitting if we get telemetry.

azurelinuxagent/common/utils/networkutil.py
nagworld9: As I said before, as per my testing `remove-passthrough` returns success regardless rule exist or not. I also removed the rule and restart the agent. Agent didn't raise any sort of errors. 

Same test I added in Dcr too.

azurelinuxagent/ga/env.py
larohra: > I feel adding periodically is too much, you end of seeing this in entire log.

Depends on the periodicity you set no? Say you set it to once a day or something, then it wont be as bad. 

> In general, on agent start you would expect to see firewall rules setup

I mean in a general case, either the `env` thread will set it or the `network-setup.service` will right? In both the cases we have enough logging to tell us when they do. I think the case that you're trying to capture is if by some weird chance we see a scenario where the rules are not set or messed around with then we should have some sort of log. Is that understanding wrong? If that indeed is the case, wont a periodic log for it make more sense? We'll have a periodic check to make sure nobody is messing up our rules, and try to capture the state of it if we do detect a change.

 > So, in this you would never see in log after restart. 

What restart are you talking about? The service or the machine? Because in both cases we would see the extra log right?

nagworld9: `env thread will set it or the network-setup.service will right?`

Yes, we log if `env `thread detect change but not on `network-setup.service`. Since this happens on boot, and we don't have access to agent logs at that time so don't log anything. Obliviously now env thread won't detect any change if network service already did setup.

larohra: we do log it, if the agent detects the service was setup, it fetches the logs from the setup service. We dont log the firewall rules but we log that the firewall rules were added properly. 

This discussion is dragging on longer than it needs to. Essentially all I'm trying to say is that if your goal is to log the firewall rules to check the state of those rules because you don't trust our services setup the rules properly or if there was some external intervention, we should do it periodically than just once on service start (periodically will also capture the first time on service start). Periodically will capture more debug state and would be more useful long term. (Additionally implementing this would be trivial too, just add a new PeriodicOperation on the env thread for logging the rules)

I personally don't understand the need for logging it once on service start especially since in your changes you're checking the availability of each of those rules and if any of them is missing adding them back and subsequently logging it and I think that's enough.

That's my 2 cents on it, I'll let you decide which route you want to take on this.

nagworld9: If you go by env thread detect change for logging, in most of the time you won't hit that scenario. I'm down to period logging once a day to avoid too much logging and also, gives enough debug data overtime. Thanks for the suggestion.

azurelinuxagent/common/utils/networkutil.py
nagworld9: Having said that, at the end of the day those are shell commands and may return exit codes between1-255. So, I handled it now.

tests/common/test_persist_firewall_rules.py
nagworld9: added

azurelinuxagent/common/conf.py
```diff
@@ -170,6 +170,7 @@ def load_conf_from_file(conf_file_path, conf=__conf__):
     "Extensions.InitialGoalStatePeriod": 6,
     "Extensions.GoalStateHistoryCleanupPeriod": 1800,
     "OS.EnableFirewallPeriod": 30,
+    "OS.FirewallRulesLogPeriod": 86400,
```
larohra, line 173: This should be a `Debug` flag as we dont want to introduce this flag to the external customers (primarily because it offers them no benefit)

azurelinuxagent/common/osutil/default.py
```diff
@@ -257,10 +255,12 @@ def remove_legacy_firewall_rule(self, dst_ip):
                 "Unable to remove legacy firewall rule, won't try removing it again. Error: {0}".format(ustr(error)))
 
     def enable_firewall(self, dst_ip, uid):
+        # This is to send telemetry when iptable rules updated
+        is_firewall_rules_updated = False
         # If a previous attempt failed, do not retry
         global _enable_firewall  # pylint: disable=W0603
         if not _enable_firewall:
-            return False
+            return False, is_firewall_rules_updated
```
larohra, line 263: Can you add a description to this function and describe a bit on what the 2 flags signify.

larohra, line 263: Since you've modified the return parameters of this function, please make sure to modify all usages of this to match the new return type. 

1 example: 

```python
def _reset_firewall_rules(osutil):
    """
    Removes MetadataServer firewall rule so IMDS can be used. Enables
    WireServer firewall rule based on if firewall is configured to be on.
    """
    osutil.remove_firewall(dst_ip=_KNOWN_METADATASERVER_IP, uid=os.getuid(), wait=osutil.get_firewall_will_wait())
    if conf.enable_firewall():
        success = osutil.enable_firewall(dst_ip=KNOWN_WIRESERVER_IP, uid=os.getuid())
```

larohra: Also, is the 2nd flag really needed? Can't we just add the telemetry inside the `enable_firewall` function?

azurelinuxagent/common/persist_firewall_rules.py
```diff
@@ -143,16 +143,30 @@ def __verify_firewall_rules_enabled(self):
...
+        try:
+            AddFirewallRules.remove_firewalld_rules(self._dst_ip, self._uid)
+        except Exception as error:
+            logger.verbose(
+                "failed to remove rule using firewalld.service: {0}".format(ustr(error)))
+
+    def __add_firewalld_rules(self):
+        try:
+            AddFirewallRules.add_firewalld_rules(self._dst_ip, self._uid)
+        except Exception as error:
```
larohra, line 156: Why is exception handling needed here? This is kind of breaking the flow of the code. For example, if we're unable to add the rules, you log the error below and handle the exception. Then as soon as you exit this function, this subsequent statement will be logged -  
> logger.info("Successfully added the firewall commands using firewalld.service")

Which is completely wrong since we were infact unable to add the firewall rules. 

Maybe I'm missing something, but could you please explain a bit on why the error handling is needed here?

azurelinuxagent/common/persist_firewall_rules.py
```diff
@@ -143,16 +143,30 @@ def __verify_firewall_rules_enabled(self):
 
         return True
 
+    def __remove_firewalld_rules(self):
+        try:
+            AddFirewallRules.remove_firewalld_rules(self._dst_ip, self._uid)
+        except Exception as error:
+            logger.verbose(
```
larohra, line 150: This should not be verbose, we would want to know about this error.

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -160,34 +160,35 @@ def __get_common_command_params(command, destination):
...
     @staticmethod
-    def get_accept_command(command, destination, owner_uid, firewalld_command="", wait=""):
-        if firewalld_command != "":
-            cmd = AddFirewallRules.__get_firewalld_base_command(firewalld_command)
-        else:
-            cmd = AddFirewallRules.__get_iptables_base_command(wait)
+    def get_wire_root_accept_rule(command, destination, owner_uid, firewalld_command="", wait=""):
+        cmd = AddFirewallRules.__get_firewall_base_command(firewalld_command, wait)
 
         cmd.extend(AddFirewallRules.__get_common_command_params(command, destination))
```
larohra, line 185: The TCP rule uses - `cmd + [-t, ..]` notations and the other rules use `cmd.extend([...])`, maybe make them both consistent for better readability?

azurelinuxagent/ga/env.py
```diff
@@ -147,6 +146,16 @@ def _operation(self):
             log_event=False)
 
 
+class LogFirewallRules(PeriodicOperation):
+    def __init__(self, osutil):
```
larohra, line 150: Can you add a small description of what this Op does and why its needed. Thanks

larohra (changes requested):

azurelinuxagent/common/utils/networkutil.py
larohra: Thanks

azurelinuxagent/common/osutil/default.py
nagworld9: yes, we can but I was skeptical to add since it's utility function.

azurelinuxagent/common/persist_firewall_rules.py
nagworld9: I didn't look at that logger.info. That's good point we should return if we unable to add rules. 

`Maybe I'm missing something, but could you please explain a bit on why the error handling is needed here? `

As per my understanding, these are shell commands, you would expect exit codes between 1-255. In this case you don't see exit code 1 but very likely you can expect others when shell command didn't run properly. Correct me if I'm wrong?

larohra: yes that's correct. But this code was kinda designed that way. Like its supposed to throw if we fail to execute those commands as that defines how the whole sequence of operations work. Does that make sense? If not we can get over it on a call if you want.

nagworld9: I understood the whole sequence and it shouldn't go to next operation and log successful message when exception happens. I was explaining why I decided to do exception handling. I will handle that sequence by return when exception happens.

nagworld9: I just checked the code we have upstream error handling so no need to do it here, thanks

azurelinuxagent/common/osutil/default.py
larohra: The 2 flags seem clunky and add no major purpose except for logging. Additionally, it changes the contract of the OSUtils and might introduce additional errors if all the code paths are not verified and fixed. Could you just move the telemetry inside here? 

> yes, we can but I was skeptical to add since it's utility function.

We're anyways logging here (the multiple `logger.info`) so having telemetry doesn't make much difference

azurelinuxagent/common/osutil/default.py
```diff
@@ -269,50 +255,58 @@ def remove_legacy_firewall_rule(self, dst_ip):
                 "Unable to remove legacy firewall rule, won't try removing it again. Error: {0}".format(ustr(error)))
 
     def enable_firewall(self, dst_ip, uid):
+        """
+        it's checks every iptable rule and add rules if not present. It's return tuple(enable firewall success status, update rules flag)
```
larohra, line 259: NIT: typos - 

```suggestion
        It checks if every iptable rule exists and add them if not present. It returns a tuple(enable firewall success status, update rules flag)
```

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -157,56 +157,41 @@ def __get_firewalld_base_command(command):
...
         return cmd
 
     @staticmethod
-    def get_iptables_drop_command(wait, command, destination):
-        cmd = AddFirewallRules.__get_iptables_base_command()
-        cmd.extend(AddFirewallRules.__get_common_drop_command_params(wait, command, destination))
+    def get_accept_tcp_rule(command, destination, firewalld_command="", wait=""):
+        # This rule allows DNS TCP request to wireserver ip for non root users
+        cmd = AddFirewallRules.__get_firewall_base_command(firewalld_command, wait)
+
```
larohra, line 176: NIT: extra line

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -157,56 +157,41 @@ def __get_firewalld_base_command(command):
...
+        cmd.extend(AddFirewallRules.__get_common_command_params(command, destination))
```
larohra, line 177: NIT: This call is common for all 3 rules, why not move it to the common function `__get_firewall_base_command`?

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -157,56 +157,41 @@ def __get_firewalld_base_command(command):
...
 
     @staticmethod
-    def get_firewalld_accept_command(command, destination, uid, wait=""):
-        cmd = AddFirewallRules.__get_firewalld_base_command(command)
-        cmd.extend(
-            AddFirewallRules.__get_common_accept_command_params(wait, AddFirewallRules.APPEND_COMMAND, destination,
-                                                                uid))
+    def get_wire_root_accept_rule(command, destination, owner_uid, firewalld_command="", wait=""):
+        cmd = AddFirewallRules.__get_firewall_base_command(firewalld_command, wait)
+
```
larohra, line 184: NIT extra line

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -157,56 +157,41 @@ def __get_firewalld_base_command(command):
...
         return cmd
 
     @staticmethod
-    def get_firewalld_drop_command(command, destination, wait=""):
-        cmd = AddFirewallRules.__get_firewalld_base_command(command)
-        cmd.extend(
-            AddFirewallRules.__get_common_drop_command_params(wait, AddFirewallRules.APPEND_COMMAND, destination))
+    def get_wire_non_root_drop_rule(command, destination, firewalld_command="", wait=""):
+        cmd = AddFirewallRules.__get_firewall_base_command(firewalld_command, wait)
+
```
larohra, line 192: NIT extra line

azurelinuxagent/common/osutil/default.py
nagworld9: `We're anyways logging here (the multiple logger.info) so having telemetry doesn't make much difference`

Adding telemetry in osutil created circular import error. I guess I need to maintain two flags. Please let me know your thoughts, I don't want to undo and redo forever :)

larohra: what's the circular import on? Why not just move the offending import to a separate common utility and fix the circular imports?

azurelinuxagent/common/osutil/default.py
```diff
@@ -269,50 +255,58 @@ def remove_legacy_firewall_rule(self, dst_ip):
                 "Unable to remove legacy firewall rule, won't try removing it again. Error: {0}".format(ustr(error)))
 
     def enable_firewall(self, dst_ip, uid):
+        """
+        It checks if every iptable rule exists and add them if not present. It returns a tuple(enable firewall success status, update rules flag)
+        enable firewall success status: Returns true if every rule check completed successfully otherwise false
```
larohra, line 260: NIT: maybe change to - 
```suggestion
        enable firewall success status: Returns True if every firewall rule exists otherwise False
```

azurelinuxagent/common/osutil/default.py
```diff
@@ -269,50 +255,58 @@ def remove_legacy_firewall_rule(self, dst_ip):
...
-                drop_rule = get_firewall_drop_command(wait, AddFirewallRules.CHECK_COMMAND, dst_ip)
-                shellutil.run_command(drop_rule)
-                logger.verbose("Firewall appears established")
-                return True
+                if not AddFirewallRules.verify_iptables_rules_exist(wait, dst_ip, uid):
+                    self.remove_firewall(dst_ip, uid, wait)
+                    AddFirewallRules.add_iptables_rules(wait, dst_ip, uid)
+                    is_firewall_rules_updated = True
+                    logger.info("Successfully added Azure fabric firewall rules")
+                    logger.info("Firewall rules:\n{0}".format(self.get_firewall_list(wait)))
```
larohra, line 281: Maybe send this as part of the telemetry too?

larohra (dismissed):
Minor comments else LGTM

larohra (approved):

kevinclark19a (approved):

PR #2486, opened 2022-01-25
Move fetch_vm_settings() to HostPluginProtocol

narrieta:
Some code cleanup. Moved the logic and data structures for the vmSettings request from the WireClient to the HostPluginProtocol.

### DIFF
--- a/azurelinuxagent/common/protocol/hostplugin.py
+++ b/azurelinuxagent/common/protocol/hostplugin.py
@@ -554,9 +554,7 @@ class _VmSettingsErrorReporter(object):
                 "timeouts":       self._timeout_count,
                 "failedRequests": self._request_failure_count
             }
-            # always send telemetry, but log errors only
-            message = json.dumps(summary)
-            add_event(op=WALAEventOperation.VmSettingsSummary, message=message, is_success=False, log_event=False)
+            add_event(op=WALAEventOperation.VmSettingsSummary, message=json.dumps(summary), is_success=False, log_event=False)
             if self._error_count > 0:
                 logger.info("[VmSettingsSummary] {0}", message)
 
--- a/azurelinuxagent/common/protocol/hostplugin.py
+++ b/azurelinuxagent/common/protocol/hostplugin.py
@@ -554,7 +554,8 @@ class _VmSettingsErrorReporter(object):
                 "timeouts":       self._timeout_count,
                 "failedRequests": self._request_failure_count
             }
-            add_event(op=WALAEventOperation.VmSettingsSummary, message=json.dumps(summary), is_success=False, log_event=False)
+            message = json.dumps(summary)
+            add_event(op=WALAEventOperation.VmSettingsSummary, message=message, is_success=False, log_event=False)
             if self._error_count > 0:
                 logger.info("[VmSettingsSummary] {0}", message)
 
--- a/azurelinuxagent/common/protocol/hostplugin.py
+++ b/azurelinuxagent/common/protocol/hostplugin.py
@@ -25,9 +25,11 @@ import uuid
 from azurelinuxagent.common import logger
 from azurelinuxagent.common.errorstate import ErrorState, ERROR_STATE_HOST_PLUGIN_FAILURE
 from azurelinuxagent.common.event import WALAEventOperation, add_event
-from azurelinuxagent.common.exception import HttpError, ProtocolError
-from azurelinuxagent.common.future import ustr
+from azurelinuxagent.common.exception import HttpError, ProtocolError, ResourceGoneError
+from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
+from azurelinuxagent.common.future import ustr, httpclient
 from azurelinuxagent.common.protocol.healthservice import HealthService
+from azurelinuxagent.common.protocol.extensions_goal_state_factory import ExtensionsGoalStateFactory
 from azurelinuxagent.common.utils import restutil
 from azurelinuxagent.common.utils import textutil
 from azurelinuxagent.common.utils.textutil import remove_bom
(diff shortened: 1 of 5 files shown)


azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -383,3 +390,174 @@ def _base64_encode(self, data):
         if PY_VERSION_MAJOR > 2:
             return s.decode('utf-8')
         return s
+
```
narrieta, line 393: Move from the WireClient, the only difference in the implementation is the use of ResourceGoneError to report 410s (added the exception here because the retry logic needs to be implemented by the caller)

tests/protocol/test_hostplugin.py
```diff
@@ -862,6 +862,125 @@ def test_should_report(self):
         self.assertEqual(False, actual)
 
 
+class TestHostPluginVmSettings(HttpRequestPredicates, AgentTestCase):
```
narrieta, line 865: Move from test_wire. The implementation is almost identical to the original, but instead of testing via  WireClient.update_goal_state() now we test via HostPluginProtocol.fetch_vm_settings()

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -383,3 +390,174 @@ def _base64_encode(self, data):
...
+    def fetch_vm_settings(self, force_update):
+        """
+        Queries the vmSettings from the HostGAPlugin and returns an (ExtensionsGoalStateFromVmSettings, bool) tuple with the vmSettings and
+        a boolean indicating if they are an updated (True) or a cached value (False).
```
larohra, line 397: NIT: typo - 

```suggestion
        a boolean indicating if they are an update (True) or a cached value (False).
```

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -383,3 +390,174 @@ def _base64_encode(self, data):
...
+        if datetime.datetime.now() >= self._next_period:
+            summary = {
+                "requests":       self._request_count,
+                "errors":         self._error_count,
+                "serverErrors":   self._server_error_count,
+                "clientErrors":   self._client_error_count,
+                "timeouts":       self._timeout_count,
+                "failedRequests": self._request_failure_count
+            }
+            # always send telemetry, but log errors only
```
larohra, line 557: NIT: typo - 

```suggestion
            # always send telemetry, but log only on errors 
```

larohra (approved):
minor comments else LGTM

azurelinuxagent/common/protocol/hostplugin.py
narrieta: "updated" or a "cached" value seems correct

azurelinuxagent/common/protocol/hostplugin.py
narrieta: thanks; actually a leftover comment (I removed the log statement previously)

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -383,3 +390,173 @@ def _base64_encode(self, data):
         if PY_VERSION_MAJOR > 2:
             return s.decode('utf-8')
         return s
+
+    def fetch_vm_settings(self, force_update):
+        """
+        Queries the vmSettings from the HostGAPlugin and returns an (ExtensionsGoalStateFromVmSettings, bool) tuple with the vmSettings and
+        a boolean indicating if they are an updated (True) or a cached value (False).
+
+        Raises TypeError if the HostGAPlugin does not support the vmSettings API, ResourceGoneError if the container ID and roleconfig name
```
nagworld9, line 399: NIT: 
```suggestion
        Raises VmSettingsNotSupported if the HostGAPlugin does not support the vmSettings API, ResourceGoneError if the container ID and roleconfig name
```

azurelinuxagent/common/protocol/wire.py
```diff
@@ -799,18 +791,23 @@ def update_goal_state(self, force_update=False):
...
+            host_ga_plugin.update_container_id(goal_state.container_id)
+            host_ga_plugin.update_role_config_name(goal_state.role_config_name)
 
             #
             # Then we fetch the vmSettings from the HostGAPlugin; the response will include the goal state for extensions.
             #
-            vm_settings_goal_state, vm_settings_goal_state_updated = (None, False)
+            vm_settings, vm_settings_updated = (None, False)
 
             if conf.get_enable_fast_track():
```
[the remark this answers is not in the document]
larohra: In that case, how are we going to handle it if the customer wants to opt-out of fast-track, will we just avoid using vmSettings altogether? In short we're considering vmSettings tantamount to fast-track?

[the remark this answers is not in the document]
narrieta: Yes, if fast track is disabled we will simply use extensionsConfig. That is already implemented in this PR. Yes, vmSettings is fundamental for FastTrack, no fast-track => do not use vmsettings, no vmsettings => do not use fast track.

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -383,3 +390,173 @@ def _base64_encode(self, data):
...
+                else:
+                    self._vm_settings_error_reporter.report_error(error_description)
+
+                raise ProtocolError(error_description)
+
+            for h in response.getheaders():
+                if h[0].lower() == 'etag':
+                    response_etag = h[1]
+                    break
+            else:  # since the vmSettings were updated, the response must include an etag
```
nagworld9, line 463: confused here, is this check supposed to be inside `for` loop?

narrieta: a python thing: the else will be executed if we don't break from the loop

azurelinuxagent/common/protocol/hostplugin.py
narrieta: thanks, will fix it

azurelinuxagent/common/protocol/hostplugin.py
nagworld9: ohh, I didn't know this.

nagworld9 (approved):

PR #2499, opened 2022-02-09
Use ExtensionsGoalState.id when checking for changes in extensions

narrieta:
Once Fast Track is enabled we cannot use the incarnation to check for changes in the extensions goal state. Using 'id' instead (this abstracts the incarnation/etag used by WireServer/HostGAPlugin).

### DIFF
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -166,7 +166,15 @@ class UpdateHandler(object):
         self._last_try_update_goal_state_failed = False
         self._report_status_last_failed_incarnation = -1
 
-        self.last_incarnation = None
+        # incarnation of the last goal state that has been fully processed
+        # (None if no goal state has been processed)
+        self._last_incarnation = None
+        # ID of the last extensions goal state that has been fully processed (incarnation for WireServer goal states or etag for HostGAPlugin goal states)
+        # (None if no extensions goal state has been processed)
+        self._last_extensions_gs_id = None
+        # Goal state that is currently been processed
+        # (None if no goal state is being processed)
+        self._goal_state = None
 
         self._extensions_summary = ExtensionsSummary()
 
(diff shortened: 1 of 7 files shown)


tests/ga/test_update.py
```diff
@@ -2482,6 +2483,17 @@ def __init__(self, return_value=0, side_effect=None):
         self.wait = Mock(return_value=return_value, side_effect=side_effect)
 
 
+class ExtensionsGoalStateMock(object):
```
narrieta, line 2486: These tests are using deep internals of the UpdateHandler object; I patched them for now, but eventually they should be updated to use the mock_protocol utilities.

azurelinuxagent/ga/update.py
```diff
@@ -540,30 +551,30 @@ def handle_updates_for_requested_version():
...
-            return
+            # update self._goal_state
+            self._try_update_goal_state(protocol)
 
-        # Update the Guest Agent if a new version is available
-        self.__update_guest_agent(protocol)
-        incarnation = protocol.get_incarnation()
+            # Update the Guest Agent if a new version is available
+            if self._goal_state is not None:
+                self.__update_guest_agent(protocol)
```
nagworld9, line 575: Not having `return` on try_update_goal_state failure case, changes the existing behavior of update guest agent.

Example previous goal state received and updated `self._goal_state is not None` which processed update guest agent. All good at this point. On a current goal state, we try to do `try_update_goal_state ` which is failed and self._goal_state is not updated and will have previous goal sate. So now the `update guest agent `processed on previous goal state information which we already done as part of previous goal state. Do we need to run update guest agent without incarnation changes?

narrieta: yes, we need to agent update and status report even if no incarnation change; the previous code was skipping agent update so this change was intentional

nagworld9 (approved):

PR #2503, opened 2022-02-11
Save vmSettings on parse errors; improve messages in parse errors

narrieta:
In the previous commit we stopped logging the vmsettings on errors, since some of the errors are persistent and saving all the details can make the log grow too large.

This PR improves the error messages of parsing errors, and ensures the vmSettings are saved to the history when a parse error occurs.

Also, this handles an empty dependsOn as setting a dependency level of 0 (this matches the behavior of extensionsConfig)


tests/common/test_exception.py
```diff
@@ -1,46 +0,0 @@
-# Copyright 2019 Microsoft Corporation
```
narrieta, line 1: These tests enforce a contract that prevents adding parameters to AgentError exceptions, and they provide little value so I removed them.

nagworld9:
LGTM

PR #2508, opened 2022-02-14
Cleanup history directory after every goal state

narrieta:
There was a race condition between the env thread cleaning up the history folder and the GoalState writing to it.

Cleaning up the history folder on a separate thread was done just for legacy reasons, but  there is no good motivation to keep it that way. I moved it to the main thread, cleaning up after each goal state.


README.md
```diff
@@ -243,15 +242,6 @@ _Note_: setting up this parameter to more than a few minutes can make the state
 the VM be reported as unresponsive/unavailable on the Azure portal. Also, this 
 setting affects how fast the agent starts executing extensions. 
 
-#### __Extensions.GoalStateHistoryCleanupPeriod__
```
narrieta, line 246: cleaning up the history is no longer a periodic operation

azurelinuxagent/common/utils/archive.py
```diff
@@ -44,7 +44,10 @@
     re.compile(r"^VmSettings.\d+\.json$"),
     re.compile(r"^(.*)\.(\d+)\.(agentsManifest)$", re.IGNORECASE),
     re.compile(r"^(.*)\.(\d+)\.(manifest\.xml)$", re.IGNORECASE),
-    re.compile(r"^(.*)\.(\d+)\.(xml)$", re.IGNORECASE)
+    re.compile(r"^(.*)\.(\d+)\.(xml)$", re.IGNORECASE),
```
narrieta, line 47: I  missed those files on a previous PR

azurelinuxagent/common/utils/archive.py
```diff
@@ -161,9 +164,11 @@ def purge(self):
         for state in states[_MAX_ARCHIVED_STATES:]:
             state.delete()
 
-        # legacy history files
-        for current_file in os.listdir(self._source):
-            full_path = os.path.join(self._source, current_file)
+    @staticmethod
```
narrieta, line 167: separated this into its own function. which is called before the main loop

azurelinuxagent/ga/update.py
```diff
@@ -590,11 +593,31 @@ def _process_goal_state(self, exthandlers_handler, remote_access_handler):
...
+                    logger.warn("Error cleaning up the goal state history: {0}", ustr(exception))
+
         finally:
             if self._goal_state is not None:
                 self._last_incarnation = self._goal_state.incarnation
                 self._last_extensions_gs_id = self._goal_state.extensions_goal_state.id
 
+    @staticmethod
+    def _cleanup_goal_state_history():
+        archiver = StateArchiver(conf.get_lib_dir())
```
kevinclark19a, line 610: This is only happening on an updated goal state, right? i.e. not every six seconds? 

Was initially a bit weary of creating a new StateArchiver every go around the loop, but I think this should be fine on the relatively sparse event of a new extensions goal state.

kevinclark19a (approved):

azurelinuxagent/ga/update.py
narrieta: the call is at line 599, protected by if self._processing_new_extensions_goal_state():

narrieta: (only on new goal state for extensions)

tests/utils/test_archive.py
```diff
@@ -136,6 +129,27 @@ def test_archive02(self):
...
+                'GoalState.1.xml',
+                'VmSettings.1.json',
+                'Prod.1.manifest.xml',
+                'ExtensionsConfig.1.xml',
+                'Microsoft.Azure.Extensions.CustomScript.1.xml',
+                'SharedConfig.xml',
+                'HostingEnvironmentConfig.xml',
+                'RemoteAccess.xml'
+            ]
+            legacy_files = [os.path.join(self.tmp_dir, file) for file in legacy_files]
```
kevinclark19a, line 144: Looks like using the variable `file` is tripping up pylint on python2.7. Wanted to save you having to look for it.

narrieta: thanks; i fixed the instances below this one, will fix it too

tests/utils/test_archive.py
```diff
@@ -53,43 +53,36 @@ def _parse_archive_name(name):
...
+            for current_file in test_files:
+                self._write_file(os.path.join(directory, current_file))
+            test_directories.append(directory)
 
         test_subject = StateArchiver(self.tmp_dir)
         test_subject.archive()
 
-        timestamp_zips = os.listdir(self.history_dir)
-        self.assertEqual(1, len(timestamp_zips))
+        for directory in test_directories[0:1]:
```
nagworld9, line 76: This index only checks one directory, so could change to test second one too?

tests/utils/test_archive.py
```diff
@@ -136,6 +129,27 @@ def test_archive02(self):
                 filename = "{0}_0.zip".format(timestamp)
             self.assertTrue(filename in archived_entries, "'{0}' is not in the list of unpurged entires".format(filename))
 
+    def test_purge_legacy_goal_state_history(self):
+        with patch("azurelinuxagent.common.conf.get_lib_dir", return_value=self.tmp_dir):
```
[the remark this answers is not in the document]
narrieta: somehow that patch was is not working here, so i went ahead and duplicated it

tests/utils/test_archive.py
narrieta: this checks items 0 & 1, item 2 is checked at line 85

kevinclark19a (dismissed):

tests/utils/test_archive.py
narrieta: fixed

kevinclark19a (approved):

nagworld9 (approved):

PR #2510, opened 2022-02-15
Handle OOM errors by stopping the periodic log collector

kevinclark19a:
## Description

A bug exists in the current code base where specific circumstances lead to the periodic log collector getting consistently OOM killed. This seems to not be recoverable in the current code base, so this PR introduces logic to turn off the log collector (until service restart) the first time this occurs.

---

### PR information


azurelinuxagent/common/logcollector.py
```diff
@@ -45,6 +45,7 @@
 
 CGROUPS_UNIT = "collect-logs.scope"
 
+OOM_KILLED_ERRCODE = -9
```
narrieta, line 48: I would not make this specific to OOM... -9 it is just a generic force kill. let's rename this and add comments that it being used to attempt to capture the termination by OOM

kevinclark19a: Updated. I added the comments where this value is used, thought that fit better.

narrieta (approved):

azurelinuxagent/ga/collect_logs.py
```diff
@@ -203,6 +203,12 @@ def exec_command(output_file):
                     if e.returncode == logcollector.INVALID_CGROUPS_ERRCODE: # pylint: disable=no-member
                         logger.info("Disabling periodic log collection until service restart due to process error.")
                         self.stop()
+                    
+                    # When the OOM killer is invoked on the log collector process, this error code is
+                    # returned. Stop the periodic operation because it seems to be persistent.
+                    elif e.returncode == logcollector.FORCE_KILLED_ERRCODE: # pylint: disable=no-member
+                        logger.info("Disabling periodic log collection until service restart due OOM error.")
```
nagworld9, line 210: "service restart due _to_ OOM error"?

nagworld9 (dismissed):

azurelinuxagent/ga/collect_logs.py
kevinclark19a: Yep, good catch

nagworld9 (approved):

PR #2513, opened 2022-02-16
Stop UTs from attempting to call to IMDS

kevinclark19a:
## Description

A bug was introduced in #2462 that caused UTs ran on non-Azure VM systems to hang while attempting to talk to IMDS.

This PR fixes this and stops the UTs from hanging.

---

### PR information


tests/ga/test_exthandlers_exthandlerinstance.py
```diff
@@ -29,6 +29,7 @@ def setUp(self):
 
     def tearDown(self):
         self.mock_get_base_dir.stop()
+        super(ExtHandlerInstanceTestCase, self).tearDown()
```
kevinclark19a, line 32: There were a couple of these where a subclass overrides the base class (AgentTestCase) and don't call into it. I added the call.

tests/tools.py
```diff
@@ -182,10 +182,16 @@ def setUp(self):
         event.init_event_status(self.tmp_dir)
         event.init_event_logger(self.tmp_dir)
 
+        self._imds_endpoint_patch = patch('azurelinuxagent.common.protocol.imds.IMDS_ENDPOINT', "NotAnIpAddr")
```
kevinclark19a, line 185: Putting a non-IP value here fixes the hanging.

tests/utils/test_archive.py
```diff
@@ -23,14 +22,11 @@
 
 class TestArchive(AgentTestCase):
     def setUp(self):
+        super(TestArchive, self).setUp()
```
kevinclark19a, line 25: Similar to above, need to add the call to setUp.

tests/utils/test_extension_process_util.py
```diff
@@ -39,8 +39,8 @@ def setUp(self):
     def tearDown(self):
         self.stderr.close()
         self.stdout.close()
-        if self.tmp_dir is not None:
-            shutil.rmtree(self.tmp_dir)
+        
+        super(TestProcessUtils, self).tearDown()
```
kevinclark19a, line 43: This was just code duplication, so I cleaned it up.

tests/protocol/mockwiredata.py
```diff
@@ -239,6 +244,8 @@ def mock_http_get(self, url, *_, **kwargs):
                 content = self.vm_settings
                 response_headers = [('ETag', self.etag)]
             self.call_counts["vm_settings"] += 1
+        elif '{0}/metadata/compute'.format(IMDS_ENDPOINT) in url:
+            content = json.dumps(self.imds_info.get("compute", "{}"))
```
kevinclark19a, line 248: Right now (perhaps always? not sure if this will ever change), this is the only type of IMDS call. Behavior if a new type of call is added to IMDS is to throw a `NotImplementedError`.

tests/ga/test_update.py
```diff
@@ -2869,7 +2814,65 @@ def test_it_should_process_goal_state_only_on_new_goal_state(self):
...
+            update_handler = get_update_handler()
+            
+            update_handler.last_telemetry_heartbeat = datetime.utcnow() - timedelta(hours=1)
+            update_handler._send_heartbeat_telemetry(mock_protocol)
+            self.assertEqual(1, patch_add_event.call_count)
+            self.assertTrue(any(call_args[0] == "[HEARTBEAT] Agent {0} is running as the goal state agent {1}"
+                            for call_args in patch_info.call_args), "The heartbeat was not written to the agent's log")
+    
+    @patch("azurelinuxagent.ga.update.add_event")
+    @patch("azurelinuxagent.common.protocol.imds.ImdsClient")
```
kevinclark19a, line 2833: This patch is higher level than the patch in `mock_wire_protocol`; we never actually call in to the `http_get` handler (for imds calls) in this test.

tests/ga/test_update.py
```diff
@@ -2869,7 +2814,65 @@ def test_it_should_process_goal_state_only_on_new_goal_state(self):
...
 
 
-class ReportStatusTestCase(AgentTestCase):
+class HeartbeatTestCase(AgentTestCase):
+
+    @patch("azurelinuxagent.common.logger.info")
+    @patch("azurelinuxagent.ga.update.add_event")
+    def test_telemetry_heartbeat_creates_event(self, patch_add_event, patch_info, *_):
+        
+        with mock_wire_protocol(mockwiredata.DATA_FILE) as mock_protocol:
```
kevinclark19a, line 2823: This test originally used `mock_protocol = WireProtocol('foo.bar')`, but I wanted to test the hook for IMDS network calls anyway, so I switched it to using the newer way of doing things.

tests/ga/test_update.py
```diff
@@ -156,13 +156,13 @@ def check_running(*val, **__):
...
 
-class UpdateTestCase(AgentTestCase):
+class UpdateTestCase(AgentTestCaseWithGetVmSizeMock):
     _test_suite_tmp_dir = None
     _agent_zip_dir = None
 
     @classmethod
     def setUpClass(cls):
-        AgentTestCase.setUpClass()
+        super(UpdateTestCase, cls).setUpClass()
```
kevinclark19a, line 165: There are a couple of these scattered throughout the code base, but I've only updated the ones that would be broken by the new intermediate base class.

tests/protocol/mockwiredata.py
```diff
@@ -36,6 +38,7 @@
         "trans_prv": "wire/trans_prv",
         "trans_cert": "wire/trans_cert",
         "test_ext": "ext/sample_ext-1.3.0.zip",
+        "imds_info": "imds/valid.json",
```
kevinclark19a, line 41: This file already existed, but a cursory search didn't show it being used in the codebase... There's also `imds/unicode.json` that seems to be identical.

tests/protocol/mockwiredata.py
```diff
@@ -180,6 +184,7 @@ def reload(self):
         self.ga_manifest = load_data(self.data_files.get("ga_manifest"))
         self.trans_prv = load_data(self.data_files.get("trans_prv"))
         self.trans_cert = load_data(self.data_files.get("trans_cert"))
+        self.imds_info = json.loads(load_data(self.data_files.get("imds_info")))
```
kevinclark19a, line 187: There's an added dependency on the `json` module here because the json files (i.e. `valid.json` and `unicode.json`) include data returned by both `compute` and `network`.

tests/protocol/mockwiredata.py
narrieta: we should probably return valid data, instead of empty

tests/protocol/mockwiredata.py
kevinclark19a: Huh, looks like it actually is in use; I had searched for `unicode.json`, but we're actually constructing the filename at runtime (i.e. `"{0}.json".format("unicode")` so I didn't see it at first.

tests/protocol/mockwiredata.py
narrieta: never mind, i misread the code

nagworld9 (approved):

PR #2517, opened 2022-02-22
Add mariner 1.0 to DCR v2

kevinclark19a:
## Description

This PR adds end-to-end testing for the mariner 1.0 image on the marketplace.

---

### PR information


dcr/scenarios/agent-bvt/check_firewall.py
```diff
@@ -24,8 +24,8 @@ def __is_firewall_enabled():
                 if update_match:
                     return update_match.groups()[0].lower() == 'y'
 
-    # The firewall is enabled by default.
-    return True
+    # The firewall is disabled by default.
+    return False
```
kevinclark19a, line 28: As seen with mariner 1.0, when there is no entry in the config file, the agent disables the firewall.

dcr/scenarios/agent-bvt/test_agent_basics.py
```diff
@@ -48,7 +48,7 @@ def check_root_login():
     print(root_passwd_line)
     root_passwd = root_passwd_line.split(":")[1]
 
-    if "!" in root_passwd or "*" in root_passwd:
+    if any(val in root_passwd for val in ("!", "*", "x")):
```
kevinclark19a, line 51: Mariner 1.0 uses a different syntax to mean the same thing; `x` is not the result of a hash and therefore root password is disabled. Maybe we want to have a less explicit check here, so that we don't need to keep updating this code? Something to the effect of checking the length of the entry _might_ work (i.e. anything a single character long is not a hash), but I'm not sure. Thoughts?

dcr/templates/deploy-linux-vm.json
```diff
@@ -234,7 +234,8 @@
             "createOption": "FromImage",
             "managedDisk": {
               "storageAccountType": "[variables('osDiskType')]"
-            }
+            },
+            "diskSizeGB": 32
```
kevinclark19a, line 238: Mariner 1.0 specifies a minimum OS disk as 1GB, which causes our testing to crash with "no space left on disk" if we leave this property blank. This approach does lock in all other distros to the same size; i.e., I believe ubuntu lets you create an 16GB OS disk, so we'd be doubling that size with this. 32GB is the largest minimum of the distros we test, so I went with that.

dcr/scenarios/agent-bvt/test_agent_basics.py
narrieta: why do we need this test at all? what is its equivalent in DCR v1?

dcr/scripts/test-vm/harvest.sh
```diff
@@ -11,6 +11,8 @@
 set -euxo pipefail
 
 ssh -o "StrictHostKeyChecking no" "$1"@"$2" "sudo tar --exclude='journal/*' --exclude='omsbundle' --exclude='omsagent' --exclude='mdsd' --exclude='scx*' --exclude='*.so' --exclude='*__LinuxDiagnostic__*' --exclude='*.zip' --exclude='*.deb' --exclude='*.rpm' -czf logs-$2.tgz /var/log /var/lib/waagent/ /etc/waagent.conf"
+# Some distros do not have "other" permissions (e.g., mariner1.0), add them here.
+ssh -o "StrictHostKeyChecking no" "$1"@"$2" "sudo chmod o+rwx logs-$2.tgz"
```
narrieta, line 15: why do we need the "other" permission

narrieta:
@kevinclark19a can you post a link to a sample run?

dcr/scenarios/agent-bvt/test_agent_basics.py
kevinclark19a: The equivalent is: https://msazure.visualstudio.com/DefaultCollection/One/_git/Compute-Runtime-Tux-Pipeline?path=%2Fdungeon_crawler%2Fscenarios%2Fagent-bvt%2F05_check_root_login.py&_a=contents&version=GBdevelop

narrieta: My first impression is that we can drop this test. Before doing so, though, could you check if the agent disables root login (maybe during provisioning)?

kevinclark19a: Sure, will do

kevinclark19a:
@narrieta, here's [the sample run](https://dev.azure.com/cplatruntime/WALinuxAgent/_build/results?buildId=985&view=ms.vss-test-web.build-test-results-tab). It has the ETP failures that are on develop.

narrieta:
@kevinclark19a run looks good, thanks

nagworld9 (approved):

dcr/scripts/test-vm/harvest.sh
narrieta: we are creating the file using sudo, so the owner would be root... maybe the call to harvest is missing "sudo"?

kevinclark19a: Yep, I think you're right; the difference is that mariner defaults new files to more restrictive permissions than other distros. (The call to copy the tar over is the last line of this file).

The equivalent of adding `sudo` to the harvest command would be using `root` as the user instead of `dcr` in the `scp` command. I think using the root user directly would be less ideal than this `chmod`, though.

narrieta: ah, ok, then $1 is the user... yes, probably the chown would be slightly better than a blanket all permissions to all users

dcr/scripts/test-vm/harvest.sh
```diff
@@ -12,7 +12,7 @@ set -euxo pipefail
 
 ssh -o "StrictHostKeyChecking no" "$1"@"$2" "sudo tar --exclude='journal/*' --exclude='omsbundle' --exclude='omsagent' --exclude='mdsd' --exclude='scx*' --exclude='*.so' --exclude='*__LinuxDiagnostic__*' --exclude='*.zip' --exclude='*.deb' --exclude='*.rpm' -czf logs-$2.tgz /var/log /var/lib/waagent/ /etc/waagent.conf"
 # Some distros do not have "other" permissions (e.g., mariner1.0), add them here.
-ssh -o "StrictHostKeyChecking no" "$1"@"$2" "sudo chmod o+rwx logs-$2.tgz"
+ssh -o "StrictHostKeyChecking no" "$1"@"$2" "sudo chown $1 logs-$2.tgz"
```
narrieta, line 15: comment on previous line needs updating

kevinclark19a: good catch!

narrieta (approved):

nagworld9 (approved):

PR #2522, opened 2022-03-01
Improve logic of unexpected processes check started by agent

nagworld9:
## Description

 The current logic of finding parent and if parent is agent, then we don't consider it as unexpected process and don't disable cgroups. That is broken if the process gets orphaned on certain scenarios after it's started by agent.
 
Design:
we should explicitly add an env variable to the processes we create and change the cgroupconfigurator to look for that var.

---

### PR information

### DIFF
--- a/azurelinuxagent/common/cgroupconfigurator.py
+++ b/azurelinuxagent/common/cgroupconfigurator.py
@@ -605,7 +605,8 @@ class CGroupConfigurator(object):
                     current = process
                     while current != 0 and current not in agent_commands:
                         current = self._get_parent(current)
-                    if current == 0:
+                    # Process started by agent will have a marker and check if that marker found in process environment.
+                    if current == 0 and not self.__is_process_env_flag_found(process):
                         unexpected.append(self.__format_process(process))
                         if len(unexpected) >= 5:  # collect just a small sample
                             break
@@ -639,6 +640,24 @@ class CGroupConfigurator(object):
             except Exception:
                 pass
             return "[PID: {0}] UNKNOWN".format(pid)
+
+        @staticmethod
+        def __is_process_env_flag_found(pid):
+            """
+            Returns True if the process env flag(PARENT_NAME) found otherwise False
+            """
+            try:
+                env = '/proc/{0}/environ'.format(pid)
+                if os.path.exists(env):
+                    with open(env, "r") as env_file:
+                        environ = env_file.read()
+                        if environ and environ[-1] == '\x00':
+                            environ = environ[:-1]
+                        match = re.search(shellutil.PARENT_PROCESS_NAME + "=" + shellutil.AZURE_GUEST_AGENT, environ)
+                        return match is not None
+            except Exception:
+                pass
+            return False
 
         @staticmethod
         def _check_agent_throttled_time(cgroup_metrics):
--- a/azurelinuxagent/common/cgroupconfigurator.py
+++ b/azurelinuxagent/common/cgroupconfigurator.py
@@ -654,8 +654,7 @@ class CGroupConfigurator(object):
                         environ = env_file.read()
                         if environ and environ[-1] == '\x00':
                             environ = environ[:-1]
-                        match = re.search(shellutil.PARENT_PROCESS_NAME + "=" + shellutil.AZURE_GUEST_AGENT, environ)
-                        return match is not None
+                        return "{0}={1}".format(shellutil.PARENT_PROCESS_NAME, shellutil.AZURE_GUEST_AGENT) in environ
             except Exception:
                 pass
             return False
--- a/azurelinuxagent/common/cgroupconfigurator.py
+++ b/azurelinuxagent/common/cgroupconfigurator.py
@@ -606,7 +606,7 @@ class CGroupConfigurator(object):
                     while current != 0 and current not in agent_commands:
                         current = self._get_parent(current)
                     # Process started by agent will have a marker and check if that marker found in process environment.
-                    if current == 0 and not self.__is_process_env_flag_found(process):
+                    if current == 0 and not self.__is_process_descendant_of_the_agent(process):
                         unexpected.append(self.__format_process(process))
                         if len(unexpected) >= 5:  # collect just a small sample
                             break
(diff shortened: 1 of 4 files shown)


tests/ga/test_multi_config_extension.py
```diff
@@ -738,7 +738,7 @@ def __assert_env_variables(handler_name, handler_version="1.0.0", seq_no="1", ex
                     self.assertFalse(any(env_var in commands['data'] for env_var in not_expected), "Unwanted env variable found")
 
         def mock_popen(cmd, *_, **kwargs):
-            if 'env' in kwargs:
+            if 'env' in kwargs and ExtCommandEnvVariable.ExtensionVersion in kwargs['env']:
```
nagworld9, line 741: This cgroupsapi mocking all other popen calls which breaking for non-extension commands. So, I added this check to call extension emulator on extension operations.

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -605,7 +605,8 @@ def _check_processes_in_agent_cgroup(self):
                     current = process
                     while current != 0 and current not in agent_commands:
                         current = self._get_parent(current)
-                    if current == 0:
+                    # Process started by agent will have a marker and check if that marker found in process environment.
+                    if current == 0 and not self.__is_process_env_flag_found(process):
```
nagworld9, line 609: Added as alternative condition If all the above checks unsuccessful.

azurelinuxagent/common/utils/shellutil.py
```diff
@@ -345,10 +345,23 @@ def quote(word_list):
...
+        # Add the environment variables
+        env = {}
+        if 'env' in kwargs:
+            env.update(kwargs['env'])
+        else:
+            env.update(os.environ)
+
+        # Set the marker before process start
+        env[PARENT_PROCESS_NAME] = AZURE_GUEST_AGENT
+        kwargs['env'] = env
```
nagworld9, line 363: If caller sends `env` arg, use the same flags otherwise extend `os.enviorn` along with new marker flag

tests/ga/test_multi_config_extension.py
narrieta: let's add this (with a little more detail) as a comment in the code itself

narrieta, line 741: we should add a unit test for the environment variable, or modify the existing tests to check that the var is in the env of the test processes

azurelinuxagent/common/cgroupconfigurator.py
nagworld9: Actually, we have special cases like demon, systemd-run checks and we need those even now. I think only place we can improve when we check parent process.  We can get rid of this and add env variable check there. 
```
                    current = process
                    while current != 0 and current not in agent_commands:
                        current = self._get_parent(current)
```

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -640,6 +641,24 @@ def __format_process(pid):
...
+            Returns True if the process env flag(PARENT_NAME) found otherwise False
+            """
+            try:
+                env = '/proc/{0}/environ'.format(pid)
+                if os.path.exists(env):
+                    with open(env, "r") as env_file:
+                        environ = env_file.read()
+                        if environ and environ[-1] == '\x00':
+                            environ = environ[:-1]
+                        match = re.search(shellutil.PARENT_PROCESS_NAME + "=" + shellutil.AZURE_GUEST_AGENT, environ)
```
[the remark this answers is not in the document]
nagworld9: this is how sample output looks. I feel regex would simplify the parsing.
```
[root@edpm26zyjw azureuser]# cat /proc/1156/environ 
OLDPWD=/etc/sysconfig/network-scripts_AZURE_GUEST_AGENT_DAEMON_VERSION_=9.9.9.9PATH=/sbin:/usr/sbin:/bin:/usr/binPWD=/etc/sysconfig/network-scriptsLANG=en_US.UTF-8PARENT_PROCESS_NAME=AZURE_GUEST_AGENTSHLVL=2_=/sbin/dhclient
```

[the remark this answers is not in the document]
nagworld9: I'm not sure what you are referring but other place like in extension env variable using this  `   Prefix = "AZURE_GUEST_AGENT"`. Is this one?

azurelinuxagent/common/cgroupconfigurator.py
nagworld9: After further look I realized we need this parent check since few places like logcollector we do popen directly not going through defined code. In those scenarios, this parent pid check would catch it. So, I would say either we need all checks or update direct popen calls to shellutil. Let me know your thoughts.

tests/ga/test_multi_config_extension.py
nagworld9: addressed

nagworld9: added

azurelinuxagent/common/cgroupconfigurator.py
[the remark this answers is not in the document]
nagworld9: updated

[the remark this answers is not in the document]
kevinclark19a: Could we just use `"{0}={1}".format(shellutil.PARENT_PROCESS_NAME, shellutil.AZURE_GUEST_AGENT) in environ`?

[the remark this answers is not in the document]
narrieta: @kevinclark19a yes, that is what i was trying to point out, sorry for not being clear... I did not see any pattern in the re.search call, so this is basically a "contains".

Now @nagworld9 if you want to be more strict than simple "contains", you should match whatever separators the file uses.

azurelinuxagent/common/cgroupconfigurator.py
narrieta: that is not quite what i meant, but we can sync offline... for now you can leave it as is

narrieta (approved):
one minor comment about the use of re.search, otherwise LGTM (please look at the comment)

azurelinuxagent/common/cgroupconfigurator.py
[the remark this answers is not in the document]
nagworld9: Now I see your point and fixed it, thanks.

narrieta (approved):

kevinclark19a (approved):

PR #2538, opened 2022-03-17
don't report CPU metrics if it's negative value

nagworld9:
## Description
The cpu usage can be negative value if we notice less current cpu ticks from last time and this could happen when service restart between two polls.

Fix is not to report telemetry if cpu usage is negative value.
 
Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/common/cgroup.py
+++ b/azurelinuxagent/common/cgroup.py
@@ -120,7 +120,9 @@ class CGroup(object):
 
     def get_tracked_metrics(self, **_):
         """
-        Retrieves the current value of the metrics tracked for this cgroup and returns them as an array
+        Retrieves the current value of the metrics tracked for this cgroup and returns them as an array.
+
+        Note: Agent won't track the metrics if the current cpu ticks less than previous value and returns empty array.
         """
         raise NotImplementedError()
 
@@ -241,11 +243,16 @@ class CpuCgroup(CGroup):
         return float(self._current_throttled_time - self._previous_throttled_time) / 1E9
 
     def get_tracked_metrics(self, **kwargs):
-        tracked = [
-            MetricValue(MetricsCategory.CPU_CATEGORY, MetricsCounter.PROCESSOR_PERCENT_TIME, self.name, self.get_cpu_usage()),
-        ]
+        tracked = []
+        cpu_usage = self.get_cpu_usage()
+        if cpu_usage >= float(0):
+            tracked.append(MetricValue(MetricsCategory.CPU_CATEGORY, MetricsCounter.PROCESSOR_PERCENT_TIME, self.name, cpu_usage))
+
         if 'track_throttled_time' in kwargs and kwargs['track_throttled_time']:
-            tracked.append(MetricValue(MetricsCategory.CPU_CATEGORY, MetricsCounter.THROTTLED_TIME, self.name, self.get_throttled_time()))
+            throttled_time = self.get_throttled_time()
+            if cpu_usage >= float(0) and throttled_time >= float(0):
+                tracked.append(MetricValue(MetricsCategory.CPU_CATEGORY, MetricsCounter.THROTTLED_TIME, self.name, throttled_time))
+
         return tracked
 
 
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/cgroup.py
```diff
@@ -241,11 +243,16 @@ def get_throttled_time(self):
...
-        ]
+        tracked = []
+        cpu_usage = self.get_cpu_usage()
+        if cpu_usage >= float(0):
+            tracked.append(MetricValue(MetricsCategory.CPU_CATEGORY, MetricsCounter.PROCESSOR_PERCENT_TIME, self.name, cpu_usage))
+
         if 'track_throttled_time' in kwargs and kwargs['track_throttled_time']:
-            tracked.append(MetricValue(MetricsCategory.CPU_CATEGORY, MetricsCounter.THROTTLED_TIME, self.name, self.get_throttled_time()))
+            throttled_time = self.get_throttled_time()
+            if cpu_usage >= float(0) and throttled_time >= float(0):
```
nagworld9, line 253: In this case, very likely throttled time could have wrong value too. So, I decided not to report both.

azurelinuxagent/common/cgroup.py
```diff
@@ -241,11 +243,16 @@ def get_throttled_time(self):
         return float(self._current_throttled_time - self._previous_throttled_time) / 1E9
 
     def get_tracked_metrics(self, **kwargs):
-        tracked = [
-            MetricValue(MetricsCategory.CPU_CATEGORY, MetricsCounter.PROCESSOR_PERCENT_TIME, self.name, self.get_cpu_usage()),
-        ]
+        tracked = []
+        cpu_usage = self.get_cpu_usage()
```
nagworld9, line 247: updating previous/current values happens inside these get functions on every poll, so no explicit logic needed to reset the values.

kevinclark19a (approved):

azurelinuxagent/common/cgroup.py
narrieta: I think we do not need to check for the 2 conditions, if the single condition cpu_usage < 0 is true, we should not report the 2 metrics

tests/common/test_cgroupstelemetry.py
```diff
@@ -374,3 +374,44 @@ def test_extension_telemetry_not_sent_for_empty_perf_metrics(self, *args):  # py
                 metrics = CGroupsTelemetry.poll_all_tracked()
                 self.assertEqual(0, len(metrics))
 
+    @patch("azurelinuxagent.common.cgroup.MemoryCgroup.get_max_memory_usage")
```
narrieta, line 377: too many mocks here

azurelinuxagent/common/cgroup.py
nagworld9: yeah, if cpu_usage < 0 is true then it won't check second condition. I added second condition for rare case.

tests/common/test_cgroupstelemetry.py
nagworld9: I think I followed old test cases. Actually, we don't need to mock memory usage as this test/change only for CPU values. I will remove them.

tests/common/test_cgroupstelemetry.py
```diff
@@ -374,3 +374,44 @@ def test_extension_telemetry_not_sent_for_empty_perf_metrics(self, *args):  # py
...
+    @patch("azurelinuxagent.common.cgroup.MemoryCgroup.get_memory_usage")
+    @patch("azurelinuxagent.common.cgroup.CpuCgroup.get_cpu_usage")
+    @patch("azurelinuxagent.common.cgroup.CpuCgroup.get_throttled_time")
+    @patch("azurelinuxagent.common.cgroup.CGroup.is_active")
+    def test_cgroup_telemetry_should_not_report_cpu_negative_value(self, patch_is_active, path_get_throttled_time, patch_get_cpu_usage, patch_get_memory_usage, patch_get_memory_max_usage, *args):# pylint: disable=unused-argument
```
narrieta, line 382: don't disable unused-argument, use "_" as the argument name instead

tests/common/test_cgroupstelemetry.py
narrieta: ok, do you need the mock for is_active?

kevinclark19a (approved):

PR #2542, opened 2022-04-05
Skip outdated Fabric goal states when HGAP does not support vmSettings

narrieta:
If the agent is using FastTrack and the VM is migrated from a node where the HGAP supports vmSettings to a node where it doesn't we need to ensure that the agent does not execute an outdated Fabric goal state.

The HostPluginProtocol now keeps track of the timestamp of the most recent FastTrack goal state. If the fetch_vm_settings() method detects that support for vmSettings is no longer available, it raises  VmSettingsSupportStopped.

GoalState.update() handles that exception by restoring the current Fabric goal state. If the timestamp of this goal state is older than the most recent FastTrack goal state, it marks the GS as "outdated".

The UpdateHandler will not execute extensions if the GS is outdated.

The timestamp for the FastTrack goal state is persisted to /var/lib/waagent/fast_track.json. This is needed to handle agent restarts. The file is deleted once the agent reports status (at that point the agent will report that it does not support FastTrack and CRP will switch to Fabric.

The PR includes cleanup of related code. I'll add comments pointing out this cleanup.

### DIFF
--- a/azurelinuxagent/common/protocol/goal_state.py
+++ b/azurelinuxagent/common/protocol/goal_state.py
@@ -23,11 +23,12 @@ import azurelinuxagent.common.conf as conf
 import azurelinuxagent.common.logger as logger
 from azurelinuxagent.common.AgentGlobals import AgentGlobals
 from azurelinuxagent.common.datacontract import set_properties
+from azurelinuxagent.common.event import add_event, WALAEventOperation
 from azurelinuxagent.common.exception import ProtocolError, ResourceGoneError
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.protocol.extensions_goal_state_factory import ExtensionsGoalStateFactory
 from azurelinuxagent.common.protocol.extensions_goal_state import VmSettingsParseError, GoalStateSource
-from azurelinuxagent.common.protocol.hostplugin import VmSettingsNotSupported
+from azurelinuxagent.common.protocol.hostplugin import VmSettingsNotSupported, VmSettingsSupportStopped
 from azurelinuxagent.common.protocol.restapi import Cert, CertList, RemoteAccessUser, RemoteAccessUsersList
 from azurelinuxagent.common.utils import fileutil, timeutil
 from azurelinuxagent.common.utils.archive import GoalStateHistory
@@ -134,9 +135,15 @@ class GoalState(object):
         incarnation, xml_text, xml_doc = GoalState._fetch_goal_state(self._wire_client)
         goal_state_updated = incarnation != self._incarnation
         if goal_state_updated:
-            logger.info('Fetched new goal state from the WireServer [incarnation {0}]', incarnation)
+            logger.info('Fetched a new incarnation for the WireServer goal state [incarnation {0}]', incarnation)
 
-        vm_settings, vm_settings_updated = GoalState._fetch_vm_settings(self._wire_client)
+        vm_settings, vm_settings_updated = None, False
+        try:
+            vm_settings, vm_settings_updated = GoalState._fetch_vm_settings(self._wire_client)
+        except VmSettingsSupportStopped as exception:  # If the HGAP stopped supporting vmSettings, we need to use the goal state from the WireServer
+            self._restore_wire_server_goal_state(incarnation, xml_text, xml_doc, exception)
+            return
+
         if vm_settings_updated:
             logger.info("Fetched new vmSettings [HostGAPlugin correlation ID: {0} eTag: {1} source: {2}]", vm_settings.hostga_plugin_correlation_id, vm_settings.etag, vm_settings.source)
         # Ignore the vmSettings if their source is Fabric (processing a Fabric goal state may require the tenant certificate and the vmSettings don't include it.)
@@ -176,6 +183,18 @@ class GoalState(object):
 
         if self._extensions_goal_state is None or most_recent.created_on_timestamp > self._extensions_goal_state.created_on_timestamp:
             self._extensions_goal_state = most_recent
+
+    def _restore_wire_server_goal_state(self, incarnation, xml_text, xml_doc, vm_settings_support_stopped_error):
+        logger.info('The HGAP stopped supporting vmSettings; will fetched the goal state from the WireServer.')
+        self._history = GoalStateHistory(timeutil.create_timestamp(), incarnation)
+        self._history.save_goal_state(xml_text)
+        self._extensions_goal_state = self._fetch_full_wire_server_goal_state(incarnation, xml_doc)
+        if self._extensions_goal_state.created_on_timestamp < vm_settings_support_stopped_error.timestamp:
+            self._extensions_goal_state.is_outdated = True
+            msg = "Fetched a Fabric goal state older than the most recent FastTrack goal state; will skip it. (Fabric: {0} FastTrack: {1})".format(
+                  self._extensions_goal_state.created_on_timestamp, vm_settings_support_stopped_error.timestamp)
+            logger.info(msg)
+            add_event(op=WALAEventOperation.VmSettings, message=msg, is_success=True)
 
     def save_to_history(self, data, file_name):
         self._history.save(data, file_name)
@@ -231,6 +250,8 @@ class GoalState(object):
                     GoalState.update_host_plugin_headers(wire_client)
                     vm_settings, vm_settings_updated = wire_client.get_host_plugin().fetch_vm_settings()
 
+            except VmSettingsSupportStopped:
+                raise
             except VmSettingsNotSupported:
                 pass
             except VmSettingsParseError as exception:
@@ -249,7 +270,7 @@ class GoalState(object):
         Returns the value of ExtensionsConfig.
         """
         try:
-            logger.info('Fetching full goal state from the WireServer')
+            logger.info('Fetching full goal state from the WireServer [incarnation {0}]', incarnation)
 
             role_instance = find(xml_doc, "RoleInstance")
             role_instance_id = findtext(role_instance, "InstanceId")
--- a/azurelinuxagent/common/protocol/hostplugin.py
+++ b/azurelinuxagent/common/protocol/hostplugin.py
@@ -453,6 +453,7 @@ class HostPluginProtocol(object):
 
         Raises
             * VmSettingsNotSupported if the HostGAPlugin does not support the vmSettings API
+            * VmSettingsSupportStopped if the HostGAPlugin stopped supporting the vmSettings API
             * VmSettingsParseError if the HostGAPlugin returned invalid vmSettings (e.g. syntax error)
             * ResourceGoneError if the container ID and roleconfig name need to be refreshed
             * ProtocolError if the request fails for any other reason (e.g. not supported, time out, server error)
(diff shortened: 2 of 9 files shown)


azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -134,9 +135,15 @@ def update(self):
...
         if goal_state_updated:
-            logger.info('Fetched new goal state from the WireServer [incarnation {0}]', incarnation)
+            logger.info('Fetched a new incarnation for the WireServer goal state [incarnation {0}]', incarnation)
+
+        vm_settings, vm_settings_updated = None, False
+        try:
+            vm_settings, vm_settings_updated = GoalState._fetch_vm_settings(self._wire_client)
+        except VmSettingsSupportStopped as exception:  # If the HGAP stopped supporting vmSettings, we need to use the goal state from the WireServer
+            self._restore_wire_server_goal_state(incarnation, xml_text, xml_doc, exception)
+            return
```
narrieta, line 145: This handles  the exception that the HostPluginProtocol raises when support for vmSettings stops.

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -86,11 +85,21 @@ def __init__(self, endpoint):
         self.status_error_state = ErrorState(min_timedelta=ERROR_STATE_HOST_PLUGIN_FAILURE)
         self.fetch_last_timestamp = None
         self.status_last_timestamp = None
-        self._host_plugin_version = FlexibleVersion("0.0.0.0")  # Version 0 means "unknown"
```
narrieta, line 89: Renamed these members to remove the "host_plugin_" prefix. These used to be members of the WireClient class in a previous iteration of the code; then, the prefix made sense. Since now they are members of HostPluginProtocol, the prefix is not needed.

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -86,11 +85,21 @@ def __init__(self, endpoint):
...
+
+        # restore the state of Fast Track
+        if not os.path.exists(self._get_fast_track_state_file()):
+            self._supports_vm_settings = False
+            self._supports_vm_settings_next_check = datetime.datetime.now()
+            self._fast_track_timestamp = None
+        else:
+            self._supports_vm_settings = True
+            self._supports_vm_settings_next_check = datetime.datetime.now()
+            self._fast_track_timestamp = HostPluginProtocol._get_fast_track_timestamp()
```
narrieta, line 102: Needed for service restart

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -405,34 +414,74 @@ def _base64_encode(self, data):
...
+    @staticmethod
+    def _get_fast_track_timestamp():
+        try:
+            with open(HostPluginProtocol._get_fast_track_state_file(), "r") as file_:
+                return json.load(file_)["timestamp"]
+        except Exception as e:
+            logger.warn("Can't retrieve the timestamp for the most recent Fast Track goal state ({0}), will assume the current time. Error: {1}",
+                HostPluginProtocol._get_fast_track_state_file(), ustr(e))
+        return timeutil.create_timestamp(datetime.datetime.utcnow())
+
```
narrieta, line 448: Helpers to maintain the timestamp of the most recent FastTrack goal state

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -519,19 +570,31 @@ def format_message(msg):
...
+    def __init__(self, timestamp):
+        super(VmSettingsSupportStopped, self).__init__()
+        self.timestamp = timestamp
 
 
 class _VmSettingsError(object):
-    ServerError   = 'ServerError'
-    ClientError   = 'ClientError'
-    Timeout       = 'Timeout'
-    RequestFailed = 'RequestFailed'
```
narrieta, line 529: Alphabetized the list and added HttpError

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -540,32 +603,32 @@ def __init__(self):
...
 
-    def report_error(self, error, category=None):
+    def report_error(self, error, category):
         self._error_count += 1
 
-        if self._error_count <= _VmSettingsErrorReporter._MaxLogErrors:
-            logger.info("[VmSettings] [Informational only, the Agent will continue normal operation] {0}", error)
-
-        if self._error_count <= _VmSettingsErrorReporter._MaxTelemetryErrors:
-            add_event(op=WALAEventOperation.VmSettings, message=error, is_success=False, log_event=False)
```
narrieta, line 559: Removed these errors from the local log. This was part of 2.7 and was useful for debugging. Now, errors in vmSettings are reported the same way as any other error in the goal state. We are still collecting telemetry while the HGAP stabilizes, but some of this telemetry will also be removed in a future release.

tests/protocol/test_goal_state.py
```diff
@@ -20,7 +23,68 @@
 from tests.tools import AgentTestCase, patch, load_data
 
 
-class GoalStateTestCase(AgentTestCase):
+class GoalStateTestCase(AgentTestCase, HttpRequestPredicates):
+    def test_it_should_use_vm_settings_by_default(self):
```
narrieta, line 27: These tests used to be in test_wire.py, as part of the tests for WireClient.update_goal_state(). The code they are testing is now in GoalState, so I moved the tests here.

They are almost identical to its original version, except that now the test the code via GoalState/GoalState.update() instead of WireClient.update_goal_state()

tests/protocol/test_goal_state.py
```diff
@@ -267,3 +331,30 @@ def test_it_should_use_the_most_recent_goal_state(self):
             goal_state.update()
 
             self._assert_goal_state(goal_state, '222', channel=GoalStateChannel.WireServer, source=GoalStateSource.Fabric)
+
```
narrieta, line 334: Test for the new logic

tests/protocol/test_hostplugin.py
```diff
@@ -974,15 +960,54 @@ def get_vm_settings_call_count():
...
+            self._fetch_vm_settings_ignoring_errors(protocol)
             self.assertEqual(1, get_vm_settings_call_count(), "Additional calls to update_goal_state should not have produced extra calls to vmSettings.")
 
             # reset the vmSettings check period; this should restart the calls to the API
-            protocol.client._host_plugin._host_plugin_supports_vm_settings_next_check = datetime.datetime.now()
+            protocol.client._host_plugin._supports_vm_settings_next_check = datetime.datetime.now()
             protocol.client.update_goal_state()
             self.assertEqual(2, get_vm_settings_call_count(), "A second call to vmSettings was expecting after the check period has elapsed.")
 
+    def test_it_should_raise_when_the_vm_settings_api_stops_being_supported(self):
```
narrieta, line 972: Tests for the new logic

tests/protocol/test_wire.py
```diff
@@ -1097,67 +1094,6 @@ def test_forced_update_should_update_the_goal_state_and_the_host_plugin_when_the
             self.assertEqual(protocol.client.get_host_plugin().container_id, new_container_id)
             self.assertEqual(protocol.client.get_host_plugin().role_config_name, new_role_config_name)
 
-    def test_it_should_retry_get_vm_settings_on_resource_gone_error(self):
```
narrieta, line 1100: moved to test_goal_state.py

tests/ga/test_update.py
```diff
@@ -1463,6 +1464,7 @@ def _test_upgrade_available(
 
         self.update_handler.protocol_util = protocol
         self.update_handler._goal_state = protocol.get_goal_state()
+        self.update_handler._goal_state.extensions_goal_state.is_outdated = False
```
narrieta, line 1467: Forcing the mock created by the test to not mark the goal state as outdated

tests/protocol/test_goal_state.py
```diff
@@ -166,13 +230,13 @@ def _create_protocol_ws_and_hgap_in_sync():
             # that may invalidate this setup.
             vm_settings, _ = protocol.client.get_host_plugin().fetch_vm_settings()
             if vm_settings.etag != etag:
-                raise Exception("The HostGAPlugin is no in sync. Expected ETag {0}. Got {1}".format(etag, vm_settings.etag))
+                raise Exception("The HostGAPlugin is not in sync. Expected ETag {0}. Got {1}".format(etag, vm_settings.etag))
```
narrieta, line 233: @kevinclark19a fix for the typo in my previous PR. thanks for pointing it out

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -477,34 +527,36 @@ def format_message(msg):
...
             # TODO: update the minimum version to 1.0.8.123 before release
             if vm_settings.host_ga_plugin_version < FlexibleVersion("1.0.8.117"):
-                raise_not_supported(reset_state=True)
+                raise_not_supported()
 
-            self._host_plugin_supports_vm_settings = True
+            self._supports_vm_settings = True
             self._cached_vm_settings = vm_settings
+
+            if vm_settings.source == GoalStateSource.FastTrack:
```
nagworld9, line 551: At this point it's all about FastTrack and source always be FastTrack otherwise we would have raised not supported exception. So, I don't see point of having `else `and clearing the state. I don't know if I'm missing something here.

azurelinuxagent/ga/update.py
```diff
@@ -670,6 +671,8 @@ def _report_status(self, exthandlers_handler):
             if self._goal_state is not None:
                 agent_status = exthandlers_handler.get_ext_handlers_status_debug_info(vm_status)
                 self._goal_state.save_to_history(agent_status, AGENT_STATUS_FILE)
+                if self._goal_state.extensions_goal_state.is_outdated:
+                    exthandlers_handler.protocol.client.get_host_plugin().clear_fast_track_state()
```
nagworld9, line 675: we cleared state but status report had failures due to network issues, so we continue to skip following goal states until CRP receives not supported feature?

azurelinuxagent/common/protocol/hostplugin.py
narrieta: The HGAP continues to return both FastTrack and Fabric goal states, nothing has changed there. GoalState.update(), which calls this method is the one that will ignore the Fabric goal states.

Clearing the state is not strictly necessary, since if we receive a Fabric goal state, its timestamp will be greater than the saved state. However, deleting the state keeps things tidy and reduces the code paths the agent goes through during service start (the persisted state is used only during service start if FastTrack is disabled or not supported by the HGAP).

azurelinuxagent/ga/update.py
narrieta: exthandlers_handler.report_ext_handlers_status() should return None on error, and then we wouldn't clear the state.

But the point you make is valid: CRP will continue sending FastTrack goal states until the agent can report status successfully. That is a requirement of the current protocol.

Those goal states would time out. Something similar would happen (regardless of FastTrack) if the agent cannot report status. CRP needs the status to know the goal state completed. One difference, though, is that in this scenario the agent would not even see the FastTrack goal states.

azurelinuxagent/common/protocol/hostplugin.py
nagworld9: Gotcha, Thanks for clarifying.

azurelinuxagent/ga/update.py
nagworld9: I see your point, in that case VM might not be in good state (whole different problem) and should not worry about Agent behavior, I guess.

nagworld9 (approved):

PR #2544, opened 2022-04-07
Detect outdated Fabric goal states when FastTrack is disabled

narrieta:
If the user disables FastTrack (in waagent.conf), when we fetch the goal state from the WireServer we need to skip it if it is older than the most recent FastTrack goal state.

Also, created mock_update_handler to help tests that need to use UpdateHandler.run

### DIFF
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -45,7 +45,7 @@ from azurelinuxagent.common.exception import ResourceGoneError, UpdateError, Exi
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.common.osutil import get_osutil, systemd
 from azurelinuxagent.common.persist_firewall_rules import PersistFirewallRulesHandler
-from azurelinuxagent.common.protocol.hostplugin import HostPluginProtocol
+from azurelinuxagent.common.protocol.hostplugin import HostPluginProtocol, VmSettingsNotSupported
 from azurelinuxagent.common.protocol.restapi import VMAgentUpdateStatus, VMAgentUpdateStatuses, ExtHandlerPackageList, \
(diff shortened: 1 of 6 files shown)


kevinclark19a (approved):

azurelinuxagent/ga/update.py
```diff
@@ -469,7 +489,10 @@ def _try_update_goal_state(self, protocol):
                 add_event(AGENT_NAME, op=WALAEventOperation.FetchGoalState, version=CURRENT_VERSION, is_success=True, message=message, log_event=False)
                 logger.info(message)
 
-            self._supports_fast_track = conf.get_enable_fast_track() and protocol.client.get_host_plugin().check_vm_settings_support()
+            try:
+                self._supports_fast_track = conf.get_enable_fast_track() and protocol.client.get_host_plugin().check_vm_settings_support()
+            except VmSettingsNotSupported:
```
nagworld9, line 494: why only handle VmSettingsNotSupported exception, don't we see other exceptions here?

narrieta: We are trying to check whether the HostGAPlugin supports the vmSettings API. If it doesn't, it raises VmSettingsNotSupported. Any other exceptions are handled by the except clause on line 497.

nagworld9: okay, what I thought was on vm migration scenario we may get VmSettingsSupportStopped exception but now I realized after looking at the code, this would be the second time call for fetch_vm_settings and first call was part of updating_goal_state. So, we handle that it first time and update the supports_fast_track to False so that it will raise VmSettingsNotSupported if we check again.

nagworld9 (approved):

PR #2548, opened 2022-04-13
fix test_wait_for_handler_completion_no_status test

nagworld9:
## Description
Seems this UT has timing issue. Our current mocked timeout is too soon and not able to capture the status which we are testing in UT. Also, we have used different timeout for the same variable in other test. So I'm increasing to that number which could  fix the problem.

Issue # 

---

### PR information

### DIFF
--- a/tests/ga/test_extension.py
+++ b/tests/ga/test_extension.py
@@ -1453,7 +1453,7 @@ class TestExtension_Deprecated(TestExtensionBase):
             return original_popen(["echo", "Yes"], *args, **kwargs)
 
         with patch('azurelinuxagent.common.cgroupapi.subprocess.Popen', side_effect=mock_popen):
-            with patch('azurelinuxagent.ga.exthandlers._DEFAULT_EXT_TIMEOUT_MINUTES', 0.001):
+            with patch('azurelinuxagent.ga.exthandlers._DEFAULT_EXT_TIMEOUT_MINUTES', 0.01):
                 exthandlers_handler.run()
                 exthandlers_handler.report_ext_handlers_status()
 


tests/ga/test_extension.py
```diff
@@ -1453,7 +1453,7 @@ def mock_popen(cmd, *args, **kwargs):
             return original_popen(["echo", "Yes"], *args, **kwargs)
 
         with patch('azurelinuxagent.common.cgroupapi.subprocess.Popen', side_effect=mock_popen):
-            with patch('azurelinuxagent.ga.exthandlers._DEFAULT_EXT_TIMEOUT_MINUTES', 0.001):
+            with patch('azurelinuxagent.ga.exthandlers._DEFAULT_EXT_TIMEOUT_MINUTES', 0.01):
```
nagworld9, line 1456:
```
 ext_completed, status = False, None

            # Keep polling for the extension status until it succeeds or times out
            while datetime.datetime.utcnow() <= wait_until:
                ext_completed, status = handler_i.is_ext_handling_complete(extension)
                if ext_completed:
                    break
                time.sleep(5)
```

Status should be retrieved from` handler_i.is_ext_handling_complete` but due to timeout it didn't execute. So status had defaults.

narrieta: but the test extension is not async, right? doesn't it completely sychronously?

narrieta: but the test extension is not async, right? doesn't it completely sychronously?

nagworld9: Maybe I didn't explain you properly. The UT is to test base extension status "transitioning" if status file not found and status of dependent extension. The timeout I mentioned is how long we wait keep polling and report status. If it's still not completed within timeout, then we report last status with timeout issue on dependent extension handler status. In this case sometimes the polling won't happen at least once by the time we call this code since wait period already completed.

narrieta (approved):

kevinclark19a (approved):

PR #2552, opened 2022-04-13
Updated tests in test_update.py to use mock_update_handler

narrieta:
Updated the tests in TestUpdate to use mock_update_handler.

mock_update_handler does not start the background threads by default (these threads tend to produce random failures in the unit test runs)

### DIFF
--- a/tests/ga/mocks.py
+++ b/tests/ga/mocks.py
@@ -1,4 +1,4 @@
-# Copyright 2018 Microsoft Corporation
+    # Copyright 2018 Microsoft Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


tests/ga/mocks.py
```diff
@@ -25,17 +25,29 @@
 
 
 @contextlib.contextmanager
-def mock_update_handler(protocol, iterations=1, on_new_iteration=lambda _: None, exthandlers_handler=None, remote_access_handler=None, enable_agent_updates=False):
+def mock_update_handler(protocol,
```
narrieta, line 28: Added new parameters to the mock in order to support more test scenarios

tests/ga/test_update.py
```diff
@@ -1294,77 +1294,6 @@ def test_get_latest_agent_should_return_latest_agent_even_on_bad_error_json(self
         latest_agent = self.update_handler.get_latest_agent_greater_than_daemon()
         self.assertEqual(latest_agent.version, dst_ver, "Latest agent version is invalid")
 
-    def _test_run(self, invocations=1, calls=1, enable_updates=False, sleep_interval=(6,)):
```
narrieta, line 1297: Moved those tests to a new class UpdateHandlerRunTestCase and updated them to use mock_update_handler

nagworld9 (approved):

tests/ga/test_update.py
```diff
@@ -1897,6 +1830,77 @@ def test_it_should_reset_legacy_blacklisted_agents_on_process_start(self):
...
+            update_handler = self._test_run(autoupdate_enabled=True)
+            self.assertEqual(0, update_handler.get_iterations_completed())
+
+    def test_run_stops_if_orphaned(self):
+        with patch('os.getppid', return_value=1):
+            update_handler = self._test_run(check_daemon_running=True)
+            self.assertEqual(0, update_handler.get_iterations_completed())
+
+    def test_run_clears_sentinel_on_successful_exit(self):
+        update_handler = self._test_run()
```
kevinclark19a, line 1891: Should we create the sentinel file before starting the run? Doesn't look like this test is explicitly guaranteeing that `run` clears the sentinel file, just that it doesn't create one on a successful termination.

narrieta: the only intention of this PR is to migrate those tests to use the mock_update_handler. no other improvements (and these tests need quite a few) were made

PR #2558, opened 2022-04-19
Improvements in waagent.log

narrieta:
This PR includes several improvements in the agent's local log that help readability and debugging of goal state issues.

* Added a new message at the beginning of the ExtensionHandler similar to the Daemon's. This helps point out when the former is actually starting
* Added blank lines around fetching the goal state and processing it
* Improved the messages around certificates to make clearer when they are being downloaded/installed
* Saved the certificates thumbprints to the history folder
* Added a check for FastTrack goal states for which certificates have not been installed yet

Lastly, I added waagent_status to the list of legacy files that need to be cleaned up from the agent's root directory.

### DIFF
--- a/azurelinuxagent/common/protocol/goal_state.py
+++ b/azurelinuxagent/common/protocol/goal_state.py
@@ -314,9 +314,9 @@ class GoalState(object):
             certs = None
             certs_uri = findtext(xml_doc, "Certificates")
             if certs_uri is not None:
-                # Note that we do not save the certificates to the goal state history
                 xml_text = self._wire_client.fetch_config(certs_uri, self._wire_client.get_header_for_cert())
                 certs = Certificates(xml_text)
+                # Save the certificate summary, which includes only the thumbprint, but not the certificate itself to the goal state history
                 self._history.save_certificates(json.dumps(certs.summary))
 
             remote_access = None
--- a/azurelinuxagent/common/protocol/goal_state.py
+++ b/azurelinuxagent/common/protocol/goal_state.py
@@ -316,7 +316,7 @@ class GoalState(object):
             if certs_uri is not None:
                 xml_text = self._wire_client.fetch_config(certs_uri, self._wire_client.get_header_for_cert())
                 certs = Certificates(xml_text)
-                # Save the certificate summary, which includes only the thumbprint, but not the certificate itself to the goal state history
+                # Save the certificate summary, which includes only the thumbprint but not the certificate itself, to the goal state history
                 self._history.save_certificates(json.dumps(certs.summary))
 
             remote_access = None
--- a/azurelinuxagent/common/protocol/goal_state.py
+++ b/azurelinuxagent/common/protocol/goal_state.py
@@ -18,6 +18,7 @@
 import os
 import re
 import time
+import json
 
 import azurelinuxagent.common.conf as conf
 import azurelinuxagent.common.logger as logger
@@ -145,6 +146,7 @@ class GoalState(object):
             return
 
         if vm_settings_updated:
+            logger.info('')
             logger.info("Fetched new vmSettings [HostGAPlugin correlation ID: {0} eTag: {1} source: {2}]", vm_settings.hostga_plugin_correlation_id, vm_settings.etag, vm_settings.source)
         # Ignore the vmSettings if their source is Fabric (processing a Fabric goal state may require the tenant certificate and the vmSettings don't include it.)
         if vm_settings is not None and vm_settings.source == GoalStateSource.Fabric:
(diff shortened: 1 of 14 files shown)


azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -184,6 +186,17 @@ def update(self):
...
 
+        # For Fast Track goal states, verify that the required certificates are in the goal state
+        if self.extensions_goal_state.source == GoalStateSource.FastTrack:
+            for extension in self.extensions_goal_state.extensions:
+                for settings in extension.settings:
+                    if settings.protectedSettings is None:
+                        continue
+                    certificates = self.certs.summary
+                    if not any(settings.certificateThumbprint == c['thumbprint'] for c in certificates):
+                        message = "Certificate {0} needed by {1} is missing from the goal state".format(settings.certificateThumbprint, extension.name)
```
larohra, line 197: In this case, do we skip GS or just log error and keep going?

azurelinuxagent/ga/update.py
```diff
@@ -324,20 +324,11 @@ def run(self, debug=False):
         """
 
         try:
+            logger.info("{0} Version: {1}", AGENT_LONG_NAME, CURRENT_AGENT)
+            logger.info("OS: {0} {1}", DISTRO_NAME, DISTRO_VERSION)
```
larohra, line 328: NIT: Repeated logs (Line 328-329 are already being logged in line 345)

larohra (approved):
some minor comments else LGTM

azurelinuxagent/common/protocol/goal_state.py
narrieta: At this point it just writes to the local log to, should this ever happen, make the error obvious, and to collect telemetry.

As far as handling the error, at this point the code already handles all the scenarios we anticipated. If we are missing other scenarios the telemetry would point this out and then we can decide how to handle those.

azurelinuxagent/ga/update.py
narrieta: Yes, this repeats some of the info on L345, but removing the duplicated info is not worth it. Also, the intention is to make it as close as possible to the same message logged by the Daemon. And L345 is also send to telemetry.

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -303,6 +317,7 @@ def _fetch_full_wire_server_goal_state(self, incarnation, xml_doc):
                 # Note that we do not save the certificates to the goal state history
```
nagworld9, line 317: Now we save the certificate to history folder, should we remove this comment?

azurelinuxagent/ga/update.py
```diff
@@ -351,8 +342,20 @@ def run(self, debug=False):
...
+            #
+            protocol = self.protocol_util.get_protocol()
+
+            self._initialize_goal_state(protocol)
+
+            # Initialize the common parameters for telemetry events
+            initialize_event_logger_vminfo_common_parameters(protocol)
+
+            # Send telemetry for the OS-specific info.
             add_event(AGENT_NAME, op=WALAEventOperation.OSInfo, message=os_info_msg)
```
nagworld9, line 359: is this intentional? why don't we keep this line next to the logger.

azurelinuxagent/common/protocol/goal_state.py
narrieta: we do not save the certificate to the history folder, we never should do that since it is a secret. we only save the thrumbprint, which can be helpful when debugging

azurelinuxagent/ga/update.py
narrieta: telemetry depends on the goal state, so we need to initialize the gs first (line 353)

azurelinuxagent/common/protocol/goal_state.py
narrieta: ah, i see the confusion now... i'll update the comment to say we save the summary

tests/protocol/test_goal_state.py
```diff
@@ -358,3 +359,18 @@ def http_get_handler(url, *_, **__):
...
+        with mock_wire_protocol(data_file) as protocol:
+            with patch("azurelinuxagent.common.protocol.goal_state.add_event") as add_event:
+                _ = GoalState(protocol.client)
+
+                expected_message = "Certificate 59A10F50FFE2A0408D3F03FE336C8FD5716CF25C needed by Microsoft.OSTCExtensions.VMAccessForLinux is missing from the goal state"
+                events = [kwargs for _, kwargs in add_event.call_args_list if kwargs['op'] == WALAEventOperation.VmSettings and kwargs['message'] == expected_message]
+
+                self.assertTrue(
+                    len(events) == 1,
+                    "Missing certificate 59A10F50FFE2A0408D3F03FE336C8FD5716CF25C was note reported. Telemetry: {0}".format([kwargs['message'] for _, kwargs in add_event.call_args_list]))
```
kevinclark19a, line 376: Nit: `note reported` vs `not reported`

azurelinuxagent/ga/update.py
kevinclark19a: I don't think it's too big of a deal, but is there any reason not to move the `logger.info` line down till after grabbing the goal state?

kevinclark19a (approved):

azurelinuxagent/ga/update.py
narrieta: yes, the log messages wrt the goal state tend to obscure the ID info. The intention is to have all the information that identifies the agent at the top of the log

tests/protocol/test_goal_state.py
narrieta: thanks, will update on my next PR

nagworld9 (approved):

PR #2560, opened 2022-04-20
Change format of history items

narrieta:
This PR changes the format of the history items from

`2022-04-11T23:34:00.563193_3-11493024124256518279.zip
`

to

`2022-04-11T23-34-00__3-11493024124256518279.zip
`

i.e. use '-' instead of ':' in the time, remove the milliseconds, and use two '_' before the incarnation.

The motivation being that the incarnation number is hard to spot, and that the ':' doesn't play well with the shell (it gets in the way of autocomplete, and it won't select as a single token with double-click).

Also, it fixes a typo in a unit test and changes the format of the initial message in the Extension Handler log

### DIFF
--- a/azurelinuxagent/common/utils/archive.py
+++ b/azurelinuxagent/common/utils/archive.py
@@ -9,7 +9,7 @@ import zipfile
 
 import azurelinuxagent.common.logger as logger
 import azurelinuxagent.common.conf as conf
-from azurelinuxagent.common.utils import fileutil
+from azurelinuxagent.common.utils import fileutil, timeutil
 
 # pylint: disable=W0105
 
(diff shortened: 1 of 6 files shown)


azurelinuxagent/common/utils/archive.py
```diff
@@ -58,13 +58,15 @@
...
 #
 # Current names
 #
-#   2018-04-06T08:21:37.142697_N-M
-#   2018-04-06T08:21:37.142697_N-M.zip
+#   2018-04-06T08-21-37__N-M
+#   2018-04-06T08-21-37__N-M.zip
 #
-_ARCHIVE_BASE_PATTERN = r"\d{4}\-\d{2}\-\d{2}T\d{2}:\d{2}:\d{2}\.\d+((_incarnation)?_(\d+|status)(-\d+)?)?"
+_ARCHIVE_BASE_PATTERN = r"\d{4}\-\d{2}\-\d{2}T\d{2}[:-]\d{2}[:-]\d{2}(\.\d+)?((_incarnation)?_+(\d+|status)(-\d+)?)?"
```
nagworld9, line 69: we use '-' instead of ':' in the time, so why do we need` :` in regex pattern

narrieta: we need to match both the current and legacy formats

nagworld9 (approved):

kevinclark19a (approved):

PR #2562, opened 2022-04-22
Refresh goal state when certificates are missing

narrieta:
In some scenarios, certificates may be missing from the goal state (see comments in the code for a description of one of those scenarios). 

When that happens, the GoalState class refreshes the goal state in an attempt to fix the inconsistency. It tries this only one time and if the issue persists it raises an exception, which would be handled and reported in the main loop of the agent. The main loop would continue trying to update the goal state every period.

The PR also includes other small changes needed to support this scenario. I added comments to the PR pointing those out.


tests/data/wire/certs-2.xml
```diff
@@ -0,0 +1,85 @@
+<?xml version="1.0" encoding="utf-8"?>
```
narrieta, line 1: This new file contains certificates that do not match ExtensionsConfig

tests/data/wire/goal_state.xml
```diff
@@ -15,12 +15,12 @@
          <InstanceId>b61f93d0-e1ed-40b2-b067-22c243233448.MachineRole_IN_0</InstanceId>
          <State>Started</State>
          <Configuration>
-         <HostingEnvironmentConfig>http://168.63.129.16:80/hostingenvuri/</HostingEnvironmentConfig>
-         <SharedConfig>http://168.63.129.16:80/sharedconfiguri/</SharedConfig>
-         <Certificates>http://168.63.129.16:80/certificatesuri/</Certificates>
-         <ExtensionsConfig>http://168.63.129.16:80/extensionsconfiguri/</ExtensionsConfig>
-         <FullConfig>http://168.63.129.16:80/fullconfiguri/</FullConfig>
-         <ConfigName>b61f93d0-e1ed-40b2-b067-22c243233448.1.b61f93d0-e1ed-40b2-b067-22c243233448.2.MachineRole_IN_0.xml</ConfigName>
+         <HostingEnvironmentConfig>http://168.63.129.16:80/machine/865a6683-91d8-450f-99ae/bc8b9d47%2Db5ed%2D4704%2D85d9%2Dfd74cc967ec2.%5Fcanary?comp=config&amp;type=hostingEnvironmentConfig&amp;incarnation=1</HostingEnvironmentConfig>
```
narrieta, line 18: I updated these test files with realistic URL, instead of the previous dummy values

tests/protocol/HttpRequestPredicates.py
```diff
@@ -11,6 +11,22 @@ class HttpRequestPredicates(object):
     def is_goal_state_request(url):
         return url.lower() == 'http://{0}/machine/?comp=goalstate'.format(restutil.KNOWN_WIRESERVER_IP)
 
+    @staticmethod
```
narrieta, line 14: I only needed is_certificates_request, but since I was here I added the rest as well

tests/protocol/mockwiredata.py
```diff
@@ -135,10 +135,10 @@ def __init__(self, data_files=None):
             "/HealthService": 0,
             "/vmAgentLog": 0,
             "goalstate": 0,
-            "hostingenvuri": 0,
-            "sharedconfiguri": 0,
-            "certificatesuri": 0,
-            "extensionsconfiguri": 0,
+            "hostingEnvironmentConfig": 0,
```
narrieta, line 138: made the name of these counters consistent with the actual URLs

azurelinuxagent/ga/update.py
```diff
@@ -600,11 +600,14 @@ def _process_goal_state(self, exthandlers_handler, remote_access_handler):
             protocol = exthandlers_handler.protocol
 
             # update self._goal_state
-            self._try_update_goal_state(protocol)
-
-            # Update the Guest Agent if a new version is available
-            if self._goal_state is not None:
```
narrieta, line 606: we were executing the goal state even when we fail to update it... it's a regression introduced a few releases ago

azurelinuxagent/common/logger.py
```diff
@@ -124,6 +125,9 @@ def write_log(log_appender):  # pylint: disable=W0612
                 finally:
                     log_appender.appender_lock = False
 
+        if self.silent:
+            return
+
```
narrieta, line 130: added a new flag to turn on/off logging

azurelinuxagent/ga/update.py
```diff
@@ -481,13 +482,16 @@ def _try_update_goal_state(self, protocol):
         Attempts to update the goal state and returns True on success or False on failure, sending telemetry events about the failures.
         """
         try:
-            protocol.update_goal_state()
+            max_errors_to_log = 3
+
```
narrieta, line 486: As part of the Fast Track changes more logging has been added to the GoalState class. I added the capability to turn off all logging done by GoalState.
 
We leaving logging turned on for at most 3 consecutive errors and then turn it off and output 1 single error every 6 hours. We turn on logging once fetching goal state succeeds.

tests/protocol/test_goal_state.py
```diff
@@ -360,17 +359,54 @@ def http_get_handler(url, *_, **__):
...
 
-                expected_message = "Certificate 59A10F50FFE2A0408D3F03FE336C8FD5716CF25C needed by Microsoft.OSTCExtensions.VMAccessForLinux is missing from the goal state"
-                events = [kwargs for _, kwargs in add_event.call_args_list if kwargs['op'] == WALAEventOperation.VmSettings and kwargs['message'] == expected_message]
+            expected_message = "Certificate 59A10F50FFE2A0408D3F03FE336C8FD5716CF25C needed by Microsoft.OSTCExtensions.VMAccessForLinux is missing from the goal state"
+            self.assertIn(expected_message, str(context.exception))
+
+    def test_it_should_refresh_the_goal_state_when_it_is_inconsistent(self):
+        #
+        # Some scenarios can produce inconsistent goal states. For example, during hibernation/resume, the Fabric goal state changes (the
+        # tenant certificate is re-generated when the VM is restarted) *without* the incarnation changing. If a Fast Track goal state
```
kevinclark19a, line 378: Not sure if this is necessarily the place for this much info, but it doesn't look like this text matches that above (in goal_state.py) w.r.t. the reset of incarnation to 1 (meaning the incarnation _might_ not change) vs. the incarnation not changing.

narrieta: This comment is trying to describe the specific scenario that the test is emulating. In this scenario the incarnation does not change.

The code in goal_state.py describes the general case, where the incarnation may or may not change.

Does that address your comment? (not sure I understand the comment 100%)

tests/ga/test_update.py
```diff
@@ -2711,10 +2711,8 @@ def create_log_and_telemetry_mocks():
             with create_log_and_telemetry_mocks() as (logger, add_event):
                 update_handler._try_update_goal_state(protocol)
 
-                w = warnings()
-                pw = periodic_warnings()
-                self.assertEqual(1, len(w), "A failure should have produced a warning: [{0}]".format(w))
-                self.assertEqual(1, len(pw), "A failure should have produced a periodic warning: [{0}]".format(pw))
+                e = errors()
+                self.assertEqual(1, len(e), "A failure should have produced a warning: [{0}]".format(e))
```
kevinclark19a, line 2715: NIT: more accurately, "a failure should have produced an error".

azurelinuxagent/ga/update.py
```diff
@@ -497,15 +501,21 @@ def _try_update_goal_state(self, protocol):
...
-                message = u"An error occurred while retrieving the goal state: {0}".format(textutil.format_exception(e))
-                logger.warn(message)
-                add_event(AGENT_NAME, op=WALAEventOperation.FetchGoalState, version=CURRENT_VERSION, is_success=False, message=message, log_event=False)
-            message = u"Attempts to retrieve the goal state are failing: {0}".format(ustr(e))
-            logger.periodic_warn(logger.EVERY_SIX_HOURS, "[PERIODIC] {0}".format(message))
+            self._update_goal_state_error_count += 1
             self._heartbeat_update_goal_state_error_count += 1
+            if self._update_goal_state_error_count <= max_errors_to_log:
+                message = u"Error fetching the goal state: {0}".format(textutil.format_exception(e))
+                logger.error(message)
```
kevinclark19a, line 508: What is the context here behind this switch from `warn` to `error`? It makes sense to me why it _would_ be an error, but do we know the reason for the `warn` before hand?

tests/ga/test_update.py
narrieta: thanks, fixed

tests/ga/test_update.py
```diff
@@ -1898,7 +1898,7 @@ def update_goal_state_and_run_handler():
 
     def test_it_should_wait_to_fetch_first_goal_state(self):
         with _get_update_handler() as (update_handler, protocol):
-            with patch("azurelinuxagent.common.logger.warn") as patch_warn:
+            with patch("azurelinuxagent.common.logger.error") as patch_error:
```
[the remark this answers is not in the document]
narrieta: yes, failures to update the goal states should be considered errors (previous code was warning), the updates in the test reflect the updates in the _try_update_goal_state method

azurelinuxagent/ga/update.py
narrieta: that was just the way coded it initially. should have been an error

kevinclark19a (approved):

azurelinuxagent/ga/update.py
nagworld9: @narrieta Sorry, I didn't get a chance to review this. Do we turn on before the goal state succeeded or will it be for next goal state? Looks to me successful goal state may not have logging if it gets updated after 3 consecutive errors. 

Ex: We had 3 consecutive failures, on 4th attempt we disable the logger but goal state updated successfully and process the same goal state. So don't we miss that process goal state log in waagent because logger disabled.

narrieta: thanks for taking a look @nagworld9 

yes, we re-enable logging at line 492

narrieta: (we reset to 0 and the condition for turning it off is "self._update_goal_state_error_count >= max_errors_to_log")

PR #2569, opened 2022-04-27
Do not mark goal state as processed when goal state fails to update

narrieta:
If we fail to update the goal state, we should not mark it as processed (i.e. we should not update UpdateHandler._last_incarnation and UpdateHandler._last_extensions_gs_id).

### DIFF
--- a/azurelinuxagent/common/protocol/goal_state.py
+++ b/azurelinuxagent/common/protocol/goal_state.py
@@ -209,7 +209,7 @@ class GoalState(object):
         else:  # vm_settings_updated
             most_recent = vm_settings
 
-        if self._extensions_goal_state is None or most_recent.created_on_timestamp > self._extensions_goal_state.created_on_timestamp:
+        if self._extensions_goal_state is None or most_recent.created_on_timestamp >= self._extensions_goal_state.created_on_timestamp:
             self._extensions_goal_state = most_recent
 
         #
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -209,7 +209,7 @@ def _update(self, force_update):
         else:  # vm_settings_updated
             most_recent = vm_settings
 
-        if self._extensions_goal_state is None or most_recent.created_on_timestamp > self._extensions_goal_state.created_on_timestamp:
+        if self._extensions_goal_state is None or most_recent.created_on_timestamp >= self._extensions_goal_state.created_on_timestamp:
```
narrieta, line 212: The vmSettings ETag captures changes changes in the FT or Fabric goal states, we one can a new ETag even when the artifacts profile does not change (e.g. the svdSeqNumber -- the incarnation -- can change if there is a new Fabric goal state with new certificates). The timestamp is in the artifacts profile blob so it would not change in that case.

Adding "=" to the condition to capture the new eTag.

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -221,7 +221,7 @@ def _update(self, force_update):
         # Track goal state comes after that, the extensions will need the new certificate. The Agent needs to refresh the goal state in that
         # case, to ensure it fetches the new certificate.
         #
-        if self.extensions_goal_state.source == GoalStateSource.FastTrack:
+        if self._extensions_goal_state.source == GoalStateSource.FastTrack:
```
narrieta, line 224: Using the private member (_extensions_goal_state) instead of the public property (extensions_goal_state) for consistency with the rest of the code.

kevinclark19a (approved):

nagworld9:
@narrieta LGTM

PR #2577, opened 2022-04-29
Bug Fix: Change fast track timestamp default from None to datetime.min

kevinclark19a:
## Description

In case of a live migration from a host supporting VmSettings to one that doesn't, the current code caches the timestamp of the last fast track goal state in memory so we can determine whether the next fabric goal state should be executed.

However, if the live migration happens before we ever receive a fast track goal state from HostGAPlugin (i.e. HostGAPlugin supports VmSettings but only ever sent Fabric goal states through), we never update the initial value of `None` for `_fast_track_timestamp`. This can cause a `NoneType` exception whenever we try to compare that cached timestamp to a fabric one.

This PR instead initializes the timestamp to datetime.min, indicating that _any_ fabric goal state is newer than the (nonexisted) cached last fast track goal state.

---

### PR information


tests/ga/test_update.py
```diff
@@ -2919,16 +2919,53 @@ def test_it_should_mark_outdated_goal_states_on_service_restart_when_host_ga_plu
     def test_it_should_clear_the_timestamp_for_the_most_recent_fast_track_goal_state(self):
         data_file = self._prepare_fast_track_goal_state()
 
-        if HostPluginProtocol.get_fast_track_timestamp() is None:
+        if HostPluginProtocol.get_fast_track_timestamp() is timeutil.create_timestamp(datetime.min):
```
narrieta, line 2922: == instead of is

tests/ga/test_update.py
```diff
@@ -2919,16 +2919,53 @@ def test_it_should_mark_outdated_goal_states_on_service_restart_when_host_ga_plu
...
                 with mock_update_handler(protocol) as update_handler:
                     update_handler.run()
 
-        self.assertIsNone(HostPluginProtocol.get_fast_track_timestamp(), "The Fast Track state was not cleared")
+        self.assertEqual(HostPluginProtocol.get_fast_track_timestamp(), timeutil.create_timestamp(datetime.min),
+            "The Fast Track state was not cleared")
 
+    def test_it_should_default_fast_track_timestamp_to_datetime_min(self):
+        data = DATA_FILE_VM_SETTINGS.copy()
+        data["vm_settings"] = "hostgaplugin/vm_settings-fabric-no_extension_manifests.json"
```
narrieta, line 2935: are the extension manifests relevant to the test case?

tests/ga/test_update.py
```diff
@@ -2919,16 +2919,53 @@ def test_it_should_mark_outdated_goal_states_on_service_restart_when_host_ga_plu
...
+                if iteration == 1:
+                    protocol.mock_wire_data.set_incarnation(2)
+                    protocol.set_http_handlers(http_get_handler=vm_settings_no_change)
+                elif iteration == 2:
+                    protocol.mock_wire_data.set_incarnation(3)
+                    protocol.set_http_handlers(http_get_handler=vm_settings_not_supported)
+            
+            with mock_update_handler(protocol, 3, on_new_iteration=mock_live_migration) as update_handler:
+                with patch("azurelinuxagent.ga.update.logger.error") as patched_error:
+                    def match_unexpected_errors():
```
narrieta, line 2960: copy n' paste error? it's checking error messages, not timestamps

tests/ga/test_update.py
```diff
@@ -2919,16 +2919,53 @@ def test_it_should_mark_outdated_goal_states_on_service_restart_when_host_ga_plu
...
                     update_handler.run()
 
-        self.assertIsNone(HostPluginProtocol.get_fast_track_timestamp(), "The Fast Track state was not cleared")
+        self.assertEqual(HostPluginProtocol.get_fast_track_timestamp(), timeutil.create_timestamp(datetime.min),
+            "The Fast Track state was not cleared")
 
+    def test_it_should_default_fast_track_timestamp_to_datetime_min(self):
+        data = DATA_FILE_VM_SETTINGS.copy()
+        data["vm_settings"] = "hostgaplugin/vm_settings-fabric-no_extension_manifests.json"
+        data['goal_state'] = 'wire/goal_state_no_certs.xml'
```
narrieta, line 2936: i'd add a comment about the test using no certs to work around a limitation on the mocks, but that certs are irrelevant to the test

tests/ga/test_update.py
kevinclark19a: No, but I'm using an extensions config with no certificates to get around a bug in the test code (the permissions issue on the certificate file we talked about). The default `vm_settings.json` fails in `_check_certificates` when doing the fetch.

tests/ga/test_update.py
kevinclark19a: good catch, will push a fix

tests/ga/test_update.py
kevinclark19a: I actually want to leave this here to catch any failures, but I agree I need to add a check on the actual timestamp.

tests/ga/test_update.py
narrieta: yes, that is ok for goal_state_**no_certs**.xml, but why vm_settings-fabric-**no_extension_manifests**.json? you could have left the extension manifest in the test data and just remove the certificate properties.

or rename the file not to draw attention to the manifests, since they are irrelevant. vm_settings-fabric.json? (though test data without extension manifests looks funny)

narrieta (approved):

azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -95,7 +95,7 @@ def __init__(self, endpoint):
         if not os.path.exists(self._get_fast_track_state_file()):
             self._supports_vm_settings = False
             self._supports_vm_settings_next_check = datetime.datetime.now()
-            self._fast_track_timestamp = None
+            self._fast_track_timestamp = timeutil.create_timestamp(datetime.datetime.min)
```
nagworld9, line 98: does this cause opposite behavior where we want to skip the goal state if current goal state is outdated and fast_track_timestamp has min value. I don't know we ever get to that case, just throwing to think of any.
```
        if self._extensions_goal_state.created_on_timestamp < vm_settings_support_stopped_error.timestamp:
            self._extensions_goal_state.is_outdated = True
            msg = "Fetched a Fabric goal state older than the most recent FastTrack goal state; will skip it.\nFabric:    {0}\nFastTrack: {1}".format(
                  self._extensions_goal_state.created_on_timestamp, vm_settings_support_stopped_error.timestamp)
            self.logger.info(msg)
```

nagworld9 (approved):

PR #2581, opened 2022-05-04
ext cpu throttling scenarios

nagworld9:
## Description

Cgroup disable on following cases:-

**Agent throttled** - reset agent cpu quota.

**Extra process in Agent cgroup** - Disable All (includes extension and its services). reset quotas and reset monitors.
 
**Systemd-run failure**: disable all, reset quotas and reset monitors.

Also have changes for logging throttled time on extension timeout case.

---

### PR information

### DIFF
--- a/azurelinuxagent/common/cgroup.py
+++ b/azurelinuxagent/common/cgroup.py
@@ -139,7 +139,6 @@ class CpuCgroup(CGroup):
         self._current_system_cpu = None
         self._previous_throttled_time = None
         self._current_throttled_time = None
-        self._throttled_time_lock = threading.RLock()  # Protect the get_throttled_time which is called from Monitor thread and cgroupconfigurator.
 
     def _get_cpu_ticks(self, allow_no_such_file_or_directory_error=False):
         """
@@ -174,7 +173,6 @@ class CpuCgroup(CGroup):
         return cpu_ticks
 
     def get_throttled_time(self):
-        self._throttled_time_lock.acquire()
         try:
             with open(os.path.join(self.path, 'cpu.stat')) as cpu_stat:
                 #
@@ -196,8 +194,6 @@ class CpuCgroup(CGroup):
             raise CGroupsException("Failed to read cpu.stat: {0}".format(ustr(e)))
         except Exception as e:
             raise CGroupsException("Failed to read cpu.stat: {0}".format(ustr(e)))
-        finally:
-            self._throttled_time_lock.release()
 
     def _cpu_usage_initialized(self):
         return self._current_cgroup_cpu is not None and self._current_system_cpu is not None
@@ -234,11 +230,15 @@ class CpuCgroup(CGroup):
 
         return round(100.0 * self._osutil.get_processor_cores() * float(cgroup_delta) / float(system_delta), 3)
 
-    def get_cpu_throttled_time(self):
+    def get_cpu_throttled_time(self, read_previous_throttled_time=True):
         """
         Computes the throttled time (in seconds) since the last call to this function.
         NOTE: initialize_cpu_usage() must be invoked before calling this function
+        Compute only current throttled time if read_previous_throttled_time set to False
         """
+        if not read_previous_throttled_time:
+            return float(self.get_throttled_time() / 1E9)
+
         if not self._cpu_usage_initialized():
             raise CGroupsException("initialize_cpu_usage() must be invoked before the first call to get_throttled_time()")
 
--- a/azurelinuxagent/common/cgroup.py
+++ b/azurelinuxagent/common/cgroup.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 #
 # Requires Python 2.6+ and Openssl 1.0+
-import threading
 from collections import namedtuple
 
 import errno
--- a/azurelinuxagent/common/cgroup.py
+++ b/azurelinuxagent/common/cgroup.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 #
 # Requires Python 2.6+ and Openssl 1.0+
+import threading
 from collections import namedtuple
 
 import errno
@@ -138,6 +139,7 @@ class CpuCgroup(CGroup):
         self._current_system_cpu = None
         self._previous_throttled_time = None
         self._current_throttled_time = None
+        self._throttled_time_lock = threading.RLock()  # Protect the get_throttled_time which is called from Monitor thread and cgroupconfigurator.
 
     def _get_cpu_ticks(self, allow_no_such_file_or_directory_error=False):
         """
@@ -171,7 +173,8 @@ class CpuCgroup(CGroup):
 
         return cpu_ticks
 
-    def _get_throttled_time(self):
+    def get_throttled_time(self):
+        self._throttled_time_lock.acquire()
         try:
             with open(os.path.join(self.path, 'cpu.stat')) as cpu_stat:
                 #
@@ -193,6 +196,8 @@ class CpuCgroup(CGroup):
             raise CGroupsException("Failed to read cpu.stat: {0}".format(ustr(e)))
         except Exception as e:
             raise CGroupsException("Failed to read cpu.stat: {0}".format(ustr(e)))
+        finally:
+            self._throttled_time_lock.release()
 
     def _cpu_usage_initialized(self):
         return self._current_cgroup_cpu is not None and self._current_system_cpu is not None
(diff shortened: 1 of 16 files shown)


azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -873,6 +913,31 @@ def start_tracking_extension_services_cgroups(self, services_list):
                     if service_name is not None:
                         self.start_tracking_unit_cgroups(service_name)
 
+        @staticmethod
+        def get_extension_services_list():
```
nagworld9, line 917: retrieve the manifest files of installed extensions to reset quotas when we disable cgroups for all.

azurelinuxagent/common/utils/extensionprocessutil.py
```diff
@@ -62,13 +65,18 @@ def handle_process_completion(process, command, timeout, stdout, stderr, error_c
...
+    :param cpu_cgroup: Reference the cpu cgroup name and path
     :return:
     """
     # Wait for process completion or timeout
-    timed_out, return_code = wait_for_process_completion_or_timeout(process, timeout)
+    timed_out, return_code, throttled_time = wait_for_process_completion_or_timeout(process, timeout, cpu_cgroup)
     process_output = read_output(stdout, stderr)
 
     if timed_out:
+        if cpu_cgroup is not None:# Report CPUThrottledTime when timeout happens
```
nagworld9, line 76: Report the Throttled time in the exception msg when timeout case.

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -479,18 +485,22 @@ def enable(self):
...
+
                 CGroupsTelemetry.reset()
+                self._agent_cgroups_enabled = False
+                self._extensions_cgroups_enabled = False
             elif disable_cgroups == DisableCgroups.AGENT:  # disable agent
                 self._agent_cgroups_enabled = False
                 self.__reset_agent_cpu_quota()
                 CGroupsTelemetry.stop_tracking(CpuCgroup(AGENT_NAME_TELEMETRY, self._agent_cpu_cgroup_path))
-            elif disable_cgroups == DisableCgroups.EXTENSIONS:  # disable extensions
-                self._extensions_cgroups_enabled = False
```
nagworld9, line 493: not needed as we don't disable particular extension

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -519,7 +529,6 @@ def __reset_agent_cpu_quota():
             """
             logger.info("Resetting agent's CPUQuota")
             if CGroupConfigurator._Impl.__try_set_cpu_quota(''):  # setting an empty value resets to the default (infinity)
-                CGroupsTelemetry.set_track_throttled_time(False)
```
nagworld9, line 522: Not needed as we are already stop tracking the cgroup and also this would set to false for extensions too which we don't want to do it on agent throttling case

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -479,18 +485,22 @@ def enable(self):
...
             if disable_cgroups == DisableCgroups.ALL:  # disable all
-                self._agent_cgroups_enabled = False
-                self._extensions_cgroups_enabled = False
+                # Reset quotas
                 self.__reset_agent_cpu_quota()
+                extension_services = self.get_extension_services_list()
+                for extension in extension_services:
+                    self.__reset_extension_cpu_quota(extension_name=extension)
+                    self.__reset_extension_services_cpu_quota(extension_services[extension])
+                self.__reload_systemd_config()
```
narrieta, line 495: can we verify whether the reload is strictly neccessary? in general, we should try to minimize the use of daemon-reload, since that is a system-wide operation

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -479,18 +485,22 @@ def enable(self):
...
 
         def disable(self, reason, disable_cgroups):
-            # Todo: disable/reset extension when ext quotas introduced
             if disable_cgroups == DisableCgroups.ALL:  # disable all
-                self._agent_cgroups_enabled = False
-                self._extensions_cgroups_enabled = False
+                # Reset quotas
                 self.__reset_agent_cpu_quota()
+                extension_services = self.get_extension_services_list()
+                for extension in extension_services:
```
narrieta, line 492: can we add some log.info about the stuff we are resetting?

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -823,13 +842,34 @@ def set_extension_services_cpu_memory_quota(self, services_list):
...
+                            drop_in_file_cpu_quota = os.path.join(drop_in_path, _DROP_IN_FILE_CPU_QUOTA)
+                            cpu_quota_contents = _DROP_IN_FILE_CPU_QUOTA_CONTENTS_FORMAT.format(cpu_quota)
+                            if os.path.exists(drop_in_file_cpu_quota):
+                                with open(drop_in_file_cpu_quota, "r") as file_:
+                                    if file_.read() == cpu_quota_contents:
+                                        return
+                                files_to_create.append((drop_in_file_cpu_quota, cpu_quota_contents))
+                            self.__create_all_files(files_to_create)
+            except Exception as exception:
+                _log_cgroup_warning('Failed to set CPUQuota: {0}', ustr(exception))
```
narrieta, line 872: let's add the extension/service name to the error message

azurelinuxagent/common/cgroupconfigurator.py
```diff
@@ -873,6 +913,31 @@ def start_tracking_extension_services_cgroups(self, services_list):
...
 
+        @staticmethod
+        def get_extension_services_list():
+            """
+            ResourceLimits for extensions are coming from <extName>/HandlerManifest.json file.
+            Use this pattern to determine all the installed extension HandlerManifest files and
+            read the extension services if ResourceLimits are present.
+            """
+            extensions_services = {}
+            for manifest_path in glob.iglob(os.path.join(conf.get_lib_dir(), "*/HandlerManifest.json")):
```
narrieta, line 924: can we abstract the extension traversal in ExtensionsHandlerHandler? that class already has some code to do this kind of stuff, and takes into account, for example, extensions that failed to install (you probably want to skip those)

tests/common/test_cgroupconfigurator.py
```diff
@@ -269,7 +269,55 @@ def test_disable_should_reset_cpu_quota_and_tracked_cgroups(self):
                 fileutil.findre_in_file(agent_drop_in_file_cpu_quota, "^CPUQuota=$"),
                 "CPUQuota was not set correctly. Expected an empty value. Got:\n{0}".format(fileutil.read_file(agent_drop_in_file_cpu_quota)))
             self.assertEqual(len(CGroupsTelemetry._tracked), 0, "No cgroups should be tracked after disable. Tracking: {0}".format(CGroupsTelemetry._tracked))
-            self.assertFalse(CGroupsTelemetry._track_throttled_time, "Throttle Time should not be tracked after disable")
+
```
narrieta, line 272: apart from the unit tests, are you planning on adding any end-to-end tests?

azurelinuxagent/common/utils/extensionprocessutil.py
nagworld9: yes, I would like to do that way but when timeout happens, we kill the process before returns ExtensionTimeout exception. That would remove the cgroups so I had to pass cpu_cgroup down the lane to compute before kill.

azurelinuxagent/common/cgroup.py
```diff
@@ -138,6 +139,7 @@ def __init__(self, name, cgroup_path):
         self._current_system_cpu = None
         self._previous_throttled_time = None
         self._current_throttled_time = None
+        self._throttled_time_lock = threading.RLock()  # Protect the get_throttled_time which is called from Monitor thread and cgroupconfigurator.
```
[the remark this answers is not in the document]
nagworld9: I think we need in the case of long running extensions which are running more than 5mins or so. In those cases, monitor thread and update handler will have race condition.

tests/common/test_cgroupconfigurator.py
nagworld9: Are you talking about end-to-end tests in DCR? if so, I'll create a task for it

azurelinuxagent/common/cgroupconfigurator.py
nagworld9: updated

azurelinuxagent/common/cgroupconfigurator.py
nagworld9: updated

azurelinuxagent/common/utils/extensionprocessutil.py
```diff
@@ -141,3 +149,17 @@ def to_s(captured_stdout, stdout_offset, captured_stderr, stderr_offset):
...
+
+
+def get_cpu_throttled_time(cpu_cgroup):
+    """
+    return the throttled time for the given cgroup.
+    """
+    throttled_time = 0
+    if cpu_cgroup is not None:
+        try:
+            throttled_time = float(cpu_cgroup.get_throttled_time() / 1E9)
```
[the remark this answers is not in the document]
nagworld9: addressed

narrieta (approved):

kevinclark19a (approved):

PR #2602, opened 2022-06-01
removing private key from logs

norakoiralamsft:
## Description

Part of bigger task of removing the logging of private keys from agent + extension logs. 
This change replaces private thumbprint with the length the thumbprint (for debugging purposes)

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/common/protocol/goal_state.py
+++ b/azurelinuxagent/common/protocol/goal_state.py
@@ -221,7 +221,7 @@ class Certificates(object):
                 tmp_file = prvs[pubkey]
                 prv = "{0}.prv".format(thumbprint)
                 os.rename(tmp_file, os.path.join(conf.get_lib_dir(), prv))
-                logger.info("Found private key matching thumbprint {0}".format(thumbprint))
+                logger.info("Found private key matching thumbprint with length: {0}".format(len(thumbprint)))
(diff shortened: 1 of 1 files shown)


azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -221,7 +221,7 @@ def __init__(self, xml_text):
                 tmp_file = prvs[pubkey]
                 prv = "{0}.prv".format(thumbprint)
                 os.rename(tmp_file, os.path.join(conf.get_lib_dir(), prv))
-                logger.info("Found private key matching thumbprint {0}".format(thumbprint))
+                logger.info("Found private key matching thumbprint with length: {0}".format(len(thumbprint)))
```
jscalev, line 224: The thumbprint itself shouldn't be a danger to log and is useful for debugging (needed it on an incident this week). Why are we changing this?

narrieta: @norakoiralamsft Yes, we shouldn't change this. As @jscalev points out, logging the thumbprint is not a concern, and it is needed for debugging some issues, especially issues related to Fast Track. The length is not enough to distinguish the certificate, since all the thumbprints are of the same length,

PR #2637, opened 2022-08-01
refactor logcollector memory usage

nagworld9:
## Description
Problem: The log collector process runs with 30MB of cgroup memory limit. Some vms reported OOM kills of the log collector process when it's reaches the limit.  

Solution: Agent monitors the memory every 2 secs in separate thread and gracefully exit the log collector process when it's reaches 30MB limit. That way we can avoid force kills by OOM killer.

```

2022-08-01T02:55:25.357959Z INFO LogCollectorMonitorHandler LogCollector Log collector memory limit 31457280 bytes exceeded. The max reported usage is 32534528 bytes.
2022-08-01T02:55:25.546377Z INFO CollectLogsHandler ExtHandler Disabling periodic log collection until service restart due to exceeded process memory limit.
```

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/ga/collect_logs.py
+++ b/azurelinuxagent/ga/collect_logs.py
@@ -172,7 +172,9 @@ class CollectLogsHandler(ThreadHandlerInterface):
     def _collect_logs(self):
         logger.info("Starting log collection...")
 
-        # Invoke the command line tool in the agent to collect logs, with resource limits on CPU and memory (RAM).
+        # Invoke the command line tool in the agent to collect logs, with resource limits on CPU.
+        # Some distros like ubuntu20 by default cpu and memory accounting enabled. Thus create nested cgroups under the logcollector slice
+        # So disabling CPU and Memory accounting prevents from creating nested cgroups, so that all the counters will be present in logcollector Cgroup
 
         systemd_cmd = [
             "systemd-run", "--property=CPUAccounting=no", "--property=MemoryAccounting=no",
@@ -189,19 +191,16 @@ class CollectLogsHandler(ThreadHandlerInterface):
             success = False
             msg = None
             try:
-                base_dir = conf.get_lib_dir()
-                with tempfile.TemporaryFile(dir=base_dir, mode="w+b") as stdout:
-                    with tempfile.TemporaryFile(dir=base_dir, mode="w+b") as stderr:
-                        shellutil.run_command(final_command, log_error=False, stdout=stdout, stderr=stderr)
-                        duration = elapsed_milliseconds(start_time)
-                        archive_size = os.path.getsize(COMPRESSED_ARCHIVE_PATH)
+                shellutil.run_command(final_command, log_error=False)
+                duration = elapsed_milliseconds(start_time)
+                archive_size = os.path.getsize(COMPRESSED_ARCHIVE_PATH)
 
-                        msg = "Successfully collected logs. Archive size: {0} b, elapsed time: {1} ms.".format(archive_size,
-                                                                                                               duration)
-                        logger.info(msg)
-                        success = True
+                msg = "Successfully collected logs. Archive size: {0} b, elapsed time: {1} ms.".format(archive_size,
+                                                                                                       duration)
+                logger.info(msg)
+                success = True
 
-                        return True
+                return True
             except Exception as e:
                 duration = elapsed_milliseconds(start_time)
                 err_msg = ustr(e)
--- a/azurelinuxagent/ga/collect_logs.py
+++ b/azurelinuxagent/ga/collect_logs.py
@@ -16,27 +16,29 @@
 #
 # Requires Python 2.6+ and Openssl 1.0+
 #
-
 import datetime
 import os
 import sys
+import tempfile
 import threading
 import time
 from azurelinuxagent.common import cgroupconfigurator, logcollector
 
 import azurelinuxagent.common.conf as conf
 from azurelinuxagent.common import logger
-from azurelinuxagent.common.event import elapsed_milliseconds, add_event, WALAEventOperation
-from azurelinuxagent.common.future import subprocess_dev_null, ustr
+from azurelinuxagent.common.cgroup import MetricsCounter
+from azurelinuxagent.common.event import elapsed_milliseconds, add_event, WALAEventOperation, report_metric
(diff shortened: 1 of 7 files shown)


azurelinuxagent/ga/collect_logs.py
```diff
@@ -170,29 +175,33 @@ def _collect_logs(self):
...
-                duration = elapsed_milliseconds(start_time)
-                archive_size = os.path.getsize(COMPRESSED_ARCHIVE_PATH)
-
-                msg = "Successfully collected logs. Archive size: {0} b, elapsed time: {1} ms.".format(archive_size,
-                                                                                                    duration)
-                logger.info(msg)
-                success = True
-
-                return True
+                base_dir = conf.get_lib_dir()
```
nagworld9, line 192: Since the main logger set to capture stdout,  systemd-run stdout to temp file instead of agent log file to avoid duplicates in agent log

azurelinuxagent/ga/collect_logs.py
```diff
@@ -128,7 +130,10 @@ def stopped(self):
     def stop(self):
         self.should_run = False
         if self.is_alive():
-            self.join()
+            try:
+                self.join()
+            except RuntimeError:
```
nagworld9, line 135: RuntimeError being raised when agent try to stop the thread to exit gracefully. Seems like expected error ignoring this

  File "bin/WALinuxAgent-9.9.9.9-py3.8.egg/azurelinuxagent/ga/collect_logs.py", line 123, in join
    self.event_thread.join()
  File "/usr/lib/python3.6/threading.py", line 1053, in join
    raise RuntimeError("cannot join current thread")
RuntimeError: cannot join current thread

azurelinuxagent/ga/collect_logs.py
```diff
@@ -261,3 +262,98 @@ def _send_logs(self):
                 is_success=success,
                 message=msg,
                 log_event=False)
+
+
+def get_log_collector_monitor_handler(cgroups):
+    return LogCollectorMonitorHandler(cgroups)
+
+
+class LogCollectorMonitorHandler(ThreadHandlerInterface):
```
nagworld9, line 271: Montor thread

azurelinuxagent/ga/collect_logs.py
```diff
@@ -261,3 +262,98 @@ def _send_logs(self):
...
+        for metric in metrics:
+            if metric.counter == MetricsCounter.TOTAL_MEM_USAGE:
+                current_usage += metric.value
+            elif metric.counter == MetricsCounter.SWAP_MEM_USAGE:
+                current_usage += metric.value
+            elif metric.counter == MetricsCounter.MAX_MEM_USAGE:
+                max_usage = metric.value
+
+        current_max = max(current_usage, max_usage)
+        if current_max > LOGCOLLECTOR_MEMORY_LIMIT:
```
nagworld9, line 351: This block checks the memory limit and sends exit

azurelinuxagent/ga/collect_logs.py
```diff
@@ -170,29 +175,33 @@ def _collect_logs(self):
         # Invoke the command line tool in the agent to collect logs, with resource limits on CPU and memory (RAM).
 
         systemd_cmd = [
-            "systemd-run", "--unit={0}".format(logcollector.CGROUPS_UNIT),
+            "systemd-run", "--property=CPUAccounting=no", "--property=MemoryAccounting=no",
```
narrieta, line 178: I'd add a comment explaining why this is set to "no"

azurelinuxagent/ga/collect_logs.py
```diff
@@ -170,29 +175,33 @@ def _collect_logs(self):
...
-                archive_size = os.path.getsize(COMPRESSED_ARCHIVE_PATH)
-
-                msg = "Successfully collected logs. Archive size: {0} b, elapsed time: {1} ms.".format(archive_size,
-                                                                                                    duration)
-                logger.info(msg)
-                success = True
-
-                return True
+                base_dir = conf.get_lib_dir()
+                with tempfile.TemporaryFile(dir=base_dir, mode="w+b") as stdout:
```
narrieta, line 193: why do we need those tmp files?

narrieta (commented):
Overall LGTM, just a few minor comments/questions

azurelinuxagent/ga/collect_logs.py
nagworld9: Nothing specific, we are doing same thing for extension systemd-run commands. Since it's working solution, I decided to use same thing.

narrieta: In the case of extensions commands we use files because we need to reset stdout/stderr when systemd-run fails

```
           # Reset the stdout and stderr
            stdout.truncate(0)
            stderr.truncate(0)

```
but here the files are not needed

azurelinuxagent/ga/collect_logs.py
narrieta: could you explain this comment? thanks

azurelinuxagent/ga/collect_logs.py
nagworld9: I see your point, I use default Subprocess.PIPE instead.

azurelinuxagent/ga/collect_logs.py
nagworld9: added comment.

azurelinuxagent/ga/collect_logs.py
nagworld9: Duplicate logs in agent log are due to `systemd-run popen `uses the agent log file for stdout to write and also, agent initialized the logger to capture the any stdout. Since both are using same log file, we would see duplicates. So, I changed the popen stdout to use standard pipe instead of log file, so that popen won't write to the file.

narrieta: sorry, still don't get it. I think this comment is misplaced, can't find the corresponding code, can you point me to it? thanks

nagworld9: this line https://github.com/Azure/WALinuxAgent/pull/2637/files#diff-35808f273042b1e0b61f6e439a3469a775aadb1e4850db80c742480435fb6258R193

narrieta (approved):

PR #2661, opened 2022-09-01
waagent.service: set ConditionVirtualization=|microsoft

cjp256:
Only start waagent service when running under Microsoft virtualization.

Set it as a triggering condition to make it easier for downstreams or
test setups to add another condition (i.e. run outside of hyperv).

Signed-off-by: Chris Patterson <cpatterson@microsoft.com>

### DIFF
--- a/bin/waagent2.0
+++ b/bin/waagent2.0
@@ -1569,6 +1569,7 @@ After=network.target
 After=sshd.service
 ConditionFileIsExecutable=/usr/sbin/waagent
 ConditionPathExists=/etc/waagent.conf
+ConditionVirtualization=|microsoft
 
 [Service]
 Type=simple
--- a/init/arch/waagent.service
+++ b/init/arch/waagent.service
@@ -5,6 +5,7 @@ After=network-online.target
 
 ConditionFileIsExecutable=/usr/bin/waagent
 ConditionPathExists=/etc/waagent.conf
+ConditionVirtualization=|microsoft
 
 [Service]
 Type=simple
--- a/init/clearlinux/waagent.service
+++ b/init/clearlinux/waagent.service
@@ -5,6 +5,7 @@ After=network-online.target
 
 ConditionFileIsExecutable=/usr/bin/waagent
 ConditionPathExists=/usr/share/defaults/waagent/waagent.conf
+ConditionVirtualization=|microsoft
 
 [Service]
 Type=simple
--- a/init/coreos/cloud-config.yml
+++ b/init/coreos/cloud-config.yml
@@ -26,6 +26,7 @@ coreos:
         Description=Microsoft Azure Agent
         Wants=network-online.target sshd-keygen.service
         After=network-online.target sshd-keygen.service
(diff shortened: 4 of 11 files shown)


Conan-Kudo (approved):

bin/waagent2.0
```diff
@@ -1569,6 +1569,7 @@ After=network.target
 After=sshd.service
 ConditionFileIsExecutable=/usr/sbin/waagent
 ConditionPathExists=/etc/waagent.conf
+ConditionVirtualization=microsoft
```
Conan-Kudo, line 1572: You probably want `ConditionVirtualization=|microsoft` instead

> Condition checks can use a pipe symbol ("|") after the equals sign ("Condition…=|…"), which causes the condition to become a triggering condition. If at least one triggering condition is defined for a unit, then the unit will be started if at least one of the triggering conditions of the unit applies and all of the regular (i.e. non-triggering) conditions apply. If you prefix an argument with the pipe symbol and an exclamation mark, the pipe symbol must be passed first, the exclamation second. If any of these options is assigned the empty string, the list of conditions is reset completely, all previous condition settings (of any kind) will have no effect.

Conan-Kudo (changes requested):

bin/waagent2.0
cjp256: @Conan-Kudo can you elaborate?  

Seems like it would make sense if we had an alternative trigger to run regardless of virtualization, e.g.:
```
ConditionVirtualization=|microsoft
ConditionPathExists=|/etc/run-waagent-even-if-not-hyperv
```

But given that no alternative trigger exists, is there a practical reason for making a lone triggering condition?

cjp256: I ask also because I recently made this change which is rolling down into RHEL: https://src.fedoraproject.org/rpms/hyperv-daemons/c/f4c66aab854882e15ce939196f076510b9c3729f?branch=rawhide :D

Conan-Kudo: I guess not? I saw that it was done that way for the VirtualBox guest agents, and I figure there was a good reason for that: 

* https://src.fedoraproject.org/rpms/virtualbox-guest-additions/blob/rawhide/f/vboxclient.service
* https://src.fedoraproject.org/rpms/virtualbox-guest-additions/blob/rawhide/f/vboxservice.service

cjp256: I updated it to use it as a triggering condition.  Thinking about it more it would make it easier for someone to extend the unit to run outside of hyper-v, if needed.

sgallagher (approved):

Conan-Kudo (approved):

narrieta:
@cjp256 I discussed this with Chris Patterson [cpatterson@microsoft.com](mailto:cpatterson@microsoft.com) internally. We don't have the infrastructure to test the change to all those distros, and some of them don't need this change since the agent is installed only on Azure images. Chris commented that the intention of these changes was to help onboard Fedora, please limit the changes to that distro. Thanks

PR #2662, opened 2022-09-03
use 30 secs timeout for put vmagent log api

nagworld9:
## Description
exercise 30 secs timeout to reduce latency failure rates on PUT VMagent log api.
Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/common/utils/restutil.py
+++ b/azurelinuxagent/common/utils/restutil.py
@@ -309,8 +309,8 @@ def redact_sas_tokens_in_urls(url):
     return SAS_TOKEN_RETRIEVAL_REGEX.sub(r"\1" + REDACTED_TEXT + r"\3", url)
 
 
-def _http_request(method, host, rel_uri, port=None, data=None, secure=False,
-                  headers=None, proxy_host=None, proxy_port=None, redact_data=False, timeout=10):
+def _http_request(method, host, rel_uri, timeout, port=None, data=None, secure=False,
+                  headers=None, proxy_host=None, proxy_port=None, redact_data=False):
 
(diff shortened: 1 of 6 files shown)


azurelinuxagent/common/utils/restutil.py
```diff
@@ -310,7 +310,7 @@ def redact_sas_tokens_in_urls(url):
 
 
 def _http_request(method, host, rel_uri, port=None, data=None, secure=False,
-                  headers=None, proxy_host=None, proxy_port=None, redact_data=False):
+                  headers=None, proxy_host=None, proxy_port=None, redact_data=False, timeout=10):
```
narrieta, line 313: We defining the default timeout in multiple places. We should probably do that on the higher level functions (e.g. http_put) and make a mandatory parameter in the other methods (http_request, _http_request)

narrieta (commented):
1 comment

azurelinuxagent/common/utils/restutil.py
nagworld9: That would force us to change every http_* API call (GET, POST, etc). I'm okay to do that too.

narrieta: yes, I think that is a better API interface (which requires more code changes, but hey...) :)

narrieta: (Unless you think otherwise)

nagworld9: Sometimes better things come with price. I'm gonna pay for it if things are clean :)

narrieta (commented):
thanks for the update

narrieta (approved):

PR #2665, opened 2022-09-16
Update waagent for hardening best practices

mfortin:
## Description
CIS Benchmark has a control that requires `/tmp` partition to be mounted with `noexec` to prevent arbitatry file executions. In order to implement this setting this change is required to use python to load the file vs loading the binary directly.

Issue # 

---

### PR information
- [X] The title of the PR is clear and informative.
- [X] New Unit tests were added for the changes made


narrieta:
@mfortin  Thanks for your contribution. What is the motivation of this change? Did you find an actual scenario where the current code actually fails?

Most of waagent2.0 is dead code, though some small portions are used by very (really) old extensions. We are not taking any changes to this module, and we don't have any infrastructure to test changes to it. If you found an scenario where this code is actually executed, we can follow up with the corresponding extension to have them update their code and remove any references to waagent2.0.

mfortin:
@narrieta I believe that is the agent available on Ubuntu 20.04 image I pulled from the marketplace.

I had issues while running our existing hardening cookbook as we have a requirement to set the `noexec` mount option on `/tmp` and `/var/tmp` and that looked like the root cause for the failure I had. Can you point me to the repo for the latest agent ?

narrieta:
@mfortin I'm closing this PR. The code in your change is dead code. Most likely the issue is coming from the extensions you are trying to execute.

mfortin:
@narrieta ok, will keep digging the root source of my issue then. Thanks.

PR #2682, opened 2022-10-07
Use common download logic for agent downloads

### LINKED ISSUE #2489: [BUG] waagent downloads updates to /var/lib/waagent but doesn't removes the old .zip files
Description: Any Linux OS running waagent for a few months will have at least a few .zip files at /var/lib/waagent that corresponds to older versions of the WALinuxAgent file that is downloaded every time Azure pushes a new update.

Over time these files start to pile up and add up to unnecessary space utilization on the root file syste or /var if LVM is in use.

The idea is to include a log rotation config file to automatically find and delete these files or even adding a cron job to this would be fine.

 - Distro and Version: All
 - WALinuxAgent version: All

It's easy to check this on any server that has been running waagent for more that a few months ( lest 6-8 months ).

Just run:

`/usr/bin/find /var/lib/waagent -name "WALinuxAgent*.zip" -type f -exec ls -ltr {} \+`

An you will see all the old updates downloaded by the waagent

As an example, an easy way to fix this is by adding a log rotate config file with these commands

```
cat << EOF > /etc/logrotate.d/waagent.cleanup
/var/lib/waagent/*.zip {
        # Ignore if there are no zip files at /var/lib/waagent/
        missingok

        # Don't send email
        nomail

postrotate
        # Remove any file that matches WALinuxAgent*.zip" and it's older than 60 days
        /usr/bin/find /var/lib/waagent -name "WALinuxAgent*.zip" -type f -mtime +60 -exec rm {} \;
endscript
}
EOF
```

narrieta:
This PR completes the refactoring of the download logic to use a common implementation for all downloads.

The downloads on this PR are for the agent's package. I made one change of behavior: now we delete the ZIP file once we expand it (both for the agent and extensions). These ZIPs are not used after we expand them and we have received reports that the current cleanup logic does not always work. I did not remove the cleanup logic, since we still need to cleanup files downloaded by previous agents or the daemon.

Fixes #2489


azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -152,13 +152,6 @@ def _fetch_manifest(self, manifest_type, name, uris):
         except Exception as e:
             raise ProtocolError("Failed to retrieve {0} manifest. Error: {1}".format(manifest_type, ustr(e)))
 
-    def download_extension(self, uris, destination, on_downloaded=lambda: True):
```
narrieta, line 155: Initially I added this method here for consistency with fetch_extension_manifest above, but it really doesn't quite fit here so I'm removing it.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -604,35 +606,39 @@ def hgap_download(uri):
 
         return self._download_with_fallback_channel(download_type, uris, direct_download=direct_download, hgap_download=hgap_download)
 
-    def download_extension(self, uris, destination, use_verify_header, on_downloaded=lambda: True):
```
narrieta, line 607: renamed this method since it can be used to download the agent as well; also, now this method can only be used to download ZIPs (we don't download any other type of files) and the expansion of the package is done here, instead of in the caller.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -658,14 +664,34 @@ def _download_with_fallback_channel(self, download_type, uris, direct_download,
...
+            logger.error("Error while unzipping {0}: {1}", package_type, ustr(exception))
+            if os.path.exists(target_directory):
+                try:
+                    shutil.rmtree(target_directory)
+                except Exception as exception:
+                    logger.warn("Cannot delete {0}: {1}", target_directory, ustr(exception))
+            raise
+        finally:
+            try:
+                os.remove(target_file)
```
narrieta, line 691: now we remove the ZIP file after expansion (or download)

azurelinuxagent/ga/update.py
```diff
@@ -1494,18 +1488,20 @@ def _reset_legacy_blacklisted_agents(self):
 
 
 class GuestAgent(object):
-    def __init__(self, path=None, pkg=None, is_fast_track_goal_state=False, host=None):
+    def __init__(self, path, pkg, protocol, is_fast_track_goal_state):
```
narrieta, line 1491: The interface for __init__ is confusing, since it can be used for 2 mutually exclusive purposes depending on the parameters given/omitted. I defined 'from_installed_agent' and 'from_installed_agent' that can be used instead of __init__ with a (hopefully) clearer interface

azurelinuxagent/ga/update.py
```diff
@@ -1529,26 +1525,12 @@ def __init__(self, path=None, pkg=None, is_fast_track_goal_state=False, host=Non
             self._ensure_downloaded()
             self._ensure_loaded()
         except Exception as e:
-            if isinstance(e, ResourceGoneError):
```
narrieta, line 1532: ResourceGoneError is already handled in the common download logic, no need to treat it as a special case here.

azurelinuxagent/ga/update.py
```diff
@@ -1529,26 +1525,12 @@ def __init__(self, path=None, pkg=None, is_fast_track_goal_state=False, host=Non
...
-                raise
-
-            # The agent was improperly blacklisting versions due to a timeout
-            # encountered while downloading a later version. Errors of type
-            # socket.error are IOError, so this should provide sufficient
-            # protection against a large class of I/O operation failures.
-            if isinstance(e, IOError):
```
narrieta, line 1539: This check seems to be leftover code that is not needed. There is a unit test that verifies we do not blacklist the agent on download error so these errors are handled correctly even without this check.

azurelinuxagent/ga/update.py
```diff
@@ -1637,39 +1632,10 @@ def _ensure_loaded(self):
         self._load_error()
 
     def _download(self):
-        uris_shuffled = self.pkg.uris
```
narrieta, line 1640: now we use the common download logic instead of this (as well as _fetch below)

tests/ga/test_extension.py
```diff
@@ -140,8 +141,6 @@ def test_cleanup_leaves_installed_extensions(self):
             exthandlers_handler.run()
             exthandlers_handler.report_ext_handlers_status()
 
-            self.assertEqual(no_of_exts, TestExtensionCleanup._count_packages(),
```
narrieta, line 143: The ZIP package is now deleted after expansion so all these tests do not need to check the count of packages

tests/ga/test_extension.py
```diff
@@ -682,120 +673,6 @@ def test_it_should_process_valid_extensions_if_present(self, mock_get, mock_cryp
             expected_handlers.remove(handler.name)
         self.assertEqual(0, len(expected_handlers), "All handlers not reported status")
 
-    def test_ext_zip_file_packages_removed_in_update_case(self, *args):
```
narrieta, line 685: The ZIP package is now deleted after expansion so all these tests are not needed

tests/protocol/test_hostplugin.py
```diff
@@ -163,7 +163,7 @@ def create_mock_protocol():
             yield protocol
 
     @patch("azurelinuxagent.common.protocol.healthservice.HealthService.report_host_plugin_versions")
-    @patch("azurelinuxagent.ga.update.restutil.http_get")
```
narrieta, line 166: This is sad. We were mocking restutil.http_get via the update module, which is not even used by this test.

tests/protocol/test_imds.py
```diff
@@ -20,18 +20,18 @@
...
 import unittest
 
-import azurelinuxagent.common.protocol.imds as imds
+from azurelinuxagent.common.protocol import imds
 
 from azurelinuxagent.common.datacontract import set_properties
 from azurelinuxagent.common.exception import HttpError, ResourceGoneError
 from azurelinuxagent.common.future import ustr, httpclient
 from azurelinuxagent.common.utils import restutil
-from tests.ga.test_update import ResponseMock
```
narrieta, line 29: There are multiple implementations of mock HTTP responses that make things confusing and cause some really weird test failures. Removing the usage of this one in favor of MockHttpResponse, which provides a more complete implementation

tests/protocol/test_hostplugin.py
nagworld9: yeah, I have seen these few places, thanks for correcting.

azurelinuxagent/ga/update.py
nagworld9: Thanks for changing this

azurelinuxagent/ga/update.py
```diff
@@ -1529,26 +1525,12 @@ def __init__(self, path=None, pkg=None, is_fast_track_goal_state=False, host=Non
...
-                raise
-
-            # If we're unable to download/unpack the agent, delete the Agent directory and the zip file (if exists) to
-            # ensure we try downloading again in the next round.
+            # If we're unable to download/unpack the agent, delete the Agent directory
             try:
                 if os.path.isdir(self.get_agent_dir()):
                     shutil.rmtree(self.get_agent_dir(), ignore_errors=True)
-                if os.path.isfile(self.get_agent_pkg_path()):
-                    os.remove(self.get_agent_pkg_path())
```
nagworld9, line 1548: I feel like we should keep this too which gives second chance to remove incase first attempt failed or didn't get to the unzip package path.

narrieta: we still have the cleanup logic that runs periodically.

also, a new goal state would find that the ZIP is there and skip the download (and remove the ZIP after expanding it)

tests/ga/test_extension.py
narrieta: yup, i should have added that test.

added now: TestWireClient.test_download_zip_package_should_expand_and_delete_the_package

azurelinuxagent/ga/update.py
nagworld9: I don't find the cleanup logic for agent ZIPs. When I check the vm, zips still there. I don't think we have cleanup thing. 

I'm more worry about if some reason agent gets partial zip or corrupted zip while downloading and received exception before expanding it. In that case on next goal state, we would skip the download if that zip is still in the vm.

narrieta: A bad download should be covered as well.

The previous logic was trying to delete on download or expand failures. That cleanup is done covered elsewhere: if we begin the download, start creating the file and then downloading the rest of the file fails, the WireClient.stream would delete the ZIP. If the download succeeds, but the ZIP is corrupt and expand fails, then WireClient._try_expand_zip_package would remove it.

I was mistaken in my previous comment: a new goal state does not check for the existence of the ZIP, but rather for the manifest file. If for any reason a bad ZIP is still there from a previous goal state, the new goal state would do the download and overwrite the bad ZIP.

Lastly, during agent update, UpdateHandler._purge_agents removes agent directories and ZIPs that are no longer in the goal state.

All of this maintains the same logic as before, except that some of the resposibilities have moved elsewhere as part of using the common download logic.

nagworld9: Sonds like we are covering all the cases, thanks for explaining.

nagworld9 (approved):

Issue #2674, opened 2022-09-30
[BUG][RHEL7.9] walinuxagent hang


Klaas-:
**Describe the bug: A clear and concise description of what the bug is.**

Note: Please add some context which would help us understand the problem better
1. there are no logs, the process simply hangs in a very uninterruptable state. I tried to gcore it even that hangs.
2. Last sign of life in journal/waagent.log: `INFO ExtHandler ExtHandler All extensions in the goal state have reached a terminal state: [(u'Microsoft.Azure.ActiveDirectory.AADSSHLoginForLinux', u'Ready'), (u'Microsoft.Azure.Monitoring.DependencyAgent.DependencyAgentLinux', u'Ready'), (u'Microsoft.Azure.RecoveryServices.VMSnapshotLinux', u'success'), (u'Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux', u'success')]`

**Distro and WALinuxAgent details (please complete the following information):**
 - Distro and Version: `Red Hat Enterprise Linux Server release 7.9 (Maipo)`
 - WALinuxAgent version:
```
WALinuxAgent-2.3.0.2 running on redhat 7.9
Python: 2.7.5
Goal state agent: 2.8.0.11
```

**Additional context**
I have opened a support case: 2209290050001897

**Log file attached**
If possible, please provide the full /var/log/waagent.log file to help us understand the problem better and get the context of the issue.
Last signs of life:
```
2022-09-25T18:29:12.141284Z INFO ExtHandler ExtHandler Started extension in unit 'enable_id.scope'
2022-09-25T18:29:12.142212Z INFO ExtHandler ExtHandler Started tracking cgroup Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux-1.14.19 [/sys/fs/cgroup/cpu,cpuacct/azure.slice/azure-vmextensions.slice/azure-vmextensions-Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux_1.14.19.slice]
2022/09/25 20:29:12 OmsAgentForLinux started to handle.
2022/09/25 20:29:12 [Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux-1.14.19] cwd is /var/lib/waagent/Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux-1.14.19
[...]

PR #2685, opened 2022-10-17
drop cgroup support for rhel

nagworld9:
## Description

CX reported the waagent hang issue https://github.com/Azure/WALinuxAgent/issues/2674. RHEL team suspecting that Kernel bug in implementation of cgroups. Since the agent uses cgroups for resource governance, dropping the rhel support to be cautious and avoid potential widespread.
  
Issue # 

---

### PR information


narrieta (approved):

Klaas-:
Hi,
two things:
1) if rhel is affected so is centos
2) at least for me this only seems to happen on 7.9, not on 8.

nagworld9:
@Klaas- It's not very clear what is the RCA and what distros really impacted.  At this point we only know rhel affected so we decide to drop the entire rhel versions to be on conservative side.

Klaas-:
there should be zero functional difference between a centos 7 and a rhel 7 kernel. with centos stream it's a little different, but for 7 it's a clone. The literally just exchange some certs and strings https://git.centos.org/rpms/kernel/c/595b4ccbe0e0e84f193c39a24f4c78e3933ebee6?branch=c7 @nagworld9

vittyvk:
I also don't quite understand why such a big hammer is being used here. RHEL kernels support cgroups and the problem (https://github.com/Azure/WALinuxAgent/issues/2674) is only observed on RHEL7. Why do we need to disable cgroup support for RHEL8 and RHEL9? I also agree that Centos and RHEL should have the same behavior.

nagworld9:
@vittyvk Today our agent only uses cgroup v1 and not supporting the RHEL9 as this by default, mounts and utilizes cgroups-v2. 
We are in the process of releasing new version of agent and we don't want to block this deployment until we know the RCA. So, with this uncertainty we felt it's safe to drop support for all rhel and centos and retake our deployment to avoid potential widespread as we are setting the cpu quota to 50% in the new version.

Klaas-:
@vittyvk they also dropped it for centos after my push via https://github.com/Azure/WALinuxAgent/pull/2689

nagworld9:
@Klaas- @vittyvk If it was miscommunicated earlier, I just want to make it clear that the changes here do not drop support for RHEL, they simply stop setting up a CPU quota on the agent.

Klaas-:
I think we understood that :)

narrieta:
Fixes #2674

PR #2688, opened 2022-10-21
Report message in handler heartbeat

maddieford:
## Description
This PR updates the agent to report the extension handler heartbeat message. Previously, any message in heartbeat.log was ignored. 

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -1001,7 +1001,8 @@ class ExtHandlersHandler(object):
                     heartbeat = ext_handler_i.collect_heartbeat()
                     if heartbeat is not None:
                         handler_status.status = heartbeat.get('status')
-                        handler_status.message = parse_formatted_message(heartbeat.get('formattedMessage'))
+                        if 'formattedMessage' in heartbeat:
+                            handler_status.message = parse_formatted_message(heartbeat.get('formattedMessage'))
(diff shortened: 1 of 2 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -1001,6 +1001,7 @@ def report_ext_handler_status(self, vm_status, ext_handler, goal_state_changed):
                     heartbeat = ext_handler_i.collect_heartbeat()
                     if heartbeat is not None:
                         handler_status.status = heartbeat.get('status')
+                        handler_status.message = parse_formatted_message(heartbeat.get('formattedMessage'))
```
narrieta, line 1004: you should check that the heartbeat actually has a 'formattedMessage')

nagworld9: why do we need extra check. The get call would handle if it's not available. It will return None if property does not present and same applies to status field.

narrieta: @nagworld9 we need the check so that we do not assign None to these properties

nagworld9: I see.  At this point we are overriding the handler status file content with heartbeat not with default None values. So makes sense to check here.

narrieta (approved):

nagworld9 (approved):

PR #2704, opened 2022-11-11
Run tests using a container

narrieta:
Installing LISA typically takes 5-7 min. These changes execute the tests using a container where LISA is pre-installed; pulling the container image typically takes 50-90 seconds.

I also fixed some Python imports that I missed in my previous PR, renamed some resources in our pipeline and made some of those resources parameters to the pipeline.


tests_e2e/azure-pipelines.yml
```diff
@@ -1,10 +1,6 @@
 variables:
   - name: azureConnection
-    value: 'AzLinux DCR Public (8e037ad4-618f-4466-8bc8-5099d41ac15b)'
-  - name: subId
-    value: '8e037ad4-618f-4466-8bc8-5099d41ac15b'
-  - name: testsSourcesDirectory
-    value: "$(Build.SourcesDirectory)/tests_e2e"
+    value: 'azuremanagement'
```
narrieta, line 3: I renamed the connection and made the actual subscription ID a parameter.

tests_e2e/docker/Dockerfile
```diff
@@ -0,0 +1,74 @@
+#
```
narrieta, line 1: This is basically the previous install_dependencies.sh now as a Dockerfile

tests_e2e/requirements.txt
```diff
@@ -10,4 +10,3 @@ cryptography
 distro
 junitparser
 msrestazure
-python-dotenv
```
narrieta, line 13: I removed this dependency. We are over-using the environment to pass data across different components. We need to reduce the use of environment variables, so this module should not be needed.

tests_e2e/scenario_utils/azure_models.py
```diff
@@ -12,8 +12,8 @@
 from azure.mgmt.resource import ResourceManagementClient
 from msrestazure.azure_exceptions import CloudError
 
-from dcr.scenario_utils.logging_utils import LoggingHandler
-from dcr.scenario_utils.models import get_vm_data_from_env, VMModelType, VMMetaData
+from tests_e2e.scenario_utils.logging_utils import LoggingHandler
```
narrieta, line 15: fixing some dependencies I missed in my previous PR

tests_e2e/scenario_utils/logging_utils.py
```diff
@@ -0,0 +1,33 @@
+# Create a base class
```
narrieta, line 1: When I fixed the dependencies, I realized I missed this file in my previous PR. This file comes from the current Azure Pipeline and needs to be rewritten

tests_e2e/scripts/install_dependencies.sh
```diff
@@ -1,43 +0,0 @@
-#!/usr/bin/env bash
```
narrieta, line 1: The Dockerfile replaces this script

tests_e2e/docker/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+                                                                                                                 \
+    #                                                                                                            \
+    # Install LISA                                                                                               \
+    #                                                                                                            \
+    cd $HOME                                                                                                  && \
+    git clone https://github.com/microsoft/lisa.git                                                           && \
+    cd lisa                                                                                                   && \
+    poetry config virtualenvs.in-project true                                                                 && \
+    make setup                                                                                                && \
+    chmod +r /home/waagent/.config/pypoetry/config.toml                                                       && \
```
narrieta, line 61: this chmod is a left-over from some experimentation I was doing, i'll remove it

tests_e2e/scripts/execute_tests_container.sh
```diff
@@ -0,0 +1,27 @@
...
+az acr login --name waagenttests > /dev/null
+
+docker pull waagenttests.azurecr.io/waagenttests:latest
+
+# Logs will be placed in this location. Make waagent (UID 1000 in the container) the owner.
+mkdir "$HOME/logs"
+sudo chown 1000 "$HOME/logs"
+
+docker run --rm \
+      --volume "$BUILD_SOURCESDIRECTORY:/home/waagent/WALinuxAgent" \
```
narrieta, line 16: "--volume <host_path>:<container_path>" shares the local <host_path> with the container as <container-path>.

we are sharing the agent's code and the the ssh key to connect to the test VMs (both of which Azure Pipelines downloaded to the VM); also, we are specifying a location where the container will drop the logs and then the host VM will pick them up from there to report results.

maddieford (approved):

tests_e2e/requirements.txt
narrieta: Use of environment variables should be restricted to the interface with the Azure pipelines and probably a few (if any) places. If used for other things other than communication with the pipeline, we need to be very explicit about what variables we are  using. There should not be a need for this module to save and restore the entire environment.

tests_e2e/scripts/execute_tests.sh
```diff
@@ -2,17 +2,19 @@
...
-  -v identity_file:$SSHKEY_SECUREFILEPATH
+# Execute the tests, this needs to be done from the LISA root directory
+cd "$HOME/lisa"
 
+./lisa.sh \
+  --runbook "$HOME/WALinuxAgent/tests_e2e/lisa/runbook/azure.yml" \
+  --log_path "$HOME/logs" \
+  --working_path "$HOME/logs" \
+  -v subscription_id:"$SUBSCRIPTION_ID" \
+  -v identity_file:"$HOME/.ssh/id_rsa.pub"
```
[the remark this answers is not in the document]
narrieta: Yes, this script is executed within the container using the 'waagent' user. $USER, $HOME, etc refer to 'waagent'.

tests_e2e/scripts/execute_tests_container.sh
```diff
@@ -0,0 +1,27 @@
...
+      --volume "$DOWNLOADSSHKEY_SECUREFILEPATH:/home/waagent/id_rsa" \
```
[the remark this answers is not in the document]
narrieta: $BUILD_SOURCESDIRECTORY and $DOWNLOADSSHKEY_SECUREFILEPATH are inputs to the container, while $HOME/logs is its output.

We should not modify the input directories, since the may be coming from a developer's machine while debugging/etc (the script that executes the tests creates a file under '/home/waagent' and changes the mode of '/home/waagent/id_rsa')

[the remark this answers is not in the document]
nagworld9: I see your point. If the file semantics changes in container path would change the input path too. Makes sense.

nagworld9 (approved):

PR #2706, opened 2022-11-30
agent update refactor

nagworld9:
## Description
- Skip update if self-update disabled 
- Retrieve Requested version from goal state
- Check if update is permitted (means we allow agent update per hour frequency) based on last_update_time for that version.
- Download agent package for requested version
- Purge agents except the current version and requested version
- Process update – Upgrade pretty straightforward and downgrade (mark the current agent as bad version) (There must be good reason for not wanting current version in downgrade scenario)
- Report status 
- Daemon continues to exclude bad agents as of today
- Reset the bad agents if those are done by older agents based on error flag.

Issue # 

---

### PR information


tests/ga/test_update.py
```diff
@@ -316,301 +318,6 @@ def replicate_agents(self,
         return dst_v
 
 
-class TestGuestAgentError(UpdateTestCase):
```
nagworld9, line 319: Moved to test_agent_update.py

tests/ga/test_update.py
```diff
@@ -316,301 +318,6 @@ def replicate_agents(self,
...
-        s = "Last Failure: {0}, Total Failures: {1}, Fatal: {2}, Reason: {3}".format(
-            WITH_ERROR["last_failure"],
-            WITH_ERROR["failure_count"],
-            WITH_ERROR["was_fatal"],
-            WITH_ERROR["reason"])
-        self.assertEqual(s, str(err))
-        return
-
-
-class TestGuestAgent(UpdateTestCase):
```
nagworld9, line 400: Moved to test_agent_update.py

tests/ga/test_update.py
```diff
@@ -1192,85 +896,6 @@ def test_shutdown_ignores_exceptions(self):
         except Exception as e:  # pylint: disable=unused-variable
             self.assertTrue(False, "Unexpected exception")  # pylint: disable=redundant-unittest-assert
 
-    def _test_upgrade_available(
```
nagworld9, line 1195: not applicable now

azurelinuxagent/ga/update.py
```diff
@@ -606,18 +519,18 @@ def _processing_new_extensions_goal_state(self):
...
+    def _process_goal_state(self, exthandlers_handler, remote_access_handler, agent_update_handler):
         protocol = exthandlers_handler.protocol
 
         # update self._goal_state
         if not self._try_update_goal_state(protocol):
             # agent updates and status reporting should be done even when the goal state is not updated
-            self.__update_guest_agent(protocol)
-            self._report_status(exthandlers_handler)
+            agent_update_handler.run()
+            self._report_status(exthandlers_handler, agent_update_handler, gs_updated=False)
```
narrieta, line 529: it seems to me like _report_status should not take a dependency on the agent_update_handler... all it needs is the update status, right? if yes, it may be better to pass the update status instead of the whole update_handler

azurelinuxagent/ga/update.py
```diff
@@ -606,18 +519,18 @@ def _processing_new_extensions_goal_state(self):
...
             return
 
-        # check for agent updates
-        self.__update_guest_agent(protocol)
+        gs_updated = self._processing_new_extensions_goal_state() or self._processing_new_incarnation()
```
narrieta, line 532: why do we need to check for self._processing_new_incarnation()? the required version will come in the extensions goal state, right?

azurelinuxagent/ga/update.py
```diff
@@ -1485,290 +1143,3 @@ def _reset_legacy_blacklisted_agents(self):
                 agent.clear_error()
         except Exception as err:
             logger.warn("Unable to reset legacy blacklisted agents due to: {0}".format(err))
-
-
-class GuestAgent(object):
```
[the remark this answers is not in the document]
narrieta: agent_update.py may not be the best location for this. This is used not only during update, but also to start the latest version. Let's move it to its own module.

azurelinuxagent/ga/update.py
```diff
@@ -1485,290 +1143,3 @@ def _reset_legacy_blacklisted_agents(self):
...
-            self.name,
-            self.get_agent_manifest_path())
-        logger.verbose(u"Successfully loaded Agent {0} {1}: {2}",
-                       self.name,
-                       AGENT_MANIFEST_FILE,
-                       ustr(self.manifest.data))
-        return
-
-
-class GuestAgentError(object):
```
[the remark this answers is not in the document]
narrieta: let's move it to its own module, it's used in more scenarios than agent update

narrieta (commented):
did a partial review, posting just a few comments for now... review still in progress

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+            logger.info(msg_)
+        add_event(op=WALAEventOperation.AgentUpgrade, is_success=success_, message=msg_, log_event=False)
+
+    def run(self):
+        try:
+            # Ignore new agents if update is disabled
+            if not self.autoupdate_enabled:
+                return
+
+            agent_manifest = self.__get_agent_manifest_from_last_gs()
```
narrieta, line 240: Expressing this in terms of manifest makes the code confusing, since we do not need the agent manifests to get the required version, but just the extensions goal state.

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+    def __get_requested_version(agent_manifest):
+        """
+        Get the requested version from agent manifest
+        Returns: Requested version if supported and available
+                 None if None of the above is True
+        """
+        if conf.get_enable_ga_versioning() and agent_manifest.is_requested_version_specified:
+            if agent_manifest.requested_version is not None:
+                return FlexibleVersion(agent_manifest.requested_version)
+        return None
```
narrieta, line 114: In the new model, the requested version should always be there and if it is missing it is an error. This should raise an exception instead of return None.

Also, let's collect telemetry if the requested version is missing (this should probably be done when fetching the goal state)

narrieta (commented):
Additional comments. My review still in progress.

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+AGENT_ERROR_FILE = "error.json"  # File name for agent error record
+AGENT_MANIFEST_FILE = "HandlerManifest.json"
+MAX_FAILURE = 3  # Max failure allowed for agent before declare bad agent
+
+
+def get_agent_update_handler(protocol):
+    return AgentUpdateHandler(protocol)
+
+
+class PersistentUpdateHandlerState(object):
```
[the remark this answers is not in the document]
nagworld9: As of today, just one member and that's string with error message. For the string values I don't see good default value. Even if we do, we still need to add check to not use default value like check for None.

Motivation for new class incase if we need to add additional members like report codes etc but as of today just one. We can always create new class whenever it makes sense to create.

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+    report_error = None
+
+
+class AgentUpdateHandler(object):
+
+    def __init__(self, protocol):
+        self.__protocol = protocol
+        self.__ga_family = conf.get_autoupdate_gafamily()
+        self.__autoupdate_enabled = conf.get_autoupdate_enabled()
+        self.__gs_id = self.protocol.get_goal_state().incarnation
```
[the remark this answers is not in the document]
nagworld9: not necessarily we update on extension goal state id changes, that could fail due to transient issue. Next attempt we could update without new goal state too.

`also, it seems like it is never updated`
what are you referring to?

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+        self._last_attempted_update_time = None
```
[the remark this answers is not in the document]
nagworld9: ok

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+        self._last_attempted_update_version = FlexibleVersion("0.0.0.0")
+        self._pkg_list = None
```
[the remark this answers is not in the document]
nagworld9: Actually, this is not a list anymore, we just need one pkg which is requested version. I'll name this but still think default is None.

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+    @property
+    def ga_family(self):
+        return self.__ga_family
+
+    @property
+    def autoupdate_enabled(self):
+        return self.__autoupdate_enabled
+
+    @property
+    def gs_id(self):
```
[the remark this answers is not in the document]
nagworld9: I don't see a need to use outside of this class.

azurelinuxagent/ga/agent_update.py
nagworld9: makes sense

azurelinuxagent/ga/agent_update.py
nagworld9: ok, I'll rename

azurelinuxagent/ga/update.py
```diff
@@ -606,18 +519,18 @@ def _processing_new_extensions_goal_state(self):
...
-            self.__update_guest_agent(protocol)
-            self._report_status(exthandlers_handler)
+            agent_update_handler.run()
+            self._report_status(exthandlers_handler, agent_update_handler, gs_updated=False)
             return
 
-        # check for agent updates
-        self.__update_guest_agent(protocol)
+        gs_updated = self._processing_new_extensions_goal_state() or self._processing_new_incarnation()
+        agent_update_handler.run()
```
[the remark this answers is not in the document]
nagworld9: If that's the case, I don't need protocol at all for agent update. All that can be retrieved from goal state if we pass it. This is going to effect on test cases and lot of test cases mocked the protocol and refreshed/updated goal state through protocol.

When we do refactor, we have to update few places like where we took protocol dependency instead of goal state like ext_handler and remote handler.

I'll happy to do that now for agent update but I'm thinking is it easy and utilize efforts very effectively if we do it together or knock it out one by one when we get a chance?

azurelinuxagent/ga/update.py
```diff
@@ -664,44 +577,8 @@ def _cleanup_legacy_goal_state_history():
...
-                           textutil.format_exception(error))
-                logger.warn(err_msg)
-                add_event(op=WALAEventOperation.AgentUpgrade, is_success=False, message=err_msg, log_event=False)
-
-        return update_status
-
-    def _report_status(self, exthandlers_handler):
-        vm_agent_update_status = self.__get_vmagent_update_status(self._processing_new_extensions_goal_state())
+    def _report_status(self, exthandlers_handler, agent_update_handler, gs_updated):
+        vm_agent_update_status = agent_update_handler.get_vmagent_update_status(gs_updated=gs_updated)
```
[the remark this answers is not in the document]
narrieta: If CRP can make the decision on whether to use the status or not, let's simply send it in each status report and avoid the dependency on gs_updated (double-check with CRP fist)

azurelinuxagent/ga/update.py
narrieta: the goal state is updated as a whole, and there are never partial updates (if there are, that is a bug and we need to fix it) due to transient or permanent issues

azurelinuxagent/ga/agent_update.py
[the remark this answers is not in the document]
narrieta: initialize the message to a string indicating success, or to empty and use a explicit boolean to mark the success or failure... using None makes harder to write the code using this data

azurelinuxagent/ga/agent_update.py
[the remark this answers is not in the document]
narrieta: self.__gs_id seems to be initialized only in init and never updated
 
 there can be changes in the goal state without changes in the incarnation

azurelinuxagent/ga/agent_update.py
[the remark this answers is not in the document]
narrieta: let's make them private or protected

azurelinuxagent/ga/agent_update.py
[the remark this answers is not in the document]
narrieta: try to avoid None, makes code harder to write correclty

azurelinuxagent/ga/agent_update.py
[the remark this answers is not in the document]
nagworld9: Still, I didn't get it. why do we need to update. At this point last goal state already updated and should have latest incarnation number. But we don't the fetch the goal state if incarnation changes. So, we update and get the requested version based on last updated goal state right. 

Since agent update depends on requested version and it's coming from extension goal state. I see your point of adding Id instead of incarnation.

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+        return self.__gs_id
+
+    def __is_update_allowed(self, requested_version):
+        """
+        check to see if update is allowed once per (as specified in the conf.get_autoupdate_frequency())
+        """
+        now = time.time()
+
+        if self._last_attempted_update_time is not None and self._last_attempted_update_version == requested_version:
```
narrieta, line 69: I don't quite get this check. If we attempt the update and fail then self._last_attempted_update_version == requested_version would be True and we would keep incrementing the next time, possibly never retrying the update, wouldn't we?

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+            requested_version = self.__get_requested_version(agent_manifest)
+
+            if not requested_version or requested_version == CURRENT_VERSION:
+                return
+
+            # Check if an update is allowed
+            if not self.__is_update_allowed(requested_version):
+                return
+
+            msg_ = "Found requested version: {0} in manifest for goal state incarnation: {1} to update agent".format(
```
narrieta, line 250: we need to update this message not to refer to the manifest, since the manifest is not involved...

maybe "Goal state {0} is requesting a new agent version {1}, will update the agent before processing the goal state." ?

narrieta: And we should also send this to telemetry

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+                str(requested_version), self.gs_id)
+            self.__log_event(LogLevel.INFO, msg_)
+
+            self.__fetch_pkg_list(agent_manifest)
```
narrieta, line 254: I'd suggest removing __fetch_pkg_list and calling fetch_agent_manifest() directly, then passing the manifest to _process_update as argument

This makes cleared that we are fetching the manifest here (i.e. we are doing a network call) and expresses the dependencies more clearly, instead of keeping _pkg_list as a data member that gets updates as the side effect of doing other operations.

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+        Get the requested version from agent manifest
+        Returns: Requested version if supported and available
+                 None if None of the above is True
+        """
+        if conf.get_enable_ga_versioning() and agent_manifest.is_requested_version_specified:
+            if agent_manifest.requested_version is not None:
+                return FlexibleVersion(agent_manifest.requested_version)
+        return None
+
+    def _process_update(self, requested_version):
```
narrieta, line 116: are using "_" instead of "__" on purpose? one convention for Python code uses the former for protected members and the latter for private.

I tend to use "_", since "__" mangles the names and makes debugging harder. Since the code in the agent is not consistent, feel free to follow any convention, but let's try to be consistent across the same module

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+        """
+        This function downloads the new agent(requested version) and updates to that version.
+        """
+        package_to_download = self.__get_agent_package_to_download(requested_version)
+        agent = self.__download_and_get_agent(package_to_download)
+        # if unable to download/unpack the agent
+        if not agent or agent.version != requested_version:
```
narrieta, line 123: Does __download_and_get_agent return None if it fails to download/unpack? that should be an error that raises an exception, instead of returning None

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+            msg = "Unable to download/unpack the agent version : {0} , skipping agent update".format(requested_version)
+            self.__log_event(LogLevel.INFO, msg)
```
narrieta, line 125: should be a warning

narrieta (commented):
a few more comments

azurelinuxagent/ga/agent_update.py
nagworld9: yes, error raise by GuestAgent class and being handled in GuestAgent class itself.  In __download_and_get_agent I'm checking if is this good agent or not after download. If there are download errors, then Agent.isAvailable false and return None.

azurelinuxagent/ga/agent_update.py
nagworld9: we don't increment the actual last_attenped_update_time and compute next_time always from last_attenped_update_time.

If we don't reach the autoupdate frequency, we don't update the last_attenped_update_time and version variables.

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+
+        if self._last_attempted_update_time is not None and self._last_attempted_update_version == requested_version:
+            next_attempt_time = self._last_attempted_update_time + conf.get_autoupdate_frequency()
+        else:
+            next_attempt_time = now
+
+        if next_attempt_time > now:
+            return False
+
+        self._last_attempted_update_time = now
```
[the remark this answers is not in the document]
nagworld9: why would this changes the behavior.  We update these variables whenever we are ready to attempt update. Updating at the start or end of update is not that significant.

azurelinuxagent/ga/agent_update.py
nagworld9: I don't want to manufacture as None rather I want to handle it here and bail out but the exceptions handled in GuestAgent and don't want to change the current behavior of that because it's used in other places like in daemon when picking the latest agent.

azurelinuxagent/ga/agent_update.py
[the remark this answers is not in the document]
narrieta: You should try to keep the state of your object consistent at all times (or as much as possible). You are updating the state of the update handler, here before actually attempting the update. If an exception alters the flow of execution you may not attempt the update, even though you already updated the state.

azurelinuxagent/ga/agent_update.py
narrieta: Please do not use None to report errors. Raise an exception and handle it here.

narrieta: The comment seems to imply you are handling an error " # if unable to download/unpack the agent"

azurelinuxagent/ga/agent_update.py
narrieta: Say 

1) we get a goal state with a new requested version. We update self._last_attempted_update_version to that version, but later the update fails. 

2) we get another goal state with the same requested version. self._last_attempted_update_version == requested_version is True so we increment next_attempt_time. If next_attempt_time > now we return False

then 2) keeps happening over and over and we never update

azurelinuxagent/ga/agent_update.py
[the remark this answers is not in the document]
narrieta: (plus you are doing the updates as a side effect of checking whether the update should be attempted, that should not be a side effect of this query)

azurelinuxagent/ga/agent_update.py
nagworld9: If the auto update frequency set to 1hr. Then we don't update till that 1hr(that's the idea right). Once time elapsed will try attempt update. I don't know what we are missing here.

narrieta: i see it now, your are ok

azurelinuxagent/ga/agent_update.py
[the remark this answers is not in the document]
narrieta: the exception may occur before the actual attempt -- we should not consider that an attempt

azurelinuxagent/ga/agent_update.py
narrieta, line 123: if we are downloading only the requested version, when would agent.version != requested_version?

narrieta (commented):
few more comments

azurelinuxagent/ga/agent_update.py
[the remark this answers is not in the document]
nagworld9: I don't have any logic in between this update state and update attempt. The next immediate step is update attempt.

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+            msg = "Unable to download/unpack the agent version : {0} , skipping agent update".format(requested_version)
+            self.__log_event(LogLevel.INFO, msg)
+            return
+        # We delete the directory and the zip package from the filesystem except current version and target version
+        self.__purge_extra_agents_from_disk(known_agents=[agent])
+        self.__proceed_with_update(requested_version)
+
+    def __get_agent_package_to_download(self, requested_version):
+        """
+        Returns the package of the requested Version found in the manifest. If not found, returns exception
```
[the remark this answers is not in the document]
nagworld9: Agent pkg and it's url is required manifest.

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+        """
+        package_to_download = self.__get_agent_package_to_download(requested_version)
+        agent = self.__download_and_get_agent(package_to_download)
+        # if unable to download/unpack the agent
+        if not agent or agent.version != requested_version:
+            msg = "Unable to download/unpack the agent version : {0} , skipping agent update".format(requested_version)
+            self.__log_event(LogLevel.INFO, msg)
+            return
+        # We delete the directory and the zip package from the filesystem except current version and target version
+        self.__purge_extra_agents_from_disk(known_agents=[agent])
```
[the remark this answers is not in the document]
nagworld9: You mean we should purge() always regardless even if we get exceptions above in the process of update.

azurelinuxagent/ga/agent_update.py
nagworld9: Technically we should not see any other version and added extra safe check because the GuestAgent object init with       version = None

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+                    "Could not find a matching agent with current version {0} to blacklist, skipping it".format(
+                        CURRENT_VERSION))
+        else:
+            # In case of an upgrade, we don't need to exclude anything as the daemon will automatically
+            # start the next available highest version which would be the target version
+            prefix = "upgrade"
+        raise AgentUpgradeExitException("Agent update found, Exiting current process to {0} to the new Agent version {1}".format(prefix, requested_version))
+
+    @staticmethod
+    def __get_available_agents_on_disk():
```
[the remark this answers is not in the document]
nagworld9: is_available checks for downloaded success and not blacklisted. In this case we should consider the agents which are not downloaded properly to remove from the disk.

tests/ga/test_agent_update.py
```diff
@@ -0,0 +1,575 @@
...
+
+    def setUp(self):
+        UpdateTestCase.setUp(self)
+        # Since ProtocolUtil is a singleton per thread, we need to clear it to ensure that the test cases do not
+        # reuse a previous state
+        clear_singleton_instances(ProtocolUtil)
+
+    @contextlib.contextmanager
+    def __get_agent_update_handler(self, test_data=None, autoupdate_frequency=0.001, autoupdate_enabled=True):
+        test_data = DATA_FILE if test_data is None else test_data
```
narrieta, line 33: let's default the test_data parameter to DATA_FILE instead of doing it here. It makes the interface more explicit

tests/ga/test_agent_update.py
```diff
@@ -0,0 +1,575 @@
...
+                         "All other agents should be purged from agent dir: {0}".format(other_agents))
+
+    def test_it_should_not_update_when_autoupdate_disabled(self):
+        self.prepare_agents(count=1)
+        with self.__get_agent_update_handler(autoupdate_enabled=False) as (agent_update_handler, mock_telemetry):
+            agent_update_handler.run()
+            self.__assert_agent_directories_exist_and_others_dont_exist(versions=[str(CURRENT_VERSION)])
+            self.assertEqual(0, len([kwarg['message'] for _, kwarg in mock_telemetry.call_args_list if
+                                     "Found requested version" in kwarg['message'] and kwarg[
+                                         'op'] == WALAEventOperation.AgentUpgrade]), "Unwanted upgrade")
```
narrieta, line 78: ca we improve this message? ("Unwanted upgrade")

tests/ga/test_agent_update.py
```diff
@@ -0,0 +1,575 @@
...
+        data_file["ext_conf"] = "wire/ext_conf_requested_version.xml"
+        version = "5.2.0.1"
+        with self.__get_agent_update_handler(test_data=data_file, autoupdate_frequency=10) as (agent_update_handler, mock_telemetry):
+            with patch.object(conf, "get_enable_ga_versioning", return_value=True):
+                agent_update_handler._protocol.mock_wire_data.set_extension_config_requested_version(version)
+                agent_update_handler._protocol.mock_wire_data.set_incarnation(2)
+                agent_update_handler._protocol.update_goal_state()
+                agent_update_handler.run()
+
+                self.assertEqual(1, len([kwarg['message'] for _, kwarg in mock_telemetry.call_args_list if
```
narrieta, line 92: This search is repeated several times with different messages. Let's capture it in a helper function to improve readability of the tests

[kwarg['message'] for _, kwarg in mock_telemetry.call_args_list if
                                         "Found requested version: {0}".format(version) in kwarg[
                                             'message'] and kwarg[
                                             'op'] == WALAEventOperation.AgentUpgrade]

tests/ga/test_agent_update.py
```diff
@@ -0,0 +1,575 @@
...
+                                         "Found requested version: {0}".format(version) in kwarg[
+                                             'message'] and kwarg[
+                                             'op'] == WALAEventOperation.AgentUpgrade]), "invalid message")
```
narrieta, line 95: need to improve these messages ("invalid message")

tests/ga/test_agent_update.py
narrieta: Also, consider checking whether the agent was updated or not by other means that telemetry messages. Using messages can break if we change the message in the agent, and some test may even fail silently.

nagworld9: > This search is repeated several times with different messages. Let's capture it in a helper function to improve readability of the tests
> 
> [kwarg['message'] for _, kwarg in mock_telemetry.call_args_list if "Found requested version: {0}".format(version) in kwarg[ 'message'] and kwarg[ 'op'] == WALAEventOperation.AgentUpgrade]

yes, I realized later, doing it in next commit

nagworld9: > Also, consider checking whether the agent was updated or not by other means that telemetry messages. Using messages can break if we change the message in the agent, and some test may even fail silently.

I believe tests are meant to break if we introduce something which are not expecting. Telemetry check I added is additional checks along with others.

`some test may even fail silently.`
I don't understand this.

azurelinuxagent/common/conf.py
```diff
@@ -626,15 +626,6 @@ def get_normal_upgrade_frequency(conf=__conf__):
     return conf.get_int("Debug.AutoUpdateNormalFrequency", 24 * 60 * 60)
 
 
-def get_enable_ga_versioning(conf=__conf__):
```
nagworld9, line 629: This is not needed as we have only requested version agent version update path

azurelinuxagent/ga/agent_update.py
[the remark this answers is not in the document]
nagworld9: After thought, checks for is_available makes more sense

tests/ga/test_agent_update.py
nagworld9: Default to DATA_FILE raising pylint warning
`tests/ga/test_agent_update.py:29:4: W0102: Dangerous default value DATA_FILE (builtins.dict) as argument (dangerous-default-value)`

We can suppress but I decided not to do it unless it's absolutely necessary

azurelinuxagent/ga/guestagent.py
```diff
@@ -0,0 +1,314 @@
...
+from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
+from azurelinuxagent.common.version import AGENT_DIR_PATTERN, AGENT_NAME, CURRENT_VERSION
+from azurelinuxagent.ga.exthandlers import HandlerManifest
+
+AGENT_ERROR_FILE = "error.json"  # File name for agent error record
+AGENT_MANIFEST_FILE = "HandlerManifest.json"
+MAX_FAILURE = 3  # Max failure allowed for agent before declare bad agent
+
+
+class GAUpdateReportState(object):
```
nagworld9, line 21: This is not best place to keep this but due to circular dependencies I kept here as this is also related to agent update and also used in this module

azurelinuxagent/ga/guestagent.py
```diff
@@ -0,0 +1,314 @@
...
+            self.name,
+            self.get_agent_manifest_path())
+        logger.verbose(u"Successfully loaded Agent {0} {1}: {2}",
+                       self.name,
+                       AGENT_MANIFEST_FILE,
+                       ustr(self.manifest.data))
+        return
+
+
+class GuestAgentError(object):
```
nagworld9, line 237: GuestAgentError used in GuestAgent and no other places have a reference, so I kept in same module.

azurelinuxagent/common/conf.py
narrieta: we should still have a "kill switch" for the feature, let's discuss this offline

tests/ga/test_agent_update.py
narrieta: I see. OK, let's add a doc string describing the default

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,238 @@
...
+    def get_vmagent_update_status(self, goal_state):
+        """
+        This function gets the VMAgent update status as per the last GoalState.
+        Returns: None if the last GS does not ask for requested version else VMAgentUpdateStatus
+        """
+        try:
+            agent_family = self.__get_agent_family_from_last_gs(goal_state)
+            target_version = self.__get_requested_version(agent_family)
+
+            if target_version is not None:
```
narrieta, line 224: move this to run()

azurelinuxagent/ga/update.py
```diff
@@ -606,18 +519,18 @@ def _processing_new_extensions_goal_state(self):
...
-    def _process_goal_state(self, exthandlers_handler, remote_access_handler):
+    def _process_goal_state(self, exthandlers_handler, remote_access_handler, agent_update_handler):
         protocol = exthandlers_handler.protocol
 
         # update self._goal_state
         if not self._try_update_goal_state(protocol):
             # agent updates and status reporting should be done even when the goal state is not updated
-            self.__update_guest_agent(protocol)
-            self._report_status(exthandlers_handler)
+            agent_update_handler.run()
```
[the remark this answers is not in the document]
narrieta: TBD

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,238 @@
...
+            if not self.__should_update_agent(requested_version):
+                return
+
+            msg_ = "Goal state {0} is requesting a new agent version {1}, will update the agent before processing the goal state.".format(self._gs_id, str(requested_version))
+            self.__log_event(LogLevel.INFO, msg_)
+
+            agent = self.__download_and_get_agent(goal_state, agent_family, requested_version)
+
+            if not agent.is_available:
+                msg = "Unable to download/unpack the agent version : {0} , skipping agent update".format(
```
narrieta, line 200: we need to update this message since this is not a download/unpack the agent

tests/ga/test_update.py
```diff
@@ -1880,7 +1484,7 @@ def __assert_agent_directories_exist_and_others_dont_exist(self, versions):
 
     def __assert_no_agent_upgrade_telemetry(self, mock_telemetry):
         self.assertEqual(0, len([kwarg['message'] for _, kwarg in mock_telemetry.call_args_list if
-                                 "Agent upgrade discovered, updating to" in kwarg['message'] and kwarg[
+                                 "Agent update found, Exiting current process" in kwarg['message'] and kwarg[
                                      'op'] == WALAEventOperation.AgentUpgrade]), "Unwanted upgrade")
```
narrieta, line 1488: Let's improve this message: "Unwanted upgrade"

tests/ga/test_agent_update.py
narrieta: If we update the telemetry message, the asserts that check for the presence of the message (self.assertEqual(1, len(upgrade_event_msgs)) will fail on those are easy to fix.

However, the tests that check for the absence of the message (len == 0) will continue to pass, event if we introduce a bug .

tests/ga/test_update.py
```diff
@@ -1802,20 +1419,16 @@ def test_run_emits_restart_event(self):
...
-                            yield
+            with patch("azurelinuxagent.common.conf.get_autoupdate_frequency", return_value=autoupdate_frequency):
+                with patch("azurelinuxagent.common.conf.get_autoupdate_gafamily", return_value="Prod"):
+                    yield
 
     @contextlib.contextmanager
-    def __get_update_handler(self, iterations=1, test_data=None, hotfix_frequency=1.0, normal_frequency=2.0,
-                             reload_conf=None):
+    def __get_update_handler(self, iterations=1, test_data=None,
+                             reload_conf=None, autoupdate_frequency=0.001):
```
maddieford, line 1431: What unit is this frequency?

azurelinuxagent/ga/agent_update.py
nagworld9: All we do is download and unpack the agent when we call GuestAgent

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+    def autoupdate_enabled(self):
+        return self.__autoupdate_enabled
+
+    @property
+    def gs_id(self):
+        return self.__gs_id
+
+    def __is_update_allowed(self, requested_version):
+        """
+        check to see if update is allowed once per (as specified in the conf.get_autoupdate_frequency())
```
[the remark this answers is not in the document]
nagworld9: Yes, per hour by default but users can configure it.

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,577 @@
...
+            if target_version is not None:
+                if CURRENT_VERSION == target_version:
+                    status = VMAgentUpdateStatuses.Success
+                    code = 0
+                    message = ""
+                else:
+                    status = VMAgentUpdateStatuses.Error
+                    code = 1
+                    message = PersistentUpdateHandlerState.report_error if PersistentUpdateHandlerState.report_error else ""
+                PersistentUpdateHandlerState.report_error = None
```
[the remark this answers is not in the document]
nagworld9: yes

tests/ga/test_update.py
nagworld9: secs

narrieta (approved):
Thanks for the updates

maddieford (approved):

PR #2707, opened 2022-11-30
Publish test results and logs

narrieta:
Publish test results and logs to the Azure Pipeline.

Also, given the limitations on LISA's TestSuite, re-organized the agent BVTs.

### DIFF
--- a/tests_e2e/lisa/testsuites/agent-bvt.py
+++ b/tests_e2e/lisa/testsuites/agent-bvt.py
@@ -1,17 +1,13 @@
 from assertpy import assert_that
-from pathlib import Path
+
+from tests_e2e.lisa.testsuites.agent_test_suite import AgentTestSuite
 from tests_e2e.lisa.tests.agent_bvt import custom_script
 
 from lisa import (
-    CustomScriptBuilder,
-    Logger,
-    Node,
     simple_requirement,
     TestCaseMetadata,
-    TestSuite,
     TestSuiteMetadata,
 )
-from lisa.sut_orchestrator.azure.common import get_node_context
 
 
 @TestSuiteMetadata(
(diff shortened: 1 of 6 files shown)


tests_e2e/lisa/testsuites/agent-bvt.py
```diff
@@ -22,20 +18,17 @@
     """,
     requirement=simple_requirement(unsupported_os=[]),
 )
-class AgentBvt(TestSuite):
```
narrieta, line 25: Given the limitations on LISA's TestSuite, now I am modeling a DCR scenario as a TestSuite with a single test (main), which invokes each of the steps in the scenario.

tests_e2e/lisa/testsuites/agent_test_suite.py
```diff
@@ -0,0 +1,53 @@
+from pathlib import Path, PurePath
```
narrieta, line 1: Currently just a skeleton, needs many improvements

tests_e2e/scripts/collect_logs.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
narrieta, line 1: executed on the test VM to collect agent and system logs into a compeessed tarball

maddieford (approved):

tests_e2e/scripts/execute_tests_container.sh
```diff
@@ -8,20 +8,34 @@ az acr login --name waagenttests
...
+#    .../20221130/20221130-160013-749
+#    .../20221130/20221130-160013-749/environments
+#    .../20221130/20221130-160013-749/lisa-20221130-160013-749.log
+#    .../20221130/20221130-160013-749/lisa.junit.xml
+#    etc
+#
+# Remove the first 2 levels of the tree (which indicate the time of the test run) to make navigation
+# in the Azure Pipelines UI easier.
+#
+mv "$BUILD_ARTIFACTSTAGINGDIRECTORY"/[0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]/*/* "$BUILD_ARTIFACTSTAGINGDIRECTORY"
```
nagworld9, line 40: good point, I noticed this.

tests_e2e/templates/execute-tests.yml
```diff
@@ -32,3 +32,14 @@ jobs:
           AZURE_CLIENT_SECRET: $(AZURE-CLIENT-SECRET)
           AZURE_TENANT_ID: $(AZURE-TENANT-ID)
           SUBSCRIPTION_ID: $(SUBSCRIPTION-ID)
+
+      - task: PublishTestResults@2
```
nagworld9, line 36: I have general question regarding junit test results. Based on the current design when do we say test_case failed, is it if all the steps failed in test case or any of the step failed? 

Like dcr today do we show warning/failed status on test case if any of them failed, so that it's easy to navigate to that test case and dig the logs to see the errors.

narrieta: each scenario needs to define what conditions should be treated as an error (i.e. the test case must fail) or as warnings (i.e. the test passes, but there are some warning messages in the logs).

dcr is too limited in that it does not make this distinction and it is hard to know what is an actual failure without checking the logs in detail

nagworld9: yes, but how do you pass this information to junit output, now junit output control by lisa and lisa determines the test case result.

narrieta: just pass or fail the test, not sure i get your question

nagworld9: I understand pass or fail. It's straightforward. I'm curious about warning case If we make test case pass for warnings, then azure pipeline shows green and it's hard to know particular test case has warnings in log without going in each test case log.

narrieta: warning would be something that does not prevent a test from executing/succeeding. when a test fails. warnings can help determine why. 

not sure if there is a way to report them to the pipeline. but i am not sure we want to since they are only interesting when something fails

nagworld9: if we consider one of step failure as warnings to continue execute rest of the steps, in those cases we still need to follow up why particular step as warning to make sure no regressions.

I'm just throwing point; this is something to think about in the grand scheme of things.

nagworld9 (approved):

tests_e2e/templates/execute-tests.yml
narrieta: not sure i understand what you are trying to achieve, but you can always reach out offline

PR #2708, opened 2022-12-06
Reorganize directory structure of end-to-end tests

narrieta:
The initial directory structure came straight from the dcr prototype. This PR creates a better structure to organize the code. The only code changes are to update paths.

* docker - Dockefile and any other files we may need to maintain the test containers
* pipeline - Configuration and scripts to support/interface with the Azure Pipeline
* scenarios - Test scenarios


nagworld9 (approved):
so, the structure for scripts outside of scenarios used/run by azure pipeline and scenario scripts used/run by lisa

maddieford (approved):

PR #2713, opened 2022-12-08
Log collector should not fetch full goal state with extensions

maddieford:
## Description
The Log Collector should not fetch the full goal state with extensions and certificates. This can cause a race condition between the log collector and extension handler when downloading certificates.

This PR adds a GoalStateProperties enum which defines the properties that can be fetched in the goal state and updates fetch_full_wire_server_goal_state to take a list of properties from the goal state to populate. This supports fetching different combinations of propreties in the goal state.

Issue # 

---

### PR information


azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -48,6 +48,22 @@
...
+    Enum for defining the properties that we fetch in the goal state
+    """
+    RoleConfig = 1
+    HostingEnv = 2
+    SharedConfig = 4
+    ExtensionsConfig_Certs = 8
+    RemoteAccessInfo = 16
+
+    @staticmethod
+    def default_properties():
```
narrieta, line 62: This is OK, though "default" in the name doesn't quite fit.

An alternative is to add an "All" item to the enum.

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -48,6 +48,22 @@
...
 
+class GoalStateProperties(object):
+    """
+    Enum for defining the properties that we fetch in the goal state
+    """
+    RoleConfig = 1
+    HostingEnv = 2
+    SharedConfig = 4
+    ExtensionsConfig_Certs = 8
+    RemoteAccessInfo = 16
```
narrieta, line 59: Minor: using hex instead of decimal can help visualize exactly which bit is being used for the value.

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -160,20 +176,20 @@ def update_host_plugin_headers(wire_client):
         # Fetching the goal state updates the HostGAPlugin so simply trigger the request
         GoalState._fetch_goal_state(wire_client)
 
-    def update(self, silent=False):
+    def update(self, goalstate_properties=GoalStateProperties.default_properties(), silent=False):
```
narrieta, line 179: The update method should probably not take the properties as parameter, since it opens the possibility of passing a value different to the one passed to init() and that would lead to an inconsistent goal state.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -72,7 +73,7 @@ def __init__(self, endpoint):
             raise ProtocolError("WireProtocol endpoint is None")
         self.client = WireClient(endpoint)
 
-    def detect(self):
+    def detect(self, goalstate_properties=GoalStateProperties.default_properties()):
```
narrieta, line 76: No need to expose the properties in this method

azurelinuxagent/ga/update.py
```diff
@@ -485,7 +485,7 @@ def _try_update_goal_state(self, protocol):
         try:
             max_errors_to_log = 3
 
-            protocol.update_goal_state(silent=self._update_goal_state_error_count >= max_errors_to_log)
+            protocol.client.update_goal_state(silent=self._update_goal_state_error_count >= max_errors_to_log)
```
maddieford, line 488: @narrieta On line 430 in update.py we discussed changing 
`self._try_update_goal_state(protocol)`
to 
`self._reset_goal_state()`

It looks like we sleep until try_update_goal_state returns True for success.. Should we implement similar logic for reset_goal_state?

azurelinuxagent/common/protocol/wire.py
maddieford: This has been removed

azurelinuxagent/common/protocol/goal_state.py
maddieford: This has been removed

azurelinuxagent/common/protocol/goal_state.py
maddieford: Updated

azurelinuxagent/common/protocol/goal_state.py
maddieford: Updated to add All item

azurelinuxagent/common/protocol/util.py
```diff
@@ -188,7 +189,7 @@ def _clear_wireserver_endpoint(self):
                 return
             logger.error("Failed to clear wiresever endpoint: {0}", e)
 
-    def _detect_protocol(self):
+    def _detect_protocol(self, goalstate_properties=GoalStateProperties.default_properties()):
```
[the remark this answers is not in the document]
maddieford: Removed goal_state_properties from detect

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -343,7 +359,7 @@ def _fetch_vm_settings(wire_client, force_update=False):
 
         return vm_settings, vm_settings_updated
 
-    def _fetch_full_wire_server_goal_state(self, incarnation, xml_doc):
+    def _fetch_full_wire_server_goal_state(self, incarnation, xml_doc, goalstate_properties=GoalStateProperties.default_properties()):
```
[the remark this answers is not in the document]
maddieford: Removed goal_state_properties parameter for this method and created private member for the goal state class

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -48,6 +48,22 @@
...
 
 
+class GoalStateProperties(object):
+    """
+    Enum for defining the properties that we fetch in the goal state
+    """
+    RoleConfig = 1
+    HostingEnv = 2
+    SharedConfig = 4
+    ExtensionsConfig_Certs = 8
```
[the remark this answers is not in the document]
maddieford: Updated to separate the two

azurelinuxagent/common/logcollector.py
```diff
@@ -118,7 +119,7 @@ def _set_resource_usage_cgroups(cpu_cgroup_path, memory_cgroup_path):
     @staticmethod
     def _initialize_telemetry():
         protocol = get_protocol_util().get_protocol()
```
nagworld9, line 121: Looks like get_protocal() still initializing goal state with all properties?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -778,18 +776,30 @@ def update_host_plugin(self, container_id, role_config_name):
             self._host_plugin.update_container_id(container_id)
             self._host_plugin.update_role_config_name(role_config_name)
 
-    def update_goal_state(self, force_update=False, silent=False):
+    def update_goal_state(self, silent=False):
```
nagworld9, line 779: I don't quite get the update_goal_state vs reset_goal_state. When do we use each of these?

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -99,35 +113,59 @@ def incarnation(self):
 
     @property
     def container_id(self):
-        return self._container_id
+        if not self._goal_state_properties & GoalStateProperties.RoleConfig:
+            raise ProtocolError("RoleConfig is not in goal state properties")
```
narrieta, line 117: container id

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -99,35 +113,59 @@ def incarnation(self):
...
+        else:
+            return self._container_id
 
     @property
     def role_instance_id(self):
-        return self._role_instance_id
+        if not self._goal_state_properties & GoalStateProperties.RoleConfig:
+            raise ProtocolError("RoleConfig is not in goal state properties")
```
narrieta, line 124: role instance id

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -99,35 +113,59 @@ def incarnation(self):
...
-        return self._role_config_name
+        if not self._goal_state_properties & GoalStateProperties.RoleConfig:
+            raise ProtocolError("RoleConfig is not in goal state properties")
+        else:
+            return self._role_config_name
 
     @property
     def extensions_goal_state(self):
-        return self._extensions_goal_state
+        if not self._goal_state_properties & GoalStateProperties.ExtensionsConfig:
```
narrieta, line 137: Ah, i did not notice this before. the extensions goal state may come from vmSettings or ExtensionsConfig. The latter is for fabric and the former for fast track.

Let's rename GoalStateProperties.ExtensionsConfig to GoalStateProperties.ExtensionsGoalState

narrieta: Also, in _update(), we need to retrieve the vmSettings only when requesting GoalStateProperties.ExtensionsGoalState

narrieta (commented):
A few comments on goal_state.py, I'm reviewing the usages of the goal state next

azurelinuxagent/common/protocol/wire.py
```diff
@@ -778,18 +776,30 @@ def update_host_plugin(self, container_id, role_config_name):
...
         """
-        Updates the goal state if the incarnation or etag changed or if 'force_update' is True
+        Updates the goal state if the incarnation or etag changed
         """
         try:
-            if force_update and not silent:
+            if not silent:
```
narrieta, line 784: we can remove this message since the force flag was removed

azurelinuxagent/common/protocol/wire.py
```diff
@@ -778,18 +776,30 @@ def update_host_plugin(self, container_id, role_config_name):
...
                 logger.info("Forcing an update of the goal state.")
 
-            if self._goal_state is None or force_update:
-                self._goal_state = GoalState(self, silent=silent)
-            else:
-                self._goal_state.update(silent=silent)
+            self._goal_state.update(silent=silent)
```
narrieta, line 787: I think we need to keep a call to reset here (minus the check for the force flag). This is needed by the initialization of the goal state on line 430 that you pointed out

```
            if self._goal_state is None:
                self._goal_state = GoalState(self, silent=silent)
            else:
                self._goal_state.update(silent=silent)

```

azurelinuxagent/ga/update.py
narrieta: good point. i added a comment on _try_update_goal_state

azurelinuxagent/common/protocol/wire.py
maddieford: We're removing the force_update parameter and using reset_goal_state in all cases where force_update was True. 

Reset will initialize the goal state instead of updating

azurelinuxagent/common/logcollector.py
maddieford: Added flag to get_protocol()

azurelinuxagent/common/protocol/goal_state.py
maddieford: Updated to add this check

azurelinuxagent/common/protocol/wire.py
maddieford: I added the None check back

narrieta (approved):

PR #2725, opened 2023-01-10
Add support to execute tests on multiple distros

narrieta:
The PR includes support to test execution on multiple distros. The working directory needs to account for multiple test suites executed concurrently, the log needs to add a field identifying the test suite and distro, and there are a few distro-compatibility fixes in the remote scripts.

The PR also includes a fix for an issue that was preventing log collection in some cases.

The list of distros in this PR matches the DCR v2 prototype, but additional changes will be needed later.


tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -40,121 +42,147 @@
 from tests_e2e.scenarios.lib.logging import log
 
 
-class AgentLisaTestContext(AgentTestContext):
```
narrieta, line 43: I made this class local to AgentTestSuite

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -40,121 +42,147 @@
...
-        """
-        Prepares the test suite for execution
-        """
-        log.info("Test Node: %s", self.context.vm.name)
-        log.info("Resource Group: %s", self.context.vm.resource_group)
-        log.info("Working directory: %s", self.context.working_directory)
+        self.__context.node = node
+        self.__context.suite_name = f"{self._metadata.full_name}:{runbook.marketplace.offer}-{runbook.marketplace.sku}"
 
-        if self.context.working_directory.exists():
```
narrieta, line 103: The cleanup of the working directory was getting in the way of debugging, since sometimes one needs to check its contents. Instead of cleaning up, now I use a unique working directory per runbook execution.

The cleanup was mainly targeted to the developer scenario, so with these changes development machines may start accumulating many test files. That can be handled by our helper scripts.

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -40,121 +42,147 @@
...
+        """
+        Returns the path to the agent package.
+        """
+        return self.context.working_directory/"eggs"/f"WALinuxAgent-{AGENT_VERSION}.zip"
 
-        return package_path
+    def _clean_up(self) -> None:
+        """
+        Cleans up any leftovers from the test suite run. Currently just an empty placeholder for future use.
+        """
```
narrieta, line 157: The cleanup used to remove the working directory. Since this is part of the protocol enforced by self.execute(), I left it here as an empty method in case we need it in the future.

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -233,15 +260,21 @@ def execute(self, node: Node, test_suite: List[Type[AgentTest]]) -> None:
...
+            raise
+
         finally:
             self._clean_up()
+            current_thread().name = thread_name
 
         # Fail the entire test suite if any test failed
-        assert_that(failed).described_as("One or more tests failed").is_length(0)
+        if len(failed) > 0:
+            fail(f"{[self.context.suite_name]} One or more tests failed: {failed}")
```
narrieta, line 277: Using fail() instead of assert)that() improves the formatting of the error message a little bit.

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -259,17 +292,16 @@ def execute_script_on_node(self, script_path: Path, parameters: str = "", sudo:
...
         if result.stdout != "":
             separator = "\n" if "\n" in result.stdout else " "
             log.info("stdout:%s%s", separator, result.stdout)
         if result.stderr != "":
             separator = "\n" if "\n" in result.stderr else " "
             log.error("stderr:%s%s", separator, result.stderr)
 
+        if result.exit_code != 0:
+            raise Exception(f"[{command_line}] failed. Exit code: {result.exit_code}")
+
```
narrieta, line 304: We need to raise the exception *after* logging the command's output

tests_e2e/orchestrator/scripts/collect-logs
```diff
@@ -2,24 +2,33 @@
 #
 # Collects the logs needed to debug agent issues into a compressed tarball.
 #
-set -euxo pipefail
 
-logs_file_name="$HOME/logs.tgz"
+# Note that we do "set -euxo pipefail" only after executing "tar". That command exits with code 1 on warnings
+# and we do not want to consider those as failures.
+
```
narrieta, line 8: This is what was preventing log collection on warnings

tests_e2e/orchestrator/scripts/install-agent
```diff
@@ -51,9 +51,22 @@ if [ "$#" -ne 0 ] || [ -z ${package+x} ] || [ -z ${version+x} ]; then
 fi
 
 #
-# The service name is walinuxagent in Ubuntu and waagent elsewhere
+# Find the command to manage services
```
narrieta, line 54: Ubuntu has a 'service' wrapper on top of 'systemctl' but some of the distros I added don't, so our script needs to handle that difference.

Note that other distros that we will add do not use systemd, and services are managed with the 'service' command.

tests_e2e/scenarios/runbooks/daily.yml
```diff
@@ -1,29 +1,23 @@
-name: azure
+name: Daily
```
narrieta, line 1: I was using the daily.yml both for creating new VMs or use existing ones. With the addition of the combinator to execute multiple distros that is no longer possible.

I split the existing VM setup into a new runbook in the samples directory.

tests_e2e/scenarios/runbooks/daily.yml
```diff
@@ -64,19 +48,27 @@ platform:
...
+concurrency: 10
+
+notifier:
+  - type: env_stats
+  - type: junit
+
+include:
+  - path: ./include/ssh_proxy.yml
 
-#
```
narrieta, line 74: The ssh proxy settings will be shared by multiple runbooks, so I refactored it into its own file

tests_e2e/docker/Dockerfile
```diff
@@ -23,14 +23,20 @@ RUN \
     #                                                                                                            \
     # Install basic dependencies                                                                                 \
     #                                                                                                            \
-    apt-get install -y git python3.10 python3.10-dev curl                                                     && \
```
narrieta, line 26: curl was previously use to download Poetry in order to setup LISA, but their new installation is not based on Poetry so curl is no longer needed

tests_e2e/docker/Dockerfile
```diff
@@ -23,14 +23,20 @@ RUN \
...
+    apt-get install -y git python3.10 python3.10-dev                                                          && \
     ln /usr/bin/python3.10 /usr/bin/python3                                                                   && \
                                                                                                                  \
     #                                                                                                            \
     # Install LISA dependencies                                                                                  \
     #                                                                                                            \
     apt-get install -y git gcc libgirepository1.0-dev libcairo2-dev qemu-utils libvirt-dev                       \
             python3-pip python3-venv                                                                          && \
+                                                                                                                 \
```
narrieta, line 34: The latest Ubuntu image on Dockerhub doesn't have zip, so installing explicitly

tests_e2e/orchestrator/scripts/collect-logs
```diff
@@ -2,24 +2,33 @@
...
-if [ "$exit_code" != "1" ] && [ "$exit_code" != "0" ]; then
+
+if [ "$exit_code" == "1" ]; then
+  echo "WARNING: tar exit code is 1"
+elif [ "$exit_code" != "0" ]; then
     exit $exit_code
 fi
 
-chmod +r "$logs_file_name"
+chmod a+r "$logs_file_name"
```
narrieta, line 32: "a" (i.e. "all") needs to be given explicitly, since Mariner's UMASK turns the read permission off

maddieford (approved):

nagworld9 (approved):
LGTM

PR #2731, opened 2023-01-17
Create individual log files for each (distro, test suite) pair

narrieta:
Now we create separate log files for each pair of (distro, test suite).

Test suites are executed concurrently on different threads. These changes include a thread-safe logging.Handler that multiplexes log output to the proper file. When writing multi-threaded tests a little extra setup is needed; I'll document that in the wiki and go over the documentation with the team.

### DIFF
--- a/tests_e2e/orchestrator/lib/agent_test_suite.py
+++ b/tests_e2e/orchestrator/lib/agent_test_suite.py
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import logging
 import re
 
 from assertpy import fail
@@ -27,6 +28,7 @@ from typing import List, Type
 #     E0401: Unable to import 'lisa.sut_orchestrator.azure.common' (import-error)
 from lisa import (  # pylint: disable=E0401
     CustomScriptBuilder,
+    Logger,
     Node,
     TestSuite,
     TestSuiteMetadata
@@ -39,7 +41,31 @@ from azurelinuxagent.common.version import AGENT_VERSION
 from tests_e2e.scenarios.lib.agent_test import AgentTest
 from tests_e2e.scenarios.lib.agent_test_context import AgentTestContext
 from tests_e2e.scenarios.lib.identifiers import VmIdentifier
-from tests_e2e.scenarios.lib.logging import log
+from tests_e2e.scenarios.lib.logging import log as agent_test_logger  # Logger used by the tests
+
+
+def _initialize_lisa_logger():
+    """
+    Customizes the LISA logger.
+
+    The default behavior of this logger is too verbose, which makes reading the logs difficult. We set up a more succinct
+    formatter and decrease the log level to INFO (the default is VERBOSE). In the future we may consider making this
+    customization settable at runtime in case we need to debug LISA issues.
+    """
+    logger: Logger = logging.getLogger("lisa")
+
+    logger.setLevel(logging.INFO)
+
+    formatter = logging.Formatter('%(asctime)s.%(msecs)03d [%(levelname)s] [%(threadName)s] %(message)s', datefmt="%Y-%m-%dT%H:%M:%SZ")
+    for handler in logger.handlers:
+        handler.setFormatter(formatter)
+
+
+#
+# We want to customize the LISA logger as early as possible, so we do it when this module is first imported. That will
+# happen early in the LISA workflow, when it loads the test suites to execute.
+#
+_initialize_lisa_logger()
 
 
 class AgentTestSuite(TestSuite):
@@ -52,6 +78,7 @@ class AgentTestSuite(TestSuite):
         def __init__(self, vm: VmIdentifier, paths: AgentTestContext.Paths, connection: AgentTestContext.Connection):
             super().__init__(vm=vm, paths=paths, connection=connection)
             # These are initialized by AgentTestSuite._set_context().
+            self.log: Logger = None
             self.node: Node = None
             self.runbook_name: str = None
             self.suite_name: str = None
@@ -61,7 +88,7 @@ class AgentTestSuite(TestSuite):
         # The context is initialized by _set_context() via the call to execute()
         self.__context: AgentTestSuite._Context = None
 
-    def _set_context(self, node: Node):
+    def _set_context(self, node: Node, log: Logger):
         connection_info = node.connection_info
         node_context = get_node_context(node)
         runbook = node.capability.get_extended_runbook(AzureNodeSchema, AZURE)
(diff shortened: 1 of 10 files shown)


tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -39,7 +41,31 @@
 from tests_e2e.scenarios.lib.agent_test import AgentTest
 from tests_e2e.scenarios.lib.agent_test_context import AgentTestContext
 from tests_e2e.scenarios.lib.identifiers import VmIdentifier
-from tests_e2e.scenarios.lib.logging import log
+from tests_e2e.scenarios.lib.logging import log as agent_test_logger  # Logger used by the tests
+
+
+def _initialize_lisa_logger():
```
narrieta, line 47: Previously this was done in tests_e2e/scenarions/lib/logging.py as part of the setup for the global logger. Since now I separated the LISA log from the test logs, I moved this to this module, which is our interface with LISA.

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -84,15 +111,23 @@ def _set_context(self, node: Node):
                 private_key_file=connection_info['private_key_file'],
                 ssh_port=connection_info['port']))
 
+        self.__context.log = log
         self.__context.node = node
-        self.__context.suite_name = f"{self._metadata.full_name}:{runbook.marketplace.offer}-{runbook.marketplace.sku}"
+        self.__context.suite_name = f"{self._metadata.full_name}_{runbook.marketplace.offer}-{runbook.marketplace.sku}"
```
narrieta, line 116: '-' instead of ':' is friendlier for file names

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -84,15 +111,23 @@ def _set_context(self, node: Node):
...
             raise Exception("The context for the AgentTestSuite has not been initialized")
         return self.__context
 
+    @property
+    def _log(self) -> Logger:
+        """
+        Returns a reference to the LISA Logger.
+        """
+        return self.context.log
+
```
narrieta, line 130: self._log is the LISA Logger, while agent_test_logger is the global test Logger (which redirects output to the log file for each suite)

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -174,103 +209,118 @@ def _install_agent_on_node(self) -> None:
...
-        failed: List[str] = []
+        failed: List[str] = []  # List of failed tests (names only)
 
+        # The thread name is added to self._log, set it to the current test suite while we execute it
         thread_name = current_thread().name
         current_thread().name = self.context.suite_name
 
+        # We create a separate log file for the test suite.
+        suite_log_file: Path = Path.home()/'logs'/f"{self.context.suite_name}.log"
+        agent_test_logger.set_current_thread_log(suite_log_file)
```
narrieta, line 258: This sets the log file for the current suite

tests_e2e/orchestrator/scripts/run-scenarios
```diff
@@ -35,10 +35,15 @@ cp "$HOME/id_rsa" "$HOME/.ssh"
 chmod 700 "$HOME/.ssh/id_rsa"
 ssh-keygen -y -f "$HOME/.ssh/id_rsa" > "$HOME/.ssh/id_rsa.pub"
 
+#
 # Now start the runbook
+#
+lisa_logs="$HOME/logs/lisa"
```
narrieta, line 41: Since now we have multiple logs for the full test run, I moved the LISA logs to their own subdirectory. The main LISA log is directly under this subdirectory.

tests_e2e/scenarios/tests/error_test.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python3
```
narrieta, line 1: We have similar tests in DCR. They are very useful when debugging issues in the test pipeline.

tests_e2e/scenarios/lib/logging.py
```diff
@@ -14,24 +14,110 @@
...
+    AgentLoggingHandler is meant to be primarily used in multithreaded scenarios and is thread-safe.
+    """
+    def __init__(self):
+        super().__init__()
+        self.formatter: Formatter = Formatter('%(asctime)s.%(msecs)03d [%(levelname)s] %(message)s', datefmt="%Y-%m-%dT%H:%M:%SZ")
+        self.default_handler = StreamHandler()
+        self.default_handler.setFormatter(self.formatter)
+        self.per_thread_handlers: Dict[int, FileHandler] = {}
+
+    def set_thread_log(self, thread_ident: int, log_file: Path) -> None:
```
maddieford, line 53: How is the thread_ident determined?

maddieford (dismissed):

tests_e2e/scenarios/lib/logging.py
narrieta: It is a property of Python's Thread objects. 

This method is meant to be used by tests that start multiple threads (currently we don't many of those) and want the child threads to write to the test's log file. The parent thread would create a new Thread, set the log file using Thread.ident and then start the child thread with Thread.start(). The child thread should call close_current_thread_log() when it completes. I'll document this protocol in the wiki.

Note that _AgentLoggingHandler is not a public class and is not meant to be used outside this module; it is simply an implementation detail of the AgentLogger. Users of the logger would call set_thread_log(), etc on the logger, which simply forwards to the _AgentLoggingHandler.

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -174,103 +209,118 @@ def _install_agent_on_node(self) -> None:
...
+                    self._log.info("******** %s", result)
+                    results.append(result)
+
+                agent_test_logger.info("")
+                agent_test_logger.info("********* [Test Results]")
+                agent_test_logger.info("")
                 for r in results:
-                    log.info("\t%s", r)
-                log.info("")
+                    agent_test_logger.info("\t%s", r)
```
nagworld9, line 303: why not added to main log? we know the summary in the main log too?

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -174,103 +209,118 @@ def _install_agent_on_node(self) -> None:
...
-                        log.info("")
-
-                        results.append(result)
+                        result = f"[Passed] {test_full_name}"
+                    except AssertionError as e:
+                        failed.append(test.__name__)
+                        result = f"[Failed] {test_full_name}"
+                        self._log.error("%s", e)
+                        agent_test_logger.error("%s", e)
+                        self._log.error("%s", e)
```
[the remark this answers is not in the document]
narrieta: thanks; removed

tests_e2e/orchestrator/lib/agent_test_suite.py
narrieta: it doesn't help there. it's a multi-line report and the output from different threads makes it hard to read, plus it won't always be at the end of the log, it'll be lost with the execution for other distros.

while executing the test suite, the purpose of the main log is just to show the progress of the test suite run. i duplicated test errors in the main log because that may be more convenient when checking the overall status of the test run.

nagworld9 (approved):

maddieford (approved):

PR #2732, opened 2023-01-19
Bug fixes for test results

narrieta:
* A test run wuth failing tests were previously reported as Success
* The distro name was missing in the test report

### DIFF
--- a/tests_e2e/orchestrator/scripts/run-scenarios
+++ b/tests_e2e/orchestrator/scripts/run-scenarios
@@ -45,5 +45,5 @@ lisa \
   --log_path "$lisa_logs" \
   --working_path "$lisa_logs" \
   -v subscription_id:"$SUBSCRIPTION_ID" \
-  -v identity_file:"$HOME/.ssh/id_rsa" \
-  || true # force a success exit code to allow execution to continue when a test fails
+  -v identity_file:"$HOME/.ssh/id_rsa"
+
--- a/tests_e2e/pipeline/templates/execute-tests.yml
+++ b/tests_e2e/pipeline/templates/execute-tests.yml
@@ -26,6 +26,7 @@ jobs:
 
       - bash: $(Build.SourcesDirectory)/tests_e2e/pipeline/scripts/execute_tests.sh
         displayName: "Execute tests"
+        continueOnError: true
         env:
           # Add all KeyVault secrets explicitly as they're not added by default to the environment vars
           AZURE_CLIENT_ID: $(AZURE-CLIENT-ID)
@@ -33,13 +34,15 @@ jobs:
           AZURE_TENANT_ID: $(AZURE-TENANT-ID)
           SUBSCRIPTION_ID: $(SUBSCRIPTION-ID)
 
+      - publish: $(Build.ArtifactStagingDirectory)
+        artifact: 'artifacts'
+        displayName: 'Publish test artifacts'
+
       - task: PublishTestResults@2
+        displayName: 'Publish test results'
         inputs:
           testResultsFormat: 'JUnit'
-          testResultsFiles: '**/*junit.xml'
+          testResultsFiles: 'lisa/agent.junit.xml'
           searchFolder: $(Build.ArtifactStagingDirectory)
-          testRunTitle: 'Publish test results'
+          failTaskOnFailedTests: true
 
-      - publish: $(Build.ArtifactStagingDirectory)
-        artifact: 'test-logs'
-        displayName: 'Publish test logs'
--- a/tests_e2e/scenarios/testsuites/agent_junit.py
+++ b/tests_e2e/scenarios/testsuites/agent_junit.py
@@ -1,0 +1,50 @@
+# Microsoft Azure Linux Agent
+#
+# Copyright 2018 Microsoft Corporation
+#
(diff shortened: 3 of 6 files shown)


tests_e2e/orchestrator/scripts/run-scenarios
```diff
@@ -45,5 +45,5 @@ lisa \
   --log_path "$lisa_logs" \
   --working_path "$lisa_logs" \
   -v subscription_id:"$SUBSCRIPTION_ID" \
-  -v identity_file:"$HOME/.ssh/id_rsa" \
-  || true # force a success exit code to allow execution to continue when a test fails
```
narrieta, line 49: now this is handled in execute_tests.sh

tests_e2e/pipeline/templates/execute-tests.yml
```diff
@@ -26,20 +26,23 @@ jobs:
 
       - bash: $(Build.SourcesDirectory)/tests_e2e/pipeline/scripts/execute_tests.sh
         displayName: "Execute tests"
+        continueOnError: true
```
narrieta, line 29: Test failures should not stop the pipeline, so we allow this task to continue on error

tests_e2e/pipeline/templates/execute-tests.yml
```diff
@@ -26,20 +26,23 @@ jobs:
...
+
       - task: PublishTestResults@2
+        displayName: 'Publish test results'
         inputs:
           testResultsFormat: 'JUnit'
-          testResultsFiles: '**/*junit.xml'
+          testResultsFiles: 'lisa/agent.junit.xml'
           searchFolder: $(Build.ArtifactStagingDirectory)
-          testRunTitle: 'Publish test results'
+          failTaskOnFailedTests: true
```
narrieta, line 47: If the Execute Tests tasks fails, the build is declared as a Partial Success. This forces the build to Fail on test failures

tests_e2e/scenarios/testsuites/agent_junit.py
```diff
@@ -0,0 +1,50 @@
...
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return AgentJUnitSchema
+
+    def _received_message(self, message: MessageBase) -> None:
+        if isinstance(message, TestResultMessage):
+            distro = message.information.get('distro_version')
+            if distro is not None:
+                message.name = distro
+        super()._received_message(message)
```
narrieta, line 50: We replace the test name (which is always "main" in our test runs) with the distro name

tests_e2e/pipeline/templates/execute-tests.yml
```diff
@@ -26,20 +26,23 @@ jobs:
...
+        continueOnError: true
         env:
           # Add all KeyVault secrets explicitly as they're not added by default to the environment vars
           AZURE_CLIENT_ID: $(AZURE-CLIENT-ID)
           AZURE_CLIENT_SECRET: $(AZURE-CLIENT-SECRET)
           AZURE_TENANT_ID: $(AZURE-TENANT-ID)
           SUBSCRIPTION_ID: $(SUBSCRIPTION-ID)
 
+      - publish: $(Build.ArtifactStagingDirectory)
+        artifact: 'artifacts'
```
maddieford, line 38: How should I be accessing the test artifacts/logs?

I thought you were going to the pipeline's published items, but when I go to check the published artifacts on the test run I don't see the same hierarchy you shared on our call:
image

maddieford (approved):

tests_e2e/pipeline/templates/execute-tests.yml
nagworld9: how this flag(failTaskOnFailedTests) is related and only checks for execute-tests status?

will this not run if I have test failures? I think we need all the time to know summary of failures and successes.

narrieta: The task will mark itself as Failed, but it won't prevent it from publishing the results

from https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/publish-test-results-v2

failTaskOnFailedTests - Fail if there are test failures
boolean. Default value: false.

Optional. When this boolean's value is true, the task will fail if any of the tests in the results file are marked as failed. The default is false, which will simply publish the results from the results file.

narrieta: (see run 5534 for an example)

nagworld9: Thanks for explaining, makes sense. I see few runs with warning sign. what is that mean? we have some warnings in test but no failures?

narrieta: Those are partial success (reported as "[Build partially succeeded] ...- WALinuxAgent - 0eeb2c4e")

The test execution failed (either because we could not execute the tests, or because some test failed), but I forced the Pipeline to continueOnError when the test execution task fails.

narrieta: BTW - In those test runs there are no test results due to a bug in LISA. I reported it this morning and they will fix it.

nagworld9 (dismissed):

nagworld9 (approved):

maddieford (approved):

PR #2751, opened 2023-02-02
Report results for individual tests

narrieta:
Currently we report the junit results at the level of a test suite. The changes in this PR report now at the level of individual tests.

### DIFF
--- a/tests_e2e/orchestrator/lib/agent_test_suite.py
+++ b/tests_e2e/orchestrator/lib/agent_test_suite.py
@@ -335,33 +335,22 @@ class AgentTestSuite(TestSuite):
         """
         suite_name = suite.name
         suite_full_name = f"{suite_name}-{self.context.image_name}"
+        suite_start_time: datetime.datetime = datetime.datetime.now()
 
         with _set_thread_name(suite_full_name):  # The thread name is added to self._log
             with set_current_thread_log(Path.home()/'logs'/f"{suite_full_name}.log"):
-                start_time: datetime.datetime = datetime.datetime.now()
-
-                message: TestResultMessage = TestResultMessage()
-                message.type = "AgentTestResultMessage"
-                message.id_ = str(uuid.uuid4())
-                message.status = TestStatus.RUNNING
-                message.suite_full_name = suite_name
-                message.suite_name = message.suite_full_name
-                message.full_name = f"{suite_name}-{self.context.image_name}"
-                message.name = message.full_name
-                message.elapsed = 0
-                notifier.notify(message)
-
                 try:
                     agent_test_logger.info("")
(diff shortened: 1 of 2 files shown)


tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -392,21 +399,51 @@ def _execute_test_suite(self, suite: TestSuiteDescription) -> bool:
                         agent_test_logger.info("\t%s", r)
                     agent_test_logger.info("")
 
-                    if len(failed) == 0:
```
narrieta, line 395: this is no longer needed. we were reporting a list of test names for failures, but now each failure would be reported individually

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -335,33 +335,22 @@ def _execute_test_suite(self, suite: TestSuiteDescription) -> bool:
         """
         suite_name = suite.name
         suite_full_name = f"{suite_name}-{self.context.image_name}"
+        suite_start_time: datetime.datetime = datetime.datetime.now()
 
         with _set_thread_name(suite_full_name):  # The thread name is added to self._log
             with set_current_thread_log(Path.home()/'logs'/f"{suite_full_name}.log"):
-                start_time: datetime.datetime = datetime.datetime.now()
```
narrieta, line 341: removed the per-test suite result and added per-test results in the 'for test in suite.tests:' loop

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -392,21 +399,51 @@ def _execute_test_suite(self, suite: TestSuiteDescription) -> bool:
                         agent_test_logger.info("\t%s", r)
                     agent_test_logger.info("")
 
-                    if len(failed) == 0:
-                        message.status = TestStatus.PASSED
-                    else:
-                        message.status = TestStatus.FAILED
-                        message.message = f"Tests failed: {failed}"
-
                 except:  # pylint: disable=bare-except
```
maddieford, line 402: Do we also need to add `failed = True` in this except block? Although it might not be necessary because the except blocks above handle this

narrieta: ah, good catch, thanks! yes, it is needed (and important, since an exception at this level of the code may prevent some tests from running and we need to be notified about that). I'm adding it now.

narrieta: added

maddieford (approved):

nagworld9 (approved):

PR #2756, opened 2023-02-09
Add distros and location to the test suites definition files

narrieta:
Now the distros and location where test should run are defined in the configuration files for the test suites, similarly to the approach followed in DCR.

* Changed the format of the definition files from JSON to YAML
* Extended definition files to include distros, location, VM size, and whether the test VMs can be shared with other tests.
* Created a LISA combinator to produce the list of test VMs that need to be created
* Updated the distro list to include the distros in DCR v1, DCR v2 + new distros.


tests_e2e/pipeline/pipeline-cleanup.yml
```diff
@@ -4,7 +4,7 @@
 # Runs every 3 hours and deletes any resource groups that are more than a day old and contain string "lisa-WALinuxAgent-"
 #
 schedules:
-  - cron: "0 */3 * * *" # Run every 3 hours
+  - cron: "0 */12 * * *" # Run twice a day (every 12 hours)
```
narrieta, line 7: I decreased the frequency of cleanup. Twice a day should be enough, we can adjust later if needed.

tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
```diff
@@ -0,0 +1,119 @@
...
+
+                # If the image specifies a list of VM sizes, use any of them. Otherwise, set the size to empty and let LISA choose it.
+                vm_size = image.vm_sizes[0] if len(image.vm_sizes) > 0 else ""
+
+                if suite_info.owns_vm:
+                    environment_list.append({
+                        "marketplace_image": image.urn,
+                        "location": location,
+                        "vm_size": vm_size,
+                        "vhd": "",
```
narrieta, line 93: support for VHDs will come in a separate PR

tests_e2e/orchestrator/lib/agent_test_loader.py
```diff
@@ -96,12 +197,53 @@ def _load_tests(source_file: Path) -> List[Type[AgentTest]]:
...
+                    raise Exception(f"Image {name} is missing the 'urn' property: {description}")
+                i.urn = description["urn"]
+                i.locations = description["locations"] if "locations" in description else []
+                i.vm_sizes = description["vm_sizes"] if "vm_sizes" in description else []
+            images[name] = [i]
+
+        # now load the image-sets, mapping them to the images that we just computed
+        for image_set_name, image_list in image_descriptions["image-sets"].items():
+            # the same name cannot denote an image and an image-set
+            if image_set_name in images:
```
maddieford, line 231: We currently only have one image_set which is 'endorsed', is that correct?

maddieford (approved):

tests_e2e/orchestrator/lib/agent_test_loader.py
narrieta: for now, yes. the bvts use 'endorsed'. dcr has many more images and image sets, but i don't want to migrate them in bulk. we'll migrate them as we migrate the tests that actually used them.

tests_e2e/test_suites/images.yml
```diff
@@ -0,0 +1,70 @@
...
+      - "ubuntu_1604"
+      - "ubuntu_1804"
+      - "ubuntu_2004"
+
+#
+# An image can be specified by a string giving its urn, as in
+#
+#    ubuntu_2004: "Canonical 0001-com-ubuntu-server-focal 20_04-lts latest"
+#
+# or by an object with 3 properties: urn, locations and vm_sizes, as in
```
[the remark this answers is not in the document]
narrieta: Mariner 2 ARM64 provides both, but they are independent. Most images will omit both, a few images will set a location, and even fewer will set both a location and vm_size (or just a vm_size).

DCR doesn't allow to specify the VM size for an image. This is a new requirement needed to support Mariner 2 ARM64, currently that image has very limited availability.

tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
```diff
@@ -0,0 +1,119 @@
...
+
+@dataclass_json()
+@dataclass
+class AgentTestSuitesCombinatorSchema(schema.Combinator):
+    test_suites: str = field(
+        default_factory=str, metadata=field_metadata(required=True)
+    )
+
+
+class AgentTestSuitesCombinator(Combinator):
```
nagworld9, line 29: when is this invoked?

tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
```diff
@@ -0,0 +1,119 @@
...
+
+        log: logging.Logger = logging.getLogger("lisa")
+        log.info("******** Environments *****")
+        for e in environment_list:
+            log.info(
+                "{ marketplace_image: '%s', location: '%s', vm_size: '%s', vhd: '%s', test_suites_info: '%s' }",
+                e['marketplace_image'], e['location'], e['vm_size'], e['vhd'], [s.name for s in e['test_suites_info']])
+        log.info("***************************")
+
+        return environment_list
```
nagworld9, line 119: who gets this return list?

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -308,9 +307,9 @@ def execute(self, node: Node, variables: Dict[str, Any], log: Logger) -> None:
                     if not self.context.skip_setup:
                         self._setup_node()
 
-                    test_suites: List[TestSuiteDescription] = AgentTestLoader(self.context.test_source_directory).load(self.context.test_suites)
-
-                    for suite in test_suites:
+                    # pylint seems to think self.context.test_suites is not iterable. Suppressing warning, since its type is List[AgentTestSuite]
+                    #  E1133: Non-iterable value self.context.test_suites is used in an iterating context (not-an-iterable)
+                    for suite in self.context.test_suites:  # pylint: disable=E1133
                         test_suite_success = self._execute_test_suite(suite) and test_suite_success
```
nagworld9, line 313: how this suite type as TestSuiteInfo. I see `self.context.test_suites` get assigned from the variable we pass to Lisa run. That would be list of commas separated or directory type of single test.

tests_e2e/test_suites/images.yml
```diff
@@ -0,0 +1,70 @@
...
+#   centos_610: "OpenLogic CentOS 6.10 latest"
+   centos_79: "OpenLogic CentOS 7_9 latest"
+#   debian_8: "credativ Debian 8 latest"
+   debian_9: "credativ Debian 9 latest"
+   debian_10: "Debian debian-10 10 latest"
+   mariner_1: "microsoftcblmariner cbl-mariner cbl-mariner-1 latest"
+   mariner_2: "microsoftcblmariner cbl-mariner cbl-mariner-2 latest"
+   mariner_2_arm64:
+      urn: "microsoftcblmariner cbl-mariner cbl-mariner-2-arm64 latest"
+      locations:
```
nagworld9, line 60: why do we need list here? I don't see a use case? and also, your implementation is only reading first value? same goes for vm_sizes?

tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
nagworld9: based on this image list, lisa creates that many vms?

tests_e2e/orchestrator/lib/agent_test_suite.py
narrieta: The string we pass to the runbook is actually an argument for the combinator:
```
  #
  # These variables parameters for the AgentTestSuitesCombinator combinator
  #
  # The test suites to execute
  - name: test_suites
    value: "agent_bvt"
    is_case_visible: true

```
Then the combinator produces an array of TestSuiteInfos that is passed to the AgentTestSuite in the "test_suites_info" variable:

```
  #
  # These variables are set by the AgentTestSuitesCombinator combinator
  #
...
  - name: test_suites_info
    value: []
    is_case_visible: true

```

My next PR already includes extra comments and renames the variables produced by the combinator so that they are prefixed by "c_", which should help.

tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
narrieta: It is invoked by LISA.

LISA will invoke the combinator, whose output feeds the the values needed to create the test VM

```
platform:
  - type: azure
     ....
      azure:
        marketplace: $(marketplace_image)
        vhd: $(vhd)
        location: $(location)
        vm_size: $(vm_size)
```

tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
narrieta: The runbook uses the values returned by the combinator to feed the values used to create the VMs (see my previous comment above)

tests_e2e/test_suites/images.yml
narrieta: An image may be available only in a few regions. 

Say an image I is available only on regions R1 and R2. Say we have test suites T1, which needs to run on R1, T2, which needs to run on R2, and T3, which can run on any region.

In that case, we would use R1 for T1, R2 for T2, and for T3 we don't care about the region, so we just use whatever value is item 0 of the list.

nagworld9: That is not implemented right where you are not taking region based on test_suite. The example you provided could be achieved by setting region on test_suite.yml itself. I still don't see multiple in images list.

tests_e2e/orchestrator/lib/agent_test_loader.py
```diff
@@ -15,77 +15,178 @@
...
     """
-    Description of the test suite loaded from its JSON file.
+    Description of a test suite
     """
+    # The name of the test suite
     name: str
+    # The tests that comprise the suite
     tests: List[Type[AgentTest]]
+    # An image or image set (as defined in images.yml) specifying the images the suite must run on.
+    images: str
```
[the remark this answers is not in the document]
narrieta: the solution is the same as in dcr: create a new image set that includes all the images that the suite needs to run on and use that set

tests_e2e/test_suites/images.yml
narrieta: it's done by the combinator:
```

            for image in images_info:
                # If the suite specifies a location, use it. Else, if the image specifies a list of locations, use
                # any of them. Otherwise, use the default location.
                if suite_info.location != '':
                    location = suite_info.location
                elif len(image.locations) > 0:
                    location = image.locations[0]
                else:
                    location = AgentTestSuitesCombinator._DEFAULT_LOCATION
```

narrieta: > The example you provided could be achieved by setting region on test_suite.yml itself.

No, setting the region on the test suite has a different meaning. That means the suite must run on that location. When set on the image it means the image is available only on a subset of regions. They are different (and could have different values)

tests_e2e/orchestrator/lib/agent_test_loader.py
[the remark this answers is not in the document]
nagworld9: In DCR, we create the new image set in specific test_suite, now we need to define all those in image.yml. If we end of doing it in one file, it becomes too much in single file. May be worth considering allowing new image set at test_suite level. For now, ok and I guess when we port more tests, we get the idea whether we need to implement this or not.

tests_e2e/test_suites/images.yml
nagworld9: My concern is this, every suite gets same location. What is the point of having multiple.

```
                elif len(image.locations) > 0:
                    location = image.locations[0]
```

tests_e2e/orchestrator/lib/agent_test_loader.py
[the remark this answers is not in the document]
narrieta: Yes, start simple and add functionality only if actually needed. Using an array instead of a single item is pretty simple, actually, but for the moment keep things simple.

tests_e2e/test_suites/images.yml
narrieta: why is that a concern? for sure we wouldn't be overloading that region with out tests :)

nagworld9: I mean you said we will have leverage to pick the location based on test_suite but implementation is not reflecting that.

nagworld9: Anyway, I'm ok for now. we will see when we have that use case.

narrieta: it's done by the combinator. see the code snippet i posted above and run it over the example i mentioned 

> Say an image I is available only on regions R1 and R2. Say we have test suites T1, which needs to run on R1, T2, which needs to run on R2, and T3, which can run on any region.

>In that case, we would use R1 for T1, R2 for T2, and for T3 we don't care about the region, so we just use whatever value is item 0 of the list.

nagworld9: According to the schema I could have R3, R4, R5 in the image location. Lets say I didn't define any region on T3. So T3 would pick R3 and say new T4 also don't have region defined. It will pick R3 only, where is the case any test case would pick R4 and R5.

narrieta: None. 

That the image is available on R3, R4, and R5 doesn't mean we need to use all of them. If T4 doesn't specify a location then using item 0 (R3) is as good as using item 1 (R4), or item 2 (R5).

If later on, another suite T9 needs to run on R5, then R5 will be used.

nagworld9: Maybe I didn't understand completely. I still don't see list is useful here. I'm approving it and we can discuss when you provide another session on this pipeline.

nagworld9 (approved):

tests_e2e/test_suites/images.yml
narrieta: The list is needed if there are test suites that need to run on different locations.

If T1 needs to run on R1 and T2 needs to run on R2, and the image is available only on R1 and R2, if we don't use a list then you need to chose between R1 and R2, and then you can run only 1 test: either T1 or T2.

Yes, we can go over this in the presentation.

PR #2761, opened 2023-02-14
Download certificates when goal state source is Fast Track

### LINKED ISSUE #2750: WALinuxAgent doesn't download all .crt and .prv files from KeyVault
Not sure if this is a bug but I will try to explain as much as I can.

Environment: Virtual Machine Scale Set - Azure (West EU, West US, Japan, Asia,.. all regions impacted)

We are doing deployments of VMSS in azure using ARM template and ansible playbook with some configurations.
Before running the ansible, we are using the following to push certficates from KV:

TEMPLATE
{
    "type": "Microsoft.Compute/virtualMachines",
    "name": "Region1VM",
    ...
    "properties": {
    ...
    "osProfile": {
            "computerName": "Region1VM",
    ...
    "secrets": [
                {
                    "sourceVault": {
                        "id": "[resourceId('Microsoft.KeyVault/vaults', Region1KeyVault)]"
                    },
                    "vaultCertificates": [
                        {
                            "certificateUrl": "[reference(resourceId('Microsoft.KeyVault/vaults/secrets', 'Region1KeyVault', 'SampleCertificateAsSecret')).secretUriWithVersion]",
                            "certificateStore": "My"
                        }
                    ]
                }
            ],
        },
    ...
    }
}

CODE
image

LINK
https://devblogs.microsoft.com/premier-developer/centralized-vm-certificate-deployment-across-multiple-regions-with-arm-templates/#part-2-push-certificate-from-the-regional-key-vault-to-the-virtual-machine

after running this part of the ARM template we do some certificate copies from var/lib/waagent to another location but it fails with the below error:

"could not find or access '/var/lib/waagent/nameexample.prv"

image

The problem is, the file that is missing should be downloaded during the push of the certificates from Keyvault but this is not happening and ansible playbook crashes.

If we restart the waagent service, the file "nameexample.prv" will be downloaded and ansible will not crash anymore.

The final lines of the ansible code, will remove this file again from the VM. The next deployment will crash again. 

We have two workarounds here:
FIRST- if we restart the waagent after the crash everything will run as expected
SECOND - if we don't delete the file after the ansible

MAIN PROBLEM - this has been working like this for last 8 months, but now we are getting this errors. 

we don't understand why, the agent doesn't push all the file in KV and we always need to restart the service to do a "complete download" let's say.

 - Distro and Version: RedHat|RHEL|7.3|
 - WALinuxAgent version: 2.9.0.4

**Additional context**
I can give you an agent log from 31st of January 4:30PM issue : 

image

**Log file attached**

[waagent.log](https://github.com/Azure/WALinuxAgent/files/10570703/waagent.log)

maddieford:
This PR includes a change to download certificates when goal state source is FastTrack

Fixes #2750


tests/protocol/test_goal_state.py
```diff
@@ -412,7 +412,7 @@ def http_get_handler(url, *_, **__):
             goal_state = GoalState(protocol.client)
 
             self.assertEqual(2, protocol.mock_wire_data.call_counts['goalstate'], "There should have been exactly 2 requests for the goal state (original + refresh)")
-            self.assertEqual(2, http_get_handler.certificate_requests, "There should have been exactly 2 requests for the goal state certificates (original + refresh)")
+            self.assertEqual(4, http_get_handler.certificate_requests, "There should have been exactly 4 requests for the goal state certificates (2x original + 2x refresh)")
```
maddieford, line 415: The previous 2 requests became 4 requests for this case because we're downloading certificates in _fetch_full_wire_server_goal_state and _update directly. 

Should I add logic to only download certificates once per goals state?

tests/protocol/test_goal_state.py
```diff
@@ -156,7 +156,7 @@ def http_get_handler(url, *_, **__):
             protocol.set_http_handlers(http_get_handler=None)
             goal_state.update()
             self._assert_directory_contents(
-                self._find_history_subdirectory("234-987"), ["VmSettings.json"])
+                self._find_history_subdirectory("234-987"), ["VmSettings.json", "Certificates.json"])
```
maddieford, line 159: This test shows that we create history subdirectory with Certificates.json when we have a fast track goal state. Should I add an additional test to verify the certificates are downloaded?

tests/protocol/test_wire.py
```diff
@@ -1101,7 +1101,7 @@ def test_forced_update_should_update_the_goal_state_and_the_host_plugin_when_the
 
     def test_reset_should_init_provided_goal_state_properties(self):
         with mock_wire_protocol(mockwiredata.DATA_FILE) as protocol:
-            protocol.client.reset_goal_state(goal_state_properties=GoalStateProperties.All & ~GoalStateProperties.ExtensionsGoalState)
+            protocol.client.reset_goal_state(goal_state_properties=GoalStateProperties.All & ~GoalStateProperties.Certificates)
```
maddieford, line 1104: I updated this test after separating ExtensionsGoalState and Certificates in GoalStateProperties.

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -435,18 +453,8 @@ def _fetch_full_wire_server_goal_state(self, incarnation, xml_doc):
...
-                # Log and save the certificates summary (i.e. the thumbprint but not the certificate itself) to the goal state history
-                for c in certs.summary:
-                    message = "Downloaded certificate {0}".format(c)
-                    self.logger.info(message)
-                    add_event(op=WALAEventOperation.GoalState, message=message)
-                if len(certs.warnings) > 0:
-                    self.logger.warn(certs.warnings)
-                    add_event(op=WALAEventOperation.GoalState, message=certs.warnings)
-                self._history.save_certificates(json.dumps(certs.summary))
+            if (GoalStateProperties.Certificates & self._goal_state_properties) and certs_uri is not None:
```
narrieta, line 456: I did not catch this before, but we should do this unconditionally, or update the method's name and documentation

```

    def _fetch_full_wire_server_goal_state(self, incarnation, xml_doc):
        """
        Issues HTTP requests (to the WireServer) for each of the URIs in the goal state (ExtensionsConfig, Certificate, Remote Access users, etc)
        and populates the corresponding properties.
```

(of course, we still need to check that "certs_uri is not None")

narrieta: never mind, this is ok

narrieta (approved):

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -289,6 +290,9 @@ def _update(self, force_update):
         # case, to ensure it fetches the new certificate.
         #
         if self._extensions_goal_state.source == GoalStateSource.FastTrack:
+            certs_uri = findtext(xml_doc, "Certificates")
+            if certs_uri is not None:
+                self._download_certificates(certs_uri)
```
nagworld9, line 295: In case if we downloaded new certs then we are not updating the state of goal_state certs in self._certs.  Next line check certificates is using self._certs which will have old certs. Am I missing something?

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -289,9 +290,14 @@ def _update(self, force_update):
         # case, to ensure it fetches the new certificate.
         #
         if self._extensions_goal_state.source == GoalStateSource.FastTrack:
-            self._check_certificates()
+            certs_uri = findtext(xml_doc, "Certificates")
```
narrieta, line 293: Minor: let's add certs_uri as a data member of self and use that instead of picking it up  from the xml_doc we just fetched. With the current code, this makes no difference functionally, but let's do that in case the code changes. That would also emphasize that we are redownloading the same certs as we did before. 

Also, we should re-download only if self._goal_state_properties & GoalStateProperties.Certificates != 0

narrieta (approved):

nagworld9 (approved):

PR #2767, opened 2023-02-21
Use unique name for test VMs. Remove hardcoded log path

narrieta:
* We have been using the distro name to distinguish test VMs in the main log, the junit results, and to create the file name for test logs. With the support for location, vm_size and "owns VM" in my previous PR, the distro name is no longer unique. This PR fixes this, adding the location, or the test suite name for suites that own the VM, to the distro name.

* The AgentTestSuite was hardcoding the location of the log files to $HOME/logs. Now this location can be passed as a parameter to the runbook.


tests_e2e/tests/lib/vm_extension.py
```diff
@@ -99,9 +99,9 @@ def enable(
                 extension_parameters
             ).result(timeout=_TIMEOUT))
 
-        if result.provisioning_state != 'Succeeded':
```
narrieta, line 102: This check is failing from time to time because we get 'Updating'; that should not be reported as a failure.

nagworld9 (approved):

PR #2771, opened 2023-02-27
Add support for Azure clouds

narrieta:
This PR adds support for executing the Azure pipeline on different clouds.

It adds a new variable to the pipeline, 'connection_info' which points to the keyvault where the connection parameters are stored. 

It also separates the identity used to connect to the container registry from the identity used for management operations (create VMs, execute extensions, etc).

Lastly, it adds a new parameter, 'cloud' to the LISA runbook.

### DIFF
--- a/tests_e2e/pipeline/pipeline.yml
+++ b/tests_e2e/pipeline/pipeline.yml
@@ -1,3 +1,9 @@
+# variables:
+  #
+  # NOTE: When creating the pipeline, "connection_info" must be added as a variable pointing to the
+  #       corresponding key vault; see wiki for details.
+  #
+
 parameters:
     # See the test wiki for a description of the parameters
   - name: test_suites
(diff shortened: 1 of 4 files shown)


tests_e2e/pipeline/pipeline.yml
```diff
@@ -44,23 +50,6 @@ parameters:
     - failed
     - no
 
-variables:
```
narrieta, line 47: These variables are used only by the "Execute tests" tasks, so I removed them and referenced the parameters directly in that task.

tests_e2e/pipeline/pipeline.yml
```diff
@@ -73,36 +62,48 @@ jobs:
...
       - task: AzureKeyVault@2
-        displayName: "Fetch secrets from KV"
+        displayName: "Fetch connection info"
         inputs:
-          azureSubscription: '$(azureConnection)'
-          KeyVaultName: 'dcrV2SPs'
+          azureSubscription: 'azuremanagement'
+          KeyVaultName: '$(connection_info)'
           SecretsFilter: '*'
-          RunAsPreJob: true
```
narrieta, line 88: 'RunAsPreJob' makes the secrets available to all the tasks, this is not needed since the only task that uses them is "Execute tests". I removed this parameter to use the default value (false), which makes the secrets available only to the next task and I moved the task to precede "Execute tests".

maddieford (approved):

nagworld9 (approved):
I tried checking connection_info variable in the pipeline, I didn't find it and we can discuss offline.

PR #2779, opened 2023-03-07
Add Flatcar to end-to-end tests, install Pypy on test VMs, etc

narrieta:
This PR includes several changes related to adding Flatcar to the test automation:

* Added Flatcar (x64 and ARM64) to the endorsed distros
* Installed Pypy 3.7 to the test VMs and made it the default Python for the user executing the tests (now test code does not need to worry about the Python version on the test VM).
* Added support to skip tests
* Added support to copy tools used by the tests to the test VM
* Fixed a couple of bugs in AgentTestSuiteCombinator


tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -44,10 +43,13 @@
 import makepkg
 from azurelinuxagent.common.version import AGENT_VERSION
 from tests_e2e.orchestrator.lib.agent_test_loader import TestSuiteInfo
+from tests_e2e.tests.lib.agent_test import TestSkipped
```
narrieta, line 46: Tests can raise TestSkipped to skip a non-supported distro, etc

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -148,17 +151,12 @@ def _set_context(self, node: Node, variables: Dict[str, Any], lisa_log_path: str
         self.__context.log_path = self._get_log_path(variables, lisa_log_path)
         self.__context.log = log
         self.__context.node = node
-        self.__context.is_vhd = self._get_required_parameter(variables, "c_vhd") != ""
-        self.__context.image_name = f"{node.os.name}-vhd" if self.__context.is_vhd else self._get_required_parameter(variables, "c_name")
+        self.__context.is_vhd = self._get_optional_parameter(variables, "c_vhd") != ""
+        self.__context.image_name = f"{node.os.name}-vhd" if self.__context.is_vhd else self._get_required_parameter(variables, "c_env_name")
```
narrieta, line 155: I added a "vm_name" parameter for existing VMs, so I renamed "c_name" to "c_env_name"

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -260,29 +265,51 @@ def _setup_node(self) -> None:
...
         else:
             self._install_agent_on_node()
 
+    def _install_tools_on_node(self) -> None:
+        """
+        Installs the test tools on the test node
+        """
+        self.context.ssh_client.run_command("mkdir -p ~/bin")
+
+        tools_path = self.context.test_source_directory/"orchestrator"/"scripts"
```
narrieta, line 281: now we copy orchestrator/scripts to the VMs in order to execute those scripts over SSH

nag, if you need scripts for your tests, probably we should create a similar subdirectory under the tests directory and copy it here

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -293,13 +320,14 @@ def _collect_node_logs(self) -> None:
...
-            self.execute_script_on_node(self.context.test_source_directory/"orchestrator"/"scripts"/"collect-logs", sudo=True)
+            stdout = self.context.ssh_client.run_command("collect-logs", use_sudo=True)
+            self._log.info(stdout)
 
             # Copy the tarball to the local logs directory
             remote_path = "/tmp/waagent-logs.tgz"
             local_path = self.context.log_path/'{0}.tgz'.format(self.context.image_name)
             self._log.info("Copying %s:%s to %s", self.context.node.name, remote_path, local_path)
-            self.context.node.shell.copy_back(remote_path, local_path)
+            self.context.ssh_client.copy(remote_path, local_path, remote_source=True)
```
narrieta, line 330: i removed the references to LISA's SSH client and now I am using the Agent's

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -310,9 +338,14 @@ def agent_test_suite(self, node: Node, variables: Dict[str, Any], log_path: str,
         """
         self._set_context(node, variables, log_path, log)
 
-        test_suite_success = True
-
         with _set_thread_name(self.context.image_name):  # The thread name is added to self._log
+            self._log.info(
```
narrieta, line 342: I move this message here from the initialization function. We set the thread name for the log after the initialization, so it was missing from the log for this message

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -332,12 +365,17 @@ def agent_test_suite(self, node: Node, variables: Dict[str, Any], log_path: str,
...
             except:   # pylint: disable=bare-except
-                # Note that we report the error to the LISA log and then re-raise it. We log it here
-                # so that the message is decorated with the thread name in the LISA log; we re-raise
-                # to let LISA know the test errored out (LISA will report that error one more time
-                # in its log)
-                self._log.exception("UNHANDLED EXCEPTION")
-                raise
+                # Report the error and raise and exception to let LISA know that the test errored out.
+                self._log.exception("TEST FAILURE DUE TO AN UNEXPECTED ERROR.")
+                self._report_test_result(
```
narrieta, line 370: I wasn't reporting those failures to the Junit file

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -332,12 +365,17 @@ def agent_test_suite(self, node: Node, variables: Dict[str, Any], log_path: str,
...
+                    self.context.image_name,
+                    "Setup",
+                    TestStatus.FAILED,
+                    start_time,
+                    message="TEST FAILURE DUE TO AN UNEXPECTED ERROR.",
+                    add_exception_stack_trace=True)
+
+                raise Exception("STOPPING TEST EXECUTION DUE TO AN UNEXPECTED ERROR.")
```
narrieta, line 378: Replaced the re-raise with this exception to eliminate the duplicated error message in the LISA log, and to make clearer that we are stopping the LISA run

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -460,32 +508,4 @@ def _report_test_result(
 
         notifier.notify(msg)
 
-    def execute_script_on_node(self, script_path: Path, parameters: str = "", sudo: bool = False) -> int:
```
narrieta, line 463: Now we use the Agent's SSH client to do this

tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
```diff
@@ -29,41 +29,59 @@ class AgentTestSuitesCombinatorSchema(schema.Combinator):
...
         default_factory=str, metadata=field_metadata(required=True)
     )
+    image: str = field(
+        default_factory=str, metadata=field_metadata(required=False)
+    )
     vm_size: str = field(
-        default_factory=str, metadata=field_metadata(required=True)
+        default_factory=str, metadata=field_metadata(required=False)
+    )
+    vm_name: str = field(
```
narrieta, line 41: new parameter for existing VMs

tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
```diff
@@ -29,41 +29,59 @@ class AgentTestSuitesCombinatorSchema(schema.Combinator):
...
     """
     def __init__(self, runbook: AgentTestSuitesCombinatorSchema) -> None:
         super().__init__(runbook)
-        self._environments = self.create_environment_list()
-        self._index = 0
-
         if self.runbook.cloud not in self._DEFAULT_LOCATIONS:
             raise Exception(f"Invalid cloud: {self.runbook.cloud}")
 
+        if self.runbook.vm_name != '' and (self.runbook.image != '' or self.runbook.vm_size != ''):
```
narrieta, line 75: The combinator was generating the full list of distros when running on an existing VM. For that scenario, it should generate a single item containing that VM

tests_e2e/orchestrator/sample_runbooks/existing_vm.yml
```diff
@@ -31,6 +31,8 @@ variable:
   #
   # These variables identify the existing VM, and the user for SSH connections
   #
+  - name: cloud
```
narrieta, line 34: forgot to add "cloud" in my previous PR

tests_e2e/orchestrator/scripts/get-agent-pythonpath
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
```
narrieta, line 1: this script can be used to set the PYTHONPATH for any Python (this PR uses it for Pypy) before invoking 'waagent' or importing any of its modules

tests_e2e/orchestrator/scripts/get-waagent-path
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
```
narrieta, line 1: in some distros (e.g. Flatcar)  'waagent' is not in PATH

tests_e2e/orchestrator/scripts/unzip.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python3
```
narrieta, line 1: we were using the "unzip" package to do this, but that package may not be available in all tested distros (e.g. Flatcar does not even have a package manager) so now we use this Python script to unzip.

tests_e2e/test_suites/images.yml
```diff
@@ -57,18 +59,19 @@ image-sets:
 # '<Publisher>:<Offer>:<Sku>:<Version>'
 #
 images:
-#
```
narrieta, line 60: removed this duplicated comment... these distros are not added to 'endorsed' yet, and there is already a TODO comment there

tests_e2e/tests/bvts/vm_access.py
```diff
@@ -38,6 +38,9 @@
 
 class VmAccessBvt(AgentTest):
     def run(self):
+        if type(self._context.node.os).__name__ == 'CoreOs' and self._context.node.os.information.full_version.startswith('Flatcar'):
```
maddieford, line 41: Should we keep a list of skipped tests by distro? Or add a TODO here so we know to remove this when VMAccess becomes supported on Flatcar.

maddieford (dismissed):

tests_e2e/tests/bvts/vm_access.py
narrieta: good point, we already have 3 or so tests/test cases that we skip. i'll create work items for those

tests_e2e/orchestrator/lib/agent_test_suite.py
nagworld9: copy to where?` ~bin` path

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -260,29 +265,51 @@ def _setup_node(self) -> None:
...
-
         self._log.info("Installing %s on %s", agent_package_path, self.context.node.name)
         agent_package_remote_path = self.context.remote_working_directory/agent_package_path.name
         self._log.info("Copying %s to %s:%s", agent_package_path, self.context.node.name, agent_package_remote_path)
-        self.context.node.shell.copy(agent_package_path, agent_package_remote_path)
-        self.execute_script_on_node(
-            self.context.test_source_directory/"orchestrator"/"scripts"/"install-agent",
-            parameters=f"--package {agent_package_remote_path} --version {AGENT_VERSION}",
-            sudo=True)
+        self.context.ssh_client.copy(agent_package_path, agent_package_remote_path, remote_target=True)
```
nagworld9, line 310: is this for zip folder?

nagworld9 (commented):
Now ssh_client is added to context, can we extend that to tests instead initializing on every test.

tests_e2e/orchestrator/lib/agent_test_suite.py
narrieta: Yes, copy to ~/bin.

 '~' means $HOME for the current user (waagent on the automation pipeline, or whatever user is passed as parameter to the runbook for dev runs).

tests_e2e/orchestrator/scripts/install-agent
```diff
@@ -76,9 +75,14 @@ echo "Service name: $service_name"
...
 #
-echo "Installing $package..."
-unzip -d "/var/lib/waagent/WALinuxAgent-$version" -o "$package"
-sed -i 's/AutoUpdate.Enabled=n/AutoUpdate.Enabled=y/g' /etc/waagent.conf
+echo "Installing $package as version $version..."
+unzip.py "$package" "/var/lib/waagent/WALinuxAgent-$version"
+
+# Ensure that AutoUpdate is enabled. some distros, e.g. Flatcar, don't have a waagent.conf
+# but AutoUpdate defaults to True so there is no need to anything in that case.
+if [[ -e /etc/waagent.conf ]]; then
```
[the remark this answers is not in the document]
narrieta: The default values are hardcoded in the agent.Specific tests may need to create /etc/waagent.conf, but the overall infrastructure should not create it, since that is not the way actual VMs would look like.

BTW, currently VMAccess breaks on Flatcar because it has a hard dependency on /etc/waagent.conf; the extension needs to fix this dependency, since user-created VMs won't have it. If the AgentTestSuite created the conf file, we would be masking this bug.

tests_e2e/orchestrator/scripts/install-agent
```diff
@@ -92,14 +96,22 @@ mv /var/log/waagent.log /var/log/waagent."$(date --iso-8601=seconds)".log
...
-# may take some time to start so give 1 minute.
+# Verify that the new agent is running and output its status.
+# Note that the extension handler may take some time to start so give 1 minute.
+# Also, note that the default Python is set to Pypy, so before executing 'waagent' we need to set the
+# Python path to the location of the azurelinuxagent module.
 #
 echo "Verifying agent installation..."
+
+PYTHONPATH=$(get-agent-pythonpath)
+export PYTHONPATH
```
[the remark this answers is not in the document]
narrieta: If tests need to import agent modules, or invoke 'waagent', they need to do the same as this script

```
PYTHONPATH=$(get-agent-pythonpath)
export PYTHONPATH

```

Let me think a little if we should also set PYTHONPATH in .bash_profile, and then tests would not need to worry about it. I can do that on a separate PR.

tests_e2e/orchestrator/scripts/install-agent
```diff
@@ -92,14 +96,22 @@ mv /var/log/waagent.log /var/log/waagent."$(date --iso-8601=seconds)".log
...
+
 check-version() {
+  waagent=$(get-waagent-path)
```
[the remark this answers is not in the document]
narrieta: flatcar forced me to do it :)

tests_e2e/orchestrator/scripts/install-agent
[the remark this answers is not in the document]
nagworld9: But why flatcar vm don't have it? what is special about it? My suspicion is that our agent source setup itself don't have it. In general distro owners copy or look at our agent setup to bootstrap agent right.

tests_e2e/orchestrator/scripts/get-agent-pythonpath
```diff
@@ -0,0 +1,74 @@
...
+
+  return 1
+}
+
+if ! python=$(find-agent-python); then
+  echo "Can't find the python command used to start the agent" >&2
+  exit 1
+fi
+
+$python -c 'import site; print(":".join(site.getsitepackages()))'
```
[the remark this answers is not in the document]
nagworld9: is this agent module python path or new python version path that we installed?

nagworld9:
Now ssh_client is added to context, can we extend that to tests as well instead initializing on every test.

tests_e2e/orchestrator/scripts/get-agent-pythonpath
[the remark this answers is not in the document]
narrieta: that sample is for my dev machine

tests_e2e/orchestrator/scripts/install-agent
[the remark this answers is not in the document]
narrieta: That is the way they setup Flatcar and is perfectly valid. Flatcar is supposed to be immutable so I imagine they are not allowing for any configuration changes by default,

tests_e2e/orchestrator/scripts/install-agent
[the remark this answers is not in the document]
narrieta: I think it should be fine to set PYTHONPATH in .bas_profile. I'll add that to my next PR

nagworld9 (approved):

PR #2786, opened 2023-03-14
Support int type for eventPid and eventTid fields

nagworld9:
## Description
Today we only support string value for eventPid and EventTid. As these fields are ids and ids are generally numeric in nature, so changing logic to support int values too. 

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/ga/collect_telemetry_events.py
+++ b/azurelinuxagent/ga/collect_telemetry_events.py
@@ -58,6 +58,8 @@ class ExtensionEventSchema(object):
            "EventTid":"2",
            "OperationId":"Guid (str)"
         }
+
+    From next version(2.10+) we accept integer values for EventPid and EventTid fields. But we still support string type for backward compatability
     """
     Version = "Version"
     Timestamp = "Timestamp"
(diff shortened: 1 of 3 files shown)


azurelinuxagent/ga/collect_telemetry_events.py
```diff
@@ -58,6 +58,8 @@ class ExtensionEventSchema(object):
            "EventTid":"2",
            "OperationId":"Guid (str)"
         }
+
+    From next version(2.10+) we accept integer values for EventPid and EventTid fields. But we still support string type for backward compatability
```
nagworld9, line 62: Assumed this go in next release and version would be 2.10

azurelinuxagent/ga/collect_telemetry_events.py
```diff
@@ -323,15 +325,20 @@ def _parse_event_and_ensure_it_is_valid(self, extension_event):
         :param extension_event: The json event from file
         :return: Verified Json event that qualifies the contract.
         """
-
-        clean_string = lambda x: x.strip() if x is not None else x
+        def _clean_value(k, v):
+            if v is not None:
```
narrieta, line 329: if v is None, _clean_value will return None and the conversion below will raise an exception

nagworld9: Sorry, where? I kept it same as we have today so that downstream code will work like before.

narrieta: never mind, you are converting the key, not the value

narrieta (approved):

maddieford (approved):

PR #2793, opened 2023-03-30
Python configuration for tests

narrieta:
This PR changes the Python configuration for the end to end tests to an explicit approach: instead of having a default Python, tests need to specify the Python to use (the system's Python or Pypy).

* Always install Pypy on test VMs
* Install the Test agent as a module for Pypy
* Add utilities to find the location of the Agent's executable (waagent) and modules (azurelinuxagent.*), as well as the location of the Python used by the system to execute the Agent.
* Install tests libraries to ~/lib, instead of ~/bin

The PR also includes some minor cleanup fixes. I'll point them out in review comments.


tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -277,54 +277,74 @@ def _setup_node(self) -> None:
         log.info("Resource Group: %s", self.context.vm.resource_group)
         log.info("")
 
-        self.context.ssh_client.run_command("mkdir -p ~/bin/tests_e2e/tests; touch ~/bin/agent-env")
```
narrieta, line 280: Now we prepare a tarball on the orchestrator machine that is already structured in the way it needs to be setup on the test VMs, instead of using SSH to create/copy these files. This should make easier to include additional files that need to be copied to the test VMs.

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -447,15 +469,15 @@ def _execute_test_suite(self, suite: TestSuiteInfo) -> bool:
                         test_start_time: datetime.datetime = datetime.datetime.now()
 
                         log.info("******** Executing %s", test_name)
-                        self.context.lisa_log.info("******** Executing %s", test_full_name)
+                        self.context.lisa_log.info("Executing test %s", test_full_name)
```
narrieta, line 472: Minor formatting change. We used to output "********" to the LISA log to make those messages stand out. My previous PR created a log for each environment, so now the LISA log is not too crowded and this marker is not needed anymore.

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -528,7 +550,7 @@ def _check_agent_log(self) -> bool:
             self.context.lisa_log.info("Checking agent log on the test node")
             log.info("Checking agent log on the test node")
 
-            output = self.context.ssh_client.run_command("check-agent-log.py -j")
+            output = self.context.ssh_client.run_command("$(get-agent-python) ~/bin/check-agent-log.py -j")
```
narrieta, line 553: get_distro() in version.py is not returning the expected values in some distros when using Pypy. I need to follow up that, and in the meanwhile I am using the system's python to execute this script.

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -551,21 +573,17 @@ def _check_agent_log(self) -> bool:
                 log.info("There are no errors in the agent log")
                 return True
 
-            log_path: Path = self.context.log_path/f"CheckAgentLog-{self.context.environment_name}.log"
```
narrieta, line 554: My previous PR create a log for each environment, so now we do not need to create this extra CheckAgentLog-*.log, we can simply log those messages to the environment log.

tests_e2e/orchestrator/scripts/get-agent-modules-path
```diff
@@ -18,34 +18,20 @@
...
 #
-# Returns the path to the Python executable.
+# Returns the PYTHONPATH on which the azurelinuxagent and associated modules are located.
+#
+# To do this, the script tries to find the python command used to start the agent and then
+# returns the value of site.getsitepackages().
 #
 set -euo pipefail
 
-# python3 is available on most distros
```
narrieta, line 25: Somehow git captures this as a rename + changes. I actually removed find-python and created a totally unrelated script, get-agent-modules-path

tests_e2e/orchestrator/scripts/uncompress.py
```diff
@@ -1,5 +1,3 @@
-#!/usr/bin/env python3
```
narrieta, line 1: This script runs using whatever Python is installed on the test VM, and that is not necessarily Python 3.

tests_e2e/pipeline/scripts/execute_tests.sh
```diff
@@ -86,18 +86,27 @@ sudo find "$LOGS_DIRECTORY" -exec chown "$USER" {} \;
 #
 # Move the relevant logs to the staging directory
 #
+# Move the logs for failed tests to a temporary location
```
maddieford, line 89: Why do we move failed tests to a temporary location here?

narrieta: we move them back on line 105

the temp location makes it easy to separate them from other logs

nagworld9 (approved):

PR #2804, opened 2023-04-14
Add tests for no outbound connectivity

narrieta:
Adding a test for scenarios where the VM has no outbound connectivity to the Internet.

This test replaces the exitsing agent-nsg* DCR scenarios.

As part of this PR, I added infrastructure to customize the ARM template used to create the test VMs, as well as a property ('blocks_suite') that allows a test to stop execution of a test suite if it fails.


tests_e2e/orchestrator/lib/agent_test_loader.py
```diff
@@ -25,20 +25,39 @@
...
-    tests: List[Type[AgentTest]]
+    tests: List[TestInfo]
     # Images or image sets (as defined in images.yml) on which the suite must run.
     images: List[str]
     # The location (region) on which the suite must run; if empty, the suite can run on any location
     location: str
     # Whether this suite must run on its own test VM
     owns_vm: bool
+    # Customization for the ARM template used when creating the test VM
+    template: str
```
narrieta, line 60: Added a new property, 'template', to allow test to use custom ARM templates to create the test VMs

tests_e2e/orchestrator/lib/agent_test_loader.py
```diff
@@ -148,18 +167,22 @@ def _load_test_suite(description_file: Path) -> TestSuiteInfo:
...
 
         * name     - A string used to identify the test suite
-        * tests    - A list of the tests in the suite. Each test is specified by the path for its source code relative to
-                     WALinuxAgent/tests_e2e/tests.
+        * tests    - A list of the tests in the suite. Each test can be specified by a string (the path for its source code relative to
+                     WALinuxAgent/tests_e2e/tests), or a dictionary with two items:
+                        * source: the path for its source code relative to WALinuxAgent/tests_e2e/tests
+                        * blocks_suite: [Optional; boolean] If True, a failure on the test will stop execution of the test suite (i.e. the
+                          rest of the tests in the suite will not be executed). By default, a failure on a test does not stop execution of
+                          the test suite.
```
narrieta, line 178: Now tests can be described with a simple string (path to their source code), or an object with 2 properties: 'source' and 'blocks_suite'

tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
```diff
@@ -218,36 +218,44 @@ def create_environment_list(self) -> List[Dict[str, Any]]:
...
                         "c_vm_size": vm_size,
                         "c_vhd": vhd,
                         "c_test_suites": [suite_info],
-                        "c_env_name": f"{name}-{suite_info.name}",
+                        "c_env_name": env_name,
                         "c_marketplace_image_information_location": self._MARKETPLACE_IMAGE_INFORMATION_LOCATIONS[self.runbook.cloud],
-                        "c_shared_resource_group_location": self._SHARED_RESOURCE_GROUP_LOCATIONS[self.runbook.cloud]
-                    })
+                        "c_shared_resource_group_location": self._SHARED_RESOURCE_GROUP_LOCATIONS[self.runbook.cloud],
+                        "c_vm_tags": tags
```
narrieta, line 238: Introduced a new variable, "c_vm_tags", which is used to pass the custom templates to the UpdateArmTemplateHook (see wiki for details)

nagworld9 (approved):

tests_e2e/tests/no_outbound_connections/nsg_template.py
maddieford, line 1: When we need additional templates in the future, will we add a class for each template in the respective test directory? I'm wondering if there will be cases where multiple test suites will need to share an arm template

maddieford (approved):

tests_e2e/tests/no_outbound_connections/nsg_template.py
narrieta: in this case this is the only test using this template, but for shared templates we can always add them to the "lib" directory (or a new "templates" directory) and use the corresponding path in the test suite yaml definition.

```
template: "no_outbound_connections/nsg_template.py"
```

PR #2810, opened 2023-04-22
Agent update refactor supports GA versioning

nagworld9:
## Description

Issue # 

---

### PR information


GabstaMSFT:
test

tests_e2e/orchestrator/scripts/install-agent
```diff
@@ -87,22 +62,25 @@ printf "\n"
...
-# but AutoUpdate defaults to True so there is no need to do anything in that case.
-if [[ -e /etc/waagent.conf ]]; then
-  sed -i 's/AutoUpdate.Enabled=n/AutoUpdate.Enabled=y/g' /etc/waagent.conf
-fi
+python=$(get-agent-python)
+# Ensure that AutoUpdate is enabled. some distros, e.g. Flatcar have a waagent.conf in different path
+waagent_conf_path=$($python -c 'from azurelinuxagent.common.osutil import get_osutil; osutil=get_osutil(); print(osutil.agent_conf_file_path)')
+echo "Agent's conf path: $waagent_conf_path"
+sed -i 's/AutoUpdate.Enabled=n/AutoUpdate.Enabled=y/g' "$waagent_conf_path"
+# By default GAUpdates flag set to True, so that agent go through update logic to look for new agents.
```
narrieta, line 70: I think this should be done only for the GA Update tests

tests_e2e/tests/lib/retry.py
```diff
@@ -57,3 +57,23 @@ def retry_ssh_run(operation: Callable[[], Any]) -> Any:
                     raise
             log.warning("The operation failed, retrying in 30 secs.\n%s", e)
         time.sleep(30)
+
+
+def retry_if_not_found(operation: Callable[[], bool], attempts: int = 5) -> bool:
```
narrieta, line 62: "if_not_found" implies some kind of search, how about "retry_if_false". Let's also make the sleep time a parameter with a default value.

tests_e2e/tests/lib/virtual_machine.py
```diff
@@ -91,11 +91,39 @@ def _begin_restart(self) -> LROPoller:
...
+        """
+        Derived classes must provide the implementation for this method using their corresponding begin_create_or_update() implementation
+        """
+
     def __str__(self):
         return f"{self._identifier}"
 
 
-class VirtualMachine(VirtualMachineBaseClass):
+class VmMachine(VirtualMachineBaseClass):
```
narrieta, line 126: What was the name conflict that motivated this rename? I can't recall.

tests_e2e/tests/scripts/rsm_goal_state.py
```diff
@@ -0,0 +1,66 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-v', '--version', required=True)
+    args = parser.parse_args()
+
+    protocol = get_protocol_util().get_protocol(init_goal_state=False)
+    protocol.client.reset_goal_state(
+        goal_state_properties=GoalStateProperties.ExtensionsGoalState)
+
+    attempts = 5
+    while attempts > 0:
```
narrieta, line 51: should this be using your new retry function?

tests_e2e/tests/scripts/rsm_goal_state.py
```diff
@@ -0,0 +1,66 @@
...
+        protocol.client.update_goal_state()
+        goal_state = protocol.client.get_goal_state()
+        requested_version = get_requested_version(goal_state)
+        if requested_version == args.version:
+            print("Latest GS includes rsm requested version : {0}.".format(requested_version))
+            break
+        print("RSM requested version GS not available yet to the agent, checking again in 30 secs.")
+        attempts -= 1
+        time.sleep(30)
+
```
narrieta, line 61: should this fail if we never get the expected goal state?

tests_e2e/tests/lib/virtual_machine.py
```diff
@@ -91,11 +91,39 @@ def _begin_restart(self) -> LROPoller:
...
+            parameters = {}
+
+        log.info("Creating/Updating VM for %s", self._identifier)
+
+        poller: LROPoller = execute_with_retry(lambda: self._begin_create_or_update(parameters))
+
+        poller.wait(timeout=timeout)
+
+        if not poller.done():
+            raise TimeoutError(f"Failed to restart {self._identifier.name} after {timeout} seconds")
```
narrieta, line 114: why "Failed to restart"?

tests_e2e/tests/scripts/rsm_goal_state.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env pypy3
```
narrieta, line 1: probably a better name for the script would be "wait_for_rsm_goal_state.py"

tests_e2e/tests/scripts/rsm_goal_state.py
```diff
@@ -0,0 +1,66 @@
...
+    protocol.client.reset_goal_state(
+        goal_state_properties=GoalStateProperties.ExtensionsGoalState)
+
+    attempts = 5
+    while attempts > 0:
+        protocol.client.update_goal_state()
+        goal_state = protocol.client.get_goal_state()
+        requested_version = get_requested_version(goal_state)
+        if requested_version == args.version:
+            print("Latest GS includes rsm requested version : {0}.".format(requested_version))
```
narrieta, line 56: these should be logger.info rather than print to include a timestamp

tests_e2e/tests/agent_update/rsm_update.py
```diff
@@ -0,0 +1,189 @@
...
+
+        ]
+        return ignore_rules
+
+    def run(self) -> None:
+        # Allow agent to send supported feature flag
+        self._verify_agent_reported_supported_feature_flag()
+
+        log.info("*******Verifying the Agent Downgrade scenario*******")
+        self._mock_rsm_update("1.3.0.0")
```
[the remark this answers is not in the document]
narrieta: also line 88

tests_e2e/tests/agent_update/rsm_update.py
```diff
@@ -0,0 +1,189 @@
...
+
+        # verify no version update. There is bug in CRP and will enable once it's fixed
+        # log.info("*******Verifying the no version update scenario*******")
+        # self._prepare_rsm_update("1.3.1.0")
+        # self._verify_guest_agent_update("1.3.1.0")
+
+    def _check_rsm_gs(self, requested_version: str) -> None:
+        # This checks if RSM GS available to the agent after we mock the rsm update request
+        output = self._ssh_client.run_command(f"rsm_goal_state.py --version {requested_version}", use_sudo=True)
+        log.info('Verifying requested version GS available to the agent \n%s', output)
```
narrieta, line 90: if we don't get the RSM goal state we should raise an exception

tests_e2e/tests/scripts/agent-update-config
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
narrieta, line 1: we need a better name for this script. "agent-update-config" is too generic

tests_e2e/tests/agent_update/rsm_update.py
```diff
@@ -0,0 +1,189 @@
...
+        # Verify downgrade scenario
+        self._verify_guest_agent_update("1.3.0.0")
+
+        # Verify upgrade scenario
+        log.info("*******Verifying the Agent Upgrade scenario*******")
+        self._mock_rsm_update("1.3.1.0")
+        self._check_rsm_gs("1.3.1.0")
+        self._verify_guest_agent_update("1.3.1.0")
+
+        # verify no version update. There is bug in CRP and will enable once it's fixed
```
narrieta, line 82: what is the current bug?

narrieta (commented):
Added some comments on the end-to-end test

azurelinuxagent/common/conf.py
```diff
@@ -502,6 +503,14 @@ def get_monitor_network_configuration_changes(conf=__conf__):
     return conf.get_switch("Monitor.NetworkConfigurationChanges", False)
 
 
+def get_ga_updates_enabled(conf=__conf__):
+    """
+    If True, the agent go through update logic to look for new agents otherwise it will stop agent updates.
+    NOTE: This option is needed in e2e tests to control agent updates.
```
narrieta, line 509: If only needed by the tests, this should be a Debug/experimental option

azurelinuxagent/common/event.py
```diff
@@ -127,6 +127,7 @@ class WALAEventOperation:
     Update = "Update"
     VmSettings = "VmSettings"
     VmSettingsSummary = "VmSettingsSummary"
+    FeatureFlag = "FeatureFlag"
```
narrieta, line 130: "FeatureFlag" is too generic, can we use a different name?

azurelinuxagent/ga/guestagent.py
```diff
@@ -0,0 +1,316 @@
+import json
```
narrieta, line 1: this module needs a better name, "guestagent.py" is too generic

azurelinuxagent/ga/guestagent.py
```diff
@@ -0,0 +1,316 @@
...
+from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
+from azurelinuxagent.common.version import AGENT_DIR_PATTERN, AGENT_NAME, CURRENT_VERSION
+from azurelinuxagent.ga.exthandlers import HandlerManifest
+
+AGENT_ERROR_FILE = "error.json"  # File name for agent error record
+AGENT_MANIFEST_FILE = "HandlerManifest.json"
+MAX_FAILURE = 3  # Max failure allowed for agent before declare bad agent
+
+
+class GAUpdateReportState(object):
```
narrieta, line 21: These 2 variables are globals. We should not introduce globals.

azurelinuxagent/ga/guestagent.py
```diff
@@ -0,0 +1,316 @@
...
+    """
+    This class is primarily used to maintain the in-memory persistent state for the agent updates.
+    This state will be persisted throughout the current service run and might be modified by external classes.
+    """
+    report_error_msg = ""
+    report_expected_version = FlexibleVersion("0.0.0.0")
+
+
+class GuestAgent(object):
```
narrieta, line 30: We should split this class in code to download the agent and code to get the information of agents that are already downloaded. We can follow offline.

narrieta (commented):
comments for guestagent.py and update.py

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,340 @@
...
+    def __get_available_agents_on_disk():
+        available_agents = [agent for agent in AgentUpdateHandler.__get_all_agents_on_disk() if agent.is_available]
+        return sorted(available_agents, key=lambda agent: agent.version, reverse=True)
+
+    @staticmethod
+    def __get_all_agents_on_disk():
+        path = os.path.join(conf.get_lib_dir(), "{0}-*".format(AGENT_NAME))
+        return [GuestAgent.from_installed_agent(path=agent_dir) for agent_dir in glob.iglob(path) if os.path.isdir(agent_dir)]
+
+    def __log_event(self, level, msg_, success_=True):
```
narrieta, line 242: Why the trailing underscore in msg_ and success_? the convention in Python is to do that when using a name that collides with a built-in name, but that does not seem to be the case here

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,340 @@
...
+                return
+
+            msg_ = "Goal state {0} is requesting a new agent version {1}, will update the agent before processing the goal state.".format(
+                self._gs_id, str(requested_version))
+            self.__log_event(LogLevel.INFO, msg_)
+
+            try:
+                agent = self.__download_and_get_agent(goal_state, agent_family, agent_manifest, requested_version)
+
+                if not agent.is_available:
```
narrieta, line 303: When would this condition be True?

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,340 @@
...
+    @staticmethod
+    def __purge_extra_agents_from_disk(known_agents):
+        """
+        Remove from disk all directories and .zip files of unknown agents
+        (without removing the current, running agent).
+        """
+        path = os.path.join(conf.get_lib_dir(), "{0}-*".format(AGENT_NAME))
+
+        known_versions = [agent.version for agent in known_agents]
+        known_versions.append(CURRENT_VERSION)
```
narrieta, line 187: I think you should pass CURRENT_VERSION explicitly as part of known_agents to make the purpose of this method clearer

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,340 @@
...
+            # (the expectation here being that if we get request to a downgrade,
+            # there's a good reason for not wanting the current version).
+            prefix = "downgrade"
+            try:
+                # We should always have an agent directory for the CURRENT_VERSION
+                agents_on_disk = AgentUpdateHandler.__get_available_agents_on_disk()
+                current_agent = next(agent for agent in agents_on_disk if agent.version == CURRENT_VERSION)
+                msg = "Marking the agent {0} as bad version since a downgrade was requested in the GoalState, " \
+                      "suggesting that we really don't want to execute any extensions using this version".format(CURRENT_VERSION)
+                logger.info(msg)
```
narrieta, line 220: can you send this to telemetry, too?

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,340 @@
...
+                current_agent.mark_failure(is_fatal=True, reason=msg)
```
narrieta, line 221: The purge method should not delete blacklisted versions (it can delete the code, but leave the directory and error file on disk), otherwise the daemon may download and use them again

narrieta (commented):
completed my review

azurelinuxagent/ga/update.py
```diff
@@ -606,18 +519,19 @@ def _processing_new_extensions_goal_state(self):
...
-    def _process_goal_state(self, exthandlers_handler, remote_access_handler):
+    def _process_goal_state(self, exthandlers_handler, remote_access_handler, agent_update_handler):
         protocol = exthandlers_handler.protocol
 
         # update self._goal_state
         if not self._try_update_goal_state(protocol):
-            # agent updates and status reporting should be done even when the goal state is not updated
-            self.__update_guest_agent(protocol)
-            self._report_status(exthandlers_handler)
+            # status reporting should be done even when the goal state is not updated
```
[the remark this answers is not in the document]
nagworld9: since now we get new goal state for every rsm request.

azurelinuxagent/ga/update.py
```diff
@@ -635,7 +549,11 @@ def _process_goal_state(self, exthandlers_handler, remote_access_handler):
                 CGroupConfigurator.get_instance().check_cgroups(cgroup_metrics=[])
 
             # report status before processing the remote access, since that operation can take a long time
-            self._report_status(exthandlers_handler)
+            agent_update_status = agent_update_handler.get_vmagent_update_status()
+            self._report_status(exthandlers_handler, agent_update_status)
+
+            # Logging after agent reports supported feature flag so this msg in sync with report status
```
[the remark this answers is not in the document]
nagworld9: yes, we do but we are logging this after we report status because e2e tests rely on this in agent update scenario

azurelinuxagent/ga/update.py
```diff
@@ -737,6 +618,24 @@ def _report_extensions_summary(self, vm_status):
                 logger.warn(msg)
                 add_event(op=WALAEventOperation.GoalState, is_success=False, message=msg)
 
+    def _log_agent_supports_versioning_or_not(self):
```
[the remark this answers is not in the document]
nagworld9: Feature is supported by default but cx can turn it off if GA versioning flag is disabled. 

and also, we introduced this logging when e2e need for it to know feature is supported or not by the agent.

tests/data/wire/ext_conf_missing_family.xml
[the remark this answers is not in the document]
nagworld9: This would be used in one of our test case which checks for missing GA family case.

tests_e2e/tests/agent_update/rsm_update.py
```diff
@@ -0,0 +1,189 @@
...
+            #
+            {
+                'message': r"Agent WALinuxAgent-9.9.9.9 is permanently blacklisted"
+            }
+
+        ]
+        return ignore_rules
+
+    def run(self) -> None:
+        # Allow agent to send supported feature flag
```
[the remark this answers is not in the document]
nagworld9: RSM update will not be attempted on vm creation as CRP requires feature flag to drive rsm request.

tests_e2e/tests/agent_update/rsm_update.py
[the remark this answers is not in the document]
nagworld9: Previous PRs I was suggested to rename to mock.

tests_e2e/tests/agent_update/rsm_update.py
nagworld9: CRP not populating GA family in GS if requested version is same as last successful attempted updated version.

Before I was on vacation, Kashif mentioned change was merged, waiting for next CRP deployment

tests_e2e/tests/lib/virtual_machine.py
nagworld9: Good catch, failed to restart doesn't make sense. I'll change

tests_e2e/tests/scripts/rsm_goal_state.py
nagworld9: Let me see If I can reuse

tests_e2e/tests/scripts/rsm_goal_state.py
nagworld9: will this log to agent log If I use logger.info? I don't want to add to agent log. This will create confusion with agent rsm related messages.

tests_e2e/tests/scripts/rsm_goal_state.py
nagworld9: yes, it should fail. I'll update.

azurelinuxagent/common/event.py
narrieta: If it is meant for multiple features I think the message should be json or something like that so that we can parse it for individual features. At this point it is used only for agent versioning. If you want to use it for all features, you should also use it for multi-config and fastrack.

I think you should make this one specific to versioning

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,340 @@
...
+        self.last_warning_time = datetime.datetime.min
+
+
+class AgentUpdateHandler(object):
+
+    def __init__(self, protocol):
+        self._protocol = protocol
+        self._ga_family = conf.get_autoupdate_gafamily()
+        self._autoupdate_enabled = conf.get_autoupdate_enabled()
+        self._gs_id = self._protocol.get_goal_state().extensions_goal_state.id
```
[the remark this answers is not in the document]
narrieta: maybe initialize it to "not-initialized", "unknown" or something like that? we are going to remove the instance of the goal state from the protocol class

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,340 @@
...
+
+            self._gs_id = goal_state.extensions_goal_state.id
+            agent_family = self.__get_agent_family_from_last_gs(goal_state)
+            requested_version = self.__get_requested_version(agent_family)
+            agent_manifest = None  # This is to make sure fetch agent manifest once per update
+
+            if requested_version is None:
+                if conf.get_enable_ga_versioning():  # log the warning only when ga versioning is enabled
+                    warn_msg = "Missing requested version in agent family: {0} for incarnation: {1}, fallback to largest version update".format(self._ga_family, self._gs_id)
+                    self.__log_event(LogLevel.WARNING, warn_msg)
```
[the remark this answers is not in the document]
narrieta: i don't think this would swamp the log, if goal states were that frequent then we have bigger problems, since we log quite a few messages in each goal state.

now, coming to think about it, we should probably log it when we actually attempt the update, otherwise i think the message is not interesting

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,340 @@
...
+            if requested_version == CURRENT_VERSION:
+                return
+
+            # Check if an update is allowed
+            if not self.__should_update_agent(requested_version):
+                return
+
+            msg_ = "Goal state {0} is requesting a new agent version {1}, will update the agent before processing the goal state.".format(
+                self._gs_id, str(requested_version))
+            self.__log_event(LogLevel.INFO, msg_)
```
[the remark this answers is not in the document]
narrieta: periodic messages have produced issues in the past. we should avoid them (i've been removing them as i touch code)

if the messages change as the code evolves, it easy to miss that, and it can create very large logs

azurelinuxagent/ga/agent_update.py
narrieta: why would it be missing? if we are concerned about that we should check during publishing, probably in the agent-update test

azurelinuxagent/ga/update.py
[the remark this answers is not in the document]
narrieta: but we still need the fallback when the requested version is not in the goal state

azurelinuxagent/ga/update.py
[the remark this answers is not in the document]
narrieta: i see... i think your test should instead check the status that we save to the history folder, would be more robust than checking the log

azurelinuxagent/ga/update.py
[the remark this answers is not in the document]
narrieta: (see above comment: the test can check the status that the agent reports rather than the log)

If we want to log that the customer disabled the feature let's add a message specific for that. we have several of those messages during agent initialization at the beginning of the run() method

azurelinuxagent/ga/agent_update.py
nagworld9: My concern is while extracting the zip, it may tamper the folder and mess up the files

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,340 @@
...
+            agent_family = self.__get_agent_family_from_last_gs(goal_state)
+            requested_version = self.__get_requested_version(agent_family)
+            agent_manifest = None  # This is to make sure fetch agent manifest once per update
+
+            if requested_version is None:
+                if conf.get_enable_ga_versioning():  # log the warning only when ga versioning is enabled
+                    warn_msg = "Missing requested version in agent family: {0} for incarnation: {1}, fallback to largest version update".format(self._ga_family, self._gs_id)
+                    self.__log_event(LogLevel.WARNING, warn_msg)
+                    GAUpdateReportState.report_error_msg = warn_msg
+                agent_manifest = goal_state.fetch_agent_manifest(agent_family.name, agent_family.uris)
```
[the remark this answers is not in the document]
nagworld9: I'll change it in different task

azurelinuxagent/common/conf.py
narrieta: @nagworld9 - did we decide this to expose this option to the user? if yes, let's add it to README, if not, let's make it Debug

narrieta: sorry, what is the difference between GAUpdates.Enabled and Debug.EnableGAVersioning

narrieta (approved):
Approved, with just one question on the 2 conf settings

azurelinuxagent/common/conf.py
nagworld9: No, this is not exposed to the user. I'll change to debug. 

Debug.EnableGAVersioning- This flag is like kill switch. Once this is trun off our code will go to self update path.

GAUpdates.Enabled- This flag is to stop the agent updates. Only used in e2e tests.

nagworld9: I'll change as part of next work item.

azurelinuxagent/ga/agent_update.py
```diff
@@ -0,0 +1,340 @@
...
+        """
+        Returns the package of the given Version found in the manifest. If not found, returns exception
+        """
+        for pkg in agent_manifest.pkg_list.versions:
+            if FlexibleVersion(pkg.version) == version:
+                # Found a matching package, only download that one
+                return pkg
+
+        raise Exception("No matching package found in the agent manifest for requested version: {0} in goal state incarnation: {1}, "
+                        "skipping agent update".format(str(version), self._gs_id))
```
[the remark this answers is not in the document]
maddieford: `raise Exception("No matching package found in the agent manifest for requested version: {0} in goal state incarnation: {1}, "   "skipping agent update".format(str(version), self._gs_id))`


I see two separate strings:
- "No matching package found in the agent manifest for requested version: {0} in goal state incarnation: {1}, "
-  "skipping agent update".format(str(version), self._gs_id)

the second string is taking two values in format() but it's the first string that needs to be formatted


Maybe this is some syntax i'm unfamiliar with

maddieford (approved):

PR #2814, opened 2023-05-02
Fix name of single IB device when provisioning RDMA

longlimsft:
The current code assumes the ipoib interface name is ib0 when single IB interface is provisioned. This is not always true when udev rules are used to rename to other names like ibPxxxxx.

Fix this by searching any interface name starting with "ib".

## Description

Issue # 

---

### PR information


longlimsft:
> It would be safer to use sysfs to enumerate infiniband devices rather than rely on device naming.

Yes, update_iboip_interfaces() is doing exactly that.

In the single IP case (in this pull request), the KVP value passed from the host has a slightly different format, so we know there is only one IP interface to configure on this VM. There is no need to enumerate on the interfaces.

cjp256:
> > It would be safer to use sysfs to enumerate infiniband devices rather than rely on device naming.
> 
> Yes, update_iboip_interfaces() is doing exactly that.
> 
> In the single IP case (in this pull request), the KVP value passed from the host has a slightly different format, so we know there is only one IP interface to configure on this VM. There is no need to enumerate on the interfaces.

Not that I'm asking for the change, it's just a thought.  As an alternative to relying on name anywhere, the inifiniband devices could be enumerated by:

1. /sys/class/net/<name>/type == 32
2. /sys/class/infiniband/<ib-device-name>,  where you can find the net device underneath, for ex:
/sys/class/infiniband/mlx5_ib0/device/net/ib0/

```
cpatterson@test-hpc-ib:/sys/class/net/ib0$ grep ^32$ /sys/class/net/*/type
/sys/class/net/ib0/type:32
cpatterson@test-hpc-ib:/sys/class/net/ib0$ cat /sys/class/net/ib0/address
00:00:01:48:fe:80:00:00:00:00:00:00:00:15:5d:ff:fd:33:ff:36
cpatterson@test-hpc-ib:/sys/class/net/ib0$ cat /sys/class/net/ib0/type
32
cpatterson@test-hpc-ib:/sys/class/net/ib0$ ls /sys/class/infiniband
mlx5_ib0
cpatterson@test-hpc-ib:/sys/class/net/ib0$ cat /sys/class/infiniband/mlx5_ib0/device/net/ib0/address
00:00:01:48:fe:80:00:00:00:00:00:00:00:15:5d:ff:fd:33:ff:36
```

longlimsft:
> > > It would be safer to use sysfs to enumerate infiniband devices rather than rely on device naming.
> > 
> > 
> > Yes, update_iboip_interfaces() is doing exactly that.
> > In the single IP case (in this pull request), the KVP value passed from the host has a slightly different format, so we know there is only one IP interface to configure on this VM. There is no need to enumerate on the interfaces.
> 
> Not that I'm asking for the change, it's just a thought. As an alternative to relying on name anywhere, the inifiniband devices could be enumerated by:
> 
> 1. `/sys/class/net/<name>/type == 32`
> 2. `/sys/class/infiniband/<ib-device-name>`,  where you can find the net device underneath, for ex:
>    /sys/class/infiniband/mlx5_ib0/device/net/ib0/
> 
> ```
> cpatterson@test-hpc-ib:/sys/class/net/ib0$ grep ^32$ /sys/class/net/*/type
> /sys/class/net/ib0/type:32
> cpatterson@test-hpc-ib:/sys/class/net/ib0$ cat /sys/class/net/ib0/address
> 00:00:01:48:fe:80:00:00:00:00:00:00:00:15:5d:ff:fd:33:ff:36
> cpatterson@test-hpc-ib:/sys/class/net/ib0$ cat /sys/class/net/ib0/type
> 32
> cpatterson@test-hpc-ib:/sys/class/net/ib0$ ls /sys/class/infiniband
> mlx5_ib0
> cpatterson@test-hpc-ib:/sys/class/net/ib0$ cat /sys/class/infiniband/mlx5_ib0/device/net/ib0/address
> 00:00:01:48:fe:80:00:00:00:00:00:00:00:15:5d:ff:fd:33:ff:36
> ```

Thank you. Yes, I agree this is the best way to look for ipoib interfaces. The current code has an assumption that all such interfaces will start with "ib". Your method can work with any interface names.

Given the prefix "ib" unlikely to change anytime soon for an IB interface, I suggest this patch goes in as is. We can work on another patch to better handle those interface names as you suggested.

cjp256 (approved):
lgtm

cjp256 (approved):

azurelinuxagent/common/rdma.py
```diff
@@ -419,28 +419,33 @@ def update_iboip_interfaces(self, mac_ip_array):
...
             n += 1
 
-        if not found_ib0:
-            raise Exception("ib0 is not available")
+        if not found_ib:
+            raise Exception("ib is not available")
+
+        ibname = found_ib.groups()[0]
+        if shellutil.run("ifconfig {0} up".format(ibname)) != 0:
+            raise Exception("Could run ifconfig {0} up".format(ibname))
```
narrieta, line 441: typo: "Could **not** run..."

azurelinuxagent/common/rdma.py
```diff
@@ -419,28 +419,33 @@ def update_iboip_interfaces(self, mac_ip_array):
...
 
         netmask = 16
         logger.info("RDMA: configuring IPv4 addr and netmask on ipoib interface")
         addr = '{0}/{1}'.format(ipv4_addr, netmask)
-        if shellutil.run("ifconfig ib0 {0}".format(addr)) != 0:
-            raise Exception("Could set addr to {0} on ib0".format(addr))
+        if shellutil.run("ifconfig {0} {1}".format(ibname, addr)) != 0:
+            raise Exception("Could set addr to {0} on {1}".format(addr, ibname))
```
narrieta, line 447: "Could **not**..."

narrieta:
@longlimsft  could you base your PR off the develop branch? thanks

azurelinuxagent/common/rdma.py
longlimsft: Fixed

PR #2816, opened 2023-05-09
Port NSBSD system to the latest version of waagent

sebastienb-stormshield:
## Description

NSBSD system was lagging behind with an old version of both python and the azure agent. We updated the agent on our appliances, but had to make some small isolated changes to make it work, most of them remain in NSBSD dedicated part.

The only "intrusive" change is the new `get_root_username` function in osutil, as NSBSD root user is named `admin` and not `root`. This causes the log directory creation to fail since by default it is created for the `root` user.

Alternatively, the directory creation function (`fileutil.mkdir`) could directly take the uid,gid of the owner instead of resolving it on its own, which would allow us to pass `0` on both and not introduce a new function.

---

### PR information
- [x] (N/A) If applicable, the PR references the bug/issue that it fixes in the description.

### DIFF
--- a/azurelinuxagent/agent.py
+++ b/azurelinuxagent/agent.py
@@ -104,7 +104,7 @@ class Agent(object):
             if os.path.isfile(ext_log_dir):
                 raise Exception("{0} is a file".format(ext_log_dir))
             if not os.path.isdir(ext_log_dir):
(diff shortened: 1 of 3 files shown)


sebastienb-stormshield:
More changes are planned. Will submit everything in a single PR.

PR #2820, opened 2023-05-16
Bug fixes for end-to-end tests

narrieta:
* Logs were not being collected on test failures
* Intermittent errors checking the provisioning state when executing for extensions
* When invoking a runbook on an existing VM, the 'c_cloud' was missing

### DIFF
--- a/tests_e2e/orchestrator/lib/agent_test_suite.py
+++ b/tests_e2e/orchestrator/lib/agent_test_suite.py
@@ -135,7 +135,7 @@ class AgentTestSuite(LisaTestSuite):
 
         self.__context = self._Context(
             vm=VmIdentifier(
-                cloud=self._get_required_parameter(variables, "c_cloud"),
+                cloud=self._get_required_parameter(variables, "cloud"),
                 location=self._get_required_parameter(variables, "c_location"),
                 subscription=node.features._platform.subscription_id,
                 resource_group=node_context.resource_group_name,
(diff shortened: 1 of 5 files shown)


tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -135,7 +135,7 @@ def _initialize(self, node: Node, variables: Dict[str, Any], lisa_working_path:
 
         self.__context = self._Context(
             vm=VmIdentifier(
-                cloud=self._get_required_parameter(variables, "c_cloud"),
```
narrieta, line 138: The 'c_cloud' variable was not being passed to runbooks when running on an existing VM.

While fixing this I realized there is no need for this variable, since the AgentTestSuite and AgentTestSuiteCombinator can just use 'cloud' directly from the runbook.

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -454,8 +454,6 @@ def _execute_test_suite(self, suite: TestSuiteInfo) -> bool:
         suite_full_name = f"{suite_name}-{self.context.environment_name}"
         suite_start_time: datetime.datetime = datetime.datetime.now()
 
-        success: bool = True  # True if all the tests succeed
```
narrieta, line 457: this variable was a leftover of a previous iteration of the code and _execute_test_suite was always returning True (i.e. Success, so VM logs were not being collected on test failures)

tests_e2e/tests/lib/vm_extension.py
```diff
@@ -107,9 +107,7 @@ def enable(
                 extension_parameters
             ).result(timeout=_TIMEOUT))
 
-        if result.provisioning_state not in ('Succeeded', 'Updating'):
```
narrieta, line 110: I copied this check from the original prototype, but it has been producing intermittent failures. Depending on the timing, it can be Succeeded, Updating or just None. I double-checked that on extension failure the call to result() in the previous line will raise an exception anyway, so this check is actually not needed and now we just log an INFO.

maddieford (approved):

PR #2824, opened 2023-05-19
Add test for extensions disabled; refactor VirtualMachine and VmExtension

narrieta:
Test for disabling extension processing; replaces the current agent-extensions-toggled DCR scenario.

As part of this change, I also did significant refactoring of the  VirtualMachine and VmExtension classes (and a few bug fixes)


tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -203,12 +203,26 @@ def context(self):
...
     # Test suites within the same runbook may be executed concurrently, and setup needs to be done only once.
-    # We use this lock to allow only 1 thread to do the setup. Setup completion is marked using the 'completed'
+    # We use these locks to allow only 1 thread to do the setup. Setup completion is marked using the 'completed'
     # file: the thread doing the setup creates the file and threads that find that the file already exists
     # simply skip setup.
     #
+    _working_directory_lock = RLock()
     _setup_lock = RLock()
 
+    def _create_working_directory(self) -> None:
```
narrieta, line 213: The runbook has an option to skip setup. If that option is set, the working directory still needs to be created, so I separated this into its own method, which is always called.

tests_e2e/orchestrator/scripts/agent-service
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env bash
```
narrieta, line 1: utility taken from #2810

tests_e2e/tests/bvts/extension_operations.py
```diff
@@ -35,7 +35,7 @@
 from tests_e2e.tests.lib.identifiers import VmExtensionIds, VmExtensionIdentifier
 from tests_e2e.tests.lib.logging import log
 from tests_e2e.tests.lib.ssh_client import SshClient
-from tests_e2e.tests.lib.vm_extension import VmExtension
```
narrieta, line 38: the rename is explained in another comment below

tests_e2e/tests/bvts/extension_operations.py
```diff
@@ -65,15 +65,15 @@ def run(self):
...
-        custom_script_2_1 = VmExtension(
+        custom_script_2_1 = VirtualMachineExtensionClient(
             self._context.vm,
             VmExtensionIdentifier(VmExtensionIds.CustomScript.publisher, VmExtensionIds.CustomScript.type, "2.1"),
             resource_name="CustomScript")
 
         if is_arm64:
             log.info("Installing %s", custom_script_2_1)
         else:
-            log.info("Updating %s to %s", custom_script_2_0, custom_script_2_1)
```
narrieta, line 76: this was just logging "Updating CustomScript to CustomScript"

tests_e2e/tests/extensions_disabled.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python3
```
narrieta, line 1: new test

tests_e2e/tests/lib/azure_client.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class AzureClient:
+    """
+    Utilities for classes using the Azure SDK.
+    """
+    _DEFAULT_TIMEOUT = 10 * 60  # (in seconds)
+
+    @staticmethod
+    def _execute_async_operation(operation: Callable[[], LROPoller], operation_name: str, timeout: int) -> Any:
```
narrieta, line 33: The calls to extensions were missing the check for timeouts. To help avoiding similar issues, I refactored the pattern to invoke async SDK methods in this function.

tests_e2e/tests/lib/virtual_machine.py
```diff
@@ -1,155 +0,0 @@
-# Microsoft Azure Linux Agent
```
narrieta, line 1: The initial prototype tried to define a common interface for Virtual Machine and Scale Set operations, and I tried to maintain that model in my first implementation of this code.

However, that model does not quite work. Though there are some similarities in the APIs and the invocation patterns, the differences and significant enough to make a common implementation awkward. I already had several TODO comments to revisit that decision. Since I needed to change those classes to support the update operation, I decided to revisit the design now. I removed the implementation for scale sets and merged the implementation of the base class into VirtualMachine.

I renamed VirtualMachine to VirtualMachineClient, because the former conflicts with class VirtualMachine in the SDK (the get_description method, needed by #2810, returns an SDK's VirtualMachine).

I made similar changes on VmExtension (now VirtualMachineExtensionClient).

@nagworld9 This refactoring affects #2810, but the impact in your test should be minimal (just a few renames).

maddieford: leaving a question here that may help me with migrating my current dcr scenario (ext-workflow)

in the dcr repo we have classes to define each extension we use for testing (https://msazure.visualstudio.com/One/_git/Compute-Runtime-Tux-Pipeline?path=/dungeon_crawler/scenarios/extensions.py)

We're currently using the GuestAgentTestDcr extension for testing, and I'm trying to decide the best place to define a similar class in the new e2e pipeline. 

I haven't seen any similar class made for RunCommand or CSE yet in the e2e pipeline, do you have any recommendations on where/if I should make this class. I'm wondering if it should be an implementation of virtual machine extension client? Let me know if you have any ideas

tests_e2e/orchestrator/lib/agent_test_suite.py
narrieta: During development when executing a test suite multiple times on the same machine. Once setup is done, skipping it saves some time when executing the test suite. This helps to follow a development model in which one does small changes when writing a test and executes the test frequently to try those changes.

tests_e2e/tests/lib/virtual_machine.py
narrieta: The new code base is organized differently: extension operations are defined in tests_e2e/tests/lib/virtual_machine_extension_client.py, extension names in tests_e2e/tests/lib/identifiers.py, SSH operations on tests_e2e/tests/lib/ssh_client.py, etc

Ping me when you get to specific items and we can figure out where they should go.

maddieford (approved):

PR #2826, opened 2023-05-24
Add test for osProfile.linuxConfiguration.provisionVMAgent

narrieta:
- Test for Add test for osProfile.linuxConfiguration.provisionVMAgent (replaces the agent-not-provisioned DCR scenario)
- Added a boolean property, install_test_agent, to the test suite description to skip installation of the test Agent
- Ignore 'No handler status found for Microsoft.Azure.Extensions.CustomScript' on extensions_disabled test


tests_e2e/test_suites/no_outbound_connections.yml
```diff
@@ -16,5 +16,5 @@ tests:
   - "bvts/vm_access.py"
   - "no_outbound_connections/check_fallback_to_hgap.py"
 images: "random(endorsed)"
-template: "no_outbound_connections/nsg_template.py"
```
narrieta, line 19: renamed 'nsg_template.py' to 'template.py' for consistency with the agent_not_provisioned suite

tests_e2e/tests/extensions_disabled.py
```diff
@@ -81,6 +82,11 @@ def run(self):
             .is_greater_than(pytz.utc.localize(disabled_timestamp))
         log.info("The VM Agent reported status after extensions were disabled, as expected.")
 
+    def get_ignore_error_rules(self) -> List[Dict[str, Any]]:
+        return [
+            {'message': 'No handler status found for Microsoft.Azure.Extensions.CustomScript'},
```
narrieta, line 87: This message showed up in today's daily run. Will be there or not depending on the time of the test

tests_e2e/tests/lib/agent_log.py
```diff
@@ -209,6 +209,15 @@ def get_errors(self) -> List[AgentLogRecord]:
                 'if': lambda r: DISTRO_NAME == 'ubuntu' and DISTRO_VERSION >= '22.00'
             },
             #
+            # Old daemons can produce this message
```
narrieta, line 212: Since the new test suite disables the agent, agent update does not happen and this message shows up in some distros.

tests_e2e/test_suites/agent_not_provisioned.yml
```diff
@@ -0,0 +1,12 @@
+#
+# Disables Agent provisioning using osProfile.linuxConfiguration.provisionVMAgent and verifies that the agent is disabled
+# and extension operations are not allowed.
+#
+name: "AgentNotProvisioned"
+tests:
+  - "agent_not_provisioned/agent_not_provisioned.py"
+images: "ubuntu_2004"
```
narrieta, line 8: SSH is not being enabled correctly in a few distros when the agent is not provisioned. A quick check gave me the impression it may be an issue with cloud-init. Restricting this suite to Ubuntu 20 while I investigate, all make it random after.

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -317,11 +317,14 @@ def _setup_node(self) -> None:
         tarball_path: Path = Path("/tmp/waagent.tar")
         log.info("Creating %s with the files need on the test node", tarball_path)
         log.info("Adding orchestrator/scripts")
-        run_command(['tar', 'cvf', str(tarball_path), '--transform=s,.*/,bin/,', '-C', str(self.context.test_source_directory/"orchestrator"/"scripts"), '.'])
```
narrieta, line 320: Adding '.' to the tarball was overwriting the attributes of the home directory on the test machine (we extract the tarball with cwd set to $HOME). This was causing SSH issues on some distros, on which the SSH server checks the mode of $HOME and refuses the connection if it is too open.

Instead of adding '.', now we cd to the source directory and add '*'. The other 2 invocations of tar were OK, but I also changed them for consistency.

maddieford (approved):

PR #2842, opened 2023-06-13
Add test for FIPS

narrieta:
Replaces DCR's agent-fips


tests_e2e/test_suites/agent_bvt.yml
```diff
@@ -1,8 +1,8 @@
 name: "AgentBvt"
 tests:
-  - "bvts/extension_operations.py"
```
narrieta, line 3: renamed directory to make it consistent with other tests: the directory is name is the same as the test suite's

tests_e2e/test_suites/extensions_disabled.yml
```diff
@@ -4,6 +4,6 @@
 #
 name: "ExtensionsDisabled"
 tests:
-  - "extensions_disabled.py"
```
narrieta, line 7: moved the test to its own directory for consistency

tests_e2e/tests/agent_bvt/extension_operations.py
```diff
@@ -58,7 +58,7 @@ def run(self):
             log.info("Installing %s", custom_script_2_0)
             message = f"Hello {uuid.uuid4()}!"
             custom_script_2_0.enable(
-                settings={
+                protected_settings={
```
narrieta, line 61: using protected settings also excercises handling of the tenant certificate

tests_e2e/tests/lib/virtual_machine_client.py
```diff
@@ -54,9 +57,9 @@ def __init__(self, vm: VmIdentifier):
             base_url=cloud.endpoints.resource_manager,
             credential_scopes=[cloud.endpoints.resource_manager + "/.default"])
 
-    def get_description(self) -> VirtualMachine:
+    def get_model(self) -> VirtualMachine:
```
narrieta, line 60: I call this method "get_description" because I wasn't sure what is returning and the documentation just said that it returns the "description" of the VM. It actually returns the model.

tests_e2e/tests/lib/virtual_machine_client.py
```diff
@@ -114,6 +117,23 @@ def restart(self, timeout: int = AzureClient._DEFAULT_TIMEOUT) -> None:
...
             timeout=timeout)
 
+        if not wait_for_boot:
+            return
+
+        start = datetime.datetime.now()
+        while datetime.datetime.now() < start + boot_timeout:
+            log.info("Waiting for VM %s to boot", self._identifier)
+            time.sleep(15)  # Note that we always sleep at least 1 time, to give the reboot time to start
+            instance_view = self.get_instance_view()
```
nagworld9, line 127: does it throw an error if vm is unavailable during restart? It the expection is expected, we are not handling it here.

tests_e2e/tests/lib/virtual_machine_client.py
```diff
@@ -114,6 +117,23 @@ def restart(self, timeout: int = AzureClient._DEFAULT_TIMEOUT) -> None:
...
+            power_state = [s.code for s in instance_view.statuses if "PowerState" in s.code]
+            if len(power_state) != 1:
+                raise Exception(f"Could not find PowerState in the instance view statuses:\n{json.dumps(instance_view.statuses)}")
+            log.info("VM's Power State: %s", power_state[0])
+            if power_state[0] == "PowerState/running":
```
nagworld9, line 132: how do we guarantee that this state is after vm restart? It's possible that vm request not performed due to something else but vm instance view is from previous state and this check could pass. I think checking timestamp of the status make sure this state is after vm restart?

narrieta: there is no guarantee, but we are not interested on the reboot itself, but rather on whether FIPS is enabled or not (which we check after this). if that checks fails because the machine did not reboot (or any other reason) then the test will fail

during my experiments, the reboot started within 2 or 3 seconds so 15 seconds wait seemed reasonable, but you are right that there is no guarantee about that

checking on some sort of timestamp is a good suggestion, though, and i'll add an extra check around that. the timestamp in the instance view is not appropriate for this check, though. i'll experiment with the service start in the agent log or the uptime of the system. thanks for the suggestion.

tests_e2e/tests/lib/virtual_machine_client.py
```diff
@@ -54,9 +57,9 @@ def __init__(self, vm: VmIdentifier):
             base_url=cloud.endpoints.resource_manager,
             credential_scopes=[cloud.endpoints.resource_manager + "/.default"])
 
-    def get_description(self) -> VirtualMachine:
+    def get_model(self) -> VirtualMachine:
         """
-        Retrieves the description of the virtual machine.
+        Retrieves the model of the virtual machine.
         """
         log.info("Retrieving description for %s", self._identifier)
```
maddieford, line 64: nit suggestion: update 'description' to be 'model' in this log

maddieford (approved):

tests_e2e/tests/lib/virtual_machine_client.py
narrieta: thanks, will do

tests_e2e/tests/lib/virtual_machine_client.py
narrieta: done

tests_e2e/tests/lib/virtual_machine_client.py
narrieta: done

nagworld9 (approved):

maddieford (approved):

PR #2844, opened 2023-06-14
Eliminate duplicate list of test suites to run

narrieta:
Currently new tests new to be added to pipeline.yml and runbook.yml. With this change, they need to be added to runbook.yml only.

The PR also fixes the paths for the tests in no_outbound_connections.yml, and adds agent_update to runbook.yml.

### DIFF
--- a/tests_e2e/pipeline/scripts/execute_tests.sh
+++ b/tests_e2e/pipeline/scripts/execute_tests.sh
@@ -39,6 +39,11 @@ docker pull waagenttests.azurecr.io/waagenttests:latest
 
 # Azure Pipelines does not allow an empty string as the value for a pipeline parameter; instead we use "-" to indicate
 # an empty value. Change "-" to "" for the variables that capture the parameter values.
+if [[ $TEST_SUITES == "-" ]]; then
+    TEST_SUITES=""  # Don't set the test_suites variable
+else
+    TEST_SUITES="-v test_suites:\"$TEST_SUITES\""
+fi
 if [[ $IMAGE == "-" ]]; then
     IMAGE=""
 fi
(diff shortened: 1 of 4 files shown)


tests_e2e/pipeline/scripts/execute_tests.sh
```diff
@@ -39,6 +39,11 @@ docker pull waagenttests.azurecr.io/waagenttests:latest
 
 # Azure Pipelines does not allow an empty string as the value for a pipeline parameter; instead we use "-" to indicate
 # an empty value. Change "-" to "" for the variables that capture the parameter values.
+if [[ $TEST_SUITES == "-" ]]; then
+    TEST_SUITES=""  # Don't set the test_suites variable
+else
+    TEST_SUITES="-v test_suites:\"$TEST_SUITES\""
```
maddieford, line 45: Are these values pulled from runbook? does this mean when we go to do a manual pipeline run, the default value will be "-" and we'll need to manually list each test suite we want to run?

maddieford (approved):

nagworld9 (approved):

tests_e2e/pipeline/scripts/execute_tests.sh
narrieta: The runbook has a variable with the default list of tests:

```
  # The test suites to execute
  - name: test_suites
    value: "agent_bvt, no_outbound_connections, extensions_disabled, agent_not_provisioned, agent_update, fips"

```

"-" is only for the pipeline (Azure Pipelines does not allow empty string for the default value of pipeline parameters).

Then the script (execute_tests.sh, which is used only by the pipeline) changes the value to empty string so the script won't override the runbook variable when invoking LISA.

 When starting a pipeline run manually, one can set the test suite to FOO (just as before), and in that case we execute_tests.sh will override the runbook variable (just as before).
 
 Runs from dev machines are not affected by this change.

PR #2846, opened 2023-06-15
Expose run name as a runbook variable

narrieta:
@maddieford @nagworld9 - Please setup this variable for developer runs. Resource groups and VMs will include this as part of their name.
`
    lisa  --runbook $runbook ...  -v name:FOO`

### DIFF
--- a/tests_e2e/orchestrator/runbook.yml
+++ b/tests_e2e/orchestrator/runbook.yml
@@ -1,4 +1,4 @@
-name: WALinuxAgent
+name: $(name)
 
 testcase:
   - criteria:
@@ -8,6 +8,8 @@ extension:
   - "./lib"
 
 variable:
+  - name: name
+    value: "WALinuxAgent"
   #
   # These variables define parameters handled by LISA.
   #


maddieford (approved):

nagworld9:
@narrieta This would break the cleanup pipeline if we provide custom names. We might as well expose it as new parameter for cleanup pipeline. So that people can run this pipeline with their names.

narrieta:
@nagworld9 it wouldn't break it, but the pipeline wouldn't clean these resources up. I've been deleting my runs from the command line, but adding the parameter to the cleanup pipeline is a good idea, I'll do that as well. Thanks for the suggestion.

narrieta:
@nagworld9 Actually, it is already exposed:

  - name: name_pattern
    displayName: Regular expression to match the name of the resource groups to delete
    type: string
    default: lisa-WALinuxAgent-.*

nagworld9:
> @nagworld9 Actually, it is already exposed:
> 
> * name: name_pattern
>   displayName: Regular expression to match the name of the resource groups to delete
>   type: string
>   default: lisa-WALinuxAgent-.*

Awesome!

nagworld9 (approved):

PR #2848, opened 2023-06-16
Collect test artifacts as a separate step in the test pipeline

narrieta:
Test execution times out from time to time; when this happens, Azure Pipelines terminates the task and artifacts are not collected. I separated the collection of test artifacts to its own separate step. Also, LISA logs may be needed to debug those timeouts; I added a pipeline parameter to collect all LISA logs and defaulted it to True. Once the investigation is completed, I'll change the default to False.

### DIFF
--- a/tests_e2e/pipeline/scripts/execute_tests.sh
+++ b/tests_e2e/pipeline/scripts/execute_tests.sh
@@ -26,7 +26,8 @@ chmod a+w "$BUILD_SOURCESDIRECTORY"
 #
 # Create the directory where the Docker container will create the test logs and give ownership to 'waagent'
 #
-LOGS_DIRECTORY="$HOME/logs"
+echo "##vso[task.setvariable variable=logs_directory]$HOME/logs"
+#LOGS_DIRECTORY="$HOME/logs"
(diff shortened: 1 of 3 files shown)


tests_e2e/pipeline/scripts/execute_tests.sh
```diff
@@ -54,10 +55,6 @@ if [[ $VM_SIZE == "-" ]]; then
     VM_SIZE=""
 fi
 
-# A test failure will cause automation to exit with an error code and we don't want this script to stop so we force the command
-# to succeed and capture the exit code to return it at the end of the script.
-echo "exit 0" > /tmp/exit.sh
```
nagworld9, line 59: So, the time outs not captured here in exit?

nagworld9 (approved):

tests_e2e/pipeline/scripts/execute_tests.sh
narrieta: azure pipelines kills the process after 60 min so the code to collect logs is not executed. the PR moves that code to a different step, which will be executed after this one is killed

PR #2850, opened 2023-06-19
Skip downgrade if requested version below daemon version

nagworld9:
## Description
Skip downgrade if requested version below daemon version.

Issue # 

---

### PR information


azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -291,6 +300,13 @@ def run(self, goal_state):
             if warn_msg != "":
                 self.__log_event(LogLevel.WARNING, warn_msg)
 
+            daemon_version = self.__get_daemon_version_for_update()
+            if requested_version < daemon_version:
+                # Don't process the update if the requested version is less than daemon version,
```
narrieta, line 305: let's explain why

narrieta, line 305: also, let's add a test case for this to agent_update

narrieta (approved):

azurelinuxagent/ga/agent_update_handler.py
narrieta: sorry, I should have been explicit: I meant add this scenario to your end-to-end tests (can be a separate PR). Having a unit test is also great, thanks for adding it.

narrieta (approved):
Approved with one comment on end-to-end tests

azurelinuxagent/ga/agent_update_handler.py
narrieta: never mind, it's there

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -305,7 +305,7 @@ def run(self, goal_state):
                     # as historically we don't support downgrades below daemon versions. So daemon will not pickup that requested version rather start with
                     # installed latest version again. When that happens agent go into loop of downloading the requested version, exiting and start again with same version.
                     #
-                    raise Exception("Can't process the update as the requested version: {0} is < current daemon version: {1}".format(
+                    raise Exception("as the requested version: {0} is < current daemon version: {1}".format(
```
narrieta, line 308: Missing "Unable to update Agent:"?

nagworld9: we append this msg when we catch the exception.

narrieta: That is confusing. Myabe instead define an AgentUpdateError exception with a "reason" property and initialize the reason to "Requested version... is < ...."? The format the message where you handle the exception.

narrieta: Or actually, you don't need the "reason" property. The type of the exception would be AgentUpdateError and the message  "Requested version... is < ...."

nagworld9: updated

narrieta (approved):

PR #2860, opened 2023-06-27
add time window for agent manifest download

nagworld9:
## Description

Added time window for agent manifest download in the case of agent needs to update with self-update path(largestVersion)
Issue # 

---

### PR information


tests_e2e/orchestrator/scripts/install-agent
```diff
@@ -115,7 +115,7 @@ echo "Agent's conf path: $waagent_conf_path"
 sed -i 's/AutoUpdate.Enabled=n/AutoUpdate.Enabled=y/g' "$waagent_conf_path"
 # By default GAUpdates flag set to True, so that agent go through update logic to look for new agents.
 # But in e2e tests this flag needs to be off in test version 9.9.9.9 to stop the agent updates, so that our scenarios run on 9.9.9.9.
-sed -i '$a GAUpdates.Enabled=n' "$waagent_conf_path"
+sed -i '$a Debug.EnableGAUpdates=n' "$waagent_conf_path"
```
nagworld9, line 118: Same update needs in agent update scenario. In this PR, I skip it since I rename the file where this needs to go in another PR.  So it will create a merge conflict and want to avoid that. I'll do it in sperate PR.

azurelinuxagent/common/conf.py
```diff
@@ -508,7 +508,7 @@ def get_ga_updates_enabled(conf=__conf__):
     If True, the agent go through update logic to look for new agents otherwise it will stop agent updates.
     NOTE: This option is needed in e2e tests to control agent updates.
     """
-    return conf.get_switch("GAUpdates.Enabled", True)
+    return conf.get_switch("Debug.EnableGAUpdates", True)
```
narrieta, line 511: The description in the comment needs improving. From reading it, it is hard to tell how it is different to AutoUpdate.Enabled

narrieta: The description for Debug.EnableGAVersioning also needs improving

azurelinuxagent/common/conf.py
```diff
@@ -506,7 +506,8 @@ def get_monitor_network_configuration_changes(conf=__conf__):
 def get_ga_updates_enabled(conf=__conf__):
     """
     If True, the agent go through update logic to look for new agents otherwise it will stop agent updates.
-    NOTE: This option is needed in e2e tests to control agent updates.
+    NOTE: This option is same like Autoupdates flag but one difference is that this will not influence the daemon to start the ext_handler of daemon(installed with image) version.
+    and it is used especially when you don't want to change the agent from a particular version
```
narrieta, line 510: why it doesn't influence the daemon?

sorry, this last part is still unclear: "and it is used especially when you don't want to change the agent from a particular version"

nagworld9: `why it doesn't influence the daemon?
`
Because the daemon won't check this flag when finding latest version to start it. 

`        latest_agent = None if not conf.get_autoupdate_enabled() else self.get_latest_agent_greater_than_daemon(
            daemon_version=CURRENT_VERSION)`

Let's say you installed 2.9 and you want to do some testing on this version and don't want to change this version even new version published. If you disable the autoupdate flag and restart the agent, then daemon launch the ext_handler of daemon version not the 2.9* version whereas this new flag helps you keep the same version 2.9 since it won't allow any upgrades and also daemon still finds 2.9 is latest version.

narrieta: from your description, seems like this doesn't have anything to do with daemon vs extension handler, but rather with run() vs run_latest(). is that the case?

nagworld9: It's the daemon who calls run_latest()

nagworld9: I mean to say it will influence the finding latest version.

narrieta: > It's the daemon who calls run_latest()

today, but that will change

if i understand this correctly, AutoUpdate.Enabled controls whether the Agent downloads and starts new updates **and** also whether any downloaded updates are started or not, while GAUpdatesEnabled controls only the former.

if yes, maybe you can explain in those terms (with a little more detail than my explanation)?

narrieta: And a table would be useful:

AutoUpdate.Enabled == false -> Agent preinstalled on the image will process extensions and will not update (regardless of GAUpdatesEnabled)

AutoUpdate.Enabled == true and  GAUpdatesEnabled == true, any update already downloaded will be started, and the agent will update

AutoUpdate.Enabled == true and  GAUpdatesEnabled == false, any update already downloaded will be started, and the agent will **not** update

All this verbosity needed because those flags are extremely confusing

narrieta: if the above is true, maybe rename GAUpdatesEnabled to something closer to DownloadAgentUpdates, too?

narrieta: InstallAgentUpdates?

nagworld9: It does both. We download when agent finds the update. When it finds the update, it will try to install by exit the process.

nagworld9: If I think loud, may be your right the GAUpdatesEnabled controls the download and while install the AutoUpdate.Enabled gets in the way and controls to start that downloaded agent or not.

nagworld9: In addition to your example if any of the flag is set to false, form that point agent stops downloading new versions if any available. 

In terms of install, your example is correct. 

In nutshell AutoUpdate.Enabled controls both download and install, GAUpdatesEnabled controls just download.

narrieta (approved):

narrieta (approved):

maddieford (approved):

PR #2862, opened 2023-06-29
ignore 75-persistent-net-generator.rules in e2e tests

nagworld9:
## Description

Issue # 

---

### PR information


tests_e2e/tests/lib/agent_log.py
```diff
@@ -92,11 +92,10 @@ def get_errors(self) -> List[AgentLogRecord]:
...
             #
-            # # This warning is expected on SUSE 12
-            # {
-            #     'message': r"WARNING EnvHandler ExtHandler Move rules file 75-persistent-net-generator.rules to /var/lib/waagent/75-persistent-net-generator.rules",
-            #     'if': lambda _: re.match(r"((sles15\.2)|suse12)\D*", DISTRO_NAME, flags=re.IGNORECASE) is not None
-            # },
+            # 2023-06-28T09:31:38.903835Z WARNING EnvHandler ExtHandler Move rules file 75-persistent-net-generator.rules to /var/lib/waagent/75-persistent-net-generator.rules
+            {
+                'message': r"WARNING EnvHandler ExtHandler Move rules file 75-persistent-net-generator.rules to /var/lib/waagent/75-persistent-net-generator.rules"
+            },
```
narrieta, line 98: Did you try this with an actual agent log including that record? "message" shouldn't include "WARNING EnvHandler ExtHandler" (there was a bug in the parser for old daemon versions that wasn't parsing the message, maybe the code in DCR was hitting that bug and that is why the level, prefix and thread are here)

nagworld9: Ok, we have similar message ignore in this file and doesn't include level prefix and thread. I'll update that to ignore 75

                'message': r"Move rules file 70-persistent-net.rules to /var/lib/waagent/70-persistent-net.rules",

narrieta: sounds good, please try those changes on a log from a failed test run... the format of the log changes with the daemon version and we should verify we are covering all the versions we test.

narrieta (approved):

maddieford (approved):

PR #2875, opened 2023-07-11
Agent status scenario

maddieford:
## Description
Migrating DCR agent_status scenario to e2e pipeline to validate VM agent status without goal state changes

Issue # 

---

### PR information


tests_e2e/tests/agent_status/agent_status.py
```diff
@@ -0,0 +1,143 @@
...
+                 .format(instance_view.vm_agent.vm_agent_version,
+                         instance_view.vm_agent.statuses[0].display_status))
+
+    return instance_view_is_valid
+
+
+class AgentStatus(AgentTest):
+    def __init__(self, context: AgentTestContext):
+        super().__init__(context)
+        self._ssh_client = SshClient(
```
nagworld9, line 111: We do have method in AgentTestContext, reusing that self._context.create_ssh_client() simplifies things.

tests_e2e/tests/agent_status/agent_status.py
```diff
@@ -0,0 +1,143 @@
...
+            instance_view = vm.get_instance_view()
+            log.info("")
+            log.info("Validating VM Instance View...")
+            log.info("Instance view of VM is:\n%s", instance_view.serialize())
+
+            instance_view_is_valid = validate_instance_view(instance_view)
+            if not instance_view_is_valid:
+                log.info("")
+                log.info("Instance view is not valid, waiting 10s before retry...")
+                sleep(10)
```
nagworld9, line 136: sleep 10 secs with timeout 5 mins is too aggressive retries. I don't think too frequent reties needed here as we decide to wait for 5 mins.

tests_e2e/test_suites/agent_status.yml
```diff
@@ -0,0 +1,9 @@
+#
+# This scenario validates the agent status is updated without any other goal state changes
```
narrieta, line 2: "any", rather than "any other"

tests_e2e/tests/agent_status/agent_status.py
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env python3
+import datetime
```
narrieta, line 2: let's move those imports below, with the rest of them

tests_e2e/tests/agent_status/agent_status.py
```diff
@@ -0,0 +1,143 @@
...
+from azure.mgmt.compute.models import VirtualMachineInstanceView
+
+from tests_e2e.tests.lib.agent_test import AgentTest
+from tests_e2e.tests.lib.agent_test_context import AgentTestContext
+from tests_e2e.tests.lib.logging import log
+from tests_e2e.tests.lib.ssh_client import SshClient
+from tests_e2e.tests.lib.virtual_machine_client import VirtualMachineClient
+
+
+def validate_instance_view_vmagent_status(instance_view: VirtualMachineInstanceView):
```
narrieta, line 36: why use global functions? shouldn't these be methods of the test class?

tests_e2e/tests/agent_status/agent_status.py
```diff
@@ -0,0 +1,143 @@
...
+
+        timeout = datetime.datetime.now() + datetime.timedelta(minutes=5)
+        instance_view_is_valid = False
+
+        # Retry validating instance view with timeout of 5 minutes
+        while datetime.datetime.now() < timeout and not instance_view_is_valid:
+            instance_view = vm.get_instance_view()
+            log.info("")
+            log.info("Validating VM Instance View...")
+            log.info("Instance view of VM is:\n%s", instance_view.serialize())
```
narrieta, line 130: can you pretty-print the serialization?

tests_e2e/tests/agent_status/agent_status.py
```diff
@@ -0,0 +1,143 @@
...
+
+            instance_view_is_valid = validate_instance_view(instance_view)
+            if not instance_view_is_valid:
+                log.info("")
+                log.info("Instance view is not valid, waiting 10s before retry...")
+                sleep(10)
+
+        log.info("")
+        assert_that(instance_view_is_valid).described_as("Timeout has expired, instance view is not valid.").is_true()
```
narrieta, line 139: the error message should include why the view is not valid

tests_e2e/tests/agent_status/agent_status.py
maddieford: pycharm suggested making them functions since they were static

maddieford: i think they should be methods of the test class though, I'll move them in the test class

tests_e2e/tests/agent_status/agent_status.py
```diff
@@ -0,0 +1,143 @@
...
+
+from tests_e2e.tests.lib.agent_test import AgentTest
+from tests_e2e.tests.lib.agent_test_context import AgentTestContext
+from tests_e2e.tests.lib.logging import log
+from tests_e2e.tests.lib.ssh_client import SshClient
+from tests_e2e.tests.lib.virtual_machine_client import VirtualMachineClient
+
+
+def validate_instance_view_vmagent_status(instance_view: VirtualMachineInstanceView):
+    is_valid = True
```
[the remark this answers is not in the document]
maddieford: updated

tests_e2e/tests/agent_status/agent_status.py
```diff
@@ -0,0 +1,120 @@
...
+        super().__init__(context)
+        self._ssh_client = self._context.create_ssh_client()
+
+    def validate_instance_view_vmagent_status(self, instance_view: VirtualMachineInstanceView):
+        status = instance_view.vm_agent.statuses[0]
+
+        # Validate message field
+        message = status.message
+        if message is None:
+            raise Exception("Instance view is missing an agent status message, waiting to retry...")
```
nagworld9, line 48: waiting to retry, bit misleading. Looks like this exception is logging after retries completed when we log timeout expired.

narrieta (changes requested):
Please review all the test messages. I believe the intention of this test to verify that the agent reports status without new goal states? The messages are not even closely related to that. Messages give the impression that we are checking a valid instance view without giving a clue as to what "valid" is.

tests_e2e/tests/agent_status/agent_status.py
```diff
@@ -0,0 +1,143 @@
...
+        # Retry validating instance view with timeout of 5 minutes
+        while datetime.datetime.now() <= timeout and not instance_view_is_valid:
+            instance_view = vm.get_instance_view()
+            log.info("")
+            log.info(
+                "Check instance view to validate that the Guest Agent reports status without any new goal states...")
+            log.info("Instance view of VM is:\n%s", json.dumps(instance_view.serialize(), indent=2))
+
+            try:
+                self.validate_instance_view(instance_view)
```
narrieta, line 127: How are we validating that the Agent is reporting status with no goal state? This seems to just be checking that the instance view has an Agent status with the expected format, but that status could have been reported as the result of a previous goal state.

maddieford: I updated to check that the agent status timestamp in the instance view updates & we check that the last processed goal state in the agent log is the same

tests_e2e/tests/agent_status/agent_status.py
```diff
@@ -0,0 +1,202 @@
...
+        Checks that instance view has vm_agent.statuses and vm_agent.vm_agent_version properties which report the Guest
+        Agent as running and Ready:
+
+        "vm_agent": {
+            "extension_handlers": [],
+            "vm_agent_version": "9.9.9.9",
+            "statuses": [
+                {
+                    "level": "Info",
+                    "time": "<class 'datetime.datetime'>",
```
narrieta, line 96: the value needs fixing

tests_e2e/tests/agent_status/agent_status.py
```diff
@@ -0,0 +1,202 @@
...
+                    "message": "Guest Agent is running",
+                    "code": "ProvisioningState/succeeded",
+                    "display_status": "Ready"
+                }
+            ]
+        }
+        """
+        if instance_view.vm_agent is None:
+            raise RetryableAgentStatusException("Instance view is missing vm agent")
```
narrieta, line 105: If you want to check that the property is present you should do '"vm_agent" in instance_view". The dot operator would raise an exception if the property is missing.

Same applies for the other checks on properties

tests_e2e/tests/agent_status/agent_status.py
narrieta: Thanks

tests_e2e/tests/agent_status/agent_status.py
```diff
@@ -0,0 +1,202 @@
...
+                prev_status_timestamp = status_timestamp
+                prev_gs_processed_log = gs_processed_log
+
+                # Sleep 30s to allow agent status to update before we check again
+                sleep(30)
+
+            except RetryableAgentStatusException as e:
+                instance_view_exception = str(e)
+                log.info("")
+                log.info("Agent status is invalid: {0}".format(instance_view_exception))
```
narrieta, line 186: the status is not invalid in this case:

`Agent status is invalid: Agent status failed to update: Current agent status timestamp 2023-08-07 15:02:27+00:00 is not greater than previous status timestamp 2023-08-07 15:02:27+00:00`

maddieford: Updated this to have more accurate error statement

tests_e2e/tests/agent_status/agent_status.py
maddieford: I updated this to have a more accurate error message "{property} in instance view is None" and also typed my instance_view variables.

I'm using the dot operator because the Model azure.mgmt.compute.models has classes for InstanceViewStatus and VirtualMachineAgentInstanceView. All of the properties I'm checking are attributes of these classes and initialized to None:
```
class InstanceViewStatus(msrest.serialization.Model):
    def __init__(
        self,
        *,
        code: Optional[str] = None,
        level: Optional[Union[str, "StatusLevelTypes"]] = None,
        **display_status: Optional[str] = None,
        message: Optional[str] = None,
        time: Optional[datetime.datetime] = None,**
        **kwargs
    ):
        super(InstanceViewStatus, self).__init__(**kwargs)
        self.code = code
        self.level = level
        self.display_status = display_status
        self.message = message
        self.time = time
```
```
class VirtualMachineAgentInstanceView(msrest.serialization.Model):
    def __init__(
        self,
        *,
        vm_agent_version: Optional[str] = None,
        extension_handlers: Optional[List["VirtualMachineExtensionHandlerInstanceView"]] = None,
        statuses: Optional[List["InstanceViewStatus"]] = None,
        **kwargs
    ):
        super(VirtualMachineAgentInstanceView, self).__init__(**kwargs)
        self.vm_agent_version = vm_agent_version
        self.extension_handlers = extension_handlers
        self.statuses = statuses
```

tests_e2e/tests/agent_status/agent_status.py
```diff
@@ -0,0 +1,202 @@
...
+                }
+            ]
+        }
+        """
+        if instance_view.vm_agent is None:
+            raise RetryableAgentStatusException("Instance view is missing vm agent")
+        else:
+            self.validate_instance_view_vmagent(instance_view=instance_view)
+
+        if instance_view.statuses is None:
```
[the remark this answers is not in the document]
maddieford: Removed and combined with validate_instance_view_vmagent

tests_e2e/tests/agent_status/agent_status.py
```diff
@@ -0,0 +1,192 @@
...
+                    "time": "2023-08-11T09:13:01.000Z",
+                    "message": "Guest Agent is running",
+                    "code": "ProvisioningState/succeeded",
+                    "display_status": "Ready"
+                }
+            ]
+        }
+        """
+        if instance_view.vm_agent is None:
+            raise RetryableAgentStatusException("Agent status is invalid: 'vm_agent' property in instance view is None")
```
narrieta, line 85: do you think these properties will ever be None? I think more likely they would be missing, as your previous code was trying to check

tests_e2e/tests/agent_status/agent_status.py
narrieta: cool. could you add that as a comment?

narrieta (approved):

nagworld9 (approved):

PR #2886, opened 2023-07-26
Install assertpy on test VMs

narrieta:
Test VMs need module assertpy. Now we prepare a Pypy tarball with this module preinstalled, keep it in storage, and install Pypy from there (see wiki for details)

### DIFF
--- a/tests_e2e/orchestrator/scripts/prepare-pypy
+++ b/tests_e2e/orchestrator/scripts/prepare-pypy
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+
+# Microsoft Azure Linux Agent
+#
+# Copyright 2018 Microsoft Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#
+# This script is used to prepare a tarball containing Pypy with the assert-py module pre-installed.
+# It needs to be run on x64 and arm64 VMs and the resulting tarballs need to be uploaded to storage,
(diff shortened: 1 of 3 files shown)


maddieford (approved):

tests_e2e/orchestrator/scripts/prepare-pypy
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
```
nagworld9, line 1: So, this is a standalone tool, someone needs to run to update the image?

tests_e2e/orchestrator/scripts/prepare-pypy
```diff
@@ -0,0 +1,56 @@
...
+
+printf "\n*** Installing assertpy...\n"
+tar xf $tarball
+./pypy3.7-v7.3.5-*/bin/pypy -m ensurepip
+./pypy3.7-v7.3.5-*/bin/pypy -mpip install assertpy
+
+printf "\n*** Creating new tarball for Pypy...\n"
+# remove the cache files created when Pypy, and set the owner to 0/0, in order to match the original tarball
+find pypy3.7-v7.3.5-* -name '*.pyc' -exec rm {} \;
+mv -v $tarball "$tarball.original"
```
nagworld9, line 51: what is .original means? does it add extension name to the zip name? or its just way of defining for something?

narrieta: It's the original tarball that we downloaded on line 37/40. I keep the original in case it is needed for debugging

tests_e2e/orchestrator/scripts/prepare-pypy
narrieta: it's described in the wiki

tests_e2e/orchestrator/scripts/prepare-pypy
nagworld9: If that's the case, to keep original this needs to move up before we install other modules? is itn't?

nagworld9: Never mind, we are not modifying the tar at this point, its effecting the extraction folder I guess. I think we are good

nagworld9 (approved):

tests_e2e/orchestrator/scripts/prepare-pypy
narrieta: yup, the modules are in the new tarball; original remains unchanged

PR #2903, opened 2023-08-23
Fix agent memory usage check

nagworld9:
## Description
Added fix for addressing agent reverted to old version due to multiple restarts too frequently because of agent exits with memory limit reach.

We revert to old version if the agent restarts 3 times in CHILD_LAUNCH_INTERVAL. That's why we delay first attempt memory usage check, so that it won't meet this condition on restart.

Issue # 

---

### PR information


azurelinuxagent/ga/update.py
```diff
@@ -1016,8 +1017,11 @@ def _check_agent_memory_usage(self):
         """
         try:
             if conf.get_enable_agent_memory_usage_check() and self._extensions_summary.converged:
-                if self._last_check_memory_usage == datetime.min or datetime.utcnow() >= (self._last_check_memory_usage + UpdateHandler.CHECK_MEMORY_USAGE_PERIOD):
-                    self._last_check_memory_usage = datetime.utcnow()
+                # we delay first attempt memory usage check, so that current agent won't get blacklisted due to multiple restarts(because of memory limit reach) too frequently
+                if (self._initial_attempt_check_memory_usage and time.time() - self._last_check_memory_usage_time > CHILD_LAUNCH_INTERVAL) or \
```
narrieta, line 1021: why the change from utcnow() to time()? - the former has the advantage that it formats nicely when, for example, running the code under the debugger

narrieta (approved):

azurelinuxagent/ga/update.py
nagworld9: Nothing specific, I just came across the idea. Let me update if it has advantage

narrieta: ok, it is not a big issue.

maddieford (approved):

PR #2907, opened 2023-08-30
Download certs on FT GS after check_certificates only when missing from disk

maddieford:
## Description
We deployed a change with 2.9.1.1 which updated the agent to always re download certs on every fast track goal state (in case a cert was removed from disk) -> see change here https://github.com/Azure/WALinuxAgent/pull/2761

This introduced an issue for the hibernate VM case. When a VM resumes from hibernate state, it is moved to a new container. The agent processes a fasttrack goal state on hibernate resume, but the certs_uri references the VMs old containerId, so wireserver returns a 410 when we try to download certs with the certs_uri. We don't handle this 410, so the GS is not failed or retried gracefully.

This PR updates this behavior to: if FastTrack GS, check that certificates required by GS are in certs summary. If certs_uri is not None, check that all certs in certs summary are present on the disk -> if not, try to re-download certs, if any exceptions are raised do nothing except log and send to telemetry 

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/common/protocol/goal_state.py
+++ b/azurelinuxagent/common/protocol/goal_state.py
@@ -208,7 +208,7 @@ class GoalState(object):
         try:
             self._update(force_update=False)
         except GoalStateInconsistentError as e:
-            message = "Detected an inconsistency in the goal state: {0}", ustr(e)
+            message = "Detected an inconsistency in the goal state: {0}".format(ustr(e))
             self.logger.warn(message)
             add_event(op=WALAEventOperation.GoalState, is_success=False, message=message)
 
--- a/azurelinuxagent/common/protocol/goal_state.py
+++ b/azurelinuxagent/common/protocol/goal_state.py
@@ -292,11 +292,9 @@ class GoalState(object):
         #
         if self._extensions_goal_state.source == GoalStateSource.FastTrack and self._goal_state_properties & GoalStateProperties.Certificates:
             self._check_certificates()
+            self._check_and_download_missing_certs_on_disk()
 
     def _check_certificates(self):
-        # Re-download certificates in case they have been removed from disk since last download
-        if self._certs_uri is not None:
-            self._download_certificates(self._certs_uri)
         # Check that certificates needed by extensions are in goal state certs.summary
         for extension in self.extensions_goal_state.extensions:
             for settings in extension.settings:
@@ -320,6 +318,32 @@ class GoalState(object):
             add_event(op=WALAEventOperation.GoalState, message=certs.warnings)
         self._history.save_certificates(json.dumps(certs.summary))
         return certs
+
+    def _check_and_download_missing_certs_on_disk(self):
+        # Re-download certificates if any have been removed from disk since last download
+        if self._certs_uri is not None:
+            certificates = self.certs.summary
+            certs_missing_from_disk = False
+
+            for c in certificates:
+                cert_path = os.path.join(conf.get_lib_dir(), c['thumbprint'] + '.crt')
+                if not os.path.isfile(cert_path):
+                    certs_missing_from_disk = True
+                    message = "Certificate required by goal state is not on disk: {0}".format(cert_path)
+                    self.logger.info(message)
+                    add_event(op=WALAEventOperation.GoalState, message=message)
+            if certs_missing_from_disk:
+                # Try to re-download certs. Sometimes download may fail if certs_uri is outdated/contains wrong
+                # container id (for example, when the VM is moved to a new container after resuming from
+                # hibernation). If download fails we should report and continue with goal state processing, as some
+                # extensions in the goal state may succeed.
+                try:
+                    self._download_certificates(self._certs_uri)
+                except Exception as e:
+                    message = "Unable to download certificates. Goal state processing will continue, some " \
+                              "extensions requiring certificates may fail. Error: {0}".format(ustr(e))
+                    self.logger.warn(message)
+                    add_event(op=WALAEventOperation.GoalState, is_success=False, message=message)
 
     def _restore_wire_server_goal_state(self, incarnation, xml_text, xml_doc, vm_settings_support_stopped_error):
         msg = 'The HGAP stopped supporting vmSettings; will fetched the goal state from the WireServer.'
--- a/azurelinuxagent/common/protocol/goal_state.py
+++ b/azurelinuxagent/common/protocol/goal_state.py
@@ -208,9 +208,15 @@ class GoalState(object):
         try:
             self._update(force_update=False)
         except GoalStateInconsistentError as e:
-            self.logger.warn("Detected an inconsistency in the goal state: {0}", ustr(e))
+            message = "Detected an inconsistency in the goal state: {0}", ustr(e)
+            self.logger.warn(message)
+            add_event(op=WALAEventOperation.GoalState, is_success=False, message=message)
+
             self._update(force_update=True)
-            self.logger.info("The goal state is consistent")
+
+            message = "The goal state is consistent"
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -292,11 +292,31 @@ def _update(self, force_update):
         #
         if self._extensions_goal_state.source == GoalStateSource.FastTrack and self._goal_state_properties & GoalStateProperties.Certificates:
             self._check_certificates()
+            # Re-download certificates if any have been removed from disk since last download
+            if self._certs_uri is not None:
+                certificates = self.certs.summary
+                certs_missing_from_disk = False
+
+                for c in certificates:
```
maddieford, line 300: This will log/send telemetry for every certificate missing from disk

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -292,11 +292,31 @@ def _update(self, force_update):
...
+                if certs_missing_from_disk:
+                    # Try to re-download certs. Sometimes download may fail if certs_uri is outdated/contains wrong
+                    # container id (for example, when the VM is moved to a new container after resuming from
+                    # hibernation). If download fails we should report and continue with goal state processing, as some
+                    # extensions in the goal state may succeed.
+                    try:
+                        self._download_certificates(self._certs_uri)
+                    except Exception as e:
+                        message = "Failed to download certificates: {0}".format(ustr(e))
+                        self.logger.warn(message)
```
maddieford, line 316: Warning, error, or info? Currently have it as a warning since missing cert may cause extension processing to fail, but not as an error because we will continue to process the goal state

tests/common/protocol/test_goal_state.py
maddieford, line 1: These unit tests were originally modified in https://github.com/Azure/WALinuxAgent/pull/2761/files#diff-191372e2d3c3eb5595dd179cb26eb8e2d59a432c68c5e6280ab35817a1fae471L136

Im changing them back since we won't download certs on every FastTrack goal state, only when certs are missing from disk

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -292,11 +292,31 @@ def _update(self, force_update):
...
+                        add_event(op=WALAEventOperation.GoalState, message=message)
+                if certs_missing_from_disk:
+                    # Try to re-download certs. Sometimes download may fail if certs_uri is outdated/contains wrong
+                    # container id (for example, when the VM is moved to a new container after resuming from
+                    # hibernation). If download fails we should report and continue with goal state processing, as some
+                    # extensions in the goal state may succeed.
+                    try:
+                        self._download_certificates(self._certs_uri)
+                    except Exception as e:
+                        message = "Failed to download certificates: {0}".format(ustr(e))
```
narrieta, line 315: since this is a warning, let's add more context, similar to the comment above: "unable to download certificates, will continue, some extensions may fail, error: blah"

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -292,11 +292,31 @@ def _update(self, force_update):
...
+                        self.logger.warn(message)
+                        add_event(op=WALAEventOperation.GoalState, message=message)
```
narrieta, line 317: the operation success should be false

tests/common/protocol/test_goal_state.py
maddieford, line 1: There are already unit tests written which test certs will be re-downloaded on FT GS after being deleted: https://github.com/Azure/WALinuxAgent/blob/ae01d59f2d43314f4c8cb2d8120a3c5892086de6/tests/common/protocol/test_goal_state.py#L388

tests/common/protocol/test_goal_state.py
```diff
@@ -464,7 +464,7 @@ def http_get_handler(url, *_, **__):
             goal_state = GoalState(protocol.client)
 
             self.assertEqual(2, protocol.mock_wire_data.call_counts['goalstate'], "There should have been exactly 2 requests for the goal state (original + refresh)")
-            self.assertEqual(4, http_get_handler.certificate_requests, "There should have been exactly 4 requests for the goal state certificates (2x original + 2x refresh)")
+            self.assertEqual(2, http_get_handler.certificate_requests, "There should have been exactly 2 requests for the goal state certificates (original + refresh)")
```
narrieta, line 467: the new code is independent of the inconsistent goal state. it should have its own test

azurelinuxagent/common/protocol/goal_state.py
narrieta: warning is correct

azurelinuxagent/common/protocol/goal_state.py
narrieta: i mean set success=False in the call to add_event

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -292,11 +292,31 @@ def _update(self, force_update):
         #
         if self._extensions_goal_state.source == GoalStateSource.FastTrack and self._goal_state_properties & GoalStateProperties.Certificates:
             self._check_certificates()
+            # Re-download certificates if any have been removed from disk since last download
```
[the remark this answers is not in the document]
maddieford: Moved to its own method

azurelinuxagent/common/protocol/goal_state.py
maddieford: Updated

azurelinuxagent/common/protocol/goal_state.py
maddieford: Updated the message with more context

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -292,11 +292,31 @@ def _update(self, force_update):
...
+                        message = "Certificate required by goal state is not on disk: (0)".format(cert_path)
+                        self.logger.info(message)
+                        add_event(op=WALAEventOperation.GoalState, message=message)
+                if certs_missing_from_disk:
+                    # Try to re-download certs. Sometimes download may fail if certs_uri is outdated/contains wrong
+                    # container id (for example, when the VM is moved to a new container after resuming from
+                    # hibernation). If download fails we should report and continue with goal state processing, as some
+                    # extensions in the goal state may succeed.
+                    try:
+                        self._download_certificates(self._certs_uri)
```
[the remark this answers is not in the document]
maddieford: this PR is trying to handle that case. When we resume from hibernation we don't force an update of the goal state/refresh the cert_uri, so after we added logic to download certs on every fasttrack GS, the download failed because cert_uri was outdated. When download failed we didn't handle the exception, this PR handles the exception in this case by reporting the failure and continuing. We also are trying to limit the # of times we hit this issue by only downloading certs when they're missing from the disk.

Cert_uri is only refreshed if we force an update or fetch new GS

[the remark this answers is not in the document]
narrieta: The goal state is refreshed by the code handling the InconsistentGoalState exception. The change to force download on every FT goal state bypassed that logic. This changes restores that path by downloading the certs only after the check for consistency has succeeded (plus includes some other improvements, like better error handling, logging and telemetry)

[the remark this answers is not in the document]
nagworld9: Thanks for explaining.

narrieta (dismissed):

nagworld9 (approved):

narrieta (approved):

narrieta (approved):

PR #2915, opened 2023-09-06
fix the ignore rule in agent update test

nagworld9:
## Description
This PR also includes change for initializing test case at once instead multiple times. 
Issue # 

---

### PR information

### DIFF
--- a/tests_e2e/orchestrator/lib/agent_test_suite.py
+++ b/tests_e2e/orchestrator/lib/agent_test_suite.py
@@ -462,19 +462,10 @@ class AgentTestSuite(LisaTestSuite):
                                 test_suite_success = False
                                 raise
 
-                        # initialize and save test classes at once to avoid multiple initialization and also keep the context of the test class around
-                        load_tests = {}
-                        for suite in self.context.test_suites:  # pylint: disable=E1133
-                            for test in suite.tests:
-                                if test.name not in load_tests:
-                                    load_tests[test.name] = test.test_class(self.context)
-
                         for suite in self.context.test_suites:  # pylint: disable=E1133
                             log.info("Executing test suite %s", suite.name)
                             self.context.lisa_log.info("Executing Test Suite %s", suite.name)
-                            test_suite_success = self._execute_test_suite(suite, load_tests) and test_suite_success
-
-                        test_suite_success = self._check_agent_log(load_tests) and test_suite_success
+                            test_suite_success = self._execute_test_suite(suite) and test_suite_success
 
                     finally:
                         collect = self.context.collect_logs
@@ -500,7 +491,7 @@ class AgentTestSuite(LisaTestSuite):
                     if not success:
                         self._mark_log_as_failed()
 
-    def _execute_test_suite(self, suite: TestSuiteInfo, load_tests: Dict[str, Any]) -> bool:
+    def _execute_test_suite(self, suite: TestSuiteInfo) -> bool:
         """
         Executes the given test suite and returns True if all the tests in the suite succeeded.
         """
@@ -519,6 +510,7 @@ class AgentTestSuite(LisaTestSuite):
                     log.info("")
 
                     summary: List[str] = []
+                    ignore_error_rules: List[Dict[str, Any]] = []
 
                     for test in suite.tests:
                         test_full_name = f"{suite_name}-{test.name}"
@@ -529,8 +521,9 @@ class AgentTestSuite(LisaTestSuite):
 
                         test_success: bool = True
 
+                        load_test = test.test_class(self.context)
                         try:
-                            load_tests[test.name].run()
+                            load_test.run()
                             summary.append(f"[Passed]  {test.name}")
                             log.info("******** [Passed] %s", test.name)
                             self.context.lisa_log.info("[Passed] %s", test_full_name)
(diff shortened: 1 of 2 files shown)


tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -462,12 +462,19 @@ def _execute(self, environment: Environment, variables: Dict[str, Any]):
                                 test_suite_success = False
                                 raise
 
+                        # initialize and save test classes at once to avoid multiple initialization and also keep the context of the test class around
```
narrieta, line 465: The code would be simpler and clearer if you move the call to self._check_agent_log() into execute_test_suite().

As test are executed, we can collect the ignore rules into an array and then just pass them as argument to check_agent_log()

tests_e2e/tests/agent_update/rsm_update.py
```diff
@@ -49,16 +50,17 @@ def __init__(self, context: AgentTestContext):
...
         ignore_rules = [
             #
             # This is expected as we validate the downgrade scenario
             #
             # WARNING ExtHandler ExtHandler Agent WALinuxAgent-9.9.9.9 is permanently blacklisted
-            #
+            # Note: Version varies depending on the pipeline branch the test is running on
             {
-                'message': r"Agent WALinuxAgent-9.9.9.9 is permanently blacklisted"
+                'message': rf"Agent WALinuxAgent-{self._installed_agent_version} is permanently blacklisted"
```
narrieta, line 63: why using the version of the installed agent? shouldn't it be the version that your test uses for the downgrade?

nagworld9: no, we blocklist the current version which is running not the downloaded version

tests_e2e/orchestrator/lib/agent_test_suite.py
nagworld9: let me work on it

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -587,6 +586,8 @@ def _execute_test_suite(self, suite: TestSuiteInfo) -> bool:
                             log.warning("%s failed and blocks the suite. Stopping suite execution.", test.name)
                             break
 
+                        ignore_error_rules.extend(load_test.get_ignore_error_rules())
```
narrieta, line 589: i think we need to add the rules even if we hit the break at 587

tests_e2e/tests/agent_update/rsm_update.py
```diff
@@ -241,6 +247,17 @@ def _verify_agent_reported_update_status(self, version: str):
...
 
+    def _retrieve_installed_agent_version(self):
+        """
+        Retrieve the installed agent version
+        """
+        log.info("Retrieving installed agent version")
+        stdout: str = self._ssh_client.run_command("waagent-version", use_sudo=True)
+        log.info("Retrieved installed agent version \n {0}".format(stdout))
+        match = re.search(r'.*Goal state agent: (\S*)', stdout)
+        if match:
```
narrieta, line 258: no match should be an error instead of defaulting to 9.9.9.9, shouldn't it

tests_e2e/orchestrator/lib/agent_test_suite.py
nagworld9: you mean for all remaining for tests? or every test regardless? I don't see a point if test run failed.

tests_e2e/tests/agent_update/rsm_update.py
nagworld9: Even we raise an error, we still check ignore rules.  Need default value either case but that won't change behavior and makes if condition false, so rule won't apply.

But I'll raise error

narrieta: or at least a warning, rather than silently default to 9.9.9.9

tests_e2e/orchestrator/lib/agent_test_suite.py
narrieta: you are already adding the rule if the test failed.

the break is hit when the test failed **and** it blocks the suite

nagworld9: you meant for that particular test? Don't care about reaming tests after hit break

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -522,9 +521,9 @@ def _execute_test_suite(self, suite: TestSuiteInfo) -> bool:
 
                         test_success: bool = True
 
+                        load_test = test.test_class(self.context)
```
narrieta, line 524: btw, could we rename this variable "load_test" doesn't have a lot of meaning, I think. Maybe "test_instance"?

tests_e2e/orchestrator/lib/agent_test_suite.py
narrieta: for the test, right? that is what you are trying to do. ignore rules are associates to specific tests

nagworld9: when we block suite run, do we need to add rules for rest of the tests in suite?Lets' say suite has 1,2,3 test and 2 has failed and has flag to block suite. So ignore rule list should contain 1 and 2 but not 3 or all 1,2,3?

narrieta (approved):
thanks for the updates

tests_e2e/orchestrator/lib/agent_test_suite.py
narrieta: it should not contain the rules for the tests that did not execute

nagworld9: That's what I thought, thanks for the clarification

maddieford (approved):

PR #2916, opened 2023-09-07
Use Mariner 2 in FIPS test

narrieta:
The FIPS test does not run end-to-end with RHEL 9; switching to Mariner 2.

It looks like the openssl settings need to be tweaked for RHEL 9 to work. I'll follow that up on separate change.

### DIFF
--- a/tests_e2e/tests/scripts/fips-check_fips_mariner
+++ b/tests_e2e/tests/scripts/fips-check_fips_mariner
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+
+# Microsoft Azure Linux Agent
+#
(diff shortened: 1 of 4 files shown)


tests_e2e/tests/scripts/fips-check_fips_mariner
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
```
narrieta, line 1: The steps in these 2 scripts come from the internal documentation for Mariner

maddieford (approved):

nagworld9 (approved):

PR #2933, opened 2023-09-27
Extension sequencing scenario

maddieford:
## Description
This PR adds the extension sequencing scenario which requires VMSS. It includes changes to the combinator and agent test suite to support running tests on scale set instances.

Issue # 

---

### PR information


.gitignore
```diff
@@ -91,3 +91,6 @@ ENV/
 # pyenv
 .python-version
 .vscode/
+
+# resource group txt created for ext_sequencing scenario in e2e pipeline
+resource_groups_to_delete.txt
```
maddieford, line 96: The combinator creates a resource group and VMSS for the extension sequencing scenario and writes the resource group name to this file. Later, a task in the pipeline runs to cleanup the resources listed in this file.

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -275,138 +278,147 @@ def _clean_up(self) -> None:
         Cleans up any leftovers from the test suite run. Currently just an empty placeholder for future use.
         """
 
-    def _setup_node(self, install_test_agent: bool) -> None:
+    def _setup_node(self, install_test_agent: bool, nodes: List[Dict[str, Any]]) -> None:
```
maddieford, line 281: _setup_node, _collect_node_logs, and _check_agent_log were all updated to take in a list of nodes which indicate which nodes the setup/log actions should be taken.

In the case of extension sequencing scenario, the nodes list represents all the instances of the VMSS. Setup and log collection/checking will happen on each instance. In the case of any other extension, the nodes list will only contain one VM.

tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
maddieford, line 1: The combinator was updated to create a VMSS for the extension sequencing scenario. We choose one instance of the VMSS to run the extension sequencing scenario on. We report a test failure if VMSS creation fails

tests_e2e/orchestrator/lib/templates/ext_seq_vmss_template.json
maddieford, line 1: Template to create virtual machine scale set

tests_e2e/orchestrator/lib/templates/ext_seq_vmss_template.json
```diff
@@ -0,0 +1,321 @@
...
+            "sku": {
+                "name": "Standard_D2s_v3",
+                "tier": "Standard",
+                "capacity": 3
+            },
+            "properties": {
+                "orchestrationMode": "Flexible",
+                "virtualMachineProfile": {
+                    "extensionProfile": {
+                        "extensions": [
```
maddieford, line 212: This template includes extensions and dependencies which will be added to the scale set at creation.

This is different from the current DCR scenario. The current DCR scenario creates the VMSS, and then repeatedly updates the extensions/dependencies on the scale set in the test to validate the order in which the extensions were enabled. This can be done because the DCR scenario creates a VMSS with 'Uniform' orchestration mode. With Uniform orchestration mode you can update the scale set model and each instance in the scale set will be updated with the new model (including extensions). 

We cannot use 'Uniform' orchestration mode for the VMSS in this scenario. Lisa requires that we use 'Flexible' orchestration mode, because while setting up the test environment, LISA looks for the VM in the resource group. In flexible orchestration mode, a VM for each instance is created in the resource group, but for uniform mode there are no VM resources created for the instances. 

As a result, this test will look a little different from DCR. We must add the extensions/dependencies at VMSS creation so that each instance will get the extensions/dependencies. We can only validate once, because we cannot update the VM instances after VMSS creation. 

Pavan mentioned Flexible VMSS will support UpgradePolicy in the next 1-2 semesters. So I added a TODO to improve this scenario when that happens

tests_e2e/pipeline/pipeline.yml
maddieford, line 1: Added job in pipeline to cleanup resources created by the combinator

tests_e2e/tests/lib/identifiers.py
maddieford, line 1: Added clients to do operations on ResourceGroups and VirtualMachineScaleSetClients in azure python sdk

tests_e2e/tests/ext_sequencing/ext_sequencing.py
```diff
@@ -0,0 +1,211 @@
...
+        ext_to_remove = None
+        for ext in dependency_map:
+            if dependency_map[ext]:
+                ext_to_remove = random.choice(dependency_map[ext])
+                break
+        if ext_to_remove:
+            try:
+                vmss.delete_extension(ext_to_remove)
+            except HttpResponseError as e:
+                if "(BadRequest)" in e.message and "provisionAfterExtensions" in e.message:
```
maddieford, line 74: When you try to remove an extension with dependencies CRP reports an error which we catch here
```
{
  "innererror": {
    "internalErrorCode": "MissingDependentExtensionInArgumentDependentExtensionExistsInVMSSInstance"
  },
  "code": "BadRequest",
  "message": "On resource 'lisawalinuxagentextseq20230923164508e0', extension 'RunCommand' specifies 'VmAccess' in its provisionAfterExtensions property, but the extension 'VmAccess' will no longer exist. First, remove the extension 'RunCommand' or remove 'VmAccess' from the provisionAfterExtensions property of 'RunCommand'."
}
```

tests_e2e/test_suites/ext_sequencing.yml
```diff
@@ -0,0 +1,10 @@
+#
+# Adds extensions with multiple dependencies to VMSS using 'provisionAfterExtensions' property and validates they are
+# enabled in order of dependencies.
+# NOTE: This scenario is executed on an instance of a scaleset created by the combinator.
+#
+name: "ExtSequencing"
+tests:
+  - "ext_sequencing/ext_sequencing.py"
+images: "random(endorsed)"
```
maddieford, line 9: Original DCR scenario runs on all endorsed images. 

I'm choosing to run the scenario on one random endorsed image because VMSS creation is time consuming and test execution cannot start until the combinator is done creating the VMSS resources and environments.

tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
```diff
@@ -192,13 +212,40 @@ def create_environment(c_env_name: str) -> Dict[str, Any]:
...
                         "c_shared_resource_group_location": self._SHARED_RESOURCE_GROUP_LOCATIONS[self.runbook.cloud],
-                        "c_vm_tags": c_vm_tags
+                        "c_vm_tags": c_vm_tags,
+                        "c_deploy": True
                     }
+                
+                def create_environment_ext_sequencing(c_env_name: str) -> Dict[str, Any]:
+                    log.info("Creating VMSS for ExtSequencing scenario")
+                    curr_datetime = datetime.datetime.now().strftime("%Y%m%d-%H%M%S")
+                    rg_name = f"lisa-WALinuxAgent-extseq-{curr_datetime}-e{self._created_rg_count}"
```
narrieta, line 222: we should use the 'name' variable from the runbook instead of hardcoding "WALinuxAgent"

tests_e2e/pipeline/scripts/delete_resource_groups.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
narrieta, line 1: I think we will need to move the cleanup code to somewhere in LISA, otherwise developer runs are not cleaned up

maddieford:
VMSS support was added in this PR: https://github.com/Azure/WALinuxAgent/pull/2954

Extension sequencing scenario will be added in a different PR

PR #2936, opened 2023-09-29
Add support for EC certificates

### LINKED ISSUE #1550: walinuxagent fails to start when host is using EC keys [BUG]

narrieta:
Fixes #1550


azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -579,26 +579,19 @@ def __init__(self, xml_text, my_logger):
 
         # The parsing process use public key to match prv and crt.
         buf = []
-        begin_crt = False  # pylint: disable=W0612
```
narrieta, line 582: pylint W0612: Unused Variable

tests_e2e/tests/keyvault_certificates/keyvault_certificates.py
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+        thumbprints = test_certificates.keys()
+        certificate_urls = [u[self._context.vm.cloud] for u in test_certificates.values()]
+
+        # The test certificates should be downloaded to these locations
+        expected_certificates = " ".join([f"/var/lib/waagent/{t}.{{crt,prv}}" for t in thumbprints])
+
+        # The test may be running on a VM that has already been tested (e.g. while debugging the test), so we need to delete any existing test certificates first
+        # (note that rm -f does not fail if the give files do not exist)
```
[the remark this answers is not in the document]
narrieta: thanks; fixed

azurelinuxagent/common/utils/cryptutil.py
```diff
@@ -53,10 +53,21 @@ def gen_transport_cert(self, prv_file, crt_file):
...
+        except shellutil.CommandError as error:
+            if not ("Not an RSA key" in error.stderr or "expecting an rsa key" in error.stderr):
+                logger.error(
+                    "Command: [{0}], return code: [{1}], stdout: [{2}] stderr: [{3}]",
+                    " ".join(command),
+                    error.returncode,
+                    error.stdout,
+                    error.stderr)
+                raise
+            return shellutil.run_command([self.openssl_cmd, "pkey", "-in", file_name, "-pubout"], log_error=True)
```
maddieford, line 70: If pkey is may not be available on older versions of openssl, should we handle any errors here?

narrieta: how should we handle them? we have no way to extract the certs so at this point i'm just letting the goal state fail by letting the exception bubble up

maddieford: If we catch and log the error, could some extensions in the GS which don't require the cert still provision successfully? And then only the extensions which require the cert fail

narrieta: We don't know what extensions require those certs. The tenant cert and ssh keys for provisioning always use RSA (at least currently). EC would be used on KeyVault certs, which are typically used with CSE or something like that, and the scripts may not be resilient to a failure not finding the cert (and may have already done partial work). At this point I would fail the goal state rather than swallowing the error. Once we start getting telemetry about pkey we can revisit this.

maddieford: Makes sense!

nagworld9 (approved):

Issue #2929, opened 2023-09-20
Running collect-logs


CalvoM:
Hello,
I am trying to run the collect-logs command and it seems to fail however I see that the daemon runs it by prefixing the command with some `systemd-run` configuration.  Is there a way of running the `collect-log` command from the cli without prefixing it with the `systemd-run` config?

I am testing v2.9.1.1 on ubuntu versions>= 18.04.

maddieford:
@CalvoM Thanks for pointing this out. We added resource monitoring on collect logs in 2.9.1.1+, which broke the command line option. We'll work on fixing that. 

In the meantime, if you have 2.8.0.11 on the machine, you can use the following command to collect logs from cli:
`python3 -u /var/lib/waagent/WALinuxAgent-2.8.0.11/bin/WALinuxAgent-2.8.0.11-py2.7.egg -collect-logs`

maddieford:
Fix is merged

PR #2939, opened 2023-10-06
skip cgroup monitoring if log collector doesn't start by the agent.

nagworld9:
## Description
Fixing https://github.com/Azure/WALinuxAgent/issues/2929

We added cgroup monitoring to track resource usage for log collector which is breaking for users who starts from command line. 

Solution: 
Instead of any new command line argument, taking the advantage of existing flag where we already skip cgroup validation if log collector does not start by the agent. Using same flag for not monitoring cgroups too. This way it's pretty transparent to the users and no need to learn about new command line option.

Issue # 

---

### PR information


azurelinuxagent/agent.py
```diff
@@ -204,11 +205,10 @@ def collect_logs(self, is_full_mode):
             logger.info("Running log collector mode normal")
 
         # Check the cgroups unit
-        cpu_cgroup_path, memory_cgroup_path, log_collector_monitor = None, None, None
-        if CollectLogsHandler.should_validate_cgroups():
```
narrieta, line 208: After removing this call, I think CollectLogsHandler.should_validate_cgroups() is not used anymore

nagworld9: No, I renamed it and still used at Line 211

maddieford (approved):

narrieta (approved):

Issue #2940, opened 2023-10-09
[BUG] CPU credits drain on B1ls after update to 2.10.0.3


gregolsky:
1. Log - after update we see CPU usage increase by 2%, which then causes CPU credits drain on small burstable instances e.g. B1ls where 5% is the baseline. It happens on multiple instances since September 27th - day of the update.

```
Sep 27 10:52:17 vmXXXXXXXXXX python3[4788]: 2023-09-27T10:52:17.140495Z INFO ExtHandler ExtHandler [HEARTBEAT] Agent WALinuxAgent-2.9.1.1 is running as the goal state agent [DEBUG HeartbeatCounter: 1102;Heart
Sep 27 11:16:49 vmXXXXXXXXXX python3[4788]: 2023-09-27T11:16:49.960035Z INFO EnvHandler ExtHandler Current Firewall rules:
Sep 27 11:16:49 vmXXXXXXXXXX python3[4788]: Chain INPUT (policy ACCEPT 0 packets, 0 bytes)
Sep 27 11:16:49 vmXXXXXXXXXX python3[4788]:     pkts      bytes target     prot opt in     out     source               destination
Sep 27 11:16:49 vmXXXXXXXXXX python3[4788]: Chain FORWARD (policy ACCEPT 0 packets, 0 bytes)
Sep 27 11:16:49 vmXXXXXXXXXX python3[4788]:     pkts      bytes target     prot opt in     out     source               destination
Sep 27 11:16:49 vmXXXXXXXXXX python3[4788]: Chain OUTPUT (policy ACCEPT 13440 packets, 1173316 bytes)
Sep 27 11:16:49 vmXXXXXXXXXX python3[4788]:     pkts      bytes target     prot opt in     out     source               destination
Sep 27 11:16:49 vmXXXXXXXXXX python3[4788]:        0        0 ACCEPT     tcp  --  *      *       0.0.0.0/0            168.63.129.16        tcp dpt:53
Sep 27 11:16:49 vmXXXXXXXXXX python3[4788]:     5241  1352079 ACCEPT     tcp  --  *      *       0.0.0.0/0            168.63.129.16        owner UID match 0
Sep 27 11:16:49 vmXXXXXXXXXX python3[4788]:        0        0 DROP       tcp  --  *      *       0.0.0.0/0            168.63.129.16        ctstate INVALID,NEW
[...]

PR #2942, opened 2023-10-12
Cache daemon version

nagworld9:
## Description
fixes https://github.com/Azure/WALinuxAgent/issues/2940
Daemon version check command create new python environment on every call. This causing more execution time.  
Fix is to cache the value as it won't change.

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/common/version.py
+++ b/azurelinuxagent/common/version.py
@@ -32,6 +32,9 @@ __DAEMON_VERSION_ENV_VARIABLE = '_AZURE_GUEST_AGENT_DAEMON_VERSION_'
     The variable is set only on versions >= 2.2.53
 """
 
+# cache the daemon version to avoid running the command multiple times
+_DAEMON_VERSION = FlexibleVersion("0.0.0.0")
+
 
 def set_daemon_version(version):
     """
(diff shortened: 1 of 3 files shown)


nagworld9:
I want to add cpu usage check in agent cgroups scenario but scenario changes not there in release branch. I'll work on that change when we merge cherry-pick commit into develop.

As far as validation, I checked with profiling and also our reported cpu usage. They come back to normal numbers.

azurelinuxagent/common/version.py
```diff
@@ -51,19 +54,25 @@ def get_daemon_version():
     If the variable is not set (because the agent is < 2.2.53, or the process was not started by the daemon and
     the process is not the daemon itself) the function returns version of agent which started by the python
     """
+    # W0603: Using the global statement (global-statement)
+    # OK to disable; _DAEMON_VERSION is used only within get_daemon_version, but needs to persist across calls.
+    global _DAEMON_VERSION  # pylint: disable=W0603
     if __DAEMON_VERSION_ENV_VARIABLE in os.environ:
         return FlexibleVersion(os.environ[__DAEMON_VERSION_ENV_VARIABLE])
-    else:
+    elif _DAEMON_VERSION == FlexibleVersion("0.0.0.0"):
```
narrieta, line 62: rather than keeping this in a global, we should just set the environment variable, i.e. call set_daemon_version(version) and then return the env variable (or do a recursive call)

narrieta (approved):

maddieford (approved):

PR #2954, opened 2023-10-24
Add support for VM Scale Sets to end-to-end tests

narrieta:
Some of our tests require VM Scale Sets and LISA does not provide support for them. This PR adds that support in the AgentTestSuite.

The test metadata now includes a flag, 'executes_on_scale_set', to indicate that the test must run on a VMSS. When the flag is True, the AgentTestSuite will create the scale set and pass it as parameter to the tests as part of their context.

The LISA runbook is now generated dynamically by the AgentTestSuiteCombinator, using the 'platform' and 'environment' variables.

I made several changes to the object model in order to accommodate scale sets: 

* Now there are 2 types of tests: AgentVmTest and AgentVmssTest, which take an AgentVmTestContext and AgentVmssTestContext, respectively
* I merged AgentTestSuite._Context into AgentTestSuite
* I merged VmIdentifier into VirtualMachineClient
* I made some changes around the ssh utility functions 

While working on this code, I noticed a race condition when creating the tarball with the test tools. Multiple instances of the AgentTestSuite may be executing concurrently and all of them were writing to the same file, so one could overwrite the tarball while another one was copying it to the test node. I moved that code to common setup, which is protected by a lock,

I'll add comments in the PR pointing out the relevant changes.


tests_e2e/tests/lib/agent_test_context.py
```diff
@@ -17,161 +17,107 @@
...
+class AgentTestContext(ABC):
     """
-    Execution context for agent tests. Defines the test VM, working directories and connection info for the tests.
-
-    NOTE: The context is shared by all tests in the same runbook execution. Tests within the same test suite
-          are executed sequentially, but multiple test suites may be executed concurrently depending on the
-          concurrency level of the runbook.
+    Base class for the execution context of agent tests; includes the working directories and SSH info for the tests.
     """
-    class Paths:
```
narrieta, line 35: I removed these 2 helper classes, they made the code too verbose

tests_e2e/tests/lib/identifiers.py
```diff
@@ -18,31 +18,16 @@
 from typing import Dict, List
 
 
-class VmIdentifier(object):
```
narrieta, line 21: I merged this class into VirtualMachineClient. Makes the code less verbose.

tests_e2e/tests/lib/logging.py
```diff
@@ -155,3 +155,18 @@ def set_current_thread_log(log_file: Path):
         log.close_current_thread_log()
         if initial_value is not None:
             log.set_current_thread_log(initial_value)
+
```
narrieta, line 158: Moved here from the AgentTestSuite module

tests_e2e/tests/lib/ssh_client.py
```diff
@@ -28,11 +28,11 @@
 
 
 class SshClient(object):
-    def __init__(self, ip_address: str, username: str, private_key_file: Path, port: int = 22):
-        self._ip_address: str = ip_address
-        self._username: str = username
-        self._private_key_file: Path = private_key_file
-        self._port: int = port
+    def __init__(self, ip_address: str, username: str, identity_file: Path, port: int = 22):
```
narrieta, line 31: "private_key_file" is LISA's terminology. The SSH command line uses "identity_file"

tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
```diff
@@ -150,89 +140,250 @@ def create_environment_list(self) -> List[Dict[str, Any]]:
...
+            resource_group=self.runbook.resource_group_name,
+            name=self.runbook.vm_name)
+
+        ip_address = vm.get_ip_address()
+
+        return {
+            "c_env_name": self.runbook.vm_name,
+            "c_platform": [
+                {
+                    "type": "ready"
```
narrieta, line 248: "Ready" can be used to pass existing machine to LISA, "ready" meaning that they can be used without LISA initializing them or performing any operations on them. They are a good fit for existing VMs/VMSSs or to indicate to the AgentTestSuite that it needs to create a VMSS

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -232,181 +309,170 @@ def _setup(self) -> None:
...
-
-    def _get_agent_package_path(self) -> Path:
+    def _clean_up(self, success: bool) -> None:
         """
-        Returns the path to the agent package.
+        Cleans up any items created by the test suite run.
         """
-        return self.context.working_directory/"eggs"/f"WALinuxAgent-{AGENT_VERSION}.zip"
+        if self._delete_scale_set:
+            if self._keep_environment == KeepEnvironment.Always:
```
maddieford, line 374: If I remember correctly, I think lisa always skips clean up if we provide an existing vm (even if keep_environment is not always). Do we want to keep that behavior here?

tests_e2e/tests/lib/agent_test.py
```diff
@@ -94,3 +101,16 @@ def _run_remote_test(self, command: str, use_sudo: bool = False, attempts: int =
     @staticmethod
     def _indent(text: str, indent: str = " " * 8):
         return "\n".join(f"{indent}{line}" for line in text.splitlines())
+
+
+class AgentVmTest(AgentTest, ABC):
```
maddieford, line 106: what is ABC here?

tests_e2e/tests/lib/identifiers.py
maddieford: Looks like this file is entirely related to extension identifiers now. Should we update the name? Or do we anticipate adding non-extension related identifiers here in the future

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -40,19 +38,27 @@
...
 from lisa.environment import EnvironmentStatus  # pylint: disable=E0401
 from lisa.messages import TestStatus, TestResultMessage  # pylint: disable=E0401
+from lisa.node import LocalNode  # pylint: disable=E0401
+from lisa.util.constants import RUN_ID  # pylint: disable=E0401
 from lisa.sut_orchestrator.azure.common import get_node_context  # pylint: disable=E0401
+from lisa.sut_orchestrator.azure.platform_ import AzurePlatform  # pylint: disable=E0401
 
 import makepkg
 from azurelinuxagent.common.version import AGENT_VERSION
+
```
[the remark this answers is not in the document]
narrieta: I separated the imports in 4 sections:

1) standard python modules (lines 17-21 and lines 23-25, former uses "import", latter "from...import")
2) LISA modules (lines 27-44)
3) Agent modules (lines 46-47)
4) Agent Tests modules (lines 49-62)

This is just a personal preference, not a required coding standard so feel free to use a different convention when you create new files. I found this convention easy to follow.

tests_e2e/orchestrator/lib/agent_test_suite.py
narrieta: That is already taken care of by the check on self._delete_scale_set, which is set to True only when the AgentTestSuite itself creates the VM.

LISA does log an additional message about not deleting existing VMs. It seems redundant to me, but if you think it is useful I can add a similar message here, though that would require keeping an extra flag.

tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
```diff
@@ -150,89 +140,250 @@ def create_environment_list(self) -> List[Dict[str, Any]]:
...
-                        "c_vm_tags": c_vm_tags
-                    }
-                # pylint: enable=W0640
-
-                if suite_info.owns_vm:
-                    # create an environment for exclusive use by this suite
-                    environments.append(create_environment(f"{image_name}-{suite_info.name}"))
+                vm_size = self._get_vm_size(image)
+
+                if test_suite_info.owns_vm or not test_suite_info.install_test_agent:
```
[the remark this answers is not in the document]
narrieta: Good point, I need to add scale sets here, too

tests_e2e/tests/agent_not_provisioned/disable_agent_provisioning.py
```diff
@@ -22,11 +22,14 @@
 from tests_e2e.tests.lib.update_arm_template import UpdateArmTemplate
 
 
-class DenyOutboundConnections(UpdateArmTemplate):
```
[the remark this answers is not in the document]
narrieta: yup, the result of copy-n-paste programming :(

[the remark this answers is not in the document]
narrieta: the incorrect name did not affect functionality, though, since we load the class based on type (inhertis from UpdateArmTemplate) and the name is not used

tests_e2e/tests/agent_status/agent_status.py
```diff
@@ -139,7 +139,12 @@ def run(self):
         log.info("")
         log.info("*******Verifying the agent status updates 3 times*******")
 
-        vm = VirtualMachineClient(self._context.vm)
+        vm: VirtualMachineClient = VirtualMachineClient(
```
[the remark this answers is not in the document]
narrieta: yes, i'll revert this change. at some point I did a global replace and this was changed. the change does not affect functionality, but the previous code was more succinct and readable

tests_e2e/tests/extensions_disabled/extensions_disabled.py
```diff
@@ -109,7 +109,12 @@ def run(self):
         #
         # Validate that the agent continued reporting status even if it is not processing extensions
         #
-        vm: VirtualMachineClient = VirtualMachineClient(self._context.vm)
+        vm: VirtualMachineClient = VirtualMachineClient(
```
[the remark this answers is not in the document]
narrieta: will revert

tests_e2e/tests/fips/fips.py
```diff
@@ -45,7 +45,12 @@ def run(self):
             raise Exception(f"Failed to enable FIPS: {e}")
 
         log.info("Restarting test VM")
-        vm: VirtualMachineClient = VirtualMachineClient(self._context.vm)
+        vm: VirtualMachineClient = VirtualMachineClient(
```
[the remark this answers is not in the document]
narrieta: same

tests_e2e/tests/keyvault_certificates/keyvault_certificates.py
```diff
@@ -59,7 +59,12 @@ def run(self):
         else:
             log.info("Some test certificates had already been downloaded to the test VM (they have been deleted now):\n%s", existing_certificates)
 
-        vm: VirtualMachineClient = VirtualMachineClient(self._context.vm)
+        vm: VirtualMachineClient = VirtualMachineClient(
```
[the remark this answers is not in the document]
narrieta: ok, already!

:)

tests_e2e/tests/lib/add_network_security_group.py
```diff
@@ -66,72 +66,106 @@ def update(self, template: Dict[str, Any]) -> None:
...
         nsg_reference = json.loads(f"""{{
           "networkSecurityGroup": {{
             "id": "[resourceId('Microsoft.Network/networkSecurityGroups', '{NETWORK_SECURITY_GROUP}')]"
           }}
         }}""")
 
-        subnets_properties = subnets_input.get("properties")
-        if subnets_properties is None:
-            subnets_input["properties"] = nsg_reference
+        if is_lisa_template:
```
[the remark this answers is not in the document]
narrieta: No, we do not assume a vm or vmss neither for LISA's template or ours. But the 2 templates (no matter if they are for VM or VMSS) have different formats for the virtual network, which is what is important here.

tests_e2e/tests/lib/agent_test.py
narrieta: Standard python class, "Abstract Base Class". You combine this with the abstractmethod decorator and if you try to instantiate one of these classes without providing an implementation for all abstract methods, Python raises an exception.

It is actually not needed here, AgentTest is already an ABC. Will remove it.

tests_e2e/tests/lib/identifiers.py
narrieta: I thought the same thing (renaming) and laziness (or maybe another issue distracted me :)) won and I did not do it. I'll do it now.

tests_e2e/orchestrator/lib/agent_test_suite.py
maddieford: I don't think the log is necessary either

maddieford (approved):

PR #2957, opened 2023-10-26
Ignore dependencies when the extension does not have any settings

narrieta:
Until recently, CRP was ignoring extension dependencies when they originated from extensions with no settings. The fix for that issue exposed this bug in the Agent. When parsing the VmSettings, if a dependency is found, the agent always tries to access settings[0], causing an index out of range error.

As a temporary fix, the Agent will ignore the dependency in this case. The next release will include the full fix.

While checking the telemetry for this fix, I noticed a confusing message from self-update. I am removing it as part of this PR, too.

### DIFF
--- a/azurelinuxagent/common/event.py
+++ b/azurelinuxagent/common/event.py
@@ -109,6 +109,7 @@ class WALAEventOperation:
     OpenSsl = "OpenSsl"
     Partition = "Partition"
     PersistFirewallRules = "PersistFirewallRules"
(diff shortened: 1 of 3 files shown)


maddieford (approved):

nouralkadi:
any idea when the fix will be released?

narrieta:
@nouralkadi It's been released as part of https://github.com/Azure/WALinuxAgent/releases/tag/pre-v2.10.0.6

There was also an update on the service side to address that issue. That release should complete within the next few days.

PR #2974, opened 2023-11-03
GA versioning refactor plus fetch new rsm properties.

nagworld9:
## Description

 Agent update handler handles two type of agent updates. Handler initializes the updater to SelfUpdateVersionUpdater and switch to appropriate updater based on below conditions:
        RSM update: This is the update requested by RSM. The contract between CRP and agent is we get following properties in the goal state:
                    version: it will have what version to update
                    isVersionFromRSM: True if the version is from RSM deployment.
                    isVMEnabledForRSMUpgrades: True if the VM is enabled for RSM upgrades.
                    if vm enabled for RSM upgrades, we use RSM update path. But if requested version is not by rsm deployment
                    we ignore the update.
        Self update: We fallback to this if above is condition not met. This update to the largest version available in the manifest
                    Note: Self-update don't support downgrade.

   Handler keeps the rsm state of last update is with RSM or not on every new goal state. Once handler decides which updater to use, then
    does following steps:
      1. Retrieve the agent version from the goal state.
      2. Check if we allowed to update for that version.
      3. Log the update message.
      4. Purge the extra agents from disk.
      5. Download the new agent.
      6. Proceed with update.

Issue # 

---

### PR information


azurelinuxagent/ga/guestagent.py
```diff
@@ -18,15 +18,6 @@
 MAX_FAILURE = 3  # Max failure allowed for agent before declare bad agent
 
 
-class GAUpdateReportState(object):
-    """
-    This class is primarily used to maintain the in-memory persistent state for the agent updates.
-    This state will be persisted throughout the current service run and might be modified by external classes.
-    """
-    report_error_msg = ""
-    report_expected_version = FlexibleVersion("0.0.0.0")
```
nagworld9, line 27: Now I added it in updatestate class

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -36,105 +36,143 @@ class AgentUpdateHandlerUpdateState(object):
...
+        self.last_attempted_regular_update_time = datetime.datetime.min
         self.last_attempted_manifest_download_time = datetime.datetime.min
+        self.last_attempted_update_error_msg = ""
+        self.last_attempted_update_version = FlexibleVersion("0.0.0.0")
 
 
 class AgentUpdateHandler(object):
-
+    """
+    This class handles two type of agent updates and chooses the appropriate updater based on the below conditions:
```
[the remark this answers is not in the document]
nagworld9: updated semantics

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -34,107 +34,145 @@ class AgentUpdateHandlerUpdateState(object):
...
+        RSM update: This is the update requested by RSM. The contract between CRP and agent is we get following properties in the goal state:
+                    version: it will have what version to update
+                    isVersionFromRSM: True if the version is from RSM deployment.
+                    isVMEnabledForRSMUpgrades: True if the VM is enabled for RSM upgrades.
+                    if vm enabled for RSM upgrades, we use RSM update path. But if requested update is not by rsm deployment
+                    we ignore the update.
+                    This update is allowed once per (as specified in the conf.get_autoupdate_frequency())
```
maddieford, line 54: is this auto update frequency only used for RSM update? I see we have separate frequencies for hotfix and regular updates. If autoupdate_frequency is only used for RSM, could we update the name to differentiate it?

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -34,107 +34,145 @@ class AgentUpdateHandlerUpdateState(object):
...
+            self._is_vm_enabled_for_rsm_upgrades = False
         else:
-            next_attempt_time = now
-
-        if next_attempt_time > now:
-            return False
-        self.update_state.last_attempted_manifest_download_time = now
-        return True
+            self._is_version_from_rsm = self._get_is_version_from_rsm()
+            self._is_vm_enabled_for_rsm_upgrades = self._get_is_vm_enabled_for_rsm_upgrades()
```
maddieford, line 73: get_autoupdate_frequency() and get_autoupdate_frequency() already return False if **not os.path.exists(self._get_rsm_version_state_file()):**

I don't think you need the if/else condition here

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -160,38 +198,142 @@ def __get_agent_family_manifests(self, goal_state):
...
-        Returns: Requested version if supported and available in the GS
-                 None if requested version missing or GA versioning not enabled
+        Get the version from agent family
+        Returns: version if supported and available in the GS
+                 None if version is missing
         """
-        if conf.get_enable_ga_versioning() and agent_family.is_requested_version_specified:
-            if agent_family.requested_version is not None:
-                return FlexibleVersion(agent_family.requested_version)
+        if agent_family.version is not None:
```
maddieford, line 207: Will version always be in agent_family? Should we check here that the version property exists

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -228,23 +366,75 @@ def __purge_extra_agents_from_disk(current_version, known_agents):
...
+        # The time limit elapsed for us to allow updates.
+        return True
+
+    def should_update_agent(self, goal_state):
+        if not self._is_update_allowed_this_time():
+            return False
+
+        # we don't allow updates if version is not from RSM or downgrades below daemon version or if version is same as current version
+        if not self._is_version_from_rsm or self._version < self._daemon_version or self._version == CURRENT_VERSION:
+            return False
```
maddieford, line 410: If we hit this case then we don't end up doing the update, but we updated our attempt time to be now already in _is_update_allowed_this_time (line 400).

Is this the behavior we want? I think we shouldn't update the 'last_attempted_rsm_version_update_time' until we actually attempt the update

nagworld9: That is debatable. What I thought was, if we don't update the clock, if rsm request made to vms same time, then we end up does the update to all vms at the time. Even though this is small window, still does some help to spread.

azurelinuxagent/ga/agent_update_handler.py
nagworld9: same frequency used for manifest download in self-update, let me think better name

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -160,38 +198,142 @@ def __get_agent_family_manifests(self, goal_state):
...
-            if pkg_version > largest_version:
-                largest_version = pkg_version
-        return largest_version
+    def run(self, goal_state):
+
+        try:
+            # Ignore new agents if update is disabled. The latter flag only used in e2e tests.
+            if not self._autoupdate_enabled or not conf.get_download_new_agents():
+                return
+
```
narrieta, line 217: As discussed previously, lines 218 to 229 are executed in the main loop every 6 seconds. Those lines invoke methods than in turn invoke other methods. That is too much logic. It's way too easy to introduce bugs that produce too much telemetry, too much log output, consume too much CPU, etc. Please work on moving the check for the timer currently at line 232 right after the check for update enable on line 215.

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -160,38 +198,142 @@ def __get_agent_family_manifests(self, goal_state):
...
+            agent_family = self._get_agent_family_manifest(goal_state)
+            version = self._get_version_from_gs(agent_family)
```
narrieta, line 219: why are we trying to get the version from the goal state at this point? so far we haven't decided yet whether we are doing self-update or RSM update and the former does not need the version

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -160,38 +198,142 @@ def __get_agent_family_manifests(self, goal_state):
...
     @staticmethod
-    def __get_requested_version(agent_family):
+    def _get_version_from_gs(agent_family):
         """
-        Get the requested version from agent family
-        Returns: Requested version if supported and available in the GS
-                 None if requested version missing or GA versioning not enabled
+        Get the version from agent family
+        Returns: version if supported and available in the GS
+                 None if version is missing
```
narrieta, line 205: Returning None makes the code harder to write/review/maintain.

For self-update we should not even be looking at the version in the goal state. If it is missing, we do not care. 

For RSM update, the version must not be missing, so you can raise an update exception here to report the error and skip the update, instead of returning None.

This is a consequence of my other comment below: that we are looking at the version before we have decided whether we are doing self-update or RSM update

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
```diff
@@ -267,10 +269,16 @@ def _parse_agent_manifests(self, vm_settings):
         for family in families:
             name = family["name"]
             version = family.get("version")
+            is_version_from_rsm = family.get("isVersionFromRSM")
+            is_vm_enabled_for_rsm_upgrades = family.get("isVMEnabledForRSMUpgrades")
             uris = family.get("uris")
             if uris is None:
                 uris = []
             agent_family = VMAgentFamily(name, version)
+            if is_version_from_rsm is not None:
```
narrieta, line 278: This is confusing. Why are you enforcing passing a version to VMAgentFamily, but not passing is_version_from_rsm and is_vm_enabled_for_rsm_updates? in other parts of the code you are checking for version to not be None, so apparently all 3 of them can be None

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
```diff
@@ -267,10 +269,16 @@ def _parse_agent_manifests(self, vm_settings):
...
+                agent_family.is_version_from_rsm = is_version_from_rsm
```
narrieta, line 279: Related comment: naming a variable "is_*" creates a strong expectation that it is a bool, but these 2 variables seem to actually be tri_state: None, False, or True

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -160,38 +198,142 @@ def __get_agent_family_manifests(self, goal_state):
...
+            # Ignore new agents if update is disabled. The latter flag only used in e2e tests.
+            if not self._autoupdate_enabled or not conf.get_download_new_agents():
+                return
+
+            agent_family = self._get_agent_family_manifest(goal_state)
+            version = self._get_version_from_gs(agent_family)
+            gs_id = goal_state.extensions_goal_state.id
+            self._update_rsm_version_state_if_changed(goal_state.extensions_goal_state.created_on_timestamp, agent_family)
+            # if version is specified and vm is enabled for rsm upgrades, use rsm update path, else sef-update
+            if version is None and self._is_vm_enabled_for_rsm_upgrades and self._is_version_from_rsm:
```
narrieta, line 223: this check should be done when we are picking up those properties from the goal state

azurelinuxagent/ga/agent_update_handler.py
narrieta: I added a comment similar to @maddieford 's in a previous iteration of this code. I also find that confusing

azurelinuxagent/common/protocol/extensions_goal_state_from_vm_settings.py
narrieta, line 279: Another related comment: Try to avoid modelling those flags as tri-state. Allowing None makes the code harder to write/review/maintain, since the consumers of those flags need to keep checking for None.

In other parts of the code, the flags are actually bools (they are never None) and that makes code hard to understand.
As an example, should agent_update_handler.py@409 be checking for None?

`if not self._is_version_from_rsm or self._version < self._daemon_version or self._version == CURRENT_VERSION`

(it shouldn't, but that is hard to understand without digging into the implementation in detail; it should not be that hard)

nagworld9: The intention for tri-state modeling is agent_family should reflect what we see in the goal state. When we parse the goal state, if property missing (possible scenario) consider as None instead false because agent_update should know the state to compute latest state. If agent update last known as True, but in next goal state property missing, how do I know I should consider last known state (True) not the new state in the case of property missing. If I convert that into False, I wouldn't know if that's due to missing property or it's actual value. 

The state that we maintain in agent_update is two-state (True/False) but the agent_family model reflects the goal state(try-state).

narrieta: Yes, but that is very hard to understand without digging deep into the code. Shouldn't be this hard for a simple flag. There are better ways to model this.

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -160,38 +198,142 @@ def __get_agent_family_manifests(self, goal_state):
...
+                If largest version is found in manifest, upgrade to that version. Downgrade is not supported.
+                Raises: AgentUpgradeExitException
+        """
+        raise NotImplementedError
+
+    def download_and_get_new_agent(self, protocol, goal_state):
+        """
+        This function downloads the new agent and returns the downloaded version.
+        """
+        if self._agent_manifest is None:  # Fetch agent manifest if it's not already done
```
narrieta, line 329: The use of None in this class is also confusing/problematic. The init method takes all these parameters:

```
    def __init__(self, gs_id, agent_family, agent_manifest, version, update_state):
        self._gs_id = gs_id
        self._agent_family = agent_family
        self._agent_manifest = agent_manifest
        self._version = version
        self._update_state = update_state

```

Does this check for None on self.agent_manifest implies than any of these properties can be None and we need to check before using any of them? Some of them can be None, and some can't? How would the reader of this code know which ones can and which ones can't?

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -160,38 +198,142 @@ def __get_agent_family_manifests(self, goal_state):
...
+        self._gs_id = gs_id
+        self._agent_family = agent_family
+        self._agent_manifest = agent_manifest
+        self._version = version
+        self._update_state = update_state
+
+    def should_update_agent(self, goal_state):
         """
-        This function downloads the new agent(requested version) and returns the downloaded version.
+        RSM version update:
```
narrieta, line 286: These comments should probably be in the derived classes to make the code easier to read instead of having to scroll up and down to see the comments

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -160,38 +198,142 @@ def __get_agent_family_manifests(self, goal_state):
...
+                return VMAgentUpdateStatus(expected_version=str(self.update_state.last_attempted_update_version), status=status, code=code, message=self.update_state.last_attempted_update_error_msg)
+        except Exception as err:
+            msg = "Unable to report agent update status: {0}".format(textutil.format_exception(err))
+            logger.warn(msg)
+            add_event(op=WALAEventOperation.AgentUpgrade, is_success=False, message=msg, log_event=True)
+        return None
 
-    def __download_and_get_agent(self, goal_state, agent_family, agent_manifest, requested_version):
+
+class GAVersionUpdater(object):
```
narrieta, line 275: GAVersionUpdater and its derived classes should probably be in a separate file. The rough rule of thumb is "one class per source file", small/related classes can share the same file)

narrieta, line 275: Thanks for creating those classes. It's a good step in the right direction. However, some of the state and logic that should be in these classes is still spread around other classes.

As an example, this state should be in these 2 derived classses:

```
    def __init__(self, gs_id, agent_family, agent_manifest, version, update_state):
        self._gs_id = gs_id
        self._agent_family = agent_family
        self._agent_manifest = agent_manifest
        self._version = version
        self._update_state = update_state

```

Same for these properties:

```
class AgentUpdateHandler(object):
    def __init__(self, protocol):
         ...
        # restore the state of rsm update
        if not os.path.exists(self._get_rsm_version_state_file()):
            self._is_version_from_rsm = False
            self._is_vm_enabled_for_rsm_upgrades = False
        else:
            self._is_version_from_rsm = self._get_is_version_from_rsm()
            self._is_vm_enabled_for_rsm_upgrades = self._get_is_vm_enabled_for_rsm_upgrades()
```

And this logic: 

```
    def _get_is_version_from_rsm(self):
    def _get_is_vm_enabled_for_rsm_upgrades(self):
    def _get_rsm_state_used_gs_timestamp(self):
    def _update_rsm_version_state_if_changed(self, goalstate_timestamp, agent_family):
```

At this point you can't move this state to the updaters because you are creating them on each iteration of the main loop and this state needs to persist across iterations. But instantiating on each iteration is not needed. You only need to instantiate them only when there is a new goal state and the "is vm enabled for RSM updates" property changes. This approach would also address my other concern about having too much code/logic executing on each iteration of the main loop.

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -160,38 +198,142 @@ def __get_agent_family_manifests(self, goal_state):
...
+        RSM version update:
+                upgrade/downgrade to the specified version.
+                Raises: AgentUpgradeExitException
+        self-update:
+                If largest version is found in manifest, upgrade to that version. Downgrade is not supported.
+                Raises: AgentUpgradeExitException
+        """
+        raise NotImplementedError
+
+    def download_and_get_new_agent(self, protocol, goal_state):
```
narrieta, line 325: Another issue with the design of this method is that it takes the goal state as parameter and the init method takes several properties from the goal state. With this design, this goal state and the properties in init may correspond to different goal states. You may want to consider passing the goal state on init and remove it from here

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -160,38 +198,142 @@ def __get_agent_family_manifests(self, goal_state):
...
+            add_event(op=WALAEventOperation.AgentUpgrade, is_success=False, message=error_msg, log_event=False)
+            self.update_state.last_attempted_update_error_msg = error_msg
+
+    def get_vmagent_update_status(self):
+        """
+        This function gets the VMAgent update status as per the last attempted update.
+        Returns: None if fail to report or update never attempted with rsm version specified in GS
+        """
+        try:
+            if conf.get_enable_ga_versioning() and self._is_vm_enabled_for_rsm_upgrades and self._is_version_from_rsm:
```
narrieta, line 260: why do we need to check for self._is_version_from_rsm here? Does the protocol with CRP require that we do not report status when _is_vm_enabled_for_rsm_upgrades is true but _is_version_from_rsm is not?

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -34,107 +34,145 @@ class AgentUpdateHandlerUpdateState(object):
...
+
+    def _save_rsm_version_state(self, isVersionFromRSM, isVMEnabledForRSMUpgrades, timestamp):
         """
-        Get the next upgrade times
-        return: Next Hotfix Upgrade Time, Next Normal Upgrade Time
+        Save the rsm state to the file
         """
+        try:
+            with open(self._get_rsm_version_state_file(), "w") as file_:
+                json.dump({"isVersionFromRSM": isVersionFromRSM,
```
narrieta, line 95: why do we need to persist isVersionFromRSM?

azurelinuxagent/ga/agent_update_handler.py
narrieta, line 223: the check on version looks suspicious. i'll go over this offline

azurelinuxagent/common/conf.py
```diff
@@ -479,8 +479,8 @@ def get_autoupdate_enabled(conf=__conf__):
     return conf.get_switch("AutoUpdate.Enabled", True)
 
 
-def get_autoupdate_frequency(conf=__conf__):
-    return conf.get_int("Autoupdate.Frequency", 3600)
+def get_agentupdate_frequency(conf=__conf__):
+    return conf.get_int("AgentUpdate.Frequency", 3600)
```
narrieta, line 483: We should not rename these parameters, since customers may be using them

azurelinuxagent/common/protocol/extensions_goal_state_from_extensions_config.py
```diff
@@ -61,9 +61,17 @@ def _parse_extensions_config(self, xml_text, wire_client):
         for ga_family in ga_families:
             name = findtext(ga_family, "Name")
             version = findtext(ga_family, "Version")
+            is_version_from_rsm = findtext(ga_family, "IsVersionFromRSM")
+            is_vm_enabled_for_rsm_upgrades = findtext(ga_family, "IsVMEnabledForRSMUpgrades")
             uris_list = find(ga_family, "Uris")
             uris = findall(uris_list, "Uri")
-            family = VMAgentFamily(name, version)
+            family = VMAgentFamily(name)
+            if version is not None:
```
narrieta, line 69: why do we need these checks for None, if VMAgentFamily initializes these properties to None anyways?

this gives the impresion that those properties should not be None

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -1,146 +1,122 @@
...
+        if not os.path.exists(self._get_rsm_update_state_file()):
+            return False
 
-        return next_hotfix_time, next_normal_time
+        try:
+            with open(self._get_rsm_update_state_file(), "r") as file_:
+                return json.load(file_)["isLastUpdateWithRSM"]
+        except Exception as e:
+            logger.warn(
+                "Can't retrieve the isLastUpdateWithRSM from rsm state file ({0}), will assume it False. Error: {1}",
```
narrieta, line 109: assuming False may be incorrect.

the file just contains a boolean; you can eliminate this error condition by using the presence of the file as the boolean: if it exits, it's RSM, otherwise it's sefl-update

azurelinuxagent/ga/rsm_version_updater.py
```diff
@@ -0,0 +1,159 @@
...
+        return True
+
+    def check_and_switch_updater_if_changed(self, agent_family, gs_id):
+        """
+        Checks if there is a new goal state and decide if we need to continue with rsm update or switch to self-update.
+        Firstly it checks agent supports GA versioning or not. If not, we raise exception to switch to self-update.
+        if vm is enabled for RSM updates and continue with rsm update, otherwise we raise exception to switch to self-update.
+        if either isVersionFromRSM or isVMEnabledForRSMUpgrades is missing in the goal state, we ignore the update as we consider it as invalid goal state.
+        """
+        if self._gs_id != gs_id:
```
narrieta, line 74: This check should be done with the value returned by UpdateHandler._processing_new_extensions_goal_state()

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -151,211 +127,55 @@ def __get_agent_family_manifests(self, goal_state):
                     agent_family_manifests.append(m)
 
         if not family_found:
-            raise AgentUpdateError(u"Agent family: {0} not found in the goal state, skipping agent update".format(family))
+            raise AgentUpdateError(u"Agent family: {0} not found in the goal state incarnation: {1}, skipping agent update".format(family, self._gs_id))
```
narrieta, line 130: "incarnation" applies only to Fabric goal states

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -151,211 +127,55 @@ def __get_agent_family_manifests(self, goal_state):
...
-                if "Missing requested version" in GAUpdateReportState.report_error_msg:
-                    GAUpdateReportState.report_error_msg = ""
-
-            # Check if an update is allowed and update next upgrade times even if no new agent is available for upgrade
-            if not self.__check_if_agent_update_allowed_and_update_next_upgrade_times(requested_version):
+            if not conf.get_autoupdate_enabled() or not conf.get_download_new_agents():
                 return
 
-            if requested_version == CURRENT_VERSION:
+            # verify if agent update is allowed this time (RSM checks 1 hr interval; self-update checks manifest download interval)
```
narrieta, line 145: RSM updates need to be checked on every goal state, the 1-hour timer is just for retries

azurelinuxagent/ga/rsm_version_updater.py
```diff
@@ -0,0 +1,159 @@
...
+                if agent_family.is_version_from_rsm is None:
+                    raise AgentUpdateError(
+                        "Received invalid goal state:{0}, missing isVersionFromRSM property. So, skipping agent update".format(
+                            gs_id))
+
+    def retrieve_agent_version(self, agent_family, goal_state):
+        """
+        Get the agent version from the goal state
+        """
+        if agent_family.version is None and agent_family.is_vm_enabled_for_rsm_upgrades and agent_family.is_version_from_rsm:
```
narrieta, line 95: We talked about encapsulating the use of those tri-state variables in 1 single method, otherwise they are difficult to use by callers (they always need to check for None). The method above, check_and_switch_updater_if_changed, is protecting itself against None, but this method is not.

azurelinuxagent/ga/self_update_version_updater.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def check_and_switch_updater_if_changed(self, agent_family, gs_id):
+        """
+        Checks if there is a new goal state and decide if we need to continue with self-update or switch to rsm update.
+        if vm is not enabled for RSM updates or agent not supports GA versioning then we continue with self update, otherwise we raise exception to switch to rsm update.
+        if isVersionFromRSM is missing but isVMEnabledForRSMUpgrades is present in the goal state, we ignore the update as we consider it as invalid goal state.
+        """
+        if self._gs_id != gs_id:
+            self._gs_id = gs_id
+            if conf.get_enable_ga_versioning() and agent_family.is_vm_enabled_for_rsm_upgrades:
```
narrieta, line 127: another method that is not protecting itself against None

azurelinuxagent/ga/self_update_version_updater.py
```diff
@@ -0,0 +1,181 @@
...
+                                                  conf.get_self_update_regular_frequency())
+
+        return next_hotfix_time, next_regular_time
+
+    def _is_new_agent_allowed_update(self):
+        """
+        This method ensure that update is allowed only once per (hotfix/Regular) upgrade frequency
+        """
+        now = datetime.datetime.now()
+        next_hotfix_time, next_regular_time = self._get_next_upgrade_times(now)
```
narrieta, line 83: why do we need to compute both the hotfix and regular times? at any given time, only one of them is valid

narrieta (commented):
thanks for the updates, code improved a lot

added some extra comments

azurelinuxagent/common/conf.py
nagworld9: I agree customer may have taken dependency, my intention is this frequency is shared between self and rsm update. So I thought auto update my confuse. Will revert it.

azurelinuxagent/ga/agent_update_handler.py
nagworld9: well, even true is not right all the time. I feel assume false and self-update more safe than true. With rsm, we need extra flags to move forward with update. If it's not there we fail the update. That will stop the update but self-update at least customers vms on latest version. Later in time switch back to RSM if flags are showed up.

As far as presence file as boolean. I already check the presence in Line #100. So at this point we can assume true if that is what we want

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -151,211 +127,55 @@ def __get_agent_family_manifests(self, goal_state):
...
-                self.__log_event(LogLevel.WARNING, warn_msg)
-
-            # Downgrades are not allowed for self-update version
-            if not self.__check_if_downgrade_is_requested_and_allowed(requested_version):
+            agent_family = self._get_agent_family_manifest(goal_state)
+            gs_id = goal_state.extensions_goal_state.id
+            try:
+                # updater will raise exception if we need to switch to self-update or rsm update
+                self._updater.check_and_switch_updater_if_changed(agent_family, gs_id)
+            except VMDisabledRSMUpdates:
```
narrieta, line 154: Opting in and out of RSM are not error conditions, but the code communicates those transitions with exceptions, which is odd. Exceptions are used to communicate errors. Consider changing that to conditionals.

azurelinuxagent/ga/agent_update_handler.py
nagworld9: I had concern we do the update on extension goal states too If we go with every goal state

azurelinuxagent/ga/rsm_version_updater.py
nagworld9: Yes, I felt version check None in check_and_switch_updater_if_changed method is not a good place. That's why I added in retrieve version method. 

Still adding confusion, I can place it in same method

azurelinuxagent/ga/self_update_version_updater.py
nagworld9: which None are you talking? is it is_vm_enabled_for_rsm_upgrades? That is intentional. In the self-update, I don't want to fail with error (stop the updates) if it's None rather I want to continue with self-update until flags showed up. Once the enabled flag is there, we should have other flag otherwise we don't switch to rsm-update.

azurelinuxagent/ga/self_update_version_updater.py
nagworld9: good point, before we compute I can determine hotfix or regualr version and depending on that compute one time. Makes sense

azurelinuxagent/ga/agent_update_handler.py
narrieta: Exactly, assuming True is not safe either.

Hence my suggestion of just using an empty file, and if the file exist the value (is RSM) is True,  otherwise it is False

narrieta: This is what I mean
```

def _get_is_last_update_with_rsm(self):
    return os.path.exists(self._get_rsm_update_state_file())

```
and change _save_rsm_update_stateto create the file when its argument is True and delete it when it is False.

azurelinuxagent/ga/agent_update_handler.py
narrieta: Not sure I understand your concern. We can go over it offline if you want.

My understanding is that the contract with CRP is that the Agent must process the goal state requesting the update, not 1 hour later (or less, depending on timing)

azurelinuxagent/ga/rsm_version_updater.py
narrieta: My suggestion is to encapsulate this in the GoalState class. Make the tri-state variables private and provide access to them only thru 1 method, let's say get_agent_update_info, which returns a struct with the two booleans and the string, and throws exceptions if the values in the goal state are invalid or missing.

Of course, there are other ways to do this. The idea is to limit the potential use of invalid values to 1 single method instead of having every single method that uses those properties checking for None.

azurelinuxagent/ga/self_update_version_updater.py
narrieta: Are you checking for None on line 127? If looks like a check on a boolean. If you are checking against None, please do an explicit comparison against None. 

Taking advantage that None evalueates as False in a bool context is OK for small, one-single-use scripts, not so good for production code in a large code base.

narrieta: Having said that, see my comment above wrt allowing those properties to be None. Please don't. Encapsulate the use of invalid values (None) in a single method.

azurelinuxagent/ga/agent_update_handler.py
nagworld9: sure, we can discuss in sync. 

My point, how do you GS related to Agent update request not the extension operation request. If we allow on every goal state, for example if we get 10 new GS within 1-hour timer, we end up fetching the agent family 10 times and validate that version is already updated or not which is unnecessary. It seems there is some tradeoffs we need to do. If we want this thing, we have to pay the price of doing extra checks

azurelinuxagent/ga/rsm_version_updater.py
```diff
@@ -0,0 +1,159 @@
...
+
+        if self._last_attempted_rsm_version_update_time != datetime.datetime.min:
+            next_attempt_time = self._last_attempted_rsm_version_update_time + datetime.timedelta(
+                seconds=conf.get_agentupdate_frequency())
+        else:
+            next_attempt_time = now
+
+        if next_attempt_time > now:
+            return False
+        self._last_attempted_rsm_version_update_time = now
```
maddieford, line 63: Why do we update the last_attempted_rsm_version_update_time before we actually proceed with the update? If for some reason we can't update to the provided version, then we would have to wait one whole frequency before attempting another update with a different version. 

Not necessarily suggesting we change this, but just wanting to understand the decision

azurelinuxagent/ga/rsm_version_updater.py
```diff
@@ -0,0 +1,159 @@
...
+        if vm is enabled for RSM updates and continue with rsm update, otherwise we raise exception to switch to self-update.
+        if either isVersionFromRSM or isVMEnabledForRSMUpgrades is missing in the goal state, we ignore the update as we consider it as invalid goal state.
+        """
+        if self._gs_id != gs_id:
+            self._gs_id = gs_id
+            if not conf.get_enable_ga_versioning():
+                raise VMDisabledRSMUpdates()
+
+            if agent_family.is_vm_enabled_for_rsm_upgrades is None:
+                raise AgentUpdateError(
```
maddieford, line 80: if we have a VM which was enabled for RSM upgrades and then is switched to a host which does not send down the is_vm_enabled_for_rsm_upgrades, then we will not process any goal states?

What is the solution for customers who may end up getting blocked by this? I'm thinking we can use the GA versioning flag in conf. Is there another way?

azurelinuxagent/ga/rsm_version_updater.py
narrieta: This has been pointed out at least twice in previous iterations of this PR. I agree with @maddieford's comment

azurelinuxagent/ga/rsm_version_updater.py
nagworld9: As we discussed offline, there isn't much we can do about it.

narrieta (approved):

maddieford (approved):

PR #2977, opened 2023-11-08
Save goal state history explicitly

narrieta:
We have multiple threads/processes fetching the goal state and all of those operations save the goal state to the history folder, which can end up messing up the history folder due to concurrency issues. 

We use the history folder to debug issues in extension handling, so we need only the main thread to save the goal state. I added a new parameter, 'save_to_history', which is False by default and only the extension handler thread sets to True. The parameter had to be propagated thru several layers in the code, as well as in the test mocks.

### DIFF
--- a/azurelinuxagent/common/protocol/util.py
+++ b/azurelinuxagent/common/protocol/util.py
@@ -188,7 +188,7 @@ class ProtocolUtil(SingletonPerThread):
                 return
             logger.error("Failed to clear wiresever endpoint: {0}", e)
 
-    def _detect_protocol(self, init_goal_state=True):
+    def _detect_protocol(self, init_goal_state=True, save_to_history=False):
         """
         Probe protocol endpoints in turn.
         """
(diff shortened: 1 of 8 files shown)


maddieford (approved):

azurelinuxagent/common/protocol/util.py
```diff
@@ -188,7 +188,7 @@ def _clear_wireserver_endpoint(self):
                 return
             logger.error("Failed to clear wiresever endpoint: {0}", e)
 
-    def _detect_protocol(self, init_goal_state=True):
+    def _detect_protocol(self, init_goal_state=True, save_to_history=False):
```
nagworld9, line 191: since this method defined as internal use and only not called outside of this. Shouldn't we take parameter from calling method instead default?

narrieta: good point, i removed the default value from this method

nagworld9 (approved):

maddieford (approved):

PR #3002, opened 2023-12-22
AMA is not supported on cbl-mariner 1.0

maddieford:
## Description
Update AMA unsupported distro list to include cbl-mariner 1.0:
"The handler for VM extension type 'Microsoft.Azure.Monitor.AzureMonitorLinuxAgent' has reported terminal failure for VM extension 'AzureMonitorLinuxAgent' with error message: '[ExtensionOperationError] Non-zero exit code: 51, /var/lib/waagent/Microsoft.Azure.Monitor.AzureMonitorLinuxAgent-1.29.4/./shim.sh -install\n[stdout]\n2023/12/14 10:Failed to check disk usage.\nor.AzureMonitorLinuxAgent-1.29.4] Install,failed,51,Unsupported operating system: mariner 1.0\n\n\n[stderr]\n'.\r\n \r\n'Install handler failed for the extension. More information on troubleshooting is available at [https://aka.ms/vmextensionlinuxtroubleshoot'"](https://aka.ms/vmextensionlinuxtroubleshoot%27%22)

From AMA team:
"We still support Mariner 1.0. This looks like a parsing error of OS. We recognize 'cbl-mariner' for Mariner 1.0. There are two ways to parse OS: 1) old python use platform.linux_distribution() to get the OS 2) manually parse OS from /etc/os-release. It seems the current parsing returns mariner instead of cbl-mariner as the OS so we didn't recoginize it as our supported distro"

Waiting on fix from AMA

Issue # 

---

### PR information

### DIFF
--- a/tests_e2e/tests/ext_sequencing/ext_sequencing.py
+++ b/tests_e2e/tests/ext_sequencing/ext_sequencing.py
@@ -159,7 +159,7 @@ class ExtSequencing(AgentVmssTest):
         for instance in instances_ip_address:
             ssh_clients[instance.instance_name] = SshClient(ip_address=instance.ip_address, username=self._context.username, identity_file=self._context.identity_file)
 
-        if not VmExtensionIds.AzureMonitorLinuxAgent.supports_distro(next(iter(ssh_clients.values())).run_command("uname -a")):
+        if not VmExtensionIds.AzureMonitorLinuxAgent.supports_distro(next(iter(ssh_clients.values())).run_command("cat /etc/*-release", use_sudo=True)):
             raise TestSkipped("Currently AzureMonitorLinuxAgent is not supported on this distro")
 
         # This is the base ARM template that's used for deploying extensions for this scenario
--- a/tests_e2e/tests/ext_sequencing/ext_sequencing.py
+++ b/tests_e2e/tests/ext_sequencing/ext_sequencing.py
@@ -159,7 +159,7 @@ class ExtSequencing(AgentVmssTest):
         for instance in instances_ip_address:
             ssh_clients[instance.instance_name] = SshClient(ip_address=instance.ip_address, username=self._context.username, identity_file=self._context.identity_file)
 
-        if not VmExtensionIds.AzureMonitorLinuxAgent.supports_distro(next(iter(ssh_clients.values())).run_command("cat /etc/*-release", use_sudo=True)):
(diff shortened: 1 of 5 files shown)


tests_e2e/tests/ext_sequencing/ext_sequencing.py
```diff
@@ -159,7 +159,7 @@ def run(self):
         for instance in instances_ip_address:
             ssh_clients[instance.instance_name] = SshClient(ip_address=instance.ip_address, username=self._context.username, identity_file=self._context.identity_file)
 
-        if not VmExtensionIds.AzureMonitorLinuxAgent.supports_distro(next(iter(ssh_clients.values())).run_command("uname -a")):
```
narrieta, line 162: We use "uname -a" to check the distro in multiple places in the code. I am concerned with updating only this one instance.

Actually, maybe we should use get_distro(), in the Agent's code to do this

maddieford: Updated to use get_distro to get distro name and version for all test suites which use "uname -a"

tests_e2e/tests/lib/vm_extension_identifier.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, publisher: str, ext_type: str, version: str):
 
     unsupported_distros: Dict[str, List[str]] = {
         "Microsoft.OSTCExtensions.VMAccessForLinux": ["flatcar"],
-        "Microsoft.Azure.Monitor.AzureMonitorLinuxAgent": ["flatcar"]
+        "Microsoft.Azure.Monitor.AzureMonitorLinuxAgent": ["flatcar", "mariner_1"]
```
narrieta, line 37: get_distro.py produces <distro>_<version> and here we are only using the name ("flatcar")

maddieford: We check if any of these are a substring in the result of get_distro.py. This check is on Line 45

maddieford: If we only put distro name in this list, then we know that extension doesn't support any version of the distro. If we put <name>_<version> then we can mark that the extension only excludes a certain version. I did this since mariner_1 is unsupported by AMA, but mariner_2 is supported

narrieta: where do we strip the version so that "flatcar" matches the value returned by get_distro.py?

narrieta (approved):

nagworld9 (approved):

PR #3007, opened 2024-01-02
Use self-hosted pool for automation runs

narrieta:
These changes use a self-hosted Azure Pipelines Agent Pool to execute the end-to-end tests. This pool is hosted within corpnet; its setup is described in the test wiki.

The test setup still adds a rule allowing SSH access from the orchestrator machine, because the rule allowing connections from corpnet is added several minutes after the rule denying all inbound access is added, which produces a very long window of time when SSH connections fail.

This rule, however, now needs to be added explicitly during test runs using the new "allow_ssh" runbook parameter.


tests_e2e/pipeline/pipeline.yml
```diff
@@ -46,7 +46,7 @@ parameters:
   - name: collect_lisa_logs
     displayName: Collect LISA logs
     type: boolean
-    default: true
```
narrieta, line 49: i've been meaning to disable this for several PRs; finally, I remember to do it in this one.

tests_e2e/pipeline/scripts/execute_tests.sh
```diff
@@ -10,7 +13,7 @@ WAAGENT_UID=1000
 #
 # Set the correct mode and owner for the private SSH key and generate the public key.
 #
-cd "$HOME"
+cd "$AGENT_TEMPDIRECTORY"
```
narrieta, line 16: initially i was re-using the agent vms (currently they are destroyed after each run), so I moved these from $HOME to the temp directory

maddieford (approved):

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -923,7 +927,8 @@ def read_file(path: str) -> str:
                         "publisher": "[parameters('publisher')]"
                     }
 
-        AddNetworkSecurityGroup().update(template, is_lisa_template=False)
+        if self._allow_ssh != '':
```
nagworld9, line 930: This give me impression this value can be empty, and we let go without failing the tests? Eventually test fails with ssh issue since rule was not added because ip empty. I think we should fail fast instead tests fails with ssh rule

narrieta: yes, it can be empty

(from runbook.yml)
```
  #
  # Takes an IP address as value; if not empty, it adds a Network Security Rule allowing SSH access from the specified IP address to any test VMs created by the runbook execution.
  #
  - name: allow_ssh
    value: ""
    is_case_visible: true
    
```

from the PR description

`
This rule, however, now needs to be added explicitly during test runs using the new "allow_ssh" runbook parameter.
`

nagworld9: My point is tests may fail with ssh issue if user not providing as runbook parameter, why don't we let user to know by raising exception that need this parameter when it's empty

narrieta: empty is not an error.

short tests execute without issue. longer runs may hit issues.

narrieta: automation runs always set this parameter.

the intention here is to force developers to think about what they are doing (adding a rule that can cause problems). developers many times forget to delete their test VMs and we get security warnings on the test subscription. if the developer needs to add the rule, I want them to be aware of what they are doing.

if the developer doesn't need to add a rule because they are just executing a short test, I don't want to force them to add a rule (i.e. I do not want to report that as an error)

nagworld9: makes sense

nagworld9 (approved):

PR #3008, opened 2024-01-02
NM should not be restarted during hostname publish if NM_CONTROLLED=y

maddieford:
## Description
In the Fedora 7.* distros if NM_CONTROLLED=y in the network interface configuration file, the NM has permissions to do operations on the network interface and will restart the interface on service start. NM_CONTROLLED=y by default. When the agent publishes an updated hostname to DNS, it restarts the NM and then restarts the interface configuration manually. This can lead to a race condition, because both the NetworkManager service and the Guest Agent service are doing operations on the network interface simultaneously. If NM_CONTROLLED=n, then the NM does not have permissions to do any operations on the interface, so there is no race condition with the agent's operations on the interface. 

It is not well documented why the NetworkManager service restart was ever added to this OSUtil (it was done many years ago). We have not identified any issues when removing this call to restart the NetworkManager in our tests, but our tests also show that there is no issue with keeping this restart when NM_CONTROLLED=n. This PR takes a cautious approach and leaves the call to restart NM only when NM_CONTROLLED=n. 

This was a customer reported issue.

Issue # 

---

### PR information


narrieta (dismissed):

azurelinuxagent/common/osutil/redhat.py
```diff
@@ -117,11 +117,38 @@ def set_hostname(self, hostname):
...
+            if result and len(result.split('=')) > 1:
+                value = result.split('=')[1]
+                if "n" in value:
+                    return False
+        except shellutil.CommandError as e:
+            # Command might fail because NM_CONTROLLED value is not in interface config file (exit code 1).
+            # Log warning for any other exit code.
+            # NM_CONTROLLED=y by default if not specified.
+            if e.returncode != 1:
+                logger.warn("[{0}] failed: {1}.\nAgent will continue to publish hostname with NM_CONTROLLED=y".format(' '.join(nm_controlled_cmd), e))
```
nagworld9, line 136: users might not know about NM_CONTROLLED=y. Do you think logging this "Agent will continue to publish hostname without NetworkManager restart" is clearer?

nagworld9: or adding both?

azurelinuxagent/common/osutil/redhat.py
```diff
@@ -117,11 +117,38 @@ def set_hostname(self, hostname):
...
+    def get_nm_controlled(self):
+        ifname = self.get_if_name()
+        filepath = "/etc/sysconfig/network-scripts/ifcfg-{0}".format(ifname)
+        nm_controlled_cmd = ['grep', 'NM_CONTROLLED=', filepath]
+        try:
+            result = shellutil.run_command(nm_controlled_cmd, log_error=False).rstrip()
+
+            if result and len(result.split('=')) > 1:
+                value = result.split('=')[1]
+                if "n" in value:
```
nagworld9, line 129: I feel this is a loose evaluation. what are the possible values? May be value start with 'n' or direct matching something like that would be a strict check

maddieford: could be n, no, y, yes, 'n', 'no', 'y', 'yes', "n", "no", "y", "yes"

maddieford: I can remove any ' or " characters and check for n or no

nagworld9: we need to be careful about trailing spaces too when we remove it. If it's complicated, then I'm ok as is.

maddieford: Updated and tested stricter check

nagworld9 (approved):

azurelinuxagent/common/osutil/redhat.py
```diff
@@ -122,11 +122,12 @@ def get_nm_controlled(self):
         filepath = "/etc/sysconfig/network-scripts/ifcfg-{0}".format(ifname)
         nm_controlled_cmd = ['grep', 'NM_CONTROLLED=', filepath]
         try:
-            result = shellutil.run_command(nm_controlled_cmd, log_error=False).rstrip()
+            result = shellutil.run_command(nm_controlled_cmd, log_error=False, encode_output=False).rstrip()
```
narrieta, line 125: why is encide_output needed?

maddieford: If encode_output is True the stdout is returned as a string, otherwise it is returned as a bytes object.

When True, the result is unicode (u'no'). It was easier to do the comparison if encode_output=False and the result is a string

narrieta (approved):

PR #3009, opened 2024-01-05
Add distros which use Python 2.6 (for reference only)

narrieta:
Added the URN (and defined an image set) for some distros that use Python 2.6.

They are not included in the daily runs, but this facilitates some dev testing and is also useful for reference.

### DIFF
--- a/tests_e2e/test_suites/images.yml
+++ b/tests_e2e/test_suites/images.yml
@@ -5,9 +5,8 @@ image-sets:
    # Endorsed distros that are tested on the daily runs
    endorsed:
 #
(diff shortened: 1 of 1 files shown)


tests_e2e/test_suites/images.yml
```diff
@@ -142,6 +142,13 @@ images:
       locations:
          AzureChinaCloud: []
          AzureUSGovernment: []
+   rocky_9:
```
narrieta, line 145: moved these to keep alphabetical order

nagworld9 (approved):

PR #3012, opened 2024-01-08
Fix hostname initialization to take hostame from the correct file

maddieford:
## Description
The agent determines the hostname at service initialization using a hostname file. If provisioning is done by the agent, the hostname will be in the published hostname record (/var/lib/waagent/published_hostname). If provisioning is done by cloud-init, the hostname will be in set-hostname (/var/lib/cloud/data/set-hostname). At every hostname update, the agent will write the updated hostname to the published hostname record (/var/lib/waagent/published_hostname)

Currently, at service initialization the agent looks for the presence of published_hostname. If it exists, then it is assumed provisioning was done by the agent, and the hostname in published_hostname should be taken. If it does not exist, then it is assumed cloud-init did provisioning, and the hostname in set-hostname should be taken. This may cause an issue when there is a stale published_hostname file on the machine at service initialization, but provisioning was done by cloud-init. 

This PR updates the logic to determine the hostname according to which service did provisioning, regardless of if there is a stale file on the machine. It also changes the agent to keep the most updated hostname in set-hostname (/var/lib/cloud/data/set-hostname), if the file exists. This change was made because the agent gets the hostname at every service initialization. So if set-hostname is not updated with the current hostname, the agent will revert to the original hostname at provisioning during service initialization.

Issue # 

---

### PR information


azurelinuxagent/common/osutil/default.py
```diff
@@ -1313,28 +1313,75 @@ def device_for_ide_port(self, port_id):
 
     def set_hostname_record(self, hostname):
         fileutil.write_file(conf.get_published_hostname(), contents=hostname)
+        # If cloud init hostname file exists, update the file with the hostname. During service initialization, if
+        # provisioning was done by cloud-init the hostname will be set to whatever value is in the cloud init hostname
+        # file, so we should keep it updated with each hostname change.
+        if os.path.exists(self._get_cloud_init_hostname_path()):
+            self._set_cloud_init_hostname(hostname)
```
maddieford, line 1320: Keep the most updated hostname in set-hostname (/var/lib/cloud/data/set-hostname), if the file exists. 

This change was made because the agent gets the hostname at every service initialization. So if set-hostname is not updated with the current hostname, the agent will revert to the original hostname at provisioning during service initialization.

azurelinuxagent/pa/provision/factory.py
```diff
@@ -23,15 +23,23 @@
 from .default import ProvisionHandler
 from .cloudinit import CloudInitProvisionHandler, cloud_init_is_enabled
 
-def get_provision_handler(distro_name=DISTRO_NAME,  # pylint: disable=W0613
-                            distro_version=DISTRO_VERSION,  # pylint: disable=W0613
-                            distro_full_name=DISTRO_FULL_NAME):  # pylint: disable=W0613
 
+def cloud_init_should_provision():
```
maddieford, line 27: I have concerns about changing files in azurelinuxagent/pa

Thoughts on if it would be better to copy this logic somewhere else, and leave this file as is?

azurelinuxagent/common/osutil/default.py
narrieta: we shouldn't modify this file, since it was not created by the agent

maddieford:
This solution does not protect against the case where set-hostname is also stale. 

Set-hostname should not be modified since it is not created by the agent. But the only other way to protect against stale files is to create another state file, which could also be stale  :) Moving this fix to the daemon ->https://github.com/Azure/WALinuxAgent/pull/3016

PR #3014, opened 2024-01-09
Do not reset the mode of a extension's log directory

narrieta:
Currently, each time an extension is executed, the Agent resets the mode of its log directory. This is causing issues for customers that set custom modes on those directories.

With this fix, the Agent still sets the initial mode of the directory, but does not reset it when the extension is executed again.

### DIFF
--- a/azurelinuxagent/common/utils/fileutil.py
+++ b/azurelinuxagent/common/utils/fileutil.py
@@ -99,13 +99,15 @@ def get_line_startingwith(prefix, filepath):
     return None
 
 
-def mkdir(dirpath, mode=None, owner=None):
+def mkdir(dirpath, mode=None, owner=None, reset_mode_and_owner=True):
     if not os.path.isdir(dirpath):
         os.makedirs(dirpath)
-    if mode is not None:
-        chmod(dirpath, mode)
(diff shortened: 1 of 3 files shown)


nagworld9:
@narrieta Why don't we set the mode to 700 to protect from non-root users when we create it first, that's what we have it for waagent directory. Anyway now we are allowing users to modify it later and not doing reset on that, that means we should be ok with 700 too.

narrieta:
@nagworld9 Sorry, I don't get your question. This does not change the mode we currently use for logs. This just does not reset it every time the extension runs.

nagworld9:
@narrieta My question is not about this pr change. More of improvement perspective, when we create first time, we set to 755(existing code). I wonder why don't we create with 700 mode? what was the intention for 755? how this is different from waagent directory mode? I believe this ext directory only used by agent/extensions and they run with root. So, root users will have full access and won't be an issue.

narrieta:
@nagworld9 

`My question is not about this pr change
`

That should probably be followed up outside this PR, then.

maddieford (approved):

nagworld9 (approved):

PR #3015, opened 2024-01-09
honor rsm update with no time when agent receives new GS

nagworld9:
## Description
The current code has an issue if previous updater is self-update and waiting for its manifest download timer(1hr) to finish to continue with rest of the update steps (that includes switching to rsm update).  That is blocking rsm update and eventually update timeout in CRP.

Moving this timer condition after we check for updater switch.

Now, in main loop we _may_ log agent family missing warnings on every iteration. So, to avoid frequent logging, added a condition to log only on new goal state.

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/common/exception.py
+++ b/azurelinuxagent/common/exception.py
@@ -80,8 +80,9 @@ class AgentUpdateError(AgentError):
     When agent failed to update.
     """
 
-    def __init__(self, msg=None, inner=None):
+    def __init__(self, msg=None, inner=None, log_error=True):
         super(AgentUpdateError, self).__init__(msg, inner)
+        self.log_error = log_error
 
 
 class CGroupsException(AgentError):
--- a/azurelinuxagent/common/exception.py
+++ b/azurelinuxagent/common/exception.py
@@ -80,9 +80,17 @@ class AgentUpdateError(AgentError):
     When agent failed to update.
     """
 
-    def __init__(self, msg=None, inner=None, log_error=True):
+    def __init__(self, msg=None, inner=None):
         super(AgentUpdateError, self).__init__(msg, inner)
-        self.log_error = log_error
+
+
+class AgentFamilyMissingError(AgentError):
+    """
+    When agent family is missing.
+    """
+
+    def __init__(self, msg=None, inner=None):
+        super(AgentFamilyMissingError, self).__init__(msg, inner)
 
 
 class CGroupsException(AgentError):
--- a/azurelinuxagent/ga/agent_update_handler.py
+++ b/azurelinuxagent/ga/agent_update_handler.py
@@ -151,8 +151,8 @@ class AgentUpdateHandler(object):
             agent_family = self._get_agent_family_manifest(goal_state)
 
             # Updater will return True or False if we need to switch the updater
-            # If self-update receives RSM update request, it will switch to RSM updater
-            # If RSM updater stopped receiving RSM request, it will switch to self-update
+            # If self-update receives RSM update enabled, it will switch to RSM updater
+            # If RSM updater receives RSM update disabled, it will switch to self-update
             # No change in updater if GS not updated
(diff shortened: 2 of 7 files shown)


azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -145,30 +143,34 @@ def run(self, goal_state, ext_gs_updated):
...
-            # updater will return RSM enabled or disabled if we need to switch to self-update or rsm update
-            updater_mode = self._updater.check_and_switch_updater_if_changed(agent_family, self._gs_id, ext_gs_updated)
+            # Updater will return True or False if we need to switch the updater
+            # If self-update receives RSM update request, it will switch to RSM updater
+            # If RSM updater stopped receiving RSM request, it will switch to self-update
+            # No change in updater if GS not updated
+            switch_to_rsm_updater = self._updater.is_gs_requested_rsm_update(agent_family, self._gs_id, ext_gs_updated)
 
-            if updater_mode == RSMUpdates.Disabled:
+            if not switch_to_rsm_updater and isinstance(self._updater, RSMVersionUpdater):
```
maddieford, line 155: This check confused me at first because the variable name says 'switch'. The value could be true, but we're not doing a switch because the updater already is RSMUpdater. I think 'is_RSM_update' is clearer/more readable

azurelinuxagent/common/exception.py
```diff
@@ -80,8 +80,9 @@ class AgentUpdateError(AgentError):
     When agent failed to update.
     """
 
-    def __init__(self, msg=None, inner=None):
+    def __init__(self, msg=None, inner=None, log_error=True):
         super(AgentUpdateError, self).__init__(msg, inner)
+        self.log_error = log_error
```
narrieta, line 85: this log_error property is really weird:

AgentUpdateHandler.run has the 'ext_gs_updated' flag, which it passes to _get_agent_family_manifest, which passes it to AgentUpdateError with a different name, 'log_error', which is handled by AgentUpdateHandler.

Seems like AgentUpdateHandler could just use ext_gs_updated directlry

azurelinuxagent/ga/agent_update_handler.py
narrieta: I also find this code confusing, but the other way around: if one looks at the documentation for this method, it begins with "Checks if there is a new goal state and decide if we need to continue with rsm update or switch to self-update.", which makes me think both the method and the variable should be renamed to something like "should_switch_to_self_update", but then one continues reading the documentation and the implementation of the method and it looks like it is whether the machine is enabled for RSM updates, which seems to be confirmed by the message we log just in one line below. In that case, probably the method and variable should be renamed to "is_rsm_update_enabled".

Another issue with that method is that it looks like a query, just a check on state, but it actually updates the state (self._gs_id = gs_id) as a side effect.

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -145,30 +143,38 @@ def run(self, goal_state, ext_gs_updated):
...
-            # verify if agent update is allowed this time (RSM checks new goal state; self-update checks manifest download interval)
-            if not self._updater.is_update_allowed_this_time(ext_gs_updated):
-                return
+            # Update the state only on new goal state
+            if ext_gs_updated:
+                self._gs_id = goal_state.extensions_goal_state.id
+                self._updater.sync_new_gs_id(self._gs_id)
 
-            self._gs_id = goal_state.extensions_goal_state.id
             agent_family = self._get_agent_family_manifest(goal_state)
```
nagworld9, line 151: I still decide to read properties from goal state on every iteration as goal state is in memory and don't see a big deal. 

Having to call only on new goal state will need to save the state of agent family and also need to maintain state of last attempt has agent family missing error, so that we don't procced with update. I don't like the idea of saving extra state, at some time it will become convoluted.

So I raise different exception and handling it in exception block.

narrieta (approved):

maddieford (approved):

PR #3022, opened 2024-01-12
Don't check Agent log from the top after each test suite

narrieta:
The AgentTestSuite was checking the Agent log from the top every time a suite completed. With that approach, errors already handled by a suite would be reported by suites that execute after it.

Now, the AgentTestSuite keeps track of the timestamp of the sections of the log that it has already checked and restarts from that point on the next check.

### DIFF
--- a/tests_e2e/orchestrator/lib/agent_test_suite.py
+++ b/tests_e2e/orchestrator/lib/agent_test_suite.py
@@ -621,6 +621,7 @@ class AgentTestSuite(LisaTestSuite):
         suite_name = suite.name
         suite_full_name = f"{suite_name}-{self._environment_name}"
         suite_start_time: datetime.datetime = datetime.datetime.now()
+        check_log_start_time_override = datetime.datetime.max  # tests can override the timestamp for the agent log check with the get_ignore_errors_before_timestamp() method
 
         with set_thread_name(suite_full_name):  # The thread name is added to the LISA log
             log_path: Path = self._log_path / f"{suite_full_name}.log"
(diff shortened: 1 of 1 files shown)


tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -759,22 +760,15 @@ def _check_agent_log_on_test_nodes(self, ignore_error_rules: List[Dict[str, Any]
...
-                log.info("Checking agent log on the test node %s", node_name)
+                message = f"Checking agent log on test node {node_name}, starting at {check_log_start_time.strftime('%Y-%m-%dT%H:%M:%S.%fZ')}"
+                self._lisa_log.info(message)
+                log.info(message)
 
                 output = ssh_client.run_command("check-agent-log.py -j")
                 errors = json.loads(output, object_hook=AgentLogRecord.from_dictionary)
 
-                # Individual tests may have rules to ignore known errors; filter those out
-                if len(ignore_error_rules) > 0:
```
narrieta, line 769: Previously the check on the timestamp was being done only if the test has ignore rules. That seems more a coding mistake than the intended design. The new code fixes this.

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -705,13 +707,11 @@ def _execute_test_suite(self, suite: TestSuiteInfo, test_context: AgentTestConte
...
-                        if test_instance.get_ignore_errors_before_timestamp() != datetime.datetime.min:
-                            if before_timestamp != datetime.datetime.min:
-                                before_timestamp = min(before_timestamp, test_instance.get_ignore_errors_before_timestamp())
-                            else:
-                                before_timestamp = test_instance.get_ignore_errors_before_timestamp()
+                        # Check if the test is requesting to override the timestamp for the agent log check.
+                        # Note that if multiple tests in the suite provide an override, we'll use the earliest timestamp.
+                        test_check_log_start_time = test_instance.get_ignore_errors_before_timestamp()
+                        if test_check_log_start_time != datetime.datetime.min:
+                            check_log_start_time = min(check_log_start_time, test_check_log_start_time)
```
nagworld9, line 714: This always be datetime.min as check_log_start_time set to min

narrieta: check_log_start_time is updated on line 581

nagworld9: But that's after we complete first test suite. what if I have override timestamp on the first suite it executes

narrieta: ah, i see what you mean. thanks, let me fix it

nagworld9 (approved):

maddieford (approved):

PR #3024, opened 2024-01-18
Recover primary nic if down after publishing hostname in RedhatOSUtil

maddieford:
## Description
Primary network interface state is at /sys/class/net/eth0/operstate. If state is 'down' after we publish hostname, perform a NetworkManager restart to recover the interface

Issue # 

---

### PR information


azurelinuxagent/common/osutil/redhat.py
```diff
@@ -140,6 +139,30 @@ def get_nm_controlled(self):
 
         return True
 
+    def check_and_recover_nic_state(self, ifname):
+        filepath = "/sys/class/net/{0}/operstate".format(ifname)
+        try:
+            if not os.path.isfile(filepath):
+                logger.warn("Unable to determine primary network interface state, because state file does not exist: "
+                            "{0}".format(filepath))
```
narrieta, line 147: we should send telemetry about those errors so we can fix them, should they ever happen

azurelinuxagent/common/osutil/redhat.py
```diff
@@ -140,6 +139,30 @@ def get_nm_controlled(self):
...
+            else:
+                content = fileutil.read_file(filepath).rstrip().encode('utf-8')
```
narrieta, line 149: why is encode needed here? other instances of file_util.read_file on system files do not use it

azurelinuxagent/common/osutil/redhat.py
```diff
@@ -140,6 +139,30 @@ def get_nm_controlled(self):
...
+                down_states = ["down", "unknown"]
```
narrieta, line 150: when do we see 'unknown'?

narrieta (commented):
Two more comments: we should do this on all distros where we reset the network, and we should have an end-to-end test for the recover logic.

azurelinuxagent/common/osutil/redhat.py
```diff
@@ -140,6 +139,30 @@ def get_nm_controlled(self):
...
+                if content in down_states:
+                    logger.warn("The primary network interface {0} is in '{1}' state after being restarted. Restarting "
+                                "the Network Manager service to recover the network interface".format(ifname, content))
+                    self.restart_network_manager()
```
narrieta, line 154: after restarting the nm, we should log/send telemetry the success/failure of the attempt to bring the iface up

azurelinuxagent/common/osutil/redhat.py
maddieford: From kernel.org documentation:

> Interface is in unknown state, neither driver nor userspace has set
>  operational state. Interface must be considered for user data as
>  setting operational state has not been implemented in every driver.
 
 If driver isn't setting operation state, a restart of the NM is unlikely to fix this. I think I should remove 'unknown' from this list of states. We can log a warning when interface is in unknown state

azurelinuxagent/common/osutil/redhat.py
maddieford: Just tested and it's not needed. Will remove

azurelinuxagent/common/osutil/redhat.py
maddieford: Updated this. In my testing I saw that it takes around 1 second after restarting the NM for the interface to come up, so I added a 1 second sleep between NM restart and checking if the attempt succeeded

azurelinuxagent/common/osutil/redhat.py
```diff
@@ -140,17 +139,83 @@ def get_nm_controlled(self):
...
+            # network interface state.
+            time.sleep(1)
+            nic_operstate, nic_general_state = self.get_nic_operational_and_general_states(ifname)
+            if nic_operstate != "up" or nic_general_state != "100 (connected)":
+                msg = "Network Manager restart failed to bring network interface {0} into 'up' and 'connected' state".format(ifname)
+                logger.warn(msg)
+                raise Exception(msg)
+            else:
+                logger.info("Network Manager restart successfully brought the network interface {0} into 'up' and 'connected' state".format(ifname))
+        elif nic_operstate != "up" or nic_general_state != "100 (connected)":
```
nagworld9, line 193: when do we hit this condition? is not nic_operstate != "up" same as "down"?

azurelinuxagent/common/osutil/default.py
```diff
@@ -1190,11 +1190,20 @@ def restart_if(self, ifname, retries=3, wait=5):
             else:
                 logger.warn("exceeded restart retries")
 
-    def publish_hostname(self, hostname):
+    def check_and_recover_nic_state(self, ifname):
+        # TODO: This should be implemented for all distros where we reset the network during publishing hostname. Currently it is only implemented in RedhatOSUtil.
+        pass
+
+    def publish_hostname(self, hostname, recover_nic=False):
```
nagworld9, line 1197: when do we not need recover_nic as we default to false?

azurelinuxagent/common/osutil/redhat.py
```diff
@@ -140,17 +139,83 @@ def get_nm_controlled(self):
...
         return True
 
-    def publish_hostname(self, hostname):
+    def get_nic_operational_and_general_states(self, ifname):
+        """
+        Checks the contents of /sys/class/net/{ifname}/operstate and the results of 'nmcli -g general.state device show {ifname}' to determine the state of the provided interface.
+        Raises an exception if the network interface state cannot be determined.
+        """
+        filepath = "/sys/class/net/{0}/operstate".format(ifname)
+        nic_general_state_cmd = ['nmcli', '-g', 'general.state', 'device', 'show', ifname]
```
nagworld9, line 148: are these expected to be there and work in all versions?

tests_e2e/tests/recover_network_interface/recover_network_interface.py
```diff
@@ -0,0 +1,124 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#
+# This test uses CSE to bring the network down and call check_and_recover_nic_state to bring the network back into an
+# 'up' and 'connected' state. The intention of the test is to alert us if there is some change in newer distros which
```
nagworld9, line 22: how do we know change in new distros as we are validating on fixed distros.

tests_e2e/tests/recover_network_interface/recover_network_interface.py
```diff
@@ -0,0 +1,124 @@
...
+        assert_that(instance_view.statuses[0].message).described_as("The primary network interface should be in a disconnected state before the attempt to recover").contains("Primary network interface state before recovering: 30 (disconnected)")
+        assert_that(instance_view.statuses[0].message).described_as("The primary network interface should be in a connected state after the attempt to recover").contains("Primary network interface state after recovering: 100 (connected)")
+
+        # Check that the agent is successfully reporting status after recovering the network
+        log.info("")
+        log.info("Checking that the agent is reporting status after recovering the network...")
+        self.check_agent_reports_status()
+
+        log.info("")
+        log.info("The primary network was successfully recovered by the agent.")
```
nagworld9, line 120: nit: primary network interface

azurelinuxagent/common/osutil/default.py
maddieford: during provisioning we call publish_hostname. We don't want to change the behavior during provisioning, so making the default false

azurelinuxagent/common/osutil/redhat.py
maddieford: documentation says there is an "unknown" state:

> Interface is in unknown state, neither driver nor userspace has set
> operational state. Interface must be considered for user data as
> setting operational state has not been implemented in every driver.

If driver isn't setting operation state, a restart of the NM is unlikely to fix this, so we are just reporting this state as a warning

tests_e2e/tests/recover_network_interface/recover_network_interface.py
maddieford: Eventually we want to attempt to recover the network interface for all distros where we restart the network interface, but this PR only implements the recovery for RedhatOSUtil (all fedora distros > 7 and < 8). 

There is a TODO in the yml for this scenario which says we should expand coverage to other distros once implemented, and then this comment will be accurate :)

azurelinuxagent/common/osutil/redhat.py
maddieford: Yes, nmcli is the NetworkManager command line management tool for fedora based systems: https://fedoraproject.org/wiki/Tools/NetworkManager/CLI

If for some unexpected reason this command returns a non zero exit code, a CommandError will be raised and we'll raise it to the MonitorHostNameChanges operation which will report it to telemetry.

azurelinuxagent/common/osutil/gaia.py
```diff
@@ -202,7 +202,7 @@ def set_hostname(self, hostname):
     def set_dhcp_hostname(self, hostname):
         logger.warn('set_dhcp_hostname is ignored on GAiA')
 
-    def publish_hostname(self, hostname):
+    def publish_hostname(self, hostname, recover_nic=False):
```
narrieta, line 205: why do we set to False explicitly?

maddieford: We don't want to have to change the call to publish_hostname in provisioning. If we don't set recover_nic to False explicitly, then we'll need to add another argument to the call in provisioning

azurelinuxagent/common/osutil/redhat.py
```diff
@@ -140,17 +139,83 @@ def get_nm_controlled(self):
...
+        if not os.path.isfile(filepath):
+            msg = "Unable to determine primary network interface {0} state, because state file does not exist: {1}".format(ifname, filepath)
+            logger.warn(msg)
+            raise Exception(msg)
+
+        try:
+            nic_operstate = fileutil.read_file(filepath).rstrip().lower()
```
narrieta, line 155: nit: rename to nic_oper_state for consistency with nic_general_state?

azurelinuxagent/common/osutil/redhat.py
```diff
@@ -140,17 +139,83 @@ def get_nm_controlled(self):
...
+            nic_general_state = shellutil.run_command(nic_general_state_cmd, log_error=True).rstrip().lower()
```
narrieta, line 156: do we need to log the error here? it's also logged at line 168 as a WARN

azurelinuxagent/common/osutil/gaia.py
narrieta: never mind, this is an override so you need to make it consistent with the definition in the base class

azurelinuxagent/common/osutil/redhat.py
```diff
@@ -140,17 +139,83 @@ def get_nm_controlled(self):
...
+        Raises an exception if an attempt to bring the interface into an 'up' state fails, or if the state
+         of the network interface cannot be determined.
+        """
+        nic_operstate, nic_general_state = self.get_nic_operational_and_general_states(ifname)
+        if nic_operstate == "down" or "disconnected" in nic_general_state:
+            logger.info("Restarting the Network Manager service to recover network interface {0}".format(ifname))
+            self.restart_network_manager()
+            # Interface does not come up immediately after NetworkManager restart. Wait 1 second before checking
+            # network interface state.
+            time.sleep(1)
```
narrieta, line 185: I wonder if this should be a little longer (3? 5?). Does the value usually come up in ms or significantly less than 1 sec? it may be worth waiting more, since this is done rarely and this is happening in the env thread, which doesn't block anything significant

azurelinuxagent/common/osutil/redhat.py
```diff
@@ -140,17 +139,83 @@ def get_nm_controlled(self):
...
+    def check_and_recover_nic_state(self, ifname):
+        """
+        Checks if the provided network interface is in an 'up' state. If the network interface is in a 'down' state,
+        attempt to recover the interface by restarting the Network Manager service.
+
+        Raises an exception if an attempt to bring the interface into an 'up' state fails, or if the state
+         of the network interface cannot be determined.
+        """
+        nic_operstate, nic_general_state = self.get_nic_operational_and_general_states(ifname)
+        if nic_operstate == "down" or "disconnected" in nic_general_state:
```
narrieta, line 180: why the difference in the conditions checked here and on line 187?

narrieta: ok, i see a similar question was asked below... maybe add your reply as a comment in the code?

tests_e2e/tests/recover_network_interface/recover_network_interface.py
```diff
@@ -0,0 +1,124 @@
...
+                 "to debug".format(last_agent_status_time))
+
+    def run(self):
+        # Add password to VM and log. This allows us to debug with serial console if necessary
+        log.info("")
+        log.info("Adding password to the VM to use for debugging in case necessary...")
+        self.add_vm_password()
+
+        # Get the primary network interface name
+        ifname = self._ssh_client.run_command("pypy3 -c 'from azurelinuxagent.common.osutil.redhat import RedhatOSUtil; print(RedhatOSUtil().get_if_name())'").rstrip()
```
narrieta, line 81: there is no need to start another instance of pypy3, you could simple import the Agent's module from this test.

tests_e2e/tests/recover_network_interface/recover_network_interface.py
```diff
@@ -0,0 +1,124 @@
...
+        # The interface name needs to be in double quotes for the pypy portion of the script
+        formatted_ifname = f'"{ifname}"'
+
+        # The script should bring the primary network interface down and use the agent to recover the interface. These
+        # commands will bring the network down, so they should be executed on the machine using CSE instead of ssh.
+        script = f"""
```
narrieta, line 87: let's add "set -euxo pipefail" to the script

azurelinuxagent/common/osutil/redhat.py
```diff
@@ -117,12 +117,11 @@ def set_hostname(self, hostname):
...
             DefaultOSUtil.set_hostname(self, hostname)
 
-    def get_nm_controlled(self):
-        ifname = self.get_if_name()
+    def get_nm_controlled(self, ifname):
         filepath = "/etc/sysconfig/network-scripts/ifcfg-{0}".format(ifname)
         nm_controlled_cmd = ['grep', 'NM_CONTROLLED=', filepath]
         try:
-            result = shellutil.run_command(nm_controlled_cmd, log_error=False, encode_output=False).rstrip()
+            result = shellutil.run_command(nm_controlled_cmd, log_error=False).rstrip()
```
maddieford, line 124: I did testing and encode_output=False is not necessary, so I removed it

azurelinuxagent/common/osutil/redhat.py
maddieford: removed

azurelinuxagent/common/osutil/redhat.py
maddieford: Added comment

azurelinuxagent/common/osutil/redhat.py
maddieford: Added comment for clarification on 187

tests_e2e/tests/recover_network_interface/recover_network_interface.py
maddieford: Added

narrieta (approved):

narrieta (approved):

nagworld9 (approved):

PR #3020, opened 2024-01-12
Adding AutoUpdate.UpdateToLatestVersion new flag support

nagworld9:
## Description

Issue # 

---

### PR information
- [ ] The title of the PR is clear and informative.
- [ ] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).
- [ ] If applicable, the PR references the bug/issue that it fixes in the description.
- [ ] New Unit tests were added for the changes made

### Quality of Code and Contribution Guidelines
- [ ] I have read the [contribution guidelines](https://github.com/Azure/WALinuxAgent/blob/master/.github/CONTRIBUTING.md).

PR #3027, opened 2024-01-19
New conf flag AutoUpdate.UpdateToLatestVersion support

nagworld9:
## Description

We have an ask from TEMU customer to update the agents at their convivence time. Current controlling with existing flag AutoUpdate.Enabled is not working how they wanted it. When you turn off, it will revert to preinstalled agent. But customer wanted to stay on latest version installed on the vm. The current flag should have been implemented that way, unfortunately it wasn't. We can't change that flag anymore as users might have taken dependency on the existing behavior.

New behavior was added as hidden feature for e2e testing. Now changing that to customer facing flag.

Also added e2e test for new flag behavior.

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/common/conf.py
+++ b/azurelinuxagent/common/conf.py
@@ -128,6 +128,7 @@ __SWITCH_OPTIONS__ = {
     "ResourceDisk.EnableSwap": False,
     "ResourceDisk.EnableSwapEncryption": False,
     "AutoUpdate.Enabled": True,
+    "AutoUpdate.UpdateToLatestVersion": True,
     "EnableOverProvisioning": True,
     #
     # "Debug" options are experimental and may be removed in later
@@ -136,7 +137,6 @@ __SWITCH_OPTIONS__ = {
     "Debug.CgroupLogMetrics": False,
     "Debug.CgroupDisableOnProcessCheckFailure": True,
     "Debug.CgroupDisableOnQuotaCheckFailure": True,
-    "Debug.DownloadNewAgents": True,
     "Debug.EnableAgentMemoryUsageCheck": False,
     "Debug.EnableFastTrack": True,
     "Debug.EnableGAVersioning": True
@@ -503,15 +503,14 @@ def get_monitor_network_configuration_changes(conf=__conf__):
     return conf.get_switch("Monitor.NetworkConfigurationChanges", False)
 
 
-def get_download_new_agents(conf=__conf__):
+def get_agent_update_to_latest_version(conf=__conf__):
     """
-    If True, the agent go through update logic to look for new agents to download otherwise it will stop agent updates.
-    NOTE: AutoUpdate.Enabled controls whether the Agent downloads new update and also whether any downloaded updates are started or not, while DownloadNewAgents controls only the former.
-    AutoUpdate.Enabled == false -> Agent preinstalled on the image will process extensions and will not update (regardless of DownloadNewAgents flag)
-    AutoUpdate.Enabled == true and DownloadNewAgents == true, any update already downloaded will be started, and agent look for future updates
-    AutoUpdate.Enabled == true and DownloadNewAgents == false, any update already downloaded will be started, but the agent will not look for future updates
+    If set to True, agent will update to the latest version
+    NOTE:
+        when turned on, both AutoEnabled.Enabled and UpdateToLatestVersion same meaning: update to latest version
+        when turned off, AutoEnabled.Enabled: reverts to pre-installed agent, UpdateToLatestVersion: uses latest version installed on the vm
     """
-    return conf.get_switch("Debug.DownloadNewAgents", True)
+    return conf.get_switch("AutoUpdate.UpdateToLatestVersion", True)
 
 
 def get_cgroup_check_period(conf=__conf__):
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -752,6 +752,7 @@ class UpdateHandler(object):
             log_if_op_disabled("OS.EnableFirewall", conf.enable_firewall())
             log_if_op_disabled("Extensions.Enabled", conf.get_extensions_enabled())
             log_if_op_disabled("AutoUpdate.Enabled", conf.get_autoupdate_enabled())
+            log_if_op_disabled("AutoUpdate.UpdateToLatestVersion", conf.get_agent_update_to_latest_version())
 
             if conf.enable_firewall():
                 log_if_int_changed_from_default("OS.EnableFirewallPeriod", conf.get_enable_firewall_period())
--- a/config/alpine/waagent.conf
+++ b/config/alpine/waagent.conf
@@ -75,7 +75,14 @@ OS.OpensslPath=None
 OS.SshDir=/etc/ssh
 
 # Enable or disable goal state processing auto-update, default is enabled
+# When turned off, it reverts to the pre-installed agent that comes with image
+# Added new option AutoUpdate.UpdateToLatestVersion, and encourage users to transition to this new option
+# However, this flag is retained for legacy reasons
 # AutoUpdate.Enabled=y
+
+# Enable or disable goal state processing auto-update, default is enabled
+# When turned off, it remains on latest version installed on the vm
+# AutoUpdate.UpdateToLatestVersion=y
 
 # Determine the update family, this should not be changed
 # AutoUpdate.GAFamily=Prod
--- a/tests/test_agent.py
+++ b/tests/test_agent.py
@@ -28,6 +28,7 @@ from tests.lib.tools import AgentTestCase, data_dir, Mock, patch
 EXPECTED_CONFIGURATION = \
 """AutoUpdate.Enabled = True
 AutoUpdate.GAFamily = Prod
+AutoUpdate.UpdateToLatestVersion = True
 Autoupdate.Frequency = 3600
 DVD.MountPoint = /mnt/cdrom/secure
 Debug.AgentCpuQuota = 50
@@ -41,7 +42,6 @@ Debug.CgroupDisableOnQuotaCheckFailure = True
 Debug.CgroupLogMetrics = False
 Debug.CgroupMonitorExpiryTime = 2022-03-31
 Debug.CgroupMonitorExtensionName = Microsoft.Azure.Monitor.AzureMonitorLinuxAgent
-Debug.DownloadNewAgents = True
 Debug.EnableAgentMemoryUsageCheck = False
 Debug.EnableFastTrack = True
 Debug.EnableGAVersioning = True
--- a/azurelinuxagent/common/agent_supported_feature.py
+++ b/azurelinuxagent/common/agent_supported_feature.py
@@ -76,7 +76,7 @@ class _ETPFeature(AgentSupportedFeature):
 
 class _GAVersioningGovernanceFeature(AgentSupportedFeature):
     """
-    CRP would drive the RSM upgrade version if agent reports that it does support RSM upgrades with this flag otherwise CRP fallback to largest version.
+    CRP would drive the RSM update if agent reports that it does support RSM upgrades with this flag otherwise CRP fallback to largest version.
     Agent doesn't report supported feature flag if auto update is disabled or old version of agent running that doesn't understand GA versioning.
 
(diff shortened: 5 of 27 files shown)


config/alpine/waagent.conf
```diff
@@ -75,8 +75,15 @@ OS.OpensslPath=None
 OS.SshDir=/etc/ssh
 
 # Enable or disable goal state processing auto-update, default is enabled
+# When turned off, it reverts to the pre-installed agent that comes with image
+# Added new option AutoUpdate.UpdateToLatestVersion, and encourage users to transition to this new option
```
maddieford, line 79: Once we know release #, i think it may be helpful to add which release added support for AutoUpdate.UpdateToLatestVersion. 

Also maybe we can mention why we encourage users to transition to this new option (for example, if you revert to pre-installed agent you lose many bug fixes, etc.)

azurelinuxagent/ga/update.py
```diff
@@ -752,6 +752,12 @@ def log_if_agent_versioning_feature_disabled():
             log_if_op_disabled("OS.EnableFirewall", conf.enable_firewall())
             log_if_op_disabled("Extensions.Enabled", conf.get_extensions_enabled())
             log_if_op_disabled("AutoUpdate.Enabled", conf.get_autoupdate_enabled())
+            log_if_op_disabled("AutoUpdate.UpdateToLatestVersion", conf.get_auto_update_to_latest_version())
+
+            if not conf.get_autoupdate_enabled():
+                msg = "AutoUpdate.Enabled property is **Deprecated** but it was used and set to False. Please switch to using AutoUpdate.UpdateToLatestVersion instead."
```
maddieford, line 758: Should we include wiki link here too? This message makes it sound like AutoUpdate.Enabled and AutoUpdate.UpdateToLatestVersion do the same thing

azurelinuxagent/common/conf.py
```diff
@@ -503,15 +503,18 @@ def get_monitor_network_configuration_changes(conf=__conf__):
...
+def get_auto_update_to_latest_version(conf=__conf__):
     """
-    If True, the agent go through update logic to look for new agents to download otherwise it will stop agent updates.
-    NOTE: AutoUpdate.Enabled controls whether the Agent downloads new update and also whether any downloaded updates are started or not, while DownloadNewAgents controls only the former.
-    AutoUpdate.Enabled == false -> Agent preinstalled on the image will process extensions and will not update (regardless of DownloadNewAgents flag)
-    AutoUpdate.Enabled == true and DownloadNewAgents == true, any update already downloaded will be started, and agent look for future updates
-    AutoUpdate.Enabled == true and DownloadNewAgents == false, any update already downloaded will be started, but the agent will not look for future updates
+    If set to True, agent will update to the latest version
+    NOTE:
+        when turned on, both AutoEnabled.Enabled and UpdateToLatestVersion same meaning: update to latest version
```
maddieford, line 510:
```suggestion
        when both turned on, both AutoEnabled.Enabled and UpdateToLatestVersion same meaning: update to latest version
```

azurelinuxagent/common/conf.py
```diff
@@ -503,15 +503,18 @@ def get_monitor_network_configuration_changes(conf=__conf__):
...
+        when turned off, AutoEnabled.Enabled: reverts to pre-installed agent, UpdateToLatestVersion: uses latest version installed on the vm
```
maddieford, line 511:
```suggestion
        when turned off, AutoEnabled.Enabled: reverts to pre-installed agent, UpdateToLatestVersion: uses latest version already installed on the vm and does not download new agents
```

azurelinuxagent/common/conf.py
```diff
@@ -503,15 +503,18 @@ def get_monitor_network_configuration_changes(conf=__conf__):
...
+        Since we are deprecating AutoEnabled.Enabled, we need to honor if existing flag explicitly set to 'n' for backward compatibility.
     """
-    return conf.get_switch("Debug.DownloadNewAgents", True)
+    auto_update_enabled = get_autoupdate_enabled()
+    if not auto_update_enabled:
```
maddieford, line 515: Does the wiki mention that if AutoUpdate.Enabled is 'n', then AutoUpdate.UpdateToLatestVersion is ignored? If not, I think it should

nagworld9: I added AutoUpdate.UpdateToLatestVersion applicable only if autoupdate set to 'y'. 

[maddieford](https://github.com/maddieford) [yesterday](https://github.com/Azure/WALinuxAgent/pull/3020#discussion_r1464202733)
Should we mention here it only applies when AutoUpdate.Enabled is 'y' OR not in the conf at all?

Member
@[maddieford](https://github.com/maddieford) maddieford [yesterday](https://github.com/Azure/WALinuxAgent/pull/3020#discussion_r1464202962)
Or alternatively, it is only applied if AutoUpdate.Enabled is not 'n'

maddieford: From this line it looks like if AutoUpdate.UpdateToLatestVersion is 'y' and AutoUpdate.Enabled is 'no' we return False here, is that correct?
In that case, I think we should add to the wiki that we ignore AutoUpdate.UpdateToLatestVersion if AutoUpdate.Enabled is 'n', even if AutoUpdate.UpdateToLatestVersion is 'y'

config/alpine/waagent.conf
```diff
@@ -75,7 +75,11 @@ OS.OpensslPath=None
 OS.SshDir=/etc/ssh
 
 # Enable or disable goal state processing auto-update, default is enabled
-# AutoUpdate.Enabled=y
+# When turned off, it remains on latest version installed on the vm
+# Added this new option AutoUpdate.UpdateToLatestVersion in place of AutoUpdate.Enabled, and encourage users to transition to this new option
+# See wiki[https://github.com/Azure/WALinuxAgent/wiki/FAQ#autoupdateenabled-vs-autoupdateupdatetolatestversion] for more details
+# TODO: Update the wiki link and point to readme page or public facing doc
```
nagworld9, line 81: Ideally, I want to point to readme link for flag definition but those changes not there in master yet

azurelinuxagent/common/conf.py
```diff
@@ -507,14 +507,15 @@ def get_auto_update_to_latest_version(conf=__conf__):
...
-    """
-    auto_update_enabled = get_autoupdate_enabled()
-    if not auto_update_enabled:
-        return auto_update_enabled
-    return conf.get_switch("AutoUpdate.UpdateToLatestVersion", True)
+        when both turned on, both AutoUpdate.Enabled and AutoUpdate.UpdateToLatestVersion same meaning: update to latest version
+        when turned off, AutoUpdate.Enabled: reverts to pre-installed agent, AutoUpdate.UpdateToLatestVersion: uses latest version already installed on the vm and does not download new agents
+        Even we are deprecating AutoUpdate.Enabled, we still need to support if users explicitly setting it instead new flag.
+        If AutoUpdate.UpdateToLatestVersion is present, it overrides any value set for AutoUpdate.Enabled (if present).
+        If AutoUpdate.Enabled is present and set to 'n', we adhere to AutoUpdate.Enabled flag's behavior
```
maddieford, line 514: Same comment here as other PR, lines 513 and 514 seem to contradict each other to me

tests/common/test_conf.py
```diff
@@ -63,6 +63,7 @@ class TestConf(AgentTestCase):
         "OS.CheckRdmaDriver": False,
         "AutoUpdate.Enabled": True,
         "AutoUpdate.GAFamily": "Prod",
+        "AutoUpdate.UpdateToLatestVersion": False,
```
narrieta, line 66: default should be true

tests/test_agent.py
```diff
@@ -28,6 +28,7 @@
 EXPECTED_CONFIGURATION = \
 """AutoUpdate.Enabled = True
 AutoUpdate.GAFamily = Prod
+AutoUpdate.UpdateToLatestVersion = False
```
narrieta, line 31: default should be true

azurelinuxagent/common/conf.py
```diff
@@ -503,15 +503,19 @@ def get_monitor_network_configuration_changes(conf=__conf__):
...
 
-def get_download_new_agents(conf=__conf__):
+def get_auto_update_to_latest_version(conf=__conf__):
     """
-    If True, the agent go through update logic to look for new agents to download otherwise it will stop agent updates.
-    NOTE: AutoUpdate.Enabled controls whether the Agent downloads new update and also whether any downloaded updates are started or not, while DownloadNewAgents controls only the former.
-    AutoUpdate.Enabled == false -> Agent preinstalled on the image will process extensions and will not update (regardless of DownloadNewAgents flag)
-    AutoUpdate.Enabled == true and DownloadNewAgents == true, any update already downloaded will be started, and agent look for future updates
-    AutoUpdate.Enabled == true and DownloadNewAgents == false, any update already downloaded will be started, but the agent will not look for future updates
+    If set to True, agent will update to the latest version
```
narrieta, line 508: we should add unit tests that codify these rules: use all combinations of y/n/absent for both flags and check the return value of UpdateToLatestVersion

nagworld9: addressed

tests/ga/test_update.py
```diff
@@ -686,12 +686,13 @@ def _test_run_latest(self, mock_child=None, mock_time=None, child_args=None):
     def test_run_latest(self):
         self.prepare_agents()
 
-        agent = self.update_handler.get_latest_agent_greater_than_daemon()
-        args, kwargs = self._test_run_latest()
-        args = args[0]
-        cmds = textutil.safe_shlex_split(agent.get_agent_cmd())
-        if cmds[0].lower() == "python":
-            cmds[0] = sys.executable
+        with patch("azurelinuxagent.common.conf.get_autoupdate_enabled", return_value=True):
```
nagworld9, line 689: This change as a result of removing global mock for autoupdate flag in AgentUpdateCase. See the comment in tools.py file

tests/ga/test_update.py
```diff
@@ -801,7 +802,8 @@ def test_run_latest_exception_blacklists(self):
         verify_string = "Force blacklisting: {0}".format(str(uuid.uuid4()))
 
         with patch('azurelinuxagent.ga.update.UpdateHandler.get_latest_agent_greater_than_daemon', return_value=latest_agent):
-            self._test_run_latest(mock_child=ChildMock(side_effect=Exception(verify_string)))
+            with patch("azurelinuxagent.common.conf.get_autoupdate_enabled", return_value=True):
```
nagworld9, line 805: Same thing, see other test comment

tests/lib/tools.py
```diff
@@ -181,7 +181,6 @@ def setUp(self):
         self.tmp_dir = tempfile.mkdtemp(prefix=prefix)
         self.test_file = 'test_file'
 
-        conf.get_autoupdate_enabled = Mock(return_value=True)
```
nagworld9, line 184: Removed global mock as tests are extended agenttest setup. This mock getting in the way for other tests especially conf tests which are importing the value from config file

azurelinuxagent/common/conf.py
```diff
@@ -228,6 +228,11 @@ def get_switch_default_value(option):
     raise ValueError("{0} is not a valid configuration parameter.".format(option))
 
 
+def get_switch_no_default_value(key, conf=__conf__):
+    # If specified, return the value else return None
+    return conf.get_switch(key, None)
+
```
narrieta, line 234: this function feels a little weird, doesn't it? 

Maybe

```
    def is_present(self, key):
        """
        Returns True if the given flag is present in the config file.
        """
        return self.values.get(key) is not None

```

tests/ga/test_update.py
narrieta: do we know why we had a global mock?

nagworld9: My theory after digging history of this change, this was added when agent update feature implemented first. That time they don't mock for update_handler and all of them extending agent setup they decided added to global setup. Now we have update_handler mock and the agent update is related to update_handler, I don't think we need a global mock and also causing problems for other tests like I mentioned before.

azurelinuxagent/common/conf.py
nagworld9: updated

narrieta (approved):

maddieford (approved):

PR #274, opened 2016-06-28
Improved download process and ensured output is not buffered

brendandixon:
Signed-off-by: Brendan Dixon <brendand@microsoft.com>

PR #3031, opened 2024-01-23
Add config parameter to wait for cloud-init (Extensions.WaitForCloudInit)

narrieta:
Adding Extensions.WaitForCloudInit to /etc/waagent.conf. See update in README.md for a description of this parameter.

Also, added some infrastructure to execute tests on custom images published to a shared gallery. Added an end-to-end test that can be executed manually from a custom image.


tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
```diff
@@ -220,18 +228,18 @@ def create_environment_list(self) -> List[Dict[str, Any]]:
                                 env_name=env_name,
                                 marketplace_image=marketplace_image,
                                 vhd=vhd,
+                                shared_gallery=shared_gallery,
                                 location=location,
                                 vm_size=vm_size,
                                 test_suite_info=test_suite_info)
                         shared_environments[env_name] = env
 
-                    if test_suite_info.template != '':
```
narrieta, line 228: We were adding vm_tags["templates"] twice: here and below @ line 342. As part of the fix, I ensured there is always an env["vm_tags"] so it is safe to use [] instead of get() plus check on None.

tests_e2e/tests/agent_not_provisioned/disable_agent_provisioning.py
```diff
@@ -62,31 +55,9 @@ def update(self, template: Dict[str, Any], is_lisa_template: bool) -> None:
         #         }
         #     }
         #
-        # The code below sets template['functions'][i]['members']['getLinuxConfiguration']['output']['value']['provisionVMAgent'] to True,
```
narrieta, line 65: This code got refactored to the base class (UpdateArmTemplate) since now it is used by multiple tests

azurelinuxagent/ga/update.py
```diff
@@ -458,6 +460,22 @@ def _initialize_goal_state(self, protocol):
                     logger.info("The current Fabric goal state is older than the most recent FastTrack goal state; will skip it.\nFabric:    {0}\nFastTrack: {1}",
                         egs.created_on_timestamp, last_fast_track_timestamp)
 
+    def _wait_for_cloud_init(self):
+        if conf.get_wait_for_cloud_init() and not self._cloud_init_completed:
+            message = "Waiting for cloud-init to complete..."
+            logger.info(message)
+            add_event(op=WALAEventOperation.CloudInit, message=message)
+            try:
+                output = shellutil.run_command(["cloud-init", "status", "--wait"], timeout=conf.get_wait_for_cloud_init_timeout())
```
nagworld9, line 469: how long we wait for python2 distros if they set to 'y'?

nagworld9, line 469: what is the command does?

README.md
```diff
@@ -261,6 +261,30 @@ without the agent. In order to do that, the `provisionVMAgent` flag must be set
...
+that conflict with each other (for example, both of them may try to install packages). Setting this option
+to 'y' ensures that VM extensions are executed only after cloud-init has completed all its stages.
+
+Note that using this option requires creating a custom image with the value of this option set to 'y', in
+order to ensure that the wait is performed during the initial deployment of the VM.
+
+#### __Extensions.WaitForCloudInitTimeout__
+
+_Type: Integer_  
+_Default: 3600_
```
nagworld9, line 283: waiting for 1 hr, isn't too long?

tests_e2e/tests/agent_wait_for_cloud_init/agent_wait_for_cloud_init.py
```diff
@@ -0,0 +1,93 @@
...
+        echo ">>> $(date) cloud-init script end" >> /var/log/waagent.log
+    """
+
+    def run(self):
+        ssh_client: SshClient = self._context.create_ssh_client()
+
+        log.info("Waiting for Agent to start processing extensions")
+        for _ in range(15):
+            try:
+                ssh_client.run_command("grep 'ProcessExtensionsGoalState started' /var/log/waagent.log")
```
nagworld9, line 57: we have retry function, can we use that?

azurelinuxagent/ga/update.py
```diff
@@ -458,6 +460,22 @@ def _initialize_goal_state(self, protocol):
                     logger.info("The current Fabric goal state is older than the most recent FastTrack goal state; will skip it.\nFabric:    {0}\nFastTrack: {1}",
                         egs.created_on_timestamp, last_fast_track_timestamp)
 
+    def _wait_for_cloud_init(self):
```
maddieford, line 463: On my PR for updating the daemon to clean up any stale files if VM image was not properly generalized, we discussed the possibility that agent and cloud-init will be racing (since the agent will detect that the vm is in a 'provisioned' state when it is not).

Could we add a call to _wait_for_cloud_init() in the daemon to protect against this?

README.md
narrieta: We are waiting for arbitrary configuration, provided by the user: install packages, setup services, or run arbitrary scripts.

I feel that 60 min is in line with the 90 min timeout for goal states, but what a good value is depends completely on what the user is doing.

This option needs to be explicitly set on a custom image; the user can decide the timeout at that time.

azurelinuxagent/common/utils/shellutil.py
```diff
@@ -246,7 +256,22 @@ def command_action():
...
+                command_stdout, command_stderr = process.communicate(input=communicate_input, timeout=timeout)
+        except TimeoutExpired:
+            if log_error:
+                logger.error(u"Command [{0}] timed out", __format_command(command))
+            try:
+                process.kill()
+            except Exception as exception:
+                if log_error:
+                    logger.error(u"Can't terminate timed out process: {0}", ustr(exception))
+            command_stdout, command_stderr = process.communicate()
```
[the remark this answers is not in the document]
narrieta: good point, i should treat it as there a possibility of it raising. will update the code

azurelinuxagent/ga/update.py
```diff
@@ -149,6 +149,8 @@ def __init__(self):
         self._last_check_memory_usage_time = time.time()
         self._check_memory_usage_last_error_report = datetime.min
 
+        self._cloud_init_completed = False  # Only used when Extensions.WaitForCloudInit is enabled; note that this variable is always reset on service start.
```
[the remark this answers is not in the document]
narrieta: once cloud-init completes the wait call returns immediately

azurelinuxagent/ga/update.py
narrieta: The race I was mentioning is against the Config stage, that is where the Azure module does the provisioning stuff. The wait I am adding is against the Final stage. PA does not wait for Final because there could be lengthy operations there that could affect TDPR (that is why the option I added needs to be explicitly set on a custom image).

Waiting for Config when 'provisioned' exists may be a good idea, but I would leave the decision of whether to implement it to PA.
```
https://cloudinit.readthedocs.io/en/latest/explanation/boot.html

There are five stages to boot:

1 Detect
2 Local
3 Network
4 Config
5 Final
```

azurelinuxagent/ga/update.py
narrieta: there is no timeout

azurelinuxagent/ga/update.py
```diff
@@ -562,6 +580,8 @@ def _process_goal_state(self, exthandlers_handler, remote_access_handler, agent_
         # check for agent updates
         agent_update_handler.run(self._goal_state, self._processing_new_extensions_goal_state())
 
+        self._wait_for_cloud_init()
```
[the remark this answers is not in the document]
narrieta: The wait I am adding is against the Final stage. PA does not wait for Final because there could be lengthy operations there that could affect TDPR (that is why the option I added needs to be explicitly set on a custom image).

```
https://cloudinit.readthedocs.io/en/latest/explanation/boot.html

There are five stages to boot:

1 Detect
2 Local
3 Network
4 Config
5 Final
```

tests_e2e/tests/agent_wait_for_cloud_init/agent_wait_for_cloud_init.py
narrieta: retry_if_false is too simplistic. I would need to write more code than this simple loop to make it work for the specific errors I am handling, with the specific error messages I am logging. It is not worth it.

azurelinuxagent/ga/update.py
narrieta: this waits for cloud-init to complete all its stages

azurelinuxagent/ga/update.py
maddieford: makes sense

tests_e2e/tests/agent_wait_for_cloud_init/agent_wait_for_cloud_init.py
```diff
@@ -0,0 +1,93 @@
...
+from tests_e2e.tests.lib.logging import log
+from tests_e2e.tests.lib.shell import CommandError
+from tests_e2e.tests.lib.ssh_client import SshClient
+
+
+class AgentWaitForCloudInit(AgentVmTest):
+    """
+    This test verifies that the Agent waits for cloud-init to complete before it starts processing extensions.
+
+    To do this, it adds 'CloudInitScript' in cloud-init's custom data. The script ensures first that the Agent
```
maddieford, line 34: Is this script run by cloud-init? How is cloud-init's custom data used

tests_e2e/tests/agent_wait_for_cloud_init/agent_wait_for_cloud_init.py
```diff
@@ -0,0 +1,93 @@
...
+            except CommandError:
+                log.info("The Agent has not started to process extensions, will check again after a short delay")
+                time.sleep(60)
+        else:
+            raise Exception("Timeout while waiting for the Agent to start processing extensions")
+
+        log.info("The Agent has started to process extensions")
+
+        output = ssh_client.run_command(
+            "grep -E 'INFO ExtHandler ExtHandler Waiting for cloud-init to complete|" +
```
maddieford, line 68: This doesn't check the timestamp of these logs to make sure they were logged before the 'ProcessExtensionsGoalState started' log. Since we sleep 60 seconds between attempts to check for extension processing log, it's possible that these get logged after extension processing starts.

azurelinuxagent/common/utils/shellutil.py
```diff
@@ -246,7 +256,26 @@ def command_action():
...
+                process.kill()
+                try:
+                    command_stdout, command_stderr = process.communicate()
+                # W0702: No exception type(s) specified (bare-except)
+                except:  # pylint: disable=W0702
+                    command_stdout, command_stderr = '', ''
+            except Exception as exception:
+                if log_error:
+                    logger.error(u"Can't terminate timed out process: {0}", ustr(exception))
+            raise CommandError(command=__format_command(command), return_code=-1, stdout=command_stdout, stderr="command timeout\n{0}".format(command_stderr))
```
nagworld9, line 277: I see command_stdout not defined error if we get timeout exception and unable to kill the process. In that case, we directly hit the line #274.

narrieta: good point, i'll fix it

azurelinuxagent/ga/update.py
narrieta: Yes, there is no timeout; it's documented in README. 

The customers requesting this feature are on Python 3. If anybody needs a timeout and is using Python 2, they could implement the timeout on their side, or give us feedback and we may implement it on our side if there is enough feedback. Python 2 has been out if support for some time already.

tests_e2e/tests/agent_wait_for_cloud_init/agent_wait_for_cloud_init.py
narrieta: Yes, run by cloud-init. 

The configuration for cloud-init is passed (in Azure) in the custom data: https://learn.microsoft.com/en-us/azure/virtual-machines/linux/tutorial-automate-vm-deployment

There's quite a few things one can do with cloud-init. Here are some samples for Azure: https://learn.microsoft.com/en-us/azure/virtual-machines/linux/using-cloud-init

There is additional information in the readthedocs official documentation for cloud-init.

In the case of this test, it is executing a bash script

tests_e2e/tests/agent_wait_for_cloud_init/agent_wait_for_cloud_init.py
narrieta: We get here only after the Agent has started processing extensions (check starts on line 55).

Once we are here, there is no need to check the timestamps, only the order. The check on line 85 verifies that all these messages were found in the log, and that they are in exactly the same order as in the 'expected' array. The Agent log is in chronological order and grep reads the file sequentially.

maddieford (approved):

tests_e2e/tests/agent_wait_for_cloud_init/agent_wait_for_cloud_init.py
narrieta: @maddieford 

Actually, the first 2 messages in the 'expected' array ('Waiting for cloud-init to complete', 'cloud-init script begin') could be in any order, since the script may begin executing before the Agent starts waiting for cloud-init. 

Strictly speaking, I only need to check that 'ProcessExtensionsGoalState started' happens after 'cloud-init script end'. I added the other messages because I am logging 'output' and the timestamps from the Agent log would be useful for debugging.

I'm going to remove 'Waiting for cloud-init to complete' from the check.

nagworld9 (approved):

maddieford (approved):

PR #3033, opened 2024-01-31
don't allow agent update if attempts reached max limit

nagworld9:
## Description

Current problem: 
agent upgrade from v2 to v3
we clean up all agents except v2
download v3 and exit v2
deamon unable to start v3 and eventually start v2
v2 repeats v3 and it's in loop

Solution:
Keep track number of updates attempts. If attempts >3 and still agent is blacklisted, we don't proceed with update.

Issue # 

---

### PR information


azurelinuxagent/ga/update.py
```diff
@@ -1173,16 +1172,3 @@ def _execute_run_command(command):
         except Exception as e:
             msg = "Error while checking ip table rules:{0}".format(ustr(e))
             logger.error(msg)
-
-    def _reset_legacy_blacklisted_agents(self):
```
nagworld9, line 1177: This reset putting agent update in loop (because it is removing error file) and also, I don't see a value now(how it helps). we still need info of blacklisted agents that are blacklisted by daemon.

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -182,12 +182,26 @@ def run(self, goal_state, ext_gs_updated):
...
-                msg = "Downloaded agent version is in bad state : {0} , skipping agent update".format(
-                    str(agent.version))
+
+            # We are not updating the agent if it is not downloaded properly or agent is blacklisted and already attempted 3 times.
+            # For some reason, if agent is blackslisted in previous attempts(<3), we clear the error state to make agent as good agent.
+            # If we allow to update, we increment the update attempt count.
+            if not agent.is_downloaded:
+                msg = "Agent: {0} is not downloaded properly, skipping agent update".format(str(agent.version))
                 raise AgentUpdateError(msg)
+            elif agent.get_update_attempt_count() >= 3 and agent.is_blacklisted:
```
narrieta, line 193: how does this interact with RSM updates? won't this prevent updates when RSM tries the update again?

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -182,12 +182,26 @@ def run(self, goal_state, ext_gs_updated):
...
+                msg = "Did {0} update attempts previously for version: {1} but still agent not recovered from bad state. So, skipping agent update".format(
+                    agent.get_update_attempt_count(), str(agent.version))
+                raise AgentUpdateError(msg)
+            else:
+                agent.clear_error()
+                agent.inc_update_attempt_count()
+                msg = "Agent update attempt count: {0} for version: {1}".format(agent.get_update_attempt_count(), str(agent.version))
+                logger.info(msg)
+                add_event(op=WALAEventOperation.AgentUpgrade, message=msg, log_event=False)
```
narrieta, line 202: this should be marked as error

azurelinuxagent/ga/agent_update_handler.py
narrieta, line 193: why the check on agent.is_blacklisted? isn't the count enough?

azurelinuxagent/ga/update.py
narrieta: at some point the blacklisting logic in the daemon changed and seems like this code was trying to get rid of errors from previous daemons

with your changes we are removing all agents other than (current, update), whether they are blacklisted or not, right?

if yes, i agree that it is safe to remove this

nagworld9: yup, we remove everything else no matter what state they are

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -182,12 +182,26 @@ def run(self, goal_state, ext_gs_updated):
...
             agent = self._updater.download_and_get_new_agent(self._protocol, agent_family, goal_state)
-            if agent.is_blacklisted or not agent.is_downloaded:
-                msg = "Downloaded agent version is in bad state : {0} , skipping agent update".format(
-                    str(agent.version))
+
+            # We are not updating the agent if it is not downloaded properly or agent is blacklisted and already attempted 3 times.
+            # For some reason, if agent is blackslisted in previous attempts(<3), we clear the error state to make agent as good agent.
+            # If we allow to update, we increment the update attempt count.
+            if not agent.is_downloaded:
+                msg = "Agent: {0} is not downloaded properly, skipping agent update".format(str(agent.version))
```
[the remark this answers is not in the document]
nagworld9: In the case of download failures, I don't want to continue with update and exit the current process. if download failure happens, we delete the directory. So, we could use directory presence for this check instead manifest since we don't use manifest anymore if that's your concern.

As far as telemetry, we do send telemetry for download failures, so checking both would give context.

azurelinuxagent/ga/agent_update_handler.py
nagworld9: That was the intention, right otherwise we will be in loop. If transient failure, this may help. If it's permanent, what's the point of honoring RSM request.

After first RSM request, every new GS is an update. So, we don't know which GS driven by RSM.

nagworld9: well, we get this condition only when agent is blacklisted, otherwise we never multiple attempts. So I want to double check that.

narrieta: ok, if we get here only when agent.is_blacklisted is True, then it is a redundant check. why do we want to double-check?

azurelinuxagent/ga/agent_update_handler.py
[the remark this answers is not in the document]
narrieta: my concerns is that it seems like a useless check. if we are trying to check for download failures, shouldn't this check or similar be done when we download the agent? then the error message can have the appropriate context

azurelinuxagent/ga/agent_update_handler.py
nagworld9: It would kill further rsm retries too.

azurelinuxagent/ga/agent_update_handler.py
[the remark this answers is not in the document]
nagworld9: The download happens inside this GuestAgent module. ahent_update_handler only way know by calling is_download checks

`        agent = GuestAgent.from_agent_package(package_to_download, protocol, is_fast_track_goal_state)
`

azurelinuxagent/ga/guestagent.py
```diff
@@ -16,6 +16,7 @@
 AGENT_ERROR_FILE = "error.json"  # File name for agent error record
 AGENT_MANIFEST_FILE = "HandlerManifest.json"
 MAX_FAILURE = 3  # Max failure allowed for agent before declare bad agent
+AGENT_UPDATE_COUNT_FILE = "update_attempt.json"  # File for tracking agent update attempt count
```
maddieford, line 19: we should also clean up this file during deprovisioning

nagworld9: This file inside agent pkg, we delete Agent directory if I'm not mistaken

azurelinuxagent/ga/agent_update_handler.py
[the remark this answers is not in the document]
narrieta: I'm confused here... we call self._updater.download_and_get_new_agent() on line 185 and then immediately check agent.is_downloaded.

Does that mean that download_and_get_new_agent() can fail to download the agent and not report the error with an exception? that does not seem correct to me.

azurelinuxagent/ga/agent_update_handler.py
narrieta: Have you talked this over with the RSM team?

azurelinuxagent/ga/agent_update_handler.py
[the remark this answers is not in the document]
nagworld9: The GuestAgent class implemented in that way before and required refactoring of changing these methods and removing some of the dead code. That is separate effort I feel.

azurelinuxagent/ga/agent_update_handler.py
nagworld9: I haven't but what is that we can do otherwise to fix this issue?

I will check with them anyway.

azurelinuxagent/ga/guestagent.py
maddieford: Yes you're right. In temu issue customer had an image that was not deprovisioned correctly. When agent detects this state, it attempts to remove some stale files to minimize unexpected issues. For the TEMU PR, we added published_hostname to that list of files. I think we should add this too.

azurelinuxagent/pa/deprovision/default.py
We attempt to remove files in del_lib_dir_files, called by setup_changed_unique_id 
https://github.com/maddieford/WALinuxAgent/blob/0cd86173d40fc97a8eb21e199c2409d2d334fee4/azurelinuxagent/pa/deprovision/default.py#L154
https://github.com/maddieford/WALinuxAgent/blob/0cd86173d40fc97a8eb21e199c2409d2d334fee4/azurelinuxagent/pa/deprovision/default.py#L216

azurelinuxagent/ga/ga_version_updater.py
```diff
@@ -104,6 +98,33 @@ def sync_new_gs_id(self, gs_id):
...
+        agent_name = "{0}-{1}".format(AGENT_NAME, package_to_download.version)
+        agent_dir = os.path.join(conf.get_lib_dir(), agent_name)
+        agent_pkg_path = ".".join((os.path.join(conf.get_lib_dir(), agent_name), "zip"))
+        agent_handler_manifest_file = os.path.join(agent_dir, AGENT_MANIFEST_FILE)
+        if not os.path.exists(agent_dir) or not os.path.isfile(agent_handler_manifest_file):
+            protocol.client.download_zip_package("agent package", package_to_download.uris, agent_pkg_path, agent_dir, use_verify_header=is_fast_track_goal_state)
+        else:
+            logger.info("Agent {0} was previously downloaded - skipping download", agent_name)
+
+        if not os.path.isfile(agent_handler_manifest_file):
```
narrieta, line 118: I still find this check confusing: why do we check for the manifest, but nor for the egg? we need both in order to execute the Agent. Or looking at this differently, do we need to check at all? the download/zip already succeed at this point.

narrieta (approved):

azurelinuxagent/ga/ga_version_updater.py
nagworld9: Intention is not to exit the current process when we know we can't run the agent without both. I agree we need both checks or could continue with update since download/zip succeeded. When Deamon finds missing, it will eventually blacklist and stop the update.  Former case, we may end up downloading agent again and again if it's more permanent issue. Latter case, we don't give second chance to download. Not see an advantage in either case though.:)

nagworld9: well, latter case not having check safer than download again?

narrieta: I am not sure I follow the "downloading agent again and again" vs "we don't give second chance to download.". 

This may be related to my other comment about checking that the directory exists and skipping the download. Every update attempt should start from scratch, I think, to prevent these issues of incomplete downloads.

If we really want to check the integrity of the downloaded agent, the check should be stronger than checking for the manifest (or the egg, or both), more like adding a checksum to the package and verifying the checksum.

PR #3057, opened 2024-02-14
Add Python 3.10 to the pylint matrix

narrieta:
I also added many suppresions of deprecated-method/deprecated-module. Those need to be addressed, but I will do it on separate PRs to facilitate the reviews.

I created 3.10.pylintrc out of 3.6.pylintrc with a few adjustments for new warnings that were added or old warnings that were removed on 3.10. I'll highlight those with comments in the review. Documentation for warnings (use search for specific warning): https://pylint.pycqa.org/en/latest/tutorial.html

### DIFF
--- a/.github/workflows/ci_pr.yml
+++ b/.github/workflows/ci_pr.yml
@@ -14,8 +14,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - python-version: 2.6
-          - python-version: 3.4
+          - python-version: "2.6"
+          - python-version: "3.4"
 
     name: "Python ${{ matrix.python-version }} Unit Tests"
     runs-on: ubuntu-20.04
@@ -42,7 +42,7 @@ jobs:
 
     - name: Test with nosetests
       run: |
-        if [[ ${{ matrix.python-version }} == 2.6 ]]; then
+        if [[ ${{ matrix.python-version }} == "2.6" ]]; then
           source /home/waagent/virtualenv/python2.6.9/bin/activate
         else
           source /home/waagent/virtualenv/python3.4.8/bin/activate
@@ -86,21 +86,24 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - python-version: 3.5
+          - python-version: "3.5"
             PYLINTOPTS: "--rcfile=ci/3.6.pylintrc --ignore=tests_e2e,makepkg.py"
 
-          - python-version: 3.6
+          - python-version: "3.6"
             PYLINTOPTS: "--rcfile=ci/3.6.pylintrc --ignore=tests_e2e"
 
-          - python-version: 3.7
+          - python-version: "3.7"
             PYLINTOPTS: "--rcfile=ci/3.6.pylintrc --ignore=tests_e2e"
 
-          - python-version: 3.8
+          - python-version: "3.8"
             PYLINTOPTS: "--rcfile=ci/3.6.pylintrc --ignore=tests_e2e"
 
-          - python-version: 3.9
+          - python-version: "3.9"
             PYLINTOPTS: "--rcfile=ci/3.6.pylintrc"
             additional-nose-opts: "--with-coverage --cover-erase --cover-inclusive --cover-branches --cover-package=azurelinuxagent"
+
+          - python-version: "3.10"
+            PYLINTOPTS: "--rcfile=ci/3.10.pylintrc --ignore=tests"
 
     name: "Python ${{ matrix.python-version }} Unit Tests"
     runs-on: ubuntu-20.04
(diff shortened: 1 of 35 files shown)


.github/workflows/ci_pr.yml
```diff
@@ -14,8 +14,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - python-version: 2.6
-          - python-version: 3.4
+          - python-version: "2.6"
```
narrieta, line 17: we were using numbers for the version, so "3.10" was actually 3.1

ci/3.10.pylintrc
```diff
@@ -0,0 +1,40 @@
+[MESSAGES CONTROL]
+
+disable=C, # (C) convention, for programming standard violation
+    broad-except, # W0703: *Catching too general exception %s*
+    broad-exception-raised, # W0719: Raising too general exception: Exception
```
narrieta, line 5: new

ci/3.10.pylintrc
```diff
@@ -0,0 +1,40 @@
...
+    consider-using-dict-comprehension, # R1717: *Consider using a dictionary comprehension*
+    consider-using-from-import, # R0402: Use 'from foo import bar' instead
```
narrieta, line 7: new

ci/3.10.pylintrc
```diff
@@ -0,0 +1,40 @@
...
+    consider-using-in, # R1714: *Consider merging these comparisons with "in" to %r*
+    consider-using-set-comprehension, # R1718: *Consider using a set comprehension*
+    consider-using-with, # R1732: *Emitted if a resource-allocating assignment or call may be replaced by a 'with' block*
+    duplicate-code, # R0801: *Similar lines in %s files*
+    fixme, # Used when a warning note as FIXME or TODO is detected
+    logging-format-interpolation, # W1202: Use lazy % formatting in logging functions
+    logging-fstring-interpolation, # W1203: Use lazy % or .format() formatting in logging functions
```
narrieta, line 14: 13 & 14 new

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -184,10 +184,6 @@ def initialize(self):
 
                 _log_cgroup_info("systemd version: {0}", systemd.get_version())
 
-                # This is temporarily disabled while we analyze telemetry. Likely it will be removed.
```
narrieta, line 187: we added those when we started using the 'azure' slice and drop in files... that has been stable for a long time, so no longer needed

.github/workflows/ci_pr.yml
```diff
@@ -133,13 +136,13 @@ jobs:
         pylint $PYLINTOPTS --jobs=0 $PYLINTFILES
 
     - name: Test with nosetests
-      if: success() || (failure() && steps.install-dependencies.outcome == 'success')
+      if: matrix.python-version != '3.10' && (success() || (failure() && steps.install-dependencies.outcome == 'success'))
```
narrieta, line 139: we are note testing on 3.10 yet

.github/workflows/ci_pr.yml
```diff
@@ -86,22 +86,25 @@ jobs:
...
+          - python-version: "3.8"
             PYLINTOPTS: "--rcfile=ci/3.6.pylintrc --ignore=tests_e2e"
 
-          - python-version: 3.9
+          - python-version: "3.9"
             PYLINTOPTS: "--rcfile=ci/3.6.pylintrc"
             additional-nose-opts: "--with-coverage --cover-erase --cover-inclusive --cover-branches --cover-package=azurelinuxagent"
 
+          - python-version: "3.10"
+            PYLINTOPTS: "--rcfile=ci/3.10.pylintrc --ignore=tests"
```
narrieta, line 106: the unit tests are not Python 3.10 compatible, no need to lint them

maddieford (approved):

nagworld9 (approved):

ci/3.10.pylintrc
nagworld9: where do we get this data?

narrieta: what do you mean?

nagworld9: I mean how do one get these pylint warnings list. Tomorow If I add for 3.11, where do I go and get these?

narrieta: I see. 

Run pylint on the code, check the documentation for what each warning means, then make a judgement call: fix the ones that *must* be fixed, for the rest fix as many as possible. When there are too many to fix, think how we address them. Do we want to suppress each one of them or suppress globally? If globally, add to this file.

In this specific case, this warning is just complaining about the syntax used to do the import. We have way too many of these and the functionality/correctness of the code is not affected. It'd be nice to fix them, but I didn't consider spending time on them was worth it so I suppressed globally.

PR #3063, opened 2024-02-21
Add distutils/version.py to azurelinuxagent

### LINKED ISSUE #2919: Deprecation of distutils
Considering that `disutils` is [deprecated](https://peps.python.org/pep-0632/) in Python 3.10 and Python 3.11. Are the plans to tackle this issue for Python 3.12 when it will no longer be supported by python since it is used couple of places.

narrieta:
distutils was removed from Python >= 12. We use Version and LooseVersion from it. 

Adding distutils/version.py to azurelinuxagent. StrictVersion was removed from the copy to  avoid adding more dependencies to this deprecated module.

Fixes #2919


azurelinuxagent/common/future.py
```diff
@@ -61,14 +61,23 @@
...
-
     if sys.version_info[1] >= 7:
         from collections import OrderedDict  # For Py 2.7+
     else:
         from ordereddict import OrderedDict  # Works only on 2.6  # pylint: disable=E0401
 else:
     raise ImportError("Unknown python version: {0}".format(sys.version_info))
 
+#
+# distutils has been removed from Python >= 3.12; use the copy from azurelinuxagent instead
```
nagworld9, line 72: if it's same copy, why don't we use it for all py versions? or eventually we remove distutils dependency when things work with copy?

narrieta: version.py is not identical across python versions; mostly syntactic differences, but i do not want to take the risk of introducing any changes in behavior across existing distros/python versions, so I am using the copy only on distros/python where there isn't a version.py

narrieta: (LooseVersion is used in the osutil hierarchy in many places, and that code is poorly tested)

nagworld9 (approved):

PR #3066, opened 2024-02-23
reset uphold setting for agent service in flatcar distro

nagworld9:
## Description

 New flatcar images set the uphold property for agent service that is causing automatic restart on stop cmd
 [Upholds= dependency on it has a continuous effect, constantly restarting the unit if necessary]
 Resetting the uphold property as workaround for now.

Issue # 

---

### PR information


tests_e2e/orchestrator/scripts/install-agent
```diff
@@ -140,6 +140,24 @@ if [[ $(uname -a) == *"flatcar"* ]]; then
   if [[ ! -f /usr/share/oem/waagent.conf ]]; then
     ln -s "$waagent_conf_path" /usr/share/oem/waagent.conf
   fi
+
```
narrieta, line 143: let's move this code to the if statement for flatcar

tests_e2e/orchestrator/scripts/install-agent
```diff
@@ -140,6 +140,24 @@ if [[ $(uname -a) == *"flatcar"* ]]; then
...
+    if [[ ! -d /etc/systemd/system/$target_name.d ]]; then
+      mkdir -p /etc/systemd/system/$target_name.d
+    fi
+    echo -e "[Unit]\nUpholds=" > /etc/systemd/system/$target_name.d/10-waagent-sysext.conf
+    systemctl daemon-reload
+  fi
+  # Flatcar images does automatic reboot without user input, so stopping the service which trigger the reboot
+  # Broadcast message from locksmithd at 2024-02-23 19:48:55.478412272 +0000 UTC m=
+  # System reboot in 5 minutes!
+  systemctl stop locksmithd
```
narrieta, line 160: let's turn it off rather than stopping it (otherwise it'll just restart on the next boot)
```

Configuration
There are three different strategies that locksmithd can use after the update engine has successfully applied an update:

etcd-lock - reboot after first taking a lock in etcd.
reboot - reboot without taking a lock.
off - causes locksmithd to exit and do nothing.
These strategies will either be followed immediately after an update, or during the next available [reboot window](https://github.com/flatcar/locksmith#reboot-windows) if one has been configured.

These strategies can be configured via /etc/flatcar/update.conf with a line that looks like:

REBOOT_STRATEGY=reboot
```

nagworld9: I tried this setting and its not effective until reboot. Maybe I need to check if service restart required.

tests_e2e/orchestrator/scripts/install-agent
nagworld9: This code already inside if statement for flatcar

narrieta (approved):

maddieford (approved):

PR #3070, opened 2024-02-27
Use legacycrypt instead of crypt on Python >= 3.13

narrieta:
The crypt module will be removed in Python 3.13. Using legacycrypt as a temporary workaround.


azurelinuxagent/common/osutil/default.py
```diff
@@ -26,23 +27,33 @@
...
 from pwd import getpwall
 
-import array
+from azurelinuxagent.common.exception import OSUtilError
+# 'crypt' was removed in Python 3.13; use legacycrypt instead
+if sys.version_info[0] == 3 and sys.version_info[1] >= 13 or sys.version_info[0] > 3:
+    try:
+        from legacycrypt import crypt
+    except ImportError:
+        def crypt(password, salt):
```
narrieta, line 46: We provide this stub implementation in case the module is not present, otherwise the entire process will crash

setup.py
```diff
@@ -314,13 +314,16 @@ def run(self):
 
 
 # Note to packagers and users from source.
-# In version 3.5 of Python distribution information handling in the platform
-# module was deprecated. Depending on the Linux distribution the
-# implementation may be broken prior to Python 3.7 wher the functionality
```
narrieta, line 319: it was actually removed on 3.7

maddieford (approved):

narrieta:
@nagworld9 

`question for my understanding, what is this requires does? will setup fail if these modules not present?`

from my experimentation, it just creates a requirements file for the Agent's package, not sure what tool does the actual install. I'll have the JIT/PA folks to follow that up, since it's those components who depend on this module. As far as the Extension Handler, the dummy 'crypt' function I define will prevent the Agent from crashing if the module is not installed.

jeremycline:
Hey, just a heads-up that WALinuxAgent is broken in Fedora Rawhide (what will be Fedora 41) since it's now shipping Python 3.13-b1. It'd be great to get this into a release well ahead of Fedora 41 and Python 3.13's release.

nagworld9:
when are they shipping? Rough estimate of our next release is end of this month

PR #3076, opened 2024-03-01
Add lock around access to fast_track.json

narrieta:
Both the monitor and extension handler threads  create instances of HostPluginProtocol, which creates a race condition on the initialization of fast_track.json. Added a lock around access to the file.

### DIFF
--- a/azurelinuxagent/common/protocol/hostplugin.py
+++ b/azurelinuxagent/common/protocol/hostplugin.py
@@ -20,6 +20,7 @@ import base64
 import datetime
 import json
 import os.path
+import threading
 import uuid
 
 from azurelinuxagent.common import logger, conf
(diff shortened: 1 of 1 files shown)


azurelinuxagent/common/protocol/hostplugin.py
```diff
@@ -446,16 +452,17 @@ def get_fast_track_timestamp():
...
 
-        try:
-            with open(HostPluginProtocol._get_fast_track_state_file(), "r") as file_:
-                return json.load(file_)["timestamp"]
-        except Exception as e:
-            logger.warn("Can't retrieve the timestamp for the most recent Fast Track goal state ({0}), will assume the current time. Error: {1}",
-                    HostPluginProtocol._get_fast_track_state_file(), ustr(e))
-        return timeutil.create_timestamp(datetime.datetime.utcnow())
+            try:
+                with open(HostPluginProtocol._get_fast_track_state_file(), "r") as file_:
```
nagworld9, line 460: don't we need here too?

narrieta: yes, the whole function body is protected

PR #3077, opened 2024-03-04
Added new options to verify if network is up at boot time

nagworld9:
## Description
As we know we do the firewall setup before network is up in order to prevent non-root user access to wireserver once network is up. 
This validation we do with cronjob script at boot time in persist firewall scenario. The script verifies the network is up and then validates goal state fetch.
Today we verify the network is up by ping to 8.8.8.8. Sometimes this check is unstable, and we suspect factors like routing tables, network security rules, etc that may be producing noise. 

This pr adding other options in order to compare which option is more stable. For now, it just logs the connection status when ping to 8.8.8.8 is failed.

Issue # 

---

### PR information

### DIFF
--- a/tests_e2e/tests/scripts/agent_persist_firewall-access_wireserver
+++ b/tests_e2e/tests/scripts/agent_persist_firewall-access_wireserver
@@ -19,6 +19,11 @@
 # Helper script which tries to access Wireserver on system reboot. Also prints out iptable rules if non-root and still
 # able to access Wireserver
 
+if [[ $# -ne 1 ]]; then
+    echo "Usage: agent_persist_firewall-access_wireserver <test-user>"
+    exit 1
+fi
+TEST_USER=$1
 USER=$(whoami)
 echo "$(date --utc +%FT%T.%3NZ): Running as user: $USER"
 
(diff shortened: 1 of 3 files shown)


narrieta (approved):

tests_e2e/tests/scripts/agent_persist_firewall-access_wireserver
```diff
@@ -48,6 +66,19 @@ while [ "$IS_ONLINE" -eq 1 ]; do
 
 done
 
+# logging other options output to compare and evaluate which option is more stable when ping to 8.8.8.8 failed
+if [ "$IS_ONLINE" -eq 1 ]; then
```
maddieford, line 70: Just confirming we're only logging these options right now. It doesn't look like we update IS_ONLINE with the results of these options

maddieford (approved):

tests_e2e/tests/scripts/agent_persist_firewall-access_wireserver
nagworld9: Just logging for now, we look at these outputs when test fail with first option.

PR #3078, opened 2024-03-04
Add DistroVersion class to compare distro versions

narrieta:
A previous PR added a copy of distuls/version.py as a temporary workaround to use LooseVersion (since it is deprecated and will be removed from Python). This PR replaces that copy with a new class: DistroVersion that implements the same behavior in terms of comparing versions.

The results of some of the comparisons done with LooseVersion/DistroVersion are not always intuitive and can lead to subtle code bugs (e.g. "1" != "1.0" != "1.0.0", etc). There are a couple of places in the code where we use similar comparisons; be careful with those.

Also, we have another class that deals with versions,  FlexibleVersion, with its own behavior. This creates confusion as to what class to use when comparing versions.

I'll do another iteration of this code to merge DistroVersion and FlexibleVersion.


tests/common/utils/test_text_util.py
```diff
@@ -68,23 +67,6 @@ def test_remove_bom(self):
         data = textutil.remove_bom(data)
         self.assertEqual(u"  ", data)
 
-    def test_version_compare(self):
```
narrieta, line 71: moved to test_distro_version.py

tests/common/utils/test_distro_version.py
```diff
@@ -0,0 +1,128 @@
...
+        if sys.version_info[0] == 2:
+            self.assertTrue(DistroVersion("3.11.2-rc.1") < DistroVersion("3.11.2-rc.a"))
+        else:
+            # TypeError: '<' not supported between instances of 'int' and 'str'
+            with self.assertRaises(TypeError):
+                _ = DistroVersion("3.11.2-rc.1") < DistroVersion("3.11.2-rc.a")
+
+        # AttributeError: 'FlexibleVersion' object has no attribute '_fragments'
+        with self.assertRaises(AttributeError):
+            _ = DistroVersion("1.0.0.0") == FlexibleVersion("1.0.0.0")
```
nagworld9, line 124: you mean it will fail with Attribute error If I use FlexibleVersion in the code. I didn't this comment

tests/common/utils/test_distro_version.py
```diff
@@ -0,0 +1,128 @@
...
+        self.assertTrue(DistroVersion("2.1-systemd-alpha") < DistroVersion("2.1-systemd-rc"))
+        self.assertTrue(DistroVersion("2308a") < DistroVersion("2308beta"))
+        self.assertTrue(DistroVersion("6.0.0.beta4") < DistroVersion("6.0.0.beta5"))
+        self.assertTrue(DistroVersion("9.13.1P8X1") < DistroVersion("9.13.1RC1"))
+        self.assertTrue(DistroVersion("a") < DistroVersion("rc"))
+        self.assertTrue(DistroVersion("Clawhammer__9.14.0"), DistroVersion("Clawhammer__9.14.1"))
+        self.assertTrue(DistroVersion("FFFF") < DistroVersion("h"))
+        self.assertTrue(DistroVersion("None") < DistroVersion("n/a"))
+
+        if sys.version_info[0] == 2:
```
nagworld9, line 115: That means in prod there a chance that code can hit typeerror and break the agent?

tests/common/utils/test_distro_version.py
```diff
@@ -0,0 +1,128 @@
...
+        Test that DistroVersion does some common comparisons correctly.
+        """
+        self.assertTrue(DistroVersion("1.0.0") < DistroVersion("2.0.0."))
+        self.assertTrue(DistroVersion("1.0.0") < DistroVersion("1.1.0"))
+        self.assertTrue(DistroVersion("1.0.0") < DistroVersion("1.0.1"))
+
+        self.assertTrue(DistroVersion("1.0.0") == DistroVersion("1.0.0"))
+        self.assertTrue(DistroVersion("1.0.0") != DistroVersion("2.0.0"))
+
+        self.assertTrue(DistroVersion("13") != DistroVersion("13.0"))
```
nagworld9, line 64: This looks to me DistroVersion is also variant of strict version and Flexible version also documented as same. Is that means both does same thing? 

13 != 13.0  !=13.0.0 is true in both modules?

narrieta: DistroVersion is not a variant of StrictVersion, but of LooseVersion

tests/common/utils/test_distro_version.py
narrieta: yes, the current code with LooseVersion has this issue

tests/common/utils/test_distro_version.py
narrieta: this comparison raises AttributeError. the comment describes includes the message in the exception

nagworld9 (approved):

maddieford (approved):

PR #3081, opened 2024-03-06
Run unit tests with pytest on Python >= 3.10

narrieta:
Our tests are compatible with pytest, so this is a good option to execute unit tests on Python >= 3.10. In fact, pytest is recommended over nose2 for new projects.

We were using 'attr' from the nose module to mark tests that require sudo. I did not see an obvious way to do the same with pytest, so I removed the attributed and moved those tests to a different file. Unit tests should not require sudo; these 3 tests are a rare exception and I do not anticipate many more tests to require it.

Also, pytest executes the entire test run within the same process, as opposed to nose, which creates a process per test file. The tests in test_agent modify the global logger and conf objects, which was causing many tests to fail when using pytest. I added cleanup code.

pytest is supported on Python >= 3.8; nose on <= 3.9. This PR uses nose for <= 3.9, pytest for >= 3.10


tests/common/test_telemetryevent.py
```diff
@@ -19,23 +19,23 @@
 from tests.lib.tools import AgentTestCase
 
 
-def get_test_event(name="DummyExtension", op="Unknown", is_success=True, duration=0, version="foo", evt_type="", is_internal=False,
```
narrieta, line 22: nose was identifying this as a test; renamed it and moved it to the TestTelemetryEvent class

tests/ga/test_cgroupconfigurator.py
```diff
@@ -526,112 +523,6 @@ def test_start_extension_command_should_disable_cgroups_and_invoke_the_command_d
 
                         self.assertEqual(len(CGroupsTelemetry._tracked), 0, "No cgroups should have been created")
 
-    @skip_if_predicate_true(is_python_version_26_or_34, "Disabled on Python 2.6 and 3.4 for now. Need to revisit to fix it")
```
narrieta, line 529: Moved those to separate file.

tests/ga/test_cgroupconfigurator_sudo.py
```diff
@@ -0,0 +1,157 @@
+# Copyright 2018 Microsoft Corporation
```
narrieta, line 1: Moved from test_cgroupconfigurator.py without changes, other than removing "@attr('requires_sudo')"

tests/ga/test_extension.py
```diff
@@ -63,10 +63,6 @@
 SUCCESS_CODE_FROM_STATUS_FILE = 1
 
 
-def do_not_run_test():
```
narrieta, line 66: nose was identifying this as a test; not used, so I removed it

maddieford (approved):

nagworld9 (approved):
wow, it seems we were lucky our tests are compatible with other frameworks with minimal changes. Maybe we happen to use native Apis instead of nose specific ones.

narrieta:
@nagworld9 nose/pytest are just the runners. We are using the standard Python unitest module and the mock module for the actual tests. From nose we were using only 'attr', to mark tests as "requires sudo". Initially we were getting tons of failures with nose2/pytest, but those turned out to be caused by AgentTest modifying logger/conf. That was a pain to debug, but once solved, most of the errors went away.

PR #3084, opened 2024-03-07
Fix pytest warnings

narrieta:
Fixing pytest warnings; most of them were due to strings representing regular expressions missing the 'r' prefix (we were suppressing those from pylint).

### DIFF
--- a/tests/common/test_singletonperthread.py
+++ b/tests/common/test_singletonperthread.py
@@ -6,7 +6,7 @@ from azurelinuxagent.common.singletonperthread import SingletonPerThread
 from tests.lib.tools import AgentTestCase, clear_singleton_instances
 
 
-class TestClassToTestSingletonPerThread(SingletonPerThread):
+class Singleton(SingletonPerThread):
     """
     Since these tests deal with testing in a multithreaded environment,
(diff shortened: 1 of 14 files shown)


tests/common/test_singletonperthread.py
```diff
@@ -6,7 +6,7 @@
 from tests.lib.tools import AgentTestCase, clear_singleton_instances
 
 
-class TestClassToTestSingletonPerThread(SingletonPerThread):
```
narrieta, line 9: Renamed this, since pytest was identifying it as a Test class

tests/lib/tools.py
```diff
@@ -156,8 +156,6 @@ def setUpClass(cls):
             cls.assertIsNone = cls.emulate_assertIsNone
         if not hasattr(cls, "assertIsNotNone"):
             cls.assertIsNotNone = cls.emulate_assertIsNotNone
-        if hasattr(cls, "assertRaisesRegexp"):
```
narrieta, line 159: I guess this was needed for some old Python versions, but it is no longer needed, and it is producing a warning about assertRaisesRegexp being deprecated.

maddieford (approved):

nagworld9 (approved):

PR #3089, opened 2024-03-08
Add keyvault test to daily run + Specify tests suite as a list

narrieta:
It turns out I forgot to add the keyvault test to the daily runbook.

As we've added more tests to the runbook, listing them as a comma-separate list makes it harder to notice such errors. I changed the test suites to be an array instead, which makes it much easy to read.

### DIFF
--- a/tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
+++ b/tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
@@ -46,6 +46,7 @@ class AgentTestSuitesCombinatorSchema(schema.Combinator):
     resource_group_name: str = field(default_factory=str, metadata=field_metadata(required=True))
     subscription_id: str = field(default_factory=str, metadata=field_metadata(required=True))
     test_suites: str = field(default_factory=str, metadata=field_metadata(required=True))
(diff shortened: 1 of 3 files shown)


tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
```diff
@@ -260,8 +266,8 @@ def create_environment_list(self) -> List[Dict[str, Any]]:
 
         return environments
 
-    def create_existing_vm_environment(self) -> Dict[str, Any]:
-        loader = AgentTestLoader(self.runbook.test_suites, self.runbook.cloud)
+    def create_existing_vm_environment(self, test_suites: List[str]) -> Dict[str, Any]:
+        loader = AgentTestLoader(test_suites, self.runbook.cloud)
```
maddieford, line 270: how is loader used in create_existing_vm_environment and create_existing_vmss_environment? I don't see it being used anywhere here

narrieta: we get the test suites from it, it's the last property in the return value

`c_test_suites": loader.test_suites,`

nagworld9 (approved):

PR #3091, opened 2024-03-13
agent publish refactor

nagworld9:
## Description

Added RSM update verification in agent publish test.

PR includes new pipeline parameter for passing test arguments. Passing all the way from pipeline to each test.
The expected format is comma-separated list of key=value pair like `publihsedVersion=2.10.0.8,hostName=Test`
This is needed especially for agent publish test to know the published version to validate. This new parameter will be useful for other tests when you want to test something dynamically without needing to change tests locally. 

Issue # 

---

### PR information

### DIFF
--- a/tests_e2e/orchestrator/lib/agent_test_suite.py
+++ b/tests_e2e/orchestrator/lib/agent_test_suite.py
@@ -149,6 +149,8 @@ class AgentTestSuite(LisaTestSuite):
 
         self._test_suites: List[AgentTestSuite]  # Test suites to execute in the environment
 
+        self._test_args: Dict[str, str]  # Additional arguments pass to the test suite
+
         self._cloud: str  # Azure cloud where test VMs are located
         self._subscription_id: str  # Azure subscription where test VMs are located
         self._location: str  # Azure location (region) where test VMs are located
@@ -209,6 +211,7 @@ class AgentTestSuite(LisaTestSuite):
         self._environment_name = variables["c_env_name"]
 
         self._test_suites = variables["c_test_suites"]
+        self._test_args = variables["test_args"]
 
         self._cloud = variables["cloud"]
         self._subscription_id = variables["subscription_id"]
@@ -565,6 +568,7 @@ class AgentTestSuite(LisaTestSuite):
 
                     try:
                         test_context = self._create_test_context()
+                        test_args = self._get_test_args()
 
                         if not self._skip_setup:
                             try:
@@ -578,7 +582,7 @@ class AgentTestSuite(LisaTestSuite):
                         for suite in self._test_suites:
                             log.info("Executing test suite %s", suite.name)
                             self._lisa_log.info("Executing Test Suite %s", suite.name)
-                            case_success, check_log_start_time = self._execute_test_suite(suite, test_context, check_log_start_time)
+                            case_success, check_log_start_time = self._execute_test_suite(suite, test_context, test_args, check_log_start_time)
                             test_suite_success = case_success and test_suite_success
                             if not case_success:
                                 failed_cases.append(suite.name)
@@ -613,7 +617,7 @@ class AgentTestSuite(LisaTestSuite):
                     if not test_suite_success or unexpected_error:
                         raise TestFailedException(self._environment_name, failed_cases)
 
-    def _execute_test_suite(self, suite: TestSuiteInfo, test_context: AgentTestContext, check_log_start_time: datetime.datetime) -> Tuple[bool, datetime.datetime]:
+    def _execute_test_suite(self, suite: TestSuiteInfo, test_context: AgentTestContext, test_args: Dict[str, str], check_log_start_time: datetime.datetime) -> Tuple[bool, datetime.datetime]:
         """
         Executes the given test suite and returns a tuple of a bool indicating whether all the tests in the suite succeeded, and the timestamp that should be used
         for the next check of the agent log.
@@ -645,7 +649,7 @@ class AgentTestSuite(LisaTestSuite):
 
                         test_success: bool = True
 
-                        test_instance = test.test_class(test_context)
+                        test_instance = test.test_class(test_context, test_args)
                         try:
                             test_instance.run()
                             summary.append(f"[Passed]  {test.name}")
@@ -841,6 +845,18 @@ class AgentTestSuite(LisaTestSuite):
                 vmss=scale_set,
                 username=self._user,
                 identity_file=self._identity_file)
+
+    def _get_test_args(self) -> Dict[str, str]:
+        """
+        Returns the arguments to be passed to the test classes
+        """
+        test_args: Dict[str, str] = {}
+        if self._test_args == "":
+            return test_args
+        for arg in self._test_args.split(','):
+            key, value = arg.split('=')
+            test_args[key] = value
+        return test_args
 
     @staticmethod
     def _mark_log_as_failed():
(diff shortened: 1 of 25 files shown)


tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -842,6 +846,18 @@ def _create_test_context(self,) -> AgentTestContext:
                 username=self._user,
                 identity_file=self._identity_file)
 
+    def _get_test_args(self) -> Dict[str, str]:
```
nagworld9, line 849: converting list of key=value to dictionary

tests_e2e/tests/lib/agent_test.py
```diff
@@ -49,8 +49,9 @@ class AgentTest(ABC):
     """
     Abstract base class for Agent tests
     """
-    def __init__(self, context: AgentTestContext):
+    def __init__(self, context: AgentTestContext, test_args: Dict[str, str]):
         self._context: AgentTestContext = context
+        self._test_args: Dict[str, str] = test_args
```
nagworld9, line 54: These args exposed to all tests

tests_e2e/tests/agent_publish/agent_publish.py
```diff
@@ -86,20 +129,6 @@ def _check_cse(self) -> None:
         )
         custom_script_2_1.assert_instance_view(expected_version="2.1", expected_message=message)
 
-    def get_ignore_error_rules(self) -> List[Dict[str, Any]]:
```
nagworld9, line 89: not needed in this test

tests_e2e/tests/agent_update/rsm_update.py
```diff
@@ -146,64 +140,6 @@ def _prepare_agent(self) -> None:
                               "update-waagent-conf AutoUpdate.UpdateToLatestVersion=y Debug.EnableGAVersioning=y AutoUpdate.GAFamily=Test", use_sudo=True)
         log.info('Successfully updated agent update config \n %s', output)
 
-    @staticmethod
```
nagworld9, line 149: moved to common helper

tests_e2e/tests/lib/agent_update_helpers.py
```diff
@@ -0,0 +1,87 @@
...
+
+def request_rsm_update(requested_version: str, vm: VirtualMachineClient) -> None:
+    """
+    This method is to simulate the rsm request.
+    First we ensure the PlatformUpdates enabled in the vm and then make a request using rest api
+    """
+    if not verify_agent_update_flag_enabled(vm):
+        # enable the flag
+        log.info("Attempting vm update to set the enableVMAgentPlatformUpdates flag")
+        enable_agent_update_flag(vm)
```
maddieford, line 61: Why is this necessary? How is it used vs. our waagent.conf flag

nagworld9: This different flag need to set in the vm model for CRP/RSM driving RSM updates

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -209,6 +211,7 @@ def _initialize(self, environment: Environment, variables: Dict[str, Any], lisa_
         self._environment_name = variables["c_env_name"]
 
         self._test_suites = variables["c_test_suites"]
+        self._test_args = variables["test_args"]
```
maddieford, line 214: looks like there may be inconsistency in type here. variables["test_argd"] is str while self._test_args is dict

nagworld9: It's still a string, later we convert to dict. The variables exposed to user in pipeline, runbook or command line is in string in the form of list of comma-separated key=value pair
Example
`publihsedVersion=2.10.0.8,hostName=Test`

nagworld9: oh, you are right. It should be string. I just noticed

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -565,6 +568,7 @@ def _execute(self) -> None:
 
                     try:
                         test_context = self._create_test_context()
+                        test_args = self._get_test_args()
```
narrieta, line 571: we should do the conversion in _initialize() and keep the result in self._test_args, to avoid adding more parameters to the execute method

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -645,7 +649,7 @@ def _execute_test_suite(self, suite: TestSuiteInfo, test_context: AgentTestConte
 
                         test_success: bool = True
 
-                        test_instance = test.test_class(test_context)
+                        test_instance = test.test_class(test_context, test_args)
```
narrieta, line 652: Rather than passing test_args as a new argument to the test, each test_arg should be added to the text_context as a new property (or if the property already exists in the context, the just set its value).

Then AgentPublishTest.run_from_command_line should read the new parameters from the command line and add them to the test context as well. That is, it would parse the publishedVersion argument, instead of the test_args argument

narrieta (commented):
Added some comments, could you also describe how you are planning to planning to pass the arguments to the publish test from the pipeline? If the new argument is required, we should probably create a new pipeline.yml for the publish pipeline.

nagworld9:
> Added some comments, could you also describe how you are planning to planning to pass the arguments to the publish test from the pipeline? If the new argument is required, we should probably create a new pipeline.yml for the publish pipeline.

My idea is to it would benefit for other test_suites(for dev runs etc) if we added to main pipeline because the internal implementation is going to be same regardless. 

When we run the pipeline, one need to pass like list of key=value pair in the pipeline arg. This converted to dict and added to agenttestcontext, so that all the tests access to dict. Whichever test needed, they read the keys to get value.

For publish test when I run the pipeline with agent_publish as test_suite, I pass publishedVersion={version}, the publish test reads this key publihsedVersion

narrieta:
> My idea is to it would benefit for other test_suites(for dev runs etc) if we added to main pipeline because the internal implementation is going to be same regardless.

Yes, I agree that it is useful and can be added as an optional parameter in the main pipeline

> For publish test when I run the pipeline with agent_publish as test_suite, I pass publishedVersion={version}, the publish test reads this key publihsedVersion

If this parameter is required by the publish test, then maybe should have a pipeline just for publish,

nagworld9:
> > My idea is to it would benefit for other test_suites(for dev runs etc) if we added to main pipeline because the internal implementation is going to be same regardless.
> 
> Yes, I agree that it is useful and can be added as an optional parameter in the main pipeline
> 
> > For publish test when I run the pipeline with agent_publish as test_suite, I pass publishedVersion={version}, the publish test reads this key publihsedVersion
> 
> If this parameter is required by the publish test, then maybe should have a pipeline just for publish,

Not sure I followed here. I wouldn't send this as parameter, rather I send this parameter as value in the new the test_arg parameter. That way I don't have deal with separate code path just for agent publish

nagworld9:
@narrieta Updated my pr. I added todo for reading it from cmd line. I couldn't get the flow right, maybe I'm missing something and also, I have different thought like we can directly use argparser for reading test specific arguments in the test itself.

tests_e2e/tests/lib/agent_test.py
```diff
@@ -73,6 +73,8 @@ def run_from_command_line(cls):
         """
         Convenience method to execute the test when it is being invoked directly from the command line (as opposed as
         being invoked from a test framework or library.)
+
+        Todo: Need to implement for reading test specific arguments from command line
```
narrieta, line 77: Let's use capitals: TODO. Then some tools highlight that comment

narrieta (approved):

maddieford (approved):

PR #3096, opened 2024-03-16
Cgroups api refactor for v2

maddieford:
## Description
This PR refactors the SystemdCgroupApi to be a base abstract class for different versions of the cgroup api. There are two different implementations (SystemdCgroupApiv1 and SystemdCgroupApiv2) which inherit _SystemdCgroupApi. With these changes, the SystemdCgroupApi class should not be instantiated, and get_cgroup_api() should be used as a factory to get the correct api version instead. 

There should be no changes to existing v1 behavior with this PR (aside from some small improvements to logging and telemetry). If the agent detects legacy or a hybrid cgroup hierarchy, then the agent will use v1 api. If the agent detects unified cgroup hierarchy, then the agent will use v2. If v2 is chosen, the agent will NOT enable any usage of cgroups for the time being. 

Prior to the changes in this PR, the agent supported any v1 cgroup controller mountpoint. With these changes, the agent will only support cgroup mountpoints which are created by systemd ('/sys/fs/cgroup'). If v1 cgroups are mounted elsewhere, the agent will not enable cgroups and clean up any agent-drop in files in case this is a VM which was previously supporting agent cgroups

This PR adds an e2e test on images which use v2 by default to verify that the agent does not enable cgroup usage on these VMs. This test will be removed once the agent supports v2. 

Issue # 

---

### PR information


azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -836,6 +819,12 @@ def start_extension_command(self, extension_name, command, cmd_name, timeout, sh
                         extension_name, ustr(exception))
                     self.disable(reason, DisableCgroups.ALL)
                     # fall-through and re-invoke the extension
+                except CGroupsException as exception:
```
maddieford, line 822: Adding this as a safeguard in case SystemdCgroupsApiv2 gets used to call start_extension_command

tests/data/cgroups/sys_fs_cgroup_unified_cgroup.controllers
maddieford, line 1: cgroup.controllers file is no longer read by the agent

tests/data/cgroups/v1_and_v2/proc_pid_cgroup
maddieford, line 1: data files in /cgroups/v1_and_v2 dir represent a machine where cpu is mounted in v1, and memory is mounted in v2

tests_e2e/test_suites/cgroups_v2_disabled.yml
maddieford, line 1: I added this scenario to ensure v2 isn't enabled for agent and extensions unexpectedly

tests/lib/cgroups_tools.py
```diff
@@ -33,17 +33,3 @@ def create_legacy_agent_cgroup(cgroups_file_system_root, controller, daemon_pid)
         fileutil.append_file(os.path.join(legacy_cgroup, "cgroup.procs"), daemon_pid + "\n")
         return legacy_cgroup
 
-    @staticmethod
```
maddieford, line 36: this wasn't being used anywhere

tests/test_agent.py
```diff
@@ -240,46 +240,105 @@ def test_calls_collect_logs_with_proper_mode(self, mock_log_collector, *args):
...
+    def test_calls_collect_logs_on_valid_cgroups_v1(self, mock_log_collector):
         try:
             CollectLogsHandler.enable_monitor_cgroups_check()
             mock_log_collector.run = Mock()
 
+            # Mock cgroup paths so process is in the log collector slice
             def mock_cgroup_paths(*args, **kwargs):
                 if args and args[0] == "self":
                     relative_path = "{0}/{1}".format(cgroupconfigurator.LOGCOLLECTOR_SLICE, logcollector.CGROUPS_UNIT)
-                    return (cgroupconfigurator.LOGCOLLECTOR_SLICE, relative_path)
```
maddieford, line 251: This was incorrectly mocking the relative paths before

tests/test_agent.py
```diff
@@ -240,46 +240,105 @@ def test_calls_collect_logs_with_proper_mode(self, mock_log_collector, *args):
...
+                        try:
+                            agent.collect_logs(is_full_mode=True)
+                        except RuntimeError as re:
+                            self.assertEqual(logcollector.INVALID_CGROUPS_ERRCODE, re.args[0])
+                        mock_exit.assert_called_once_with(logcollector.INVALID_CGROUPS_ERRCODE)
+        finally:
+            CollectLogsHandler.disable_monitor_cgroups_check()
+
+    @patch("azurelinuxagent.agent.LogCollector")
+    def test_doesnt_call_collect_logs_when_controllers_mounted_in_different_hierarchies(self, mock_log_collector):
```
maddieford, line 316: log collector should only run when both cpu and memory mounted in v1

tests/ga/test_cgroupapi.py
```diff
@@ -81,79 +102,144 @@ def test_cgroups_should_be_supported_only_on_ubuntu16_centos7dot4_redhat7dot4_an
...
-            cpu_accounting = systemd.get_unit_property("walinuxagent.service", "CPUAccounting")
-
-            self.assertEqual(cpu_accounting, "no", "Property {0} of {1} is incorrect".format("CPUAccounting", "walinuxagent.service"))
+        # We expect same behavior for v1 and v2
+        mock_envs = [mock_cgroup_v1_environment(self.tmp_dir), mock_cgroup_v2_environment(self.tmp_dir)]
+        for env in mock_envs:
+            with env:
+                cpu_accounting = systemd.get_unit_property("walinuxagent.service", "CPUAccounting")
 
-    def assert_cgroups_created(self, extension_cgroups):
```
maddieford, line 123: this wasn't being used anywhere

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -166,58 +155,67 @@ def initialize(self):
                                                      agent_drop_in_file_memory_accounting, agent_drop_in_file_cpu_quota])
                             self.__cleanup_all_files(files_to_cleanup)
                             self.__reload_systemd_config()
-                            logger.info("Agent reset the quotas if distro: {0} goes from supported to unsupported list", get_distro())
+                            log_cgroup_info("Agent reset the quotas if distro: {0} goes from supported to unsupported list".format(get_distro()), send_event=False)
                     except Exception as err:
-                        logger.warn("Unable to delete Agent drop-in files while resetting the quotas: {0}".format(err))
+                        logger.warn("[CGW] Unable to delete Agent drop-in files while resetting the quotas: {0}".format(err))
```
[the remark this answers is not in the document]
narrieta: let's not mark it as CGW (cgroup warning), since it is already a warning

azurelinuxagent/ga/cgroupstelemetry.py
```diff
@@ -17,10 +17,23 @@
 import threading
 
 from azurelinuxagent.common import logger
+from azurelinuxagent.common.event import WALAEventOperation, add_event
 from azurelinuxagent.ga.cgroup import CpuCgroup
 from azurelinuxagent.common.future import ustr
 
 
+def log_cgroup_info(formatted_string, op=WALAEventOperation.CGroupsInfo, send_event=True):
```
narrieta, line 25: this should probably be in CgroupsApi. CgroupsTelemetry handles collecting the resource metrics

narrieta (approved):

narrieta (changes requested):
Sorry, I clicked Approve by mistake.

I added comments to the Agent's code. I will review the tests on a second pass

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -342,11 +397,139 @@ def start_extension_command(self, extension_name, command, cmd_name, timeout, sh
...
-        starting from version 2.2.41 we track the agent service in walinuxagent.service instead of WALinuxAgent/WALinuxAgent. If
-        we find that any of the legacy groups include the PID of the daemon then we need to disable data collection for this
-        instance (under systemd, moving PIDs across the cgroup file system can produce unpredictable results)
+        Returns True if the provided controller is enabled at the provided cgroup.
+
+        There are two ways to determine if a controller is enabled at the provided cgroup:
+
+        1. For non-leaf cgroups, the cgroup.subtree_control shows space separated list of the controllers which are
+        enabled to control resource distribution from the cgroup to its children. All non-root "cgroup.subtree_control"
+        files can only contain controllers which are enabled in the parent's "cgroup.subtree_control" file.
```
nagworld9, line 414: As you stated for the given cgroup path, controllers enabled information present in parent cgroup.subtree_control file but you are checking in current path file and also, why do we need to check for every cgroup path. Isn't this check needed only for root?

Documentation stated this

`Top-down Constraint[¶](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#top-down-constraint)
Resources are distributed top-down and a cgroup can further distribute a resource only if the resource has been distributed to it from the parent. This means that all non-root "cgroup.subtree_control" files can only contain controllers which are enabled in the parent's "cgroup.subtree_control" file. A controller can be enabled only if the parent has the controller enabled and a controller can't be disabled if one or more children have it enabled.`

Looks like root will have what controllers enabled, without here, you can't enable it in non-root cgroups

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -342,11 +397,139 @@ def start_extension_command(self, extension_name, command, cmd_name, timeout, sh
...
+        #     controlgroup_path[1:] = azure.slice/walinuxagent.service
+        controlgroup_path = systemd.get_unit_property(unit_name, "ControlGroup")
+        cpu_mount_point, memory_mount_point = self.get_cgroup_mount_points()
+
+        # Since v2 is a unified hierarchy, we need to check if each controller is enabled for the cgroup. If a
+        # controller is not enabled, then its controller interface files won't exist at the cgroup path
+        cpu_cgroup_path = None
+        if cpu_mount_point is not None:
+            cgroup_path = os.path.join(cpu_mount_point, controlgroup_path[1:])
+            if self.is_controller_enabled('cpu', cgroup_path):
```
nagworld9, line 481: why do we need to check this here? This method should return what api returns?

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -342,11 +397,139 @@ def start_extension_command(self, extension_name, command, cmd_name, timeout, sh
...
+    def get_process_cgroup_paths(self, process_id):
+        cpu_cgroup_relative_path, memory_cgroup_relative_path = self.get_process_cgroup_relative_paths(process_id)
+        cpu_mount_point, memory_mount_point = self.get_cgroup_mount_points()
+
+        # Since v2 is a unified hierarchy, we need to check if each controller is enabled for the cgroup. If a
+        # controller is not enabled, then its controller interface files won't exist at the cgroup path
+        cpu_cgroup_path = None
+        if cpu_mount_point is not None and cpu_cgroup_relative_path is not None:
+            cgroup_path = os.path.join(cpu_mount_point, cpu_cgroup_relative_path)
+            if self.is_controller_enabled('cpu', cgroup_path):
```
nagworld9, line 501: same thing here

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -342,11 +397,139 @@ def start_extension_command(self, extension_name, command, cmd_name, timeout, sh
...
+        for line in fileutil.read_file("/proc/{0}/cgroup".format(process_id)).splitlines():
+            match = re.match(r'\d+::(?P<path>\S+)', line)
+            if match is not None:
+                path = match.group('path').lstrip('/') if match.group('path') != '/' else None
+                memory_path = path
+                cpu_path = path
+
+        return cpu_path, memory_path
+
+    def start_extension_command(self, extension_name, command, cmd_name, timeout, shell, cwd, env, stdout, stderr, error_code=ExtensionErrorCodes.PluginUnknownFailure):   # pylint: disable=W0613
```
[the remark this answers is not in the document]
nagworld9: This call has protection and only called if cgroups enabled. I believe we don't enable for v2 now, this won't be called

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -166,58 +155,67 @@ def initialize(self):
...
 
                 # check whether cgroup monitoring is supported on the current distro
                 self._cgroups_supported = CGroupsApi.cgroups_supported()
                 if not self._cgroups_supported:
-                    logger.info("Cgroup monitoring is not supported on {0}", get_distro())
+                    log_cgroup_info("Cgroup monitoring is not supported on {0}".format(get_distro()), send_event=False)
```
nagworld9, line 165: I feel this is also informant info and should send this. It just one time log(on service start)

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -166,58 +155,67 @@ def initialize(self):
...
                 self.__setup_azure_slice()
 
-                cpu_controller_root, memory_controller_root = self.__get_cgroup_controllers()
-                self._agent_cpu_cgroup_path, self._agent_memory_cgroup_path = self.__get_agent_cgroups(agent_slice,
+                cpu_controller_root, memory_controller_root = self.__get_cgroup_controllers_mount_points()
+                self._agent_cpu_cgroup_path, self._agent_memory_cgroup_path = self.__get_agent_cgroup_paths(agent_slice,
                                                                                                        cpu_controller_root,
                                                                                                        memory_controller_root)
 
+                if self.cgroup_v2_enabled():
```
nagworld9, line 197: One thing I noticed when I was following cgroup issues is that we create slices first and later we check whether we need to enable the cgroups for monitoring and enforcement. Just pointing that out, at some point we may need to think about this strategy.

azurelinuxagent/ga/cgroupconfigurator.py
nagworld9: Not needed, protected against enable flag

azurelinuxagent/ga/cgroupapi.py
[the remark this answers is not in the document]
maddieford: It won't be called, I added this as a safeguard for my own sanity. Particularly if something unexpected happens with enable logic in the configurator in future PRs when we enable the log collector on v2. 

Right now it won't be called. If you both think it's best to remove and just raise NotImplementedError, I'll update this

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -342,11 +397,139 @@ def start_extension_command(self, extension_name, command, cmd_name, timeout, sh
...
+                cgroup.freeze       cgroup.procs            cpu.idle        cpu.uclamp.max  memory.current       memory.max        memory.pressure   memory.swap.high     pids.current
+                cgroup.kill         cgroup.stat             cpu.max         cpu.uclamp.min  memory.events        memory.min        memory.reclaim    memory.swap.max      pids.events
+                cgroup.max.depth    cgroup.subtree_control  cpu.max.burst   cpu.weight      memory.events.local  memory.numa_stat  memory.stat       memory.swap.peak     pids.max
+
+        If either check is True, the controller is enabled at the cgroup. Check 1 is necessary because no controller
+        interface files exist at the root cgroup, even if the controller is enabled.
         """
-        return CGroupsApi._foreach_legacy_cgroup(lambda *_: None)
+        if cgroup_path is not None and controller is not None:
+            # Check that the controller is enabled in the cgroup.subtree_control file
```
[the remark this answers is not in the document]
maddieford: Comment should probably be updated. There isn't any logic to decide between the two. If the first condition isn't true then we fall through to the second condition. Second condition can be applied for leaf or non-leaf.

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -342,11 +397,139 @@ def start_extension_command(self, extension_name, command, cmd_name, timeout, sh
...
+            enabled_controllers_file = os.path.join(cgroup_path, 'cgroup.subtree_control')
+            if os.path.exists(enabled_controllers_file):
+                enabled_controllers = fileutil.read_file(enabled_controllers_file).rstrip().split(" ")
+                if controller in enabled_controllers:
+                    return True
+
+            # Check that the controller interface files exist in the cgroup
+            if os.path.exists(cgroup_path):
```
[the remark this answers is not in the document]
maddieford: cpu.stat is a special case. From controllers documentation, it's the only controller interface file which exists whether controller enabled or not.

Also CpuAccounting is a bit different in v2. From systemd documentation:
> 
> CPUAccounting=
> Turn on CPU usage accounting for this unit. Takes a boolean argument. Note that turning on CPU accounting for one unit will also implicitly turn it on for all units contained in the same slice and for all its parent slices and the units contained therein. The system default for this setting may be controlled with DefaultCPUAccounting= in [systemd-system.conf(5)](https://man.archlinux.org/man/systemd-system.conf.5.en).
> **Under the unified cgroup hierarchy, CPU accounting is available for all units and this setting has no effect.**

When we specify CPUAccounting=yes in slice/drop-in files in v2, the cpu controller isn't enabled. CPU controller will only be enabled in our use case once CPUQuota is set. cpu usage is available for all units, so I figured it was safe that we consider it enabled here since cpu.stat is available. But I understand it may lead to confusion

If this is removed, cgroupconfigurator would need to have some other logic to determine if cpu controller is available and cpu quota could be set.

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -342,11 +397,139 @@ def start_extension_command(self, extension_name, command, cmd_name, timeout, sh
...
+                for item in os.listdir(cgroup_path):
+                    if item.startswith(controller + '.'):
+                        return True
+
+        return False
+
+    def get_cgroup_mount_points(self):
```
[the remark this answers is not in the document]
maddieford: maybe rename to get_cgroup_controller_mount_points

In v2 api, get_cgroup_controller_mount_points just calls some get_cgroup_root_mount_point

azurelinuxagent/ga/cgroupapi.py
maddieford: If controller is in cgroup.subtree_control file then we know it is enabled in the current cgroup and any of its children due to top down constraint. That's why I had that check. If this is unclear then maybe cgroup.controllers is a better check for non-root cgroups, and this method would be better divided by root vs non-root instead of leaf vs non-leaf.

Answer to the second question in your other comments

azurelinuxagent/ga/cgroupapi.py
maddieford: the v1 api gives us different paths for each controller in use by the cgroup. the v2 api result doesn't tell us if the controller is enabled for the cgroup, because it just returns one path at which any controller might be enabled. 

later we use this method to get the cgroup_paths for tracking. if cpu/memory isn't enabled, then we initialize CpuCgroup or MemoryCgroup with path that does have corresponding counter interface files.

nagworld9: we call these paths for tracking and other purposes once we start usingthe cgroups for v2 that means we enable the controllers already. why do we need extra checks? or maybe I'm missing something here.

azurelinuxagent/ga/cgroupapi.py
nagworld9: For example, if subtree_contorl is empty in agent service cgroup directory but controllers enabled in azure.slice, then this fails and says that controllers are not enabled when it's enabled

maddieford: It won't fail since it checks for existence of <controller>.* interface files, which only exist if controller is enabled

azurelinuxagent/ga/cgroupapi.py
maddieford: we attempt to enable the controllers by adding drop-in files, updating slice, etc., but we don't have an explicit check that they were successfully enabled and that the controllers are in use

I think there should be an explicit check, if not here then somewhere else. I added the check here because in v1 we know the cgroup is in cpu/memory because path is provided. in v2 we're not guaranteed the same

azurelinuxagent/ga/cgroupapi.py
maddieford: I agree cgroup.controllers would be clearer check though

nagworld9: I just want to make sure what I speak correct here. WhenI check doc for cgroup.controller, it just says what controllers available to enable but it does not say what enabled. 

`Each cgroup has a "cgroup.controllers" file which lists all controllers available for the cgroup to enable:
`
If we want to use this, then we need to check is this file containing all the controllers minus what's mounted in v1? or will it show everything regardless v1 status?

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -37,10 +37,32 @@
...
+    the corresponding Api. If the required controllers are not mounted in v1 or v2, return None.
+    """
+    v1 = SystemdCgroupsApiv1()
+    v2 = SystemdCgroupsApiv2()
+
+    log_cgroup_info("Controllers mounted in v1: {0}. Controllers mounted in v2: {1}".format(v1.get_mounted_controllers(), v2.get_mounted_controllers()))
+
+    # It is possible for different controllers to be simultaneously mounted under v1 and v2. If any are mounted under
+    # v1, use v1.
+    if v1.is_cpu_or_memory_mounted():
```
nagworld9, line 55: is this supposed to be and? we use v1 if both cpu and memory mounted?

I just thought one scenario we use v1 for cpu, and cx enabled memory in v2 for their application. Then we start using v1 for memory then that will create discprencies

azurelinuxagent/ga/cgroupapi.py
maddieford: If there is a controller (cpu for example) mounted in v1, then it will not be mounted in v2. Controllers cannot be mounted in both hierarchies simultaneously, so it wouldn't appear in the list of available controllers for v2. 

cgroup.controllers file lists all controllers available for the cgroup to enable to its children. A cgroup can only enable a controller to its children if it is enabled in the current cgroup itself, due to top down constraint.

azurelinuxagent/ga/cgroupconfigurator.py
narrieta, line 165: There are 2 checks for CGroupsApi.cgroups_supported(); one here and one on line 143. Can we merge them?

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -166,58 +155,67 @@ def initialize(self):
...
                 self._cgroups_supported = CGroupsApi.cgroups_supported()
                 if not self._cgroups_supported:
-                    logger.info("Cgroup monitoring is not supported on {0}", get_distro())
+                    log_cgroup_info("Cgroup monitoring is not supported on {0}".format(get_distro()), send_event=False)
+                    return
+
+                # Determine which version of the Cgroup API should be used. If the correct version can't be determined,
+                # do not enable resource monitoring/enforcement.
+                self._cgroups_api = get_cgroup_api()
+                if self._cgroups_api is None:
```
narrieta, line 171: let's log and send telemetry

azurelinuxagent/ga/cgroupapi.py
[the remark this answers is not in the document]
narrieta: if you feel the protection is needed, my comment above applies. If you think it is not needed, I'd remove the code above not provide an implementation (i.e. raise NotImplemented). To safeguard your sanity, you could add a unit test that verifies it is not called.

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -136,23 +158,106 @@ def get_systemd_run_commands(self):
...
         #     etc
         #
-        if self._cgroup_mountpoints is None:
+        if not self._cgroup_mountpoints:
             cpu = None
             memory = None
-            for line in shellutil.run_command(['mount', '-t', 'cgroup']).splitlines():
-                match = re.search(r'on\s+(?P<path>/\S+(memory|cpuacct))\s', line)
+            for line in shellutil.run_command(['findmnt', '-t', 'cgroup', '--noheadings']).splitlines():
+                match = re.search(r'(?P<path>/\S+(memory|cpuacct))\s', line)
```
narrieta, line 260: on line 263 we should probably be looking for "cpu,cpuacct", rather than just "cpuacct"

azurelinuxagent/ga/cgroupapi.py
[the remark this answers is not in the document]
narrieta: "mount point" was appropriate for v1, since the controllers are mounted independently of each other. This is basically the path of the root cgroup, maybe we should change to that?

get_root_cgroups, etc ?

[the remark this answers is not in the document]
narrieta: Also, the code for v1 is overcomplicated because it keeps track of the paths for each controller separately. This causes a lot of code repetition in the code, logs & telemetry, making the code and logs harder to read.

v2 has a simpler model. 

Instead of making v2 as complicated as v1, shouldn't we make v1 as simple as v2? Can you explore this option?

Initially the code was using the paths for each controller separately, since it was using the file system API directly and, for example, it needed to write the PID of the Agent to both the cpu and memory paths. Now we are using the systemd API and we work in terms of slices and scopes, which are 1 single entity (instead of 1 per controller). I think we use the actual filesystem paths only to collect the metrics.

It has been proposed that we should also monitor io and maybe other metrics. If we keep in the path of tracking each controller separately, this is going to get ever more complicated that it already is. v2 is a good chance to simplify this

azurelinuxagent/ga/cgroupapi.py
narrieta: Seems like we simplify this asking for the file system type

```If you wonder how to detect which of these three modes is currently used, use statfs() on /sys/fs/cgroup/. If it reports CGROUP2_SUPER_MAGIC in its .f_type field, then you are in unified mode. If it reports TMPFS_MAGIC then you are either in legacy or hybrid mode. To distinguish these two cases, run statfs() again on /sys/fs/cgroup/unified/. If that succeeds and reports CGROUP2_SUPER_MAGIC you are in hybrid mode, otherwise not. From a shell, you can check the Type in stat -f /sys/fs/cgroup and stat -f /sys/fs/cgroup/unified.```

Then if unified use v2, else v1.

Results from ubuntu 20 (hybrid) and 22 (unified)

```
root@nam-u20:/home/nam# stat -f --format=%T /sys/fs/cgroup
tmpfs

root@nam-ubuntu22:/home/nam# stat -f --format=%T /sys/fs/cgroup
cgroup2fs
```

File: azurelinuxagent/ga/cgroupconfigurator.py [hunk 20]
```diff
@@ -166,58 +155,67 @@ def initialize(self):
...
+                        logger.warn("[CGW] Unable to delete Agent drop-in files while resetting the quotas: {0}".format(err))
 
                 # check whether cgroup monitoring is supported on the current distro
                 self._cgroups_supported = CGroupsApi.cgroups_supported()
                 if not self._cgroups_supported:
-                    logger.info("Cgroup monitoring is not supported on {0}", get_distro())
+                    log_cgroup_info("Cgroup monitoring is not supported on {0}".format(get_distro()), send_event=False)
+                    return
+
+                # Determine which version of the Cgroup API should be used. If the correct version can't be determined,
```
narrieta, line 168: the check for v1 v2 v2 should come after the check for systemd on line 175 below

azurelinuxagent/ga/cgroupapi.py
narrieta: I am not sure we need this check other than for the root cgroup. I think we need to check whether cpu & memory are enabled (cgroup.subtree_control) on root. If they are enabled, we should use them. If not. we should not enable them. at least in the initial releases -- we can reconsider this later.

We own the non-root cgroups (walinuxagent.service, azure-walinuxagent-logcollector.slice, azure-vmextensions.slice, azure-vmextensions-Extension.Foo-1.0.0.0.slice) and we should enable the controllers we want to use on those cgroups.

azurelinuxagent/ga/cgroupapi.py
maddieford: will remove this check and only check if controllers are enabled at root

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -342,11 +397,78 @@ def start_extension_command(self, extension_name, command, cmd_name, timeout, sh
...
+class SystemdCgroupApiv2(_SystemdCgroupApi):
+    """
+    Cgroup v2 interface via systemd
+    """
+    def __init__(self):
+        super(SystemdCgroupApiv2, self).__init__()
+        self._root_cgroup_path = None
+        self._controllers_enabled_at_root = []
+
+    def is_controller_enabled_at_root(self, controller):
```
narrieta, line 410: looks like this can be marked as private

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -342,11 +397,78 @@ def start_extension_command(self, extension_name, command, cmd_name, timeout, sh
...
-    def cleanup_legacy_cgroups(self):
+
+class SystemdCgroupApiv2(_SystemdCgroupApi):
+    """
+    Cgroup v2 interface via systemd
+    """
+    def __init__(self):
+        super(SystemdCgroupApiv2, self).__init__()
+        self._root_cgroup_path = None
+        self._controllers_enabled_at_root = []
```
narrieta, line 408: Seems like these two properties are initialized as a side effect of calling get_controller_root_paths(). That does not seems correct. For example, if somebody calls is_controller_enabled_at_root() before calling get_controller_root_paths() then self._root_cgroup_path won't be initialized.

azurelinuxagent/ga/cgroupapi.py
narrieta, line 410: Looks like this can be marked as private

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -103,28 +109,78 @@ def _foreach_legacy_cgroup(operation):
...
+    Uses 'stat -f --format=%T /sys/fs/cgroups' to get the cgroup hierarchy in use.
+        If the result is 'cgroup2fs', cgroup v2 is being used.
+        If the result is 'tmpfs', cgroup v1 or a hybrid mode is being used.
+            If the result of 'stat -f --format=%T /sys/fs/cgroup/unified' is 'cgroup2fs', then hybrid mode is being used.
+
+    Raises exception if an unknown mode is detected. Also raises exception if hybrid mode is detected and there are
+    controllers available to be enabled in the unified hierarchy (the agent does not support cgroups if there are
+    controllers simultaneously attached to v1 and v2 hierarchies).
+    """
+    root_hierarchy_mode = shellutil.run_command(["stat", "-f", "--format=%T", CGROUP_FILE_SYSTEM_ROOT]).rstrip()
```
nagworld9, line 156: we assumed the root file system will be /sys/fs/cgroup. But one can mount the cgroup filesystem anywhere right, so according to this check if its not found in expected path we don't use cgroups? will this check fails if path not found?

nagworld9: The other piece that is contradicting is we don't assume root path there and finding root path with mount command in each cgroups api implementation.

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -166,94 +156,91 @@ def initialize(self):
...
                 self.__setup_azure_slice()
 
-                cpu_controller_root, memory_controller_root = self.__get_cgroup_controllers()
-                self._agent_cpu_cgroup_path, self._agent_memory_cgroup_path = self.__get_agent_cgroups(agent_slice,
+                cpu_controller_root, memory_controller_root = self.__get_cgroup_controller_roots()
+                self._agent_cpu_cgroup_path, self._agent_memory_cgroup_path = self.__get_agent_cgroup_paths(agent_slice,
                                                                                                        cpu_controller_root,
                                                                                                        memory_controller_root)
 
+                if self.cgroup_v2_enabled():
```
nagworld9, line 197: Before we don't have this sort of implementation, we don not enable cgroups v2 based on the controller's check. Now do we need to run 193, can we move this check above?

azurelinuxagent/ga/cgroupapi.py
maddieford: I'll be adding a check that the path exists. If not, we won't use cgroups since we expect systemd to mount cgroup filesystem there. We'll collect telemetry so we can get an idea for how many VMs are mounting the cgroup filesystem elsewhere.

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -103,28 +109,84 @@ def _foreach_legacy_cgroup(operation):
...
+
+    Raises exception if cgroup filesystem mountpoint is not '/sys/fs/cgroup', or an unknown mode is detected. Also
+    raises exception if hybrid mode is detected and there are controllers available to be enabled in the unified
+    hierarchy (the agent does not support cgroups if there are controllers simultaneously attached to v1 and v2
+    hierarchies).
     """
+    if not os.path.exists(CGROUP_FILE_SYSTEM_ROOT):
+        v1_mount_point = shellutil.run_command(['findmnt', '-t', 'cgroup', '--noheadings'])
+        v2_mount_point = shellutil.run_command(['findmnt', '-t', 'cgroup2', '--noheadings'])
+        raise CGroupsException("Expected cgroup filesystem to be mounted at '/sys/fs/cgroup', but it is not.\n v1 mount point: {0}\n v2 mount point: {1}".format(v1_mount_point, v2_mount_point))
```
narrieta, line 160: Minor: for v1 it will be a list of mountpoints, let's add a newline v1 mount point: **\n**{0}\n. Maybe the same for v2 for consistency.

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -103,28 +109,99 @@ def _foreach_legacy_cgroup(operation):
...
+                available_unified_controllers = fileutil.read_file(available_unified_controllers_file).rstrip()
+                if available_unified_controllers != "":
+                    raise CGroupsException("Detected hybrid cgroup mode, but there are controllers available to be enabled in unified hierarchy: {0}".format(available_unified_controllers))
+
+        cgroup_api = SystemdCgroupApiv1()
+        # Previously the agent supported users mounting cgroup v1 controllers in locations other than the systemd
+        # default ('/sys/fs/cgroup'). The agent no longer supports this scenario. If either the cpu or memory
+        # controller is mounted in a location other than the systemd default, raise Exception.
+        if not cgroup_api.are_mountpoints_systemd_created():
+            raise InvalidCgroupMountpointException("Expected cgroup controllers to be mounted at '{0}', but at least one is not. v1 mount points: \n{1}".format(CGROUP_FILE_SYSTEM_ROOT, json.dumps(cgroup_api.get_controller_root_paths())))
```
narrieta, line 193: we want to do the cleanup only for v1, but InvalidCgroupMountpointException may also be raised for other reasons and the caller has no way to know why it was raised. maybe use a different exception? InvalidV1CgroupMountpointException

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -342,11 +440,85 @@ def start_extension_command(self, extension_name, command, cmd_name, timeout, sh
...
+        #
+        for line in shellutil.run_command(['findmnt', '-t', 'cgroup2', '--noheadings']).splitlines():
+            # Systemd mounts the cgroup filesystem at '/sys/fs/cgroup'. The agent does not support cgroups if the
+            # filesystem is mounted elsewhere, so search specifically for '/sys/fs/cgroup' in the findmnt output.
+            match = re.search(r'(?P<path>\/sys\/fs\/cgroup)\s+cgroup2', line)
+            if match is not None:
+                root_cgroup_path = match.group('path')
+                if root_cgroup_path is not None:
+                    return root_cgroup_path
+        return None
```
narrieta, line 473: Using None as a special value makes the code harder to writer. Every piece of code that uses self._root_cgroup_path  needs to check for None. Please reconsider this design. Worst case, add a comment in init marking very clearly that self._root_cgroup_path  can be None

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -342,11 +440,85 @@ def start_extension_command(self, extension_name, command, cmd_name, timeout, sh
...
+
+    def _get_controllers_enabled_at_root(self):
```
narrieta, line 475: this depends on _get_root_cgroup_path being called first, in order to self._root_cgroup_path to be initialized. I'd make that dependency explicit by making this a static method and passing the path as argument, instead of referencing self._root_cgroup_path,

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -342,11 +440,85 @@ def start_extension_command(self, extension_name, command, cmd_name, timeout, sh
...
+        return root_cpu_path, root_memory_path
+
+    def get_process_cgroup_relative_paths(self, process_id):
+        # The contents of the file are similar to
+        #    # cat /proc/1218/cgroup
+        #    0::/azure.slice/walinuxagent.service
+        cpu_path = None
+        memory_path = None
+        for line in fileutil.read_file("/proc/{0}/cgroup".format(process_id)).splitlines():
+            match = re.match(r'\d+::(?P<path>\S+)', line)
```
narrieta, line 514: you should probably match 0 rather than \d+

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -478,6 +469,9 @@ def agent_enabled(self):
         def extensions_enabled(self):
             return self._extensions_cgroups_enabled
 
+        def cgroup_v2_enabled(self):
```
narrieta, line 472: rename to using_cgroup_v2 or similar?

currently cgroup_v2_enabled() can return True when enabled() returns False. That is just weird.

azurelinuxagent/ga/cgroupapi.py
maddieford: InvalidCgroupMountpointException is only raised if v1 controllers are not systemd default or if /sys/fs/cgroup path does not exist 

I think we want to do cleanup in both cases. Previously we supported users not using /sys/fs/cgroup. So if the path doesn't exist, we should do agent drop in cleanup in case we were supporting cgroups at a different path before

azurelinuxagent/ga/cgroupapi.py
maddieford: There are a few places in the current cgroup implementation that return None instead of empty string when looking for paths. For example, get_process_cgroup_relative_path. All callers currently check for None. 

I'll leave it as None and add a comment in init for consistency. When I do the controller abstraction in a future PR I will revisit this.

azurelinuxagent/ga/cgroupapi.py
narrieta: `or if /sys/fs/cgroup path does not exist
`
this could happen under v2 and there is no need to cleanup in that case

azurelinuxagent/ga/cgroupapi.py
narrieta: `When I do the controller abstraction in a future PR I will revisit this.`

Please do. Continue to use bad patterns on new code is not good, but I understand why to postpone this (as long as it gets done)

azurelinuxagent/ga/cgroupapi.py
maddieford: That's true but I think it is safer to also cleanup those files on v2 machines which meet this condition than to keep the files on v1 machines which meet this condition.

narrieta: do you mean we have been creating those files under v2 as well? i thought this cleanup was v1-specific

maddieford: yes drop in files are currently being created on v2 machines, with the exception of the CPUQuota drop-in file. This is because in cgroupconfigurator.initialize the azure slice/drop-in files are created (__setup_azure_slice) before the controller mountpoints are checked. The CPUQuota drop in file is only created after the controller mountpoints are checked. As a result, v2 machines have cpu/memory accounting drop-in files. 

Since quota drop-in file is not created on v2 machines, then it is not necessary to cleanup drop-in files on v2 (in my opinion). But, if /sys/fs/cgroup path does not exist, then we don't know whether the machine is using v1 or v2, so we should clean up the drop-in files regardless. If the machine is using v1, we should definitely clean up the drop in files since quota could exist. If the machine is using v2, there is no harm in cleaning up the drop- in files.

maddieford: BTW with this PR, the azure slice/drop in files are not created until after we check if v2 is in use. So on new machines with this update, drop-in files won't be created if v2 is in use.

narrieta: ok, sounds good then, thanks

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -150,39 +139,58 @@ def initialize(self):
...
                 if agent_slice not in (AZURE_SLICE, "system.slice"):
-                    _log_cgroup_warning("The agent is within an unexpected slice: {0}", agent_slice)
+                    log_cgroup_warning("The agent is within an unexpected slice: {0}".format(agent_slice))
                     return
 
                 self.__setup_azure_slice()
 
-                cpu_controller_root, memory_controller_root = self.__get_cgroup_controllers()
-                self._agent_cpu_cgroup_path, self._agent_memory_cgroup_path = self.__get_agent_cgroups(agent_slice,
+                if self.cgroup_v2_enabled():
```
nagworld9, line 188: `BTW with this PR, the azure slice/drop in files are not created until after we check if v2 is in use. So on new machines with this update, drop-in files won't be created if v2 is in use.`

@maddieford  you made this comments, do you have a plans to change this condition and place above `self.__setup_azure_slice()` As of now it's not matching what you said.

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -342,11 +440,85 @@ def start_extension_command(self, extension_name, command, cmd_name, timeout, sh
...
+
+                $ cat /sys/fs/cgroup/cgroup.subtree_control
+                cpuset cpu io memory hugetlb pids rdma misc
         """
-        return CGroupsApi._foreach_legacy_cgroup(lambda *_: None)
+        controllers_enabled_at_root = []
+        if self._root_cgroup_path is not None:
+            enabled_controllers_file = os.path.join(self._root_cgroup_path, 'cgroup.subtree_control')
+            if os.path.exists(enabled_controllers_file):
+                controllers_enabled_at_root = fileutil.read_file(enabled_controllers_file).rstrip().split(" ")
```
nagworld9, line 489: This only handle single space on split, should we consider multiple whitespaces and just call split without delimiter?

azurelinuxagent/ga/cgroupconfigurator.py
maddieford: You're right, I thought I moved it. I'll move it in next commit

azurelinuxagent/ga/cgroupapi.py
maddieford: systemd mentions it is a space separated list. Anything else would be unexpected. 

I'll remove delimiter to be safe

narrieta (approved):

nagworld9 (approved):

PR #3124, opened 2024-05-10
Reduce the frequency of firewall telemetry

narrieta:
Report the first 5 instance of ResetFirewall as they occur, then report at most every 3 hours.

Also changed the Message field to be more concise and easier to read.

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -266,16 +266,16 @@ class DefaultOSUtil(object):
 
     def enable_firewall(self, dst_ip, uid):
         """
-        It checks if every iptable rule exists and add them if not present. It returns a tuple(enable firewall success status, update rules flag)
+        It checks if every iptable rule exists and add them if not present. It returns a tuple(enable firewall success status, missing rules array)
         enable firewall success status: Returns True if every firewall rule exists otherwise False
-        update rules flag: Returns True if rules are updated otherwise False
+        missirg rules: array with names of the missing rules ("ACCEPT DNS", "ACCEPT", "DROP")
         """
-        # This is to send telemetry when iptable rules updated
-        is_firewall_rules_updated = False
         # If a previous attempt failed, do not retry
         global _enable_firewall  # pylint: disable=W0603
         if not _enable_firewall:
-            return False, is_firewall_rules_updated
+            return False, []
+
+        missing_rules = []
 
         try:
             wait = self.get_firewall_will_wait()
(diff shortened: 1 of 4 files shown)


azurelinuxagent/common/osutil/default.py
```diff
@@ -266,27 +266,27 @@ def remove_legacy_firewall_rule(self, dst_ip):
 
     def enable_firewall(self, dst_ip, uid):
         """
-        It checks if every iptable rule exists and add them if not present. It returns a tuple(enable firewall success status, update rules flag)
+        It checks if every iptable rule exists and add them if not present. It returns a tuple(enable firewall success status, missing rules array)
```
narrieta, line 269: Instead of returning a boolean indicating that the rules were updated, now this method returns an array containing the rules that were missing when the firewall was updated.

azurelinuxagent/common/osutil/default.py
```diff
@@ -266,27 +266,27 @@ def remove_legacy_firewall_rule(self, dst_ip):
...
         enable firewall success status: Returns True if every firewall rule exists otherwise False
-        update rules flag: Returns True if rules are updated otherwise False
+        missirg rules: array with names of the missing rules ("ACCEPT DNS", "ACCEPT", "DROP")
```
maddieford, line 271: Maybe we should clarify in a comment here that these are the missing rules before reapply operation.

azurelinuxagent/common/osutil/default.py
```diff
@@ -266,27 +266,27 @@ def remove_legacy_firewall_rule(self, dst_ip):
...
         """
-        # This is to send telemetry when iptable rules updated
-        is_firewall_rules_updated = False
         # If a previous attempt failed, do not retry
         global _enable_firewall  # pylint: disable=W0603
```
maddieford, line 274: Just curious, where is this global var instantiated?

azurelinuxagent/common/osutil/default.py
narrieta: any suggestions? "It checks if every iptable rule exists" above seems enough to me

azurelinuxagent/common/osutil/default.py
narrieta: @ line 113

azurelinuxagent/common/osutil/default.py
maddieford: something like?
```suggestion
        missing rules: array with names of the missing rules before attempt to add them ("ACCEPT DNS", "ACCEPT", "DROP")
```
I was just thinking of case where we attempt to re-add the rules and some get added but agent fails to add others. Then we should clarify if the missing rules array was created before or after that attempt.

It's clear from the code though, so might be overkill to add the comment.

narrieta: OK, I'll leave it as is. Anyways, all this code is already changed in the fix for nftables and this return value is no longer needed.

maddieford (dismissed):

tests/common/osutil/test_default.py
```diff
@@ -822,13 +822,13 @@ def test_enable_firewall_should_not_use_wait_when_iptables_does_not_support_it(s
                 success, _ = osutil.DefaultOSUtil().enable_firewall(dst_ip=mock_iptables.destination, uid=mock_iptables.uid)
 
                 self.assertTrue(success, "Enabling the firewall was not successful")
-                # Exactly 8 calls have to be made.
-                # First check rule, delete 4 rules,
+                # Exactly 10 calls have to be made.
```
narrieta, line 825: Previously verify_iptables_rules_exist (now named get_missing_iptables_rules) would return on the first failure; now it always checks all the rules to find all the missing ones. The call count needs to be updated.

maddieford (approved):

PR #3133, opened 2024-05-30
Telemetry pipeline improvements (iteration 1)

nagworld9:
## Description

This pr address below issues to improve data reliability

1. _Sending important events_: Implemented as sending this event directly to wireserver and bypass the collecton_and_send channel. Added new parameter immediate_flush to add_event. Callers need to set to true if this is important event and needs to flush immediately. As a fallback, if we fail to send, we will save to disk for collector to pick up as like a normal event.
2. _Throttling issues/retries on telemetrydata API_:  Today we have aggressive retries on throttling errors. This fix will reduce the number of retries with better delay.
3. _Retry on reading event files_: Collector may pick up and read the file while file is modified. 

I added necessary comments, still questions/concerns happy to go over a call. 

Issue # 

---

### PR information


tests/ga/test_update.py
```diff
@@ -106,17 +106,20 @@ def _get_update_handler(iterations=1, test_data=None, protocol=None, autoupdate_
     This function returns a mocked version of the UpdateHandler object to be used for testing. It will only run the
     main loop [iterations] no of times.
     """
+    # Added try-finally block to fix false positive pylint warning contextmanager-generator-missing-cleanup
```
nagworld9, line 109: pylint has new warning even on develop branch https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/contextmanager-generator-missing-cleanup.html

azurelinuxagent/common/event.py
```diff
@@ -548,11 +544,26 @@ def add_metric(self, category, counter, instance, value, log_event=False):
...
-            logger.periodic_error(logger.EVERY_FIFTEEN_MINUTES, "[PERIODIC] {0}".format(ustr(e)))
+        self.flush_or_save_event(event, message, immediate_flush)
+
+    def flush_or_save_event(self, event, message, immediate_flush):
+        """
+        Flush the event to wireserver if immediate_flush to set to true, else
+        save it disk if we fail to send or not required to flush immediately.
+        """
+        report_success = immediate_flush
+        if immediate_flush and self.protocol is not None:
```
maddieford, line 555: if self.protocol is None and immediate_flush is True then we never report the event. Is this intentional?

azurelinuxagent/common/event.py
```diff
@@ -666,18 +678,20 @@ def report_metric(category, counter, instance, value, log_event=False, reporter=
...
     try:
-        reporter.add_metric(category, counter, instance, float(value), log_event)
+        reporter.add_metric(category, counter, instance, float(value), log_event, immediate_flush)
     except ValueError:
         logger.periodic_warn(logger.EVERY_HALF_HOUR, "[PERIODIC] Cannot cast the metric value. Details of the Metric - "
                                                      "{0}/{1} [{2}] = {3}".format(category, counter, instance, value))
 
 
-def initialize_event_logger_vminfo_common_parameters(protocol, reporter=__event_logger__):
+def initialize_event_logger_vminfo_common_parameters_and_protocal(protocol, reporter=__event_logger__):
```
maddieford, line 687:
```suggestion
def initialize_event_logger_vminfo_common_parameters_and_protocol(protocol, reporter=__event_logger__):
```

azurelinuxagent/common/protocol/wire.py
```diff
@@ -62,6 +63,14 @@
 
 _DOWNLOAD_TIMEOUT = timedelta(minutes=5)
 
+# telemetrydata api max calls per 15 secs
+TELEMETRY_MAX_CALLS_PER_INTERVAL = 12
```
maddieford, line 67: PR description indicates max is 15 calls per 15 seconds. Why 12 here?

maddieford: nevermind I see explanation below. Maybe add another comment here too

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1033,6 +1043,20 @@ def report_health(self, status, substatus, description):
...
+        Construct the encoded event and url for telemetry api call
+        Before calling telemetry api, ensure calls are under throttling limits and checks if the number of telemetry api calls 12 in the last 15 seconds
+        (Considered 12 instead actual limit 15 as a conservative approach plus it helps for directly flushing events not hit the throttling issues)
+        (Trade off between delay vs successful event delivery for immediate_flush events)
+        Note: throttling limit is 15 calls in 15 seconds
+        """
+        def can_make_wireserver_call():
+            current_time = datetime.utcnow()
+            fifteen_seconds_ago = current_time - timedelta(seconds=TELEMETRY_INTERVAL)
+            while self.telemetry_api_calls_timestamps and self.telemetry_api_calls_timestamps[0] < fifteen_seconds_ago:
```
maddieford, line 1056: ` while self.telemetry_api_calls_timestamps`

is this checking that the queue is empty? maybe add explicity length check for readability?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1033,6 +1043,20 @@ def report_health(self, status, substatus, description):
...
+        """
+        Construct the encoded event and url for telemetry api call
+        Before calling telemetry api, ensure calls are under throttling limits and checks if the number of telemetry api calls 12 in the last 15 seconds
+        (Considered 12 instead actual limit 15 as a conservative approach plus it helps for directly flushing events not hit the throttling issues)
+        (Trade off between delay vs successful event delivery for immediate_flush events)
+        Note: throttling limit is 15 calls in 15 seconds
+        """
+        def can_make_wireserver_call():
+            current_time = datetime.utcnow()
+            fifteen_seconds_ago = current_time - timedelta(seconds=TELEMETRY_INTERVAL)
```
maddieford, line 1055: nit: change this variable name to 'interval_start_timestamp' or something similar in case telemetry interval changes from 15seconds in the future.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1041,10 +1065,23 @@ def send_encoded_event(self, provider_id, event_str, encoding='utf8'):
...
             # NOTE: The call to wireserver requests utf-8 encoding in the headers, but the body should not
             #       be encoded: some nodes in the telemetry pipeline do not support utf-8 encoding.
-            resp = self.call_wireserver(restutil.http_post, uri, data, header)
+            header = self.get_header_for_xml_content()
+
+            # wait until throttling limit reset to make next call
+            while not can_make_wireserver_call():
+                next_call_time = self.telemetry_api_calls_timestamps[0] + timedelta(seconds=TELEMETRY_INTERVAL)
+                logger.verbose("Reached telemetry api throttling limit: {0}, so waiting to make next call after : {1}".format(TELEMETRY_MAX_CALLS_PER_INTERVAL, next_call_time))
+                time.sleep(TELEMETRY_DELAY)
```
maddieford, line 1076: this sleeps for 1s, shouldn't it continue sleep until 'next_call_time'?

nagworld9: well it will sleep and check and eventually it will wait until next time. Initially total seconds computation not available in 2.6 to sleep until_next_time.  Now i found we have done manual computation in periodic operation, so using same formula

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1033,6 +1044,20 @@ def report_health(self, status, substatus, description):
                                                       resp.read()))
 
     def send_encoded_event(self, provider_id, event_str, encoding='utf8'):
```
narrieta, line 1046: while you are here, could mark this one as private?

narrieta (commented):
Review for event.py

azurelinuxagent/common/event.py
```diff
@@ -548,11 +544,26 @@ def add_metric(self, category, counter, instance, value, log_event=False):
...
+    def flush_or_save_event(self, event, message, immediate_flush):
+        """
+        Flush the event to wireserver if immediate_flush to set to true, else
+        save it disk if we fail to send or not required to flush immediately.
+        """
+        report_success = False
+        if immediate_flush and self.protocol is not None:
+            report_success = self.protocol.report_event([event])
+            if not report_success:
+                logger.error("Failed to send event: '{0}' directly to Wireserver. So, agent will save it to disk for periodic flush.", message)
```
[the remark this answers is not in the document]
narrieta: where do we delete the events when we fail to send them? important events should probably not be deleted, in case the send failure was transient and the next poll from the telemetry thread can send them

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1033,6 +1044,20 @@ def report_health(self, status, substatus, description):
...
+        """
+        Construct the encoded event and url for telemetry endpoint call
+        Before calling telemetry endpoint, ensure calls are under throttling limits and checks if the number of telemetry endpoint calls 12 in the last 15 seconds
+        (Considered 12 instead actual limit 15 as a conservative approach plus it helps for directly flushing events not hit the throttling issues)
+        (Trade off between delay vs successful event delivery for immediate_flush events)
+        Note: throttling limit is 15 calls in 15 seconds
+        """
+        def can_make_wireserver_call():
```
narrieta, line 1054: I am not sure we need this logic. I think we just need to keep track the timestamp for the most recent send, then check if WireServer fails with a throttling error and just sleep enough time to go over the 15 secs.

Throttling errors are reported with a 403 or 429 and this response body:

```
<Error xmlns:xsi=\http://www.w3.org/2001/XMLSchema-instance\ xmlns:xsd=\http://www.w3.org/2001/XMLSchema\>
    <Code> RequestRateLimitExceeded </Code>
    <Message> The request was denied because of too many requests by this client. Please retry the request after a few seconds. </Message>\
    <Details> </Details>
</Error>

```

azurelinuxagent/ga/collect_telemetry_events.py
```diff
@@ -248,17 +253,35 @@ def _ensure_all_events_directories_empty(extension_events_directories):
...
+        error_count = 0
+        while True:
+            try:
+                # Read event file and decode it properly
+                with open(event_file_path, "rb") as event_file_descriptor:
+                    event_data = event_file_descriptor.read().decode("utf-8")
+
+                # Parse the string and get the list of events
+                return json.loads(event_data)
+            except Exception:
```
narrieta, line 273: this catches all kinds of errors. what is the error we were trying to handle? empty file?

azurelinuxagent/common/event.py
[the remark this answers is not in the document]
nagworld9: yes, we save below #563 when we fail to send them

azurelinuxagent/common/protocol/wire.py
nagworld9: well, sleeping 15 secs always when throttling happens adds delay when it's not needed to wait that long. There is a chance that reset could happen before that. For example, we make 15 calls in 14 secs and each call at each min and hit the throttling. My next call would be reset after 1/2 mins instead 15 mins long. This helps for important events as well since we call wireserver directly and agent would wait on it to move forward with next steps. So, I want to deliver as soon as possible.

The new approach looks to me like we want to remediate the problem after we hit issue. but since we know the limits ahead, I went with that approach.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1054,13 +1097,27 @@ def send_encoded_event(self, provider_id, event_str, encoding='utf8'):
...
         buf = {}
         debug_info = CollectOrReportEventDebugInfo(operation=CollectOrReportEventDebugInfo.OP_REPORT)
         events_per_provider = defaultdict(int)
 
         def _send_event(provider_id, debug_info):
+            # we retry few times and eventually drop the events if we are not able to send them out
+            error_count = 0
             try:
-                self.send_encoded_event(provider_id, buf[provider_id])
+                while True:
```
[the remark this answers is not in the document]
nagworld9: Now, there are no retries in wireserver telemetry call. I explicitly placed here and it's needed for other events as well as some errors could be transient

azurelinuxagent/ga/collect_telemetry_events.py
nagworld9: I don't remember exception type. The scenario I know retry for reading the event file in case file is modified while reading. There could be josn deserialize error or file error.

azurelinuxagent/common/protocol/wire.py
narrieta: Yes, and that is the issue I see. This approach is tightly tied to the current implementation of WireServer and won't work is that changes.

maddieford (commented):
Thanks for addressing comments, changes lgtm! Will give another review after Norberto's comments are addressed

azurelinuxagent/ga/collect_telemetry_events.py
nagworld9: Found FileNotFoundError and jsondecodeerror from telemetry and added those for retry

```
DroppedEventsCount: 1
Reasons (first 5 errors): [EventError] Error parsing event: not well-formed (invalid token): line 1, column 1: Traceback (most recent call last):
  File "bin/WALinuxAgent-2.11.1.12-py3.9.egg/azurelinuxagent/common/event.py", line 214, in parse_event
    return parse_json_event(data_str)
  File "bin/WALinuxAgent-2.11.1.12-py3.9.egg/azurelinuxagent/common/event.py", line 204, in parse_json_event
    data = json.loads(data_str)
  File "/usr/lib64/python3.6/json/__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python3.6/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib64/python3.6/json/decoder.py", line 355, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 25 column 124 (char 874)

During handling of the above exception, another exception occurred:

  File "bin/WALinuxAgent-2.11.1.12-py3.9.egg/azurelinuxagent/common/event.py", line 237, in parse_xml_event
    xml_doc = parse_doc(data_str)
  File "bin/WALinuxAgent-2.11.1.12-py3.9.egg/azurelinuxagent/common/utils/textutil.py", line 41, in parse_doc
    return minidom.parseString(xml_text)
  File "/usr/lib64/python3.6/xml/dom/minidom.py", line 1968, in parseString
    return expatbuilder.parseString(string)
  File "/usr/lib64/python3.6/xml/dom/expatbuilder.py", line 925, in parseString
    return builder.parseString(string)
  File "/usr/lib64/python3.6/xml/dom/expatbuilder.py", line 223, in parseString
    parser.Parse(string, True)
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 1, column 1

During handling of the above exception, another exception occurred:

  File "bin/WALinuxAgent-2.11.1.12-py3.9.egg/azurelinuxagent/common/event.py", line 216, in parse_event
    return parse_xml_event(data_str)
  File "bin/WALinuxAgent-2.11.1.12-py3.9.egg/azurelinuxagent/common/event.py", line 247, in parse_xml_event
    raise ValueError(ustr(e))
ValueError: not well-formed (invalid token): line 1, column 1

During handling of the above exception, another exception occurred:

  File "bin/WALinuxAgent-2.11.1.12-py3.9.egg/azurelinuxagent/ga/collect_telemetry_events.py", line 437, in process_events
    event = parse_event(event_data)
  File "bin/WALinuxAgent-2.11.1.12-py3.9.egg/azurelinuxagent/common/event.py", line 218, in parse_event
    raise EventError("Error parsing event: {0}".format(ustr(e)))
azurelinuxagent.common.exception.EventError: [EventError] Error parsing event: not well-formed (invalid token): line 1, column 1
```
```
DroppedEventsCount: 583
Reasons (first 5 errors): [Errno 2] No such file or directory: '/var/lib/waagent/events/1726660508442709.tld': Traceback (most recent call last):
  File "bin/WALinuxAgent-2.11.1.12-py3.9.egg/azurelinuxagent/ga/collect_telemetry_events.py", line 434, in process_events
    with open(event_file_path, "rb") as event_fd:
FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/waagent/events/1726660508442709.tld'

During handling of the above exception, another exception occurred:

  File "bin/WALinuxAgent-2.11.1.12-py3.9.egg/azurelinuxagent/ga/collect_telemetry_events.py", line 461, in process_events
    os.remove(event_file_path)
FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/waagent/events/1726660508442709.tld'
, [Errno 2] No such file or directory: '/var/lib/waagent/events/1726660507946074.tld': Traceback (most recent call last):
  File "bin/WALinuxAgent-2.11.1.12-py3.9.egg/azurelinuxagent/ga/collect_telemetry_events.py", line 434, in process_events
    with open(event_file_path, "rb") as event_fd:
FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/waagent/events/1726660507946074.tld'

During handling of the above exception, another exception occurred:

  File "bin/WALinuxAgent-2.11.1.12-py3.9.egg/azurelinuxagent/ga/collect_telemetry_events.py", line 461, in process_events
    os.remove(event_file_path)
FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/waagent/events/1726660507946074.tld'
, [Errno 2] No such file or directory: '/var/lib/waagent/events/1726660507845022.tld': Traceback (most recent call last):
  File "bin/WALinuxAgent-2.11.1.12-py3.9.egg/azurelinuxagent/ga/collect_telemetry_events.py", line 434, in process_events
    with open(event_file_path, "rb") as event_fd:
FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/waagent/events/1726660507845022.tld'

During handling of the above exception, another exception occurred:

  File "bin/WALinuxAgent-2.11.1.12-py3.9.egg/azurelinuxagent/ga/collect_telemetry_events.py", line 461, in process_events
    os.remove(event_file_path)
FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/waagent/events/1726660507845022.tld'
, [Errno 2] No such file or directory: '/var/lib/waagent/events/1726660508044223.tld': Traceback (most recent call last):
  File "bin/WALinuxAgent-2.11.1.12-py3.9.egg/azurelinuxagent/ga/collect_telemetry_events.py", line 434, in process_events
    with open(event_file_path, "rb") as event_fd:
FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/waagent/events/1726660508044223.tld'

During handling of the above exception, another exception occurred:

  File "bin/WALinuxAgent-2.11.1.12-py3.9.egg/azurelinuxagent/ga/collect_telemetry_events.py", line 461, in process_events
    os.remove(event_file_path)
FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/waagent/events/1726660508044223.tld'
, [Errno 2] No such file or directory: '/var/lib/waagent/events/1726660502444669.tld': Traceback (most recent call last):
  File "bin/WALinuxAgent-2.11.1.12-py3.9.egg/azurelinuxagent/ga/collect_telemetry_events.py", line 434, in process_events
    with open(event_file_path, "rb") as event_fd:
FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/waagent/events/1726660502444669.tld'

During handling of the above exception, another exception occurred:

  File "bin/WALinuxAgent-2.11.1.12-py3.9.egg/azurelinuxagent/ga/collect_telemetry_events.py", line 461, in process_events
    os.remove(event_file_path)
FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/waagent/events/1726660502444669.tld'
```

azurelinuxagent/common/protocol/wire.py
narrieta: The body size is different. It is there to account for the limits in the telemetry infrastructure.

The problem we were trying to solve for throttling is not that we hit the throttling limit, it is that when we hit it, we keep retrying over and over (I think we also increase the number and possibility frequency of the retries) instead of pausing for a little bit and then trying again.

narrieta: You may be able to fix the throttling issue by fixing the logic around

```
DEFAULT_RETRIES = 6
DELAY_IN_SECONDS = 1

THROTTLE_RETRIES = 25
THROTTLE_DELAY_IN_SECONDS = 1

```

azurelinuxagent/common/protocol/wire.py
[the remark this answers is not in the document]
narrieta: There are retries at the resutil.http_request level

azurelinuxagent/common/protocol/wire.py
narrieta: The other reason we tend to hit the throttling limit is because we accumulate events for 5 minutes and then try to send them all at once, instead of streaming events as they are produced

nagworld9: > You may be able to fix the throttling issue by fixing the logic around
> 
> ```
> DEFAULT_RETRIES = 6
> DELAY_IN_SECONDS = 1
> 
> THROTTLE_RETRIES = 25
> THROTTLE_DELAY_IN_SECONDS = 1
> ```

since these limits apply for other endpoints, changing here affect other endpoints as well. I was not so sure why it was mentioned 25 at first place.

narrieta: yes, you need to account for the fact that this code is shared with other requests, but that should not be hard to do

nagworld9: > The other reason we tend to hit the throttling limit is because we accumulate events for 5 minutes and then try to send them all at once, instead of streaming events as they are produced

I checked wiresever doc, the other endpoints don't have throttle limit like telemeteyapi. It makes sense why we don't have such logic since we don't know when we hit the throttling but here we know beforehand.  Even we stream I feel, we still need to go within limits because api has ratelimit. 

The other approach looks to me we are not solving the problem, we are just reducing problem

azurelinuxagent/common/event.py
```diff
@@ -213,13 +213,9 @@ def parse_json_event(data_str):
...
-        try:
-            return parse_json_event(data_str)
-        except ValueError:
-            return parse_xml_event(data_str)
-    except Exception as e:
-        raise EventError("Error parsing event: {0}".format(ustr(e)))
-
+        return parse_json_event(data_str)
+    except ValueError:
+        return parse_xml_event(data_str)
```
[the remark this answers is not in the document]
nagworld9: yes, I want to add retry but adding here will get behavior for all the callers who don't need it.

maddieford (dismissed):

azurelinuxagent/common/event.py
```diff
@@ -549,11 +543,25 @@ def add_metric(self, category, counter, instance, value, log_event=False):
...
+
+    def report_or_save_event(self, event, flush=False):
+        """
+        Flush the event to wireserver if flush to set to true, else
+        save it disk if we fail to send or not required to flush immediately.
+        TODO: pickup as many events as possible and send them in one go.
+        """
+        report_success = False
+        if flush and self.protocol is not None:
+            report_success = self.protocol.report_event([event], flush)
```
maddieford, line 556: Does report_event handle all exceptions, or any that need to be handled here?

azurelinuxagent/common/protocol/wire.py
```diff
@@ -1059,7 +1061,12 @@ def send_encoded_event(self, provider_id, event_str, encoding='utf8'):
             header = self.get_header_for_xml_content()
             # NOTE: The call to wireserver requests utf-8 encoding in the headers, but the body should not
             #       be encoded: some nodes in the telemetry pipeline do not support utf-8 encoding.
-            resp = self.call_wireserver(restutil.http_post, uri, data, header)
+
+            # if it's important event flush, we use less throttle delay(to avoid long delay to complete this operation)) on throttling errors
+            if flush:
+                resp = self.call_wireserver(restutil.http_post, uri, data, header, max_retry=3, telemetry_throttle_delay=TELEMETRY_FLUSH_THROTTLE_DELAY_IN_SECONDS)
+            else:
+                resp = self.call_wireserver(restutil.http_post, uri, data, header, max_retry=3, telemetry_throttle_delay=TELEMETRY_THROTTLE_DELAY_IN_SECONDS)
```
maddieford, line 1069: Default retry was previously 6, we're allowing less retries now?

azurelinuxagent/common/utils/restutil.py
```diff
@@ -364,6 +373,7 @@ def http_request(method,
                  max_retry=None,
                  retry_codes=None,
                  retry_delay=DELAY_IN_SECONDS,
```
maddieford, line 375: What's the difference between retry_delay and telemetry_throttle_delay?

azurelinuxagent/common/utils/restutil.py
```diff
@@ -468,7 +481,10 @@ def http_request(method,
                     # retry attempts
                     if _is_throttle_status(resp.status):
                         was_throttled = True
-                        max_retry = max(max_retry, THROTTLE_RETRIES)
+                        # Today, THROTTLE_RETRIES set to big number as opposite to slow down the retry attempts
```
maddieford, line 484: I don't follow this comment, could you please explain:
> Today, THROTTLE_RETRIES set to big number as opposite to slow down the retry attempts

azurelinuxagent/common/event.py
nagworld9: it handled all types of exception. We have the same call in send_telemtry_handler for reference

azurelinuxagent/common/protocol/wire.py
nagworld9: My intention for lower number is higher retires contribute to reach throttle limit very frequently. Do you think default 6 would help on any general errors?

azurelinuxagent/common/utils/restutil.py
nagworld9: retly_delay used for non-telemetry calls for both retry errors and throttle errors

but telemetry_throttle_delay used telemetry calls throttle errors

RETRY_CODES = [
    httpclient.RESET_CONTENT,
    httpclient.PARTIAL_CONTENT,
    httpclient.FORBIDDEN,
    httpclient.INTERNAL_SERVER_ERROR,
    httpclient.NOT_IMPLEMENTED,
    httpclient.BAD_GATEWAY,
    httpclient.SERVICE_UNAVAILABLE,
    httpclient.GATEWAY_TIMEOUT,
    httpclient.INSUFFICIENT_STORAGE,
    429,  # Request Rate Limit Exceeded
]

THROTTLE_CODES = [
    httpclient.FORBIDDEN,
    httpclient.SERVICE_UNAVAILABLE,
    429, # Request Rate Limit Exceeded
]

azurelinuxagent/common/utils/restutil.py
nagworld9: Today when any api return throttle error, we make 26 attempts for retying as opposed to back off and make fewer attempts with good delay. 

My fix will change this behavior for only telemetry calls and keeping same for non-telemetry calls as I wasn't sure why it was implemented that way. 

I'll improve the comment

azurelinuxagent/common/utils/restutil.py
maddieford: I see, so for errors in retry_codes, we will use the retry_delay (for non-telemetry and telemetry calls)

Instead of adding telemetry_throttled_delay parameter, would it make sense to add throttle_delay parameter?
default value for throttle_delay parameter can be same as retry_delay. We can override throttle_delay default only for telemetry calls

azurelinuxagent/common/utils/restutil.py
maddieford: Thanks, makes sense

azurelinuxagent/common/utils/restutil.py
narrieta: +1

narrieta (approved):
LGTM; one minor comment

narrieta (dismissed):

maddieford (dismissed):

maddieford (approved):

narrieta (approved):

PR #3135, opened 2024-05-31
refactor cgroup controllers

maddieford:
## Description
The current code for cgroups is overcomplicated because it keeps track of each controller path separately. If we want to track additional controllers in the future with the existing code, we would need to update each block of code that references controller paths. In cgroup v2, there is a unified cgroup path, which makes it unnecessary to track different paths for each controller. 

The changes in this PR simplify the code by creating an abstraction for a cgroup which handles any operations that require cgroup paths. The changes include renaming the CGroup/CpuCGroup/MemoryCGroup classes to ControllerMetrics/CpuMetrics/MemoryMetrics and introduces the Cgroup/CgroupV1/CgroupV2 classes in the cgroupapi to represent a cgroup. The cgroupapi was updated to return instances of a Cgroup instead of tuples of controller paths. 

Note this PR does not yet add logic to track cpu/memory metrics in v2. Those changes will be included in a future PR.

Test run of all cgroup scenarios: https://dev.azure.com/cplatruntime/WALinuxAgent/_build/results?buildId=9290&view=results

Issue # 

---

### PR information


azurelinuxagent/ga/cgroupapi.py
```diff
@@ -470,55 +476,227 @@ def _get_root_cgroup_path():
...
+        return [CgroupV2.CPU_CONTROLLER, CgroupV2.MEMORY_CONTROLLER]
+
+    def check_in_expected_slice(self, expected_slice):
+        if expected_slice not in self._cgroup_path:
+            log_cgroup_warning("The {0} cgroup is not in the expected slice. Expected slice: {1}. Actual cgroup path: {2}".format(self._cgroup_name, expected_slice, self._cgroup_path), send_event=False)
+            return False
+
+        return True
+
+    def get_controller_metrics(self, expected_relative_path=None):
```
maddieford, line 690: Next step for cgroup v2 support will be adding logic to track metrics for cpu and memory in v2

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -612,24 +561,20 @@ def _check_processes_in_agent_cgroup(self):
...
+
             try:
                 daemon = os.getppid()
                 extension_handler = os.getpid()
                 agent_commands = set()
                 agent_commands.update(shellutil.get_running_commands())
                 systemd_run_commands = set()
                 systemd_run_commands.update(self._cgroups_api.get_systemd_run_commands())
-                agent_cgroup = self._cgroups_api.get_processes_in_cgroup(cgroup_path)
+                agent_cgroup_proccesses = self._agent_cgroup.get_processes()
```
maddieford, line 572: Note, before we were getting processes in only CPU cgroup.procs (or only Memory cgroups.procs if cpu wasn't mounted). Now this gets processes under cpu and memory if both mounted.

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -780,59 +725,37 @@ def _get_parent(pid):
             return 0
 
         def start_tracking_unit_cgroups(self, unit_name):
-            """
-            TODO: Start tracking Memory Cgroups
```
maddieford, line 784: Removed this comment because memory was already being tracked.

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -780,59 +725,37 @@ def _get_parent(pid):
...
-                    CGroupsTelemetry.track_cgroup(MemoryCgroup(unit_name, memory_cgroup_path))
+                for metric in metrics:
+                    CGroupsTelemetry.track_cgroup(metric)
 
             except Exception as exception:
                 log_cgroup_info("Failed to start tracking resource usage for the extension: {0}".format(ustr(exception)), send_event=False)
 
         def stop_tracking_unit_cgroups(self, unit_name):
-            """
-            TODO: remove Memory cgroups from tracked list.
```
maddieford, line 804: Removed this comment because memory was already being tracked.

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -780,59 +725,37 @@ def _get_parent(pid):
...
-                    CGroupsTelemetry.stop_tracking(MemoryCgroup(unit_name, memory_cgroup_path))
+                for metric in metrics:
+                    CGroupsTelemetry.stop_tracking(metric)
 
             except Exception as exception:
                 log_cgroup_info("Failed to stop tracking resource usage for the extension service: {0}".format(ustr(exception)), send_event=False)
 
         def stop_tracking_extension_cgroups(self, extension_name):
-            """
-            TODO: remove extension Memory cgroups from tracked list
```
maddieford, line 820: Removed this comment because memory was already being tracked.

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -470,55 +476,227 @@ def _get_root_cgroup_path():
...
+    def check_in_expected_slice(self, expected_slice):
+        in_expected_slice = True
+        for controller, path in self._controller_paths.items():
+            if expected_slice not in path:
+                log_cgroup_warning("The {0} controller for the {1} cgroup is not mounted in the expected slice. Expected slice: {2}. Actual controller path: {3}".format(controller, self._cgroup_name, expected_slice, path), send_event=False)
+                in_expected_slice = False
+
+        return in_expected_slice
+
+    def get_controller_metrics(self, expected_relative_path=None):
```
nagworld9, line 623: when do we send expected_path? and also not case. Cgroup instance created from relative paths, so controller_paths should have those paths, so why we need expected_path?

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -470,55 +476,227 @@ def _get_root_cgroup_path():
...
+            controller_metrics = None
+            controller_path = self._controller_paths.get(controller)
+
+            if controller_path is None:
+                log_cgroup_warning("{0} is not mounted for the {1} cgroup; will not track metrics".format(controller, self._cgroup_name), send_event=False)
+                continue
+
+            if expected_relative_path is not None:
+                expected_path = ""
+                controller_mountpoint = self._controller_mountpoints[controller]
```
nagworld9, line 636: this may fail if controller not found in mountpoint. Consider switch to get and handle None case and log the warning if it;s none

maddieford: My reasoning here was that if the controller exists in self._controller_paths, then it must be mounted. But I will add none check here to be cautious

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -189,28 +188,30 @@ def initialize(self):
...
-                    CGroupsTelemetry.track_cgroup(CpuCgroup(AGENT_NAME_TELEMETRY, self._agent_cpu_cgroup_path))
-
-                if self._agent_memory_cgroup_path is not None:
-                    log_cgroup_info("Agent Memory cgroup: {0}".format(self._agent_memory_cgroup_path))
-                    self._agent_memory_cgroup = MemoryCgroup(AGENT_NAME_TELEMETRY, self._agent_memory_cgroup_path)
-                    CGroupsTelemetry.track_cgroup(self._agent_memory_cgroup)
+                for metric in metrics:
+                    for prop in metric.get_unit_properties():
+                        log_cgroup_info('{0}: {1}'.format(prop, systemd.get_unit_property(systemd.get_agent_unit_name(), prop)))
+                    CGroupsTelemetry.track_cgroup(metric)
```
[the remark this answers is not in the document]
maddieford: gotcha, I'll move this below set_cpu_quota

azurelinuxagent/ga/cgroupapi.py
maddieford: We use expected_path when we _only_ want to track the metrics if the process cgroup is at the expected path. 

For example, in cgroupconfigurator.initialize() we only want to track the agent metrics if the cgroup is at this path: 'azure.slice/waagent.service'. If it's at any other path, we don't want to track metrics and we disable cgroups. 
See lines 202-205 in cgroupconfigurator for usage

tests_e2e/tests/scripts/agent_cgroups_process_check-unknown_process_check.py
```diff
@@ -62,8 +62,8 @@ def disable_agent_cgroups_with_unknown_process(pid):
     Note: System may kick the added process out of the cgroups, keeps adding until agent detect that process
     """
 
-    def unknown_process_found(cpu_cgroup):
-        cgroup_procs_path = os.path.join(cpu_cgroup, "cgroup.procs")
+    def unknown_process_found(cgroup_path):
+        cgroup_procs_path = os.path.join(cgroup_path, "cgroup.procs")
```
nagworld9, line 66: can we get cgroup.procs directly from systemd api in agent, instead of building path here?

tests_e2e/tests/lib/cgroup_helpers.py
```diff
@@ -164,9 +164,14 @@ def check_log_message(message, after_timestamp=datetime.datetime.min):
...
     """
-    Returns the cgroup paths for the given unit
+    Returns the cgroup path for the given unit. For v1, gets the path of the cpu controller.
     """
     cgroups_api = get_cgroup_api()
-    return cgroups_api.get_unit_cgroup_paths(unit_name)
+    unit_controlgroup_path = systemd.get_unit_property(unit_name, "ControlGroup")
+
+    if isinstance(cgroups_api, SystemdCgroupApiv1):
+        return os.path.join(cgroups_api.get_controller_mountpoints().get('cpu,cpuacct'), unit_controlgroup_path[1:])
```
nagworld9, line 175: I think we can abstract this cgroup class?
why can't use cgroups_api. get_unit_cgroup() and get cgroup object and add method to get cgroup cpu or memory path based on function parameters or something like that?

maddieford: We could but it doesn't really make sense for v2 since all controllers are at the same path

nagworld9: Ok, I think method we have here for constructing cgroup.procs path, at least we should change that

tests_e2e/tests/scripts/agent_cgroups_process_check-unknown_process_check.py
```diff
@@ -81,9 +80,7 @@ def unknown_process_found(cpu_cgroup):
                 pid)), attempts=3)
         return found and retry_if_false(check_agent_quota_disabled, attempts=3)
 
-    cpu_cgroup, _ = get_unit_cgroup_paths(AGENT_SERVICE_NAME)
-
-    found: bool = retry_if_false(lambda: unknown_process_found(cpu_cgroup), attempts=3)
+    found: bool = retry_if_false(lambda: unknown_process_found(), attempts=3)       # pylint: disable=W0108
```
nagworld9, line 83: can you check if lambda needed if function don't have args?

maddieford: the first arg to retry_if_false needs to be Callable. it won't accept bool as first arg

maddieford: i'll check if i can remove () and lambda

maddieford: removed

nagworld9 (approved):

azurelinuxagent/ga/controllermetrics.py
```diff
@@ -169,10 +169,16 @@ def get_tracked_metrics(self, **_):
         """
         raise NotImplementedError()
 
+    def get_unit_properties(self):
```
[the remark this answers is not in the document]
maddieford: Yeah they are only being used to log the properties when we initialize cgroups. We log the properties before setting any quotas. Then after setting cpu quota, we log only the cpuquota property value. 

It was logging we were doing before so I kept it, but since we log the properties after setting quotas it might not be necessary.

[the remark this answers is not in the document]
narrieta: maybe remove them altogether? the log message doesn't have any context about what those are

or add more context to the message, if you think this could be useful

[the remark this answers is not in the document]
maddieford: I'll open a PR to improve the message. I think it may be useful to include, so that we can see from the logs if there were any quotas previously set before the current agent attempts to set quotas

PR #3137, opened 2024-06-06
Switching to SNI based authentication for aad app

nagworld9:
## Description

Fixing security alert - removing pinned certs and switching to subject name issuer-based authentication

Issue # 

---

### PR information

### DIFF
--- a/tests_e2e/orchestrator/docker/Dockerfile
+++ b/tests_e2e/orchestrator/docker/Dockerfile
@@ -67,7 +67,7 @@ RUN \
     cd $HOME                                                                                                  && \
     git clone https://github.com/microsoft/lisa.git                                                           && \
     cd lisa                                                                                                   && \
-    git checkout 2c16e32001fdefb9572dff61241451b648259dbf                                                     && \
+    git checkout 95c09ff7d5b6e71d1642a628607ac9bb441c69f5                                                     && \
                                                                                                                  \
     python3 -m pip install --upgrade pip                                                                      && \
     python3 -m pip install --editable .[azure,libvirt] --config-settings editable_mode=compat                 && \
--- a/tests_e2e/tests/lib/network_security_rule.py
+++ b/tests_e2e/tests/lib/network_security_rule.py
@@ -17,7 +17,7 @@
 
 import json
 
-from typing import Any, Dict, List
+from typing import Any, Dict
 
 from tests_e2e.tests.lib.update_arm_template import UpdateArmTemplate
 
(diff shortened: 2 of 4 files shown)


tests_e2e/orchestrator/docker/Dockerfile
```diff
@@ -67,7 +67,7 @@ RUN \
     cd $HOME                                                                                                  && \
     git clone https://github.com/microsoft/lisa.git                                                           && \
     cd lisa                                                                                                   && \
-    git checkout 2c16e32001fdefb9572dff61241451b648259dbf                                                     && \
+    git checkout 95c09ff7d5b6e71d1642a628607ac9bb441c69f5                                                     && \
```
nagworld9, line 70: Pulling lisa latest changes since they updated azure-identity version which supports SNI based authentication

tests_e2e/tests/lib/network_security_rule.py
```diff
@@ -55,7 +55,7 @@ def add_security_rule(self, security_rule: Dict[str, Any]) -> None:
         self._get_network_security_group()["properties"]["securityRules"].append(security_rule)
 
     def _get_network_security_group(self) -> Dict[str, Any]:
-        resources: List[Dict[str, Any]] = self._template["resources"]
+        resources: Dict[str, Dict[str, Any]] = self._template["resources"]
```
nagworld9, line 58: Schema updated in new lisa template; resources type changed to dict(dict) from list(dict)

tests_e2e/tests/lib/network_security_rule.py
```diff
@@ -76,14 +76,14 @@ def _get_network_security_group(self) -> Dict[str, Any]:
...
+        resources[nsg_reference] = network_security_group
 
         #
         # Add a dependency on the NSG to the virtual network
         #
         network_resource = UpdateArmTemplate.get_resource(resources, "Microsoft.Network/virtualNetworks")
         network_resource_dependencies = network_resource.get("dependsOn")
-        nsg_reference = f"[resourceId('Microsoft.Network/networkSecurityGroups', '{self._NETWORK_SECURITY_GROUP}')]"
         if network_resource_dependencies is None:
             network_resource["dependsOn"] = [nsg_reference]
```
maddieford, line 88: now the virtual network has a dependency on "network_security_groups", is that correct? I think the dependency should still be nsg_reference = f"[resourceId('Microsoft.Network/networkSecurityGroups', '{self._NETWORK_SECURITY_GROUP}')]"

nagworld9: no change in dependency, the reference we need to use will change instead resource id, it will be dict key where this resource id kept.

nagworld9: for example, https://github.com/microsoft/lisa/pull/3163/files#diff-6ade80587709e19354bf9ad7752936686e13956b007be37baadc9baa14b2d733R699

maddieford (approved):

PR #3138, opened 2024-06-06
suppress pylint warn contextmanager-generator-missing-cleanup

nagworld9:
## Description
Fix for new warning from latest pylint.
_`_tests/ga/test_update.py:1203:12: W0135: The context used in function '_setup_test_for_ext_event_dirs_retention' will not be exited. (contextmanager-generator-missing-cleanup)_`_

It looks like false positive according to documentation https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/contextmanager-generator-missing-cleanup.html

_The check can create false positives if yield is used inside an if-else block without custom cleanup. Use pylint: disable for these._

This warning not implemented in earlier versions of python (3.0 -3.7) So I disabled the warning with (bad-option-value) in GitHub workflow

Issue # 

---

### PR information

### DIFF
--- a/tests/ga/test_multi_config_extension.py
+++ b/tests/ga/test_multi_config_extension.py
@@ -284,6 +284,7 @@ class TestMultiConfigExtensions(_MultiConfigBaseTestClass):
         third_ext = extension_emulator(name="OSTCExtensions.ExampleHandlerLinux.thirdExtension")
         fourth_ext = extension_emulator(name="Microsoft.Powershell.ExampleExtension")
 
+        # pylint: disable=contextmanager-generator-missing-cleanup
         with self._setup_test_env(mock_manifest=True) as (exthandlers_handler, protocol, no_of_extensions):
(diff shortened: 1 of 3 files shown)


tests/ga/test_multi_config_extension.py
```diff
@@ -284,6 +284,7 @@ def __setup_generic_test_env(self):
         third_ext = extension_emulator(name="OSTCExtensions.ExampleHandlerLinux.thirdExtension")
         fourth_ext = extension_emulator(name="Microsoft.Powershell.ExampleExtension")
 
+        # pylint: disable=contextmanager-generator-missing-cleanup
```
narrieta, line 287: this needs to be inline, on the line producing the warning

narrieta: also, if you can add a comment to each suppression pointing out what generator is triggering the warning, it'd be great

narrieta (approved):

maddieford (approved):

PR #3151, opened 2024-06-20
redhat: Use NetworkManager to set DHCP hostnames on recent RHEL distros

ani-sinha:
Recent versions of RHEL 8 and RHEL 9 started using NetworkManager to configure network intefaces instead of sysconfig files. Configurations are no longer stored in /etc/sysconfig/ifcfg-{interface}.

Fix this for setting DHCP hostnames in those RHEL versions.

### DIFF
--- a/azurelinuxagent/common/osutil/redhat.py
+++ b/azurelinuxagent/common/osutil/redhat.py
@@ -266,6 +266,19 @@ class RedhatOSModernUtil(RedhatOSUtil):
         # TODO: Implement and test a way to recover the network interface for RedhatOSModernUtil
         pass
 
+
+    def set_dhcp_hostname(self, hostname):
+        """
+        Recent RHEL distributions use network manager instead of sysconfig files
+        to configure network interfaces
+        """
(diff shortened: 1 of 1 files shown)


maddieford:
@ani-sinha Thank you for the PR. Hostname publishing on Rhel 8/9 is a high priority item for us. I'll work on stress testing this solution over the next few weeks.

ani-sinha:
@maddieford aftyallt we have been using this patch downstream for a long time on RHEL 9. I just wanted you to know push this upstream so that we can drop this downstream only patch.

vittyvk:
I somehow missed this PR and sent my own version: https://github.com/Azure/WALinuxAgent/pull/3243
the difference between the two is that I do not believe "nmcli device modify" is sufficient and thus I do "nmcli connection modify" instead. This way, the change is supposed to be persistent.

ani-sinha:
I am closing this in favor of https://github.com/Azure/WALinuxAgent/pull/3243

PR #3162, opened 2024-07-16
Update ext_seq message pattern for new VMExtensionProvisioningError pattern

maddieford:
## Description
The extension sequencing scenario has a test case to add two extensions to the VM where one extension's enable operation fails, and the other is dependent on the failing extension. The scenario adds these extensions to the VM using an ARM template, and expects the template deployment to fail with code 'VMExtensionProvisioningError' and this message:
`Multiple VM extensions failed to be provisioned on the VM. Please see the VM extension instance view for other failures. The first extension failed due to the error: VM Extension 'AzureMonitorLinuxAgent' is marked as failed since it depends upon the VM Extension 'CustomScript' which has failed.`
 
Previously CRP was only reporting the failure for the first extension. There was an update to report the failures for all extensions. So now CRP reports a 'VMExtensionProvisioningError' with this message:
`Multiple VM extensions failed to be provisioned on the VM. The Extensions failed due to the errors: \n[Extension Name: 'AzureMonitorLinuxAgent'\nError Message: VM Extension 'AzureMonitorLinuxAgent' is marked as failed since it depends upon the VM Extension 'CustomScript' which has failed.]\n\n[Extension Name: 'CustomScript'\nError Message: VM has reported a failure when processing extension 'CustomScript' (publisher 'Microsoft.Azure.Extensions' and type 'CustomScript').`

This PR updates the regex the test uses to check that the extension provisioning failed as expected.

Issue # 

---

### PR information

### DIFF
--- a/tests_e2e/tests/ext_sequencing/ext_sequencing.py
+++ b/tests_e2e/tests/ext_sequencing/ext_sequencing.py
@@ -226,7 +226,7 @@ class ExtSequencing(AgentVmssTest):
                 # fail. We know an extension should fail if "failing" is in the case name. Otherwise, report the
                 # failure.
                 deployment_failure_pattern = r"[\s\S]*\"details\": [\s\S]* \"code\": \"(?P<code>.*)\"[\s\S]* \"message\": \"(?P<msg>.*)\"[\s\S]*"
-                msg_pattern = r"Multiple VM extensions failed to be provisioned on the VM. Please see the VM extension instance view for other failures. The first extension failed due to the error: VM Extension '.*' is marked as failed since it depends upon the VM Extension 'CustomScript' which has failed."
+                msg_pattern = r"Multiple VM extensions failed to be provisioned on the VM.*VM Extension '.*' is marked as failed since it depends upon the VM Extension 'CustomScript' which has failed."
                 deployment_failure_match = re.match(deployment_failure_pattern, str(e))
                 if "failing" not in case.__name__:
                     fail("Extension template deployment unexpectedly failed: {0}".format(e))


tests_e2e/tests/ext_sequencing/ext_sequencing.py
```diff
@@ -226,7 +226,7 @@ def run(self):
                 # fail. We know an extension should fail if "failing" is in the case name. Otherwise, report the
                 # failure.
                 deployment_failure_pattern = r"[\s\S]*\"details\": [\s\S]* \"code\": \"(?P<code>.*)\"[\s\S]* \"message\": \"(?P<msg>.*)\"[\s\S]*"
-                msg_pattern = r"Multiple VM extensions failed to be provisioned on the VM. Please see the VM extension instance view for other failures. The first extension failed due to the error: VM Extension '.*' is marked as failed since it depends upon the VM Extension 'CustomScript' which has failed."
+                msg_pattern = r"Multiple VM extensions failed to be provisioned on the VM.*VM Extension '.*' is marked as failed since it depends upon the VM Extension 'CustomScript' which has failed."
```
narrieta, line 229: should we also match "VM Extension 'AzureMonitorLinuxAgent' is marked as failed "?

maddieford: The next test case expects RunCommandLinux to fail because it's dependent on Custom script. The pattern matches for both test cases

narrieta: maybe we should parameterize the extension name?

narrieta: the test would be more robust if we check the extension that failed and the extension we did not execute

maddieford: We could, but I'm not sure it matters. This check mainly exists to confirm that the deployment failure is due to extension dependency failures, and not some other failure. 
There's additional checks below this one that checks the agent log to confirm the correct extensions failed due to their dependencies.

narrieta (approved):

PR #3187, opened 2024-08-14
Add Regorus policy engine framework

mgunnala:
## Description

This is PR 1 for policy enforcement within the guest agent. This PR adds the test Regorus executable and initial policy engine framework to call the executable. Guest agent functionality is not yet affected. 
- Regorus is utilized as the policy engine. [Regorus](https://github.com/microsoft/regorus) is maintained as a separate open-source repository. Currently, static Rust MUSL-based executable is built separately and copied into the tests/data folder. The binary will eventually be added to the package via official build pipeline but for now, is kept in the test directory.
- regorus.py implements add_policy, add_data, set_input, and eval_query functions by calling the Regorus executable via subprocess. 
- policy_engine.py implements a PolicyEngineConfigurator (responsible for conf checks and import) and PolicyEngine (responsible for higher level policy functions). 
- new "Debug.EnableExtensionPolicy" conf flag added to enable policy enforcement. Regorus is only imported if policy is enabled via conf _and_ the platform is supported.
- In this PR, extension installation is _not_ affected (no policy enforcement code added to exthandlers.py). This functionality will be added in the next PR. Once functionality is complete, any errors thrown in policy enforcement code will block extension installation. 
- Unit tests were added for policy_engine.py and regorus.py. There is no change to agent functionality (PolicyEngine is not instantiated anywhere) so no end-to-end tests have been added yet. 

---

### PR information


azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,175 @@
...
+
+        except (ImportError, NameError) as ex:
+            log_policy("Error: Failed to import Regorus module and initialize policy engine. {0}".format(ex), is_success=False)
+        except Exception as ex:
+            log_policy("Error: Failed to enable policy enforcement. '{0}'".format(ex), is_success=False)
+        finally:
+            PolicyEngineConfigurator._initialized = True
+
+    @staticmethod
+    def _is_policy_supported():
```
mgunnala, line 85: Need to add a check for ARM64 here

azurelinuxagent/ga/policy/regorus.py
```diff
@@ -0,0 +1,88 @@
...
+import subprocess
+import os
+
+
+def get_regorus_path():
+    """
+    Returns path to Regorus executable. Currently, the executable is copied into ga/policy/regorus
+    during testing. It is not yet officially released as part of the agent package.
+    """
+    regorus_exe = os.path.join(os.getcwd(), "azurelinuxagent", "ga", "policy", "regorus")
```
mgunnala, line 28: need to define another path, because we don't unzip the egg

mgunnala: subdirectory within bin inside the package, update makepkg.py

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,175 @@
...
+        """Data_file should be a path to a valid JSON data file."""
+        if not self.policy_engine_enabled:
+            return
+
+        try:
+            self._engine.add_data(data_file)
+        except Exception as ex:
+            log_policy("Error: Failed to add data to Regorus policy engine. '{0}'".format(ex), is_success=False)
+
+    def set_input(self, input_file):
```
mgunnala, line 155: create a temp file for set_input and clean it up

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,175 @@
...
+        # we only attempt to import Regorus if it is both enabled via conf and supported by the platform
+        try:
+            if PolicyEngineConfigurator._initialized:
+                return
+
+            if not conf.get_extension_policy_enabled():
+                log_policy("Policy enforcement is disabled via configuration file.")
+                return
+
+            if not self._is_policy_supported():
```
mgunnala, line 67: In this case, we want to allow all extensions to be installed, but log that customer is trying to enable policy on unsupported distro.

tests/ga/test_policy_engine.py
```diff
@@ -0,0 +1,138 @@
...
+    @classmethod
+    def setUpClass(cls):
+        # Currently, ga/policy/regorus contains a dummy binary. The unit tests require a real binary,
+        # so we replace the dummy with a copy from the tests_e2e folder.
+        regorus_source_path = os.path.abspath(os.path.join(data_dir, "policy/regorus"))
+        cls.regorus_dest_path = os.path.abspath(os.path.join(test_dir, "..", "azurelinuxagent/ga/policy/regorus"))
+        if not os.path.exists(cls.regorus_dest_path):
+            shutil.copy(regorus_source_path, cls.regorus_dest_path)
+        # Patch the path to regorus for all unit tests.
+        cls.patcher = patch('azurelinuxagent.ga.policy.regorus.get_regorus_path', return_value=cls.regorus_dest_path)
```
mgunnala, line 42: Add a comment explaining how binary needs to be in the agent package for real VMs, as opposed to in the egg (as it is here for unit testing purposes)

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: Added

azurelinuxagent/ga/policy/regorus.py
mgunnala: Once Regorus is published, makepkg.py will be updated to include it as part of agent package. Updated path and comment to explain. Follow-up task is being tracked under ADO #29072449

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: Added this functionality in regorus.py, set_input in policy_engine.py accepts a dict or json string.

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,190 @@
...
+
+class PolicyEngineConfigurator:
+    """
+    This Singleton class is responsible for checking if policy is enabled for the agent.
+    Policy can only be enabled/disabled via configuration file, and can only be enabled on supported distros.
+    If an error occurs during Regorus import, policy will be globally disabled for the agent via the _policy_enabled flag.
+    Use: PolicyEngineConfigurator.get_instance()
+    """
+    _instance = None  # configurator is implemented as a singleton
+    _initialized = False  # set to true on first init, even if Regorus import fails, we don't want to retry.
```
narrieta, line 56: I don't think this state should be static, since we are trying to initialize only once per waagent service start.

Under some scenarios, for example the user tries to use the feature on an unsupported distro, the user should not need to restart the agent to fix the issue.

This initialization should probably be done on a per-goal state basis, unless it is very expensive.

Alternatively, some initialization could be done on a per-service-start basis, and some on a per-goal-state basis, but those should be clearly defined.

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,190 @@
...
+        return conf.get_extension_policy_enabled()
+
+    @staticmethod
+    def _is_policy_supported():
+        """Return True if the platform/distro supports policy enforcement."""
+        arch = platform.machine().lower()
+        if arch not in POLICY_SUPPORTED_ARCHITECTURE:
+            return False
+        distro_info = get_distro()
+        distro_name = distro_info[0]
```
narrieta, line 102: prefer DISTRO_NAME and DISTRO_VERSION over distro_info[0], [1]. Makes the code easier to read.

they are initialized like this

```
__distro__ = get_distro()
DISTRO_NAME = __distro__[0]
DISTRO_VERSION = __distro__[1]

```

narrieta (commented):
Some comments on the Agent code. I did not review the tests, since I think addressing some of my comment will require changes in the implementation/behavior

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,190 @@
...
+                log_policy("Policy enforcement is unsupported on this platform.")
+                return
+
+            # Regorus import should only be attempted after completing the above checks within
+            # the configurator, but the module itself needs to be accessible outside this class.
+            global regorus  # pylint: disable=global-statement
+            import azurelinuxagent.ga.policy.regorus as regorus
+            PolicyEngineConfigurator._policy_enabled = True
+
+        except (ImportError, NameError) as ex:
```
narrieta, line 82: How could ImportError and NameError happen?

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,190 @@
...
+
+            if not self._is_policy_supported():
+                # If distro is unsupported but feature is enabled, we never set _policy_enabled=True.
+                # Any methods will be no-ops, and query evaluation will return an empty allowlist
+                # so no extensions will be allowed.
+                log_policy("Policy enforcement is unsupported on this platform.")
+                return
+
+            # Regorus import should only be attempted after completing the above checks within
+            # the configurator, but the module itself needs to be accessible outside this class.
```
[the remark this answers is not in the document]
narrieta: I think importing here, only after checking that the feature is enabled, etc, used to make sense when the interface with regorus was going to be a set of Python bindings that may crash on some distros. Not sure what the motivation is with the current implementation.

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,190 @@
...
+    all methods will be no-ops.
+
+    If any errors are thrown in regorus.py, they will be caught and handled here. add_policy, add_data,
+    and set_input will be no-ops, eval_query will return an empty dict
+    """
+    def __init__(self):
+        self._policy_engine_enabled = False
+        self._engine = None
+        try:
+            if PolicyEngineConfigurator.get_instance().get_policy_enabled():
```
narrieta, line 138: This usage is weird, because get_policy_enabled() is an static method, so one does not need an instance to invoke it...

Why the call to get_instance()?

This is related to one of my comments above about all methods/properties of PolicyEngineConfigurator being static, so there is no need for a singleton instance

azurelinuxagent/ga/policy/regorus.py
```diff
@@ -0,0 +1,99 @@
...
+    def __init__(self):
+        self._engine = None
+        self._policy_file = None
+        self._data_file = None
+        self._input_json = None
+
+    def add_policy(self, policy_path):
+        """Policy_path is expected to point to a valid Regorus file."""
+        if not os.path.exists(policy_path) or not policy_path.endswith('.rego'):
+            raise Exception("Policy path {0} is not a valid .rego file.".format(policy_path))
```
narrieta, line 49: this check seems extremely weak... seems like we accept as valid any file, no matter its contents, as long as its name ends with ".rego"

narrieta, line 49: what's in a *.rego file? JSON content? if yes, let's use ".json" instead. The fact that regorus is being used under the covers is an implementation detail and ".json" will be more familiar to users than "*.rego"

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,190 @@
...
+        if not self.policy_engine_enabled:
+            return
+
+        try:
+            self._engine.add_data(data_file)
+        except Exception as ex:
+            log_policy("Error: Failed to add data to Regorus policy engine. '{0}'".format(ex), is_success=False)
+
+    def set_input(self, input_json):
+        """Input_json should be a JSON object."""
```
narrieta, line 171: can you describe what is in that JSON object?

azurelinuxagent/ga/policy/regorus.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def add_data(self, data):
+        """Data parameter is expected to point to a valid json file."""
+        if not os.path.exists(data) or not data.endswith(".json"):
+            raise Exception("Data path {0} is not a valid JSON file.".format(data))
+        self._data_file = data
+
+    def eval_query(self, query):
+        missing_files = []
+        if self._policy_file is None:
```
narrieta, line 69: this is weird. callers to eval_query() need to know that they must call add_policy(), set_input() and add_data() f first. How do they know that? why if the caller invokes those 3 methods to evaluate a query and then forgets to call one of them when trying to evaluate another query?

why not 

`def eval_query(self, policy_path, input_json, data, query):`

azurelinuxagent/ga/policy/regorus.py
```diff
@@ -0,0 +1,99 @@
...
+        self._input_json = None
+
+    def add_policy(self, policy_path):
+        """Policy_path is expected to point to a valid Regorus file."""
+        if not os.path.exists(policy_path) or not policy_path.endswith('.rego'):
+            raise Exception("Policy path {0} is not a valid .rego file.".format(policy_path))
+        self._policy_file = policy_path
+
+    def set_input(self, input_json):
+        """Input_path is expected to point to a valid JSON object."""
```
narrieta, line 53: can you describe what is in that JSON object?

azurelinuxagent/ga/policy/regorus.py
```diff
@@ -0,0 +1,99 @@
...
+        if isinstance(input_json, dict):
+            self._input_json = input_json
+        elif isinstance(input_json, str):
+            self._input_json = json.loads(input_json)
+        else:
+            raise Exception("Input to policy engine is not a valid JSON object.")
+
+    def add_data(self, data):
+        """Data parameter is expected to point to a valid json file."""
```
narrieta, line 62: can you describe what is in that JSON file?

azurelinuxagent/ga/policy/regorus.py
```diff
@@ -0,0 +1,99 @@
...
+        with tempfile.NamedTemporaryFile(mode='w', suffix='.json', delete=True) as input_file:
+            json.dump(self._input_json, input_file, indent=4)
+            input_file.flush()
+
+            regorus_exe = get_regorus_path()
+            command = [regorus_exe, "eval", "-d", self._policy_file, "-d", self._data_file,
+                       "-i", input_file.name, query]
+
+            # use subprocess.Popen instead of subprocess.run for Python 2 compatibility
+            process = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
narrieta, line 89: use shellutil.run_command() instead of Popen()

narrieta (commented):
I missed one file in my initial review. Additional comments.

azurelinuxagent/ga/policy/regorus.py
narrieta: i see, the actual validation would be done by regorus and then we would need to handle the error reporting here.

a TODO comment would clarify thin

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,190 @@
...
+from azurelinuxagent.common import conf
+import platform
+
+# Define support matrix for Regorus and policy engine feature.
+# Dict in the format: { distro:min_supported_version }
+POLICY_SUPPORTED_DISTROS = {
+    'ubuntu': FlexibleVersion('16.04'),
+    'mariner': FlexibleVersion('2')
+}
+POLICY_SUPPORTED_ARCHITECTURE = ['x86_64', 'amd64']
```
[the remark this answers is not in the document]
maddieford: We should have more telemetry on this now. I'll share with Manu

azurelinuxagent/ga/policy/policy_engine.py
maddieford: +1, the extension status message should be clear that extension processing is blocked because this is an unsupported distro (or whatever the reason is)

tests/data/policy/agent_extension_policy.rego
```diff
@@ -0,0 +1,125 @@
+# Copyright (c) Microsoft Corporation.
```
narrieta, line 1: should this be under the normal Agent license instead?

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+# Common logging function across PolicyEngineConfigurator and PolicyEngine classes,
+# so should be a module-level function.
+def log_policy(formatted_string, is_success=True, op=WALAEventOperation.Policy, send_event=True):
+    """
+    Log information to console and telemetry.
+    """
+    if is_success:
+        logger.info("[Policy] " + formatted_string)
```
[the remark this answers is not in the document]
mgunnala: I thought it made the log file a little easier to read/parse, can remove if you think it adds too much clutter.

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,190 @@
...
+from azurelinuxagent.common.event import WALAEventOperation, add_event
+from azurelinuxagent.common import conf
+import platform
+
+# Define support matrix for Regorus and policy engine feature.
+# Dict in the format: { distro:min_supported_version }
+POLICY_SUPPORTED_DISTROS = {
+    'ubuntu': FlexibleVersion('16.04'),
+    'mariner': FlexibleVersion('2')
+}
```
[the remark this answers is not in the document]
mgunnala: Changed to DistroVersion.

azurelinuxagent/ga/policy/policy_engine.py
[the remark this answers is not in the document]
mgunnala: We're not supporting arm64 until public preview. Added a TODO to update this list once we enable support.

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,190 @@
...
+from azurelinuxagent.common import logger
+from azurelinuxagent.common.version import get_distro
+from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
+from azurelinuxagent.common.event import WALAEventOperation, add_event
+from azurelinuxagent.common import conf
+import platform
+
+# Define support matrix for Regorus and policy engine feature.
+# Dict in the format: { distro:min_supported_version }
+POLICY_SUPPORTED_DISTROS = {
```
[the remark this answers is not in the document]
mgunnala: Fixed

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,190 @@
...
+    'ubuntu': FlexibleVersion('16.04'),
+    'mariner': FlexibleVersion('2')
```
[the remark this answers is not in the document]
mgunnala: Added

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,190 @@
...
+        logger.info("[Policy] " + formatted_string)
+    else:
+        logger.error("[Policy] " + formatted_string)
+    if send_event:
+        add_event(op=op, message=formatted_string, is_success=is_success)
+
+
+class PolicyEngineConfigurator:
+    """
+    This Singleton class is responsible for checking if policy is enabled for the agent.
```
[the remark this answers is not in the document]
mgunnala: removed the configurator

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,190 @@
...
+
+        except (ImportError, NameError) as ex:
+            log_policy("Error: Failed to import Regorus module and initialize policy engine. {0}".format(ex), is_success=False)
+        except Exception as ex:
+            log_policy("Error: Failed to enable policy enforcement. '{0}'".format(ex), is_success=False)
+        finally:
+            PolicyEngineConfigurator._initialized = True
+
+    @staticmethod
+    def is_policy_configured():
```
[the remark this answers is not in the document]
mgunnala: renamed to is_policy_enforcement_enabled( ), to avoid confusion with the policy_engine_enabled property.

azurelinuxagent/ga/policy/policy_engine.py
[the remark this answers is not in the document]
maddieford: I think we should remove 'amd64' from this list. We don't have telemetry for this arch type:

cluster("azcore.centralus.kusto.windows.net").database("Fa").GuestAgentExtensionEvents
| where PreciseTimeStamp >= ago(30d)
| where OSVersion startswith "linux"
| summarize dcount(ContainerId) by KeywordName

KeywordName | dcount_ContainerId
-- | --
{"CpuArchitecture": "x86_64"} | 739,107,991
  | 639,940,276
{"CpuArchitecture": "aarch64"} | 3,092,416
{"CpuArchitecture": "i686"} | 280

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,144 @@
...
+    @staticmethod
+    def is_policy_enforcement_enabled():
+        """Check whether user has opted into policy enforcement feature"""
+        # TODO: Remove conf flag post private preview and add other checks here.
+        return conf.get_extension_policy_enabled()
+
+    @staticmethod
+    def is_policy_enforcement_supported():
+        """Check that both platform architecture and distro/version are supported."""
+        arch = platform.machine().lower()
```
maddieford, line 91: We have an osutil function for this: osutil.get_vm_arch()

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: Updated error handling to raise exceptions

azurelinuxagent/ga/policy/policy_engine.py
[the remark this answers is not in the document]
mgunnala: removed amd64

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: Moved initialization to PolicyEngine. Exthandlers.py will initialize the PolicyEngine on a per-goal state basis.

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: Updated

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: Removed these errors, not relevant since we're no longer using a python wrapper

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: Removed PolicyEngineConfigurator

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: Added

azurelinuxagent/ga/policy/regorus.py
mgunnala: Discussed offline - this is how the regorus library implements it, we're keeping the functions the same in case we want to change the underlying implementation in the future (ex switch to a python library). PolicyEngine init function calls add_policy() and add_data(), PolicyEngine.evaluate_query() takes input as a parameter and calls set_input().

azurelinuxagent/ga/policy/regorus.py
mgunnala: Added

azurelinuxagent/ga/policy/regorus.py
mgunnala: Added, though I haven't included full details here, more explanation in the spec doc. Let me know if more detail is needed here specifically.

tests/data/policy/agent_extension_policy.rego
mgunnala: updated to normal agent license since this will be packaged with agent

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: updated to use osutil

azurelinuxagent/ga/policy/regorus.py
```diff
@@ -0,0 +1,169 @@
...
+                }, ...
+        }
+        """
+        if isinstance(input_json, dict):
+            self._input_json = input_json
+        elif isinstance(input_json, str):
+            self._input_json = json.loads(input_json)
+        else:
+            raise TypeError("Input to policy engine is not a valid JSON object.")
+        # TODO: Add JSON schema and validate input against it, return detailed error message
```
narrieta, line 84: Do we need to do our own validation here? wouldn't the regorus engine do its own validation?

narrieta: same comment for line 109 below

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,165 @@
...
+        Expected format for query: "data.agent_extension_policy.extensions_to_download"
+        """
+        # This method should never be called if policy is not enabled, this would be a developer error.
+        if not self.policy_engine_enabled:
+            raise PolicyError("Policy enforcement is disabled, cannot evaluate query.")
+
+        try:
+            self._engine.set_input(input_json)
+            full_result = self._engine.eval_query(query)
+            value = full_result['result'][0]['expressions'][0]['value']
```
narrieta, line 157: How does the output of the engine when the query indicates an extension violates a policy? what is the exit code in that case? the error exist codes documented in regorus.py seem to indicate more issues in the inputs to the engine than extensions not complying with policy.

narrieta (commented):
Adding some comments. 

I think I am still missing something in the way the extension handler will invoke regorus and handle its output and errors (exceptions). If we go over that offline, that may help answer some of my questions. In the meanwhile, added comments where I still need some clarification.

azurelinuxagent/ga/policy/regorus.py
```diff
@@ -0,0 +1,169 @@
...
+                        "extensionSigned": <true, false>
+                    }
+                },
+                "<extension_name_2>": {
+                    "signingInfo": {
+                        "extensionSigned": <true, false>
+                    }
+                }, ...
+        }
+        """
```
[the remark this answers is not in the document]
mgunnala: policy_engine.py will have ExtensionPolicyEngine class that take in an Extension object, create this JSON, and query against the engine (added a TODO for this). Extension Handler will be responsible for initializing that class and passing the Extension object.

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,165 @@
...
+            self.log_policy(msg=msg)
+            raise PolicyError(msg)
+
+    @classmethod
+    def log_policy(cls, msg, is_success=True, op=WALAEventOperation.Policy, send_event=True):
+        """
+        Log information to console and telemetry.
+        """
+        if is_success:
+            logger.info("[Policy] " + msg)
```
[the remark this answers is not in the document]
mgunnala: I thought it made the log file easier to parse/read, and I saw a similar prefix for cgroups so followed the same format. Can remove if it adds clutter.

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,165 @@
...
+        else:
+            logger.error("[Policy] " + msg)
+        if send_event:
+            add_event(op=op, message=msg, is_success=is_success)
+
+    @staticmethod
+    def is_policy_enforcement_enabled():
+        """Check whether user has opted into policy enforcement feature"""
+        # TODO: Remove conf flag post private preview and add other checks here.
```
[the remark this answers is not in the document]
mgunnala: We'll have a better idea after the discussion with PMs and v-team about user experience, but this change is relatively far-out since we're keeping the conf flag for now.

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: The engine will execute successfully (no error). "value" will look something like: 

"Microsoft.Azure.ActiveDirectory.AADSSHLoginForLinux": {
       "downloadAllowed": true 
}, 
"Microsoft.Azure.Extensions.CustomScript": { 
       "downloadAllowed": false
}

ExtensionPolicyEngine.is_extension_download_allowed( ) will be responsible for parsing this output and returning true/false based on the value of "downloadAllowed". I've added examples to the spec doc that will hopefully make this easier to understand.

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,165 @@
...
+            return  # policy_engine_enabled is not set to True
+
+        if not self.is_policy_enforcement_supported():
+            # TODO: Update this error message when conf flag is removed post private preview.
+            msg = "Attempted to enable policy enforcement, but feature is not supported on this platform."
+            self.log_policy(msg=msg)
+            raise PolicyError(msg)
+
+        try:
+            self._engine = regorus.Engine()
```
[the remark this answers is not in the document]
narrieta: OK, I see. You should hide the details of invoking the regorus API or command line in Engine

Let's have the Engine's init method take the rule and policy files, and then have eval_query take the input as parameter

azurelinuxagent/ga/policy/policy_engine.py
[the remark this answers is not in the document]
narrieta: Yes, the prefix for cgroups has a different motivation. Let's remove them.

azurelinuxagent/ga/policy/policy_engine.py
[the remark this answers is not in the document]
narrieta: Can you improve the comment a litte bit in that case? At this point it says that something needs to be done ("add other checks"), but not exactly what. Somebody trying to implement that TODO would be confused

azurelinuxagent/ga/policy/policy_engine.py
narrieta: thanks for the sample

the code above should check that 'result', 'expressions' and 'value' are in the output and issue an appropriate error message

azurelinuxagent/ga/policy/regorus.py
mgunnala: Regorus engine does do its own validation, but the error messages/codes that are popped up are not very clear to the user. The intention behind pre-validating the input is to clearly identify any issues and provide detailed error messages. 

But I am removing these validations now. We can request the Regorus team to improve clarity of error messages and return codes.

azurelinuxagent/ga/policy/policy_engine.py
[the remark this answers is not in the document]
mgunnala: Removed

azurelinuxagent/common/exception.py
```diff
@@ -153,6 +153,12 @@ class MultiConfigExtensionEnableError(ExtensionError):
     """
 
 
+class PolicyError(AgentError):
```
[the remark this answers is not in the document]
mgunnala: Moved

azurelinuxagent/ga/policy/policy_engine.py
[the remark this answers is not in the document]
mgunnala: Updated

azurelinuxagent/ga/policy/regorus.py
```diff
@@ -0,0 +1,169 @@
...
+        self._engine = None
+        self._rule_file = None
+        self._policy_file = None
+        self._input_json = None
+
+    def add_policy(self, rule_file):
+        """Rule_file is expected to point to a valid Regorus file."""
+        if not os.path.exists(rule_file):
+            raise IOError("Policy rule file path {0} does not exist".format(rule_file))
+        if not rule_file.endswith('.rego'):
```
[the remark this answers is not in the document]
mgunnala: issues with rule file should be caught by our unit tests, I removed this validation

azurelinuxagent/ga/policy/regorus.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def __init__(self):
+        self._engine = None
+        self._rule_file = None
+        self._policy_file = None
+        self._input_json = None
+
+    def add_policy(self, rule_file):
+        """Rule_file is expected to point to a valid Regorus file."""
+        if not os.path.exists(rule_file):
```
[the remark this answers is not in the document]
mgunnala: intention was to provide clear debug messages for developers, since regorus doesn't. but unit tests should catch this, removed the check.

azurelinuxagent/ga/policy/regorus.py
```diff
@@ -0,0 +1,169 @@
...
+            },
+            "azureGuestExtensionsPolicy": {
+                "allowed_ext_1": {
+                    "signingRules": {
+                        "extensionSigned": <true, false>
+                    }
+                }
+        }
+        """
+        if not os.path.exists(policy_file):
```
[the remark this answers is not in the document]
mgunnala: removed

azurelinuxagent/ga/policy/policy_engine.py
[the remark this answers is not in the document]
mgunnala: removed

azurelinuxagent/ga/policy/policy_engine.py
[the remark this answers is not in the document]
mgunnala: Updated

azurelinuxagent/ga/policy/regorus.py
mgunnala: Done

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: added this validation

azurelinuxagent/ga/policy/regorus.py
```diff
@@ -0,0 +1,169 @@
...
+        if not rule_file.endswith('.rego'):
+            raise TypeError("Policy rule file path {0} is not a valid .rego file.".format(rule_file))
+        self._rule_file = rule_file
+        # TODO: Currently, eval_query will raise error if policy file is invalid.
+        #  Consider a dummy "regorus eval" or "regorus parse" call here to validate policy rule file beforehand.
+
+    def set_input(self, input_json):
+        """
+        This sets the input to the policy engine query, for example, a list of extensions we want to download.
+        Input_json should be a JSON object.
```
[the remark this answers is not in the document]
mgunnala: just accepts a dictionary now. updated.

azurelinuxagent/ga/policy/regorus.py
```diff
@@ -0,0 +1,169 @@
...
+                    "signingRules": {
+                        "extensionSigned": <true, false>
+                    }
+                }
+        }
+        """
+        if not os.path.exists(policy_file):
+            raise IOError("Policy parameter file path {0} does not exist.".format(policy_file))
+        if not policy_file.endswith(".json"):
+            raise TypeError("Policy parameter file path {0} is not a valid JSON file.".format(policy_file))
```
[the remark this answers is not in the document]
mgunnala: I originally had a todo to add json schema validation here, removed it

invalid json format will cause engine to fail. unexpected json schema will result in an empty query result, but still execute successfully. the query output validation should be done in PolicyEngine/ExtensionPolicyEngine and error will be raised to the user there.

azurelinuxagent/ga/policy/regorus.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+
+def get_regorus_path():
+    """
+    Returns path to Regorus executable. The executable is not yet officially released as part of the agent package.
+    After release, executable will be placed at /var/lib/waagent/WALinuxAgent-x.x.x.x/bin/regorus.
+    Currently, the executable is copied into the agent directory for unit testing, and this method is mocked.
+    """
+    regorus_exe = os.path.join(os.getcwd(), "bin", "regorus")
```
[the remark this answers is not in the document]
mgunnala: Will address in the next PR, marked as a todo here. I left a dummy method here for unit testing purposes, but let me know if you need me to remove it completely.

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,172 @@
...
+                "policyVersion": "0.1.0",
+                "signingRules": {
+                    "extensionSigned": <true, false>
+                },
+                "allowListOnly": <true, false>
+            },
+            "azureGuestExtensionsPolicy": {
+                "allowed_ext_1": {
+                    "signingRules": {
+                        "extensionSigned": <true, false>
```
maddieford, line 64: Will the agent just ignore 'azureGuestExtensionsPolicy' if 'allowListOnly' is false?

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,172 @@
...
+        }
+        """
+        self._policy_engine_enabled = False
+        self._engine = None
+        if not self.is_policy_enforcement_enabled():
+            self.log_policy(msg="Policy enforcement is not enabled.")
+            return  # policy_engine_enabled is not set to True
+
+        if not self.is_policy_enforcement_supported():
+            msg = "Attempted to enable policy enforcement, but feature is not supported on this platform."
```
maddieford, line 76: We should probably specify what isn't supported in this message (arch or distro/version)

narrieta: +1

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,172 @@
...
+        osutil = get_osutil()
+        arch = osutil.get_vm_arch()
+        if arch not in POLICY_SUPPORTED_ARCHITECTURE:
+            return False
+        __distro__ = get_distro()
+        DISTRO_NAME = __distro__[0]
+        DISTRO_VERSION = __distro__[1]
+        try:
+            distro_version = DistroVersion(DISTRO_VERSION)
+        except ValueError:
```
narrieta, line 120: is this left-over code?

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,172 @@
...
+            return distro_version >= min_version
+        else:
+            return False
+
+    @property
+    def policy_engine_enabled(self):
+        """This property tracks whether the feature is enabled and Regorus engine has been successfully initialized"""
+        return self._policy_engine_enabled
+
+    def evaluate_query(self, input_dict, query):
```
narrieta, line 135: can we use a better name than "input_dict"?

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,172 @@
...
+        {
+            "extensions": {
+                "<extension_name_1>": {
+                    "signingInfo": {
+                        "extensionSigned": <true, false>
+                    }
+                }, ...
+        }
+
+        Expected format for query: "data.agent_extension_policy.extensions_to_download"
```
narrieta, line 147: what does "data.agent_extension_policy.extensions_to_download"? an extension name?

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,172 @@
...
+        """
+        # This method should never be called if policy is not enabled, this would be a developer error.
+        if not self.policy_engine_enabled:
+            raise PolicyError("Policy enforcement is disabled, cannot evaluate query.")
+
+        try:
+            full_result = self._engine.eval_query(input_dict, query)
+            if not full_result.get('result') or not isinstance(full_result['result'], list):
+                raise PolicyError("query returned unexpected output with no 'result' list. Please validate rule file.")
```
narrieta, line 156: could you add the full_result to the message? it would be needed for debugging.

while doing that, adding the path to the rule file would also be helpful (it was mentioned before that there may be multiple fules files)

narrieta: similar comment for other messages

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: "input_to_query" or "input_to_check" maybe?

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: This is the value we're querying for in the regorus engine. We can query for "data", "data.agent_extension_policy", "data.agent_extension_policy.extensions_to_download", etc.

"data.agent_extension_policy.extensions_to_download" will return output in the format:   
```
"extensions_to_download": { 
        "Microsoft.Azure.ActiveDirectory.AADSSHLoginForLinux": { 
            "downloadAllowed": true 
        }, 
        "Microsoft.Azure.Extensions.CustomScript": { 
            "downloadAllowed": false 
        } 
    }
```
...
I've put a full example in the spec.

tests/ga/test_policy_engine.py
```diff
@@ -0,0 +1,142 @@
...
+                                           msg="Policy should not be enabled on unsupported distro {0} {1}".format(distro_name, version)):
+                        PolicyEngine(self.default_rule_path, self.default_policy_path)
+
+    def test_should_raise_exception_on_unsupported_architecture(self):
+        """Policy should NOT be enabled on ARM64."""
+        # TODO: remove this test when support for ARM64 is added.
+        with patch('azurelinuxagent.ga.policy.policy_engine.get_osutil') as mock_get_osutil:
+            with patch('azurelinuxagent.ga.policy.policy_engine.conf.get_extension_policy_enabled', return_value=True):
+                with self.assertRaises(PolicyError, msg="Policy should not be enabled on unsupported architecture ARM64, should have raised exception."):
+                    mock_osutil = MagicMock()
```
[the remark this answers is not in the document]
mgunnala: Good catch, thanks. I've fixed it.

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: Added!

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: I changed this function implementation based on the other comments, removed this block too.

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,172 @@
...
+
+    @staticmethod
+    def is_policy_enforcement_supported():
+        """Check that both platform architecture and distro/version are supported."""
+        osutil = get_osutil()
+        arch = osutil.get_vm_arch()
+        if arch not in POLICY_SUPPORTED_ARCHITECTURE:
+            return False
+        __distro__ = get_distro()
+        DISTRO_NAME = __distro__[0]
```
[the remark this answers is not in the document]
mgunnala: Ah, that makes sense. Updated.

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: Changed the implementation so we raise an error with information about what isn't supported.

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,175 @@
...
+POLICY_SUPPORTED_ARCHITECTURE = ['x86_64']
+
+
+class PolicyEngine(object):
+    """
+    Implements policy engine API. Class will always be initialized, but if the Regorus import fails,
+    all methods will be no-ops.
+
+    If any errors are thrown in regorus.py, they will be caught and handled here. add_policy, add_data,
+    and set_input will be no-ops, eval_query will return an empty dict
```
maddieford, line 44: Is this comment still true? 
> add_policy, add_data, and set_input will be no-ops, eval_query will return an empty dict
Seems like we updated to raise exc instead of no-op

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,175 @@
...
+                        "extensionSigned": <true, false>
+                    }
+                }, ...
+        }
+
+        The query parameter specifies the value we want to retrieve from the policy engine.
+        Example format for query: "data.agent_extension_policy.extensions_to_download"
+        """
+        # This method should never be called if policy is not enabled, this would be a developer error.
+        if not self.policy_engine_enabled:
```
maddieford, line 152:
```suggestion
        if not self._policy_engine_enabled:
```
or 
```suggestion
        if not self.policy_engine_enabled():
```

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,175 @@
...
+            raise PolicyError("Policy enforcement is disabled, cannot evaluate query.")
+
+        try:
+            full_result = self._engine.eval_query(input_to_check, query)
+            debug_info = "Rule file is located at '{0}'. \nFull query output: {1}".format(self._engine.rule_file, full_result)
+            if not full_result.get('result') or not isinstance(full_result['result'], list):
```
maddieford, line 158: `not full_result.get('result')`
Are we checking for None here? Looks like 'result' value type is list so we should probably do explicit check for None here instead of 'not full_result...'
```suggestion
            if full_result.get('result') is None or not isinstance(full_result['result'], list):
```

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,175 @@
...
+                raise PolicyError("query returned unexpected output with no 'result' list. Please validate rule file. {0}".format(debug_info))
+            expressions = full_result['result'][0].get('expressions')
+            if not expressions or not isinstance(expressions, list) or len(expressions) == 0:
```
maddieford, line 161: same comment here, check if expressions is None explicitly

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,175 @@
...
+                raise PolicyError("query returned unexpected output with no 'expressions' list. {0}".format(debug_info))
+            value = expressions[0].get('value')
+            if not value:
```
maddieford, line 164: same here

azurelinuxagent/ga/policy/regorus.py
```diff
@@ -0,0 +1,146 @@
...
+            0 - successful query. optional parameters may be missing
+            1 - file error: unsupported file type, error parsing file, file not found
+                ex: "Error: Unsupported data file <file>. Must be rego or json."
+                ex: "Error: Failed to read <file>. No such file or directory."
+            2 - usage error: missing query argument, unexpected or unlabeled parameter
+                ex: "Error: the following required arguments were not provided: <QUERY>"
+                ex: "Error: Unexpected argument <arg> found."
+        """
+        # Write input_to_check to a temp file, because Regorus requires input to be a file path.
+        # Tempfile is automatically cleaned up at the end of with block
```
maddieford, line 122: Is there a typo here? 

> Tempfile is automatically cleaned up at the end of (?) with block

azurelinuxagent/common/conf.py
```diff
@@ -683,7 +684,16 @@ def get_firewall_rules_log_period(conf=__conf__):
     """
     return conf.get_int("Debug.FirewallRulesLogPeriod", 86400)
 
+  
+def get_extension_policy_enabled(conf=__conf__):
+    """
+    Determine whether extension policy is enabled. If true, Regorus will be installed on the VM and
```
narrieta, line 690: nit: missed this one before: "If true, Regorus will be installed on the VM". Regorus is always installed on the VM, no matter the value of the flag

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,175 @@
...
+    'mariner': DistroVersion('2'),
+    'azurelinux': DistroVersion('3')
+}
+# TODO: add 'arm64', 'aarch64' here once support is enabled for ARM64
+POLICY_SUPPORTED_ARCHITECTURE = ['x86_64']
+
+
+class PolicyEngine(object):
+    """
+    Implements policy engine API. Class will always be initialized, but if the Regorus import fails,
```
narrieta, line 40: left-over comment; we removed the try/except of ImportError in a previous iteration

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,175 @@
...
+    all methods will be no-ops.
+
+    If any errors are thrown in regorus.py, they will be caught and handled here. add_policy, add_data,
```
narrieta, line 43: left-over comment -- i think those methods are now gone

narrieta, line 43: there are some related left-over comments in the unit tests as well

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,175 @@
...
+            self.log_policy(msg="Policy enforcement is not enabled.")
+            return  # policy_engine_enabled is not set to True
+
+        # If unsupported, this call will raise an error
+        self.check_policy_enforcement_supported()
+
+        try:
+            self._engine = regorus.Engine(policy_file=policy_file, rule_file=rule_file)
+            self._policy_engine_enabled = True
+        except Exception as ex:
```
narrieta, line 81: is this try/catch a leftover of a previous iteration? at this point the initialization is just

```
        self._engine = None
        self._rule_file = rule_file
        self._policy_file = policy_file
```

azurelinuxagent/ga/policy/policy_engine.py
narrieta: Whoa! great catch @maddieford! 

pylint disappoints me more often than I'd like. But if the linter fails, the next step would be unit tests. @mgunnala - would you review unit tests and see why they missed this basic scenario? thanks!

azurelinuxagent/ga/policy/policy_engine.py
narrieta: +1

@mgunnala We avoid those implicit comparisons. We check "if foo/if not foo" only for actual bools. Comparisons against None, empty list, empty string, etc, should be explicit

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,175 @@
...
+        """
+        if is_success:
+            logger.info(msg)
+        else:
+            logger.error(msg)
+        if send_event:
+            add_event(op=op, message=msg, is_success=is_success)
+
+    @staticmethod
+    def is_policy_enforcement_enabled():
```
narrieta, line 99: should this method be private, or are we expecting callers to check this before invoking other methods in PolicyENgine?

narrieta: similar question for check_policy_enforcement_supported()

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,175 @@
...
+                msg += " distro " + DISTRO_NAME + " " + DISTRO_VERSION + ". Policy is only supported on version " + \
+                        str(min_version) + " and above."
+            else:
+                return  # do nothing if platform is supported
+        raise PolicyError(msg)
+
+
+    @property
+    def policy_engine_enabled(self):
+        """This property tracks whether the feature is enabled and Regorus engine has been successfully initialized"""
```
narrieta, line 132: left-over comment: "and Regorus engine has been successfully initialized""

azurelinuxagent/ga/policy/policy_engine.py
narrieta: I see, looks like policy_engine_enabled() is a left-over from a previous iteration of the code and that the other 2 methods are invoked from init() and should be private

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: @maddieford not sure I'm understanding the issue - this came from a previous comment to make policy_engine_enabled( ) into a @property method. My understanding is we can just access "self.policy_engine_enabled" since it's acting as a property. Is this not the case?

azurelinuxagent/ga/policy/regorus.py
mgunnala: No typo, the tempfile is cleaned up at the end of the "with" block

azurelinuxagent/ga/policy/policy_engine.py
narrieta: @mgunnala - the parenthesis are missing from the function call. As written currently, that condition is always True

maddieford points out that you should add the parenthesis or use the private property

narrieta: This is a bad bug, easy to miss. The linter (nor the unit tests) didn't point it out.

narrieta: ah, it's a property!

mgunnala: I thought @property methods in python didn't require the parentheses.

I have a unit test looking for exactly this scenario - "test_eval_query_should_throw_error_when_disabled()". It passes, and when I step through it, the function is being called as expected.

narrieta: yes, i think we misread the code

narrieta: but, related to this property.. is this needed at all? see my other comment. looks like a left-over

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: policy_engine_enabled( ) will be used in the future by ExtensionPolicyEngine, in the function is_extension_download_allowed( ). 

The other two methods can be made private, since ExtensionPolicyEngine can just check the policy_engine_enabled propety.

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: Replied to your other comment

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -0,0 +1,175 @@
...
+        try:
+            self._engine = regorus.Engine(policy_file=policy_file, rule_file=rule_file)
+            self._policy_engine_enabled = True
+        except Exception as ex:
+            msg = "Failed to initialize Regorus policy engine: '{0}'".format(ex)
+            self.log_policy(msg=msg)
+            raise PolicyError(msg)
+
+    @classmethod
+    def log_policy(cls, msg, is_success=True, op=WALAEventOperation.Policy, send_event=True):
```
narrieta, line 87: should be private

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: removed this property

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: Added explicit None checks for full_result, result, expression and value

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: done

azurelinuxagent/ga/policy/policy_engine.py
narrieta: thanks

narrieta (approved):

maddieford (approved):

PR #3189, opened 2024-08-21
send updatemode in heartbeat and don't send RSM supported feature flag if versioning disabled in agent

nagworld9:
## Description

PR includes minor changes related to RSM

- Added updatedmode: RSM/SelfUpdate flag in heartbeat (Telemetry helps to determine which/how may vms enrolled into RSM)
- Fixing supported feature flag for rsm, when agent does not support GA versioning, we still send as agent supports RSM to CRP. In that case, CRP may send rsm requests and those requests will not be honored by agent. So now don't send RSM supported feature flag if versioning disabled in agent.
- For instance, agent is on version N and rsm update attempt failed few times for version N+1 and we send the status to CRP with failed status. Later if we get new request rsm version is same as current version(N), we still sending cached last error msg as rsm status. Fixing that too.

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/ga/agent_update_handler.py
+++ b/azurelinuxagent/ga/agent_update_handler.py
@@ -29,6 +29,14 @@ from azurelinuxagent.ga.rsm_version_updater import RSMVersionUpdater
 from azurelinuxagent.ga.self_update_version_updater import SelfUpdateVersionUpdater
 
 
+class UpdateMode(object):
+    """
+    Enum for Update modes
+    """
+    RSM = "RSM"
+    SelfUpdate = "SelfUpdate"
+
+
 def get_agent_update_handler(protocol):
     return AgentUpdateHandler(protocol)
 
@@ -138,6 +146,15 @@ class AgentUpdateHandler(object):
                     family, self._gs_id))
         return agent_family_manifests[0]
 
+    def get_current_update_mode(self):
+        """
+        Returns current update mode whether RSM or Self-Update
+        """
+        if isinstance(self._updater, RSMVersionUpdater):
+            return UpdateMode.RSM
+        else:
+            return UpdateMode.SelfUpdate
+
     def run(self, goal_state, ext_gs_updated):
 
         try:
@@ -180,6 +197,9 @@ class AgentUpdateHandler(object):
             self._updater.retrieve_agent_version(agent_family, goal_state)
 
             if not self._updater.is_retrieved_version_allowed_to_update(agent_family):
+                # Reset the last error report msg since when this condition met there is no issue with update, either requested
+                # version is same as current version or below than daemon version. In that case, we shouldn't report any error msg in the status report
+                self._last_attempted_update_error_msg = ""
                 return
             self._updater.log_new_agent_update_message()
             agent = self._updater.download_and_get_new_agent(self._protocol, agent_family, goal_state)
--- a/azurelinuxagent/ga/agent_update_handler.py
+++ b/azurelinuxagent/ga/agent_update_handler.py
@@ -197,7 +197,7 @@ class AgentUpdateHandler(object):
             self._updater.retrieve_agent_version(agent_family, goal_state)
 
             if not self._updater.is_retrieved_version_allowed_to_update(agent_family):
-                # Reset the last error report msg since when this condition met there is no issue with update, either requested
+                # Reset the last error report msg  when this condition met as there is no issue with update since requested
                 # version is same as current version or below than daemon version. In that case, we shouldn't report any error msg in the status report
                 self._last_attempted_update_error_msg = ""
                 return
--- a/azurelinuxagent/ga/agent_update_handler.py
+++ b/azurelinuxagent/ga/agent_update_handler.py
@@ -164,6 +164,8 @@ class AgentUpdateHandler(object):
 
             # Update the state only on new goal state
             if ext_gs_updated:
+                # Reset the last reported update state on new goal state before we attempt update otherwise we keep reporting the last update error if any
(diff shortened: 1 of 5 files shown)


azurelinuxagent/ga/update.py
```diff
@@ -1016,22 +1016,23 @@ def _write_pid_file(self):
...
 
-    def _send_heartbeat_telemetry(self, protocol):
+    def _send_heartbeat_telemetry(self, protocol, agent_update_handler):
         if self._last_telemetry_heartbeat is None:
             self._last_telemetry_heartbeat = datetime.utcnow() - UpdateHandler.TELEMETRY_HEARTBEAT_PERIOD
 
         if datetime.utcnow() >= (self._last_telemetry_heartbeat + UpdateHandler.TELEMETRY_HEARTBEAT_PERIOD):
             dropped_packets = self.osutil.get_firewall_dropped_packets(protocol.get_endpoint())
-            auto_update_enabled = 1 if conf.get_autoupdate_enabled() else 0
+            auto_update_enabled = 1 if conf.get_auto_update_to_latest_version() else 0
```
nagworld9, line 1025: we forgot update this flag before, it should check new flag

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -180,6 +197,9 @@ def run(self, goal_state, ext_gs_updated):
             self._updater.retrieve_agent_version(agent_family, goal_state)
 
             if not self._updater.is_retrieved_version_allowed_to_update(agent_family):
+                # Reset the last error report msg  when this condition met as there is no issue with update since requested
+                # version is same as current version or below than daemon version. In that case, we shouldn't report any error msg in the status report
```
maddieford, line 201: This comment makes it sound like 'is_retrieved_version_allowed_to_update' is only False if requested version is same as current version or less than daemon version. We could also hit this condition if agent_family.is_version_from_rsm is False (RSM Update) or if update is not allowed due to upgrade frequency (self update). 

Logic makes sense to me, but comment is a little confusing

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -180,6 +197,9 @@ def run(self, goal_state, ext_gs_updated):
...
+                self._last_attempted_update_error_msg = ""
```
maddieford, line 202: Should we also reset the _last_attempted_update_error_msg in the condition that update is not allowed right now (line 194)?

azurelinuxagent/ga/agent_update_handler.py
narrieta: Since to me more like we need to reset it on a new goal state. 

@nagworld9 CRP needs this message for the goal state requesting the update, but does it still need it for other goal states? If it does, we are losing it on service restart, since it's only in-memory.

narrieta: I think you are answering my question in the next comment below.

However, we should not rely on that and just stop sending it.

`In other times, we are ok send cache value since nobody looking at that status (not even crp)
`

azurelinuxagent/ga/update.py
```diff
@@ -1016,22 +1016,23 @@ def _write_pid_file(self):
...
+            update_mode = agent_update_handler.get_current_update_mode()
 
-            telemetry_msg = "{0};{1};{2};{3};{4}".format(self._heartbeat_counter, self._heartbeat_id, dropped_packets,
+            telemetry_msg = "{0};{1};{2};{3};{4};{5}".format(self._heartbeat_counter, self._heartbeat_id, dropped_packets,
```
narrieta, line 1028: I think we need to add the names to this message, we are just sending the values and relying on their position to parse.

My firewall changes stop reporting the dropped_packets, and that is going to change the position of each value.

Are we parsing these values anywhere currently? (I know you want to start parsing for the update mode)

nagworld9: I doubt we parse it anywhere. why I need to look for position and I just need to know string present in message. 

`I think we need to add the names to this message,`

For some reason, we only log detail log with names and sending telemetry without names. I can change that and send debug_log_msg instead telemetry_msg

```
            telemetry_msg = "{0};{1};{2};{3};{4}".format(self._heartbeat_counter, self._heartbeat_id, dropped_packets,
                                                         self._heartbeat_update_goal_state_error_count,
                                                         auto_update_enabled)
            debug_log_msg = "[DEBUG HeartbeatCounter: {0};HeartbeatId: {1};DroppedPackets: {2};" \
                            "UpdateGSErrors: {3};AutoUpdate: {4}]".format(self._heartbeat_counter,
                                                                          self._heartbeat_id, dropped_packets,
                                                                          self._heartbeat_update_goal_state_error_count,
                                                                          auto_update_enabled)

            # Write Heartbeat events/logs
            add_event(name=AGENT_NAME, version=CURRENT_VERSION, op=WALAEventOperation.HeartBeat, is_success=True,
                      message=telemetry_msg, log_event=False)
            logger.info(u"[HEARTBEAT] Agent {0} is running as the goal state agent {1}", CURRENT_AGENT, debug_log_msg)
```

azurelinuxagent/ga/agent_update_handler.py
nagworld9: makes sense, we should reset on new goal state and we should send last report msg until new goal state

azurelinuxagent/ga/update.py
narrieta: sounds good

narrieta (approved):

maddieford (approved):

PR #3190, opened 2024-08-24
Fix JIT for FIPS 140-3

narrieta:
Extensions with protected settings and JIT requests are both broken under FIPS 140-3.

The fix for extensions requires fixes on other Azure components, including CRP, Fabric and the Extensions themselves. The corresponding fixes in the Agent will be addressed along with those.

Fixing JIT requests is a lot simpler, though, since it only involves changing the cipher user for RemoteAccess requests. This PR does that. The PR also includes a fix in protocol detection caused by the mentioned issues in Extensions (more details are given in the code).


nagworld9 (approved):

nagworld9:
changes LGTM, I don't know how to validate JIT flow though

nagworld9:
Discussed offline, changes looks good

maddieford (approved):

PR #3199, opened 2024-08-28
minor fixes in e2e tests ext_cgroups, agent_update and agent_publish

nagworld9:
## Description

ext_cgrups: run_remote_test swallowing the assertion error, so test failed to validate if assertion failed due to cgroups disabled. So removed calling run_remote_test

agent_update: when rsm_request sent, whole pipeline can take some time to update the goal state with the requested version, so increasing the timeout on that check

agent_publish: Tests are running before agent is ready. Added check for wait for agent to complete provisioning

Issue # 

---

### PR information


tests_e2e/tests/agent_publish/agent_publish.py
```diff
@@ -68,6 +71,22 @@ def get_ignore_errors_before_timestamp(self) -> datetime:
         timestamp = self._ssh_client.run_command("agent_publish-get_agent_log_record_timestamp.py")
         return datetime.strptime(timestamp.strip(), u'%Y-%m-%d %H:%M:%S.%f')
 
+    def _wait_for_agent_to_complete_provisioning(self):
```
narrieta, line 74: I'd consider moving this somewhere in testss/lib, since it looks like a useful function that other tests could use

tests_e2e/tests/agent_publish/agent_publish.py
```diff
@@ -68,6 +71,22 @@ def get_ignore_errors_before_timestamp(self) -> datetime:
...
+        for _ in range(5):
+            time.sleep(30)
+            try:
+                self._ssh_client.run_command("[ -f /var/lib/waagent/provisioned  ] && exit 0 || exit 1", use_sudo=True)
+                break
+            except CommandError:
+                log.info("Waiting for agent to complete provisioning, will check again after a short delay")
+
+        else:
+            fail("Timeout while waiting for the Agent to complete provisioning")
```
narrieta, line 88: I think this should be "raise Exception" rather than "fail". 

Both of them raise an exception, but fail/assert/etc should be used when the verifications the test is doing completed, but were not True. Other exceptions would be used when the test was trying to setup something, or waiting for something before doing its verifications like in this case, etc, and some error prevented the test to perform its verifications.

maddieford (approved):

PR #3208, opened 2024-09-09
handle non-boolean values in handlermanifest

nagworld9:
## Description
We learned recently that some extensions are sending string value "true" or "false" for boolean flags in handlermanifest. This led to wrong matching in the agent. For ex: string "false" matched as true since we assume they are boolean value and condition _not flag_ became as flag has value or not instead flag has true or false

Addressing this pr with reporting warning plus default to false if they send non-boolean value except "true". For backward compatibility, 'true' (case insensitive) is accepted.

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -2308,7 +2308,7 @@ class HandlerManifest(object):
         for backward compatibility, 'true' (case-insensitive) is accepted, and other values default to False
         """
         if not isinstance(value, bool):
-            return True if value.lower() == "true" else default_val
+            return True if isinstance(value, str) and value.lower() == "true" else default_val
         return value
 
 
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -2296,10 +2296,10 @@ class HandlerManifest(object):
         for key in ['reportHeartbeat', 'continueOnUpdateFailure', 'supportsMultipleExtensions']:
             value = self.data['handlerManifest'].get(key, False)
             if not isinstance(value, bool):
-                msg =("In the handler manifest: '{0}' has a non-boolean value {1} for boolean type. Please change it to a boolean value. "
+                msg =("In the handler manifest: '{0}' has a non-boolean value [{1}] for boolean type. Please change it to a boolean value. "
                       "For backward compatibility, 'true' (case insensitive) is accepted, and other values default to False.").format(key, value)
                 logger.warn(msg)
-                add_event(name=ext_name, is_success=False, message=msg, op=WALAEventOperation.ExtensionHandlerManifest)
+                add_event(name=ext_name, is_success=False, message=msg, op=WALAEventOperation.ExtensionHandlerManifest, log_event=False)
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


maddieford (dismissed):

maddieford (dismissed):

azurelinuxagent/ga/exthandlers.py
```diff
@@ -2283,6 +2289,28 @@ def get_resource_limits(self, extension_name, str_version):
 
         return ResourceLimits(self.data.get('resourceLimits', None))
 
+    def report_invalid_boolean_properties(self, ext_name):
+        """
+        Check that the specified keys in the handler manifest has boolean values.
+        """
+        for key in ['reportHeartbeat', 'continueOnUpdateFailure', 'supportsMultipleExtensions']:
+            value = self.data['handlerManifest'].get(key, False)
+            if not isinstance(value, bool):
```
narrieta, line 2298: I think this expresses the intention of the check a little better:

```
            value = self.data['handlerManifest'].get(key)
            if value is not None and not isinstance(value, bool):
```

azurelinuxagent/ga/exthandlers.py
```diff
@@ -2283,6 +2289,28 @@ def get_resource_limits(self, extension_name, str_version):
...
+                msg =("In the handler manifest: '{0}' has a non-boolean value [{1}] for boolean type. Please change it to a boolean value. "
+                      "For backward compatibility, 'true' (case insensitive) is accepted, and other values default to False.").format(key, value)
```
narrieta, line 2300: I would not advertise this, since we do not want to encourage its use

azurelinuxagent/ga/exthandlers.py
```diff
@@ -2283,6 +2289,28 @@ def get_resource_limits(self, extension_name, str_version):
...
+                logger.warn(msg)
```
narrieta, line 2301: Many times users/support are concerned about Warning in the Agent's log. I'd so this as INFO and just send telemetry. After all, this message is not meant for the user, but for the extension publisher

azurelinuxagent/ga/exthandlers.py
```diff
@@ -2283,6 +2289,28 @@ def get_resource_limits(self, extension_name, str_version):
...
+                add_event(name=ext_name, is_success=False, message=msg, op=WALAEventOperation.ExtensionHandlerManifest, log_event=False)
+
+    @staticmethod
+    def _parse_boolean_value(value, default_val=False):
```
narrieta, line 2305: let's not have a default for default_val; in this case, forcing the caller to explicitly specify the default value makes the code more readable

azurelinuxagent/ga/exthandlers.py
nagworld9: you mean this part `"For backward compatibility, 'true' (case insensitive) is accepted, and other values default to False.`

narrieta: Yup, we don't want to encourage the use of strings as booleans. We need to supported for extensions that already do it, but we should not let other extensions know that we do.

narrieta (approved):

maddieford (approved):

PR #3211, opened 2024-09-13
unicode type check

nagworld9:
## Description

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -2305,9 +2305,11 @@ class HandlerManifest(object):
         """
         Expects boolean value but
         for backward compatibility, 'true' (case-insensitive) is accepted, and other values default to False
+        Note: In py2, string can be unicode or str type. In py3, unicode removed
+        ustr is a unicode object for Py2 and a str object for Py3.
         """
         if not isinstance(value, bool):
-            return True if isinstance(value, str) and value.lower() == "true" else default_val
+            return True if isinstance(value, ustr) or isinstance(value, str) and value.lower() == "true" else default_val
         return value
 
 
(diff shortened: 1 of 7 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -2305,9 +2305,11 @@ def _parse_boolean_value(value, default_val):
         """
         Expects boolean value but
         for backward compatibility, 'true' (case-insensitive) is accepted, and other values default to False
+        Note: In py2, string can be unicode or str type. In py3, unicode removed
+        ustr is a unicode object for Py2 and a str object for Py3.
         """
         if not isinstance(value, bool):
-            return True if isinstance(value, str) and value.lower() == "true" else default_val
+            return True if isinstance(value, (ustr, str)) and value.lower() == "true" else default_val
```
narrieta, line 2312:
```suggestion
            return True if isinstance(value, (ustr, str)) and value.lower() == "true" else default_val
```
```suggestion
            return True if value.lower() == ustr("true") else default_val
```

narrieta, line 2312: need unit test

nagworld9: Initial intention was protecting against non-string/non-boolean and use default otherwise value.lower() will raise an error and will not process the extension.

nagworld9: is your intention we are ok break the extension in that case? non-string/boolean totally unacceptable?

tests/ga/test_extension.py
```diff
@@ -3527,6 +3527,23 @@ def test_handler_manifest_boolen_fields(self):
         self.assertFalse(manifest.is_report_heartbeat())
         self.assertFalse(manifest.supports_multiple_extensions())
 
+    def test_unicode_values_for_boolean_fields_in_handler_manifest(self):
+        handler_json = {
```
narrieta, line 3531: We should do this test loading the json from a file, which is what exposed the problem

Other similar tests should probably also be modified

narrieta (approved):

PR #3221, opened 2024-09-19
Add support for nftables + refactoring of firewall code

narrieta:
Adding support for nftables.

The code for setting up the firewall was scattered across multiple places and was not straightforward to follow up. Adding support for nftables on top of the existing code would have made things worse, so this PR also includes significant refactoring of that code and associated tests.

I also moved the code from the 'common' subdirectory to 'ga', since it is not needed by the provisioning agent.


azurelinuxagent/agent.py
```diff
@@ -270,15 +270,16 @@ def collect_logs(self, is_full_mode):
...
+        logger.set_prefix("Firewall")
+        threading.current_thread().name = "Firewall"
+        event.info(event.WALAEventOperation.Firewall, "Setting up firewall after boot. Endpoint: {0}", ustr(endpoint))
         try:
-            AddFirewallRules.add_iptables_rules(firewall_metadata['wait'], firewall_metadata['dst_ip'],
-                                                firewall_metadata['uid'])
-            print("Successfully set the firewall rules")
+            firewall_manager = FirewallManager.create(endpoint)
+            firewall_manager.setup()
+            event.info(event.WALAEventOperation.Firewall, "Successfully set the firewall rules")
```
narrieta, line 280: I create some convenience functions to log and send telemetry (see event.py)

azurelinuxagent/agent.py
```diff
@@ -383,20 +380,17 @@ def parse_args(sys_args):
...
-            firewall_metadata['dst_ip'] = re.match(regex_cmd_format.format("dst_ip=(?P<dst_ip>[\\d.]{7,})"), arg).group(
-                'dst_ip')
-        elif re.match(regex_cmd_format.format("uid=(?P<uid>[\\d]+)"), arg):
-            firewall_metadata['uid'] = re.match(regex_cmd_format.format("uid=(?P<uid>[\\d]+)"), arg).group('uid')
-        elif re.match(regex_cmd_format.format("(w|wait)$"), arg):
-            firewall_metadata['wait'] = "-w"
         else:
-            cmd = AgentCommands.Help
-            break
+            regex_cmd = regex_cmd_format.format("{0}=(?P<endpoint>[\\d.]{{7,}})".format(AgentCommands.SetupFirewall))
```
narrieta, line 384: now the command line takes only "-setup-firewall=<IP>"

azurelinuxagent/common/event.py
```diff
@@ -331,10 +329,10 @@ def report_dropped_events_error(count, errors, operation_name):
         report_dropped_events_error(self.__unicode_error_count, self.__unicode_errors, self.__unicode_error_event)
 
     @staticmethod
-    def _update_errors_and_get_count(error_count, errors, error):
+    def _update_errors_and_get_count(error_count, errors, error_msg):
```
narrieta, line 332: rename needed because now there is a global "error" function

azurelinuxagent/common/event.py
```diff
@@ -750,9 +772,9 @@ def dump_unhandled_err(name):
         last_type = getattr(sys, 'last_type')
         last_value = getattr(sys, 'last_value')
         last_traceback = getattr(sys, 'last_traceback')
-        error = traceback.format_exception(last_type, last_value,
+        trace = traceback.format_exception(last_type, last_value,
```
narrieta, line 775: rename needed because now there is a global "error" function

azurelinuxagent/common/osutil/default.py
```diff
@@ -69,60 +68,9 @@ def crypt(password, salt):
 if needed.
 """
 
-_IPTABLES_VERSION_PATTERN = re.compile(r"^[^\d\.]*([\d\.]+).*$")
```
narrieta, line 72: all this logic is now in FirewallManager, with several changes

azurelinuxagent/common/protocol/metadata_server_migration_util.py
```diff
@@ -55,20 +59,17 @@ def _cleanup_metadata_protocol_certificates():
...
 
-def _reset_firewall_rules(osutil):
+
+def _reset_firewall_rules():
     """
     Removes MetadataServer firewall rule so IMDS can be used. Enables
     WireServer firewall rule based on if firewall is configured to be on.
     """
-    osutil.remove_firewall(dst_ip=_KNOWN_METADATASERVER_IP, uid=os.getuid(), wait=osutil.get_firewall_will_wait())
-    if conf.enable_firewall():
```
narrieta, line 64: setting up the firewall is already done in the main and env threads; no need to do it here

azurelinuxagent/common/protocol/metadata_server_migration_util.py
```diff
@@ -77,3 +78,102 @@ def _ensure_file_removed(directory, file_name):
     path = os.path.join(directory, file_name)
     if os.path.isfile(path):
         os.remove(path)
+
```
narrieta, line 81: The way the code for cleaning up the metadata is written, it needs to run on the daemon/pa. 

I moved the firewall code from common/ostutil to ga/FirewallManager. There should not be dependencies across the pa/ga code, so I am duplicating a little bit of code here. That code comes practically unchanged from osutil.

This clean up is/was needed when migrating Azure Stack VMs to Azure. Azure Stack used to have its own version of the WireServer (it was named the Metadata protocol) and some differences between those protocols require this cleanup. I do not know if this code is still needed, but it is owned and tested by the Azure Stack team.

azurelinuxagent/common/utils/networkutil.py
```diff
@@ -98,175 +93,3 @@ def __str__(self):
         if len(self.ipv6) > 0:
             entries.append('"ipv6": {0}'.format(self._json_array(self.ipv6)))
         return "{{ {0} }}".format(", ".join(entries))
-
```
narrieta, line 101: Moved to FirewallManager

azurelinuxagent/ga/update.py
```diff
@@ -806,21 +803,6 @@ def _ensure_no_orphans(self, orphan_wait_interval=ORPHAN_WAIT_INTERVAL):
                     ustr(e))
         return
 
-    def _ensure_partition_assigned(self):
```
narrieta, line 809: It seems like at some point there was some functionality to split agent updates across "partitions", and VMs were assigned to a particular partition. Not used anymore, so deleting this code.

azurelinuxagent/ga/update.py
```diff
@@ -1016,22 +991,18 @@ def _write_pid_file(self):
...
             update_mode = agent_update_handler.get_current_update_mode()
 
             # Note: When we add new values to the heartbeat message, please add a semicolon at the end of the value.
             # This helps to parse the message easily in kusto queries with regex
-            heartbeat_msg = "HeartbeatCounter: {0};HeartbeatId: {1};DroppedPackets: {2};" \
-                            "UpdateGSErrors: {3};AutoUpdate: {4};UpdateMode: {5};".format(self._heartbeat_counter,
-                                                                          self._heartbeat_id, dropped_packets,
-                                                                          self._heartbeat_update_goal_state_error_count,
-                                                                          auto_update_enabled, update_mode)
+            heartbeat_msg = "HeartbeatCounter: {0};HeartbeatId: {1};UpdateGSErrors: {2};AutoUpdate: {3};UpdateMode: {4};".format(
```
narrieta, line 1004: We don't use the DroppedPackets telemetry. It could be added to nftables too, but since we don't use it, I am deleting it instead.

tests/common/osutil/test_default.py
```diff
@@ -643,327 +639,6 @@ def test_conf_sudoer(self, mock_dir):
         print("WRITING TO {0}".format(waagent_sudoers))
         self.assertEqual(1, count)
 
-    @staticmethod
```
narrieta, line 646: moved to test_firewall_manager, or deleted, since some of those test don't make sense for the new code

tests/common/utils/test_network_util.py
```diff
@@ -64,34 +60,3 @@ def test_nic_ordinary(self):
         self.assertEqual(str(nic), '{ "name": "test0", "link": "link INFO", "ipv4": ["ipv4-1"], "ipv6": ["ipv6-1"] }')
 
 
-class TestAddFirewallRules(AgentTestCase):
```
narrieta, line 67: Moved to test_firewall_manager

tests_e2e/tests/scripts/agent_persist_firewall-access_wireserver
```diff
@@ -90,15 +90,21 @@ echo "Finally online, Time: $(date --utc +%FT%T.%3NZ)"
...
+else
+  sudo nft list table walinuxagent
+fi
 echo ""
 
 WIRE_IP=$(cat /var/lib/waagent/WireServerEndpoint 2>/dev/null || echo '168.63.129.16' | tr -d '[:space:]')
-if command -v wget >/dev/null 2>&1; then
-    wget --tries=3 "http://$WIRE_IP/?comp=versions" --timeout=5 -O "/tmp/wire-versions-$USER.xml"
-else
+if command -v curl >/dev/null 2>&1; then
```
narrieta, line 102: For some reason, on Azure Linux 3, wget returns success (0) on timeout. I am giving preference to curl instead.

tests_e2e/tests/scripts/agent_persist_firewall-access_wireserver
```diff
@@ -90,15 +90,21 @@ echo "Finally online, Time: $(date --utc +%FT%T.%3NZ)"
...
     curl --retry 3 --retry-delay 5 --connect-timeout 5 "http://$WIRE_IP/?comp=versions" -o "/tmp/wire-versions-$USER.xml"
+elif command -v wget >/dev/null 2>&1; then
+    wget --tries=3 "http://$WIRE_IP/?comp=versions" --timeout=5 -O "/tmp/wire-versions-$USER.xml"else
+else
+    http_get.py "http://168.63.129.16/?comp=versions" --timeout 5 --delay 5 --tries 3
```
narrieta, line 107: I created this simple tool by doing surgery on some of the networking code in the agent. I used it to debug the above issue with wget. Since I already created it, I'm leaving it here as a fallback to curl/wget (initially it was meant as a replacement for them, but just changing the order of curl/wget fixed the original issue)

azurelinuxagent/ga/persist_firewall_rules.py
```diff
@@ -101,67 +98,60 @@ def _is_firewall_service_running():
...
+                WALAEventOperation.PersistFirewallRules,
+                "The permanent firewall rules for Azure Fabric are not setup correctly ({0}), will reset them. Current state:\n{1}", ustr(e), firewall_manager.get_state())
+            firewall_manager.remove()
+            firewall_manager.setup()
+            event.info(WALAEventOperation.PersistFirewallRules, "Reset permanent firewall rules for Azure Fabric:\n{0}", firewall_manager.get_state())
+
+        # Remove custom service if exists to avoid problems with firewalld
         try:
-            AddFirewallRules.remove_firewalld_rules(self._dst_ip, self._uid)
+            fileutil.rm_files(*[self.get_service_file_path(), os.path.join(conf.get_lib_dir(), self.BINARY_FILE_NAME)])
```
nagworld9, line 152: Adding here somewhat confusing to me. since we setup before firewallmanger and then I had to figure there is return to believe that we don't come to here when firewallmanager not running. May be move above as part of custom firewall setup

nagworld9:
I didn't complete my review, will give another round trmw

azurelinuxagent/common/event.py
```diff
@@ -693,6 +691,30 @@ def add_event(name=AGENT_NAME, op=WALAEventOperation.Unknown, is_success=True, d
...
+    """
+     Creates a telemetry event and logs the message as INFO; convenience wrapper over add_event(log_level=logger.LogLevel.INFO...);
+    """
+    logger.info(fmt, *args)
+    add_event(op=op, message=fmt.format(*args), is_success=True)
+
+
+def warn(op, fmt, *args):
+    """
+    Creates a telemetry event and logs the message as INFO; convenience wrapper over add_event(log_level=logger.LogLevel.WARNING...)
```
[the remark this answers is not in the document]
narrieta: thanks; fixed

azurelinuxagent/common/event.py
```diff
@@ -693,6 +691,30 @@ def add_event(name=AGENT_NAME, op=WALAEventOperation.Unknown, is_success=True, d
...
+    """
+    logger.warn(fmt, *args)
+    add_event(op=op, message="[WARNING] " + fmt.format(*args), is_success=False, log_event=False)
+
+
+def error(op, fmt, *args):
+    """
+    Creates a telemetry event and logs the message as INFO; convenience wrapper over add_event(log_level=logger.LogLevel.ERROR...)
```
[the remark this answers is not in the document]
narrieta: thanks; fixed

azurelinuxagent/common/event.py
```diff
@@ -693,6 +691,30 @@ def add_event(name=AGENT_NAME, op=WALAEventOperation.Unknown, is_success=True, d
...
+    """
+    logger.warn(fmt, *args)
```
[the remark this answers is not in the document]
narrieta: thanks; fixed

azurelinuxagent/common/protocol/metadata_server_migration_util.py
```diff
@@ -77,3 +78,102 @@ def _ensure_file_removed(directory, file_name):
...
+    try:
+        # This rule was <= 2.2.25 only, and may still exist on some VMs.  Until 2.2.25
+        # has aged out, keep this cleanup in place.
+        _delete_rule(_get_firewall_delete_conntrack_accept_command(wait, dst_ip))
+
+        _delete_rule(_get_delete_accept_tcp_rule(wait, dst_ip))
+        _delete_rule(_get_firewall_delete_owner_accept_command(wait, dst_ip, uid))
+        _delete_rule(_get_firewall_delete_conntrack_drop_command(wait, dst_ip))
+
+        return True
```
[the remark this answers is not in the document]
narrieta: I decided to copy the existing code and leave here as-is, no cleanup attempted.

azurelinuxagent/ga/env.py
```diff
@@ -99,85 +100,42 @@ def _get_dhcp_client_pid(self):
...
-            is_success=success,
-            log_event=False)
-
-        self._is_first_setup = False
-
-    def _get_firewall_state(self):
         try:
-            return self._osutil.get_firewall_list()
+            if self._firewall_manager is None:
+                self._firewall_manager = FirewallManager.create(self._wire_server_address)
```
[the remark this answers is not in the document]
narrieta: init is invoked on the main thread, while the operation is invoked on the env thread.

Having it here allows for uniform error handling on the env thread, and also retries for creating the firewall manager if for some reason the iniatiliazation fails.

azurelinuxagent/ga/firewall_manager.py
```diff
@@ -0,0 +1,499 @@
...
+    """
+    FirewallManager abstracts the interface for managing the firewall rules on the WireServer address. Concrete implementations
+    provide the underlying functionality using command-line tools that vary across distros (e.g. iptables, firewalld, and nftables.)
+    If a concrete implementation cannot be instantiated because the underlying command-line tool is not available, it must raise
+    a FirewallManagerNotAvailable exception.
+
+    Each implementation must set three rules on the WireServer address:
+
+        * "ACCEPT DNS" - Azure DNS runs on the WireServer address, so all traffic on port 53 must be allowed for all users.
+        * "ACCEPT"     - All traffic from the Agent (which runs as root) must be accepted.
```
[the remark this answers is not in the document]
narrieta: It scopes to root. The GPA will allow scoping to the Agent. The intention of the rule is to allow the Agent to access the WireServer, hence the comment.

azurelinuxagent/ga/firewall_manager.py
```diff
@@ -0,0 +1,499 @@
...
+        if use_wait_option:
+            self._base_command = ["iptables", "-w", "-t", "security"]
+        else:
+            self._base_command = ["iptables", "-t", "security"]
+
+    def _execute_delete_command(self, command):
+        """
+        Continually execute the delete operation until the return code is non-zero or the limit has been reached.
+        """
+        for _ in range(1, 100):
```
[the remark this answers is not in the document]
narrieta: Yes, same thing here. I have no idea why this written this way; might be related to the -w option (for those versions of iptables that do not have it). If time allows, I will look more into this for this release, otherwise will do on a subsequent release.

azurelinuxagent/ga/firewall_manager.py
```diff
@@ -0,0 +1,499 @@
...
+        return "--passthrough"
+
+    def _get_check_command_option(self):
+        return "--query-passthrough"
+
+    def _get_delete_command_option(self):
+        return "--remove-passthrough"
+
+
+class NfTables(FirewallManager):
```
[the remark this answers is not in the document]
narrieta: FirewallManager.create() instantiate the correct class depending on the tools present on the system. The derived classes can be instantiated directly when the caller already knows what tool to use.

azurelinuxagent/ga/persist_firewall_rules.py
narrieta: Not sure I understand the concern. I think the intention of the original code was to remove the file when initially firewalld was not running and the Agent created the custom service, then on a subsequent Agent start, firewalld is running and we need to delete the file.

azurelinuxagent/ga/env.py
```diff
@@ -99,85 +100,42 @@ def _get_dhcp_client_pid(self):
...
+                self._report(event.warn, "The firewall is not configured correctly. {0}. Will reset it. Current state:\n{1}", ustr(e), self._firewall_manager.get_state())
+                self._firewall_manager.remove()
+            self._firewall_manager.setup()
+            self._report(event.info, "The firewall was setup successfully:\n{0}", self._firewall_manager.get_state())
         except Exception as e:
-            return "Failed to get the firewall state: {0}".format(ustr(e))
+            self._report(event.warn, "An error occurred while setting up the firewall: {0}", ustr(e))
 
+    def _report(self, report_function, message, *args):
+        # Report the first 6 messages, then stop reporting for 12 hours
```
maddieford, line 128: Comment says 6 messages, but limit is 3 (line 133)

azurelinuxagent/ga/persist_firewall_rules.py
nagworld9: I understand the intention, my comment on organizing the code. I would prefer to place it before we setup the firewall rules with firewalld. That wat it looks cleaner

azurelinuxagent/ga/firewall_manager.py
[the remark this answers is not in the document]
narrieta: Not quite. Current code tries iptables, then nft, then raises if none present

```
        try:
            manager = IpTables(wire_server_address)
            logger.info("Using iptables to manage firewall rules")
            return manager
        except FirewallManagerNotAvailableError:
            pass

        try:
            manager = NfTables(wire_server_address)
            logger.info("Using nftables to manage firewall rules")
            return manager
        except FirewallManagerNotAvailableError:
            pass

        raise FirewallManagerNotAvailableError("Cannot create a firewall manager; no known command-line tools are available")

```

azurelinuxagent/ga/persist_firewall_rules.py
narrieta: I think I'm missing something here. Do you mean moving the delete before line 132? that makes no difference as far as code readability/organization, I think.

azurelinuxagent/ga/firewall_manager.py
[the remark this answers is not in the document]
nagworld9: Assume iptables and nftables both tools not present. First, we try iptables and got FirewallManagerNotAvailableError. Then we try nft but who is throwing  FirewallManagerNotAvailableError error when we initialize NfTables(wire_server_address). It just takes nftables as firewall manger and try to setup the rules, will get different exception during setup.

azurelinuxagent/ga/firewall_manager.py
```diff
@@ -0,0 +1,499 @@
...
+            use_wait_option = False
+
+        if use_wait_option:
+            self._base_command = ["iptables", "-w", "-t", "security"]
+        else:
+            self._base_command = ["iptables", "-t", "security"]
+
+    def _execute_delete_command(self, command):
+        """
+        Continually execute the delete operation until the return code is non-zero or the limit has been reached.
```
maddieford, line 298: Exit code 1 indicates successful deletion?

azurelinuxagent/ga/env.py
```diff
@@ -99,85 +100,42 @@ def _get_dhcp_client_pid(self):
...
+                self._firewall_manager = FirewallManager.create(self._wire_server_address)
+
+            try:
+                if self._firewall_manager.check():
+                    return  # The firewall is configured correctly
+                self._report(event.warn, "The firewall has not been setup. Will set it up.")
+            except FirewallStateError as e:
+                self._report(event.warn, "The firewall is not configured correctly. {0}. Will reset it. Current state:\n{1}", ustr(e), self._firewall_manager.get_state())
+                self._firewall_manager.remove()
+            self._firewall_manager.setup()
```
maddieford, line 122: We no longer remove legacy rules here?

azurelinuxagent/ga/firewall_manager.py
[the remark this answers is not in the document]
narrieta: Ah, I see what you mean. NfTables is missing its init method to do the check. I'll add it. Thanks.

maddieford, line 368: Could you explain the distinction between FirewallManager and FirewallManagerMultipleRules? i want to understand why they have separate implementations

azurelinuxagent/agent.py
narrieta: Makes sense, though I need the result of the match, so I do not think I can use an elif (unless I evaluate twice). Any suggestions?

azurelinuxagent/common/event.py
```diff
@@ -693,6 +691,30 @@ def add_event(name=AGENT_NAME, op=WALAEventOperation.Unknown, is_success=True, d
...
+    logger.info(fmt, *args)
+    add_event(op=op, message=fmt.format(*args), is_success=True)
+
+
+def warn(op, fmt, *args):
+    """
+    Creates a telemetry event and logs the message as WARNING.
+    """
+    logger.warn(fmt, *args)
+    add_event(op=op, message="[WARNING] " + fmt.format(*args), is_success=False, log_event=False)
```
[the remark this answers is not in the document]
narrieta: OperationSuccess in Kusto will be False both for warnings and errors. When checking telemetry, often people are concerned about OperationSuccess being False. I want to separate warning messages from actual errors.

azurelinuxagent/ga/env.py
narrieta: removing the legacy rules is done only once, at initialization

tests/ga/test_firewall_manager.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def test_remove_should_not_attempt_to_delete_rules_that_do_not_exist(self):
+        self._test_remove_should_not_attempt_to_delete_rules_that_do_not_exist(FirewallCmd, MockFirewallCmd)
+
+    def test_check_should_verify_all_rules(self):
+        self._test_check_should_verify_all_rules(FirewallCmd, MockFirewallCmd)
+
+    def test_remove_legacy_rule_should_delete_the_legacy_rule(self):
+        self._test_remove_legacy_rule_should_delete_the_legacy_rule(FirewallCmd, MockFirewallCmd)
+
```
nagworld9, line 152: do you plan to add test cases for ntftbales later?

azurelinuxagent/ga/firewall_manager.py
narrieta: Indicates some kind of error. See comment below - keeping existing implementation until there is time to look into it in more detail. (Note that the end-to-end tests work perfectly, hence this is not an immediate concern for now)

From the manual
```

The exit code is 0 for correct functioning.  Errors which appear to be caused by invalid or  abused  command line parameters cause an exit code of 2, and other errors cause an exit code of 1.
       
```

azurelinuxagent/ga/firewall_manager.py
narrieta: This is my attempt to explain that in the current comments. 

`    FirewallManager abstracts the interface for managing the firewall rules on the WireServer address. Concrete implementations provide the underlying functionality using command-line tools that vary across distros (e.g. iptables, firewalld, and nftables.)
`

FirewallManagerMultipleRules
`
    Base class for firewall managers that handle multiple rules, each rule being manipulated with a different command line (e.g. iptables, firewalld)
 `
 
iptables and firewallcmd maintains 3 separate rules, that need to be manipulated with 3 separate command variations

nft maintains 1 single rule, with 1 single command

azurelinuxagent/ga/firewall_manager.py
```diff
@@ -0,0 +1,499 @@
...
+    FirewallManager based on the nft command-line tool.
+    """
+    def setup(self):
+        shellutil.run_command(["nft", "add", "table", "ip", "walinuxagent"])
+        shellutil.run_command(["nft", "add", "chain", "ip", "walinuxagent", "output", "{", "type", "filter", "hook", "output", "priority", "0", ";", "policy", "accept", ";", "}"])
+        shellutil.run_command([
+            "nft", "add", "rule", "ip", "walinuxagent", "output", "ip", "daddr", self._wire_server_address,
+            "tcp", " dport", "!=", "53",
+            "skuid", "!=", str(os.getuid()),
+            "ct", "state", "invalid,new", "counter", "drop"])
```
nagworld9, line 379: I came across some interesting functionality for nfttables, I don't know it's worth in our code. Unlike iptables it provides atomicity when execute multiple rules at once.  

https://wiki.nftables.org/wiki-nftables/index.php/Atomic_rule_replacement

`With iptables it was common to use a bash script comprised of multiple iptables commands to configure a firewall. This is sub-optimal because it is not atomic, that is to say that during the few fractions of a second that your bash script takes to run your firewall is in a partially configured state. Nftables introduces atomic rule replacement with the -f option. This is different from bash scripts because nftables will read in all of the included config files, create the config object in memory alongside the existing config, and then in one atomic operation it swaps the old config for the new one meaning there is no moment when the firewall is partially configured`

narrieta: Yes, it is something we should take advantage of. Same thing with firewalld, we are not taking advantage of all its functionality. I posted this in our internal chat when I posted my PR. Will be a new iteration in this code.

azurelinuxagent/ga/persist_firewall_rules.py
```diff
@@ -101,67 +98,60 @@ def _is_firewall_service_running():
...
-        logger.info(
-            "Firewalld service not running/unavailable, trying to set up {0}".format(self._network_setup_service_name))
+        event.info(WALAEventOperation.PersistFirewallRules,"Firewalld.service is running, setting up permanent rules on the VM")
+        # In case of a failure, this would throw. In such a case, we don't need to try to setup our custom service
+        # because on system reboot, all iptables rules are reset by firewalld.service, so it would be a no-op.
+        # setup permanent firewalld rules
+        firewall_manager = FirewallCmd(self._dst_ip)
 
-        self._setup_network_setup_service()
+        firewall_manager.remove_legacy_rule()
```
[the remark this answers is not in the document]
narrieta: yes, good point, will add it

azurelinuxagent/agent.py
maddieford: I see, maybe do the match before the first if condition. Although that may defeat the intent of making this more readable. If moving it creates confusion, i'm ok with it as is

narrieta: Yes, moving it to the top of the if does not help in terms of readability. On the other hand, evaluating twice is not such a big deal in this case, so if it helps for readability in your opinion, I can add it to the next iteration of this code (I still need to add unit test for the env thread)

azurelinuxagent/ga/firewall_manager.py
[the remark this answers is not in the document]
narrieta: Thanks again for pointing this out. 

Both the firewall-cmd and nft implementations were not honoring the contract required by the base class:

`    If a concrete implementation cannot be instantiated because the underlying command-line tool is not available, it must raise a FirewallManagerNotAvailable exception.
` 

This is not an issue in the currently supported distros (and the end-to-end tests prove that). but it could be an issue in the future, and unit tests should have caught this, except there were no unit tests for this!

Issue has been fixed now, and unit tests added.

tests/ga/test_firewall_manager.py
narrieta: I had to add unit tests for the issue you mentioned above, wrt ntf not raising an exception when the command is not available. As part of the fix for that, I added all unit test I had planned for nftables.

Unit tests for the env thread coming soon in a separate PR

azurelinuxagent/ga/persist_firewall_rules.py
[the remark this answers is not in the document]
narrieta: added

azurelinuxagent/ga/firewall_manager.py
narrieta: (the way this work in practice is that, when the rule exists, the command succeeds and returns 0, then it is called again and fails with 1 (because the rule does not exist now) and that breaks the loop.)

azurelinuxagent/ga/firewall_manager.py
```diff
@@ -0,0 +1,521 @@
...
+            except CommandError as e:
+                if e.returncode == 1:  # rule does not exist
+                    missing_rules.append(rule)
+                else:
+                    raise
+
+        if len(missing_rules) == 0:  # all rules are present
+            return True
+
+        if len(existing_rules) > 0:  # some rules are present, but not all
```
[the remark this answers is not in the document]
narrieta: I want to distinguish between 3 different cases, mainly to log the proper message
* The firewall is fully setup (there are no missing rules, return True)
* The firewall is partially setup (there some missing rules, and some existing rules, throw)
* The firewall is not setup at all (there are no existing rules, return False)

The calling code handles these 3 conditions differently, e.g.

```
            try:
                    return  # The firewall is configured correctly
                self._report(event.warn, "The firewall has not been setup. Will set it up.")
            except FirewallStateError as e:
                self._report(event.warn, "The firewall is not configured correctly. {0}. Will reset it. Current state:\n{1}", ustr(e), self._firewall_manager.get_state())
```

nagworld9 (approved):

azurelinuxagent/ga/firewall_manager.py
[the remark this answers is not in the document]
maddieford: Makes sense

maddieford (approved):

PR #3223, opened 2024-09-19
move setupslice after cgroupsv2 check, remove unit file for log collector and remove fiirewall daemon-reload

nagworld9:
## Description

This pr addressing the walmart issue. Their systems hitting performance issues after agent update. As part of agent cgroup init, we use system daemon-load to update cgroup settings for agent service. Since it's global command, it's resetting the CPU affinity that that walmart Kubernetes uses. We observed this in ubunut22 specific containerD version.

Current fix targeting to walmart scenario

Changes are:

- Their systems(uv22) come with cgroupV2 and we don't support v2 yet, so we can stop creating unit files as part of cgroup init that will avoid daemon-reload. Move the setup slices after cgroupv2 check
- Remove the unit file creation for logcollector and that was the reason setup slice setup was added before cgroupsv2 check and can run independently without agent cgroupv2 support
-  Removed daemon-load in firewall service setup

Issue # 

---

### PR information


tests/lib/mock_command.py
```diff
@@ -2,12 +2,13 @@
...
 import sys
 
-if len(sys.argv) != 4:
+if len(sys.argv) < 4:
     sys.stderr.write("usage: {0} <stdout> <return_value> <stderr>".format(os.path.basename(__file__)))
 
 # W0632: Possible unbalanced tuple unpacking with sequence: left side has 3 label(s), right side has 0 value(s) (unbalanced-tuple-unpacking)
 # Disabled: Unpacking is balanced: there is a check for the length on line 5
-stdout, return_value, stderr = sys.argv[1:]  # pylint: disable=W0632
+# test adding extra arguments like pycharm helpers (/snap/pycharm-professional/412/plugins/python-ce/helpers/pydev/pydevd.py)
```
nagworld9, line 10: weird issue observed in my runs, it took me long time to figure out this. As far as mocking args it will be present at the end in the args list

narrieta: could you add more details in the comments about what the issue was? I am not sure I fully understand it.

azurelinuxagent/ga/persist_firewall_rules.py
```diff
@@ -199,8 +199,8 @@ def _setup_network_setup_service(self):
 
             # Create unit file with default values
             self.__set_service_unit_file()
-            # Reload systemd configurations when we setup the service for the first time to avoid systemctl warnings
-            self.__reload_systemd_conf()
+            # Initially systemd daemon-reload added here to prevent systemctl warnings after unit file changes. But that warning is local to that service and it doesn't affect other services
```
narrieta, line 202: Minor: I would rephrase the comment to say that after creating the service, systemctl may issue a warning when checking the service, and daemon-reload should not be used to clear the warning, since it can affect other services

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -191,19 +181,17 @@ def initialize(self):
                     log_cgroup_warning("The agent is within an unexpected slice: {0}".format(agent_slice))
                     return
 
+                if self.using_cgroup_v2():
```
narrieta, line 184: shouldn't this check be done as part of the "if cgroups supported" logic that starts at line 138?

narrieta: if we move it over there, careful with the reset at line 144, which also does a reload

maddieford: This v2 check needs to happen after we get the cgroup api instance on line 160. using_cgroup_v2() checks if self._cgroups_api is an instance of the v2 implementation 

I think this check can happen before we check which slice the agent is in

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -295,9 +283,8 @@ def __setup_azure_slice():
             if not os.path.exists(vmextensions_slice):
                 files_to_create.append((vmextensions_slice, _VMEXTENSIONS_SLICE_CONTENTS))
 
-            # Update log collector slice contents
-            slice_contents = _LOGCOLLECTOR_SLICE_CONTENTS_FMT.format(cpu_quota=LOGCOLLECTOR_CPU_QUOTA_FOR_V1_AND_V2)
-            files_to_create.append((logcollector_slice, slice_contents))
+            # New agent will setup limits for scope instead slice, so removing existing logcollector slice.
+            CGroupConfigurator._Impl.__cleanup_unit_file(logcollector_slice)
```
maddieford, line 287: Do we need to check if log collector slice exists or does __cleanup_unit_file do that check

azurelinuxagent/ga/cgroupconfigurator.py
nagworld9: we can't move to 144 line, we should check systemd system before we call cgroups api. Because those api calls based on systemd system. I think we can move to before agent slice but not 144

azurelinuxagent/ga/cgroupconfigurator.py
nagworld9: yes, __cleanup_unit_file already had that check

azurelinuxagent/ga/cgroupconfigurator.py
narrieta: the systemd check itself needs to be moved to the logic for "supported".

seems like we need some cleanup of the logic here. we can add a TODO comment here and address this on the next iteration

nagworld9: we can combine all in place but series of steps will be same as now

azurelinuxagent/ga/persist_firewall_rules.py
nagworld9: `after creating the service, systemctl may issue a warning when checking the service, `

This statement not true, when creating the service, it will not give warning, it's only issue a warning when we modify the service

`root@ub20euap:/lib/systemd/system# vi hello.service

root@ub20euap:/lib/systemd/system# systemctl status hello
● hello.service - Hello World Service
     Loaded: loaded (/lib/systemd/system/hello.service; static; vendor preset: enabled)
     Active: inactive (dead)
`

azurelinuxagent/ga/cgroupconfigurator.py
narrieta: the concern is not about the steps, but about the organization of the code. 

we do not support v2; we do not support systems on which systemd is not running; those conditions should be part of "is supported" logic

if you feel ok making those changes now, great. otherwise it can be marked to do later

narrieta: actually, this a hotfix for the release that is blocked. let's postpone those changes (but mark them as todo)

narrieta (approved):

maddieford (approved):

PR #3227, opened 2024-09-21
Add Azure Linux 3 FIPS image to daily tests

narrieta:


maddieford (commented):
Does this image have FIPS 140-2 enabled?

narrieta:
@maddieford FIPS 140-3, but OpenSSL uses a different provider (https://github.com/microsoft/SymCrypt-OpenSSL) that allows older cyphers, so Azure Linux 3 does not hit the same issues as Ubuntu/RHEL.

maddieford (approved):

nagworld9 (approved):

PR #3238, opened 2024-10-08
Fix test timing issue + minor code cleanup

narrieta:
The agent_firewall test may attempt to check the firewall before the Agent has set it up. For iptables, this was not an issue because the firewall is setup by the daemon, but that is not the case for nft. 

Added an extra check to verify that the walinuxagent table is already created in NfTables.get_missing_rules().

The Agent may still be in the middle of setting up the table when the test runs. My next PR will add atomic updated for nft, which should take care of that condition.

The PR also includes some minor cleanup that I pointed out as comments.

### DIFF
--- a/azurelinuxagent/ga/firewall_manager.py
+++ b/azurelinuxagent/ga/firewall_manager.py
@@ -291,6 +291,7 @@ class IpTables(_FirewallManagerMultipleRules):
             if isinstance(exception, OSError) and exception.errno == errno.ENOENT:  # pylint: disable=no-member
                 raise FirewallManagerNotAvailableError("iptables is not available")
             event.warn(WALAEventOperation.Firewall, "Unable to determine version of iptables; will not use -w option. --version output: {0}", ustr(exception))
+            self._version = "unknown"
             use_wait_option = False
 
         if use_wait_option:
--- a/azurelinuxagent/ga/firewall_manager.py
+++ b/azurelinuxagent/ga/firewall_manager.py
@@ -72,19 +72,26 @@ class FirewallManager(object):
         """
         try:
             manager = IpTables(wire_server_address)
-            logger.info("Using iptables to manage firewall rules")
+            event.info(WALAEventOperation.Firewall, "Using iptables [version {0}] to manage firewall rules", manager.version)
             return manager
         except FirewallManagerNotAvailableError:
             pass
 
         try:
             manager = NfTables(wire_server_address)
-            logger.info("Using nftables to manage firewall rules")
+            event.info(WALAEventOperation.Firewall, "Using nft [version {0}] to manage firewall rules", manager.version)
             return manager
         except FirewallManagerNotAvailableError:
             pass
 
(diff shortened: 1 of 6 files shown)


azurelinuxagent/ga/firewall_manager.py
```diff
@@ -72,20 +72,27 @@ def create(wire_server_address):
         """
         try:
             manager = IpTables(wire_server_address)
-            logger.info("Using iptables to manage firewall rules")
+            event.info(WALAEventOperation.Firewall, "Using iptables [version {0}] to manage firewall rules", manager.version)
```
narrieta, line 75: added telemetry to get the version of iptables/nft/firewall-cmd

tests/ga/test_env.py
```diff
@@ -169,7 +169,8 @@ def test_it_should_restore_missing_firewall_rules(self):
             test_cases = [  # Exit codes for the "-C" (check) command
                 {"accept_dns": 1, "accept": 0, "drop": 0, "legacy": 0},
                 {"accept_dns": 0, "accept": 1, "drop": 0, "legacy": 0},
-                {"accept_dns": 0, "accept": 0, "drop": 1, "legacy": 0},
+                {"accept_dns": 0, "accept": 1, "drop": 0, "legacy": 0},
+                {"accept_dns": 1, "accept": 1, "drop": 1, "legacy": 0},
```
narrieta, line 173: added test case for all rules missing

tests_e2e/tests/agent_firewall/agent_firewall.py
```diff
@@ -33,9 +32,7 @@ def __init__(self, context: AgentVmTestContext):
         self._ssh_client = self._context.create_ssh_client()
 
     def run(self):
-        log.info("Checking firewall rules added by the agent")
```
narrieta, line 36: Removed those messages; they are sort of confusing

tests_e2e/tests/lib/agent_log.py
```diff
@@ -369,18 +369,6 @@ def get_errors(self) -> List[AgentLogRecord]:
                 'if': lambda r: r.prefix == 'ExtHandler' and r.thread == 'ExtHandler'
             },
             #
-            # TODO: Currently Ubuntu minimal does not include the 'iptables' command. Remove this rule once this has been addressed.
```
narrieta, line 372: no longer needed

tests_e2e/tests/lib/firewall_manager.py
```diff
@@ -264,17 +264,19 @@ def get_state(self) -> str:
     }
 
     def get_missing_rules(self) -> List[str]:
-        missing = []
+        if "table ip walinuxagent" not in shellutil.run_command(["sudo", "nft", "list", "tables"]):
```
narrieta, line 267: add check for case when the table has not been created yet

nagworld9 (approved):

maddieford (approved):

Issue #143, opened 2016-01-03
Read VM location from WALinuxAgent


smereczynski:
There should be the possibility to read location where VM is running. It can be WALinuxAgent command or node in XML configuration parsed on deployment or incarnation. In many cases the user service's configuration is based on VM location so if VM clould read it's own location it will be very usefull.

yuezh:
That's reasonable requirement. We will work with service team to add that. @andyzhangx may provide you some workaround to get the location info from existing api.

andyzhangx:
the location info in VM is under file /var/lib/waagent/ExtensionsConfig*, e.g.
<Location>eastus2</Location>
You cloud use the following command line to get location info:
grep '<Location>' /var/lib/waagent/ExtensionsConfig* --no-filename | sed 's/<Location>//g' | sed 's/<\/Location>//g' | sed 's/ //g' | tr 'A-Z' 'a-z' | uniq

smereczynski:
Hah, I did not noticed that because of differences between ExtensionsConfig iterations. I have 4 iterations of this file:

ExtensionsConfig.1.xml
ExtensionsConfig.2.xml
ExtensionsConfig.3.xml
ExtensionsConfig.4.xml

and I was looking for something like <Location> in the fourth one. But this is just in the first one. Thank You for hint!

ahmetalpbalkan:
There are plans to launch an Instance Metadata API for Azure. It will give you an HTTP API to gather more information about the VM and the subscription it is running in. Please stay tuned.

Closing as the issue not about to the agent.

PR #3269, opened 2024-12-03
Cgroup rewrite: uses systemctl for expressing desired configuration instead drop-in files

nagworld9:
## Description

Current agents use drop-in file mechanism to set cgroup desired configuration. Creating drop-in files required daemon-reload run. This cmd can affect other services in the system like have seen for wlamart incident

The main changes in the PR:

- Removes the drop-in files, deamon-reload and uses `systemctl set-property --runtime` for resource control settings.https://www.man7.org/linux/man-pages/man1/systemctl.1.html
- New behavior: Raise the unexpected process if we see same process second time
- CPU throttle metric calculated by default instead explicit enabling
- In e2e test, added timeout for cmds/script runs over ssh

Added comments in the code and in the PR as well for other changes.

Issue # 

---

### PR information


azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -746,40 +773,43 @@ def _get_parent(pid):
             return 0
 
         def start_tracking_unit_cgroups(self, unit_name):
-            try:
-                cgroup = self._cgroups_api.get_unit_cgroup(unit_name, unit_name)
-                controllers = cgroup.get_controllers()
+            if self.enabled():
```
narrieta, line 776: why do we need the extra check?

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -811,141 +841,131 @@ def start_extension_command(self, extension_name, command, cmd_name, timeout, sh
...
-        def __reset_extension_services_cpu_quota(self, services_list):
+                        files_to_remove.append(drop_in_file_memory_accounting)
+
+                        drop_in_file_cpu_quota = os.path.join(drop_in_path, _DROP_IN_FILE_CPU_QUOTA)
+                        files_to_remove.append(drop_in_file_cpu_quota)
+                        self._cleanup_all_files(files_to_remove)
+
+                        cpu_quota = service.get('cpuQuotaPercentage')
+                        cpu_quota = "{0}%".format(cpu_quota) if cpu_quota is not None else ""  # setting an empty value resets to the default (infinity)
+                        properties_to_update = self._get_unit_properties_requiring_update(service_name, cpu_quota)
```
narrieta, line 940: why the extra check for the properties instead of just setting them up with no check?

tests_e2e/tests/lib/shell.py
```diff
@@ -46,9 +46,17 @@ def run_command(command: Any, shell=False) -> str:
     NOTE: The command's stdout and stderr are read as text streams.
     """
     process = Popen(command, stdout=PIPE, stderr=PIPE, shell=shell, text=True)
+    timer = threading.Timer(15 * 60, process.kill)  # Kill process after timeout
```
narrieta, line 49: this is applied to all commands, not only ssh

also, for ssh, we should terminate the remote process, not the local process

tests_e2e/tests/scripts/get-agent-name
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
```
narrieta, line 1: where is this script being invoked?

tests_e2e/tests/scripts/get-agent-name
```diff
@@ -0,0 +1,28 @@
...
+#
+
+#
+# Echos the agent name
+#
+
+set -euo pipefail
+
+PYTHON=$(get-agent-python)
+agent_name=$($PYTHON -c 'from azurelinuxagent.common.osutil import get_osutil; print(get_osutil().service_name)')
```
narrieta, line 27: script should be named "get-**service**-name" or similar

azurelinuxagent/common/osutil/systemd.py
```diff
@@ -84,3 +84,39 @@ def get_unit_property(unit_name, property_name):
...
+    Note: --runtime settings only apply until the next reboot
+    """
+    try:
+        # Ex: systemctl set-property foobar.service CPUWeight=200 --runtime
+        shellutil.run_command(["systemctl", "set-property", unit_name, "{0}={1}".format(property_name, value), "--runtime"])
+    except shellutil.CommandError as e:
+        raise ValueError("Can't set property {0} of {1}: {2}".format(property_name, unit_name, e))
+
+
+def set_unit_properties_run_time(unit_name, properties):
```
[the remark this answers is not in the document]
nagworld9: That's just gets complicated, we need to parse properties and match the corresponding values. Callers need to ensure they are sending the properties and values in the order.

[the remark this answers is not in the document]
narrieta: you can use a tuple

tests_e2e/tests/lib/shell.py
nagworld9: yes, I would think we even don't want the commands to stuck. 

is not kill child process which includes remote process? In my experiment processes in the vm also terminated.

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -432,74 +459,61 @@ def enable(self):
...
+        def _try_set_cpu_quota(unit_name, quota):  # pylint: disable=unused-private-member
             try:
-                drop_in_file = os.path.join(systemd.get_agent_drop_in_path(), _DROP_IN_FILE_CPU_QUOTA)
-                contents = _DROP_IN_FILE_CPU_QUOTA_CONTENTS_FORMAT.format(quota)
-                if os.path.exists(drop_in_file):
-                    with open(drop_in_file, "r") as file_:
-                        if file_.read() == contents:
-                            return True  # no need to update the file; return here to avoid doing a daemon-reload
-                CGroupConfigurator._Impl.__create_unit_file(drop_in_file, contents)
+                cpu_quota_per_sec_usec = systemd.get_unit_property(unit_name, "CPUQuotaPerSecUSec")
```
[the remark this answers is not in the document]
nagworld9: To add more context, this check was added for agent, extensions and extension services. 

My reasons are:

1. We have done same thing with drop-in files before
2. If there is no update, there is no point to touch the configuration again
3. Extensions and extensions services setup called on install and enable cmd. When we setup in install, it doesn't make sense to touch the config again in enable and also on extension install. The reason for adding it in enable is publishers dont need to publish their pkg for testing, they can update manifest in the backend and send the config change. At this point lot of them not boarded, they still need a way to test.

azurelinuxagent/common/osutil/systemd.py
```diff
@@ -84,3 +84,39 @@ def get_unit_property(unit_name, property_name):
...
+    try:
+        # Ex: systemctl set-property foobar.service CPUWeight=200 MemoryMax=2G IPAccounting=yes --runtime
+        shellutil.run_command(["systemctl", "set-property", unit_name] + properties + ["--runtime"])
+    except shellutil.CommandError as e:
+        raise ValueError("Can't set properties {0} of {1}: {2}".format(properties, unit_name, e))
+
+
+def is_unit_loaded(unit_name):
+    """
+    Determine if a unit is loaded
```
maddieford, line 116: What does it mean for a unit to be loaded

nagworld9: Unit being loaded means unit is parsed and available to the systemd process. We check this for extension services when we set quotas. If systemd is unaware of extension services and not loaded in the system yet, we get error while setting quotas. Hence, I added loaded check.

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -284,71 +293,73 @@ def __setup_azure_slice():
...
-            # reload the systemd configuration; the new slices will be used once the agent's service restarts
+        def _enable_accounting(unit_name):
+            """
+            Enable CPU and Memory accounting for the unit
+            """
             try:
-                log_cgroup_info("Executing systemctl daemon-reload...", send_event=False)
-                shellutil.run_command(["systemctl", "daemon-reload"])
+                # since we don't use daemon-reload and drop-files for accounting, so it will be enabled with systemctl set-property
+                accounting_properties = ["CPUAccounting=yes", "MemoryAccounting=yes"]
```
maddieford, line 354: What if only one controller is not mounted?  does the set_unit_properties_run_time cmd work for the other controller which is mounted?

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -128,52 +128,21 @@ def __init__(self):
...
-
-                # TODO: Move this and systemd system check to cgroups_supported logic above
-                if self.using_cgroup_v2():
-                    log_cgroup_info("Agent and extensions resource monitoring is not currently supported on cgroup v2")
+                    if not self.using_cgroup_v2():
+                        log_cgroup_info("Agent will reset the quotas in case cgroup usage went from enabled to disabled")
+                        self._reset_agent_cgroup_setup()
                     return
 
                 # We check the agent unit 'Slice' property before setting up azure.slice. This check is done first
```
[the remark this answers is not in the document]
nagworld9: yes, vm reboot

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -284,71 +293,73 @@ def __setup_azure_slice():
...
             else:
                 if not os.path.exists(agent_drop_in_file_slice):
                     files_to_create.append((agent_drop_in_file_slice, _AGENT_DROP_IN_FILE_SLICE_CONTENTS))
 
-            if fileutil.findre_in_file(agent_unit_file, r"CPUAccounting=") is not None:
-                CGroupConfigurator._Impl.__cleanup_unit_file(agent_drop_in_file_cpu_accounting)
-            else:
-                if not os.path.exists(agent_drop_in_file_cpu_accounting):
-                    files_to_create.append((agent_drop_in_file_cpu_accounting, _DROP_IN_FILE_CPU_ACCOUNTING_CONTENTS))
+            # New agent will setup limits for scope instead slice, so removing existing logcollector slice.
```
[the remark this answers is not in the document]
nagworld9: yes, logical slices still created based on slice name we provided

`The name consists of a dash-separated series of names, which describes the path to the slice from the root slice. The root slice is named -.slice. Example: foo-bar.slice is a slice that is located within foo.slice, which in turn is located in the root slice -.slice.`

azurelinuxagent/ga/cgroupconfigurator.py
nagworld9: good point, I haven't tested it. But we have done same thing with drop-in files. Regardless of controllers mounted we simply enabled controllers for CPU and Memory. We never seen an issue. Do you know how to unmount one controller and check this?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1337,11 +1337,10 @@ def set_extension_resource_limits(self):
         # setup the resource limits for extension operations and it's services.
         man = self.load_manifest()
         resource_limits = man.get_resource_limits()
-        if not CGroupConfigurator.get_instance().is_extension_resource_limits_setup_completed(extension_name,
```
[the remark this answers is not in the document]
nagworld9: Similar check was added in cgroupconfigurator setup_slice method. We don't touch the configuration again if limits already in place

azurelinuxagent/ga/cgroupconfigurator.py
maddieford: I've tried in the past, but haven't been successful because cgroup has to have no processes in order to unmount. I've seen telemetry from VMs which indicate some in the fleet only have one or the other mounted.

If we're unable to mount cpu or memory, then maybe we could try to add a unit property for a controller which isn't mounted by default on our test VM? And then see if the command returns any error codes. If not, I think it's safe to leave it as is in the PR.

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -128,52 +128,21 @@ def __init__(self):
...
-                    self._reset_agent_cgroup_setup()
-                    return
-                except CGroupsException as e:
-                    log_cgroup_warning("Unable to determine which cgroup version to use: {0}".format(ustr(e)), send_event=True)
-                    return
-
-                # TODO: Move this and systemd system check to cgroups_supported logic above
-                if self.using_cgroup_v2():
-                    log_cgroup_info("Agent and extensions resource monitoring is not currently supported on cgroup v2")
+                    if not self.using_cgroup_v2():
```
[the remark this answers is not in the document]
nagworld9: We did cleanup when we determine distro is not in supported list. At that point, I doubt we have enabled for v2 machines and later disabled. 

Now we combined several checks for cgroups not supported, so we may enter Line #143 due to other conditions as well. With that in mind, I added only v1. But no issue doing it for every vm. We only touch config if leftover files present, otherwise it's no-op

azurelinuxagent/common/osutil/systemd.py
narrieta: `If systemd is unaware of extension services and not loaded in the system yet, we get error while setting quotas. Hence, I added loaded check.`

can you add this as a comment where you call this function?

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -225,7 +195,45 @@ def initialize(self):
...
+                log_cgroup_warning("systemd was not detected on {0}".format(get_distro()), send_event=True)
+                return False
+
+            log_cgroup_info("systemd version: {0}".format(systemd.get_version()))
+
+            if not self._check_no_legacy_cgroups():
+                return False
+
+            try:
+                self._cgroups_api = get_cgroup_api()
```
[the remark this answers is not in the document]
narrieta: sorry, should have been explicit. the info we log in get_group_api

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -377,35 +388,51 @@ def __cleanup_all_files(files_to_cleanup):
...
-            return False
+        @staticmethod
+        def _calculate_current_cpu_quota(cpu_quota_per_sec_usec):
+            """
+            Calculate the CPU percentage from CPUQuotaPerSecUSec.
+            Params:
+                cpu_quota_per_sec_usec (str): The value of CPUQuotaPerSecUSec (e.g., "1s", "500ms", "500us", or "infinity").
+
+            Returns:
+                str: CPU percentage, or empty if 'infinity'.
```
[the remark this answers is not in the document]
narrieta: empty string is not a good choice. maybe we should match both to 'infinity' or some other value

azurelinuxagent/ga/cgroupconfigurator.py
nagworld9: I unmount the memory controller and set the MemoryAccounting=yes, seems no issue 

```
root@lisa--017-e0-n0:/sys/fs/cgroup/memory# findmnt -t cgroup --noheadings
/sys/fs/cgroup/systemd          cgroup cgroup rw,nosuid,nodev,noexec,relatime,xattr,name=systemd
/sys/fs/cgroup/cpu,cpuacct      cgroup cgroup rw,nosuid,nodev,noexec,relatime,cpu,cpuacct
/sys/fs/cgroup/devices          cgroup cgroup rw,nosuid,nodev,noexec,relatime,devices
/sys/fs/cgroup/net_cls,net_prio cgroup cgroup rw,nosuid,nodev,noexec,relatime,net_cls,net_prio
/sys/fs/cgroup/hugetlb          cgroup cgroup rw,nosuid,nodev,noexec,relatime,hugetlb
/sys/fs/cgroup/misc             cgroup cgroup rw,nosuid,nodev,noexec,relatime,misc
/sys/fs/cgroup/pids             cgroup cgroup rw,nosuid,nodev,noexec,relatime,pids
/sys/fs/cgroup/blkio            cgroup cgroup rw,nosuid,nodev,noexec,relatime,blkio
/sys/fs/cgroup/cpuset           cgroup cgroup rw,nosuid,nodev,noexec,relatime,cpuset
/sys/fs/cgroup/perf_event       cgroup cgroup rw,nosuid,nodev,noexec,relatime,perf_event
/sys/fs/cgroup/rdma             cgroup cgroup rw,nosuid,nodev,noexec,relatime,rdma
/sys/fs/cgroup/freezer          cgroup cgroup rw,nosuid,nodev,noexec,relatime,freezer
```

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -554,18 +577,24 @@ def check_cgroups(self, cgroup_metrics):
...
-            Raises a CGroupsException if the check fails
+            Raises a CGroupsException only when current unexpected process seen last time.
+
+            report_immediately - flag to switch to old behavior and report immediately if any unexpected process found.
+
+            Note: Process check was added as conservative approach before cgroups feature stable. Now it's producing noise due to race issues, some of those issues are extra process before systemd move to new cgroup or process about to die.
+            So now changing the behavior to raise an issue only when we see the same unexpected process on last check. Later we will remove the check if no issues reported.
             """
-            unexpected = []
+            current_unexpected = {}
```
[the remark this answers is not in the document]
nagworld9: Matching with process_id is more solid than format process (process name) when we validate the process is still in the agent cgroup.  Sometimes the name may be empty when we query for name

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -377,35 +397,51 @@ def __cleanup_all_files(files_to_cleanup):
...
-                                                              cpu_quota=cpu_quota)
-            if os.path.exists(extension_slice_path):
-                with open(extension_slice_path, "r") as file_:
-                    if file_.read() == slice_contents:
-                        return True
-            return False
+        @staticmethod
+        def _get_current_cpu_quota(unit_name):
+            """
+            Calculate the CPU percentage from CPUQuotaPerSecUSec for given unit.
```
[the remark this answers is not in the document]
nagworld9: yes, same. Fyi, CPUQuotaPerSecUSec is systemd property, I don't think cgroup version matter.

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -225,7 +195,45 @@ def initialize(self):
...
+            distro_supported = CGroupUtil.distro_supported()
+            if not distro_supported:
+                log_cgroup_info("Cgroup monitoring is not supported on {0}".format(get_distro()), send_event=True)
+                return False
+
+            if not systemd.is_systemd():
+                log_cgroup_warning("systemd was not detected on {0}".format(get_distro()), send_event=True)
+                return False
+
+            log_cgroup_info("systemd version: {0}".format(systemd.get_version()))
```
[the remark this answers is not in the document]
nagworld9: Today we already send systemd system or not in osinfo. Do we want to add new property systemd_version to that info? or separate log? I want to add to the osinfo but output seems too long. What do you think appropriate?

`systemd 239 (239-82.el8_10.2)
+PAM +AUDIT +SELINUX +IMA -APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 +SECCOMP +BLKID +ELFUTILS +KMOD +IDN2 -IDN +PCRE2 default-hierarchy=legacy`

```
            os_info_msg = u"Distro: {dist_name}-{dist_ver}; "\
                u"OSUtil: {util_name}; "\
                u"AgentService: {service_name}; "\
                u"Python: {py_major}.{py_minor}.{py_micro}; "\
                u"Arch: {vm_arch}; "\
                u"systemd: {systemd}; "\
                u"LISDrivers: {lis_ver}; "\
                u"logrotate: {has_logrotate};".format(
                    dist_name=DISTRO_NAME, dist_ver=DISTRO_VERSION,
                    util_name=type(self.osutil).__name__,
                    service_name=self.osutil.service_name,
                    py_major=PY_VERSION_MAJOR, py_minor=PY_VERSION_MINOR,
                    py_micro=PY_VERSION_MICRO, vm_arch=vm_arch, systemd=systemd.is_systemd(),
                    lis_ver=get_lis_version(), has_logrotate=has_logrotate()
```

tests_e2e/tests/lib/shell.py
narrieta: ah, ok, as long as the remote process is killed too, we are ok

tests_e2e/tests/lib/shell.py
```diff
@@ -46,9 +46,17 @@ def run_command(command: Any, shell=False) -> str:
...
 
-    stdout, stderr = process.communicate()
+    try:
+        timer.start()
```
narrieta, line 52: start() should probably be outside the try statement (if it throws, I think we do not want to call cancel())

narrieta (approved):
LGTM, approved with one minor comment on shell.py (e2e tests)

narrieta (dismissed):

maddieford (approved):

narrieta (approved):

PR #3272, opened 2024-12-06
Pickup latest goal state on tenant certificate rotation + Avoid infinite loop when the tenant certificate is missing

narrieta:
Fixes for a couple of bugs on the goal state that surfaced on recent incidents.

First, if there is a problem with the tenant certificate, UpdateHandler._initialize_goal_state() can go into an infinite loop. Second, during rotation of the tenant certificate, the Agent needs to pick up the latest extensions goal state.

For both issues, comments in the code provide more details.

The PR also includes a fix for a bug logging in UpdateHandler._try_update_goal_state().


azurelinuxagent/ga/update.py
```diff
@@ -491,45 +496,93 @@ def _check_threads_running(self, all_thread_handlers):
...
-                message = u"Error fetching the goal state: {0}".format(textutil.format_exception(e))
-                logger.error(message)
-                add_event(op=WALAEventOperation.FetchGoalState, is_success=False, message=message, log_event=False)
-                self._update_goal_state_last_error_report = datetime.now()
+                # Report up to 'max_errors_to_log' immediately
+                self._update_goal_state_next_error_report = datetime.now()
+                event.error(WALAEventOperation.FetchGoalState, "Error fetching the goal state: {0}", textutil.format_exception(e))
             else:
-                if self._update_goal_state_last_error_report + timedelta(hours=6) > datetime.now():
-                    self._update_goal_state_last_error_report = datetime.now()
```
narrieta, line 527: while working on these fixes, I noticed that these messages were being logged every 6 secs, rather than every 6 hours as was intended.

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -61,14 +61,6 @@ class GoalStateProperties(object):
     All = RoleConfig | HostingEnv | SharedConfig | ExtensionsGoalState | Certificates | RemoteAccessInfo
 
 
-class GoalStateInconsistentError(ProtocolError):
```
narrieta, line 64: Instead of checking the tenant cert at the GoalState level and raising an exception when there is a mismatch, now the callers of GoalState that are actually interested in this situation do the check and handle the error. The caller, in this case the main loop of the extension handler, can do a better job handling and reporting the error.

tests/common/protocol/test_goal_state.py
```diff
@@ -381,20 +380,6 @@ def http_get_handler(url, *_, **__):
             self.assertEqual(initial_timestamp, goal_state.extensions_goal_state.created_on_timestamp, "The timestamp of the updated goal state is incorrect")
             self.assertTrue(goal_state.extensions_goal_state.is_outdated, "The updated goal state should be marked as outdated")
 
-    def test_it_should_raise_when_the_tenant_certificate_is_missing(self):
```
narrieta, line 384: not needed anymore

tests/common/protocol/test_goal_state.py
```diff
@@ -444,45 +429,6 @@ def test_it_should_download_certs_on_a_new_fabric_goal_state(self):
             goal_state.update()
             self.assertTrue(os.path.isfile(crt_path))
 
-    def test_it_should_refresh_the_goal_state_when_it_is_inconsistent(self):
```
narrieta, line 447: moved to test_update

tests/common/protocol/test_goal_state.py
```diff
@@ -506,39 +452,6 @@ def http_get_handler(url, *_, **__):
             self.assertEqual(0, len(goal_state.certs.summary), "Cert list should be empty")
             self.assertEqual(1, http_get_handler.certificate_requests, "There should have been exactly 1 requests for the goal state certificates")
 
-    def test_it_should_refresh_the_goal_state_when_it_is_fails_to_decrypt_cert(self):
```
narrieta, line 509: this test was not needed to begin with

PR #3273, opened 2024-12-09
Pickup latest goal state on tenant certificate rotation + Avoid infin…

narrieta:
Fixes for a couple of bugs on the goal state that surfaced on recent incidents.

First, if there is a problem with the tenant certificate, UpdateHandler._initialize_goal_state() can go into an infinite loop. Second, during rotation of the tenant certificate, the Agent needs to pick up the latest extensions goal state.

For both issues, comments in the code provide more details.

The PR also includes a fix for a bug logging in UpdateHandler._try_update_goal_state().


azurelinuxagent/ga/update.py
```diff
@@ -491,45 +496,93 @@ def _check_threads_running(self, all_thread_handlers):
...
+                # Report up to 'max_errors_to_log' immediately
+                self._update_goal_state_next_error_report = datetime.now()
+                event.error(WALAEventOperation.FetchGoalState, "Error fetching the goal state: {0}", textutil.format_exception(e))
             else:
-                if self._update_goal_state_last_error_report + timedelta(hours=6) > datetime.now():
-                    self._update_goal_state_last_error_report = datetime.now()
-                    message = u"Fetching the goal state is still failing: {0}".format(textutil.format_exception(e))
-                    logger.error(message)
-                    add_event(op=WALAEventOperation.FetchGoalState, is_success=False, message=message, log_event=False)
+                # Report one single periodic error every 6 hours
```
narrieta, line 566: while working on these fixes, I noticed that these messages were being logged every 6 secs, rather than every 6 hours as was intended.

maddieford (approved):

azurelinuxagent/ga/update.py
```diff
@@ -491,45 +496,93 @@ def _check_threads_running(self, all_thread_handlers):
                 logger.warn("{0} thread died, restarting".format(thread_handler.get_thread_name()))
                 thread_handler.start()
 
+    errors = 0
+
     def _try_update_goal_state(self, protocol):
```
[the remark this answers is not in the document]
narrieta: Thanks; with this change the elif is not needed (in fact, it is dead code. Removing it.

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -282,8 +260,13 @@ def _update(self, force_update):
...
-        if goal_state_updated and vm_settings_updated:
-            most_recent = vm_settings if vm_settings.created_on_timestamp > extensions_config.created_on_timestamp else extensions_config
+        if goal_state_updated:
+            # On rotation of the tenant certificate the vmSettings and extensionsConfig are not updated. However, the incarnation of the WS goal state is update so 'goal_state_updated' will be True.
+            # In this case, we should use the most recent of vmSettigns and extensionsConfig.
+            if vm_settings is not None:
+                most_recent = vm_settings if vm_settings.created_on_timestamp > extensions_config.created_on_timestamp else extensions_config
+            else:
+                most_recent = extensions_config
         elif goal_state_updated:
```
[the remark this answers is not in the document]
narrieta: Thanks; with this change the elif is not needed (in fact, it is dead code. Removing it.

azurelinuxagent/ga/update.py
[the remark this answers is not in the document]
narrieta: Yes, _initialize_goal_state also calls this method

azurelinuxagent/ga/update.py
```diff
@@ -491,45 +496,93 @@ def _check_threads_running(self, all_thread_handlers):
                 logger.warn("{0} thread died, restarting".format(thread_handler.get_thread_name()))
                 thread_handler.start()
 
+    errors = 0
```
nagworld9, line 499: where is this used?

azurelinuxagent/ga/update.py
```diff
@@ -491,45 +496,93 @@ def _check_threads_running(self, all_thread_handlers):
...
+
+                if not new_goal_state or goal_state.extensions_goal_state.source != GoalStateSource.FastTrack:
+                    break
 
-            protocol.client.update_goal_state(silent=self._update_goal_state_error_count >= max_errors_to_log, save_to_history=True)
+                if self._check_certificates(goal_state):
+                    if attempt > 0:
+                        event.info(WALAEventOperation.FetchGoalState, "The extensions goal state is now in sync with the tenant cert.")
+                    break
+
```
nagworld9, line 533: is it worth to add log "refreshing goal state due to tenant cert out of sync" something like that on every retry. We will know why we did retry, otherwise it just has series of fetch logs.

azurelinuxagent/ga/update.py
narrieta: thanks; left-over debugging code. Removing it

azurelinuxagent/ga/update.py
narrieta: Thanks; added

azurelinuxagent/ga/update.py
```diff
@@ -491,45 +496,94 @@ def _check_threads_running(self, all_thread_handlers):
...
+            # a delay (to address situations in which the HGAP and the WireServer are temporarily out of sync).
+            #
+            for attempt in range(3):
+                protocol.client.update_goal_state(force_update=attempt > 0, silent=self._update_goal_state_error_count >= max_errors_to_log, save_to_history=True)
+
+                goal_state = protocol.get_goal_state()
+                new_goal_state = self._goal_state is None or self._goal_state.extensions_goal_state.id != goal_state.extensions_goal_state.id
+
+                if not new_goal_state or goal_state.extensions_goal_state.source != GoalStateSource.FastTrack:
+                    break
```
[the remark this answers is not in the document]
narrieta: `Current code stops at retry 1, since previous fetch and current fetch same incarnation
`
This won't happen, since we update self._goal_state is updated only after all retries have been done

azurelinuxagent/ga/update.py
narrieta: thanks; updated

azurelinuxagent/ga/update.py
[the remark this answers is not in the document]
nagworld9: Oh, Github PR UI deceived me :). self._goal_state is outside of loop, thanks for confirming, All good then

maddieford (approved):

PR #3274, opened 2024-12-09
 Suppress known error in GuestConfiguration

narrieta:


tests_e2e/tests/lib/agent_log.py
```diff
@@ -103,50 +103,6 @@ def get_errors(self) -> List[AgentLogRecord]:
         #     * 'if' - A lambda that takes as parameter an AgentLogRecord representing an error and returns true if the error should be ignored
         #
         ignore_rules = [
-            #
```
narrieta, line 106: I copied these temporarily during the transition from DCR; no longer needed.

nagworld9 (approved):

PR #3284, opened 2025-01-06
Create Azure Pipeline for Python 2.6 & 3.4 Unit Tests

narrieta:
We use an Ubuntu 16 container to run the unit tests using Python 2.6 (the tests for 3.4 were also added to that container), but Github Actions does not support this version of Ubuntu anymore.

Create a Pipeline within our DevOps project to execute those tests.

### DIFF
--- a/azurelinuxagent/ga/cgroupconfigurator.py
+++ b/azurelinuxagent/ga/cgroupconfigurator.py
@@ -437,7 +437,7 @@ class CGroupConfigurator(object):
 
                 # Calculate CPU percentage
                 cpu_percentage = (cpu_quota_us / 1000000) * 100
-                return "{:g}%".format(cpu_percentage)  # :g Removes trailing zeros after decimal point
+                return "{0:g}%".format(cpu_percentage)  # :g Removes trailing zeros after decimal point
             except Exception as e:
                 log_cgroup_warning("Error parsing current CPUQuotaPerSecUSec: {0}".format(ustr(e)))
                 return "unknown"
--- a/tests/data/ext/sample_ext-1.3.0/python.sh
+++ b/tests/data/ext/sample_ext-1.3.0/python.sh
@@ -4,7 +4,7 @@
 #
 python=$(command -v python 2> /dev/null)
 
-if [ -z "$PYTHON" ]; then
+if [ -z "$python" ]; then
   python=$(command -v python3)
 fi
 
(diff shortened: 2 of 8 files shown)


azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -437,7 +437,7 @@ def _get_current_cpu_quota(unit_name):
 
                 # Calculate CPU percentage
                 cpu_percentage = (cpu_quota_us / 1000000) * 100
-                return "{:g}%".format(cpu_percentage)  # :g Removes trailing zeros after decimal point
```
narrieta, line 440: This syntax is not supported on 2.6

tests/data/ext/sample_ext-1.3.0/python.sh
```diff
@@ -4,7 +4,7 @@
 #
 python=$(command -v python 2> /dev/null)
 
-if [ -z "$PYTHON" ]; then
+if [ -z "$python" ]; then
```
narrieta, line 7: fixing typo

tests/ga/test_cgroupapi.py
```diff
@@ -301,7 +301,7 @@ def mock_popen(command, *args, **kwargs):
                         shell=True,
                         timeout=300,
                         cwd=self.tmp_dir,
-                        env={},
+                        env={}.update(os.environ),
```
narrieta, line 304: These tests were failing on the new runs for 3.4. The reason is that they have a dependency on the "python3" executable;  on Github actions we were installing Python 3 even on runs that were using our 3.4 venv, so there was a "python3" in the system PATH and the tests were able to run. Now, we no longer install Python 3 for the 3.4 test runs, and "python3" is only in the PATH for the venv. These tests need to inherit the environment from the test runner (nosetests) in order to have the correct PATH. The Agent does something similar (adding os.environ) when invoking extensions.

tests/ga/test_cgroupconfigurator.py
```diff
@@ -868,7 +868,7 @@ def get_completed_process():
                 return completed
 
             agent_processes = [os.getppid(), os.getpid()] + agent_command_processes + [start_extension.systemd_run_pid]
-            other_processes = [1, get_completed_process()] + extension_processes
```
narrieta, line 871: When starting a process in a container, it gets PID 1. Github Actions starts containers using a dummy process ("tail -f /dev/null", which gets PID 1) and then uses "docker exec" to execute the actual tests. 

In the new pipeline, we invoke "nosetests" directly, and that gets PID 1. 

As a result, the Agent code identifies nosetests (PID 1) as the Daemon and it does not include it in the list of unexpected processes, causing this test to fail. "1" does not add anything significant to this test, so I just removed it.

PR #2729, opened 2023-01-17
Fixes #2728 Change rhel service file to not use python

Klaas-:
## Description

Issue #2782

---

### PR information
- [X] The title of the PR is clear and informative.
- [X] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).
- [X] If applicable, the PR references the bug/issue that it fixes in the description.
- [ ] New Unit tests were added for the changes made

### Quality of Code and Contribution Guidelines
- [x] I have read the [contribution guidelines](https://github.com/Azure/WALinuxAgent/blob/master/.github/CONTRIBUTING.md).

Issue #2728, opened 2023-01-17
[BUG][RHEL8] systemd service should not use python3


Klaas-:
**Describe the bug: A clear and concise description of what the bug is.**
Hi,
it seems on systems where python3 is not pointing to the same python version as platform-python waagent fails to start. This happens because you try to directly use python3 in the systemd service file.

1. Section of the log where the error occurs.
```
Traceback (most recent call last):
  File "/usr/sbin/waagent", line 35, in <module>
    import azurelinuxagent.agent as agent
ModuleNotFoundError: No module named 'azurelinuxagent'
```
2. Steps to reproduce the behavior.
I'd guess this happens on any rhel8 system where /usr/bin/python3 is not pointing to python 3.6 (platform-python)

**Distro and WALinuxAgent details (please complete the following information):**
 - Distro and Version: Red Hat Enterprise Linux release 8.7 (Ootpa)
 - WALinuxAgent version 
```
WALinuxAgent-2.7.0.6 running on redhat 8.7
Python: 3.6.8
Goal state agent: 2.8.0.11
```

```
$ python3 --version
Python 3.9.13
```

**Additional context**
Workaround: Create an override for the waagent.service file, something like this:
```
# /etc/systemd/system/waagent.service.d/override.conf
[Service]
ExecStart=
ExecStart=/usr/libexec/platform-python -u /usr/sbin/waagent -daemon
```

This happens because WALinuxAgent uses platform-python (3.6 on RHEL8).
```
head -n1 /usr/sbin/waagent
#!/usr/libexec/platform-python
```

I am guessing you could also change it to start /usr/sbin/waagent directly, without python, then it will work as well because /usr/sbin/waagent has the correct shebang and is executable

narrieta:
@Klaas- Installation of Python and the setup in the service file are done by the publisher of the image. If this is the standard image coming published by Red Hat, could you report the issue on their site? At some point they were using the platform Python, but there were some issues and they changed that after 1 release (AFAIK).

Klaas-:
@narrieta yeah, but it's your file that is creating the issues and easily fixed by my PR #2729 - then it'll trickle down into RHEL, I'll create a backport request once it's merged here. About which python version rhel uses, the beauty is I don't care :D Look at the PR, there is no need to specify which python in the service file, as long as the shebang is correct that service file will work.

Klaas-:
@narrieta any progress here? :)

narrieta:
[...]

PR #3286, opened 2025-01-06
Drop the interpreter for execution in the service file

rjschwei:
## Description

The waagent script is executable and has the interpreter set. Drop the use of the Python interpreter to execute the file from the systemd unit file. We achieve unbuffered I/O by setting the PYTHONUNBUFFERED environment variable instead. This makes it easier for distribution packaging as packagers will only have to update the pathto the interpreter used on the given distro in one file.

Issue # 

---

### PR information
- [ x] Ensure development PR is based on the `develop` branch.
- [x ] The title of the PR is clear and informative.

### DIFF
--- a/init/waagent.service
+++ b/init/waagent.service
@@ -8,7 +8,8 @@ ConditionPathExists=/etc/waagent.conf
 
 [Service]
 Type=simple
-ExecStart=/usr/bin/python -u /usr/sbin/waagent -daemon
+Environment="PYTHONUNBUFFERED=true"
+ExecStart=/usr/sbin/waagent -daemon
 Restart=always
 RestartSec=5
 


Klaas-:
I suggested drop the python execution a couple of years ago, MSFT were not interested :D
PR: https://github.com/Azure/WALinuxAgent/pull/2729
Original issue: https://github.com/Azure/WALinuxAgent/issues/2728

narrieta:
@rjschwei - Extensions inherit the environment of the Agent. This change ended up breaking the Microsoft.CPlat.Core.LinuxPatchExtension extension since it changes the timing of the extension. Although this is actually a bug in the extension and a fix is currently being deployed, we do not know if other extensions have similar issues, plus not all VMs may be configured to take extension updates. 

The PR makes a change that applies to distros other than SUSE, so we consider it to be high risk. Probably it may be worth considering whether this change can be undone in SUSE, but if it cannot, we suggest updating this PR to apply only to SUSE.

PR #3300, opened 2025-01-15
Cleanup Github Actions workflow for Unit Tests

narrieta:
The venvs for 2.6 & 3.4 can run directly on the VM (instead of running within a container), just as the venv for 2.7 already does.

This makes the test run a little faster, since there is no need to download the container image, but, more importantly, it allows for a nice cleanup of the YAML file, which too fragemented and hard to understand.

### DIFF
--- a/.github/workflows/ci_pr.yml
+++ b/.github/workflows/ci_pr.yml
@@ -8,92 +8,23 @@ on:
   workflow_dispatch:
 
 jobs:
-  test-python-2_6-and-3_4-versions:
-
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - python-version: "2.6"
-          - python-version: "3.4"
-
+  execute-tests:
     name: "Python ${{ matrix.python-version }} Unit Tests"
     runs-on: ubuntu-20.04
-    container:
-      image: ubuntu:24.04
-      volumes:
-        - /home/waagent:/home/waagent
-    defaults:
-      run:
-        shell: bash -l {0}
-
-    env:
-      NOSEOPTS: "--verbose"
-
-    steps:
-    - uses: actions/checkout@v3
-
-    - name: Install Python ${{ matrix.python-version }} Virtual Environment
-      run: |
-        apt-get update
-        apt-get install -y curl bzip2 sudo
-        curl -sSf --retry 5 -o /tmp/python-${{ matrix.python-version }}.tar.bz2 https://dcrdata.blob.core.windows.net/python/python-${{ matrix.python-version }}.tar.bz2
-        sudo tar xjf /tmp/python-${{ matrix.python-version }}.tar.bz2 --directory /
-        #
-        # TODO: Some unit tests create helper scripts that use 'python3' as shebang; we should probably port them to Bash, but installing Python 3 as a workaround for now.
-        #
-        if [[ "${{ matrix.python-version }}" == "2.6" ]]; then
-            apt-get -y install python3
-        fi
-        #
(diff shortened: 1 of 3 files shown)


.github/workflows/ci_pr.yml
```diff
@@ -106,33 +37,69 @@ jobs:
...
+    - name: Setup Python ${{ matrix.python-version }} Virtual Environment
+      if: matrix.use_virtual_environment
+      id: install-venv
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y curl bzip2 sudo
+        curl -sSf --retry 5 -o /tmp/python-${{ matrix.python-version }}.tar.bz2 https://dcrdata.blob.core.windows.net/python/python-${{ matrix.python-version }}.tar.bz2
+        sudo tar xjf /tmp/python-${{ matrix.python-version }}.tar.bz2 --directory /
+        #
+        # The virtual environments for 2.6 and 3.4 have dependencies on OpenSSL 1.0, which is not available beyond Ubuntu 16. We use this script to patch the environments.
```
narrieta, line 73: Interestingly, the venv for 2.7 does not require patching when running directly on the VM (it does if one uses a container). The venvs for 2.6 and 3.4 require patching either when running directly on the VM or within a container.

I'll dig into this later to see if can simplify the setup further.

.github/workflows/ci_pr.yml
```diff
@@ -8,92 +8,23 @@ on:
...
-      matrix:
-        include:
-          - python-version: "2.6"
-          - python-version: "3.4"
-
+  execute-tests:
     name: "Python ${{ matrix.python-version }} Unit Tests"
     runs-on: ubuntu-20.04
-    container:
-      image: ubuntu:24.04
```
narrieta, line 23: The python-setup action breaks when the VM is Ubuntu > 20. At some point, we will need to look into that.

.github/workflows/ci_pr.yml
```diff
@@ -106,33 +37,69 @@ jobs:
...
-    - name: Checkout WALinuxAgent repo
+    - name: Checkout WALinuxAgent
       uses: actions/checkout@v3
-
+    #
+    # We either install Python and the test dependencies, or download a pre-built virtual environment, depending on the
+    # use_virtual_environment flag.
+    #
     - name: Setup Python ${{ matrix.python-version }}
+      if: (!matrix.use_virtual_environment)
```
narrieta, line 48: the workflow parser has this funny behaviour that seems to require parenthesis in order to use the ! operator

tests/python_eol/patch_python_venv.sh
```diff
@@ -18,15 +18,20 @@
 #
 set -euo pipefail
 
-if [[ "$#" -ne 1 || ! "$1" =~ ^2\.6|3\.4$ ]]; then
-  echo "Usage: patch_python_venv.sh 2.6|3.4"
+if [[ "$#" -ne 1 || ! "$1" =~ ^2\.6|2\.7|3\.4$ ]]; then
+  echo "Usage: patch_python_venv.sh 2.6|2.7|3.4"
```
maddieford, line 22: 2.7 left here for container patching?

maddieford (approved):

tests/python_eol/patch_python_venv.sh
narrieta: yes, for using the 2.7 venv in a container... more patching may be needed, too (I'm getting some warnings about hashlib, even though we do not use it) and 2 test failures (not sure if the failures are related to those warnings, i'll check that later)

nagworld9 (approved):

PR #3305, opened 2025-01-21
Fix unit test failures when run on containers

narrieta:
TestOSUtil.test_get_nic_state and CGroupConfiguratorSystemdTestCase.test_start_extension_command_should_capture_only_the_last_subprocess_output were failing when run on our container images. This PR fixes them.

### DIFF
--- a/azurelinuxagent/common/osutil/default.py
+++ b/azurelinuxagent/common/osutil/default.py
@@ -82,7 +82,7 @@ IOCTL_SIOCGIFFLAGS = 0x8913
 IOCTL_SIOCGIFHWADDR = 0x8927
 IFNAMSIZ = 16
 
-IP_COMMAND_OUTPUT = re.compile(r'^\d+:\s+(\w+):\s+(.*)$')
+IP_COMMAND_OUTPUT = re.compile(r'^\d+:\s+([\w@]+):\s+(.*)$')
 
 STORAGE_DEVICE_PATH = '/sys/bus/vmbus/devices/'
 GEN2_DEVICE_ID = 'f8b3781a-1e82-4818-a1c3-63d806ec15bb'
--- a/tests/ga/test_cgroupconfigurator.py
+++ b/tests/ga/test_cgroupconfigurator.py
@@ -542,39 +542,37 @@ class CGroupConfiguratorSystemdTestCase(AgentTestCase):
     @patch('time.sleep', side_effect=lambda _: mock_sleep())
     def test_start_extension_command_should_capture_only_the_last_subprocess_output(self, _):
         with self._get_cgroup_configurator() as configurator:
-            pass  # release the mocks used to create the test CGroupConfigurator so that they do not conflict the mock Popen below
+            original_popen = subprocess.Popen
 
-        original_popen = subprocess.Popen
+            def mock_popen(command, *args, **kwargs):
+                # Inject a syntax error to the call
 
-        def mock_popen(command, *args, **kwargs):
-            # Inject a syntax error to the call
-            
-            # Popen can accept both strings and lists, handle both here.
-            if isinstance(command, str):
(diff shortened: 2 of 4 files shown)


azurelinuxagent/common/osutil/default.py
```diff
@@ -82,7 +82,7 @@ def crypt(password, salt):
 IOCTL_SIOCGIFHWADDR = 0x8927
 IFNAMSIZ = 16
 
-IP_COMMAND_OUTPUT = re.compile(r'^\d+:\s+(\w+):\s+(.*)$')
+IP_COMMAND_OUTPUT = re.compile(r'^\d+:\s+([\w@]+):\s+(.*)$')
```
narrieta, line 85: the NIC name can also include '@'

tests/ga/test_cgroupconfigurator.py
```diff
@@ -542,39 +542,37 @@ def test_start_extension_command_should_disable_cgroups_and_invoke_the_command_d
     @patch('time.sleep', side_effect=lambda _: mock_sleep())
     def test_start_extension_command_should_capture_only_the_last_subprocess_output(self, _):
         with self._get_cgroup_configurator() as configurator:
-            pass  # release the mocks used to create the test CGroupConfigurator so that they do not conflict the mock Popen below
+            original_popen = subprocess.Popen
```
narrieta, line 545: Recent changes in the cgroup code added new calls to systemctl to set quotas; those calls are already handled by the mocks created in self._get_cgroup_configurator, but previously those mocks were not needed and we were releasing them. Now we use them.

tests/ga/test_cgroupconfigurator.py
```diff
@@ -542,39 +542,37 @@ def test_start_extension_command_should_disable_cgroups_and_invoke_the_command_d
...
-            # Inject a syntax error to the call
-            
-            # Popen can accept both strings and lists, handle both here.
-            if isinstance(command, str):
-                command = command.replace('systemd-run', 'systemd-run syntax_error')
-            elif isinstance(command, list) and command[0] == 'systemd-run':
-                command = ['systemd-run', 'syntax_error'] + command[1:]
+                # Popen can accept both strings and lists, handle both here.
+                if isinstance(command, str) and command.startswith('systemd-run'):
+                    command = 'systemd-run syntax_error'
```
narrieta, line 552: We don't want this failure to be handled by the mocks created in self._get_cgroup_configurator, so we change the command completely so that the regexes in the mocks do not match it.

tests/python_eol/Dockerfile
```diff
@@ -31,7 +31,7 @@ RUN <<..
     # Install the Python venv
     #
     apt-get update
-    apt-get -y install curl bzip2 sudo
+    apt-get -y install curl bzip2 sudo iproute2
```
narrieta, line 34: Needed to get the 'ip' command (to get the NIC state)

tests/python_eol/patch_python_venv.sh
```diff
@@ -26,12 +26,7 @@ fi
 PYTHON_VERSION=$1
 
 if [[ "${PYTHON_VERSION}" =~ ^2\.6|2\.7$ ]]; then
-  if [[ "${PYTHON_VERSION}" == "2.6" ]]; then
-    file_to_patch="/opt/python/2.6.9/lib/python2.6/httplib.py"
-  else
-    file_to_patch="/opt/python/2.7.16/lib/python2.7/httplib.py"
-  fi
-  cat >> "$file_to_patch" << ...
+  cat >> /opt/python/${PYTHON_VERSION}/lib/python${PYTHON_VERSION}/httplib.py << ...
```
narrieta, line 29: I fixed some broken symbolic links on the venvs and now this is a little simpler

nagworld9 (approved):

maddieford (approved):

PR #3316, opened 2025-02-06
Enable resource monitoring in cgroup v2 machines

nagworld9:
## Description

This PR just enable the resource monitoring in v2(> ubuntu 22) machines. 

Couple of distinctions between V1 and V2 
1) By default, CPU throttle counter not present in v2, it's only available when we use quotas. So, now throttle time tracked if quota set.
2) Cgroup.procs file is not present in the slice cgroup, only in leaf cgroup in v2, as opposed to v1.  This file used to determine if cgroup is active or not. In the case of extension, we track metrics at the slice level but processes in scope cgroup. Now we check in scope too.

Issue # 

---

### PR information


azurelinuxagent/ga/cgroupapi.py
```diff
@@ -389,73 +502,8 @@ def log_root_paths(self):
             else:
                 log_cgroup_info("The {0} controller is mounted at {1}".format(controller, mount_point))
 
-    def start_extension_command(self, extension_name, command, cmd_name, timeout, shell, cwd, env, stdout, stderr,
```
nagworld9, line 392: Moved to abstract class

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -407,41 +423,6 @@ def _create_all_files(files_to_create):
                     CGroupConfigurator._Impl._cleanup_unit_file(unit_file)
                 return
 
-        @staticmethod
-        def _get_current_cpu_quota(unit_name):
```
nagworld9, line 411: Moved to util class

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -250,12 +294,81 @@ def log_root_paths(self):
         """
         raise NotImplementedError()
 
-    def start_extension_command(self, extension_name, command, cmd_name, timeout, shell, cwd, env, stdout, stderr,
-                                error_code=ExtensionErrorCodes.PluginUnknownFailure):
+    def get_enabled_controllers_at_root(self):
```
narrieta, line 297: raise NotImplementedError()?

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -389,73 +502,8 @@ def log_root_paths(self):
...
-                # no-member: Instance of 'ExtensionError' has no 'exit_code' member (no-member) - Disabled: e is actually an ExtensionOperationError
-                err_msg = 'Systemd process exited with code %s and output %s' % (
-                    e.exit_code, process_output)  # pylint: disable=no-member
-            else:
-                err_msg = "Systemd timed-out, output: %s" % process_output
-            raise SystemdRunError(err_msg)
-        finally:
-            with self._systemd_run_commands_lock:
-                self._systemd_run_commands.remove(process.pid)
+    def get_enabled_controllers_at_root(self):
```
maddieford, line 505: Should we add this to abstract class too? And Maybe we should add a comment in the abstract class that this only returns the agent supported controllers enabled at root (cpu and/or memory)

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -351,6 +351,22 @@ def _reset_agent_cgroup_setup(self):
             except Exception as err:
                 logger.warn("Error while resetting the quotas: {0}".format(err))
 
+        def _is_enforcement_supported(self, controller):
```
narrieta, line 354: this does not check whether enforcement is supported, but rather whether we are enforcing limits using that controller, right? if yes, can we rename the function

maddieford, line 354: What do you think about moving this to the controller implementations? 
CpuControllerV1 -> return True
CpuControllerV2 -> return False
_MemoryController -> return False

Then during cgroup initialization, we can just check controller.is_enforcement_supported()

azurelinuxagent/ga/cgroupcontroller.py
```diff
@@ -142,21 +143,31 @@ def _get_parameters(self, parameter_name, first_line_only=False):
...
+                    else:
+                        raise
+                return False
+
+            # In v1, the cgroup.procs file is present in the service/slice cgroup directory.
+            if _found_cgroup_procs(os.path.join(self.path, "cgroup.procs")):
+                return True
+
+            # In v2, the cgroup.procs file is present in the scope cgroup for extensions
+            for cgroup_file in glob.iglob(os.path.join(self.path, "*/cgroup.procs")):
```
narrieta, line 167: what is the * matching?

narrieta (commented):
posted some minor comments/questions

azurelinuxagent/ga/cgroupconfigurator.py
nagworld9: It does two things
1) If Agent supports enforcement (like feature support) 
2) If above is true, it will check controller is mounted/enabled to enforce the limits

nagworld9: Good idea, but for extensions we setup the limits before we start the extension, and we will not have controllers objects to get this information

azurelinuxagent/ga/cgroupcontroller.py
nagworld9: * matches sub-directories, here it will be child cgroups

ex:

/sys/fs/cgroup/azure.slice/azure-vmextensions.slice/azure-vmextensions-microsoft.azure.customscript.slice/**enable_123.scope**/cgroup.procs

nagworld9: I think we can scope it to *.scope/cgroup.procs

azurelinuxagent/ga/cgroupconfigurator.py
narrieta: where is 1 checked?

azurelinuxagent/ga/cgroupcontroller.py
```diff
@@ -164,7 +164,7 @@ def _found_cgroup_procs(file):
                 return True
 
             # In v2, the cgroup.procs file is present in the scope cgroup for extensions
-            for cgroup_file in glob.iglob(os.path.join(self.path, "*/cgroup.procs")):
+            for cgroup_file in glob.iglob(os.path.join(self.path, "*.scope/cgroup.procs")):
```
narrieta, line 167: is it possible to match more specifically the scopes we create?

azurelinuxagent/ga/cgroupconfigurator.py
narrieta: i see, that is my point. this is not checking for "supported" (e.g. CPU Quotas are supported on V2); it's checking whether we are using the feature or not. name is misleading

azurelinuxagent/ga/cgroupcontroller.py
narrieta: "*" would match any of those 3 examples, and any other scope, if there is ever one.

my ask is if we can make it match only the scope that we know we created

narrieta: actually, never mind, sorry, your original change was ok. we want to continue tracking as long as there are any processes, no matter who created them, so * is OK

azurelinuxagent/ga/cgroupconfigurator.py
nagworld9: renamed to` _can_use_controller`

narrieta: thanks

maddieford: Each controller implementation has a static can_enforce() check:
CpuControllerV1 -> return True
CpuControllerV2 -> return False
_MemoryController -> return False

Cgroupapi implementations have a can_enforce_cpu() check:
CgroupApiV1 -> return CgroupV1.CPU_CONTROLLER in self._cgroup_mountpoints and CpuControllerV1.can_enforce() 
CgroupApiV2 -> return CgropuV2.CPU_CONTROLLER in self._controllers_enabled_at_root and CpuControllerV2.can_enforce()

Then cgroupconfigurator just calls self._cgroupapi.can_enforce_cpu() 


^ This allows us to avoid these if/else conditions, but I'm also ok with leaving as is if this overcomplicates

nagworld9: I think we can even simply without static can_enforce(), not sure if we need this

Cgroupapi implementations have a can_enforce_cpu() check:
CgroupApiV1 -> return CgroupV1.CPU_CONTROLLER in self._cgroup_mountpoints
CgroupApiV2 -> return False

can_enforce_memory() -> return False

maddieford: Sounds good to me

tests/ga/test_cgroupconfigurator.py
```diff
@@ -176,6 +172,43 @@ def test_initialize_should_not_enable_cgroups_when_the_agent_is_not_in_the_syste
             self.assertEqual(len(tracked), 0, "No cgroups should be tracked. Tracked: {0}".format(tracked))
             self.assertFalse(os.path.exists(agent_drop_in_file_cpu_quota), "{0} should not have been created".format(agent_drop_in_file_cpu_quota))
 
+    def test_initialize_should_not_enable_cgroups_v2(self):
```
maddieford, line 175:
```suggestion
    def test_initialize_should_enable_cgroups_v2(self):
```
should enable cgroups v2?

narrieta (dismissed):

narrieta (approved):

maddieford (approved):

narrieta:
@ani-sinha - we follow the gitflow branching model. Once we release, we'll merge to master.

ani-sinha:
> @ani-sinha - we follow the gitflow branching model. Once we release, we'll merge to master.

Its very confusing because whereas 
[1] 
```
commit 0d7c8a7e41c14d05f2a3047dd4d5f28fdb971b48
Author: Nageswara Nandigam <84482346+nagworld9@users.noreply.github.com>
Date:   Mon Mar 3 15:35:03 2025 -0800

    handle systemd errors in cgroup (#3322) (#3325)
    
    (cherry picked from commit ffda9f45aceeaed3675e421f79763dc932d16db2)```

```
is in `main`  branch and 

[2] 
```
commit 1080ad05159dcc365ef27aa44319c6aa49b3051b
Author: Nageswara Nandigam <84482346+nagworld9@users.noreply.github.com>
Date:   Wed Feb 19 11:21:47 2025 -0800

    Enable resource monitoring in cgroup v2 machines (#3316)
    
    * enable resource monitoring in v2
    
    * address comments
    
    * address comment
    
    * new comments
    
    * ut fix

```
is not, in the release branch [1] is after [2] but in `main` branch [2] must be applied after [1] (since its not already there). So the order of the changes will be reversed between the release branch and  `main` branch.

PR #3321, opened 2025-02-20
Overwrite extension heartbeat with policy error

mgunnala:
## Description

Issue # 

This PR addresses the following TODOs for the policy/allowlist feature: 
- in the case of an extension disallowed error, overwrite extension heartbeat with the error
- instead of logging policy locally with each goal state, copy the policy file to the history folder
---

### PR information


azurelinuxagent/ga/exthandlers.py
```diff
@@ -508,9 +508,11 @@ def handle_ext_handlers(self, goal_state_id):
 
         # Instantiate policy engine, and use same engine to handle all extension handlers. If an error is thrown during
         # policy engine initialization, we block all extensions and report the error via handler status for each extension.
+        # Save policy to history folder.
         policy_error = None
         try:
-            policy_engine = ExtensionPolicyEngine()
+            gs_history = self.protocol.get_goal_state().history
+            policy_engine = ExtensionPolicyEngine(gs_history)
```
narrieta, line 515: you may want to have ExtensionPolicyEngine expose the path to the policy file and save it here, in order to avoid adding a dependency on GoalStateHistory

azurelinuxagent/common/utils/archive.py
```diff
@@ -309,3 +310,7 @@ def save_shared_conf(self, text):
 
     def save_manifest(self, name, text):
         self.save(text, _MANIFEST_FILE_NAME.format(name))
+
+    def save_policy(self, text):
```
narrieta, line 314: let's make this more generic, in case we need to add other files later... maybe add_file(self, path)?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -773,10 +775,27 @@ def __handle_ext_disallowed_error(self, ext_handler_i, error_code, report_op, me
 
         ext_handler_i.set_handler_status(status=ExtHandlerStatusValue.not_ready, message=message, code=error_code)
 
+        # If extension is reporting heartbeat, overwrite the heartbeat file with error message.
+        heartbeat_file = os.path.join(conf.get_lib_dir(), ext_handler_i.get_heartbeat_file())
+        if os.path.isfile(heartbeat_file):
```
narrieta, line 780: we need to do this every time we report heartbeat, not only on when the error happens

tests/ga/test_extension.py
```diff
@@ -3715,6 +3715,61 @@ def test_uninstall_should_succeed_if_extension_allowed(self):
                 vm_status = args[0]
                 self.assertEqual(0, len(vm_status.vmAgent.extensionHandlers))
 
+    def test_should_report_policy_failure_instead_of_heartbeat(self):
```
narrieta, line 3718: did we do any end-to-end testing?

mgunnala: I only added unit tests, since heartbeat isn't currently being e2e tested and none of the extensions we e2e test report heartbeat

If we want to add an e2e test for this, I could add AzureSecurityLinuxAgent to test heartbeat

mgunnala: Added an e2e test with AzureSecurityLinuxAgent

azurelinuxagent/common/utils/archive.py
mgunnala: Updated this to save_file(self, path)

azurelinuxagent/ga/exthandlers.py
mgunnala: Updated

azurelinuxagent/ga/exthandlers.py
mgunnala: Updated

azurelinuxagent/common/utils/archive.py
```diff
@@ -83,6 +83,7 @@
 _REMOTE_ACCESS_FILE_NAME = "RemoteAccess.xml"
 _EXT_CONF_FILE_NAME = "ExtensionsConfig.xml"
 _MANIFEST_FILE_NAME = "{0}.manifest.xml"
+_POLICY_FILE_NAME = "waagent_policy.json"
```
mgunnala, line 86: Can remove this because we added save_file()

narrieta: please do

azurelinuxagent/common/utils/archive.py
```diff
@@ -309,3 +310,11 @@ def save_shared_conf(self, text):
 
     def save_manifest(self, name, text):
         self.save(text, _MANIFEST_FILE_NAME.format(name))
+
+    def save_file(self, path):
+        """
+        Save contents of the specified file to the history folder. Output file will have the same name as the original file.
+        """
+        text = fileutil.read_file(path)
```
narrieta, line 318: file copy would be better to maintain the permissions of the original file

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1089,14 +1093,25 @@ def report_ext_handler_status(self, vm_status, ext_handler, goal_state_changed):
...
+            # If extension is disallowed, concatenate the heartbeat message to the existing handler status message, and
+            # do not override handler error code or status with heartbeat.
             if handler_state != ExtHandlerState.NotInstalled:
-                # Heartbeat is a handler level thing only, so we dont need to modify this
+                # Heartbeat is a handler level thing only, so we don't need to modify this
                 try:
                     heartbeat = ext_handler_i.collect_heartbeat()
                     if heartbeat is not None:
-                        handler_status.status = heartbeat.get('status')
+                        if not ext_disallowed:
```
narrieta, line 1103:
```
                        if ext_disallowed:
                            pass  # The status already describes why the extension is disallowed
                        else:
                            handler_status.status = heartbeat.get('status')
```

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -44,18 +44,27 @@ def __init__(self, msg, inner=None):
...
         if not self.policy_enforcement_enabled:
             return
 
         _PolicyEngine._log_policy_event("Policy enforcement is enabled.")
-        self._policy = self._parse_policy(self.__read_policy())
+        policy_contents = self.__read_policy()
+        self._policy = self._parse_policy(policy_contents)
+
+    @property
+    def policy(self):
```
narrieta, line 66: where is this being used?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1089,14 +1093,25 @@ def report_ext_handler_status(self, vm_status, ext_handler, goal_state_changed):
...
+                            handler_status.status = heartbeat.get('status')
+
                         if 'formattedMessage' in heartbeat:
-                            handler_status.message = parse_formatted_message(heartbeat.get('formattedMessage'))
+                            heartbeat_message = parse_formatted_message(heartbeat.get('formattedMessage'))
+                            if ext_disallowed:
+                                handler_status.message += " Extension was previously enabled and reported the following heartbeat:\n{0}".format(heartbeat_message)
```
maddieford, line 1109: handler_status.message is set to None in ExtHandlerStatus init. I don't think handler_status.message will be None since we write status on behalf of the Handler when it's disallowed, but we should probably add None check here before concatenating in case

mgunnala: If handler_status.message is None here, what should it be set to? 

report_ext_handler_status() doesn't have the exact cause of the ext disallowed error (policy failure, ext disabled, etc.) - we could set the message to something like "Extension will not be processed because it is disallowed. Extension was previously enabled ... "

azurelinuxagent/ga/exthandlers.py
```diff
@@ -508,8 +508,13 @@ def handle_ext_handlers(self, goal_state_id):
 
         # Instantiate policy engine, and use same engine to handle all extension handlers. If an error is thrown during
         # policy engine initialization, we block all extensions and report the error via handler status for each extension.
+        # Save policy to history folder.
         policy_error = None
         try:
+            gs_history = self.protocol.get_goal_state().history
+            policy_file = conf.get_policy_file_path()
+            if gs_history is not None and os.path.isfile(policy_file):
+                gs_history.save_file(policy_file)
```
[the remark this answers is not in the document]
maddieford: @narrieta @mgunnala what do you think of keeping the contents of the policy file when it's read as a property in ExtensionPolicyEngine. Then we can dump the contents to the history folder here in exthandlers. 

If we keep it this way, it will be unlikely that we save a different policy to the history folder than what is being applied, but if it does happen it will be difficult to debug

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: It's no longer being used currently, but I kept it here because we'll need to expose the policy for the extension-specific policy feature

azurelinuxagent/ga/exthandlers.py
maddieford: I think we can update the check to 'if ext_disallowed and handler_status.message is not None'

If ext is disallowed and the handler_status.message is None then we've hit an unexpected case and should just write the heartbeat to the status imo. 

We can add a comment that handler_status.message should not be None if the ext is disallowed (because the agent writes handler status on behalf of the ext in that case)

azurelinuxagent/common/utils/archive.py
mgunnala: Removed

azurelinuxagent/ga/exthandlers.py
```diff
@@ -508,8 +508,13 @@ def handle_ext_handlers(self, goal_state_id):
 
         # Instantiate policy engine, and use same engine to handle all extension handlers. If an error is thrown during
         # policy engine initialization, we block all extensions and report the error via handler status for each extension.
+        # Save policy to history folder.
         policy_error = None
         try:
+            gs_history = self.protocol.get_goal_state().history
+            policy_file = conf.get_policy_file_path()
```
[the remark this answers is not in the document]
mgunnala: Updated the implementation - ExtensionPolicyEngine exposes a property "policy_file_contents" that can be copied to history here

azurelinuxagent/ga/exthandlers.py
mgunnala: Updated

azurelinuxagent/ga/policy/policy_engine.py
mgunnala: I decided to remove this for now, I will add it back as needed in the future

azurelinuxagent/ga/exthandlers.py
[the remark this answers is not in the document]
mgunnala: Updated the implementation - ExtensionPolicyEngine stores raw policy file contents in "policy_file_contents" property immediately after reading, it is copied to history here.

azurelinuxagent/ga/exthandlers.py
mgunnala: Updated

narrieta: `
if ext_disallowed and handler_status.message is not None:
`
We should not just pass the heartbeat, but rather add a generic message explaining the extension was not execute. Also, since this is an unexpected condition, we should probably send telemetry to alert us of the issue. If sending telemetry, be careful not send on every single heartbeat

mgunnala: We do have unit tests asserting that status is appropriately reported (not None) when extension is disallowed. 

Is it sufficient to add a message like "Extension was not processed: failed to report status." ?

narrieta: when not None, we use

>Status + Extension was previously enabled and reported the following heartbeat:\n{0}

how about

> Extension was not executed, but it was previously enabled and reported the following heartbeat:\n{0}

mgunnala: That sounds good to me

mgunnala: Updated

narrieta (dismissed):

maddieford (approved):

narrieta (approved):

PR #3334, opened 2025-03-21
Require Hyper-V to start waagent service

vittyvk:
## Description  

 WALinuxAgent is Azure specific and Azure runs on Hyper-V. Add  

   ConditionVirtualization=microsoft
 
to service files to avoid starting WALA in other environments. In particular, this avoids polluting the console when the  
 same image is used both for Azure and some other env (e.g. local QEMU).

### PR information
- [ X ] Ensure development PR is based on the `develop` branch.
- [ X ] The title of the PR is clear and informative.
- [ X ] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).
- [ - ] If applicable, the PR references the bug/issue that it fixes in the description.
- [ - ] New Unit tests were added for the changes made

- [ X ] I have read the [contribution guidelines](https://github.com/Azure/WALinuxAgent/blob/master/.github/CONTRIBUTING.md).

### DIFF
--- a/init/redhat/waagent.service
+++ b/init/redhat/waagent.service
@@ -5,6 +5,7 @@ After=network-online.target
 
 ConditionFileIsExecutable=/usr/sbin/waagent
 ConditionPathExists=/etc/waagent.conf
+ConditionVirtualization=microsoft
 
 [Service]
 Type=simple
--- a/init/arch/waagent.service
+++ b/init/arch/waagent.service
@@ -5,6 +5,7 @@ After=network-online.target
 
 ConditionFileIsExecutable=/usr/bin/waagent
 ConditionPathExists=/etc/waagent.conf
+ConditionVirtualization=microsoft
 
 [Service]
 Type=simple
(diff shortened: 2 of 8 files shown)


init/redhat/waagent.service
```diff
@@ -5,6 +5,7 @@ After=network-online.target
 
 ConditionFileIsExecutable=/usr/sbin/waagent
 ConditionPathExists=/etc/waagent.conf
+ConditionVirtualization=microsoft
```
jeremycline, line 8: This looks fine to me, although it does differ from the [patch we're carrying in Fedora](https://src.fedoraproject.org/rpms/WALinuxAgent/blob/rawhide/f/0001-waagent.service-set-ConditionVirtualization-microsof.patch) in that it's not a triggering conditional. 

@cjp256 do you still feel that it should be a triggering conditional? In my opinion, it's just as easy to `systemctl --edit --full waagent.service` as it is to drop in an additional triggering conditional, but I've not thought hard about it.

jeremycline (approved):
For what it's worth, this looks good to me and helps reduce the patches distributions have to carry. There's a slight deviation from Fedora's downstream patch by Chris, but I think it's okay unless he has strong feelings about it being a triggering conditional.

PR #3335, opened 2025-03-27
redact telemetry events

nagworld9:
## Description

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/common/utils/restutil.py
+++ b/azurelinuxagent/common/utils/restutil.py
@@ -473,7 +473,7 @@ def http_request(method,
 
             if request_failed(resp):
                 if _is_retry_status(resp.status, retry_codes=retry_codes):
-                    msg = '[HTTP Retry] {0} {1} -- Status Code {2}'.format(method, url, resp.status)
+                    msg = '[HTTP Retry] {0} {1} -- Status Code {2}'.format(method, redact_sas_tokens_in_urls(url), resp.status)
                     # Note if throttled and ensure a safe, minimum number of
                     # retry attempts
                     if _is_throttle_status(resp.status):
@@ -503,7 +503,7 @@ def http_request(method,
             if return_raw_response:  # skip all error handling
                 raise
             clean_url = _trim_url_parameters(url)
-            msg = '[HTTP Failed] {0} {1} -- HttpException {2}'.format(method, clean_url, e)
+            msg = '[HTTP Failed] {0} {1} -- HttpException {2}'.format(method, redact_sas_tokens_in_urls(clean_url), e)
             if _is_retry_exception(e):
                 continue
             break
@@ -513,7 +513,7 @@ def http_request(method,
                 raise
             IOErrorCounter.increment(host=host, port=port)
             clean_url = _trim_url_parameters(url)
-            msg = '[HTTP Failed] {0} {1} -- IOError {2}'.format(method, clean_url, e)
+            msg = '[HTTP Failed] {0} {1} -- IOError {2}'.format(method, redact_sas_tokens_in_urls(clean_url), e)
             continue
 
     raise HttpError("{0} -- {1} attempts made".format(msg, attempt))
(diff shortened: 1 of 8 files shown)


maddieford (approved):

azurelinuxagent/common/utils/restutil.py
```diff
@@ -513,7 +513,7 @@ def http_request(method,
                 raise
             IOErrorCounter.increment(host=host, port=port)
             clean_url = _trim_url_parameters(url)
-            msg = '[HTTP Failed] {0} {1} -- IOError {2}'.format(method, clean_url, e)
+            msg = '[HTTP Failed] {0} {1} -- IOError {2}'.format(method, redact_sas_tokens_in_urls(clean_url), e)
```
narrieta, line 516: this should be done in event.py instead, right?

azurelinuxagent/ga/extensionprocessutil.py
```diff
@@ -159,11 +156,7 @@ def read_output(stdout, stderr):
         stderr = ustr(stderr.read(TELEMETRY_MESSAGE_MAX_LEN), encoding='utf-8',
                       errors='backslashreplace')
 
-        def redact(s):
-            # redact query strings that look like SAS tokens
-            return SAS_TOKEN_RE.sub(r'\1<redacted>', s)
-
-        return format_stdout_stderr(redact(stdout), redact(stderr))
+        return format_stdout_stderr(redact_sas_token(stdout), redact_sas_token(stderr))
```
narrieta, line 159: event.py instead?

azurelinuxagent/common/utils/restutil.py
nagworld9: we shouldn't write to agent logs as well

azurelinuxagent/ga/extensionprocessutil.py
nagworld9: This output goes to status file if extension fails with error code, so we still need it here

azurelinuxagent/common/utils/restutil.py
nagworld9: Updated: Now, every event and log will go through sanitization

azurelinuxagent/common/logger.py
```diff
@@ -135,13 +138,23 @@ def write_log(log_appender):  # pylint: disable=W0612
             msg = msg_format.format(*args)
         else:
             msg = msg_format
+
+        # using local method here to avoid import textutil cyclic dependency
```
narrieta, line 142: we shouldn't duplicate this.

to break the cycle, refactor to another module that textutil and logger can use

azurelinuxagent/ga/extensionprocessutil.py
narrieta: should we then redact in the method that creates the status report?

nagworld9: I don't like the idea; we need to do at handler status and extension status etc. But if that's the way, I'll do it in all those places

narrieta: can we do it as part of preparing the report?


```
    def prepare(self, blob_type):
        logger.verbose("Prepare status blob")
        self.data = self.to_json()
        self.type = blob_type

```

nagworld9: This will not prevent writing sensitive information to status file. May be sanitizing before we write is correct place

narrieta: what status file?

nagworld9: Handlerstatus and extension status file in status folder

azurelinuxagent/ga/collect_telemetry_events.py
```diff
@@ -459,6 +462,7 @@ def process_events(self):
                     logger.verbose("Processing event file: {0}", event_file_path)
 
                     event = self._read_and_parse_event_file(event_file_path)
+                    redact_event_msg(event)
```
narrieta, line 465: how is this cleanup related to the one at line 363?

nagworld9: We have two separate collections, one for agent events and one for extension events.

Line 363 for extension events and 465 for agent events. Even though we cleanup the agent events before we write to disk, we still need here just in case if old daemons include sas tokens in the events.

narrieta: ah, yes, "legacy" events. makes sense

narrieta (approved):

maddieford (approved):

PR #3350, opened 2025-04-21
Add e2e agent removal scenario

maddieford:
## Description
Agent versions are regularly removed from PIR (either as a part of regular maintenance or if a version needs to be rolled back). When an agent version is removed from PIR, it is expected that the version will be removed from all regional manifests so the agent is no longer available for download/update. There have been scenarios in the past where a removed agent version was unexpectedly still available in the manifest. There is also a risk of accidentally removing an agent version from the manifest that should still be available. 

This e2e test scenario can be ran after PIR removal to ensure that the expected agent version was removed from the manifest and/or to ensure that no additional versions were unexpectedly removed from the manifest. 

Scenario details:
- The scenario runs on all locations in agent_removal.yml for the particular cloud the pipeline is ran for. The scenario can also be ran on a single location if provided as a pipeline parameter. 
- The scenario checks Prod manifests by default, but can also check if Test manifests if the following option test argument is provided: GAFamily=Test
- If the expected_versions test argument is provided, the scenario will check that the manifest versions exactly match the provided list of expected versions. This is useful to ensure that no additional versions were unexpectedly removed from the manifest.
> expected_versions=2.11.1.12;2.12.0.2;2.13.1.1
- If the removed_version test argument is provided, the scenario will check that the manifest does not contain the removed version. This is useful if there are different sets of versions published across regions (a release is in progress, for example).
> removed_version=2.9.1.1

Issue # 

---

### PR information

### DIFF
--- a/tests_e2e/orchestrator/lib/agent_test_loader.py
+++ b/tests_e2e/orchestrator/lib/agent_test_loader.py
@@ -225,7 +225,7 @@ class AgentTestLoader(object):
               - "bvts/vm_access.py"
             images: "endorsed"
             locations: "AzureCloud:eastuseaup"
-            runs_on_all_locations: false
+            runs_on_all_suite_locations: false
             owns_vm: true
             install_test_agent: true
             template: "bvts/template.py"
@@ -245,7 +245,7 @@ class AgentTestLoader(object):
         * locations - [Optional; string or list of strings] If given, the test suite must be executed on that cloud location(e.g. "AzureCloud:eastus2euap").
                      If not specified, or set to an empty string, the test suite will be executed in the default location. This is useful
                      for test suites that exercise a feature that is enabled only in certain regions.
-        * runs_on_all_locations - [Optional; boolean] By default a test suite runs on one location per image. If this
+        * runs_on_all_suite_locations - [Optional; boolean] By default a test suite runs on one location per image. If this
                     value is True, then the test will run on all listed locations for each image.
         * owns_vm - [Optional; boolean] By default all suites in a test run are executed on the same test VMs; if this
                     value is set to True, new test VMs will be created and will be used exclusively for this test suite.
--- a/tests_e2e/orchestrator/lib/agent_test_loader.py
+++ b/tests_e2e/orchestrator/lib/agent_test_loader.py
@@ -55,6 +55,8 @@ class TestSuiteInfo(object):
     images: List[str]
     # The locations (regions) on which the suite must run; if empty, the suite can run on any location
     locations: List[str]
+    # If locations is provided, this determines whether the suite should run on all the suite locations or just one
+    runs_on_all_suite_locations: bool
     # Whether this suite must run on its own test VM
     owns_vm: bool
     # If True, the suite must run on a scale set (instead of a single VM)
@@ -223,6 +225,7 @@ class AgentTestLoader(object):
               - "bvts/vm_access.py"
             images: "endorsed"
             locations: "AzureCloud:eastuseaup"
+            runs_on_all_locations: false
             owns_vm: true
             install_test_agent: true
             template: "bvts/template.py"
@@ -242,6 +245,8 @@ class AgentTestLoader(object):
         * locations - [Optional; string or list of strings] If given, the test suite must be executed on that cloud location(e.g. "AzureCloud:eastus2euap").
                      If not specified, or set to an empty string, the test suite will be executed in the default location. This is useful
                      for test suites that exercise a feature that is enabled only in certain regions.
+        * runs_on_all_locations - [Optional; boolean] By default a test suite runs on one location per image. If this
+                    value is True, then the test will run on all listed locations for each image.
         * owns_vm - [Optional; boolean] By default all suites in a test run are executed on the same test VMs; if this
                     value is set to True, new test VMs will be created and will be used exclusively for this test suite.
                     This is useful for suites that modify the test VMs in such a way that the setup may cause problems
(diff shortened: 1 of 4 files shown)


tests_e2e/orchestrator/lib/agent_test_loader.py
```diff
@@ -55,6 +55,8 @@ class TestSuiteInfo(object):
     images: List[str]
     # The locations (regions) on which the suite must run; if empty, the suite can run on any location
     locations: List[str]
+    # If locations is provided, this determines whether the suite should run on all the suite locations or just one
+    runs_on_all_suite_locations: bool
```
narrieta, line 59: I don't feel this parameter is needed, one can already set a specific location in the runbook parameters.

tests_e2e/tests/agent_removal/agent_removal.py
```diff
@@ -0,0 +1,118 @@
...
+        # been written to history folder yet.
+        manifest_uris = []
+        for _ in range(4):
+            current_gs_history_path = self._ssh_client.run_command("find /var/lib/waagent/history -mindepth 1 -type d", use_sudo=True).rstrip()
+
+            # Try to get extensions config for current goal state
+            ext_config_path = os.path.join(current_gs_history_path, _EXT_CONF_FILE_NAME)
+            try:
+                ext_config_text = self._ssh_client.run_command("cat {0}".format(ext_config_path), use_sudo=True)
+                xml_doc = parse_doc(ext_config_text)
```
nagworld9, line 86: We have all this logic in agent, can we import agent module instead of duplicate similar logic here? If we need to change logic around this area then we need it here as well

We have done similarly in other tests. Reference: https://github.com/Azure/WALinuxAgent/blob/master/tests_e2e/tests/scripts/agent_update-wait_for_rsm_gs.py#L32

tests_e2e/orchestrator/lib/agent_test_loader.py
nagworld9: +1

tests_e2e/tests/agent_removal/agent_removal.py
```diff
@@ -0,0 +1,118 @@
...
+        for uri in manifest_uris:
+            log.info("")
+            log.info("URI: {0}".format(uri))
+            agent_versions = self._ssh_client.run_command("curl -s {0} | grep -oP '(?<=<Version>).*?(?=</Version>)'".format(uri)).splitlines()
+            log.info("Agent versions in manifest: {0}".format(agent_versions))
+            if self._expected_versions is not None:
+                log.info("Expected versions: {0}".format(self._expected_versions))
+                assert_that(self._expected_versions).described_as("Expected agent versions in manifest does not match actual agent versions in manifest. Expected={0}, Actual={1}".format(self._expected_versions, agent_versions)).is_equal_to(agent_versions)
+                log.info("Expected versions match actual agent versions in manifest")
+            else:
```
nagworld9, line 67: can we support expected versions and removed versions test in the same run?

maddieford: What is the advantage of supporting both in the same run? Expected versions list should not contain the versions we removed

nagworld9: I'm with you on Expected vessions but I don't get why we need removed version test. Expected version helping all the cases right, i dont see when we need removed version check

maddieford: The versions may be different in canary than rest of Prod. If running on all regions, removed_version should be used if canary versions are different than rest of prod

tests_e2e/orchestrator/lib/agent_test_loader.py
maddieford: Removed, and updated combinator to always run on all locations listed in the test suite. Also updated the runbook parameter to allow a comma separated list of locations

tests_e2e/tests/agent_removal/agent_removal.py
maddieford: Updated

tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
```diff
@@ -265,9 +266,10 @@ def create_environment_list(self, test_suites: List[str]) -> List[Dict[str, Any]
     def create_existing_vm_environment(self, test_suites: List[str]) -> Dict[str, Any]:
         loader = AgentTestLoader(test_suites, self.runbook.cloud)
 
+        location = self.runbook.locations.split(',')[0]
```
narrieta, line 269: here we expect a single location. rather than split, passing a list should be an error

narrieta: similar for vmss

tests_e2e/tests/scripts/agent_removal-verify_manifest_versions.py
```diff
@@ -0,0 +1,84 @@
...
+from azurelinuxagent.common.protocol.goal_state import GoalStateProperties, ExtensionManifest
+from azurelinuxagent.common.protocol.wire import WireProtocol
+from tests_e2e.tests.lib.logging import log
+from tests_e2e.tests.lib.remote_test import run_remote_test
+from tests_e2e.tests.lib.retry import retry
+from tests_e2e.tests.lib.shell import run_command
+
+
+def _get_manifest_uris(wire_protocol: WireProtocol, family: str) -> List[str]:
+    wire_protocol.client.update_goal_state()
```
narrieta, line 36: we should do a few retries here

tests_e2e/tests/scripts/agent_removal-verify_manifest_versions.py
```diff
@@ -0,0 +1,84 @@
...
+    retry(lambda: protocol.client.reset_goal_state(goal_state_properties=GoalStateProperties.ExtensionsGoalState))
+    manifest_uris = _get_manifest_uris(protocol, ga_family)
+    log.info("Successfully retrieved manifest uris from goal state.")
+
+    log.info("")
+    log.info("Validating agent versions in manifest URIs from goal state...")
+    for uri in manifest_uris:
+        log.info("")
+        log.info("URI: {0}".format(uri))
+        xml_text = run_command(["curl", "-s", "{0}".format(uri)])
```
narrieta, line 67: curl is not present on all distros, you should this in python, or use http_get.py

narrieta (approved):

nagworld9 (approved):

PR #3352, opened 2025-04-22
Mark images not available on all clouds

narrieta:


tests_e2e/test_suites/images.yml
```diff
@@ -190,8 +190,14 @@ images:
...
          - "Standard_B2s"
-   rhel_810: "RedHat RHEL 8_10 latest"
-   rhel_95: "RedHat RHEL 9_5 latest"
+   rhel_810:
+      urn: "RedHat RHEL 8_10 latest"
+      locations:
+         AzureChinaCloud: []
+   rhel_95:
+      urn: "RedHat RHEL 9_5 latest"
+   locations:
```
nagworld9, line 199: it should be in subsection

narrieta: thanks; fixed

PR #3378, opened 2025-05-19
Recogninze Alpaquita Linux, as Alpine derivative

nbuwe:
## Description

Issue # N/A

This PR introduces minimal changes to make the agent recognize [Alpaquita Linux](https://github.com/bell-sw/alpaquita-aports) as Alpine derivative.

While Alpine is not an [endorsed distribution](https://learn.microsoft.com/en-us/azure/virtual-machines/linux/endorsed-distros) on Azure, some support for Alpine is included in the agent and these PR fills a few gaps in that support.

The proposed changes should not affect endorsed distributions and should not pose a maintenance burden.  The "if alpine" block in `setup.py` is the only block that intrudes into the common code.

---

### PR information


nbuwe:
@microsoft-github-policy-service agree company="BellSoft"

nagworld9 (approved):

nagworld9:
LGTM. The changes have seemed isolated to the Alpine distro. Our test automation doesn't currently support Alpine, but the tests run on other distros look good.

azurelinuxagent/common/osutil/alpine.py
```diff
@@ -32,11 +32,12 @@ def is_dhcp_enabled(self):
         return True
 
     def get_dhcp_pid(self):
-        return self._get_dhcp_pid(["pidof", "dhcpcd"])
+        return sorted(self._get_dhcp_pid(["pidof", "dhcpcd"]))
```
narrieta, line 35: Just want to confirm that these changes are also meant to apply to Alpine. not only to Alpaquita

nbuwe: Yes.

It's still a kludge overall, but that `sorted` is a counterpart to the `[0]` in the next hunk (which is absolutely required since `pid` must be a single number).  The whole thing requires a proper rework, but I didn't want to get nerdsniped about it.  With per-interface dhcpcd instances and multiple processes for privilege separation it's a whole separate can of worms, b/c "_get dhcp pid_" is a question that can't be meaningfully answered without more context.  With the current numerology that sorted gives a slightly better chance for that `[0]` to return a "better" answer.

So here I was just aiming to get the bare minimum fix to make it run at all.

PR #3400, opened 2025-06-13
Fix object reference when comparing against the updated goal state

narrieta:
In UpdateHandler._try_update_goal_state(), the local goal_state and UpdateHandler._goal_state were pointing to the same object, so this condition was never True:

`
self._goal_state.extensions_goal_state.id != goal_state.extensions_goal_state.id
`

As a result, self_check_certificates() was not executed when a new FastTrack goal state is available. This breaks scenarios like a hibernate/resume cycle, when the goal state needs to be refreshed to get the updated certificates.

This was a regression introduced by #3273. The previous unit test was focused on the retry logic, while it should have focused on the actual scenario. This PR also updates the test.

### DIFF
--- a/tests/ga/test_update.py
+++ b/tests/ga/test_update.py
@@ -2054,125 +2054,6 @@ class TryUpdateGoalStateTestCase(HttpRequestPredicates, AgentTestCase):
             update_handler._try_update_goal_state(protocol)
             self.assertEqual(update_handler._goal_state.incarnation, '6789', "The goal state was not updated (received unexpected incarnation)")
 
-    def test_it_should_refresh_certificates_on_fast_track_goal_state_after_hibernate_resume_cycle(self):
-        #
-        # A hibernate/resume cycle is a special case in that it produces a new Fabric goal state with incarnation 1 on resume. Since the VM is re-allocated,
-        # the goal state will include a new tenant encryption certificate. If the incarnation was also 1 before hibernation, the Agent won't detect this new
-        # goal state and subsequent Fast Track goal states will fail because the Agent has not fetched the new certificate.
-        #
-        # To address this issue, before executing any Fast Track goal state, _try_update_goal_state() checks that the current goal state includes the
-        # certificate used by extensions to decrypt their protected settings and forces a refresh if it does not.
-        #
-        # The test data below uses files captured from an actual scenario (minus edits to remove irrelevant/sensitive data) and consists of 3 goal states:
-        #
-        # * goal_state_1: WireServer + HGAP (Fast Track) goal state before hibernation; incarnation 1.
-        # * goal_state_2: WireServer + HGAP (Fabric) goal state after resume; also incarnation 1, but new certificates.
-        # * goal_state_3: Fast Track goal state (requires new certificates)
-        #
-        update_handler = get_update_handler()
-
-        goal_state_1 = wire_protocol_data.DATA_FILE.copy()
-        goal_state_1.update({
-            "goal_state": "hibernate/goal_state_1/GoalState.xml",
-            "hosting_env": "hibernate/goal_state_1/HostingEnvironmentConfig.xml",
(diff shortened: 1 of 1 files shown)


azurelinuxagent/ga/update.py
```diff
@@ -531,13 +531,12 @@ def _try_update_goal_state(self, protocol):
             for attempt in range(3):
                 protocol.client.update_goal_state(force_update=attempt > 0, silent=self._update_goal_state_error_count >= max_errors_to_log, save_to_history=True)
 
-                goal_state = protocol.get_goal_state()
-                new_goal_state = self._goal_state is None or self._goal_state.extensions_goal_state.id != goal_state.extensions_goal_state.id
+                self._goal_state = protocol.get_goal_state()
```
maddieford, line 534: Is this a no-op? Seems that self._goal_state already pointed to protocol.client._goal_state before?

narrieta: In the current implementation it is, but that implementation needs to change and then this assignment will be needed, so I decided to be explicit about it.

tests/data/hibernate/TransportCert.pem
maddieford, line 1: Is a TransportCert unique to this test scenario necessary?

tests/ga/test_update.py
```diff
@@ -2053,52 +2054,124 @@ def test_it_should_update_the_goal_state(self):
...
+            #
+            # On resume, the Agent will receive goal_state_2, but since the incarnation is also 1, it won't detect it as a new goal state and
+            # _try_update_goal_state won't fetch the new data.
+            #
+            # Note that the Agent does detect the new VmSettings, but since they represent a Fabric goal state, it ignores them.
+            #
+            protocol.mock_wire_data = wire_protocol_data.WireProtocolData(goal_state_2)
+
+            with patch('azurelinuxagent.common.protocol.goal_state.add_event') as patch_add_event:
+                update_handler._try_update_goal_state(protocol)
```
maddieford: on this second invocation of _try_update_goal_state, self._processing_new_extensions_goal_state() is returning True incorrectly because self._last_extensions_gs_id only gets updated in _process_goal_state.

tests/ga/test_update.py
```diff
@@ -2053,52 +2054,124 @@ def test_it_should_update_the_goal_state(self):
...
+            if not any(message == e for e in telemetry_events):
+                raise Exception('Unexpected Agent behavior. Expected a telemetry event matching "{0}"; got: {1}'.format(message, telemetry_events))
+
+            #
+            # This is the actual test: when a Fast Track goal state shows up, the Agent should pull the certificates that originated in the previous
+            # Fabric goal state, and the updated goal state should include all the certificates referenced by the extensions in the new goal state.
+            #
+            protocol.mock_wire_data = wire_protocol_data.WireProtocolData(goal_state_3)
+
+            update_handler._try_update_goal_state(protocol)
```
maddieford: same comment here about self._last_extensions_gs_id being None in self._processing_new_extensions_goal_state() check

tests/data/hibernate/TransportCert.pem
narrieta: yes, all certs under the 'hibernate' directory are encrypted with this specific Transport cert

tests/ga/test_update.py
narrieta: Thank you so much for catching this! This should teach me to double-check those tests exercise the test subject as expected :)

I'm testing this logic now with UpdateHandler._process_goal_state() instead of UpdateHandler._try_update_goal_state. That way all the state involved in this logic is updated correctly (including self._last_extensions_gs_id).

I moved the test to the ProcessGoalState class

tests/ga/test_update.py
narrieta: Thanks again!

maddieford (approved):

nagworld9 (approved):

PR #3404, opened 2025-06-19
Pick latest version based on new and legay autoupdates flag

nagworld9:
## Description
The issue is when a user explicitly sets AutoUpdate.UpdateToLatestVersion=y and AutoUpdate.Enabled=n. In this case, the extension handler downloads the latest agent based on the new flag. However, the daemon code responsible for starting the agent doesn't consider the new flag, it still relies on the legacy logic using AutoUpdate.Enabled to determine whether to use the latest or the currently installed version. As a result, even after downloading the new version, the daemon ends up starting the installed version.

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -199,7 +199,10 @@ class UpdateHandler(object):
         if self.signal_handler is None:
             self.signal_handler = signal.signal(signal.SIGTERM, self.forward_signal)
 
-        latest_agent = None if not conf.get_autoupdate_enabled() else self.get_latest_agent_greater_than_daemon(
+        # If new flag explicitly set, agent will use latest agent downloaded and will not fall back to installed agent. See the new flag definition in conf.py
+        use_latest_agent = conf.is_present("AutoUpdate.UpdateToLatestVersion") or conf.get_autoupdate_enabled()
+
+        latest_agent = None if not use_latest_agent else self.get_latest_agent_greater_than_daemon(
             daemon_version=CURRENT_VERSION)
         if latest_agent is None:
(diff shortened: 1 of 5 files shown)


azurelinuxagent/ga/update.py
```diff
@@ -199,7 +199,10 @@ def run_latest(self, child_args=None):
         if self.signal_handler is None:
             self.signal_handler = signal.signal(signal.SIGTERM, self.forward_signal)
 
-        latest_agent = None if not conf.get_autoupdate_enabled() else self.get_latest_agent_greater_than_daemon(
+        # If new flag explicitly set, agent will use latest agent downloaded and will not fall back to installed agent. See the new flag definition in conf.py
+        use_latest_agent = conf.is_present("AutoUpdate.UpdateToLatestVersion") or conf.get_autoupdate_enabled()
```
narrieta, line 215: needs to be present and be True

nagworld9: No, we don't want to fallback to old behavior. We want to stay on latest download agent event if its No. Thats one of the reason why we add new flag.

nagworld9: No- for preventing further new agent downloads

nagworld9: Once agent is on particular version, AutoUpdate.UpdateToLatestVersion=n should not revert to installed version.

narrieta: yes, value does not matter

we should add a warning message here as well... if autoupdate.enabled is also present, log/telemetry that it will be ignored

nagworld9: added log/telemetry

maddieford (approved):

PR #3406, opened 2025-06-24
Log collector should not re-create transport cert

narrieta:
The log collector and the extension handler can be in a race where the collector is resetting the transport cert while the handler is trying to use it.

Added flag 'create_transport_certificate' to initialize the protocol without creating/resetting the transport cert.

### DIFF
--- a/azurelinuxagent/common/protocol/util.py
+++ b/azurelinuxagent/common/protocol/util.py
@@ -188,7 +188,7 @@ class ProtocolUtil(SingletonPerThread):
                 return
             logger.error("Failed to clear wiresever endpoint: {0}", e)
 
-    def _detect_protocol(self, save_to_history, init_goal_state=True):
+    def _detect_protocol(self, init_goal_state, create_transport_certificate, save_to_history):
         """
         Probe protocol endpoints in turn.
         """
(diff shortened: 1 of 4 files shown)


azurelinuxagent/common/protocol/util.py
```diff
@@ -188,7 +188,7 @@ def _clear_wireserver_endpoint(self):
                 return
             logger.error("Failed to clear wiresever endpoint: {0}", e)
 
-    def _detect_protocol(self, save_to_history, init_goal_state=True):
+    def _detect_protocol(self, init_goal_state, create_transport_certificate, save_to_history):
```
narrieta, line 191: the parameters to this function were not in the same order as the parameters for

`WireProtocol.detect(self, init_goal_state=True, create_transport_certificate=True, save_to_history=False):`

Switched the order of init_goal_state and save_to_history

nagworld9: removed default value for init_goal_state, is that intentional?

narrieta: yes, private methods should not have defaults in general

nagworld9 (approved):

PR #3410, opened 2025-07-02
Add test for goal state after hibernate-resume

narrieta:
Adding test to verify that the certificates in the goal state are refreshed after a hibernate-resume cycle.

Hibernate-resume can change the IP address of the VM, and the Agent was maintaining a static list of test nodes and their IP addresses. I changed that to get that information from the AgentTestContext, and added methods in the context to refresh the address.

At this point, only this test changes the IP address, and I decided to make the test's responsibility to keep the AgentTestContext up to date. If we need to this functionality for other tests, we can switch that responsibility to the AgentTestSuite and, for example, refresh the context after executing each test.

### DIFF
--- a/tests_e2e/orchestrator/lib/agent_test_suite.py
+++ b/tests_e2e/orchestrator/lib/agent_test_suite.py
@@ -39,8 +39,7 @@ from lisa import (  # pylint: disable=E0401
 )
 from lisa.environment import EnvironmentStatus  # pylint: disable=E0401
 from lisa.messages import TestStatus, TestResultMessage  # pylint: disable=E0401
-from lisa.node import LocalNode  # pylint: disable=E0401
-# from lisa.node import Nodes  # pylint: disable=E0401
+from lisa.node import LocalNode, Nodes  # pylint: disable=E0401
 from lisa.util.constants import RUN_ID  # pylint: disable=E0401
 from lisa.sut_orchestrator.azure.common import get_node_context  # pylint: disable=E0401
 from lisa.sut_orchestrator.azure.platform_ import AzurePlatform  # pylint: disable=E0401
@@ -134,7 +133,7 @@ class AgentTestSuite(LisaTestSuite):
         self._runbook_name: str  # name of the runbook execution, used as prefix on ARM resources created by the AgentTestSuite
 
         self._lisa_log: Logger  # Main log for the LISA run
-        # self._lisa_nodes: Nodes  # List of test nodes maintained by LISA
+        self._lisa_nodes: Nodes  # List of test nodes maintained by LISA
 
         self._lisa_environment_name: str  # Name assigned by LISA to the test environment, useful for correlation with LISA logs
         self._environment_name: str  # Name assigned by the AgentTestSuiteCombinator to the test environment
@@ -196,7 +195,7 @@ class AgentTestSuite(LisaTestSuite):
         self._runbook_name = variables["name"]
 
         self._lisa_log = lisa_log
-        # self._lisa_nodes = environment.nodes
+        self._lisa_nodes = environment.nodes
 
         self._lisa_environment_name = environment.name
         self._environment_name = variables["c_env_name"]
@@ -387,13 +386,24 @@ class AgentTestSuite(LisaTestSuite):
         finally:
             self._setup_lock.release()
 
+    def _refresh_lisa_nodes(self, test_context: AgentVmTestContext) -> None:
+        """
+        Tests can change the IP address of the test VM, and LISA does SSH connections during test cleanup, so we refresh the IP address of the LISA node
+        with the current value from the test context.
+        """
+        for n in self._lisa_nodes.list():
+            if (n.is_remote):
+                n.set_connection_info(
+                    use_public_address=True,
+                    public_address=test_context.ip_address,
+                    port=n.connection_info["port"],
+                    username=n.connection_info["username"],
+                    private_key_file=n.connection_info["private_key_file"])
+
     def _clean_up(self, success: bool) -> None:
         """
         Cleans up any items created by the test suite run.
         """
-        # for n in self._lisa_nodes:
-        #     n.close()
-        #
         if self._delete_scale_set:
             if self._keep_environment == KeepEnvironment.Always:
                 log.info("Won't delete the scale set %s, per the test suite configuration.", self._vmss_name)
@@ -588,6 +598,8 @@ class AgentTestSuite(LisaTestSuite):
                     finally:
                         if self._collect_logs == CollectLogs.Always or self._collect_logs == CollectLogs.Failed and not test_suite_success:
                             self._collect_logs_from_test_nodes(test_context)
+                        if isinstance(test_context, AgentVmTestContext):  # Scale sets are not managed by LISA, so do this only for VM tests.
+                            self._refresh_lisa_nodes(test_context)
 
                 except Exception as e:   # pylint: disable=bare-except
                     # Report the error and raise an exception to let LISA know that the test errored out.
--- a/tests_e2e/orchestrator/lib/agent_test_suite.py
+++ b/tests_e2e/orchestrator/lib/agent_test_suite.py
@@ -39,7 +39,8 @@ from lisa import (  # pylint: disable=E0401
 )
 from lisa.environment import EnvironmentStatus  # pylint: disable=E0401
 from lisa.messages import TestStatus, TestResultMessage  # pylint: disable=E0401
-from lisa.node import LocalNode, Nodes  # pylint: disable=E0401
+from lisa.node import LocalNode  # pylint: disable=E0401
+# from lisa.node import Nodes  # pylint: disable=E0401
 from lisa.util.constants import RUN_ID  # pylint: disable=E0401
 from lisa.sut_orchestrator.azure.common import get_node_context  # pylint: disable=E0401
 from lisa.sut_orchestrator.azure.platform_ import AzurePlatform  # pylint: disable=E0401
@@ -133,7 +134,7 @@ class AgentTestSuite(LisaTestSuite):
         self._runbook_name: str  # name of the runbook execution, used as prefix on ARM resources created by the AgentTestSuite
 
         self._lisa_log: Logger  # Main log for the LISA run
-        self._lisa_nodes: Nodes  # List of test nodes maintained by LISA
+        # self._lisa_nodes: Nodes  # List of test nodes maintained by LISA
 
         self._lisa_environment_name: str  # Name assigned by LISA to the test environment, useful for correlation with LISA logs
         self._environment_name: str  # Name assigned by the AgentTestSuiteCombinator to the test environment
@@ -195,7 +196,7 @@ class AgentTestSuite(LisaTestSuite):
         self._runbook_name = variables["name"]
 
         self._lisa_log = lisa_log
-        self._lisa_nodes = environment.nodes
+        # self._lisa_nodes = environment.nodes
 
         self._lisa_environment_name = environment.name
         self._environment_name = variables["c_env_name"]
@@ -390,9 +391,9 @@ class AgentTestSuite(LisaTestSuite):
         """
         Cleans up any items created by the test suite run.
         """
-        for n in self._lisa_nodes:
-            n.close()
-
+        # for n in self._lisa_nodes:
+        #     n.close()
+        #
         if self._delete_scale_set:
             if self._keep_environment == KeepEnvironment.Always:
                 log.info("Won't delete the scale set %s, per the test suite configuration.", self._vmss_name)
--- a/tests_e2e/orchestrator/lib/agent_test_suite.py
+++ b/tests_e2e/orchestrator/lib/agent_test_suite.py
@@ -388,8 +388,8 @@ class AgentTestSuite(LisaTestSuite):
 
     def _refresh_lisa_nodes(self, test_context: AgentVmTestContext) -> None:
         """
-        Tests can change the IP address of the test VM, and LISA does SSH connections during test cleanup, so we refresh the IP address of the LISA node
-        with the current value from the test context.
+        Tests can change the IP address of the test VM. We refresh the IP address of the LISA node before we return control from our test, since LISA performs SSH operations on the test
+        nodes before completing the test run.
         """
         for n in self._lisa_nodes.list():
             if (n.is_remote):
--- a/tests_e2e/orchestrator/lib/agent_test_suite.py
+++ b/tests_e2e/orchestrator/lib/agent_test_suite.py
@@ -39,7 +39,7 @@ from lisa import (  # pylint: disable=E0401
 )
 from lisa.environment import EnvironmentStatus  # pylint: disable=E0401
 from lisa.messages import TestStatus, TestResultMessage  # pylint: disable=E0401
-from lisa.node import LocalNode  # pylint: disable=E0401
+from lisa.node import LocalNode, Nodes  # pylint: disable=E0401
 from lisa.util.constants import RUN_ID  # pylint: disable=E0401
 from lisa.sut_orchestrator.azure.common import get_node_context  # pylint: disable=E0401
 from lisa.sut_orchestrator.azure.platform_ import AzurePlatform  # pylint: disable=E0401
@@ -114,18 +114,6 @@ class TestFailedException(Exception):
         super().__init__(msg)
 
 
-class _TestNode(object):
-    """
-    Name and IP address of a test VM
-    """
-    def __init__(self, name: str, ip_address: str):
-        self.name = name
-        self.ip_address = ip_address
-
-    def __str__(self):
-        return f"{self.name}:{self.ip_address}"
-
-
 @TestSuiteMetadata(area="waagent", category="", description="")
 class AgentTestSuite(LisaTestSuite):
     """
@@ -145,6 +133,7 @@ class AgentTestSuite(LisaTestSuite):
         self._runbook_name: str  # name of the runbook execution, used as prefix on ARM resources created by the AgentTestSuite
 
         self._lisa_log: Logger  # Main log for the LISA run
+        self._lisa_nodes: Nodes  # List of test nodes maintained by LISA
 
         self._lisa_environment_name: str  # Name assigned by LISA to the test environment, useful for correlation with LISA logs
         self._environment_name: str  # Name assigned by the AgentTestSuiteCombinator to the test environment
@@ -177,8 +166,6 @@ class AgentTestSuite(LisaTestSuite):
         self._vm_ip_address: str
         self._vmss_name: str
 
-        self._test_nodes: List[_TestNode]  # VMs or scale set instances the tests will run on
-
         # Whether to create and delete a scale set.
         self._create_scale_set: bool
         self._delete_scale_set: bool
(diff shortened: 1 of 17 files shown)


tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -114,18 +114,6 @@ def __init__(self, env_name: str, test_cases: List[str]):
         super().__init__(msg)
 
 
-class _TestNode(object):
```
narrieta, line 117: The AgentTestSuite used a static list of test nodes and their IP addresses. I removed that list and now this information can be retrieved, as well as refreshed, from the AgentTestContext.

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -208,6 +195,7 @@ def _initialize(self, environment: Environment, variables: Dict[str, Any], lisa_
         self._runbook_name = variables["name"]
 
         self._lisa_log = lisa_log
+        self._lisa_nodes = environment.nodes
```
narrieta, line 198: The IP address used by those nodes needs to be refreshed, since LISA uses them for SSH during test cleanup.

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -663,56 +665,31 @@ def _execute_test_suite(self, suite: TestSuiteInfo, test_context: AgentTestConte
                             summary.append(f"[Passed]  {test.name}")
                             log.info("******** [Passed] %s", test.name)
                             self._lisa_log.info("[Passed] %s", test_full_name)
-                            self._report_test_result(
-                                suite_full_name,
-                                test.name,
-                                TestStatus.PASSED,
-                                test_start_time)
+                            self._report_test_result(suite_full_name, test.name, TestStatus.PASSED, test_start_time)
```
narrieta, line 668: no changes in these except clauses, just collapsing the method calls to a single line for easier reading

tests_e2e/tests/lib/vm_extension_identifier.py
```diff
@@ -59,14 +59,15 @@ class VmExtensionIds(object):
     Only the major version is specified, and the minor version is set to 0 (set autoUpgradeMinorVersion to True in the call to enable
     to use the latest version)
     """
+    AzureMonitorLinuxAgent: VmExtensionIdentifier = VmExtensionIdentifier(publisher='Microsoft.Azure.Monitor', ext_type='AzureMonitorLinuxAgent', version="1.5")
```
narrieta, line 62: Alphabetized these and added Hibernation extension

tests_e2e/tests/hibernation/hibernation.py
```diff
@@ -0,0 +1,188 @@
...
+                }
+            )
+            log.info("Reallocating %s after enabling hibernation", self._context.vm)
+            self._context.vm.start()
+            self._refresh_ip_address()
+
+        #
+        # The test may be running on an existing machine where the Hibernate extension has already been installed, so we check that before installing it.
+        #
+        if model.instance_view.extensions is not None and any(e for e in model.instance_view.extensions if e.type == "Microsoft.CPlat.Core.LinuxHibernateExtension"):
```
nagworld9, line 183: do we need to install hibernateextension to enable hibernation capability?

tests_e2e/tests/lib/agent_test_context.py
```diff
@@ -103,6 +140,21 @@ class AgentVmssTestContext(AgentTestContext):
...
+        """
+        if self._test_nodes is None:
+            self.refresh_ip_addresses()
+        return self._test_nodes
+
+    def refresh_ip_addresses(self) -> None:
+        """
+        Updates self.ip_address to reflect the current IP address of the VM.
+        """
+        self._test_nodes = [TestNode(name=i.instance_name, ip_address=i.ip_address) for i in self.vmss.get_instances_ip_address()]
```
nagworld9, line 157: This is confusing, refesh_ip_addresses updates the test nodes?

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -402,6 +386,20 @@ def _setup_test_run(self) -> None:
         finally:
             self._setup_lock.release()
 
+    def _refresh_lisa_nodes(self, test_context: AgentVmTestContext) -> None:
+        """
+        Tests can change the IP address of the test VM, and LISA does SSH connections during test cleanup, so we refresh the IP address of the LISA node
+        with the current value from the test context.
+        """
+        for n in self._lisa_nodes.list():
+            if (n.is_remote):
```
nagworld9, line 395: what does is_remote checks?

tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -597,7 +597,9 @@ def _execute(self) -> None:
 
                     finally:
                         if self._collect_logs == CollectLogs.Always or self._collect_logs == CollectLogs.Failed and not test_suite_success:
-                            self._collect_logs_from_test_nodes()
+                            self._collect_logs_from_test_nodes(test_context)
+                        if isinstance(test_context, AgentVmTestContext):  # Scale sets are not managed by LISA, so do this only for VM tests.
+                            self._refresh_lisa_nodes(test_context)
```
nagworld9, line 602: I assume we refresh this for LISA cleanup. It seems this place is not appropriate for that, and better place would be in self._clean_up or after  self._execute(). That way it clearly indicates our intention

tests_e2e/tests/hibernation/hibernation.py
```diff
@@ -0,0 +1,188 @@
...
+            raise Exception(f"Unexpected behavior: The incarnation is not 1 after a hibernate-resume cycle. Incarnation is {incarnation}")
+        log.info("The current incarnation is 1")
+
+        log.info("Checking tenant certificate after resume...")
+        post_hibernation_tenant_certificate = self._get_tenant_certificate()
+        if post_hibernation_tenant_certificate == pre_hibernation_tenant_certificate:
+            raise Exception("Unexpected behavior: hibernate-resume did not create a new tenant certificate")
+        log.info("Tenant certificate: %s", post_hibernation_tenant_certificate)
+        log.info("Checking that the new tenant certificate has not been downloaded yet...")
+        downloaded = self._ssh_client.run_command(f"ls /var/lib/waagent/{post_hibernation_tenant_certificate}.* || true", use_sudo=True).rstrip()
```
nagworld9: This is not checking latest goal state processed by agent and didn't download new certs. We need to check both, otherwise if test check happens before goal state processed, we always get new certs not downloaded.

tests_e2e/tests/lib/agent_test_context.py
narrieta: yes, the IPs of the test nodes

any suggestions?

tests_e2e/orchestrator/lib/agent_test_suite.py
narrieta: whether it is a RemoteNode and therefore has a set_connection_info method

tests_e2e/orchestrator/lib/agent_test_suite.py
narrieta: it needs to be done after executing tests and before lisa cleans ups, and it requires the test context. our cleanup method does not have access to the context

tests_e2e/tests/hibernation/hibernation.py
narrieta: yes, requirements are in the link at the top of the file

tests_e2e/tests/hibernation/hibernation.py
narrieta: yes, you are right, there is a race condition with the agent processing the goal state.

i'll remove the check, I added it just as debugging info (that is the reason there are no asserts here). The real test is below: the extension must succeed

tests_e2e/orchestrator/lib/agent_test_suite.py
narrieta: actually, I'll update the comment in _refresh_lisa_nodes to remove any reference to cleanup, since actually, I do not know what LISA is doing, some kind of check on bash

```
2025-07-02T16:52:22Z.983 [INFO] [ThreadPoolExecutor-27_0] initializing node 'lisa-WALinuxAgent-20250702-164623-076-e0-n0' w****t@4.155.234.62:22
2025-07-02T16:52:23Z.532 [INFO] [Thread-42] lisa test for bash prompt
2025-07-02T16:52:24Z.099 [INFO] [ThreadPoolExecutor-5_0] deleting resource group: lisa-WALinuxAgent-20250702-164623-076-e0, wait: False
```

so the intention is not to do help with LISA cleanup, but rather to update the LISA data structure before our test returns control to LISA, since that data structure can be out of date and LISA would fail.

nagworld9: In that case, I would place it in at the end of execute or after execute, that's when we return control. But as you mentioned, it requires test context, we can do it at the end of self._execute

tests_e2e/tests/lib/agent_test_context.py
nagworld9: In the case of single vm, we update the ip_address and for vmss, we update the nodes, that's where it confuses me.

tests_e2e/orchestrator/lib/agent_test_suite.py
narrieta: The end of self._execute cannot access the test context safely either.

The scope of test_context is the try/finally on lines 578-602. Outside that scope, test_context could be undefined (e.g. due to an exception).

tests_e2e/tests/lib/agent_test_context.py
narrieta: Do you mean the implementation rather than the method name? 

Yes, the implementation is different, AgentVmTestContext has a self.ip_address member, since it is a single VM. Scale sets have multiple instances, which are represented as self._test_nodes in AgentVmssTestContext.

The signature of the method is consistent across AgentVmTestContext and AgentVmssTestContext, since it is the implementation of the AgentTestContext.refresh_ip_addresses() abstract method.

tests_e2e/orchestrator/lib/agent_test_suite.py
nagworld9: yeah, we could little refactor to change scope. But that's fine since we added a great comment why we need to refresh it. I'm ok as its.

tests_e2e/tests/lib/agent_test_context.py
nagworld9: yeah, implementation.

tests_e2e/orchestrator/lib/agent_test_suite.py
narrieta: yeah, the refactor is not worth it... keeping the scope as-is is a better option. Using smaller scopes makes the code easier to read and understand

tests_e2e/tests/lib/virtual_machine_client.py
```diff
@@ -183,32 +185,69 @@ def restart(
             return
 
         start = datetime.datetime.now(UTC)
+
+        self._wait_for_status("PowerState/running", boot_timeout)
+
+        # self._wait_for_power_state() works by checking the instance view, and we may capture a view from before the reboot actually happened, so we verify
```
maddieford: Do you mean self._wait_for_status()?
```suggestion
        # self._wait_for_status() works by checking the instance view, and we may capture a view from before the reboot actually happened, so we verify
```

tests_e2e/tests/hibernation/hibernation.py
```diff
@@ -0,0 +1,214 @@
...
+    The Agent has logic to detect this scenario and fetch the required certificates; this test verifies that logic.
+
+    See https://learn.microsoft.com/en-us/azure/virtual-machines/linux/hibernate-resume-linux for details on hibernation
+    """
+    def __init__(self, context: AgentVmTestContext):
+        super().__init__(context)
+        self._ssh_client: SshClient = self._context.create_ssh_client()
+
+    def run(self):
+        log.info("Executing test on %s - IP address: %s", self._context.vm, self._context.ip_address)
```
maddieford, line 53: could we add a new line after this to make the test logs more readable?
`log.info("")`

tests_e2e/tests/hibernation/hibernation.py
```diff
@@ -0,0 +1,214 @@
...
+
+        #
+        # The test may be running on an existing machine where the Hibernate extension has already been installed, so we check that before installing it.
+        #
+        if model.instance_view.extensions is not None and any(e for e in model.instance_view.extensions if e.type == "Microsoft.CPlat.Core.LinuxHibernateExtension"):
+            log.info("The Hibernation extension is already installed")
+        else:
+            log.info("Installing the Hibernation extension")
+            hibernate = VirtualMachineExtensionClient(self._context.vm, VmExtensionIds.Hibernate, resource_name="Hibernate")
+            hibernate.enable(auto_upgrade_minor_version=True)
```
maddieford, line 188: suggestion: add `log.info("")` at the end of this method. there's lots of logs as a part of enabling hibernation

tests_e2e/tests/hibernation/hibernation.py
```diff
@@ -0,0 +1,214 @@
...
+                raise Exception(f"The incarnation was not reset to 1 after a hibernate-resume cycle. Incarnation is {incarnation}")
+        log.info("The current incarnation is 1")
+
+        #
+        # Do a hibernate-resume cycle; this will generate a new tenant certificate but, since the incarnation has not changed,
+        # the Agent will not download it yet.
+        #
+        log.info("Retrieving tenant certificate before hibernation...")
+        pre_hibernation_tenant_certificate = self._get_tenant_certificate()
+        log.info("Tenant certificate: %s", pre_hibernation_tenant_certificate)
```
maddieford, line 82:
```suggestion
        log.info("Tenant certificate: %s", pre_hibernation_tenant_certificate)
        log.info("")
```

tests_e2e/tests/hibernation/hibernation.py
```diff
@@ -0,0 +1,214 @@
...
+        # Check that the current incarnation is 1; if that is not the case, do a hibernate-resume cycle to reset the incarnation to 1
+        #
+        log.info("Retrieving current incarnation...")
+        incarnation = self._ssh_client.run_command("get_goal_state.py --tag Incarnation", use_sudo=True).rstrip()
+        if incarnation != "1":
+            self._do_hibernate_resume_cycle()
+            incarnation = self._ssh_client.run_command("get_goal_state.py --tag Incarnation", use_sudo=True).rstrip()
+            if incarnation != "1":
+                raise Exception(f"The incarnation was not reset to 1 after a hibernate-resume cycle. Incarnation is {incarnation}")
+        log.info("The current incarnation is 1")
```
maddieford, line 73:
```suggestion
        log.info("The current incarnation is 1")
        log.info("")
```

tests_e2e/tests/hibernation/hibernation.py
```diff
@@ -0,0 +1,214 @@
...
+        # Do a hibernate-resume cycle; this will generate a new tenant certificate but, since the incarnation has not changed,
+        # the Agent will not download it yet.
+        #
+        log.info("Retrieving tenant certificate before hibernation...")
+        pre_hibernation_tenant_certificate = self._get_tenant_certificate()
+        log.info("Tenant certificate: %s", pre_hibernation_tenant_certificate)
+
+        log.info("Triggering a hibernate-resume cycle to test the tenant certificate...")
+        hibernate_time = self._ssh_client.get_time()
+        self._do_hibernate_resume_cycle()
```
maddieford, line 87:
```suggestion
        self._do_hibernate_resume_cycle()
        log.info("")
```

tests_e2e/tests/hibernation/hibernation.py
```diff
@@ -0,0 +1,214 @@
...
+
+        log.info("Retrieving current incarnation...")
+        incarnation = self._ssh_client.run_command("get_goal_state.py --tag Incarnation", use_sudo=True).rstrip()
+        if incarnation != "1":
+            raise Exception(f"Unexpected behavior: The incarnation is not 1 after a hibernate-resume cycle. Incarnation is {incarnation}")
+        log.info("The current incarnation is 1")
```
maddieford, line 94:
```suggestion
        log.info("The current incarnation is 1")
        log.info("")
```

tests_e2e/tests/hibernation/hibernation.py
```diff
@@ -0,0 +1,214 @@
...
+        post_hibernation_tenant_certificate = self._get_tenant_certificate()
+        if post_hibernation_tenant_certificate == pre_hibernation_tenant_certificate:
+            raise Exception("Unexpected behavior: hibernate-resume did not create a new tenant certificate")
+        log.info("Tenant certificate: %s", post_hibernation_tenant_certificate)
+        post_hibernation_tenant_certificate_path = f"{os.path.join(conf.get_lib_dir(), post_hibernation_tenant_certificate)}.crt"
+        log.info(f"Checking that the new tenant certificate has not been downloaded yet ({post_hibernation_tenant_certificate_path})...")
+        downloaded = self._ssh_client.run_command(f"ls {post_hibernation_tenant_certificate_path} || true", use_sudo=True).rstrip()
+        if downloaded != "":
+            raise Exception(f"Unexpected behavior: The new tenant certificate was downloaded after resume: {downloaded}")
+        log.info("The new tenant certificate has not been downloaded yet.")
```
maddieford, line 107:
```suggestion
        log.info("The new tenant certificate has not been downloaded yet.")
        log.info("")
```

tests_e2e/tests/hibernation/hibernation.py
```diff
@@ -0,0 +1,214 @@
...
+
+        #
+        # Execute an extension with protected settings and verify that the new tenant certificate was downloaded.
+        #
+        log.info("Executing an extension with protected settings to verify it can use the new tenant certificate")
+        custom_script = VirtualMachineExtensionClient(self._context.vm, VmExtensionIds.CustomScript, resource_name="CustomScript")
+        message = str(uuid.uuid4())
+        custom_script.enable(protected_settings={'commandToExecute': f"echo \'{message}\'"}, force_update=True)
+        custom_script.assert_instance_view(expected_message=message)
```
maddieford, line 117:
```suggestion
        custom_script.assert_instance_view(expected_message=message)
        log.info("")
```

tests_e2e/tests/hibernation/hibernation.py
```diff
@@ -0,0 +1,214 @@
...
+
+        log.info("Checking that the new tenant certificate was downloaded...")
+        downloaded = self._ssh_client.run_command(f"find {conf.get_lib_dir()} -name '*.crt'", use_sudo=True)
+        if post_hibernation_tenant_certificate not in downloaded:
+            fail(f"The new tenant certificate ({post_hibernation_tenant_certificate}) was not downloaded:\n{downloaded}")
+        log.info("The new tenant certificate was downloaded.")
```
maddieford, line 124:
```suggestion
        log.info("The new tenant certificate was downloaded.")
        log.info("")
```

maddieford (commented):
Test changes lgtm. The hibernation scenario logs are verbose and a little difficult to read right now. I just suggested adding some new lines after each code block in the test to help

nagworld9 (dismissed):

tests_e2e/tests/hibernation/hibernation.py
narrieta: Agree, but I added the blank line in the main test method instead. Thanks for the suggestions

nagworld9 (approved):

PR #3414, opened 2025-07-09
report update errorcode for extension update failures

nagworld9:
## Description

Today we don't populate the terminal(update) error code when ext update fails, as a result CRP waits for 90mins timeout.  The PR fixing that and report the update errorcode.
Issue # 

---

### PR information

### DIFF
--- a/tests_e2e/test_suites/ext_update.yml
+++ b/tests_e2e/test_suites/ext_update.yml
@@ -5,4 +5,3 @@ name: "ExtensionUpdate"
 tests:
   - "ext_update/ext_update_failure.py"
 images: "random(endorsed)"
-owns_vm: true
--- a/tests_e2e/test_suites/ext_update.yml
+++ b/tests_e2e/test_suites/ext_update.yml
@@ -1,0 +1,8 @@
+#
+#    This test verifies that the agent reports correct status and code when operation fails during the ext update.
+#
+name: "ExtensionUpdate"
+tests:
+  - "ext_update/ext_update_failure.py"
+images: "random(endorsed)"
+owns_vm: true
--- a/tests_e2e/tests/ext_update/ext_update_failure.py
+++ b/tests_e2e/tests/ext_update/ext_update_failure.py
@@ -114,7 +114,7 @@ class ExtensionUpdateFailureTest(AgentVmTest):
                 'message': r"name=Microsoft.Azure.Extensions.CustomScript, op=Disable, message=\[ExtensionOperationError\] .*disablefailed"
             },
 
-            # In some distros, the scope not being found is not an error, so we consider as systemd error
+            # In some distros, scope not included in the error, so we consider as systemd error
             # 2025-07-21T22:07:23.541880Z INFO ExtHandler ExtHandler [CGW] Disabling resource usage monitoring. Reason: Failed to start Microsoft.Azure.Extensions.CustomScript-2.0.7 using systemd-run, will try invoking the extension directly. Error: [SystemdRunError] Systemd process exited with code 1 and output [stdout]
             # [stderr]
             # Failed to find executable /var/lib/waagent/Microsoft.Azure.Extensions.CustomScript-2.0.7/disablefailed: No such file or directory
--- a/tests_e2e/tests/ext_update/ext_update_failure.py
+++ b/tests_e2e/tests/ext_update/ext_update_failure.py
@@ -36,30 +36,24 @@ class ExtensionUpdateFailureTest(AgentVmTest):
 
     def run(self):
         ssh_client: SshClient = self._context.create_ssh_client()
-        is_arm64: bool = ssh_client.get_architecture() == "aarch64"
-
         custom_script_2_0 = VirtualMachineExtensionClient(
             self._context.vm,
             VmExtensionIds.CustomScript,
             resource_name="CustomScript")
 
-        if is_arm64:
-            log.info("Will skip the update scenario, since currently there is only 1 version of CSE on ARM64")
-            return
-        else:
-            log.info("Installing %s", custom_script_2_0)
-            message = f"Hello {uuid.uuid4()}!"
-            custom_script_2_0.enable(
-                protected_settings={
-                    'commandToExecute': f"echo \'{message}\'"
-                },
-                auto_upgrade_minor_version=False
-            )
-            custom_script_2_0.assert_instance_view(expected_version="2.0", expected_message=message)
+        log.info("Installing %s", custom_script_2_0)
+        message = f"Hello {uuid.uuid4()}!"
+        custom_script_2_0.enable(
+            protected_settings={
+                'commandToExecute': f"echo \'{message}\'"
+            },
+            auto_upgrade_minor_version=False
+        )
+        custom_script_2_0.assert_instance_view(expected_version="2.0", expected_message=message)
 
         log.info("Modifying existing handler commands (disable operation) in HandlerManifest.json to fail the disable operation during update")
 
-        output = ssh_client.run_command(f"ext_update-modify_handler_manifest.py --extension-name '{custom_script_2_0._identifier}' --cmd-name disableCommand --cmd-value 'disablefailed'", use_sudo=True)
+        output = ssh_client.run_command(f"ext_update-modify_handler_manifest.py --extension-name '{custom_script_2_0._identifier}' --properties disableCommand=disablefailed continueOnUpdateFailure=false", use_sudo=True)
         log.info("Modified handlerManifest.json:\n%s", output)
 
         custom_script_2_1 = VirtualMachineExtensionClient(
--- a/tests_e2e/tests/ext_update/ext_update_failure.py
+++ b/tests_e2e/tests/ext_update/ext_update_failure.py
@@ -95,6 +95,7 @@ class ExtensionUpdateFailureTest(AgentVmTest):
             if ext_name in extension_names_on_vm:
                 self._ssh_client.run_command(f"ext_update-modify_handler_manifest.py --extension-name '{ext_name}' --reset", use_sudo=True)
                 ext.delete()
+                log.info("Removed extension %s", ext_name)
 
     def get_ignore_error_rules(self) -> List[Dict[str, Any]]:
         ignore_rules = [
--- a/tests_e2e/tests/ext_update/ext_update_failure.py
+++ b/tests_e2e/tests/ext_update/ext_update_failure.py
@@ -1,0 +1,104 @@
+#!/usr/bin/env python3
+
+# Microsoft Azure Linux Agent
+#
+# Copyright 2018 Microsoft Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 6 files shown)


tests_e2e/tests/ext_update/ext_update_failure.py
```diff
@@ -0,0 +1,104 @@
...
+        message = f"Hello {uuid.uuid4()}!"
+        try:
+            custom_script_2_1.enable(
+                protected_settings={
+                    'commandToExecute': f"echo \'{message}\'"
+                }
+            )
+            fail("The agent should have reported an error processing the extension update")
+        except Exception as error:
+            assert_that(str(error)).described_as(f"Expected VMExtensionHandlerNonTransientError/ExtensionUpdateError for {custom_script_2_1._identifier}, but actual error was: {error}").contains("VMExtensionHandlerNonTransientError").contains(f"{custom_script_2_1._identifier}").contains("ExtensionUpdateError")
```
maddieford: Ext status shows as transitioning: 
image
Shouldn't it be failed?

tests_e2e/tests/ext_update/ext_update_failure.py
```diff
@@ -0,0 +1,104 @@
...
+                protected_settings={
+                    'commandToExecute': f"echo \'{message}\'"
+                },
+                auto_upgrade_minor_version=False
+            )
+            custom_script_2_0.assert_instance_view(expected_version="2.0", expected_message=message)
+
+        log.info("Modifying existing handler commands (disable operation) in HandlerManifest.json to fail the disable operation during update")
+
+        output = ssh_client.run_command(f"ext_update-modify_handler_manifest.py --extension-name '{custom_script_2_0._identifier}' --cmd-name disableCommand --cmd-value 'disablefailed'", use_sudo=True)
```
maddieford: should we also update the manifest so continueOnUpdate is False?

tests_e2e/tests/ext_update/ext_update_failure.py
```diff
@@ -0,0 +1,104 @@
...
+        ssh_client: SshClient = self._context.create_ssh_client()
+        is_arm64: bool = ssh_client.get_architecture() == "aarch64"
+
+        custom_script_2_0 = VirtualMachineExtensionClient(
+            self._context.vm,
+            VmExtensionIds.CustomScript,
+            resource_name="CustomScript")
+
+        if is_arm64:
+            log.info("Will skip the update scenario, since currently there is only 1 version of CSE on ARM64")
```
maddieford: Is there an extension we can use with more than 1 version in arm64?

tests_e2e/test_suites/ext_update.yml
```diff
@@ -0,0 +1,8 @@
+#
+#    This test verifies that the agent reports correct status and code when operation fails during the ext update.
+#
+name: "ExtensionUpdate"
+tests:
+  - "ext_update/ext_update_failure.py"
+images: "random(endorsed)"
+owns_vm: true
```
maddieford: is it necessary for this test to own the vm?

Could we add cleanup at the end of the test to uninstall the failed ext

nagworld9: Yes, its needed because other scenarios might install CSE already in the vm, then this scenario won;t work

tests_e2e/tests/ext_update/ext_update_failure.py
nagworld9: This scenario is independent to distro or type. Hence, I use random(endorsed). So it doesn't matter vm type but I'll see if we can use other extension

tests_e2e/tests/ext_update/ext_update_failure.py
nagworld9: Don't know how transitioning coming in the portal, we dont report that in status. As far CRP operation, its failed with error immediately

nagworld9: did it show as extension not finished execution? or operation completed but showing status transitioning?

tests_e2e/test_suites/ext_update.yml
maddieford: For the Ext policy scenarios, we've been cleaning up the existing extensions at the top of the test run to avoid those issues

nagworld9: ok, in other scenarios we don't do it. If vm runs other scenarios first, we will be a trouble. I'll add cleanup but still required own_vm:true

maddieford: We do it in the Ext Policy scenario because pre-installed extensions may cause issues for that scenario, sounds like the same case for this scenario. 

If we cleanup CSE at the start of this test, why do we still need owns_vm:true?

nagworld9: I was thinking about cleanup after test run. are we going down the path of cleanup other scenario mess.

maddieford: Only if the existence of the extension would cause issues for the current test (that is rare for most scenarios)

Cleanup of single extension seems less costly than creating another environment, but I'm ok either way

tests_e2e/tests/ext_update/ext_update_failure.py
maddieford: It seems that the operation completed. I checked on a windows machine and it also shows transitioning in this case, so I guess that is the expected behavior

nagworld9: I don't know who parses the report status, they may need to change it.

tests_e2e/test_suites/ext_update.yml
nagworld9: I just realized we can't let this vm shared by others as it may create issues for other scenarios if they use CSE. Once we modified the handlermanfiest, we need to reset it to uninstall extension. That required to save the state before change. It just makes complex for simple scenario. Given we run in single vm, I wouldn't worry about creating new environment.

tests_e2e/tests/ext_update/ext_update_failure.py
nagworld9: I checked other extensions, they are not supported in all distros as well. Then we need write logic to skip on those. Since scenario not distro dependent, I'll skip for arm64

tests_e2e/tests/ext_update/ext_update_failure.py
nagworld9: done

tests_e2e/tests/ext_update/ext_update_failure.py
maddieford: It may be worth checking with CRP folks

maddieford (dismissed):

tests_e2e/tests/ext_update/ext_update_failure.py
nagworld9: I checked with Joe on this; he pointed that in general handler status of NotReady maps to an extension status of Transitioning in instanceview. This is unrelated to update scenario. Whether its need an update or not is a separate discussion I guess.

https://msazure.visualstudio.com/One/_git/Compute-CPlat-Core?path=/src/CRP/Dev/ComputeResourceProvider/Service/API/Operations/InVMArtifactsInstanceViewHelpers.cs&version=GBmaster&_a=contents

tests_e2e/test_suites/ext_update.yml
narrieta: we should try to re-use the VM

restoring the original manifest is very simple, but if for some reason it is not, deleting CSE would clean up

tests_e2e/tests/ext_update/ext_update_failure.py
```diff
@@ -0,0 +1,98 @@
...
+            protected_settings={
+                'commandToExecute': f"echo \'{message}\'"
+            },
+            auto_upgrade_minor_version=False
+        )
+        custom_script_2_0.assert_instance_view(expected_version="2.0", expected_message=message)
+
+        log.info("Modifying existing handler commands (disable operation) in HandlerManifest.json to fail the disable operation during update")
+
+        output = ssh_client.run_command(f"ext_update-modify_handler_manifest.py --extension-name '{custom_script_2_0._identifier}' --properties disableCommand=disablefailed continueOnUpdateFailure=false", use_sudo=True)
```
narrieta: seems like we are using a private property (_identifier)

tests_e2e/test_suites/ext_update.yml
nagworld9: deleting after manifest change is not possible because disable failure comes in the way of delete. Restore requires to add get function. We can do that

tests_e2e/tests/ext_update/ext_update_failure.py
nagworld9: what is the diff between self._context and this _identifier. We been using it all over the place

narrieta: self._context is a property of the current object, and objects can access their private members.
 
 custom_script_2_0._identifier is accessing a private property of another object

tests_e2e/test_suites/ext_update.yml
narrieta: you can make a copy of the manifest before changing it, then restore the copy

tests_e2e/tests/ext_update/ext_update_failure.py
nagworld9: We are not doing nested accessing, custom_script_2_0 is the object for VirtualMachineExtensionClient and _identifier is member of that class

narrieta: ExtensionUpdateFailureTest is using private members of VirtualMachineExtensionClient

narrieta: you could just use VmExtensionIds.CustomScript, btw

nagworld9: yes, thats what I'm going to do

tests_e2e/tests/ext_update/ext_update_failure.py
maddieford: For the policy related failures, we report at the handler level and also create extension level status. See azurelinuxagent.ga.exthandlers.ExtHandlersHandler.__handle_ext_disallowed_error comments for more details. I'm wondering if that's the correct approach here too..

nagworld9: what is correct approach? In this current approach, there is no issue in CRP around status polling. The translation between crp and vminstance view got implemented that way, changing our side will not change between them. The areas like these are very critical paths, we should minimize the changes as much possible. Joe has concern around changing the current logic in crp for vminstance as well, the chances of breaking someone are high.

maddieford: The reason we made the change for policy/disallowed ext scenario is because of stale status issues. 

Stale status issue can come up in the following scenario:
1. customer attempts to update CSE to 2.1.13.
2. CSE 2.1.13 fails during install. 
3. guest agent reports failure at handler level 
4. CRP sees terminal handler error code reported for CSE and fails the operation 
5. Customer tries to install CSE 2.1.13 again
6. CSE 2.1.13 is successfully installed this time
7. CRP polls status blob before agent has a chance to report the success
8. CRP sees terminal failure at handler level (from prev GS) and fails the operation, but extension actually succeeded for this goal state

For policy, we attempt to work around this by always reporting the failure at the extension and handler level, since CRP will poll for the specific seq num for extensions with settings. Of course, this does not solve the problem for extensions with no settings, or cases where the goal state is reapplied without the seq num increasing. 

I approved the PR since this issue is orthogonal to the one that was reported in the incident

narrieta (approved):

PR #3416, opened 2025-07-15
dnsutils is not available on debian_10

maddieford:
## Description
The publish_hostname scenario has been failing on debian 10 because the test tries to install the dnsutils package, but it is not available anymore:
```
# apt install -y dnsutils
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package dnsutils is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'dnsutils' has no installation candidate
```

This issue was already seen on debian 9, where the host cmd is being used to get the hostname. Making the same change for debian 10 with this PR

Issue # 

---

### PR information

### DIFF
--- a/tests_e2e/tests/publish_hostname/publish_hostname.py
+++ b/tests_e2e/tests/publish_hostname/publish_hostname.py
@@ -65,7 +65,7 @@ class PublishHostname(AgentVmTest):
         except CommandError as e:
             if "dig: command not found" in e.stderr:
                 distro = self._ssh_client.run_command("get_distro.py").rstrip().lower()
-                if "debian_9" in distro:
+                if "debian_9" in distro or "debian_10" in distro:
                     # Debian 9 hostname look up needs to be done with "host" instead of dig
                     lookup_cmd = "host {0}".format(self._private_ip)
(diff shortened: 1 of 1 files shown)


tests_e2e/tests/publish_hostname/publish_hostname.py
```diff
@@ -65,7 +65,7 @@ def check_and_install_dns_tools(self):
         except CommandError as e:
             if "dig: command not found" in e.stderr:
                 distro = self._ssh_client.run_command("get_distro.py").rstrip().lower()
-                if "debian_9" in distro:
+                if "debian_9" in distro or "debian_10" in distro:
```
narrieta: why was it that we use in instead of ==?

maddieford: the get_distro.py script returns a string which includes the minor version of the distro too

distro is something like debian_1013, so the test is using in since is it only cares about the major version

narrieta: ok... should we be using startswith instead?

maddieford: Updated

nagworld9 (approved):

PR #3433, opened 2025-08-15
Copy pypy tarball from orchestrator machine

nagworld9:
## Description

Added a fallback for pypy pkg to copy it from orchestrator machine when direct download fails in the test vm

Issue # 

---

### PR information

### DIFF
--- a/tests_e2e/orchestrator/lib/agent_test_suite.py
+++ b/tests_e2e/orchestrator/lib/agent_test_suite.py
@@ -428,7 +428,7 @@ class AgentTestSuite(LisaTestSuite):
             ssh_client.run_command("rm -rvf ~/{bin,lib,tmp}", use_sudo=True)
 
             #
-            # Copy Pypy, the test Agent, and the test tools to the test node
+            # Copy the test Agent, and the test tools to the test node
             #
             ssh_client = SshClient(ip_address=node.ip_address, username=self._user, identity_file=Path(self._identity_file))
             target_path = Path("~")/"tmp"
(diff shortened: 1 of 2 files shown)


tests_e2e/orchestrator/lib/agent_test_suite.py
```diff
@@ -452,8 +452,20 @@ def _setup_test_nodes(self, test_context: AgentTestContext) -> None:
             command = f"tar xvf {target_path/self._test_tools_tarball_path.name}"
             log.info("Remote command [%s] completed:\n%s", command, ssh_client.run_command(command))
 
-            log.info("Downloading Pypy to %s", node.name)
-            log.info(ssh_client.run_command(f"~/bin/download-pypy {self._cloud}"))
+            base_path = Path("/tmp")
+            if ssh_client.get_architecture() == "aarch64":
```
narrieta: Let's move this logic into download-pypy itself to avoid a call over SSH (decreases chances of failures)

nagworld9: We need to know pypy_name in exception case as well. You mean to say move it to downloa-pypy and also call in exception block as well?

narrieta: Yes. I see that uname was already in the previous version of download-pypy, so we would need to put it back and then move this one to the exception case

nagworld9: Done

PR #3435, opened 2025-08-18
RSM downgrade fix

nagworld9:
## Description

RSM sends additional flag (FromVersion) for rsm downgrade to determine if its real downgrade or not. Agent checks for FromVersion and if it matches with current version running, agent honor the downgrade.

Sample scenario which causes unintended downgrade before

1. Create VM. Agent uses v1 from PIR
2. New version published v2, RSM drives upgrades, and VM upgraded to v2
3. I month later, vm still on v2
4. New version published v3, RSM drives upgrades, and VM upgraded to v3
5. Found bug in v3, RSM drives downgrades, and VM downgrade to v2.
6. 1 month later, vm still on v2
7. 9 am: New version published v4 to PIR
8. VM service healed/re-image, got new disk- agent moved to v4. But RSM goal state was still on old one with v2.
9. v4 agent will do unintended downgrade to v2

Issue # 

---

### PR information


narrieta (dismissed):

azurelinuxagent/common/protocol/restapi.py
```diff
@@ -48,6 +48,8 @@ def __init__(self, name):
         self.name = name
         # Two-state: None, string. Set to None if version not specified in the GS
         self.version = None
+        #  Tri-state: None, True, False. Set to None if this property not specified in the GS.
```
maddieford: Is this comment correct? I would expect this to be two-state: None, string

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -48,6 +48,7 @@ class AgentUpdateHandler(object):
                     version: it will have what version to update
                     isVersionFromRSM: True if the version is from RSM deployment.
                     isVMEnabledForRSMUpgrades: True if the VM is enabled for RSM upgrades.
+                    fromVersion: This property specifies the version to update from. It is populated only for downgrade requests and subsequent goal states thereafter, until an upgrade request.
```
maddieford, line 51: Once an update request comes, what value will this property be in the goal state? None or not provided?

azurelinuxagent/common/protocol/restapi.py
nagworld9: Thanks for catching that, yes two states

azurelinuxagent/ga/agent_update_handler.py
nagworld9: not provided for upgrades

nagworld9: only present in downgrade requests

azurelinuxagent/ga/agent_update_handler.py
```diff
@@ -236,7 +237,7 @@ def get_vmagent_update_status(self):
                 else:
                     status = VMAgentUpdateStatuses.Error
                     code = 1
-                return VMAgentUpdateStatus(expected_version=str(self._updater.version), status=status, code=code, message=self._last_attempted_update_error_msg)
+                return VMAgentUpdateStatus(expected_version=str(CURRENT_VERSION), status=status, code=code, message=self._last_attempted_update_error_msg)
```
nagworld9, line 240: CRP expects current version running and status in the updatestatus report. The name expected_version confused me before, changing now to reflect as per protocol.

On the crp side, they check both version and status to verify the update operation status

azurelinuxagent/ga/rsm_version_updater.py
```diff
@@ -87,14 +87,23 @@ def retrieve_agent_version(self, agent_family, goal_state):
...
+        if not agent_family.is_version_from_rsm or self._version == CURRENT_VERSION:
             return False
 
+        # If the version is below daemon version or if it is a downgrade and the current agent version is not the one we are downgrading from, we don't allow update
+        elif self._version < self._daemon_version:
+            raise AgentUpdateError("Received invalid update request:{0}, new version {1} is below than daemon version {2}".format(
+                self._gs_id, str(self._version), str(self._daemon_version)))
+
+        elif self._version < CURRENT_VERSION and CURRENT_VERSION != FlexibleVersion(agent_family.from_version):
+            raise AgentUpdateError("Received invalid update request:{0}, downgrade {1} is not allowed to update from {2}. Current agent version running: {3}".format(
```
nagworld9, line 104: Report the error in the updatestatus. Caveat is that for non-rsm goal states we log and report same error too, which is ok to me as we handling the important thing stopping unintended downgrade.

maddieford: This is fine with me, but in get_vmagent_update_status() we have a comment which indicates "We report the status only when vm enrolled into RSM". 

Should we update that comment with this caveat?

azurelinuxagent/ga/agent_update_handler.py
narrieta, line 240: I'm a little concerned about the straight conversion of the FlexibleVersion to str (should probably be ustr, btw).

*Should* not be an issue for Agent versions, but in general, Flexible Version can have a pretty arbitrary format that CRP may not be able to parse correctly.

I would do an explicit conversion of the W.X.Y.Z components of the flexible version

narrieta: Never mind, the concern was about DistroVersion. In fact, previously I recommended (see distroversion.py)
`
 NOTE: For entities with simpler version schemes (e.g. extensions and the Agent), use FlexibleVersion.`

narrieta (approved):

azurelinuxagent/ga/rsm_version_updater.py
nagworld9: its still true we only reportstaus for vm enrolled to RSM but for all goal states at that point

azurelinuxagent/common/protocol/restapi.py
nagworld9: updated

azurelinuxagent/ga/rsm_version_updater.py
maddieford: Ah makes sense

maddieford (approved):

PR #3439, opened 2025-08-22
Set a timeout on vm deployment for end-to-end tests

narrieta:
Updating LISA to take advantage of the new provisioning_timeout parameter to set a timeout on VM provisioning.

I also added a new parameter to the Azure Pipeline to select the Docker container image for a test run. This is useful to point the test run to a new image while testing changes in the image setup.


tests_e2e/orchestrator/lib/agent_junit.py
```diff
@@ -33,21 +33,14 @@
 )
 
 
-@dataclass_json()
```
narrieta, line 36: This class needed to be kept in sync with JUnitSchema, or the AgentJUnit would break. JUnitSchema has additional properties on this LISA update and that broke test results reporting.

Taking a deeper look at the LISA source code, strictly speaking we don't need a custom AgentJUnitSchema and we can simply use JUnitSchema. Doing so, though, prevents us from customizing the file name for the junit report, but that is just a small inconvenience.

tests_e2e/orchestrator/lib/update_arm_template_hook.py
```diff
@@ -50,9 +50,6 @@ def azure_update_arm_template(self, template: Any, environment: Environment) ->
         #
         allow_ssh: str = vm_tags.get("allow_ssh")
         network_security_rule = NetworkSecurityRule(template, is_lisa_template=True)
-        # Disabling the default outbound access due to security requirement
```
narrieta, line 53: Now LISA disables the default outbound access, so we do not need to do it on our side.

tests_e2e/orchestrator/runbook.yml
```diff
@@ -82,6 +83,8 @@ variable:
     value: ""
   - name: vm_size
     value: ""
+  - name: vm_provisioning_timeout
```
narrieta, line 86: Starting with 15 minutes, but we can adjust as needed.

tests_e2e/pipeline/scripts/execute_tests.sh
```diff
@@ -2,6 +2,30 @@
 
 set -euxo pipefail
 
+#
```
narrieta, line 5: Moved this block to the top of the file and added CONTAINER_IMAGE

maddieford (approved):

nagworld9 (approved):

PR #3457, opened 2025-09-17
Remove references to conf that may be uninitialized

narrieta:
These 2 modules may use conf before it is initialized if they are imported before load_conf_from_file() is invoked.

### DIFF
--- a/azurelinuxagent/common/event.py
+++ b/azurelinuxagent/common/event.py
@@ -26,7 +26,6 @@ import time
 import traceback
 from datetime import datetime
 
-import azurelinuxagent.common.conf as conf
 import azurelinuxagent.common.logger as logger
 from azurelinuxagent.common.AgentGlobals import AgentGlobals
 from azurelinuxagent.common.exception import EventError, OSUtilError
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/event.py
```diff
@@ -176,7 +175,7 @@ def event_succeeded(self, name, version, op):
             return True
         return self._status[event] is True
 
-    def initialize(self, status_dir=conf.get_lib_dir()):
```
narrieta, line 179: the default value is never used

azurelinuxagent/ga/update.py
```diff
@@ -125,7 +125,6 @@ def get_update_handler():
 
 class UpdateHandler(object):
     TELEMETRY_HEARTBEAT_PERIOD = timedelta(minutes=30)
-    CHECK_MEMORY_USAGE_PERIOD = timedelta(seconds=conf.get_cgroup_check_period())
```
narrieta, line 128: this member is never used

maddieford (approved):

nagworld9 (approved):

PR #3398, opened 2025-06-04
Enforce extension signature validation with policy

mgunnala:
## Description

Issue # 

---
This PR allows enforcement of extension package signature validation using the "signatureRequired" option in the policy file. 

### PR information
- [ ] Ensure development PR is based on the `develop` branch.
- [ ] The title of the PR is clear and informative.
- [ ] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).
- [ ] If applicable, the PR references the bug/issue that it fixes in the description.
- [ ] New Unit tests were added for the changes made

### Quality of Code and Contribution Guidelines
- [ ] I have read the [contribution guidelines](https://github.com/Azure/WALinuxAgent/blob/master/.github/CONTRIBUTING.md).

PR #3458, opened 2025-09-17
Improve error handling in policy engine

mgunnala:
## Description

Issue # 

---

This PR updates error handling in policy_engine.py, based on a comment/TODO from previous PR: https://github.com/Azure/WALinuxAgent/pull/3398
- in __get_policy_enforcement_enabled(), if there is an error checking for the existence of the policy file, return True. 
- in __read_policy(), wrap the "with" block in a try/except, so that any errors opening the policy file can be caught and reported to the user appropriately. 

### PR information

### DIFF
--- a/azurelinuxagent/ga/policy/policy_engine.py
+++ b/azurelinuxagent/ga/policy/policy_engine.py
@@ -130,13 +130,19 @@ class _PolicyEngine(object):
             1. policy file exists at the expected location
             2. the conf flag "Debug.EnableExtensionPolicy" is set to True.
 
-        This method runs during policy engine initialization and should not raise errors. If an error occurs, return False.
+        This method runs during policy engine initialization and should not raise errors.
         """
         try:
-            return conf.get_extension_policy_enabled() and os.path.exists(conf.get_policy_file_path())
+            policy_file_exists = os.path.isfile(conf.get_policy_file_path())
         except Exception as ex:
-            logger.warn("Error checking if policy is enabled: {0}".format(ex))
-            return False
+            # os.path.isfile() should only raise an error if the customer-specified policy file path is invalid.
+            # In that case, we assume the customer intends to enable policy. Any error from the invalid
+            # path will be surfaced to the user when reading the file in update_policy().
+            _PolicyEngine._log_policy_event("Error checking if policy file exists: {0}".format(ex), is_success=False)
+            return True
+
+        return conf.get_extension_policy_enabled() and policy_file_exists
+
 
     @staticmethod
     def __read_policy(goal_state_history):
@@ -149,16 +155,15 @@ class _PolicyEngine(object):
         Raise InvalidPolicyError if JSON is invalid, or any exceptions are thrown while reading the file.
         """
         policy_file_contents = None
-        with open(conf.get_policy_file_path(), 'r') as f:
-            try:
+        try:
+            with open(conf.get_policy_file_path(), 'r') as f:
                 policy_file_contents = f.read()
 
                 # Save policy file contents to goal state history folder
-                if goal_state_history is not None:
-                    try:
-                        goal_state_history.save(policy_file_contents, "waagent_policy.json")
-                    except Exception as ex:
-                        logger.warn("Failed to save policy to history, continuing with goal state processing. Error: {0}", ustr(ex))
+                try:
+                    goal_state_history.save(policy_file_contents, "waagent_policy.json")
+                except Exception as ex:
+                    logger.warn("Failed to save policy to history, continuing with goal state processing. Error: {0}", ustr(ex))
 
                 _PolicyEngine._log_policy_event(
                     "Enforcing policy using policy file found at '{0}'.".format(conf.get_policy_file_path()))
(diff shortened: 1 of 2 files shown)


tests_e2e/tests/ext_cgroups/install_extensions.py
```diff
@@ -46,7 +46,7 @@ def run(self):
     def _install_ama(self):
         ama_extension = VirtualMachineExtensionClient(
             self._context.vm, VmExtensionIds.AzureMonitorLinuxAgent,
-            resource_name="AMAAgent")
+            resource_name="AzureMonitorLinuxAgent")
```
mgunnala: Note: 

The ExtPolicy test cleans up existing extensions on the VM, but it expects the resource name to match the extension type. For AMA, ExtPolicy removes "AzureMonitorLinuxAgent", but ExtCGroups leaves behind "AMAAgent". Since this is not cleaned up and AMA is not multi-config, the ExtPolicy test fails with a deployment error.

> '(BadRequest) Multiple VMExtensions per handler not supported for OS type 'Linux'. VMExtension 'AzureMonitorLinuxAgent' with handler 'Microsoft.Azure.Monitor.AzureMonitorLinuxAgent' already added or specified in input.
> Code: BadRequest
> Message: Multiple VMExtensions per handler not supported for OS type 'Linux'. VMExtension 'AzureMonitorLinuxAgent' with handler 'Microsoft.Azure.Monitor.AzureMonitorLinuxAgent' already added or specified in input.']

I am renaming this extension to "AzureMonitorLinuxAgent" (to match its type) so that ExtPolicy can clean it up correctly during the test.

narrieta: Seems like this change is unrelated to the changes in policy_engine.py.

I'd suggest submitting this on a different PR, so that we can do reviews faster and with less iterations.

Thanks

mgunnala: I've moved this change to another PR

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -130,13 +130,19 @@ def __get_policy_enforcement_enabled():
...
-        This method runs during policy engine initialization and should not raise errors. If an error occurs, return False.
+        This method runs during policy engine initialization and should not raise errors.
         """
         try:
-            return conf.get_extension_policy_enabled() and os.path.exists(conf.get_policy_file_path())
+            policy_file_exists = os.path.isfile(conf.get_policy_file_path())
         except Exception as ex:
-            logger.warn("Error checking if policy is enabled: {0}".format(ex))
-            return False
+            # os.path.isfile() should only raise an error if the customer-specified policy file path is invalid.
```
narrieta: It seems to me like an invalid policy file path should be a fatal error, instead of us making any assumptions.

But regardless of that, I seem to remember that the presence of a policy file to enable enforcement is only meant to be used for the private build (?). If this is correct, this change should be on the branch for that private build, not on the main develop branch.

mgunnala: The policy file will also be used for the telemetry release, to allow CVM users to opt-in to signature validation enforcement.

By fatal error, do you mean that an invalid policy file path should result in all extensions being blocked?

narrieta: > The policy file will also be used for the telemetry release, to allow CVM users to opt-in to signature validation enforcement.

OK

>By fatal error, do you mean that an invalid policy file path should result in all extensions being blocked?

Yes. We do the same for invalid file content, right?

mgunnala: With the current implementation, this does result in a fatal error. When update_policy() tries to read the invalid file path, an exception is thrown, causing all extensions to be blocked. We just don't want __get_policy_enforcement_enabled() to raise errors since it runs during policy engine initialization.

An alternative approach would be to validate the path earlier in conf.py and raise an AgentConfigError if the path is invalid.

Tagging @maddieford for feedback as well.

maddieford: What is meant by 'invalid policy file path' in this discussion? Does that just mean the file doesn't exist? In that case, I do not think the file not existing should block extensions

narrieta: @maddieford 

I am assuming an actually invalid file path. Non-existing would return False rather than raising an exception.

Some python versions can raise

>Changed in version 3.8: [exists()](https://docs.python.org/3/library/os.path.html#os.path.exists), [lexists()](https://docs.python.org/3/library/os.path.html#os.path.lexists), [isdir()](https://docs.python.org/3/library/os.path.html#os.path.isdir), [isfile()](https://docs.python.org/3/library/os.path.html#os.path.isfile), [islink()](https://docs.python.org/3/library/os.path.html#os.path.islink), and [ismount()](https://docs.python.org/3/library/os.path.html#os.path.ismount) now return False instead of raising an exception for paths that contain characters or bytes unrepresentable at the OS level.

mgunnala: Correct - I mean an invalid path that would raise an error, not a non-existing file.

narrieta: OK, we all agree on that...

So then, why the difference? If the Agent is given invalid json in the policy file, it prevents extensions from running (it fails on their behalf). But if the Agent is given an invalid path/file name, we want to assume policy enforcement is disabled?

mgunnala: There is no difference. If the agent is given an invalid file path, we assume policy enforcement is enabled here and raise the error in update_policy() so extensions are blocked.

narrieta: That is not a good approach. __get_policy_enforcement_enabled is detecting the error, ignoring it, and assuming somebody else will detect and report it at some point

mgunnala: Removed this error handling as discussed offline

PR #3459, opened 2025-09-17
Update test extension name in ExtCgroups

mgunnala:
## Description

Issue # 

---

The ExtPolicy test suite cleans up existing extensions that it will test on the VM, but it expects the resource name to match the extension type for these leftover extensions.

For the AzureMonitorLinuxAgent extension, ExtPolicy removes "AzureMonitorLinuxAgent". But the ExtCGroups test suite leaves behind "AMAAgent" on the same VM. Since the AzureMonitorLinuxAgent extension is not cleaned up properly and it is not a multi-config extension, the ExtPolicy test fails with a deployment error.

> '(BadRequest) Multiple VMExtensions per handler not supported for OS type 'Linux'. VMExtension 'AzureMonitorLinuxAgent' with handler 'Microsoft.Azure.Monitor.AzureMonitorLinuxAgent' already added or specified in input.
> Code: BadRequest
> Message: Multiple VMExtensions per handler not supported for OS type 'Linux'. VMExtension 'AzureMonitorLinuxAgent' with handler 'Microsoft.Azure.Monitor.AzureMonitorLinuxAgent' already added or specified in input.']

This PR renames the extension in ExtCGroups to "AzureMonitorLinuxAgent" (to match its type) so that ExtPolicy can clean it up correctly.

### PR information

### DIFF
--- a/tests_e2e/tests/ext_cgroups/install_extensions.py
+++ b/tests_e2e/tests/ext_cgroups/install_extensions.py
@@ -46,7 +46,7 @@ class InstallExtensions:
     def _install_ama(self):
         ama_extension = VirtualMachineExtensionClient(
             self._context.vm, VmExtensionIds.AzureMonitorLinuxAgent,
-            resource_name="AMAAgent")
+            resource_name="AzureMonitorLinuxAgent")
         log.info("Installing %s", ama_extension)
         ama_extension.enable()
         ama_extension.assert_instance_view()


tests_e2e/tests/ext_cgroups/install_extensions.py
```diff
@@ -46,7 +46,7 @@ def run(self):
     def _install_ama(self):
         ama_extension = VirtualMachineExtensionClient(
             self._context.vm, VmExtensionIds.AzureMonitorLinuxAgent,
-            resource_name="AMAAgent")
+            resource_name="AzureMonitorLinuxAgent")
```
narrieta, line 49: Thanks for this change, but this specific failure only points to a larger issue: since we are sharing VMs across tests, we need to standardize on the resource names used to install extensions. Please bring this up in the next team sync.

An idea that comes to mind is to add a 'PreferredResourceName' property to VmExtensionIdentifier and use that all across the code, but the entire team needs to be in sync with this (or any other approach we may use).

PR #3463, opened 2025-09-23
Enable FIPS test on AzureUSGovernment cloud

narrieta:
Added the cloud as a prefix to skip_on_images in the test suite configuration. This allowed me to skip the Ubuntu images not available in the gov cloud, and enable the FIPS test on that cloud.

### DIFF
--- a/tests_e2e/orchestrator/lib/agent_test_loader.py
+++ b/tests_e2e/orchestrator/lib/agent_test_loader.py
@@ -176,16 +176,27 @@ class AgentTestLoader(object):
                 if image not in self.images:
                     raise Exception(f"Invalid image reference in test suite {suite.name}: Can't find {image} in images.yml or image from a shared gallery")
 
-            # If the suite specifies a cloud and it's location<cloud:location>, validate that location string is start with <cloud:> and then validate that the images it uses are available in that location
+            valid_clouds = ["AzureCloud", "AzureChinaCloud", "AzureUSGovernment"]
+
+            def split_on_cloud(value):
+                split = value.split(":")
+                if len(split) == 2:
+                    if split[0] not in valid_clouds:
+                        raise Exception(f"{value} does not contain a valid cloud: {split[0]}")
+                    return split
+                if len(split) == 1:
+                    return "", split[0]
+                raise Exception(f"{value} is not a valid value in the test configuration")
+
+            # If the test suite specifies any locations, validate that the images it uses are available in those locations
             for suite_location in suite.locations:
-                if suite_location.startswith(self.__cloud + ":"):
-                    suite_location = suite_location.split(":")[1]
-                else:
+                cloud, suite_location = split_on_cloud(suite_location)
+                if self.__cloud != cloud:
                     continue
                 for suite_image in suite.images:
                     suite_image = _parse_image(suite_image)
-                    # skip validation if suite image from gallery image
-                    if CustomImage._is_image_from_gallery(suite_image):
+                    # skip validation if suite image is a gallery image or is in 'skip_images'
+                    if CustomImage._is_image_from_gallery(suite_image) or f"{self.__cloud}:{suite_image}" in suite.skip_on_images:
                         continue
                     for image in self.images[suite_image]:
                         # If the image has a location restriction, validate that it is available on the location the suite must run on
@@ -196,11 +207,12 @@ class AgentTestLoader(object):
 
             # if the suite specifies skip clouds, validate that cloud used in our tests
             for suite_skip_cloud in suite.skip_on_clouds:
-                if suite_skip_cloud not in ["AzureCloud", "AzureChinaCloud", "AzureUSGovernment"]:
+                if suite_skip_cloud not in valid_clouds:
                     raise Exception(f"Invalid cloud {suite_skip_cloud} for in {suite.name}")
 
-            # if the suite specifies skip images, validate that images used in our tests
+            # if the suite specifies skip_on_images, validate that the images are valid
             for suite_skip_image in suite.skip_on_images:
+                _, suite_skip_image = split_on_cloud(suite_skip_image)
                 if suite_skip_image not in self.images:
                     raise Exception(f"Invalid image reference in test suite {suite.name}: Can't find {suite_skip_image} in images.yml")
 
--- a/tests_e2e/orchestrator/lib/agent_test_loader.py
+++ b/tests_e2e/orchestrator/lib/agent_test_loader.py
@@ -23,6 +23,7 @@ from pathlib import Path
 from typing import Any, Dict, List, Type
 
 import tests_e2e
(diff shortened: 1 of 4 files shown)


tests_e2e/orchestrator/lib/agent_test_loader.py
```diff
@@ -166,6 +166,8 @@ def _parse_image(image: str) -> str:
                 return match.group('image_set')
             return image
 
+        valid_clouds = ["AzureCloud", "AzureChinaCloud", "AzureUSGovernment"]
```
maddieford: Can we use AZURE_CLOUDS dict (tests_e2e/tests/lib/azure_clouds.py) instead of creating new list here?

tests_e2e/orchestrator/lib/agent_test_loader.py
```diff
@@ -176,16 +178,25 @@ def _parse_image(image: str) -> str:
...
+                    return "", split[0]
+                raise Exception(f"{value} is not a valid value in the test configuration")
+
+            # If the test suite specifies any locations, validate that the images it uses are available in those locations
             for suite_location in suite.locations:
-                if suite_location.startswith(self.__cloud + ":"):
-                    suite_location = suite_location.split(":")[1]
-                else:
+                cloud, suite_location = split_on_cloud(suite_location)
+                if self.__cloud != cloud:
```
maddieford: Shouldn't we do the validation even if the following condition is true
`cloud == ""`
?

tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
```diff
@@ -497,15 +497,20 @@ def _get_test_suite_images(suite: TestSuiteInfo, loader: AgentTestLoader) -> Lis
...
 
         A test suite may reference multiple image sets and sets can intersect; this method eliminates any duplicates.
         """
         skip_unique: Dict[str, VmImageInfo] = {}
         for image in suite.skip_on_images:
+            cloud = ""
+            split = image.split(':')
+            if len(split) == 2:
+                cloud, image = split
+            if self.runbook.cloud != cloud:
```
maddieford: cloud will be "" if the image should be skipped on all clouds. It looks like we aren't handling that case

tests_e2e/test_suites/fips.yml
```diff
@@ -19,17 +19,16 @@ images:
...
 locations:
   - "AzureCloud:eastus2euap"
-  # - "AzureUSGovernment:usdodcentral"
+  - "AzureUSGovernment:usdodcentral"
 
 skip_on_clouds:
   - "AzureChinaCloud"
-  - "AzureUSGovernment"
+
+skip_on_images:
```
maddieford, line 31: Can we add this comment:
`# The Ubuntu images are not available on AzureUSGovernment, so they are skipped.`

tests_e2e/orchestrator/lib/agent_test_loader.py
```diff
@@ -176,16 +178,25 @@ def _parse_image(image: str) -> str:
...
                     continue
                 for suite_image in suite.images:
                     suite_image = _parse_image(suite_image)
-                    # skip validation if suite image from gallery image
-                    if CustomImage._is_image_from_gallery(suite_image):
+                    # skip validation if suite image is a gallery image or is in 'skip_images'
+                    if CustomImage._is_image_from_gallery(suite_image) or f"{self.__cloud}:{suite_image}" in suite.skip_on_images:
```
nagworld9: do expect cloud name in suite.skip_on_images?

tests_e2e/orchestrator/lib/agent_test_loader.py
narrieta: thanks; fixed

tests_e2e/orchestrator/lib/agent_test_loader.py
narrieta: thanks; fixed

tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
narrieta: whoa! thanks; fixed

tests_e2e/test_suites/fips.yml
narrieta: thanks; fixed

tests_e2e/orchestrator/lib/agent_test_loader.py
narrieta: thanks; fixed

tests_e2e/orchestrator/lib/agent_test_suite_combinator.py
```diff
@@ -253,6 +253,12 @@ def create_environment_list(self, test_suites: List[str]) -> List[Dict[str, Any]
 
         environments.extend(shared_environments.values())
 
+        if len(skip_test_suites) > 0:
+            self._log.info("Skipping test suites %s", skip_test_suites)
+
+        if len(skip_test_suites_images) > 0:
+            self._log.info("Skipping test suits run on images \n %s", '\n'.join([f"\t{skip}" for skip in skip_test_suites_images]))
+
```
narrieta, line 270: I moved this code here, just before we raise the exception, since these messages can help understand why the exception was raised.

tests_e2e/orchestrator/lib/agent_test_loader.py
```diff
@@ -176,16 +177,25 @@ def _parse_image(image: str) -> str:
...
-                else:
+                cloud, suite_location = split_on_cloud(suite_location)
+                if cloud != "" and cloud != self.__cloud:
                     continue
                 for suite_image in suite.images:
                     suite_image = _parse_image(suite_image)
-                    # skip validation if suite image from gallery image
-                    if CustomImage._is_image_from_gallery(suite_image):
+                    # skip validation if suite image is a gallery image or is in 'skip_images'
+                    if CustomImage._is_image_from_gallery(suite_image) or suite_image in suite.skip_on_images or f"{self.__cloud}:{suite_image}" in suite.skip_on_images:
```
nagworld9, line 198: images:
ubuntu_2004
rhel_9
suse_15

skip_on_images:
Usgov:suse_15

For public run, we shouldn't skip validation for suse_15. It should only happen for usgov run. Isn't it?

narrieta: Yes, same case as we have for ubuntu, which we skip for gov, but not for public

  - "AzureUSGovernment:ubuntu_pro_2204"
  - "AzureUSGovernment:ubuntu_pro_fips_2204"

nagworld9: NM, somehow I was thinking `skip_on_images` is string but its List

nagworld9 (approved):

maddieford (approved):

PR #3464, opened 2025-09-25
Add skipped tests to test results

narrieta:
When we skip a test due to restrictions on clouds or images, we log an INFO message, but it is hard to keep track of all the skipped tests.

This PR adds those tests as SKIPPED test results, and they will show up on the test run results.


tests_e2e/orchestrator/lib/agent_test_result.py
```diff
@@ -0,0 +1,65 @@
+# Microsoft Azure Linux Agent
```
narrieta, line 1: Refactored out from agent_test_suite.py into its own module. No changes.

nagworld9 (approved):

PR #3470, opened 2025-10-08
Do no fetch entire goal state when initializing telemetry

narrieta:
When initializing telemetry on the Daemon process, it is not necessary to fetch the entire goal state. Fetching the entire goal state can lead to issues if, for example, there are blocking issues downloading certificates.

### DIFF
--- a/azurelinuxagent/common/future.py
+++ b/azurelinuxagent/common/future.py
@@ -85,11 +85,11 @@ else:
     from datetime import tzinfo, timedelta
 
     class _UTC(tzinfo):
-        def utcoffset(self, dt):
+        def utcoffset(self, _):
             return timedelta(0)
-        def tzname(self, dt):
+        def tzname(self, _):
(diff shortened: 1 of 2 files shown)


azurelinuxagent/common/protocol/wire.py
```diff
@@ -852,26 +849,6 @@ def get_goal_state(self):
             raise ProtocolError("Trying to fetch goal state before initialization!")
         return self._goal_state
 
-    def get_hosting_env(self):
```
narrieta, line 855: We need to detach the goal state from the WireClient/WireProtocol, since ProtocolUtil is a singleton, which basically makes WireClient._goal_state a global. Different callers may need different properties from the goal state, and if a caller updates the goal state, all other callers are affected.

For now, I removed the usage of these helper functions, but the real work is to remove WireClient._goal_state altogether.

PR #3473, opened 2025-10-15
Skipping test validation when network not ready on startup

nagworld9:
## Description

During the boot we check for network ready to verify our custom service did the firewall setup. Occasionally we get network not ready, in that case we want to skip test validation as next checks not relevant

Issue # 

---

### PR information

### DIFF
--- a/azurelinuxagent/ga/signing_certificate_util.py
+++ b/azurelinuxagent/ga/signing_certificate_util.py
@@ -73,10 +73,10 @@ def _write_certificate(cert_string, output_path):
         umask = os.umask(0o077)
         with open(output_path, "w") as cert_file:
             cert_file.write(cert_string)
-        logger.info("Signing certificate written to {0}".format(output_path))
+        logger.info("Signing certificate used for extension signature validation has been written to {0}".format(output_path))
 
     except Exception as err:
-        msg = "Failed to write signing certificate to file ('{0}'). Error details:\n{1}".format(output_path, err)
+        msg = "Failed to write extension signing certificate to file ('{0}'). Error details:\n{1}".format(output_path, err)
         event.error(op=WALAEventOperation.SignatureValidation, fmt=msg)
 
     finally:
--- a/tests_e2e/pipeline/pipeline.yml
+++ b/tests_e2e/pipeline/pipeline.yml
@@ -48,11 +48,6 @@ parameters:
     - failed
     - no
 
-  - name: collect_lisa_logs
-    displayName: Collect LISA logs
-    type: boolean
-    default: false
-
   - name: keep_environment
     displayName: Keep the test VMs (do not delete them)
     type: string
@@ -143,8 +138,6 @@ jobs:
         displayName: "Collect test artifacts"
         # Collect artifacts even if the previous step is cancelled (e.g. timeout)
         condition: always()
-        env:
-          COLLECT_LISA_LOGS: ${{ parameters.collect_lisa_logs }}
 
       - publish: $(Build.ArtifactStagingDirectory)
         artifact: 'artifacts'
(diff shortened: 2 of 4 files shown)


tests_e2e/tests/lib/remote_test.py
```diff
@@ -23,10 +23,17 @@
 from tests_e2e.tests.lib.logging import log
 
 SUCCESS_EXIT_CODE = 0
+SKIP_EXIT_CODE = 50
 FAIL_EXIT_CODE = 100
 ERROR_EXIT_CODE = 200
 
 
+class RemoteTestSkipped(Exception):
```
narrieta: Why do we need to distinguish remote tests?

nagworld9: I had circular-import issues using TestSkipped. So I created a workaround and add remote name to use it for remote tests but now that I think about it, I can move TestSkipped to here or new module if that makes sense?

narrieta: yes, it does. thanks!

tests_e2e/tests/lib/agent_test.py
```diff
@@ -28,25 +28,12 @@
...
 from tests_e2e.tests.lib.agent_test_context import AgentTestContext, AgentVmTestContext, AgentVmssTestContext
+from tests_e2e.tests.lib.exception import RemoteTestError, TestSkipped, CommandError
 from tests_e2e.tests.lib.logging import log
-from tests_e2e.tests.lib.remote_test import FAIL_EXIT_CODE
-from tests_e2e.tests.lib.shell import CommandError
+from tests_e2e.tests.lib.remote_test import FAIL_EXIT_CODE, SKIP_EXIT_CODE
 from tests_e2e.tests.lib.ssh_client import ATTEMPTS, ATTEMPT_DELAY, SshClient
 
 
-class TestSkipped(Exception):
```
nagworld9, line 37: Moved to own module

tests_e2e/tests/lib/exception.py
```diff
@@ -0,0 +1,28 @@
+from typing import Any
```
narrieta: We shouldn't have an exception module with all exceptions. Exceptions should be defined in the module that raises them, does CommandError need to be here?

If you need to refactor out the other 2 execeptions to a separate module, I'd suggest naming it "TestResult" or something similar.

nagworld9: ` Exceptions should be defined in the module that raises them`

what does this mean? If I raise same exception in multiple modules, how this holds?

narrieta: if it is raised by multiple modules, it can be refactored to a separate file.

why does CommandError need to be moved?

nagworld9: I thought of circular-import issue at some point and moved here; I'll check and keep it in original place if its not craeaing issue now

tests_e2e/tests/lib/agent_test.py
```diff
@@ -28,25 +28,13 @@
 from azurelinuxagent.common.future import datetime_min_utc
 
 from tests_e2e.tests.lib.agent_test_context import AgentTestContext, AgentVmTestContext, AgentVmssTestContext
+from tests_e2e.tests.lib.test_result import RemoteTestError, TestSkipped
 from tests_e2e.tests.lib.logging import log
-from tests_e2e.tests.lib.remote_test import FAIL_EXIT_CODE
+from tests_e2e.tests.lib.remote_test import FAIL_EXIT_CODE, SKIP_EXIT_CODE
```
narrieta: To complete this refactoring, we should remove the dependency of agent_test on remote_test. We should move the exit codes to test_result (and prefix them with REMOTE_TEST, or define them as a RemoteTestExitCode enum) and now agent_test and remote_test would depend on test_result, but not on each other.

tests_e2e/tests/lib/test_result.py
```diff
@@ -18,6 +18,13 @@
 from tests_e2e.tests.lib.shell import CommandError
 
 
+class RemoteTestExitCode(object):
+    SUCCESS_EXIT_CODE = 0
```
narrieta: We should remove the "_EXIT_CODE" suffix, since the enum name already has it

nagworld9: updated

PR #3474, opened 2025-10-16
E2E test: Do not report exceptions in extensions

narrieta:
Exceptions in extensions were being reported as agent failures by check-agent-log.

### DIFF
--- a/tests_e2e/orchestrator/runbook.yml
+++ b/tests_e2e/orchestrator/runbook.yml
@@ -57,10 +57,7 @@ variable:
       - publish_hostname
       - recover_network_interface
       - log_collector
-#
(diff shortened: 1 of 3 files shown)


tests_e2e/tests/lib/agent_log.py
```diff
@@ -494,7 +494,19 @@ def get_errors(self) -> List[AgentLogRecord]:
         ]
 
         def is_error(r: AgentLogRecord) -> bool:
-            return r.level in ('ERROR', 'WARNING') or any(err in r.text for err in ['Exception', 'Traceback', '[CGW]'])
```
narrieta, line 497: this was checking r.text, should be r.message

nagworld9 (approved):

PR #3475, opened 2025-10-17
collect lisa logs by default

nagworld9:
## Description

Collect lisa logs by default to get serial logs for failed tests. This log useful for test cases where test performs vm boot and vm not started after

Issue # 

---

### PR information


tests_e2e/pipeline/pipeline.yml
```diff
@@ -51,7 +51,7 @@ parameters:
   - name: collect_lisa_logs
     displayName: Collect LISA logs
     type: boolean
-    default: false
+    default: true
```
narrieta: We should probably remove this option altogether. Initially I added it in case somebody needed the LISA logs, but seems it seems like it is good to have them, I don't see the need to have an option to turn this off.

nagworld9: you mean remove pipeline parameter altogether?

narrieta: yes, i don't think it is useful

we can keep it as well, if you prefer so

narrieta (approved):

maddieford (approved):

PR #3484, opened 2025-10-30
Do not fetch full goal state on Daemon process + Do no send provisioning signals from CloudInitProvisionHandler

narrieta:
Currently the Daemon process fetches the entire goal state, while it only needs a few properties from it. If an error occurs fetching the properties that are not actually needed (e.g. a decryption error on the certificates download), then then Daemon goes into an infinite loop.

Also, currently the CloudInitProvisionHandler retrieves the goal state in order to send the provisioning signals. Those signals are already sent by cloud-init, so this is not needed. Furthermore, sending those signals in the WALinuxAgent and cloud-init can lead to inconsistent signals (and inconsistent VM provisioning state), since both services run asynchronously and can send different statuses. In addition, retrieving the goal state on the CloudInitProvisionHandler can lead to an infinite loop if there is a failure.

This PR changes that behavior to retrieve only the properties of the goal state that are actually needed by the Daemon process, ignores errors fetching the goal state when they do not affect the functionality of the WALinuxAgent (e.g. initialize telemetry), and removed the provisioning signals emitted by the CloudInitProvisionHandler.

### DIFF
--- a/tests_e2e/tests/hibernation/hibernation.py
+++ b/tests_e2e/tests/hibernation/hibernation.py
@@ -82,16 +82,24 @@ class Hibernation(AgentVmTest):
         log.info("Tenant certificate: %s", pre_hibernation_tenant_certificate)
         log.info("")
 
-        log.info("Triggering a hibernate-resume cycle to test the tenant certificate...")
-        hibernate_time = self._ssh_client.get_time()
-        self._do_hibernate_resume_cycle()
-        log.info("")
+        # Azure policy may install an extension while the machine is resuming, which would increment the incarnation, so try a couple of times
+        for i in range(2):
+            log.info("Triggering a hibernate-resume cycle to test the tenant certificate...")
+            hibernate_time = self._ssh_client.get_time()
+            self._do_hibernate_resume_cycle()
+            log.info("")
 
-        log.info("Verifying that the current incarnation is 1 after resume...")
-        incarnation = self._ssh_client.run_command("get_goal_state.py --tag Incarnation", use_sudo=True).rstrip()
-        if incarnation != "1":
-            raise Exception(f"Unexpected behavior: The incarnation is not 1 after a hibernate-resume cycle. Incarnation is {incarnation}")
-        log.info("The current incarnation is 1")
+            log.info("Verifying that the current incarnation is 1 after resume...")
+            incarnation = self._ssh_client.run_command("get_goal_state.py --tag Incarnation", use_sudo=True).rstrip()
+            if incarnation == "1":
+                log.info("The current incarnation is 1")
+                break
+            if i > 0:
+                raise Exception(f"Unexpected behavior: The incarnation is not 1 after a couple of hibernate-resume cycles. Incarnation is {incarnation}")
+            log.info(f"The current incarnation is not 1, as was expected. A new goal state was generated after resuming the VM. An extension installed by policy may have been the reason. Incarnation: {incarnation}")
+            log.info("Will retry the hibernate-resume cycle to reset it to 1...")
+            log.info("")
+
         log.info("")
 
         log.info("Checking tenant certificate after resume...")


azurelinuxagent/common/event.py
```diff
@@ -89,6 +89,7 @@ class WALAEventOperation:
     ExtensionHandlerManifest = "ExtensionHandlerManifest"
     ExtensionPolicy = "ExtensionPolicy"                                 # Event for any extension policy-related operations (e.g., extension not in allowlist).
     ExtensionSignaturePolicy = "ExtensionSignaturePolicy"               # Event for unsigned extension blocked due to extension signature policy.
+    ExtensionSigned = "ExtensionSigned"                                 # Event indicating whether an extension is signed.
```
narrieta, line 92: Alphabetized the list of telemetry events. I was looking for a particular event and the list was hard to read since it was not in alpabetical order.

azurelinuxagent/daemon/main.py
```diff
@@ -132,19 +132,38 @@ def daemon(self, child_args=None):
...
+        # detection, but in the meanwhile we keep the approach of forcing protocol detection during initialization.
+        #
         protocol_util = get_protocol_util()
         protocol_util.clear_protocol()
 
-        self.provision_handler.run()
-
-        # Create an instance of the protocol without initializing the goal state
-        protocol = protocol_util.get_protocol(init_goal_state=False)
+        #
```
narrieta, line 147: I moved the telemetry initialization before the call to self.provision_handler.run(), so now we initialize correctly any telemetry events produced by that method. Also, I am allowing the Daemon process to continue execution if there are any errors initializing telemetry: a few telemetry events will be lost, but the Agent (or at least the Daemon process) will continue functioning correctly.

tests/daemon/test_daemon.py
```diff
@@ -117,27 +117,28 @@ def stop_daemon(child_args):  # pylint: disable=unused-argument
...
-            daemon.start()
-            daemon.join(timeout=5)
-
-            self.assertTrue(daemon.is_alive())
-            daemon.terminate()
-
-            # disable_agent was written, run_latest was not called
-            self.assertTrue(os.path.exists(conf.get_disable_agent_file_path()))
-            self.assertEqual(0, patch_run_latest.call_count)
+        with patch('azurelinuxagent.daemon.main.get_protocol_util'):
```
narrieta, line 124: Since now we initialize telemetry before invoking ProvisionHandler.run, we need to mock both the protocol (since it is used to fetch the goal state) and the telemetry initialization (since it fetches the goal state).

azurelinuxagent/daemon/main.py
```diff
@@ -153,10 +172,7 @@ def daemon(self, child_args=None):
 
             logger.info("RDMA capabilities are enabled in configuration")
             try:
-                # Ensure the most recent SharedConfig is available
-                # - Changes to RDMA state may not increment the goal state
-                #   incarnation number. A forced update ensures the most
-                #   current values.
+                protocol = protocol_util.get_protocol(init_goal_state=False)
```
nagworld9, line 175: can we simply a bit? fetch protocol again if protocol is None previously? This will avoid duplicate calls as well.

narrieta: this was intentional, i want to minimize changes on code we do not test in the automated runs

narrieta: plus creating another instance of the protocol does not do any network requests, as long as the previous instance was created successfully. second call simply checks the existence of the protocol file

nagworld9: It seems this call was not there in the rdma block. We added as new code. If you want to minimize the changes, this is not doing that.

narrieta: it is needed here in case the first initialization fails... since initializing the telemetry fields is not critical, i catch the exception, log it and move on

azurelinuxagent/pa/provision/default.py
```diff
@@ -78,13 +77,12 @@ def run(self):
             logger.info("Copying ovf-env.xml")
             ovf_env = self.protocol_util.copy_ovf_env()
 
-            self.protocol_util.get_protocol()  # Trigger protocol detection
```
maddieford, line 81: Why are we able to remove this protocol detection? Don't we need it for the reporting below

narrieta: It'll be done when the goal state is actually needed (the call to get_protocol will trigger it)

For example:

```
    def report_not_ready(self, sub_status, description):
        status = ProvisionStatus(status="NotReady", subStatus=sub_status,
                                 description=description)
        try:
            protocol = self.protocol_util.get_protocol()
            protocol.report_provision_status(status)
        except ProtocolError as e:
            logger.error("Reporting NotReady failed: {0}", e)
            self.report_event(ustr(e))
```

(WireProtocol.report_provision_status fetches the goal state)

```
    def report_provision_status(self, provision_status):
        validate_param("provision_status", provision_status, ProvisionStatus)

        goal_state = GoalState(self.client, goal_state_properties=GoalStateProperties.RoleConfig)
```

azurelinuxagent/daemon/main.py
nagworld9: `it is needed here in case the first initialization fails.`

But code looks like it's always needed here regardless of previous initialization, I l know it will not do steps of wire server discovery and all that if we do protocol initialization before, it will be kind of no-op but that someone know after dig into protocol initialization logic. I'm not pushing it for a change, it's just thought and I leave it for your judgement.

narrieta: I don't think I understand your question/concern. Maybe you can clarify offline?

Previously this line was not here because protocol was always initialized. Now it is needed here before protocol can be None if the first initialization fails.

nagworld9: I'm not talking why new code added initialization is here and my point is why can't you use None check and do the initialization, why it has to be call initialize anyway. If you don't feel like adding it None check, I'm good with it as long as protocol initialization protecting it from not calling all those IP discovery and other steps.

narrieta: Additional check on None is not strictly needed, since this is a no-op. Again, I did minimal changes (avoiding even an extra check on None) on code we do not test.

PR #3486, opened 2025-11-03
Restrict extension signature validation to CVMs

mgunnala:
## Description

Issue # 

---

This PR adds an IMDS check for VM security type, and restricts extension signature validation to confidential VMs only.

### PR information

### DIFF
--- a/azurelinuxagent/ga/confidential_vm_info.py
+++ b/azurelinuxagent/ga/confidential_vm_info.py
@@ -42,7 +42,10 @@ class ConfidentialVMInfo(object):
     #
     # TODO: Remove once extension signature validation is supported on all VMs.
 
-    # Tri-state boolean: True if CVM, False if not a CVM, None if not yet initialized (fetch_and_initialize_cvm_info has not been called).
+    # Tri-state boolean:
+    # - True if CVM
+    # - False if not a CVM or unable to determine security type (e.g., IMDS call fails)
+    # - None if not yet initialized (fetch_and_initialize_cvm_info has not been called)
     _is_confidential_vm = None
 
     @staticmethod
--- a/tests/ga/test_confidential_vm_info.py
+++ b/tests/ga/test_confidential_vm_info.py
@@ -86,3 +86,12 @@ class TestConfidentialVMInfo(AgentTestCase):
 
             # Verify IMDS was only called once
             self.assertEqual(mock_get_metadata.call_count, 1)
+
+    def test_should_raise_error_when_cvm_info_uninitialized(self):
+        # Reset to uninitialized state
+        ConfidentialVMInfo._is_confidential_vm = None
+
+        # Calling is_confidential_vm() before initialization should raise RuntimeError
+        with self.assertRaises(RuntimeError) as context:
+            ConfidentialVMInfo.is_confidential_vm()
+            self.assertIn("not initialized", str(context.exception))


azurelinuxagent/ga/update.py
```diff
@@ -399,6 +401,9 @@ def run(self, debug=False):
             self._cleanup_legacy_goal_state_history()
             write_signing_certificates()
 
+            # Check security type and update if VM is a confidential VM. This should be done on each agent start.
+            ConfidentialVMInfo.set_security_type()
```
narrieta: This needs to happen before we retrieve the first goal state, since the behavior for that also needs to be aware of whether the VM is a CVM or not

azurelinuxagent/ga/confidential_vm_util.py
```diff
@@ -0,0 +1,46 @@
...
+
+class ConfidentialVMInfo(object):
+    _security_type = None
+
+    @staticmethod
+    def is_confidential_vm():
+        return ConfidentialVMInfo._security_type == SecurityType.ConfidentialVM
+
+    @staticmethod
+    def set_security_type():
```
narrieta: rename to 'initialize' or similar

azurelinuxagent/ga/confidential_vm_util.py
```diff
@@ -0,0 +1,46 @@
...
+    TrustedVM = "TrustedLaunch"
+    ConfidentialVM = "ConfidentialVM"
+
+
+class ConfidentialVMInfo(object):
+    _security_type = None
+
+    @staticmethod
+    def is_confidential_vm():
+        return ConfidentialVMInfo._security_type == SecurityType.ConfidentialVM
```
narrieta: raise if not initialized

azurelinuxagent/ga/confidential_vm_util.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @staticmethod
+    def set_security_type():
+        try:
+            compute_info = get_imds_client().get_compute()
+            security_type = compute_info.securityProfile.get('securityType')
+            logger.info("VM security type: {0}".format(security_type))
```
narrieta: send telemetry

azurelinuxagent/ga/confidential_vm_info.py
```diff
@@ -0,0 +1,46 @@
...
+#
+
+from azurelinuxagent.common.protocol.imds import get_imds_client
+from azurelinuxagent.common import logger
+from azurelinuxagent.common import event
+from azurelinuxagent.common.future import ustr
+
+
+class SecurityType(object):
+    TrustedVM = "TrustedLaunch"
```
narrieta: document where those are set on the vm model. will help readers

azurelinuxagent/ga/confidential_vm_info.py
```diff
@@ -0,0 +1,46 @@
...
+# limitations under the License.
+#
+# Requires Python 2.6+ and Openssl 1.0+
+#
+
+from azurelinuxagent.common.protocol.imds import get_imds_client
+from azurelinuxagent.common import logger
+from azurelinuxagent.common import event
+from azurelinuxagent.common.future import ustr
+
```
narrieta, line 29: Please add a comment that this is just temporary (telemetry release?) 

It will help in order to review some behavior, like how should be handle errors trying to reach IMDS

azurelinuxagent/ga/confidential_vm_util.py
```diff
@@ -0,0 +1,46 @@
...
+            ConfidentialVMInfo._security_type = security_type
+        except Exception as ex:
+            event.warn("Failed to get virtual machine security type from IMDS: {0}", ustr(ex))
```
narrieta: TODO (note to myself): need to review this error handling is appropriate

azurelinuxagent/ga/confidential_vm_info.py
```diff
@@ -0,0 +1,46 @@
+# Windows Azure Linux Agent
```
narrieta, line 1: Rename file to confidential_vm_info.py

azurelinuxagent/common/protocol/imds.py
```diff
@@ -190,38 +190,86 @@ class ComputeInfo(DataContract):
     __matcher = ImageInfoMatcher(ENDORSED_IMAGE_INFO_MATCHER_JSON)
 
     def __init__(self,
+                 additionalCapabilities=None,
```
narrieta: why do we need all those uninitialized properties?

azurelinuxagent/common/protocol/imds.py
```diff
@@ -12,7 +12,7 @@
 from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
 
 IMDS_ENDPOINT = '169.254.169.254'
-APIVERSION = '2018-02-01'
+APIVERSION = '2025-04-07'
```
narrieta: Is this API version deployed to every single host on all cloud where the agent runs?

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -279,13 +280,15 @@ def update(self, force_update=False, silent=False):
...
         # For each extension in the goal state being executed, we emit telemetry to indicate whether a signature is present
         # for the extension. The "is_success" field reflects whether the extension was signed.
-        # If signature is missing, skip telemetry in the following cases:
-        #   - Extension requested state is 'uninstall' (uninstall goal states never include signature).
-        #   - The goal state API does not support the 'encoded_signature' property (e.g., fast track goal states where HGAP version does not support signature).
-        for ext in self._extensions_goal_state.extensions:
-            if ext.state == "uninstall" or not self._extensions_goal_state.supports_encoded_signature():
-                continue
-            add_event(op=WALAEventOperation.ExtensionSigned, message="", name=ext.name, version=ext.version, is_success=ext.encoded_signature != "", log_event=False)
+        # Only send telemetry if the following conditions are met:
```
narrieta: Please mark every piece of code that is just temporary. Add a TODO in the comment and open a work item to change it,

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -279,13 +280,15 @@ def update(self, force_update=False, silent=False):
...
+        #   - Goal state API supports 'encoded_signature' property (e.g., for fast track goal states, HGAP version should support signature)
+        #   - Agent is running on a CVM (signature is currently only supported for CVMs)
+        #   - Extension requested state is *not* 'uninstall' (uninstall goal states never include signature).
+        if self._extensions_goal_state.supports_encoded_signature() and ConfidentialVMInfo.is_confidential_vm():
```
narrieta: ExtensionsGoalState.supports_encoded_signature needs to take into account the check for CVM. Currently it always returns True on extensionsConfig, but should be False on non-CVMs

azurelinuxagent/ga/signature_validation_util.py
```diff
@@ -280,9 +282,9 @@ def validate_attribute(attribute, extension_value):
 
 def signature_validation_enabled():
     """
-    Returns True if signature validation is enabled in conf file and OpenSSL version supports all validation parameters.
+    Returns True if signature validation is enabled in conf file, OpenSSL version supports all validation parameters, and agent is running on a Confidential VM.
```
narrieta, line 285: mark temporary code

azurelinuxagent/common/protocol/goal_state.py
narrieta: shouldn't this check also involve signature_validation_enabled?

azurelinuxagent/ga/confidential_vm_info.py
maddieford: Is TrustedVM type used for anything?

azurelinuxagent/ga/confidential_vm_util.py
maddieford: Right now we send telem on VM architecture in the 'KeywordName' column of each telemetry event. Right now the 'KeywordName' column is a json, should we add a new key for security type? 

May be helpful to add that telemetry as we continue creating cvm specific features

azurelinuxagent/ga/update.py
maddieford: Would it also be needed in the daemon? since we init/update the goal state there?

Or should we add an additional check in GoalState.update() to only send telem for ext signature if GoalStateProperties.ExtensionsGoalState was requested

tests_e2e/tests/ext_policy/ext_policy.py
maddieford, line 1: should the test suite be updated to run on cvm-endorsed images?

azurelinuxagent/ga/update.py
narrieta: > Would it also be needed in the daemon? since we init/update the goal state there?

Adding stuff in the Daemon is always scary, so let's not do it unless absolutely needed. As of today only the ExtensionsGoalState/Certificates and RemoteUsers are in scope for the CVM work stream. The Daemon does not fetch  any of those 2.

maddieford: Agreed we probably shouldn't touch the daemon, but wouldn't it exercise this code block in GoalState.update()?

```
        if self._extensions_goal_state.supports_encoded_signature() and ConfidentialVMInfo.is_confidential_vm():
            for ext in self._extensions_goal_state.extensions:
                if ext.state == "uninstall":
                    continue
                add_event(op=WALAEventOperation.ExtensionSigned, message="", name=ext.name, version=ext.version, is_success=ext.encoded_signature != "", log_event=False)

```
Should we add extra check for `self._goal_state_properties & GoalStateProperties.ExtensionsGoalState` in this condition?

narrieta: > Or should we add an additional check in GoalState.update() to only send telem for ext signature if GoalStateProperties.ExtensionsGoalState was requested

Do you think that extra check is needed? the current code sends that telemetry iterating over self._extensions_goal_state.extensions and that will be an empty array if GoalStateProperties.ExtensionsGoalState was not requested

```
            for ext in self._extensions_goal_state.extensions:
                if ext.state == "uninstall":
                    continue
                add_event(op=WALAEventOperation.ExtensionSigned, message="", name=ext.name, version=ext.version, is_success=ext.encoded_signature != "", log_event=False)
```

narrieta: oh, i posted my comment after you posted your second comment

it won't exercise it, the array will be empty

maddieford: self._extensions_goal_state.extensions is an empty array for now, since we ignore VmSettings if source is Fabric, but that will eventually change in the future when we start accepting Fabric GS from VmSettings

Though my main concern is checking ConfidentialVMInfo.is_confidential_vm() before the ConfidentialVMInfo has been initialized. Right now it's returning False if not initialized, but I'm not sure that's the correct behavior

narrieta: Yes, I get your point now. Under some scenarios this code is using ConfidentialVMInfo.is_confidential_vm() before it is initialized.

```
        if self._extensions_goal_state.supports_encoded_signature() and ConfidentialVMInfo.is_confidential_vm():
            for ext in self._extensions_goal_state.extensions:
                if ext.state == "uninstall":
                    continue
                add_event(op=WALAEventOperation.ExtensionSigned, message="", name=ext.name, version=ext.version, is_success=ext.encoded_signature != "", log_event=False)

```

azurelinuxagent/ga/confidential_vm_info.py
mgunnala: Not currently

azurelinuxagent/common/protocol/imds.py
mgunnala: I added all these properties to match the full, updated IMDS schema. If a property from the schema is not defined on the class, DataContract.set_properties() will log a warning about an unknown property. I can remove that warning instead if we prefer.

azurelinuxagent/ga/update.py
mgunnala: I have updated ConfidentialVMInfo.is_confidential_vm() to initialize the _security_type property, if not already initialized. This removes the need for a set_security_type() function. Let me know if this fully addresses both the above concern and the concerns discussed offline.

azurelinuxagent/ga/confidential_vm_util.py
mgunnala: I've removed this function

azurelinuxagent/ga/confidential_vm_util.py
mgunnala: I have updated the function to initialize the property, if not already initialized

azurelinuxagent/ga/confidential_vm_util.py
mgunnala: I've updated the code to send telemetry. If we decide to add a new key for security type, I'll take that as a follow-up task.

azurelinuxagent/ga/confidential_vm_info.py
mgunnala: I've updated the comment

azurelinuxagent/ga/confidential_vm_info.py
mgunnala: Added TODO comments here

azurelinuxagent/ga/confidential_vm_info.py
mgunnala: Updated

azurelinuxagent/common/protocol/goal_state.py
mgunnala: Updated

tests_e2e/tests/ext_policy/ext_policy.py
mgunnala: Updated

azurelinuxagent/common/protocol/goal_state.py
mgunnala: Updated

azurelinuxagent/ga/signature_validation_util.py
mgunnala: Updated

azurelinuxagent/common/protocol/goal_state.py
mgunnala: Updated and created a work item: https://msazure.visualstudio.com/One/_workitems/edit/35865513

azurelinuxagent/common/protocol/imds.py
mgunnala: I've changed this to API version '2021-12-13' and confirmed with the IMDS team that this build has been fully deployed to all clouds (mooncake, fairfax, and public)

azurelinuxagent/ga/update.py
narrieta: with this approach, if the call to imds fails callers will still get a value based on an uninitialized member; what is worse, if the failure is transient. the first caller will get False and the next caller can possibly get True

azurelinuxagent/ga/confidential_vm_info.py
narrieta: sorry, i meant document where this is setup in the VM's ARm template

azurelinuxagent/common/protocol/imds.py
narrieta: This is better, but I am still nervous about making this global change without at least 1 release to collect telemetry on possible failures using the newer API. Especially considering we are doing this just for a temporary feature. You do not need to do a change global to the Agent, you can do a separate query using the new API version just for the is CVM check.

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -279,13 +280,16 @@ def update(self, force_update=False, silent=False):
...
+        # Only send telemetry if the following conditions are met:
+        #   - Goal state API supports 'encoded_signature' property (e.g., for fast track goal states, HGAP version should support signature)
+        #   - Signature validation is enabled
+        #   - Extension requested state is *not* 'uninstall' (uninstall goal states never include signature).
+        #
+        if self._extensions_goal_state.supports_encoded_signature() and signature_validation_enabled():
+            for ext in self._extensions_goal_state.extensions:
+                if ext.state == "uninstall":
+                    continue
+                add_event(op=WALAEventOperation.ExtensionSigned, message="", name=ext.name, version=ext.version, is_success=ext.encoded_signature != "", log_event=False)
```
narrieta: add to local log as well

mgunnala: Added

azurelinuxagent/ga/confidential_vm_info.py
maddieford: Do we have plans to use it? Wondering if it's needed if it's not used anywhere

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -279,13 +280,18 @@ def update(self, force_update=False, silent=False):
...
-        for ext in self._extensions_goal_state.extensions:
-            if ext.state == "uninstall" or not self._extensions_goal_state.supports_encoded_signature():
-                continue
-            add_event(op=WALAEventOperation.ExtensionSigned, message="", name=ext.name, version=ext.version, is_success=ext.encoded_signature != "", log_event=False)
+        # Only send telemetry if the following conditions are met:
+        #   - Goal state API supports 'encoded_signature' property (e.g., for fast track goal states, HGAP version should support signature)
+        #   - Signature validation is enabled
+        #   - Extension requested state is *not* 'uninstall' (uninstall goal states never include signature).
+        #
+        if self._extensions_goal_state.supports_encoded_signature() and signature_validation_enabled():
```
narrieta: the idea was that is the feature is disabled in waagent.conf, we should avoid calling any code for this feature, so I think the order of these checks needs to be inverted

`signature_validation_enabled() and self._extensions_goal_state.supports_encoded_signature() 
`

azurelinuxagent/ga/confidential_vm_info.py
```diff
@@ -0,0 +1,80 @@
...
+
+        # Get securityType attribute
+        security_type = security_profile.get('securityType')
+        if security_type is None:
+            raise ValueError("missing field 'securityProfile'")
+
+        return security_type
+
+    @staticmethod
+    def is_confidential_vm():
```
narrieta: the new implementation still has the issue that was previously pointed out

> ... if the failure is transient. the first caller will get False and the next caller can possibly get True

narrieta: this method can potentially do a network request and the caller needs to be aware of that

most of the code uses "fetch_*" to name such methods

tests/ga/test_confidential_vm_info.py
```diff
@@ -0,0 +1,104 @@
...
+from tests.lib.tools import AgentTestCase, MagicMock, patch
+
+
+class TestConfidentialVMInfo(AgentTestCase):
+
+    def setUp(self):
+        ConfidentialVMInfo._security_type = None
+        AgentTestCase.setUp(self)
+
+    def _get_mock_imds_response(self, security_type, success=True):
```
narrieta: should we move this to a couple of data files similarly to tests/data/imds/valid.json?

azurelinuxagent/ga/confidential_vm_info.py
mgunnala: @narrieta I'm still working on updating this implementation, PR isn't yet ready for review

azurelinuxagent/common/protocol/goal_state.py
mgunnala: Good point, thanks. I've changed the order.

azurelinuxagent/ga/confidential_vm_info.py
mgunnala: I've updated the implementation to initialize only once, even on transient failure, and I've added a UT to test this scenario. Let me know if this addresses your concern.

mgunnala: Updated

tests/ga/test_confidential_vm_info.py
mgunnala: Yes, I've added the data files

azurelinuxagent/common/protocol/extensions_goal_state_from_extensions_config.py
```diff
@@ -575,9 +576,13 @@ def __parse_and_add_extension_settings(settings_node, name, extension, depends_o
 
     def supports_encoded_signature(self):
         """
-        Returns bool indicating if the ExtensionsConfig API supports the 'encoded_signature' extension property. Should always return True.
+        Return True if the ExtensionsConfig API supports the 'encoded_signature' extension property and agent is
```
narrieta, line 579: supports_encoded_signature can trigger a network request. it should be reflected in its name

azurelinuxagent/ga/confidential_vm_info.py
```diff
@@ -0,0 +1,86 @@
...
+
+        # Get securityType attribute
+        security_type = security_profile.get('securityType')
+        if security_type is None:
+            raise ValueError("missing field 'securityType'")
+
+        return security_type
+
+    @staticmethod
+    def fetch_is_confidential_vm():
```
narrieta: did you check the telemetry you are emitting? it's missing the operation name

azurelinuxagent/ga/signature_validation_util.py
```diff
@@ -280,9 +282,12 @@ def validate_attribute(attribute, extension_value):
...
```
narrieta, line 285: signature_validation_enabled can trigger a network request; should be reflected on its name

mgunnala:
Offline comment from Norberto: lazy initialization may not be the best approach for ConfidentialVMInfo, might make more sense to do an explicit initialization

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -279,13 +280,18 @@ def update(self, force_update=False, silent=False):
...
+        # Only send telemetry if the following conditions are met:
+        #   - Goal state API supports 'encoded_signature' property (e.g., for fast track goal states, HGAP version should support signature)
+        #   - Signature validation is enabled
+        #   - Extension requested state is *not* 'uninstall' (uninstall goal states never include signature).
+        #
+        if signature_validation_enabled() and self._extensions_goal_state.supports_encoded_signature():
+            for ext in self._extensions_goal_state.extensions:
+                if ext.state == "uninstall":
+                    continue
+                msg = "Goal state {0} signature for extension package '{1}'".format("contains" if ext.encoded_signature else "does not contain", ext.name)
```
maddieford: nit: If 'name' column already contains ext name, do we need it in the message too?

mgunnala: We don't, but I just wanted to use the same message in the telemetry and log for consistency. I don't have a problem removing it from the telemetry, though.

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -261,38 +262,45 @@ def update(self, force_update=False, silent=False):
...
-            add_event(op=WALAEventOperation.ExtensionSigned, message="", name=ext.name, version=ext.version, is_success=ext.encoded_signature != "", log_event=False)
-
-        # Ensure all certificates are downloaded on Fast Track goal states in order to maintain backwards compatibility with previous
-        # versions of the Agent, which used to download certificates from the WireServer on every goal state. Some customer applications
-        # depend on this behavior (see https://github.com/Azure/WALinuxAgent/issues/2750).
-        #
-        if self._extensions_goal_state.source == GoalStateSource.FastTrack and self._goal_state_properties & GoalStateProperties.Certificates:
-            self._check_and_download_missing_certs_on_disk()
+        # Process only if extensions goal state was requested
+        if self._goal_state_properties & GoalStateProperties.ExtensionsGoalState:
```
mgunnala, line 266: offline comment from Maddie: we could consider putting just the signature validation telemetry code block inside this if statement, for now.

azurelinuxagent/ga/confidential_vm_info.py
mgunnala: I removed since we're not using it anywhere, we can eventually add it back if we need to enable a feature on TVMs

azurelinuxagent/ga/signature_validation_util.py
mgunnala: I've updated to use an explicit initialization for ConfidentialVMInfo so this function will no longer make a network request

azurelinuxagent/common/protocol/extensions_goal_state_from_extensions_config.py
mgunnala: I've updated to use an explicit initialization for ConfidentialVMInfo so this function will no longer make a network request

azurelinuxagent/ga/confidential_vm_info.py
mgunnala: Thank you for checking and catching this, I've updated the telemetry here and double-checked that the signature validation telemetry looks correct: 

https://azcrp.kusto.windows.net/crp_allprod?query=H4sIAAAAAAAEAGVQS0%2fCQBC%2bm%2fgfNnuhJLShReSESQ9ITHw0YvA8bseyUnabnVmLxh%2fvtjGAepv5XvNQtSdGF0n4VNZhotCwgwAm20DYpNWmtC0lBlkOkxIYXoAwktcQ2qVH4rwKlsWe0ZC2ZvEeOjo%2f%2bxLtBh2KwqHShE96hyuGXSOuBFQ2mpbDo2aZr9F1ZkEMjqnVvBHyOb%2fVxu%2f7eHkUPzTogDuxNiKSh8ErXRks5UjIrgL2DtdQ67LXdnABagsVHthHJF9zx%2fzK6LnC1lp9%2fHFpU92YV%2fvjO9k%2fINY7hUtnfXMPOxTzuRjUmiA%2bPSLOxtk0TbNpKCaTi1mczsYxZpfpoItqnH1Dxf8eNjpePBJ3SBS2%2bQYTl%2b0gtQEAAA%3d%3d&web=0

cluster("azcore.centralus.kusto.windows.net").database("Fa").GuestAgentExtensionEvents
| where PreciseTimeStamp > ago(5d)
| where GAVersion startswith "WALinuxAgent"
| where Operation in ("ExtensionSigned", "SignatureValidation", "PackageSignatureResult", "ExtensionSignaturePolicy", "PackageSigningInfoResult")
| where ResourceGroupName == 'lisa-WALinuxAgent-20251125-203347-170-e261'
| project PreciseTimeStamp, Operation, Message

azurelinuxagent/ga/confidential_vm_info.py
```diff
@@ -0,0 +1,80 @@
...
+
+    @staticmethod
+    def fetch_and_initialize_security_type():
+        try:
+            security_type = ConfidentialVMInfo._fetch_security_type_from_imds()
+            event.info(event.WALAEventOperation.SignatureValidation, "VM security type: {0}", security_type)
+            ConfidentialVMInfo._security_type = security_type
+        except Exception as ex:
+            event.warn(event.WALAEventOperation.SignatureValidation, "Failed to get virtual machine security type from IMDS: {0}", ustr(ex))
+            ConfidentialVMInfo._security_type = False
```
narrieta: sometimes this variable holds a string, sometimes a boolean. this makes introducing a code bug easier

azurelinuxagent/ga/confidential_vm_info.py
```diff
@@ -0,0 +1,80 @@
...
+        return security_type
+
+    @staticmethod
+    def fetch_and_initialize_security_type():
+        try:
+            security_type = ConfidentialVMInfo._fetch_security_type_from_imds()
+            event.info(event.WALAEventOperation.SignatureValidation, "VM security type: {0}", security_type)
+            ConfidentialVMInfo._security_type = security_type
+        except Exception as ex:
+            event.warn(event.WALAEventOperation.SignatureValidation, "Failed to get virtual machine security type from IMDS: {0}", ustr(ex))
```
narrieta:
```suggestion
            event.warn(event.WALAEventOperation.SignatureValidation, "Failed to get virtual machine security type from IMDS, will assume this is not a Confidential VM: {0}", ustr(ex))
```

azurelinuxagent/ga/confidential_vm_info.py
```diff
@@ -0,0 +1,80 @@
...
+        # Get securityType attribute
+        security_type = security_profile.get('securityType')
+        if security_type is None:
+            raise ValueError("missing field 'securityType'")
+
+        return security_type
+
+    @staticmethod
+    def fetch_and_initialize_security_type():
+        try:
```
narrieta, line 77: Add TODO comment that for the moment we assume non-CVM on IDMS failure and for the rest of this instance of the service execution, but that this needs to be improved.

azurelinuxagent/ga/update.py
```diff
@@ -360,6 +361,8 @@ def run(self, debug=False):
             #
             protocol = self.protocol_util.get_protocol(save_to_history=True)
 
+            ConfidentialVMInfo.fetch_and_initialize_security_type()
```
narrieta: this needs to be initialized before the goal state is fetched. from previous conversation, remember get_protocol() triggers fetching the goal state

azurelinuxagent/ga/confidential_vm_info.py
```diff
@@ -0,0 +1,80 @@
...
+            security_type = ConfidentialVMInfo._fetch_security_type_from_imds()
+            event.info(event.WALAEventOperation.SignatureValidation, "VM security type: {0}", security_type)
+            ConfidentialVMInfo._security_type = security_type
+        except Exception as ex:
+            event.warn(event.WALAEventOperation.SignatureValidation, "Failed to get virtual machine security type from IMDS: {0}", ustr(ex))
+            ConfidentialVMInfo._security_type = False
+
+    @staticmethod
+    def is_confidential_vm():
+        return ConfidentialVMInfo._security_type == SecurityType.ConfidentialVM
```
narrieta: should raise if called before initialization

azurelinuxagent/ga/confidential_vm_info.py
mgunnala: Since we're not using the security type for anything other than determining if this is a CVM, I've changed this variable to a boolean _is_confidential_vm.

azurelinuxagent/ga/confidential_vm_info.py
mgunnala: Updated

azurelinuxagent/ga/confidential_vm_info.py
mgunnala: Updated

azurelinuxagent/ga/confidential_vm_info.py
mgunnala: Updated

azurelinuxagent/ga/update.py
```diff
@@ -354,6 +355,8 @@ def run(self, debug=False):
                 )
             logger.info(os_info_msg)
 
+            ConfidentialVMInfo.fetch_and_initialize_security_type()
```
maddieford: This telem can't be associated with the VM since it's getting sent before the common properties are initialized: 

cluster("azcore.centralus.kusto.windows.net").database("Fa").GuestAgentExtensionEvents
| where PreciseTimeStamp > ago(6d)
| where Message contains 'VM security type:'
| where GAVersion == "WALinuxAgent-9.9.9.9"
| project OpcodeName, Operation, OperationSuccess, Message, TenantName, RoleName, RoleInstanceName, ContainerId, SubscriptionId, ResourceGroupName, VMId
| sort by OpcodeName asc
| take 5

Image

maddieford: I think it might be appropriate to include the security type in the os_info_msg which we construct on line 339 and send telemetry for on line 379 after the common properties are initialized. 

Right now fetch_and_initialize_security_type() does not return or store the security_type itself. I would either update the method to return the security type or add another class variable to keep the security type

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -261,38 +262,45 @@ def update(self, force_update=False, silent=False):
...
+                self._extensions_goal_state = most_recent
+
+            # For each extension in the goal state being executed, we emit telemetry to indicate whether a signature is present
+            # for the extension. The "is_success" field reflects whether the extension was signed.
+            # Only send telemetry if the following conditions are met:
+            #   - Goal state API supports 'encoded_signature' property (e.g., for fast track goal states, HGAP version should support signature)
+            #   - Signature validation is enabled
+            #   - Extension requested state is *not* 'uninstall' (uninstall goal states never include signature).
+            #
+            if signature_validation_enabled() and self._extensions_goal_state.supports_encoded_signature():
```
maddieford, line 290: The code as-is will hit these checks without having initialized the _is_confidential_cvm variable when the provisioning agent is used. Please check these code paths:
1. main() -> Agent.provision() -> ProvisionHandler.run() -> ProvisionHandler.report_ready()/ProvisionHandler.report_not_ready() -> ProtocolUtil.get_protocol() -> ProtocolUtil._detect_protocol() -> WireProtocol.detect() -> WireClient.reset_goal_state() -> GoalState.__init__() -> GoalState.update()
2. main() -> Agent.daemon() -> DaemonHandler.run() -> DaemonHandler.daemon() -> ProvisionHandler.run() -> ProvisionHandler.report_ready()/ProvisionHandler.report_not_ready() -> ProtocolUtil.get_protocol() -> ProtocolUtil._detect_protocol() -> WireProtocol.detect() -> WireClient.reset_goal_state() -> GoalState.__init__() -> GoalState.update()

azurelinuxagent/ga/update.py
mgunnala: I've updated this code to send telemetry after the common properties are initialized, and updated the telemetry to just say whether the VM is confidential or not, since we're not actually using the security type for anything else. Let me know if this looks okay to you. 

cluster("azcore.centralus.kusto.windows.net").database("Fa").GuestAgentExtensionEvents
| where PreciseTimeStamp > ago(2d)
| where Operation == "SignatureValidation"
| where Message contains 'confidential virtual machine'
| where GAVersion == "WALinuxAgent-9.9.9.9"
| project OpcodeName, Operation, OperationSuccess, Message, TenantName, RoleName, RoleInstanceName, ContainerId, SubscriptionId, ResourceGroupName, VMId
| sort by OpcodeName asc
| take 10

image

azurelinuxagent/ga/update.py
mgunnala: Updated

azurelinuxagent/common/protocol/goal_state.py
mgunnala: Discussed in our sync

azurelinuxagent/ga/confidential_vm_info.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class ConfidentialVMInfo(object):
+    # This class temporarily provides a way to detect whether the VM is a Confidential VM (CVM) via IMDS.
+    # It is used to limit certain features to CVMs while we build confidence in the feature
+    # before enabling it across the broader fleet (telemetry/preview releases only).
+    #
+    # TODO: Remove once extension signature validation is supported on all VMs.
+
+    _is_confidential_vm = None  # Boolean value to track whether this is a CVM
```
narrieta: add comment saying this is tri-state (and describing what None means)

narrieta (dismissed):
lgtm, just one minor comment

azurelinuxagent/ga/confidential_vm_info.py
mgunnala: Thanks, I've updated this comment

azurelinuxagent/ga/confidential_vm_info.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class ConfidentialVMInfo(object):
+    # This class temporarily provides a way to detect whether the VM is a Confidential VM (CVM) via IMDS.
+    # It is used to limit certain features to CVMs while we build confidence in the feature
+    # before enabling it across the broader fleet (telemetry/preview releases only).
+    #
+    # TODO: Remove once extension signature validation is supported on all VMs.
+
+    # Tri-state boolean: True if CVM, False if not a CVM, None if not yet initialized (fetch_and_initialize_cvm_info has not been called).
```
maddieford: Minor comment, False if not a CVM or call to IMDS fails or is otherwise inconclusive

tests/ga/test_confidential_vm_info.py
maddieford, line 1: should we add unit test that we raise if called before initialized?

mgunnala: Added!

azurelinuxagent/ga/confidential_vm_info.py
mgunnala: Updated, thanks

PR #3488, opened 2025-11-04
Set separate CPU quota for extension signature validation

mgunnala:
## Description

Issue # 

---

This PR updates the extension signature validation code to run with a separate CPU quota (currently set to 50%), to improve validation performance and reduce impact on goal state processing time.

### PR information

### DIFF
--- a/tests_e2e/tests/agent_update/rsm_update.py
+++ b/tests_e2e/tests/agent_update/rsm_update.py
@@ -196,13 +196,7 @@ class RsmUpdateBvt(AgentVmTest):
         """
 
         log.info("Executing verify_versioning_supported_feature.py remote script to verify agent reported supported feature flag")
-        # In these test VMs, direct artifact blob downloads are failing, and our retry
-        # logic takes a long time to exhaust all attempts. As a result, the agent is
-        # blocked from continuing execution and cannot report status to CRP quickly.
-        #
-        # Increasing the number of attempts and the delay is added here as a temporary
-        # workaround while we investigate the root cause of the blob download failures.
-        self._run_remote_test(self._ssh_client, f"agent_update-verify_versioning_supported_feature.py --supported {supported}", attempts=10, attempt_delay=60, use_sudo=True)
+        self._run_remote_test(self._ssh_client, f"agent_update-verify_versioning_supported_feature.py --supported {supported}", use_sudo=True)
         if supported:
             log.info("Successfully verified that Agent reported VersioningGovernance supported feature flag")
         else:
--- a/tests_e2e/tests/scripts/agent_update-verify_versioning_supported_feature.py
+++ b/tests_e2e/tests/scripts/agent_update-verify_versioning_supported_feature.py
@@ -51,13 +51,19 @@ def main():
     args = parser.parse_args()
 
     log.info("checking agent status file for VersioningGovernance supported feature flag")
+    # In these test VMs, direct artifact blob downloads are failing, and our retry
+    # logic takes a long time to exhaust all attempts. As a result, the agent is
+    # blocked from continuing execution and cannot report status to CRP quickly.
+    #
+    # Increasing the number of attempts and the delay is added here as a temporary
+    # workaround while we investigate the root cause of the blob download failures.
     if args.supported == "True":
-        found: bool = retry_if_false(check_agent_supports_versioning)
+        found: bool = retry_if_false(check_agent_supports_versioning, attempts=10, delay=60)
         if not found:
             raise Exception("Agent failed to report supported feature flag. So, skipping agent update validations "
                             "since CRP will not send RSM requested version in GS if feature flag not found in status")
     else:
-        found: bool = retry_if_true(check_agent_supports_versioning)
+        found: bool = retry_if_true(check_agent_supports_versioning, attempts=10, delay=60)
         if found:
             raise Exception("Agent should not report supported feature flag while agent updates disabled in the vm.")
 


azurelinuxagent/ga/signature_validation_util.py
```diff
@@ -191,6 +193,19 @@ def validate_signature(package_path, signature, package_full_name):
             '-CAfile', microsoft_root_cert_file,  # Path to the trusted root certificate file used for verification
             '-no_check_time'  # Skips checking whether the certificate is expired
         ]
+
+        # Signature validation is a CPU-intensive operation. If cgroups are enabled and the agent's CPU quota is set too low,
+        # the validation process may take an excessive amount of time. As a workaround, if cgroups are enabled,
+        # we run extension signature validation in a separate cgroup with its own dedicated CPU quota.
+        if CGroupConfigurator.get_instance().enabled():
```
narrieta: Why this dependency on cgroups configurator?

azurelinuxagent/ga/signature_validation_util.py
```diff
@@ -191,6 +193,19 @@ def validate_signature(package_path, signature, package_full_name):
...
+            systemd_cmd = [
```
narrieta: Errors contacting the dbus API are not uncommon, In that case, the extension should still be executed. See the error handling in start_extension_command

azurelinuxagent/ga/signature_validation_util.py
mgunnala: This is to check whether cgroups are enabled/supported, before creating a separate cgroup for the signature validation command

azurelinuxagent/ga/signature_validation_util.py
```diff
@@ -191,7 +193,41 @@ def validate_signature(package_path, signature, package_full_name):
             '-CAfile', microsoft_root_cert_file,  # Path to the trusted root certificate file used for verification
             '-no_check_time'  # Skips checking whether the certificate is expired
         ]
-        run_command(command, encode_output=False)
+
+        # If cgroups are enabled, attempt to run the command in a dedicated systemd-run scope with a dedicated CPU quota.
+        # This is because signature validation is CPU-intensive and may take excessive time if the agent's CPU quota is low.
+        # If the systemd-run invocation fails, fall back to running the OpenSSL command directly.
+        use_cgroups = CGroupConfigurator.get_instance().enabled()
```
mgunnala, line 203: could move this code block to a helper function like _validate_signature_with_systemd_run_fallback(), but I found it simpler to follow the logic without the helper.

azurelinuxagent/ga/signature_validation_util.py
mgunnala: Thanks for pointing this out, I've updated the error handling logic to re-run the openssl command without 'systemd-run' in the case of systemd failures.

azurelinuxagent/ga/signature_validation_util.py
```diff
@@ -191,7 +193,41 @@ def validate_signature(package_path, signature, package_full_name):
...
+                '--unit={0}'.format(EXT_SIGNATURE_VALIDATION_CGROUPS_UNIT),
+                '--slice={0}'.format(EXT_SIGNATURE_VALIDATION_SLICE),
+                '--scope',
+                '--property=CPUAccounting=yes',
+                '--property=CPUQuota={0}'.format(EXT_SIGNATURE_VALIDATION_CPU_QUOTA)
+            ] + base_command
+            try:
+                run_command(systemd_cmd, encode_output=False)
+            except CommandError as ex:
+                # If the systemd-run invocation itself failed (e.g., systemd not available, access denied, bus errors),
```
nagworld9: nit: D-bus

azurelinuxagent/ga/signature_validation_util.py
```diff
@@ -191,7 +193,41 @@ def validate_signature(package_path, signature, package_full_name):
...
+                # log a warning and fall back to running command directly. If the openssl command failed, re-raise and do not retry.
+                stderr_str = ex.stderr.decode('utf-8') if isinstance(ex.stderr, bytes) else ex.stderr
+                unit_not_found = "Unit {0} not found.".format(EXT_SIGNATURE_VALIDATION_CGROUPS_UNIT)
+                is_systemd_failure = unit_not_found in stderr_str or EXT_SIGNATURE_VALIDATION_CGROUPS_UNIT not in stderr_str
```
nagworld9: can we share the same logic what we do for extensions-run? If we need to change at some point how we detect, we may miss or required to change in all places.

azurelinuxagent/ga/signature_validation_util.py
```diff
@@ -191,7 +193,41 @@ def validate_signature(package_path, signature, package_full_name):
...
+                
+                if is_systemd_failure:
+                    report_validation_event(op=WALAEventOperation.SignatureValidation, level=logger.LogLevel.WARNING,
+                        message="'systemd-run' invocation failed for signature validation, falling back to direct execution. Error: '{0}'".format(ex.stderr),
+                        name=name, version=version, duration=0)
+                    # Run without systemd
+                    run_command(base_command, encode_output=False)
```
nagworld9: I already mentioned to you, just pointing out again to let others know. When we run directly, process now placed in agent cgroup and it will get agent limits. Today agent cpu at 50% which is matching what you are doing but quota may change later. We need to think about long term solution what we want to do. 

Context: On the extension run, when systemd-run failures, we reset the quota and disable the cgroups feature in the agent. So that extensions run in the agent cgroup without limits.

azurelinuxagent/ga/signature_validation_util.py
maddieford, line 1: Are we actually tracking the metrics in this new cgroup for signature validation? If not, could you remind me the motivation for doing the validation in another cgroup?

mgunnala: We need to assign a separate CPU quota for the signature validation process so when the agent’s quota is eventually reduced, validation won't take an excessive amount of time.

azurelinuxagent/ga/signature_validation_util.py
mgunnala: Updated, thanks

azurelinuxagent/ga/signature_validation_util.py
mgunnala: I've added a util function "is_systemd_failure()" in cgroupsapi.py, and used that function both in start_extension_command() and in validate_signature(). Let me know if this looks okay to you.

azurelinuxagent/ga/signature_validation_util.py
mgunnala: Discussed in our sync: we should reset the quota and disable the cgroups feature in the agent, in the case of systemd-run failures for signature validation.

azurelinuxagent/ga/signature_validation_util.py
narrieta: the concern was about whether we want to use the same distros for signature validation work as we use for agent quotas work. most likely this is ok.

azurelinuxagent/ga/signature_validation_util.py
mgunnala: Discussed offline with @nagworld9: tracking metrics will need to be discussed further and taken up in a separate PR

azurelinuxagent/ga/signature_validation_util.py
mgunnala: Updated accordingly

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -42,6 +42,28 @@
 EXTENSION_SLICE_PREFIX = "azure-vmextensions"
 
 
+def is_systemd_failure(unit_name, stderr):
```
narrieta: Move to azurelinuxagent/common/osutil/systemd.py and rename to is_systemd_run_failure. Also, since now this is a public method, document how it decides whether it is a systemd-run error or not.

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -42,6 +42,28 @@
...
+    
+    :param unit_name: The name of the systemd unit/scope
+    :param stderr: Error output as str, bytes, or file-like object
+    :return: True if this is a systemd failure
+    """
+    # Handle different types of stderr input
+    if hasattr(stderr, 'seek') and hasattr(stderr, 'read'):
+        stderr.seek(0)
+        stderr_str = ustr(stderr.read(TELEMETRY_MESSAGE_MAX_LEN), encoding='utf-8', errors='backslashreplace')
+    elif isinstance(stderr, bytes):
```
narrieta: why do we need both bytes and str?

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -42,6 +42,28 @@
...
 
 
+def is_systemd_failure(unit_name, stderr):
+    """
+    Determines if stderr from a systemd-run command indicates a systemd failure (vs a command failure).
+    Systemd failures include: unit not found, systemd not available, D-bus errors, etc.
+    
+    :param unit_name: The name of the systemd unit/scope
+    :param stderr: Error output as str, bytes, or file-like object
+    :return: True if this is a systemd failure
```
narrieta:
```suggestion
    :return: True if this is a systemd-run failure
```

azurelinuxagent/ga/signature_validation_util.py
```diff
@@ -191,7 +194,32 @@ def validate_signature(package_path, signature, package_full_name):
...
+        # If cgroups are enabled, attempt to run the command in a dedicated systemd-run scope with a dedicated CPU quota.
+        # This is because signature validation is CPU-intensive and may take excessive time if the agent's CPU quota is low.
+        # If the systemd-run invocation fails, disable cgroups entirely and fall back to running the OpenSSL command directly.
+        use_cgroups = CGroupConfigurator.get_instance().enabled()
+        if use_cgroups:
+            systemd_cmd = ['systemd-run', '--unit={0}'.format(EXT_SIGNATURE_VALIDATION_CGROUPS_UNIT),
+                            '--slice={0}'.format(EXT_SIGNATURE_VALIDATION_SLICE), '--scope', '--property=CPUAccounting=yes',
+                            '--property=CPUQuota={0}'.format(EXT_SIGNATURE_VALIDATION_CPU_QUOTA)] + base_command
+            try:
+                run_command(systemd_cmd, encode_output=False)
```
narrieta: why encode_output=False?

tests/ga/test_signature_validation_sudo.py
```diff
@@ -42,6 +44,20 @@ def tearDown(self):
...
         AgentTestCase.tearDown(self)
 
+    @staticmethod
+    @contextlib.contextmanager
+    def create_mock_run_command():
+        calls = []
+
+        original_run_command = shellutil.run_command
+
+        def run_command_mock(cmd, *args, **kwargs):
```
narrieta: you can use the existing 'wraps' and 'call_list' instead of implementing your own, there are several samples in the code

also, we usually mock at the level of Popen, rather than run_command (protects against changes in run_command)

tests/ga/test_signature_validation_sudo.py
```diff
@@ -77,3 +93,84 @@ def test_should_validate_signature_for_package_signed_with_leaf_root_cert(self):
...
+    def test_validate_signature_should_use_systemd_run(self):
+        self.assertTrue(i_am_root(), "Test does not run when non-root")
+        with patch("azurelinuxagent.ga.signature_validation_util.CGroupConfigurator.get_instance") as mock_get_instance:
+            mock_instance = mock_get_instance.return_value
+            mock_instance.enabled.return_value = True
+            with self.create_mock_run_command() as run_command_calls:
+                validate_signature(self.vm_access_zip_path, self.vm_access_signature, self.package_name_and_version)
+
+            # Check if 'openssl cms -verify' was called with systemd-run
+            systemd_run_called = any(
```
narrieta, line 101: if we ever add another call to run_command that is executed in the scope of signature validation, this is going to break

see samples of Popen mocks in the existing code for how to handle this

narrieta: similar comments below

tests/ga/test_signature_validation_sudo.py
```diff
@@ -77,3 +93,84 @@ def test_should_validate_signature_for_package_signed_with_leaf_root_cert(self):
...
+            mock_instance = mock_get_instance.return_value
+            mock_instance.enabled.return_value = False
+            
+            with self.create_mock_run_command() as run_command_calls:
+                validate_signature(self.vm_access_zip_path, self.vm_access_signature, self.package_name_and_version)
+                
+                # Verify openssl was called directly (not through systemd-run)
+                self.assertEqual(1, len(run_command_calls))
+                cmd_args = run_command_calls[0][0]
+                self.assertTrue(cmd_args[0].endswith('openssl'))
```
narrieta: add a message to this assert

tests/ga/test_signature_validation_sudo.py
```diff
@@ -77,3 +93,84 @@ def test_should_validate_signature_for_package_signed_with_leaf_root_cert(self):
...
+            mock_instance = mock_get_instance.return_value
+            mock_instance.enabled.return_value = True
+            with self.create_mock_run_command() as run_command_calls:
+                validate_signature(self.vm_access_zip_path, self.vm_access_signature, self.package_name_and_version)
+
+            # Check if 'openssl cms -verify' was called with systemd-run
+            systemd_run_called = any(
+                all(s in " ".join(cmd) for s in ["systemd-run", "openssl", "cms", "-verify"])
+                for cmd, _, _ in run_command_calls
+            )
```
narrieta, line 104: add a message to this assert

tests/ga/test_signature_validation_sudo.py
```diff
@@ -77,3 +93,84 @@ def test_should_validate_signature_for_package_signed_with_leaf_root_cert(self):
...
+            
+            def mock_run_command_with_error(cmd, *args, **kwargs):
+                # Simulate OpenSSL failure (unit name in stderr means it's not a systemd failure)
+                if 'openssl' in ' '.join(cmd):
+                    raise shellutil.CommandError(command=cmd, return_code=1, stdout='',
+                                                 stderr='Running as unit: {0}\nVerification failure'.format(EXT_SIGNATURE_VALIDATION_CGROUPS_UNIT))
+                return shellutil.run_command(cmd, *args, **kwargs)
+            
+            with patch("azurelinuxagent.ga.signature_validation_util.run_command", side_effect=mock_run_command_with_error):
+                with self.assertRaises(SignatureValidationError):
```
narrieta: add a message to this assert

tests/ga/test_signature_validation_sudo.py
```diff
@@ -77,3 +93,84 @@ def test_should_validate_signature_for_package_signed_with_leaf_root_cert(self):
...
+                # Fail if command contains systemd-run (simulating systemd not available)
+                if 'systemd-run' in ' '.join(cmd):
+                    raise shellutil.CommandError(command=cmd, return_code=1, stdout='',
+                                                 stderr='Unit {0} not found.'.format(EXT_SIGNATURE_VALIDATION_CGROUPS_UNIT))
+                return original_run_command(cmd, *args, **kwargs)
+
+            with patch("azurelinuxagent.ga.signature_validation_util.run_command", side_effect=run_command_with_systemd_failure):
+                validate_signature(self.vm_access_zip_path, self.vm_access_signature, self.package_name_and_version)
+
+                self.assertEqual(2, len(calls))
```
narrieta: add a message to this assert

tests/ga/test_signature_validation_sudo.py
```diff
@@ -77,3 +93,84 @@ def test_should_validate_signature_for_package_signed_with_leaf_root_cert(self):
...
+                # First command should be invoked via systemd-run
+                self.assertIn('systemd-run', ' '.join(calls[0]))
```
narrieta: add a message to this assert

tests/ga/test_signature_validation_sudo.py
```diff
@@ -77,3 +93,84 @@ def test_should_validate_signature_for_package_signed_with_leaf_root_cert(self):
...
+                # Second command should be a direct openssl call (no systemd-run)
+                self.assertNotIn('systemd-run', ' '.join(calls[1]))
```
narrieta: add a message to this assert

tests/ga/test_signature_validation_sudo.py
```diff
@@ -77,3 +93,84 @@ def test_should_validate_signature_for_package_signed_with_leaf_root_cert(self):
...
+
+                # Verify that cgroups were disabled
+                self.assertEqual(1, mock_instance.disable.call_count, "disable() should have been called exactly once")
+                reason = mock_instance.disable.call_args[1]['reason']
+                self.assertIn("'systemd-run' invocation failed for signature validation", reason)
```
narrieta: in several checks you do "foo in formatted_command" (or similar), could you be more strict and so "startswith" or command[0] == 'foo' (or similar)

maddieford (commented):
No additional comments from my side. I'll give it another review after the open comments are resolved

azurelinuxagent/ga/cgroupapi.py
mgunnala: Updated and added a comment detailing how it decides systemd-run errors.

tests/ga/test_signature_validation_sudo.py
mgunnala: Thanks for the suggestion, I've updated this.

azurelinuxagent/ga/signature_validation_util.py
mgunnala: Updated to use default value True

azurelinuxagent/ga/cgroupapi.py
mgunnala: Updated

tests/ga/test_signature_validation_sudo.py
mgunnala: Updated following the examples of other Popen mocks, let me know if this addresses your concern sufficiently.

tests/ga/test_signature_validation_sudo.py
mgunnala: Added

tests/ga/test_signature_validation_sudo.py
mgunnala: Added

tests/ga/test_signature_validation_sudo.py
mgunnala: Added

tests/ga/test_signature_validation_sudo.py
mgunnala: Added

tests/ga/test_signature_validation_sudo.py
mgunnala: Added

tests/ga/test_signature_validation_sudo.py
mgunnala: Good point, I've updated the tests to use regex matching and "startswith" instead

azurelinuxagent/common/osutil/systemd.py
```diff
@@ -130,4 +132,43 @@ def is_unit_loaded(unit_name):
...
+    :param unit_name: The name of the systemd unit/scope that was used with systemd-run
+    :param stderr: Error output from the systemd-run command, expected to be a file-like object or a string.
+    :return: True if this is a systemd-run failure, False if it's a command execution failure
+    """
+    if hasattr(stderr, 'seek') and hasattr(stderr, 'read'):
+        stderr.seek(0)
+        stderr_str = ustr(stderr.read(TELEMETRY_MESSAGE_MAX_LEN), encoding='utf-8', errors='backslashreplace')
+    else:
+        stderr_str = stderr
+
```
narrieta, line 172:
```
    if isinstance(stderr, ustr):
        stderr_str = stderr
    else:
        stderr.seek(0)
        stderr_str = ustr(stderr.read(TELEMETRY_MESSAGE_MAX_LEN), encoding='utf-8', errors='backslashreplace')
```

tests/lib/tools.py
```diff
@@ -495,6 +500,25 @@ def patch_builtin(target, *args, **kwargs):
     return patch("{0}.{1}".format(prefix, target), *args, **kwargs)
 
 
+def patch_encode_command_output():
```
narrieta: we don't want to add this to tools

this is not a pattern other tests should follow

tests/ga/test_extension.py
```diff
@@ -3878,6 +3878,8 @@ def setUp(self):
         self.patch_conf_flag.start()
         self.patch_is_cvm = patch('azurelinuxagent.ga.confidential_vm_info.ConfidentialVMInfo.is_confidential_vm', return_value=True)
         self.patch_is_cvm.start()
+        self.patch_encode_output = patch_encode_command_output()
```
narrieta: we shouldn't patch at this level; this touches too many tests

azurelinuxagent/ga/signature_validation_util.py
```diff
@@ -193,7 +195,32 @@ def validate_signature(package_path, signature, package_full_name):
...
+            try:
+                run_command(systemd_cmd)
+            except CommandError as ex:
+                # If the systemd-run invocation itself failed, disable cgroups entirely and fall back to running openssl command directly.
+                # If the openssl command failed, re-raise and do not retry.
+                if is_systemd_run_failure(EXT_SIGNATURE_VALIDATION_CGROUPS_UNIT, ex.stderr):
+                    error_msg = "'systemd-run' invocation failed for signature validation, disabling cgroups and falling back to direct execution. Error: '{0}'".format(ex.stderr)
+                    report_validation_event(op=WALAEventOperation.SignatureValidation, level=logger.LogLevel.WARNING,
+                        message=error_msg,
+                        name=name, version=version, duration=0)
```
nagworld9, line 218: is this name and version we log for extension?

azurelinuxagent/ga/signature_validation_util.py
```diff
@@ -193,7 +195,32 @@ def validate_signature(package_path, signature, package_full_name):
...
+        if use_cgroups:
+            systemd_cmd = ['systemd-run', '--unit={0}'.format(EXT_SIGNATURE_VALIDATION_CGROUPS_UNIT),
+                            '--slice={0}'.format(EXT_SIGNATURE_VALIDATION_SLICE), '--scope', '--property=CPUAccounting=yes',
+                            '--property=CPUQuota={0}'.format(EXT_SIGNATURE_VALIDATION_CPU_QUOTA)] + base_command
+            try:
+                run_command(systemd_cmd)
+            except CommandError as ex:
+                # If the systemd-run invocation itself failed, disable cgroups entirely and fall back to running openssl command directly.
+                # If the openssl command failed, re-raise and do not retry.
+                if is_systemd_run_failure(EXT_SIGNATURE_VALIDATION_CGROUPS_UNIT, ex.stderr):
```
nagworld9: EXT_SIGNATURE_VALIDATION_CGROUPS_UNIT includes .scope suffix in the string whereas for extension run, we are not passing .scope. can we match with that?

azurelinuxagent/ga/signature_validation_util.py
mgunnala: Yes, it is the extension name and version (same parameters we use for extensions)

azurelinuxagent/ga/signature_validation_util.py
mgunnala: Updated to follow the extension run pattern

azurelinuxagent/common/osutil/systemd.py
mgunnala: Updated, thanks for the suggestion

tests/lib/tools.py
mgunnala: Removed this workaround

tests/ga/test_extension.py
mgunnala: Removed this workaround

azurelinuxagent/ga/signature_validation_util.py
```diff
@@ -184,16 +186,43 @@ def validate_signature(package_path, signature, package_full_name):
...
+        # This is because signature validation is CPU-intensive and may take excessive time if the agent's CPU quota is low.
+        # If the systemd-run invocation fails, disable cgroups entirely and fall back to running the OpenSSL command directly.
+        use_cgroups = CGroupConfigurator.get_instance().enabled()
+        if use_cgroups:
+            slice_name = EXT_SIGNATURE_VALIDATION_SLICE_NAME + ".slice"
+            scope_name = EXT_SIGNATURE_VALIDATION_CGROUPS_UNIT_NAME + ".scope"
+            systemd_cmd = ['systemd-run', '--unit={0}'.format(scope_name), '--slice={0}'.format(slice_name), '--scope', '--property=CPUAccounting=yes',
+                           '--property=CPUQuota={0}'.format(EXT_SIGNATURE_VALIDATION_CPU_QUOTA)] + base_command
+            try:
+                run_command(systemd_cmd)
```
narrieta, line 210: we need to discard stdout

mgunnala: I updated the command to suppress output altogether (line 197)

nagworld9 (approved):

tests/ga/test_signature_validation_sudo.py
```diff
@@ -77,3 +83,105 @@ def test_should_validate_signature_for_package_signed_with_leaf_root_cert(self):
         # Signature validation should still pass, because the signature was generated when the root certificate was unexpired.
         self.assertTrue(i_am_root(), "Test does not run when non-root")
         TestSignatureValidationSudo._validate_signature_in_another_year(2026, self.vm_access_zip_path, self.vm_access_signature, self.package_name_and_version)
+
+    def test_validate_signature_should_use_systemd_run(self):
+        self.assertTrue(i_am_root(), "Test does not run when non-root")
+        with patch("azurelinuxagent.ga.signature_validation_util.CGroupConfigurator.get_instance") as mock_get_instance:
```
narrieta:
```
        with patch("azurelinuxagent.ga.signature_validation_util.CGroupConfigurator._Impl.enabled", return_value=False):
```
instead of

```
        with patch("azurelinuxagent.ga.signature_validation_util.CGroupConfigurator.get_instance") as mock_get_instance:
            mock_instance = mock_get_instance.return_value
            mock_instance.enabled.return_value = False
```

Same in multiple instances below

tests/ga/test_signature_validation_sudo.py
```diff
@@ -77,3 +83,105 @@ def test_should_validate_signature_for_package_signed_with_leaf_root_cert(self):
...
+            original_popen = subprocess.Popen
+
+            def mock_openssl_failure(*args, **kwargs):
+                # Match: openssl cms -verify
+                cmd = ' '.join(args[0])
+                if self.openssl_cmd_pattern.search(cmd) is not None:
+                    # Simulate OpenSSL failure (unit name in stderr means it's NOT a systemd failure)
+                    error_msg = 'Running as unit: {0}\nVerification failure'.format(EXT_SIGNATURE_VALIDATION_CGROUPS_UNIT_NAME)
+                    proc = MagicMock()
+                    proc.communicate.return_value = (b"", error_msg.encode())
```
narrieta: this is too tightly coupled with the current implementation of run_command; see other instances of mocks of Popen and how they create command output elsewhere in the code

narrieta: more instances below

narrieta: see, for example, test_goal_state_should_try_legacy_cypher_and_then_fail_when_no_cyphers_are_supported_by_openssl

narrieta: see MockIpTables for another approach... mocking run_command itself is also acceptable (currently you are mocking the implementation of run_command)

azurelinuxagent/common/osutil/systemd.py
```diff
@@ -130,4 +132,43 @@ def is_unit_loaded(unit_name):
         value = get_unit_property(unit_name, "LoadState")
         return value.lower() == "loaded"
     except shellutil.CommandError:
-        return False
\ No newline at end of file
+        return False
+
+
+def is_systemd_run_failure(unit_name, stderr):
```
maddieford, line 138: Thanks for the nice comments :)

tests/ga/test_signature_validation_sudo.py
mgunnala: Updated

tests/ga/test_signature_validation_sudo.py
mgunnala: Updated to mock run_command() itself

narrieta (approved):

PR #3490, opened 2025-11-05
Do not delete Certificates.pem on certificate download failures

narrieta:
Agents older than 2.13.1.1 can go into an infinite loop during initialization of the Daemon if the certificates cannot be downloaded and the Certificates.pem file does not exist.

This PR ensures the file is created even if there is an error fetching/decrypting the certificates.

### DIFF
--- a/azurelinuxagent/common/protocol/goal_state.py
+++ b/azurelinuxagent/common/protocol/goal_state.py
@@ -525,21 +525,27 @@ class Certificates(LogEvent):
         pfx_file = os.path.join(conf.get_lib_dir(), PFX_FILE_NAME)
         pem_file = os.path.join(conf.get_lib_dir(), PEM_FILE_NAME)
 
+        create_empty_pem_file = False
+
         try:
             if not self._try_download_certificates_pfx(wire_client, uri, pfx_file):
+                create_empty_pem_file = True
                 return  # The response from the WireServer did not have any certificates, or they were not in the expected format
             self._convert_certificates_pfx_to_pem(pfx_file, pem_file)
         except Exception as e:
-            self.error(WALAEventOperation.GoalStateCertificates, "Error fetching the goal state certificates: {0}", ustr(e))
-            # Agents older than 2.13.1.1 can go into an infinite loop during initialization of the Daemon if the certificates cannot
-            # be downloaded and the PEM file does not exist. Create an empty file or overwrite any existing file from previous
-            # goal states.
-            open(pem_file, "w").close()
             # A failure to download the certificates won't necessarily produce an error. Certificates do not change often and they may have
-            # already been saved to disk on a previous goal state. We continue with an empty summary and later on we check whether the
-            # certificates needed by extensions are already on disk and refresh the goal state if they are not.
+            # already been saved to disk on a previous goal state. We simply report the error and continue processing the goal_state; later on,
+            # before extensions are processed, the Agent checks whether the required certificates are already on disk and refreshes the goal
+            # state if they are not.
+            self.error(WALAEventOperation.GoalStateCertificates, "Error fetching the goal state certificates: {0}", ustr(e))
+            create_empty_pem_file = True
             return
         finally:
+            if create_empty_pem_file:
+                # Agents older than 2.13.1.1 can go into an infinite loop during initialization of the Daemon if the certificates cannot
+                # be downloaded/decrypted and the PEM file does not exist. Create an empty file (or overwrite any existing file from previous
+                # goal states) to ensure it exists.
+                open(pem_file, "w").close()
             self._remove_file(pfx_file)
 
         try:
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -597,11 +597,7 @@ class UpdateHandler(object):
         # Check that the certificates needed by extensions are in the goal state certificates summary
         for extension in goal_state.extensions_goal_state.extensions:
             for settings in extension.settings:
-                if settings.protectedSettings is None:
-                    continue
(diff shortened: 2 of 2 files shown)


azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -522,23 +522,33 @@ def __init__(self, wire_client, uri, logger_):
...
+        pfx_file = os.path.join(conf.get_lib_dir(), PFX_FILE_NAME)
+        pem_file = os.path.join(conf.get_lib_dir(), PEM_FILE_NAME)
 
-            try:
-                pem_file = self._convert_certificates_pfx_to_pem(pfx_file)
-            finally:
-                self._remove_file(pfx_file)
+        try:
+            if not self._try_download_certificates_pfx(wire_client, uri, pfx_file):
+                return  # The response from the WireServer did not have any certificates, or they were not in the expected format
```
maddieford, line 533: Shouldn't the Pem also get created in this case? Or does daemon not care about PEM file if there aren't any certs in the correct format in the WireServer response

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -522,23 +522,33 @@ def __init__(self, wire_client, uri, logger_):
...
+            self._convert_certificates_pfx_to_pem(pfx_file, pem_file)
+        except Exception as e:
+            self.error(WALAEventOperation.GoalStateCertificates, "Error fetching the goal state certificates: {0}", ustr(e))
+            # Agents older than 2.13.1.1 can go into an infinite loop during initialization of the Daemon if the certificates cannot
+            # be downloaded and the PEM file does not exist. Create an empty file or overwrite any existing file from previous
+            # goal states.
+            open(pem_file, "w").close()
+            # A failure to download the certificates won't necessarily produce an error. Certificates do not change often and they may have
+            # already been saved to disk on a previous goal state. We continue with an empty summary and later on we check whether the
+            # certificates needed by extensions are already on disk and refresh the goal state if they are not.
```
maddieford: Previously the agent didn't check if the certs that are missing from the goal state are already on disk. Why update that behavior now? Is it just to give the agent the best chance to process the goal state?

narrieta: It just removes this extra message from the Agent log in some scenarios 

`event.warn(WALAEventOperation.FetchGoalState, "The extensions goal state is out of sync with the tenant cert. Certificate {0}, needed by {1}, is missing.", settings.certificateThumbprint, extension.name)`

Even after issuing that warning, the Agent just moves on after a retry and the extensions will find the certificate if it is already on disk, but seeing the difficulties people are having enabling FIPS, I thought it was worth eliminating that extra warning from the log to simplify debugging a little bit.

One such scenario is that the VM just opted-in to FIPS and enabled FIPS on the OS, but it has not forced a new PFX. The certificates from the previous PFX (before FIPS was enabled/opted-in) would still be on disk.

azurelinuxagent/common/protocol/goal_state.py
narrieta: Thanks, I think I'll add it.

2.7.0.6 returns early as well in these 2 cases. so it does not hit the infinite loop.

However, some corner scenarios may still go into the loop. E.g. a goal state with dsts certificates, followed by a goal state with on pfx format, followed by an agent restart.

nagworld9: I didn't follow corner case. If we reach this code means, daemon already started with pem file presence and then new ext_handler ensure pem file exist. So, agent restarts in any case should not make Deamon in loop.

narrieta: 2.7.0.6 returns early when the certificates are not in PFX format. It neither creates Certificates.pem, nor tries to use it.

Then comes this code that returns without creating an empty file.

Then new goal state on PFX format, followed by an Agent restart before the new extension handler has had a chance to process the new goal state.  Then 2.7.0.6 fails decryption and tries to use a Certificates.pem that does not exist.

Very unlikely, I would think, but addressing it is not difficult.

azurelinuxagent/common/protocol/goal_state.py
narrieta: I could simplify the check for missing certificates to use only the state on disk and forget about the certificates summary. The summary would only be used for logging purposes.

```
    def _check_certificates(goal_state):
        # Check that the certificates needed by extensions are in the goal state certificates summary
        for extension in goal_state.extensions_goal_state.extensions:
            for settings in extension.settings:
                if settings.protectedSettings is None:
                    continue
                certificates = goal_state.certs.summary
                if not any(settings.certificateThumbprint == c['thumbprint'] for c in certificates):
                    # check if the certificate was saved to disk on a previous goal state.
                    certificate_path = os.path.join(conf.get_lib_dir(), settings.certificateThumbprint + '.crt')
                    if not os.path.isfile(certificate_path):
                        event.warn(WALAEventOperation.FetchGoalState, "The extensions goal state is out of sync with the tenant cert. Certificate {0}, needed by {1}, is missing.", settings.certificateThumbprint, extension.name)
                        return False
```

maddieford: Makes sense, I forgot that the Agent just moves on after retry anyway

narrieta: I changed _check_certificates to check only the state on disk. Makes the code simpler.

azurelinuxagent/common/protocol/goal_state.py
narrieta: Created the empty file on early return

nagworld9 (approved):

PR #3493, opened 2025-11-06
Ignore records written by extensions when parsing waagent.log

narrieta:
Some extensions write messages to the Agent's log. When parsing the log file, those messages can get in the way of tests because their format is arbitrary. This PR attempts to skip those messages when parsing the log.

We identify those messages looking at the format of their timestamps (it does not match any of the formats used by the Agent). When the Agent executes an extension, it captures its stdout/stderr and logs them. If the extension outputs timestamps to its stdout/stderr, those messages are incorrectly identified as single log records, when they should be part of the log record emitted by the Agent. This PR also fixes this issue.

### DIFF
--- a/azurelinuxagent/common/protocol/goal_state.py
+++ b/azurelinuxagent/common/protocol/goal_state.py
@@ -522,23 +522,39 @@ class Certificates(LogEvent):
         self.summary = []
         self._crypt_util = CryptUtil(conf.get_openssl_cmd())
 
+        pfx_file = os.path.join(conf.get_lib_dir(), PFX_FILE_NAME)
+        pem_file = os.path.join(conf.get_lib_dir(), PEM_FILE_NAME)
+
+        create_empty_pem_file = False
+
         try:
-            pfx_file = self._download_certificates_pfx(wire_client, uri)
-            if pfx_file is None:  # The response from the WireServer may not have any certificates
-                return
-
-            try:
-                pem_file = self._convert_certificates_pfx_to_pem(pfx_file)
-            finally:
-                self._remove_file(pfx_file)
-
-            self.summary = self._extract_certificate(pem_file)
-
-            for c in self.summary:
-                self.info(WALAEventOperation.GoalStateCertificates, "Downloaded certificate {0}", c)
-
+            if not self._try_download_certificates_pfx(wire_client, uri, pfx_file):
+                create_empty_pem_file = True
+                return  # The response from the WireServer did not have any certificates, or they were not in the expected format
+            self._convert_certificates_pfx_to_pem(pfx_file, pem_file)
         except Exception as e:
+            # A failure to download the certificates won't necessarily produce an error. Certificates do not change often and they may have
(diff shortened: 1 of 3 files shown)


tests_e2e/tests/lib/agent_log.py
```diff
@@ -65,30 +65,12 @@ def from_dictionary(dictionary: Dict[str, str]):
 
     @property
     def timestamp(self) -> datetime:
-        # Extension logs may follow different timestamp formats
```
narrieta, line 68: Dealing with extension timestamps is no longer needed

Agent timestamps come in two flavors, though.

tests_e2e/tests/lib/agent_log.py
```diff
@@ -505,9 +487,7 @@ def is_error(r: AgentLogRecord) -> bool:
             # if they are coming from an extension
             for err in ['Exception', 'Traceback', '[CGW]']:
                 if err in r.message:
-                    extension_prefix_re = r'\[.+]'  # The prefix for extension-related messages is the name of the extension in brackets
-                    extension_message_re = r'Command:[^\n]+\n\[stdout]\n.*\n\[stderr].*'  # The message logged by the agent includes the extension command and its stdout and stderr
-                    if r.prefix is not None and re.match(extension_prefix_re, r.prefix) is not None and re.match(extension_message_re, r.message, re.DOTALL) is not None:
+                    if r.prefix is not None and self._EXTENSION_COMMAND_PREFIX.match(r.prefix) is not None and self._EXTENSION_COMMAND_MESSAGE.match(r.message, re.DOTALL) is not None:
```
narrieta, line 490: No changes here, I just defined _CONSTANTS for those 2 regexps

tests_e2e/tests/lib/agent_log.py
```diff
@@ -65,30 +65,12 @@ def from_dictionary(dictionary: Dict[str, str]):
...
-        # Extension logs may follow different timestamp formats
-        # 2023/07/10 20:50:13.459260
-        ext_timestamp_regex_1 = r"\d{4}/\d{2}/\d{2} \d{2}:\d{2}:\d{2}[.\d]+"
-
-        # 2023/07/10 20:50:13 OR 2023/07/10 20:50:1
-        # 2023/07/10 20:50 OR 2023/07/10 20:5
-        ext_timestamp_regex_2 = r"(\d{4}/\d{2}/\d{2} \d{2}:\d{1,2})(?::(\d{1,2}))?"
-
-        if re.match(ext_timestamp_regex_1, self.when):
+        # The timestamps emitted by the Agent can follow two formats: 2025-11-05T23:52:48.037713Z (newer) or 2025-11-05T23:52:48.790673Z (older)
```
maddieford: Is this comment correct? Looks like both formats are the same

narrieta: thanks; copy-n-paste error. should be 2025/11/07 00:56:06.858951 (older)

maddieford (approved):

PR #3497, opened 2025-12-01
Do not change firewall rules in waagent-network-setup.service when firewalld is enabled

narrieta:
When firewalld is not installed/running on the VM, the Agent will install the waagent-network-setup service to setup the firewall rules during boot. However, if firewalld is installed or started after the Agent has installed waagent-network-setup, both services will end up running concurrently during boot, which can lead to an incorrect firewall setup. The waagent-network-setup service should detect this condition and exit without setting up any rules. 

In addition, some RHEL 9 images in the marketplace include a set of stale firewall rules in /etc/firewalld/direct.xml. The test added in this PR verifies that those rules are handled appropriately.

### DIFF
--- a/azurelinuxagent/agent.py
+++ b/azurelinuxagent/agent.py
@@ -47,6 +47,7 @@ from azurelinuxagent.ga.logcollector import LogCollector, OUTPUT_RESULTS_FILE_PA
 from azurelinuxagent.common.osutil import get_osutil
 from azurelinuxagent.common.utils import fileutil, textutil
 from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
+from azurelinuxagent.common.utils.shellutil import run_command, CommandError
 from azurelinuxagent.common.version import AGENT_NAME, AGENT_LONG_VERSION, AGENT_VERSION, \
     DISTRO_NAME, DISTRO_VERSION, \
     PY_VERSION_MAJOR, PY_VERSION_MINOR, \
@@ -305,9 +306,22 @@ class Agent(object):
         threading.current_thread().name = "Firewall"
         event.info(event.WALAEventOperation.Firewall, "Setting up firewall after boot. Endpoint: {0}", ustr(endpoint))
         try:
+            try:
+                stdout = run_command(['systemctl', 'list-unit-files', '--type=service', '--no-legend', 'firewalld.service']).rstrip()
+                event.info(event.WALAEventOperation.Firewall, "Firewalld is installed (state: {0}). Will not setup the firewall rules.", stdout)
+                sys.exit(0)
+            except CommandError as command_error:
+                if command_error.returncode == 1 and (command_error.stdout, command_error.stderr) == ('', ''):
+                    pass  # Not installed, continue
+                else:
+                    raise
+        except Exception as error:
+            event.error(event.WALAEventOperation.Firewall, "Unable to determine whether firewalld is installed. Will not setup the firewall rules. Error: {0}", ustr(error))
+            sys.exit(1)
+        try:
             firewall_manager = FirewallManager.create(endpoint)
             firewall_manager.setup()
-            event.info(event.WALAEventOperation.Firewall, "Successfully set the firewall rules")
+            event.info(event.WALAEventOperation.Firewall, "Successfully set up the firewall rules:\n{0}", firewall_manager.get_state())
         except Exception as error:
             event.error(event.WALAEventOperation.Firewall, "Unable to add firewall rules. Error: {0}", ustr(error))
             sys.exit(1)
--- a/azurelinuxagent/agent.py
+++ b/azurelinuxagent/agent.py
@@ -301,29 +301,49 @@ class Agent(object):
                 log_collector_monitor.stop()
 
     @staticmethod
+    def _is_firewalld_enabled():
+        #
+        # Check first whether it's installed
+        #
+        try:
+            run_command(['systemctl', 'list-unit-files', '--type=service', '--no-legend', 'firewalld.service']).rstrip()
+        except CommandError as command_error:
+            if command_error.returncode == 1 and (command_error.stdout, command_error.stderr) == ('', ''):  # exit code 1 with no output means it is not installed
+                return False
+            raise
+        #
+        # Now check it is enabled
+        #
+        try:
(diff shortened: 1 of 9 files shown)


tests_e2e/orchestrator/lib/agent_test_loader.py
```diff
@@ -153,6 +153,9 @@ def images(self) -> Dict[str, List[VmImageInfo]]:
     # Matches a reference to a random subset of images within a set with an optional count: random(<image_set>, [<count>]), e.g. random(endorsed, 3), random(endorsed)
     RANDOM_IMAGES_RE = re.compile(r"random\((?P<image_set>[^,]+)(\s*,\s*(?P<count>\d+))?\)")
 
+    # Matches an image URN, which must consist of 4 components separated by ':', e.g. "Canonical:ubuntu-24_04-lts:server:latest"
```
narrieta, line 156: Added support for URNs to the -image parameter

tests_e2e/test_suites/firewalld.yml
```diff
@@ -0,0 +1,10 @@
+#
+# Set of tests around configuration of the firewalld and waagent-network-setup services
+#
+name: "Firewalld"
+tests:
+  - "firewalld/boot_conflicts.py"
+images:
+  - "centos_82"
+  - "oracle_95"
+  - "rhel_95"
```
narrieta, line 10: The issue at hand has been reported on RHEL and CentOS. Alma and Rocky are also derivatives of Red Hat, but the firewalld configuration is different, so I did not add them here.

azurelinuxagent/agent.py
```diff
@@ -303,13 +304,23 @@ def _validate_log_collector_cgroup_slice():
...
+            pass
+
         try:
             firewall_manager = FirewallManager.create(endpoint)
             firewall_manager.setup()
-            event.info(event.WALAEventOperation.Firewall, "Successfully set the firewall rules")
+            logger.info("Successfully set up the firewall rules:\n{0}", firewall_manager.get_state())
         except Exception as error:
-            event.error(event.WALAEventOperation.Firewall, "Unable to add firewall rules. Error: {0}", ustr(error))
+            logger.warn("Unable to add firewall rules. Error: {0}", ustr(error))
```
nagworld9: can we add current state when error happens? It helps if setup left undesirable state.

tests_e2e/tests/firewalld/boot_conflicts.py
```diff
@@ -0,0 +1,132 @@
...
+        self._context.vm.restart(wait_for_boot=True, ssh_client=self._ssh_client)
+        log.info("Boot completed")
+
+        waagent_log = self._wait_for_log_message(waagent_log_size, r"^[0-9TZ:.-]+ INFO Firewall Firewall", "Will not setup the firewall rules")
+        log.info(f"waagent-network-setup.service did not setup the firewall rules, as expected:\n{indent(waagent_log)}")
+
+        waagent_log = self._wait_for_log_message(waagent_log_size, r"legacy firewall rule.+--destination-port., .53.", "Removed legacy firewall rule")
+        log.info(f"The Agent removed the legacy firewall rule, as expected:\n{indent(waagent_log)}")
+
+        log.info("Checking permanent firewall rules...")
```
nagworld9, line 66: can you also verify we disable the waagent-network-setup.service at this point?

azurelinuxagent/agent.py
narrieta: thanks; updated

tests_e2e/tests/firewalld/boot_conflicts.py
narrieta: thanks; updated

nagworld9 (dismissed):

azurelinuxagent/agent.py
```diff
@@ -303,13 +304,28 @@ def _validate_log_collector_cgroup_slice():
...
         threading.current_thread().name = "Firewall"
-        event.info(event.WALAEventOperation.Firewall, "Setting up firewall after boot. Endpoint: {0}", ustr(endpoint))
+        logger.info("Setting up firewall during boot. Endpoint: {0}", ustr(endpoint))
+
+        try:
+            run_command(['systemctl', 'is-enabled', '--type=service', 'firewalld.service']).rstrip()
+            logger.info("Firewalld is enabled. Will not setup the firewall rules to avoid conflicts.")
+            sys.exit(0)
+        except CommandError:
+            # Differences across versions of systemd make hard to determine whether the command failed because firewalld is not installed
```
maddieford, line 314: Is it worth logging and/or collecting telem on the exception so we can see the types of failures we're getting?

tests_e2e/tests/firewalld/boot_conflicts.py
```diff
@@ -0,0 +1,138 @@
...
+    def _wait_for_log_message(self, offset: int, selector_re: str, message: str) -> str:
+        log.info(f"Checking waagent.log starting at offset {offset}")
+
+        limit = datetime.now() + timedelta(minutes=5)
+
+        while True:
+            waagent_log = self._ssh_client.run_command(r"tail --bytes=+{0} /var/log/waagent.log | tr -d '\000' | grep -E '{1}' || true".format(offset, selector_re))  # some tests write NULL characters to the log; we delete them
+            if re.search(message, waagent_log) is not None:
+                return waagent_log
+            if datetime.now() > limit - timedelta(seconds=30):
```
maddieford: is `timedelta(seconds=30)` intentional? or should it be `timedelta(seconds=15)` to match the sleep time

azurelinuxagent/agent.py
narrieta: not worth it; issue is that exit code 1 can happen because it is not enabled, not installed, or some other reason (e.g. failure to contact dbus). differences across versions of systemd make difficult attempting to parse any error message

for this purpose, the logic is good enough as-is: if we are certain firewalld is enabled (i.e. command success) we quit, otherwise we continue just as we doing before and set up the firewall.

tests_e2e/tests/firewalld/boot_conflicts.py
narrieta: thanks; updated

maddieford (approved):

PR #3498, opened 2025-12-02
DowngradeFix value of AZURE_GUEST_AGENT_UPDATING_FROM_VERSION on downgrades

narrieta:
In a downgrade of CSE from 1.3.26 to 1.3.18, the Agent invoked 1.3.26.update() with  
 
    "VERSION": "1.3.18"  
    "AZURE_GUEST_AGENT_UPDATING_FROM_VERSION": "1.3.18"  
    "AZURE_GUEST_AGENT_EXTENSION_VERSION": "1.3.26"  

According to the spec (Publisher's guide)

Name | Description
-- | --
AZURE_GUEST_AGENT_EXTENSION_VERSION | The current version of the extension
VERSION | Target version to be updated to
AZURE_GUEST_AGENT_UPDATING_FROM_VERSION | Old plugin version updating/downgrading from

So the correct values should have been

"VERSION": "1.3.18"
"AZURE_GUEST_AGENT_UPDATING_FROM_VERSION": "1.3.26"
"AZURE_GUEST_AGENT_EXTENSION_VERSION": "1.3.26"

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -929,15 +929,19 @@ class ExtHandlersHandler(object):
                     func=partial(old_ext_handler_i.disable, extension=old_ext))
 
         ext_handler_i.copy_status_files(old_ext_handler_i)
+        # Invoke the update method of the greater version, both on upgrades and downgrades
         if ext_handler_i.version_gt(old_ext_handler_i):
-            ext_handler_i.update(disable_exit_codes=disable_exit_codes,
-                                 updating_from_version=old_ext_handler_i.ext_handler.version,
-                                 extension=extension)
+            ext_handler_i.update(
+                updating_from_version=old_ext_handler_i.ext_handler.version,
+                updating_to_version=ext_handler_i.ext_handler.version,
+                extension=extension,
+                disable_exit_codes=disable_exit_codes)
         else:
-            updating_from_version = ext_handler_i.ext_handler.version
-            old_ext_handler_i.update(handler_version=updating_from_version,
-                                     disable_exit_codes=disable_exit_codes, updating_from_version=updating_from_version,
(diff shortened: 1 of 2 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -929,15 +929,19 @@ def execute_old_handler_command_and_return_if_succeeds(func):
...
+                updating_from_version=old_ext_handler_i.ext_handler.version,
+                updating_to_version=ext_handler_i.ext_handler.version,
+                extension=extension,
+                disable_exit_codes=disable_exit_codes)
         else:
-            updating_from_version = ext_handler_i.ext_handler.version
-            old_ext_handler_i.update(handler_version=updating_from_version,
-                                     disable_exit_codes=disable_exit_codes, updating_from_version=updating_from_version,
-                                     extension=extension)
+            old_ext_handler_i.update(
```
nagworld9, line 940: is update operation supposed to happen on greater version? or version we update to? The reason I ask we do update on new version for upgrade and old version for downgrade?

narrieta: Greater version per spec. 

I know, it is counter-intuitive, but the rationale is that the greater version has the more recent code and knows about the behavior of previous version, while previous versions do not know about the behavior of new versions.

I made some small changes in the code and associated tests to make this behavior cleare.

narrieta: spec

> Azure VM Agent **will invoke the update command on the handler with the higher version** under a separate process with ADMINISTRATIVE privileges. During update the handler has an opportunity to transfer any state information from the previous handler. VM VM Agent will wait 15 mins for update command to finish.

nagworld9 (approved):

maddieford (approved):

Issue #160, opened 2016-03-09
hostname will not set by dhcp after waagent -deprovision


Fooobar23:
Hi,

I created a new capture. First I do waagent -deprovision, stop the vm, generalize the machine and create a capture. When I created a new VM from the capture vhd, the hostname is always localhost. What going wrong?

I tested it with ubuntu 14.04 lts and debian 8.

karataliu:
what's output for 'grep -i hostname /var/log/waagent.log'?

Fooobar23:
root@localhost:~# grep -i hostname /var/log/waagent.log
root@localhost:~#

karataliu:
1. Would a rebooting solve the issue?

2. What's 'Provisioning.MonitorHostName' configuration in /etc/waagent.conf, you'll need set it to y to get the hostname setup properly:
```
Provisioning.MonitorHostName=y
```

Fooobar23:
Hey,

1. no, the same behavior after reboot. 

2. yes, i have configured Provisioning.MonitorHostName=y

karataliu:
What's the VM OS type you're referring to here?

- For Debian, it follows previous discussion.
- For Ubuntu, [cloud-init] (https://launchpad.net/cloud-init) is responsible for setting the hostname.
So for debugging, you can consider checking following information.
```
grep -i hostname /var/log/cloud-init.log
```

Fooobar23:
~# grep -i hostname /var/log/cloud-init.log
2016-03-15 09:28:02,642 - util.py[DEBUG]: Running command hostname with allowed return codes [0] (shell=False, capture=True)
2016-03-15 09:28:02,644 - util.py[DEBUG]: Running command ['hostname', u'TEMPLATE-UBUNTU'] with allowed return codes [0] (shell=False, capture=True)
2016-03-15 09:28:02,646 - DataSourceAzure.py[DEBUG]: pubhname: publishing hostname [hostname=TEMPLATE-UBUNTU policy=True interface=eth0]
2016-03-15 09:28:03,065 - util.py[DEBUG]: publishing hostname took 0.418 seconds (0.41)
2016-03-15 09:28:04,744 - util.py[DEBUG]: Running command ['hostname', 'ubuntu'] with allowed return codes [0] (shell=False, capture=True)
2016-03-15 09:28:05,182 - importer.py[DEBUG]: Looking for modules ['cc_set_hostname', 'cloudinit.config.cc_set_hostname'] that have attributes ['handle']
[...]

PR #3499, opened 2025-12-02
Set soft limit for agents' memory cgroup in v2 machines

nagworld9:
## Description

Set soft limit for agents' memory cgroup in v2 machines to 300 MB as an initial value and also added code to track memory pressure events. This will help evaluate memory usage metrics and pressure events to understand how the agent performs under the limit. Based on the findings, we will define the next plan of action for a long-term solution.

Issue # 
---

### PR information

---

### DIFF
--- a/azurelinuxagent/ga/cgroupapi.py
+++ b/azurelinuxagent/ga/cgroupapi.py
@@ -159,6 +159,31 @@ class CGroupUtil(object):
             return "unknown"
 
     @staticmethod
+    def get_current_memory_quota(unit_name):
+        """
+        Returns the memory quota for the given unit in bytes, or 'infinity' or 'unknown' if not set.
+        """
+        try:
+            mem_quota = systemd.get_unit_property(unit_name, "MemoryHigh").strip().lower()
+            if mem_quota == "infinity":
+                return mem_quota  # No limit on memory usage
+            elif mem_quota.endswith("k"):
+                mem_quota_bytes = int(mem_quota[:-1]) * 1024
+            elif mem_quota.endswith("m"):
+                mem_quota_bytes = int(mem_quota[:-1]) * 1024 * 1024
+            elif mem_quota.endswith("g"):
+                mem_quota_bytes = int(mem_quota[:-1]) * 1024 * 1024 * 1024
+            elif mem_quota.endswith("t"):
+                mem_quota_bytes = int(mem_quota[:-1]) * 1024 * 1024 * 1024 * 1024
+            else:
+                mem_quota_bytes = mem_quota
+
+            return str(mem_quota_bytes)
+        except Exception as e:
+            log_cgroup_warning("Error parsing current MemoryHigh: {0}".format(ustr(e)))
+            return "unknown"
+
+    @staticmethod
     def has_cpu_quota(unit_name):
         """
         Returns True if quota set for the unit.
@@ -298,6 +323,12 @@ class _SystemdCgroupApi(object):
         raise NotImplementedError()
 
     def can_enforce_cpu(self):
+        """
+        Cgroup version specific. Returns if controller can be used for enforcement
+        """
+        raise NotImplementedError()
+
+    def can_enforce_memory(self):
         """
         Cgroup version specific. Returns if controller can be used for enforcement
         """
@@ -509,6 +540,9 @@ class SystemdCgroupApiv1(_SystemdCgroupApi):
     def can_enforce_cpu(self):
         return CgroupV1.CPU_CONTROLLER in self._cgroup_mountpoints
 
+    def can_enforce_memory(self):
+        return False
+
 
 class SystemdCgroupApiv2(_SystemdCgroupApi):
     """
@@ -623,6 +657,9 @@ class SystemdCgroupApiv2(_SystemdCgroupApi):
 
     def can_enforce_cpu(self):
         return CgroupV2.CPU_CONTROLLER in self._controllers_enabled_at_root
+
+    def can_enforce_memory(self):
+        return CgroupV2.MEMORY_CONTROLLER in self._controllers_enabled_at_root
 
 
 class Cgroup(object):
--- a/azurelinuxagent/ga/cgroupapi.py
+++ b/azurelinuxagent/ga/cgroupapi.py
@@ -161,7 +161,7 @@ class CGroupUtil(object):
     @staticmethod
     def get_current_memory_quota(unit_name):
         """
-        Returns the memory quota for the given unit in bytes, or 'infinity' or 'unknown' if not set.
+        Returns the memory quota for the given unit in bytes, or 'infinity' if not set, or 'unknown' if an error occurs.
         """
         try:
             mem_quota = systemd.get_unit_property(unit_name, "MemoryHigh").strip().lower()
@@ -180,7 +180,7 @@ class CGroupUtil(object):
 
             return str(mem_quota_bytes)
         except Exception as e:
-            log_cgroup_warning("Error parsing current MemoryHigh: {0}".format(ustr(e)))
+            log_cgroup_warning("Error in getting current MemoryHigh: {0}".format(ustr(e)))
             return "unknown"
 
     @staticmethod
--- a/azurelinuxagent/ga/cgroupconfigurator.py
+++ b/azurelinuxagent/ga/cgroupconfigurator.py
@@ -36,6 +36,7 @@ from azurelinuxagent.common.version import get_distro
 from azurelinuxagent.common.utils import shellutil, fileutil
 from azurelinuxagent.ga.extensionprocessutil import handle_process_completion
 from azurelinuxagent.common.event import add_event, WALAEventOperation
+from azurelinuxagent.ga.resourcequota import CpuQuota, MemoryQuota, ResourceName
 
 AZURE_SLICE = "azure.slice"
 _AZURE_SLICE_CONTENTS = """
@@ -186,9 +187,10 @@ class CGroupConfigurator(object):
                     for prop in controller.get_unit_properties():
                         log_cgroup_info('Agent {0} unit property value: {1}'.format(prop, systemd.get_unit_property(systemd.get_agent_unit_name(), prop)))
                     if isinstance(controller, _CpuController) and self._cgroups_api.can_enforce_cpu():
-                        self._set_cpu_quota(agent_unit_name, conf.get_agent_cpu_quota())
+                        self._set_resource_quota(agent_unit_name, ResourceName.CPU, conf.get_agent_cpu_quota())
                         controller.track_throttle_time(True)  # CPU controller track the throttle time only when CPU quota is set
-                    elif isinstance(controller, _MemoryController):
+                    elif isinstance(controller, _MemoryController) and self._cgroups_api.can_enforce_memory():
+                        self._set_resource_quota(agent_unit_name, ResourceName.MEMORY, conf.get_agent_memory_quota())
                         self._agent_memory_metrics = controller
                     CGroupsTelemetry.track_cgroup_controller(controller)
 
@@ -201,7 +203,7 @@ class CGroupConfigurator(object):
                 if self._cgroups_api is not None and not self._cgroups_api.can_enforce_cpu():
                     # If agent cgroups are not enabled or quotas not enabled, reset the quota for the agent unit
                     log_cgroup_info("Reset CPU quota if agent cgroups were not enabled for enforcement")
-                    self._reset_cpu_quota(systemd.get_agent_unit_name())
+                    self._reset_resource_quota(systemd.get_agent_unit_name(), ResourceName.CPU, ignore_enforce_check=True)
 
         def _check_cgroups_supported(self):
             distro_supported = CGroupUtil.distro_supported()
@@ -355,7 +357,7 @@ class CGroupConfigurator(object):
                     if len(files_to_cleanup) > 0:
                         log_cgroup_info("Found drop-in files; attempting agent cgroup setup cleanup", send_event=False)
                         self._cleanup_all_files(files_to_cleanup)
-                        self._reset_cpu_quota(systemd.get_agent_unit_name())
+                        self._reset_resource_quota(systemd.get_agent_unit_name(), ResourceName.CPU, ignore_enforce_check=True)
 
             except Exception as err:
                 logger.warn("Error while resetting the quotas: {0}".format(err))
@@ -448,7 +450,7 @@ class CGroupConfigurator(object):
             try:
                 if disable_cgroups == DisableCgroups.ALL:  # disable all
                     # Reset quotas
-                    self._reset_cpu_quota(systemd.get_agent_unit_name())
+                    self._reset_resource_quota(systemd.get_agent_unit_name(), ResourceName.ALL)
                     extension_services = self.get_extension_services_list()
                     for extension in extension_services:
                         log_cgroup_info("Resetting extension : {0} and it's services: {1} Quota".format(extension, extension_services[extension]), send_event=False)
@@ -458,7 +460,7 @@ class CGroupConfigurator(object):
                     self._agent_cgroups_enabled = False
                     self._extensions_cgroups_enabled = False
                 elif disable_cgroups == DisableCgroups.AGENT:  # disable agent
-                    self._reset_cpu_quota(systemd.get_agent_unit_name())
+                    self._reset_resource_quota(systemd.get_agent_unit_name(), ResourceName.ALL)
                     agent_controllers = self._agent_cgroup.get_controllers()
                     for controller in agent_controllers:
                         if isinstance(controller, _CpuController):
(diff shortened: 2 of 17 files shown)


azurelinuxagent/ga/resourcequota.py
```diff
@@ -0,0 +1,64 @@
...
+    2. property: The cgroup property name associated with the resource quota
+    3. format: A function that formats the quota value setting
+    4. can_enforce: A function that checks if the resource quota can be enforced
+    5. get_current_quota: A function that retrieves the current quota setting
+    """
+    def __init__(self, name, property_name, format_func, can_enforce_func, get_current_func):
+        self.name = name
+        self.property = property_name
+        self.format = format_func
+        self.can_enforce = can_enforce_func
```
narrieta: why use function pointers instead of inherited/overridden methods?

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -158,6 +158,31 @@ def get_current_cpu_quota(unit_name):
             log_cgroup_warning("Error parsing current CPUQuotaPerSecUSec: {0}".format(ustr(e)))
             return "unknown"
 
+    @staticmethod
+    def get_current_memory_quota(unit_name):
+        """
+        Returns the memory quota for the given unit in bytes, or 'infinity' or 'unknown' if not set.
```
narrieta:
```suggestion
        Returns the memory quota for the given unit in bytes, or 'infinity' if not set, or 'unknown' if an error occurrs.
```

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -158,6 +158,31 @@ def get_current_cpu_quota(unit_name):
...
+            elif mem_quota.endswith("g"):
+                mem_quota_bytes = int(mem_quota[:-1]) * 1024 * 1024 * 1024
+            elif mem_quota.endswith("t"):
+                mem_quota_bytes = int(mem_quota[:-1]) * 1024 * 1024 * 1024 * 1024
+            else:
+                mem_quota_bytes = mem_quota
+
+            return str(mem_quota_bytes)
+        except Exception as e:
+            log_cgroup_warning("Error parsing current MemoryHigh: {0}".format(ustr(e)))
```
narrieta: an exception could be raised by errors other than parsing (e.g. failure to connect to the dbus)

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -158,6 +158,31 @@ def get_current_cpu_quota(unit_name):
             log_cgroup_warning("Error parsing current CPUQuotaPerSecUSec: {0}".format(ustr(e)))
             return "unknown"
 
+    @staticmethod
+    def get_current_memory_quota(unit_name):
+        """
+        Returns the memory quota for the given unit in bytes, or 'infinity' or 'unknown' if not set.
+        """
+        try:
+            mem_quota = systemd.get_unit_property(unit_name, "MemoryHigh").strip().lower()
```
narrieta, line 167: you should probably get the value from the cgroups filesystem to avoid errors invoking systemctl

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -470,45 +472,81 @@ def disable(self, reason, disable_cgroups):
             except Exception as exception:
                 log_cgroup_warning("Error disabling cgroups: {0}".format(ustr(exception)))
 
-        def _set_cpu_quota(self, unit_name, quota):
-            """
-            Sets CPU quota to the given percentage (100% == 1 CPU)
+        def _get_resource_quotas(self):
+            if self._cgroups_api is None:
+                return []
+            return [CpuQuota(self._cgroups_api), MemoryQuota(self._cgroups_api)]
```
narrieta: a named tuple may be a better choice (instead of an array)

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -470,45 +472,81 @@ def disable(self, reason, disable_cgroups):
...
-
-        def _reset_cpu_quota(self, unit_name):
+            Sets the quota for the given resource type ('CPU', 'Memory', or 'All').
             """
-            Removes any CPUQuota on the agent
+            try:
+                quotas = self._get_resource_quotas()
+                property_names = []
+                values = []
+                if resource_type == ResourceName.ALL:
```
narrieta: let's standardize in one of 'type' or 'name'. mixing them makes the code confusing

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -470,45 +472,81 @@ def disable(self, reason, disable_cgroups):
...
+                            if current != value:
+                                log_cgroup_info("Setting {0}'s {1} to {2}".format(unit_name, rq.property, value))
+                                systemd.set_unit_run_time_property(unit_name, rq.property, value)
+                            break
+            except Exception as exception:
+                log_cgroup_warning('Failed to set resource quota: {0}'.format(ustr(exception)))
 
-            NOTE: This resets the quota on the agent's default dropin file; any local overrides on the VM will take precedence
-            over this setting.
+        def _reset_resource_quota(self, unit_name, resource_type, ignore_enforce_check=False):
```
narrieta: what is the meaning of 'ignore_enforce_check'? (when does one to set it to True (or False)?

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -470,45 +472,81 @@ def disable(self, reason, disable_cgroups):
...
-            """
-            Sets CPU quota to the given percentage (100% == 1 CPU)
+        def _get_resource_quotas(self):
+            if self._cgroups_api is None:
+                return []
+            return [CpuQuota(self._cgroups_api), MemoryQuota(self._cgroups_api)]
 
-            NOTE: This is done using a systemtcl set-property --runtime; any local overrides in /etc folder on the VM will take precedence
-            over this setting.
+        def _set_resource_quota(self, unit_name, resource_type, quota):
```
narrieta: explain what 'quota' is - apparently it can be a dictionary

azurelinuxagent/ga/cgroupconfigurator.py
narrieta: you can coalesce the blocks of code in if and then into a single block with a little refactoring; there is too much code repetition

narrieta: similar comment for the reset function

azurelinuxagent/ga/memorycontroller.py
```diff
@@ -158,6 +158,39 @@ def get_memory_usage(self):
...
+                        percentage = float(match.group(1))
+                        time_spent = 300 * (percentage / 100)
+                        return round(time_spent,  2)
+        except (IOError, OSError) as e:
+            if e.errno != errno.ENOENT:  # File is not present in some distros, so we do not raise in that case
+                raise CGroupsException("Failed to read memory.pressure: {0}".format(ustr(e)))
+        except Exception as e:
+            raise CGroupsException("Failed to read memory.pressure: {0}".format(ustr(e)))
+
+        return float(0)
```
narrieta, line 194: how do we distinguish file not present versus actual value of 0?

nagworld9: I'm not sending telemetry if value 0 anyway as its too verbose and not providing value. So, I don't need to distinguish between file not present and actual value 0

azurelinuxagent/ga/cgroupconfigurator.py
nagworld9: we have scenario where we enabled agent cgroup feature for rhel7 at one point and later we disabled due to kernel bug. When we disable, we need to reset quotas without checking enforce is enabled or not.  In that case, it will be True

azurelinuxagent/ga/resourcequota.py
nagworld9: There is no separate logic that I need to implement in inherited/overridden methods. It just function calls which vary. I don't get your question if it's an issue

azurelinuxagent/ga/cgroupconfigurator.py
nagworld9: When resource quota called for all, then we expect quota in dictionary with resource_name and quota. If it's for specific resource, it just value. I'll add better comments.

azurelinuxagent/ga/cgroupconfigurator.py
nagworld9: I;m not sure how they work

azurelinuxagent/ga/cgroupapi.py
nagworld9: sure, you meant update the error message? Like

`Error in getting current MemoryHigh`

azurelinuxagent/ga/cgroupapi.py
nagworld9: We are using systemctl all over the place for different properties. Changing for this will not solve general problem. That will be a TODO and required refactoring to do it for all

maddieford, line 167: is MemoryHigh a valid unit property in v1? What would this return for a v1 system?

azurelinuxagent/ga/cgroupapi.py
```diff
@@ -158,6 +158,31 @@ def get_current_cpu_quota(unit_name):
...
+            if mem_quota == "infinity":
+                return mem_quota  # No limit on memory usage
+            elif mem_quota.endswith("k"):
```
maddieford, line 170: is this case insensitive? If not, should the check be case insensitive or will unit always be lowercase?

azurelinuxagent/ga/cgroupapi.py
narrieta: my take is that we should not make an existing problem worse and start by using the file system here

but if you want to address all the instance on a different PR, that is ok to me, hopefully that PR coming soon

azurelinuxagent/ga/cgroupapi.py
narrieta: yes, sorry, i meant the message. new one lgtm

azurelinuxagent/ga/cgroupconfigurator.py
narrieta: we have some samples in the code, but they are not a good option for your usage, since apparently you are using the return value just to iterate thru the items

the design of this function makes it harder for the caller to use (in the sense that it is easy to introduce bugs if one uses the incorrect index to access an item). Your current usage does not have this issue, but future callers may

azurelinuxagent/ga/cgroupconfigurator.py
narrieta: ok, please document it

azurelinuxagent/ga/memorycontroller.py
narrieta: maybe not today, but future callers may
if you want to leave it as is, please document that behavior

azurelinuxagent/ga/resourcequota.py
narrieta: Type safety, less chance of introducing code bugs

```
class ResourceQuota(object):
    def __init__(self, name, property_name):
        self.name = name
        self.property = property_name

    def format(self, quota):
        return "{0}%".format(quota)

    def can_enforce(self):
        raise NotImplementedError()

    def get_current_quota(self, unit_name):
        raise NotImplementedError()


class CpuQuota(ResourceQuota):
    def __init__(self, cgroups_api):
        super(CpuQuota, self).__init__(ResourceName.CPU, "CPUQuota")
        self._cgroups_api = cgroups_api

    def can_enforce(self):
        return self._cgroups_api.can_enforce_cpu()
    
    def get_current_quota(self, unit_name):
        return CGroupUtil.get_current_cpu_quota(unit_name)


class MemoryQuota(ResourceQuota):
    def __init__(self, cgroups_api):
        super(MemoryQuota, self).__init__(ResourceName.MEMORY, "MemoryHigh")
        self._cgroups_api = cgroups_api

    def can_enforce(self):
        return self._cgroups_api.can_enforce_memory()

    def get_current_quota(self, unit_name):
        return CGroupUtil.get_current_memory_quota(unit_name)

```

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -201,7 +203,7 @@ def initialize(self):
                 if self._cgroups_api is not None and not self._cgroups_api.can_enforce_cpu():
```
maddieford, line 203: Could you remind me why we do this check? Do we need to do something similar to reset the memory limit?

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -355,7 +357,7 @@ def _reset_agent_cgroup_setup(self):
                     if len(files_to_cleanup) > 0:
                         log_cgroup_info("Found drop-in files; attempting agent cgroup setup cleanup", send_event=False)
                         self._cleanup_all_files(files_to_cleanup)
-                        self._reset_cpu_quota(systemd.get_agent_unit_name())
+                        self._reset_resource_quota(systemd.get_agent_unit_name(), ResourceName.CPU, ignore_enforce_check=True)
```
maddieford, line 363: Same question here, do we need to reset memory limit in this case?

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -448,7 +450,7 @@ def disable(self, reason, disable_cgroups):
             try:
                 if disable_cgroups == DisableCgroups.ALL:  # disable all
                     # Reset quotas
-                    self._reset_cpu_quota(systemd.get_agent_unit_name())
+                    self._reset_resource_quota(systemd.get_agent_unit_name(), ResourceName.ALL)
```
maddieford: nit: seems like we always reset resource quota, should we move it above the if/elif conditions?

azurelinuxagent/ga/cgroupapi.py
nagworld9: Using filesystem is tricky here, because we need to know cgroups paths. I can't make this code work with simple fix.  It's required a refactoring.

azurelinuxagent/ga/cgroupconfigurator.py
maddieford: Would it make more sense for quota to always be a dict? If only setting one thing between cpu or memory, then there would only be one value in the dict. That way we don't need to pass ResourceName in either

azurelinuxagent/ga/cgroupapi.py
narrieta: I see; sounds good as-is

narrieta: (but let's improve the code later)

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -733,6 +773,23 @@ def _report_agent_cgroups_procs(agent_cgroup_proc_names, unexpected):
...
 
+        @staticmethod
+        def _report_agent_cgroups_memory_usage(cgroup_metrics):
+            """
+            Reports the agent's current memory usage when it exceeds the configured limit.
+            """
+            limit_in_bytes = conf.get_agent_memory_quota()
+            current_usage = 0
+            for metric in cgroup_metrics:
+                if metric.counter == MetricsCounter.TOTAL_MEM_USAGE:
```
maddieford: could we combine this if/elif into one if condition?

```suggestion
                if metric.counter == MetricsCounter.TOTAL_MEM_USAGE or metric.counter == MetricsCounter.SWAP_MEM_USAGE:
```

azurelinuxagent/ga/cgroupcontroller.py
```diff
@@ -40,12 +39,12 @@ class MetricValue(object):
...
     """
 
-    def __init__(self, category, counter, instance, value, report_period=_DEFAULT_REPORT_PERIOD):
+    def __init__(self, category, counter, instance, value, report_period=None):
         self._category = category
         self._counter = counter
         self._instance = instance
         self._value = value
-        self._report_period = report_period
+        self._report_period = timedelta(seconds=conf.get_cgroup_check_period()) if report_period is None else report_period
```
maddieford, line 47: why make this change? Doesn't keeping report_period=_DEFAULT_REPORT_PERIOD achieve the same thing?

azurelinuxagent/ga/memorycontroller.py
```diff
@@ -158,6 +158,39 @@ def get_memory_usage(self):
...
+        return properties
+
+    def get_memory_pressure_events(self):
+        """
+        We use the share of time in which processes of the cgroup have experienced memory pressure.
+        :return: Total time some processes stalled due to memory pressure in last 300 seconds
+        :rtype: int
+        """
+        try:
+            with open(os.path.join(self.path, 'memory.pressure')) as memory_events:
```
maddieford:
```suggestion
            with open(os.path.join(self.path, 'memory.pressure')) as memory_pressure:
```

azurelinuxagent/ga/memorycontroller.py
```diff
@@ -158,6 +158,39 @@ def get_memory_usage(self):
         # In v2, cache memory is reported in the 'file' counter
         return self._get_memory_stat_counter("anon"), self._get_memory_stat_counter("file")
 
+    def get_unit_properties(self):
+        properties = super(MemoryControllerV2, self).get_unit_properties()
+        properties.append("MemoryHigh")
+        return properties
+
+    def get_memory_pressure_events(self):
```
maddieford, line 166: Will the agent reset the quota if process experienced too much time under memory pressure? Or are we just collecting this for telem only right now

azurelinuxagent/ga/cgroupapi.py
nagworld9: Property not error from systemctl query if its not present in v1.

azurelinuxagent/ga/cgroupapi.py
nagworld9: I already convert it to lowercase in line #160

azurelinuxagent/ga/cgroupconfigurator.py
nagworld9: This we added workaround when distro introduced a patch where they set the limits before we enable the feature.

azurelinuxagent/ga/cgroupconfigurator.py
nagworld9: Memory not needed, since this if for cleanup old setup files and we never add memory before

azurelinuxagent/ga/cgroupconfigurator.py
nagworld9: not sure If I followed. If it always dict and no resouceName as the key, then how that works?

azurelinuxagent/ga/cgroupcontroller.py
nagworld9: Actually, this is a bug. This _DEFAULT_REPORT_PERIOD global loads during module imports, before config loaded from the waagent.conf. As a result, _DEFAULT_REPORT_PERIOD gets a default value even config is present in the file. We addressed similar issue in other places where we reference config in the global imports, I missed here somehow.

azurelinuxagent/ga/memorycontroller.py
nagworld9: collecting telem right now

azurelinuxagent/ga/cgroupapi.py
maddieford: Would `systemd.get_unit_property(unit_name, "MemoryHigh")` raise exception in that case?

azurelinuxagent/ga/cgroupconfigurator.py
maddieford: Could we add that as a comment anywhere we reset cpu but not mem?

azurelinuxagent/ga/cgroupconfigurator.py
maddieford: Sorry I meant you wouldn't need to pass in resource_type parameter. resourceName would still be the key for the dict 

If only setting cpu, then quota would dict with a single key/value pair like this: {ResourceName.CPU: cpu_quota)

If only setting memory, then quota would be a dict with a single key/value pair like this:{ResourceName.MEMORY: memory_quota}

If setting both, then quota would be a dict with multiple key/value pairs like this: {ResourceName.CPU: cpu_quota, ResourceName.MEMORY: memory_quota}

If you do it this way, I don't think you would need the resource_type parameter

azurelinuxagent/ga/cgroupcontroller.py
maddieford: Would changing the default value of report_period from the global to `conf.get_cgroup_check_period()` fix the issue?

 `def __init__(self, category, counter, instance, value, report_period=conf.get_cgroup_check_period()):`

azurelinuxagent/ga/memorycontroller.py
maddieford: Just curious, why not reset the quota if too much time under memory pressure? Is it because we don't know how much time under memory pressure is 'too much' time without collecting telemetry first?

azurelinuxagent/ga/cgroupconfigurator.py
nagworld9: I added because set and reset_quota will have similar signature. Quota is additional parameter that we need to pass for set. I can simplify like you mentioned

azurelinuxagent/ga/cgroupapi.py
nagworld9: yes, you get commanderror from shellutil and we catch that.

azurelinuxagent/ga/cgroupcontroller.py
nagworld9: My understanding defaults also gets loaded during import module time.

azurelinuxagent/ga/memorycontroller.py
nagworld9: yes. If we reset, will lose the goal of it

azurelinuxagent/ga/cgroupapi.py
maddieford: Do we ever call this function on v1 machines? I'm wondering if we will get misleading telemetry 'error parsing current MemoryHigh' on those v1 machines. 

Or is the only called on v2 machines?

nagworld9: We call this method during set/rese and we have protection against enforcement check, I'll double check.

nagworld9: I verified get function not called in v1

azurelinuxagent/ga/cgroupconfigurator.py
nagworld9: updated

azurelinuxagent/ga/memorycontroller.py
nagworld9: added

narrieta (dismissed):

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -470,45 +475,70 @@ def disable(self, reason, disable_cgroups):
...
+                values = []
+                for rq in quotas:
+                    if rq.can_enforce() and rq.name in new_quotas:
+                        q = new_quotas.get(rq.name)
+                        value = rq.format(q)
+                        current = rq.get_current_quota(unit_name)
+                        if current != value:
+                            property_names.append(rq.property)
+                            values.append(value)
+                if property_names:
```
maddieford: Can we be more explicit with this check? 

```suggestion
                if len(property_names) > 0:
```

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -470,45 +475,70 @@ def disable(self, reason, disable_cgroups):
...
+                quotas = self._get_resource_quotas(resource_name)
+                property_names = []
+                values = []
+                for rq in quotas:
+                    if ignore_enforce_check or rq.can_enforce():
+                        current = rq.get_current_quota(unit_name)
+                        if current != "infinity":
+                            property_names.append(rq.property)
+                            values.append("")  # systemd convention
+                if property_names:
```
maddieford:
```suggestion
                if len(property_names) > 0:
```

maddieford (approved):

narrieta (approved):

PR #3501, opened 2025-12-09
Fixing CSE deletion in ext_update scenario

nagworld9:
## Description
In ext_update scenario, we make CSE update fail to test some scenario and at the end we clean up cse, so that other tests won't carry this CSE behavior. But cleanup logic has issue, before delete we check if extension that is to be deleted on the vm. Test use full extension name but get extension on vm query returns resource_name of the extension, causing a mismatch.

Issue # 
---

### PR information

---

### DIFF
--- a/tests_e2e/tests/scripts/ext_update-modify_handler_manifest.py
+++ b/tests_e2e/tests/scripts/ext_update-modify_handler_manifest.py
@@ -35,7 +35,7 @@ def main():
     operation = parser.add_argument_group('Main operation')
     operation_parser = operation.add_mutually_exclusive_group(required=True)
     operation_parser.add_argument('--properties', dest='properties', nargs='+', help='List of property=value to update in the handlerManifest file')
-    operation_parser.add_argument("--reset", dest='reset', help='Reset the handlerManifest file to the default values')
+    operation_parser.add_argument("--reset", dest='reset', action='store_true', help='Reset the handlerManifest file to the default values')
 
     args, _ = parser.parse_known_args()
     extension_name = args.extension_name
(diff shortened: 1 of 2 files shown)


narrieta:
@nagworld9 could you add what the issue was to the description? thanks

nagworld9:
> @nagworld9 could you add what the issue was to the description? thanks

done

narrieta (approved):

tests_e2e/tests/scripts/ext_update-modify_handler_manifest.py
```diff
@@ -35,7 +35,7 @@ def main():
     operation = parser.add_argument_group('Main operation')
     operation_parser = operation.add_mutually_exclusive_group(required=True)
     operation_parser.add_argument('--properties', dest='properties', nargs='+', help='List of property=value to update in the handlerManifest file')
-    operation_parser.add_argument("--reset", dest='reset', help='Reset the handlerManifest file to the default values')
+    operation_parser.add_argument("--reset", dest='reset', action='store_true', help='Reset the handlerManifest file to the default values')
```
maddieford, line 38: Is the new action arg used anywhere?

nagworld9: what do you mean? do we have samples in the code or something else you asking?

nagworld9: If you are checking for samples, we are using it in check-agent-log.py, get_goal_state.py etcc

maddieford: What does it do? Why was it added here

nagworld9: if command-line flag is present, it set as true or if absent, default to False. If I don't use action, I need to pass some value to --reset flag, which does not make sense.

It seems cleanup code never called and --reset code never called before

maddieford: I see, thanks for the explanation

maddieford (approved):

PR #3504, opened 2025-12-12
Use "iptables -L" to verify the results of "iptables -C"

narrieta:
Fix for "iptables -C" not being able to detect the DROP rule in some environments. More details in the comment for the IpTables.check() method.

The PR also includes some refactoring that I started doing when I was planning to change the way we delete the firewall rules. It makes the code easier to work with and a little more readable.

### DIFF
--- a/azurelinuxagent/agent.py
+++ b/azurelinuxagent/agent.py
@@ -36,7 +36,7 @@ from azurelinuxagent.ga import logcollector, cgroupconfigurator
 from azurelinuxagent.ga.cgroupcontroller import AGENT_LOG_COLLECTOR
 from azurelinuxagent.ga.cpucontroller import _CpuController
 from azurelinuxagent.ga.cgroupapi import create_cgroup_api, InvalidCgroupMountpointException
-from azurelinuxagent.ga.firewall_manager import FirewallManager
+from azurelinuxagent.ga.firewall_manager import FirewallManager, IpTables
 
 import azurelinuxagent.common.conf as conf
 import azurelinuxagent.common.event as event
@@ -317,6 +317,13 @@ class Agent(object):
 
         try:
             firewall_manager = FirewallManager.create(endpoint)
+            if isinstance(firewall_manager, IpTables):
+                try:
+                    logger.info("Attempting to load the conntrack module (iptables -C should not find any matching rules)...")
+                    logger.info("{0}", firewall_manager.load_conntrack())
+                except Exception as e:
+                    logger.warn("Failed to load the conntrack module: {0}", ustr(e))
+
         except Exception as error:
             logger.warn("{0}", ustr(error))
             sys.exit(1)
--- a/azurelinuxagent/agent.py
+++ b/azurelinuxagent/agent.py
@@ -319,7 +319,11 @@ class Agent(object):
             firewall_manager = FirewallManager.create(endpoint)
             if isinstance(firewall_manager, IpTables):
                 try:
-                    logger.info("Attempting to load the conntrack module (iptables -C should not find any matching rules)...")
+                    #
+                    # We execute "iptables -C -m conntrack" to force loading of the conntrack module with the intention of avoiding the
+                    # issue described in IpTables.check().
+                    #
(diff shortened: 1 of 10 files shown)


azurelinuxagent/ga/firewall_manager.py
```diff
@@ -138,88 +157,98 @@ def _get_state_command(self):
...
+        #
+        # Each item in the array is a tuple with the friendly name of the rule and a function that returns the command used to process that rule. This function
+        # takes as argument the option that is passed to the corresponding command  (-A, -C, and -D for iptables, and --passthrough, --query-passthrough, and --remove-passthrough
+        # for firewallcmd)
+        #
+        self._firewall_commands = [
+            (FirewallManager.ACCEPT_DNS, self._get_accept_dns_rule_command),
+            (FirewallManager.ACCEPT, self._get_accept_rule_command),
+            (FirewallManager.DROP, self._get_drop_rule_command)
+        ]
```
narrieta, line 179: This list used to be returned by the __get_commands() generator method (now deleted). Using an array allows more flexibility.

azurelinuxagent/ga/firewall_manager.py
```diff
@@ -138,88 +157,98 @@ def _get_state_command(self):
...
 
-    def remove(self):
-        existing_rules = []
+    @property
+    def _append(self):
+        """
+        Command-line option to append a firewall rule.
+        """
+        raise NotImplementedError()
```
narrieta, line 186: These properties used to be methods. Properties make the code more readable.

azurelinuxagent/agent.py
```diff
@@ -317,6 +317,17 @@ def setup_firewall(endpoint):
...
         try:
             firewall_manager = FirewallManager.create(endpoint)
+            if isinstance(firewall_manager, IpTables):
+                try:
+                    #
+                    # We execute "iptables -C -m conntrack" to force loading of the conntrack module with the intention of avoiding the
+                    # issue described in IpTables.check().
+                    #
+                    logger.info("Attempting to load the conntrack module (iptables -C should not find any matching rules, so that error can be ignored)...")
+                    logger.info("{0}", firewall_manager.load_conntrack())
```
nagworld9, line 327: will this work running -C command on drop rule will load conntrack module? I guess not because it didn't happen when agent used -C command in the main loop. If it had happened, env thread would have been successful in detecting drop rule

azurelinuxagent/ga/env.py
```diff
@@ -116,16 +116,19 @@ def _operation(self):
             try:
                 if self._firewall_manager.check():
                     return  # The firewall is configured correctly
-                self._report(event.warn, "The firewall has not been setup. Will set it up.")
+                self._report(event.warn, WALAEventOperation.Firewall, "The firewall has not been setup. Will set it up.")
+            except IptablesInconsistencyError as e:
```
nagworld9, line 120: Once we are in this condition, we report log every 30 seconds as env firewall setup checks every 30 seconds. Telemetry flood with this log until conntrack module issue resolved. Should we change logging period?

narrieta: self._report takes care of that. see the sample log that i posted in our internal chat

azurelinuxagent/agent.py
narrieta: the idea here is to load the module before creating (-A) the rule . see the sample log I posted in our internal chat

azurelinuxagent/ga/env.py
nagworld9: ok, I see...only 3 times reported until next period.Thanks

azurelinuxagent/agent.py
nagworld9: ok, -A will not load the module and only -C does that?

narrieta: the first command that uses the module will load it, but if it is -A or -I, the rule will not be visible, so we use -C to load it

nagworld9 (approved):

maddieford (approved):

PR #3508, opened 2025-12-16
Skip firewall tests when the Proxy Agent manages the WireServer endpoint

narrieta:
The GPA intercepts the HTTP request before they reach the firewall, so in that case these tests are not relevant.

### DIFF
--- a/tests_e2e/tests/lib/retry.py
+++ b/tests_e2e/tests/lib/retry.py
@@ -110,6 +110,8 @@ def retry(operation: Callable[[], Any], attempts: int = 5, delay: int = 30) -> A
         attempts -= 1
         try:
             return operation()
+        except AssertionError:
+            raise
         except Exception as e:
             if attempts == 0:
                 raise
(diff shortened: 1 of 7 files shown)


tests_e2e/tests/lib/retry.py
```diff
@@ -110,6 +110,8 @@ def retry(operation: Callable[[], Any], attempts: int = 5, delay: int = 30) -> A
         attempts -= 1
         try:
             return operation()
+        except AssertionError:
```
narrieta, line 113: don't retry on test failure

tests_e2e/tests/scripts/agent_persist_firewall-verify_firewall_rules_on_boot.py
```diff
@@ -70,17 +70,21 @@ def cron_log_checks():
         if not os.path.exists(cron_log):
             raise Exception("Cron log file not found: {0}".format(cron_log))
         with open(cron_log) as f:
-            cron_logs_lines = list(map(lambda _: _.strip(), f.readlines()))
-        if not cron_logs_lines:
+            cron_log_content = f.read()
```
narrieta, line 73: including the log of the cron job in the log of the test makes debugging a lot easier/faster

maddieford (approved):

nagworld9 (approved):

Issue #3510, opened 2025-12-17
[BUG] firewalld.service fails due to missing modules


litian1992:
WALA invokes a few firewall rules which depend on certain kernel modules. e.g. "-m owner" and "-m conntrack" rely on xt_owner and xt_conntrack. These modules reside in kernel-modules-extra package for distros like RHEL.
Because kernel-modules-extra is not a dependency of iptables or firewalld, these rules can fail.
Culprit: https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/ga/firewall_manager.py#L377

This is not an issue previously because kernel package pulls in kernel-modules-extra with kernel-modules-extra-matched which is a dependency of iptables. Unfortunately this is not the case in term of UKI.

Example failure messages:
```
firewalld[8037]: 
ERROR: COMMAND_FAILED: Direct: '/usr/sbin/iptables-restore -w -n' failed: Warning: Extension tcp revision 0 not supported, missing kernel module? 
Warning: Extension owner revision 0 not supported, missing kernel module? 
Warning: Extension conntrack revision 0 not supported, missing kernel module? 
iptables-restore v1.8.11 (nf_tables): 
line 3: RULE_APPEND failed (No such file or directory): rule in chain OUTPUT 
line 4: RULE_APPEND failed (No such file or directory): rule in chain OUTPUT 
```

Therefore, WALA should add some checks/fallbacks when these kernel modules don't exist.

narrieta:
@litian1992 Thanks for reporting this. Sounds like kernel-modules-extra should be a dependency of the waagent package.

litian1992:
>  Sounds like kernel-modules-extra should be a dependency of the waagent package.

I'm not certain of that. The 'kernel-modules-extra' package in distro like RHEL has a hard dependency on 'kernel-modules'. Neither of which are desired in a UKI system. We are discussing internally of moving those mentioned modules to modules-extra-matched. But fallbacks seem reasonable for WALA.

narrieta:
@litian1992 let me try to start an internal conversation with our contacts in Red Hat, I'll add you to it.

litian1992:
> We are discussing internally of moving those mentioned modules to modules-extra-matched.

Scratch this. I amended the fallback logic for both iptables and firewall-cmd to nftables and network-setup service respectively in the PR. Hope it helps.

[...]

PR #3511, opened 2025-12-17
firewall_manager: Add fallbacks when missing kernel modules

litian1992:
## Description
There are firewall rules invoked without checking the existence of the dependent kernel modules, e.g. xt_owner and xt_conntrack. These modules reside in kernel-modules-extra in distros like RHEL. The kernel-modules-extra package is not a dependency of iptables in terms of UKI. Thus the existence deserves checking. In case they are not present, fallback to nftables from iptables; to network-setup service from firewall-cmd respectively.

Issue #3510
---

### PR information

---

### DIFF
--- a/azurelinuxagent/ga/firewall_manager.py
+++ b/azurelinuxagent/ga/firewall_manager.py
@@ -446,6 +446,14 @@ class FirewallCmd(_FirewallManagerIndividualRules):
     def __init__(self, wire_server_address):
         super(FirewallCmd, self).__init__(wire_server_address)
 
+        # firewall-cmd --permanent --direct passthrough uses iptables-style rules; the stack relies on a working
+        # iptables userland, so probe listing rules before we depend on FirewallCmd.
+        try:
+            shellutil.run_command(["iptables", "-L"], log_error=False)
+        except Exception as exception:
+            if isinstance(exception, OSError) and exception.errno == errno.ENOENT:  # pylint: disable=no-member
+                raise FirewallManagerNotAvailableError("iptables is not available")
+
         try:
             self._version = shellutil.run_command(["firewall-cmd", "--version"]).strip()
         except Exception as exception:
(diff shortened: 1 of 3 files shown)


azurelinuxagent/ga/firewall_manager.py
```diff
@@ -446,6 +446,14 @@ class FirewallCmd(_FirewallManagerIndividualRules):
...
         super(FirewallCmd, self).__init__(wire_server_address)
 
+        # firewall-cmd --permanent --direct passthrough uses iptables-style rules; the stack relies on a working
+        # iptables userland, so probe listing rules before we depend on FirewallCmd.
+        try:
+            shellutil.run_command(["iptables", "-L"], log_error=False)
+        except Exception as exception:
+            if isinstance(exception, OSError) and exception.errno == errno.ENOENT:  # pylint: disable=no-member
+                raise FirewallManagerNotAvailableError("iptables is not available")
+
```
narrieta, line 456: Relly sorry for the huge delay.

But my understanding is that firewalld could also be based on nft, can't it? We want to use firewalld instead of our custom service installed by _setup_network_setup_service. This would not use implementation of firewalld based on nft.

azurelinuxagent/ga/persist_firewall_rules.py
```diff
@@ -132,7 +132,14 @@ def setup(self):
         # In case of a failure, this would throw. In such a case, we don't need to try to setup our custom service
         # because on system reboot, all iptables rules are reset by firewalld.service, so it would be a no-op.
         # setup permanent firewalld rules
-        firewall_manager = FirewallCmd(self._dst_ip)
+        try:
+            firewall_manager = FirewallCmd(self._dst_ip)
```
nagworld9, line 136: Before we initialize FirewallCmd, we check firewalld service is running in the vm L#123. is it not enough check to switch to custom service setup?

maddieford:
We are seeing distros starting to publish images where the required iptables modules are missing. Since we have not gotten any response on this PR in several months, I have opened a new PR to address the issue: #3671

PR #3526, opened 2025-12-31
Improve telemetry for firewall setup

narrieta:
Added a periodic report of the status of the firewall rules.

Also, if the status is incorrect, we use a "verbose" mode where all the commands related to the firewall are logged, along with their output.

### DIFF
--- a/azurelinuxagent/ga/env.py
+++ b/azurelinuxagent/ga/env.py
@@ -105,7 +105,7 @@ class FirewallState(object):
     OK = "OK"  # The firewall rules for the WireServer are setup correctly
     NotSet = "NotSet"  # The firewall rules have not been set
     Invalid = "Invalid"  # The state of the firewall rules is not as expected, e.g. because some rules are missing
-    Inconsistent = "Inconsistent"  # The stare of the firewall is reported differently by different tools, e.g. "iptables -C" vs "iptables -L"
+    Inconsistent = "Inconsistent"  # The state of the firewall is reported differently by different tools, e.g. "iptables -C" vs "iptables -L"
     Unknown = "Unknown"  # The state of the firewall has not been checked yet
 
 
--- a/azurelinuxagent/ga/env.py
+++ b/azurelinuxagent/ga/env.py
@@ -32,6 +32,7 @@ from azurelinuxagent.ga.firewall_manager import FirewallManager, FirewallStateEr
 from azurelinuxagent.common.future import ustr
 from azurelinuxagent.ga.interfaces import ThreadHandlerInterface
 from azurelinuxagent.common.osutil import get_osutil
+from azurelinuxagent.common.utils import textutil
 from azurelinuxagent.common.protocol.util import get_protocol_util
 from azurelinuxagent.common.version import AGENT_NAME
 from azurelinuxagent.ga.periodic_operation import PeriodicOperation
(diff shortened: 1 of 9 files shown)


azurelinuxagent/ga/env.py
```diff
@@ -100,46 +101,89 @@ def _get_dhcp_client_pid(self):
         return pid
 
 
+class FirewallState(object):
+    OK = "OK"  # The firewall rules for the WireServer are setup correctly
+    NotSet = "NotSet"  # The firewall rules have not been set
+    Invalid = "Invalid"  # The state of the firewall rules is not as expected, e.g. because some rules are missing
+    Inconsistent = "Inconsistent"  # The stare of the firewall is reported differently by different tools, e.g. "iptables -C" vs "iptables -L"
```
nagworld9: nit: state

azurelinuxagent/ga/env.py
```diff
@@ -100,46 +101,89 @@ def _get_dhcp_client_pid(self):
...
+            #
+            self._firewall_manager.verbose = self._firewall_state != FirewallState.OK and self._should_report
+
             try:
                 if self._firewall_manager.check():
-                    return  # The firewall is configured correctly
+                    self._update_firewall_state(FirewallState.OK)
+                    self._report(event.info, WALAEventOperation.Firewall, "The firewall is configured correctly. Current state:\n{0}", self._firewall_manager.get_state())
+                    return
+                self._update_firewall_state(FirewallState.NotSet)
```
nagworld9, line 145: This verbose logging happens on next iteration, is itnt? if previous iteration detects that firewall state not ok, then next iteration logs the commands that we run.

If so, I think this will not help right as we might have already done corrective action to reset rules.  Let me know if I'm missing something here.

azurelinuxagent/ga/env.py
```diff
@@ -100,46 +101,89 @@ def _get_dhcp_client_pid(self):
...
+            self._update_reporting_state()
+
             if self._firewall_manager is None:
                 self._firewall_manager = FirewallManager.create(self._wire_server_address)
 
+            #
+            # Setting up the Firewall Manager to verbose will make it log each command it executes, along with its output. We do this only when the state of the firewall
+            # is not OK and the report limit has not been reached.
+            #
+            self._firewall_manager.verbose = self._firewall_state != FirewallState.OK and self._should_report
```
nagworld9, line 138: On first instance, verbose supposed to be True. So that if `self._firewall_manager.check()`: detects firewall sate no ok, we should log those as well

azurelinuxagent/ga/env.py
narrieta: thanks; fixed

azurelinuxagent/ga/env.py
narrieta: see reply to below comment

azurelinuxagent/ga/env.py
narrieta: Before we execute the commands, we do not know whether they will succeed or fail, so we start with verbose turned off. Once we hit a bad configuration of the firewall, we turn on verbose mode. The next iteration should give us more information about the commands executed and their result.

azurelinuxagent/ga/env.py
```diff
@@ -100,46 +101,89 @@ def _get_dhcp_client_pid(self):
...
 class EnableFirewall(PeriodicOperation):
+    _REPORTING_PERIOD = datetime.timedelta(hours=24)  # we set a limit on the number of messages logged within this period
+
     def __init__(self, wire_server_address):
         super(EnableFirewall, self).__init__(conf.get_enable_firewall_period())
         self._wire_server_address = wire_server_address
         self._firewall_manager = None  # initialized on demand in the _operation method
-        self._message_count = 0
-        self._report_after = datetime.datetime.now(UTC)
+        self._firewall_state = FirewallState.OK
```
maddieford: Should we initialize to Unknown state since it hasn't been ch.cked yet?

```
class FirewallState(object):
    ...
    Unknown = "Unknown"  # The state of the firewall has not been checked yet
```

azurelinuxagent/ga/env.py
narrieta: Also, note that the first setup, either by waagent-network-setup or the main thread, is verbose.

azurelinuxagent/ga/env.py
narrieta: I initialized it to OK to avoid turning on verbose mode in the first iteration:

`self._firewall_manager.verbose = self._firewall_state != FirewallState.OK and self._should_report
`

I can add a comment to that respect in the initialization

nagworld9: I feel first iteration may be critical given the existing duplicate cases we have seen so far, where env thread sees different result first time than we expected

narrieta: the initial setup is verbose, the initial periodic check is not - if the condition persists, the 2nd iteration of the env thread should produce verbose output

narrieta: @maddieford - added comment

nagworld9 (approved):

maddieford (approved):

PR #3533, opened 2026-01-09
Do not fetch goal state from RemoteAccessHandler

narrieta:
This PR continues the refactoring to detach the goal state from the WireClient (eliminate the calls to WireClient.get_goal_state())

Instead of the RemoteAccessHandler getting the goal state, the UpdateHandler passes it the remote access section of the goal state.

### DIFF
--- a/azurelinuxagent/common/protocol/goal_state.py
+++ b/azurelinuxagent/common/protocol/goal_state.py
@@ -479,6 +479,7 @@ class GoalState(object):
 
             remote_access = None
             if GoalStateProperties.RemoteAccessInfo & self._goal_state_properties:
+                container = find(xml_doc, "Container")
                 remote_access_uri = findtext(container, "RemoteAccessInfo")
                 if remote_access_uri is not None:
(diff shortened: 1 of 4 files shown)


azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -479,6 +479,7 @@ def _fetch_full_wire_server_goal_state(self, incarnation, xml_doc):
 
             remote_access = None
             if GoalStateProperties.RemoteAccessInfo & self._goal_state_properties:
+                container = find(xml_doc, "Container")
```
narrieta, line 482: While updating the unit tests, I noticed that this code depends on the 'container' initialized at line 441. If GoalStateProperties.RoleConfig is not requested, that variable will not be initialized (it turns out that PyCharm was also flagging this issue, but I had not noticed the warning before).

maddieford (approved):

nagworld9 (approved):

PR #3534, opened 2026-01-12
Decouple initialization of the Transport certificate from protocol detection

narrieta:
Refactoring the initialization of the Transport certificate out of protocol detection. Now initialization is done by the UpdateHandler as part of extension processing.

### DIFF
--- a/azurelinuxagent/ga/update.py
+++ b/azurelinuxagent/ga/update.py
@@ -1,3 +1,4 @@
+
 # Windows Azure Linux Agent
 #
 # Copyright 2018 Microsoft Corporation
@@ -48,7 +49,7 @@ from azurelinuxagent.common.protocol.goal_state import GoalStateSource, TRANSPOR
 from azurelinuxagent.common.protocol.hostplugin import HostPluginProtocol, VmSettingsNotSupported
 from azurelinuxagent.common.protocol.restapi import VERSION_0
 from azurelinuxagent.common.protocol.util import get_protocol_util
-from azurelinuxagent.common.protocol.wire import TransportCertificateError
+from azurelinuxagent.common.protocol.wire import WireProtocol, TransportCertificateError
 from azurelinuxagent.common.utils import shellutil
 from azurelinuxagent.common.utils.archive import StateArchiver, AGENT_STATUS_FILE
 from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
@@ -367,17 +368,7 @@ class UpdateHandler(object):
             # Initialize the goal state; some components depend on information provided by the goal state and this
             # call ensures the required info is initialized (e.g. telemetry depends on the container ID.)
             #
-            # Also, refresh the transport certificate each time the Agent starts.
-            #
-            protocol = self.protocol_util.get_protocol(save_to_history=True)
-            try:
-                protocol.create_transport_certificate()
-            except TransportCertificateError as e:
-                # Report the error and continue execution; we retry in the main loop
-                message = ustr(e)
-                logger.warn(message)
-                add_event(op=WALAEventOperation.TransportCertificate, message=message, is_success=False, log_event=False)
-
+            protocol = self.protocol_util.get_protocol(init_goal_state=False)
             self._initialize_goal_state(protocol)
 
             # Initialize the common parameters for telemetry events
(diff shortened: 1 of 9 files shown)


azurelinuxagent/ga/update.py
```diff
@@ -366,7 +367,16 @@ def run(self, debug=False):
...
+            # Also, refresh the transport certificate each time the Agent starts.
+            #
             protocol = self.protocol_util.get_protocol(save_to_history=True)
+            try:
+                protocol.create_transport_certificate()
+            except TransportCertificateError as e:
+                # Report the error and continue execution; we retry in the main loop
+                message = ustr(e)
+                logger.warn(message)
+                add_event(op=WALAEventOperation.TransportCertificate, message=message, is_success=False, log_event=False)
```
narrieta: In the original implementation, a failure to create the Transport certificate would send the Agent initialization into an infinite loop. In the new implementation, we try once during initialization and then retry as part of the main goal state loop. Failure to create the certificate will still block processing of the goal state. Similarly to the way to allow execution to continue when decryption fails, this should not be a blocking error either. This would require additional refactoring that I will do on a later PR.

maddieford (dismissed):

azurelinuxagent/ga/update.py
nagworld9: This add_event not helpful as initialize_event happens after protocol detection, it will not map to the vm at this point.

azurelinuxagent/ga/update.py
```diff
@@ -547,6 +557,17 @@ def _try_update_goal_state(self, protocol):
...
         try:
+            #
+            # Ensure the transport certificate exists (in case we could not create it during Agent initialization)
+            #
+            trans_cert_file = os.path.join(conf.get_lib_dir(), TRANSPORT_CERT_FILE_NAME)
+            if not os.path.exists(trans_cert_file):
+                try:
+                    protocol.create_transport_certificate()
+                except TransportCertificateError as e:
+                    add_event(op=WALAEventOperation.TransportCertificate, message=ustr(e), is_success=False, log_event=False)
```
nagworld9: This event will flood telemetry if it happens in the loop. In the outside exception, we already sending telemetry and that has protection to log max errors. Probably we can remove here as we log this error msg there as well

azurelinuxagent/ga/update.py
narrieta: That is a fair point but note that the previous implementation had the same problem, and moreover, the current implementation loses all telemetry emitted before we initialize the first goal state (which initializes the container ID) and then it is not fully associated with the VM (resource group, vmid, role instance name) until we initialize telemetry.

To help address that, we should check each event before pushing it to WireServer and assign values to any uninitialized fields.

For now, I'm moving this code to UpdateHandler._initialize_goal_state and adding the above comments.

azurelinuxagent/ga/update.py
narrieta: Thanks; that is exactly the reason I added this here (to re-use the reporting logic) and then I forgot to remove this event. Will remove it.

azurelinuxagent/ga/update.py
```diff
@@ -483,11 +484,25 @@ def _log_openssl_info():
...
+        # container ID will be populated only after the goal state is actually fetched, and the Resource Group, VM Name, and VM ID will be
+        # populated only after we initialize telemetry (which happens after this function returns).
+        #
+        # To address issue, we should populate any uninitialized fields before pushing the telemetry events to the WireServer.
         #
         event.info(WALAEventOperation.GoalState, "Initializing the goal state...")
+
+        try:
+            event.info(WALAEventOperation.TransportCertificate, "Refreshing/creating the Transport Certificate.")
+            WireProtocol.create_transport_certificate()
```
nagworld9, line 499: can we use existing protocol object passed in _initialize_goal_state as we do it in line 570?

narrieta: the whole point is to detach create transport from protocol detection. using a protocol instance requires detection

narrieta: i'll change 570 to use the static method as well

azurelinuxagent/ga/update.py
```diff
@@ -547,6 +562,13 @@ def _try_update_goal_state(self, protocol):
         max_errors_to_log = 3
 
         try:
+            #
+            # Ensure the transport certificate exists
+            #
+            trans_cert_file = os.path.join(conf.get_lib_dir(), TRANSPORT_CERT_FILE_NAME)
```
nagworld9, line 568: why don't we check other file(private) exists? don't we need to attempt to create if thats also not present?

narrieta: we create both at the same time using 1 single command

nagworld9 (approved):

maddieford (approved):

PR #3538, opened 2026-01-15
Fix CVM unit tests failures

mgunnala:
## Description

Issue # 
---

This PR updates the unit test mock protocol to correctly mock IMDS HTTP calls and cleans up state after test_confidential_vm_info.py. 
### PR information

---

### DIFF
--- a/tests/ga/test_confidential_vm_info.py
+++ b/tests/ga/test_confidential_vm_info.py
@@ -21,15 +21,8 @@ from azurelinuxagent.ga.confidential_vm_info import ConfidentialVMInfo
 from tests.lib.tools import AgentTestCase, MagicMock, patch, data_dir
 
 
+@patch('azurelinuxagent.ga.confidential_vm_info.ConfidentialVMInfo._is_confidential_vm', None)
 class TestConfidentialVMInfo(AgentTestCase):
-
-    def setUp(self):
-        ConfidentialVMInfo._security_type = None
-        AgentTestCase.setUp(self)
-
-    def tearDown(self):
-        ConfidentialVMInfo._is_confidential_vm = None
-        AgentTestCase.tearDown(self)
 
     @staticmethod
     def _setup_mock_imds_from_file(mock_get_metadata, file_path):
--- a/tests/ga/test_confidential_vm_info.py
+++ b/tests/ga/test_confidential_vm_info.py
@@ -27,6 +27,10 @@ class TestConfidentialVMInfo(AgentTestCase):
         ConfidentialVMInfo._security_type = None
         AgentTestCase.setUp(self)
 
+    def tearDown(self):
+        ConfidentialVMInfo._is_confidential_vm = None
+        AgentTestCase.tearDown(self)
+
     @staticmethod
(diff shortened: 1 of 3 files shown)


tests/ga/test_confidential_vm_info.py
```diff
@@ -27,6 +27,10 @@ def setUp(self):
         ConfidentialVMInfo._security_type = None
         AgentTestCase.setUp(self)
 
+    def tearDown(self):
+        ConfidentialVMInfo._is_confidential_vm = None
```
narrieta: why does it have to be setup to None on setup and teardown?

tests/ga/test_extension.py
```diff
@@ -4369,6 +4372,9 @@ def test_should_enable_existing_zip_package_if_manifest_validation_fails(self):
 
     def test_should_not_validate_signature_on_non_cvm(self):
         self.patch_is_cvm.stop()
+        # Initialize CVM info to False to simulate a non-CVM
+        ConfidentialVMInfo._is_confidential_vm = False
```
narrieta: should be done with a mock within a 'with' statement or try/finally

tests/lib/wire_protocol_data.py
```diff
@@ -249,8 +249,12 @@ def mock_http_get(self, url, *_, **kwargs):
                 response_headers = [('ETag', self.etag)]
                 self.prev_etag = self.etag
             self.call_counts["vm_settings"] += 1
-        elif '{0}/metadata/compute'.format(IMDS_ENDPOINT) in url:
-            content = json.dumps(self.imds_info.get("compute", "{}"))
+        elif '{0}/metadata/instance/compute'.format(IMDS_ENDPOINT) in url:
+            # Handle standard IMDS '/metadata/instance/compute' endpoint used by agent
+            compute_data = self.imds_info.get("compute", {}).copy()
+            if "securityProfile" not in compute_data:
+                compute_data["securityProfile"] = {"securityType": "ConfidentialVM"}
```
narrieta: why hardcode to confidential VM?

tests/ga/test_confidential_vm_info.py
narrieta: It may be better to do this with a mock and add a decorator to the entire class, or if done manually, seems to me setup needs to capture the initial value, set it to None (assuming that is what all the tests in this class), and restore the initial value on teardown

mgunnala: Thanks for the idea, I've updated this to add a decorator to the class and set the value of ConfidentialVMInfo._is_confidential_vm to None.

tests/ga/test_extension.py
mgunnala: Updated, thanks

tests/lib/wire_protocol_data.py
mgunnala: It doesn't need to be confidential VM, I just used this as a default value. I've updated this to add a "security_type" attribute to WireProtocolData and set the default value to an empty string instead, following the pattern of the other attributes in the class.

narrieta (approved):

nagworld9 (approved):

PR #3542, opened 2026-01-26
Use passlib module instead of crypt on Python >= 3.13

narrieta:
The crypt module was removed on Python 3.13. The fallbacks we were using (legacycrypt, crypt-r) are not available on some distros, such as Ubuntu. Using passlib instead.


narrieta:
FYI @mirespace

.gitignore
```diff
@@ -90,4 +90,7 @@ ENV/
 
 # pyenv
 .python-version
-.vscode/
\ No newline at end of file
+.vscode/
+
+# vscode
+.github/copilot-instructions.md
```
narrieta, line 96: for now, i am only experimenting with this, so added the file to the ignore list temporarily

maddieford (approved):

nagworld9 (approved):

PR #3548, opened 2026-01-30
Cgroup2fs filesystem detection in cgroups api check

nagworld9:
## Description
Distro Ubuntu 25.10 ship Rust-based coreutils. Its 'stat' lacks a human-readable mapping for the cgroup v2 (cgroup2fs) magic number, so it prints UNKNOWN with that magic value 0x63677270. So, adding an explicit check for this hex to detect cgroup v2
```

 # stat -f --format=%T /sys/fs/cgroup/
  UNKNOWN (0x63677270)
```
Issue # 
---

### PR information

---

### DIFF
--- a/tests_e2e/tests/agent_cgroups/agent_cgroups_process_check.py
+++ b/tests_e2e/tests/agent_cgroups/agent_cgroups_process_check.py
@@ -42,6 +42,9 @@ class AgentCgroupsProcessCheck(AgentVmTest):
         3. Restart the ext_handler process to re-initialize the cgroups setup
         4. Verify that agent detects extension processes and will not enable the cgroups
         """
+        if self._ssh_client.get_distro() == "ubuntu_2510":
+            log.info("Skipping test on ubuntu_2510 as AMA extension is not supported to test this scenario")
+            return
 
         log.info("=====Validating agent cgroups process check")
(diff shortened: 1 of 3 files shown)


tests_e2e/tests/agent_cgroups/agent_cgroups_process_check.py
```diff
@@ -42,6 +42,9 @@ def run(self):
         3. Restart the ext_handler process to re-initialize the cgroups setup
         4. Verify that agent detects extension processes and will not enable the cgroups
         """
+        if self._ssh_client.get_distro() == "ubuntu_2510":
```
maddieford, line 45: Should we add ubuntu_2510 to the unsupported_distros list in vm_extension_identifier and then skip on any distro where AMA is not supported?

nagworld9: I didn't want to do global change and make it to endorsed distro list in this pr. That's why i added explicit check. If we want to make global change, we should probably add ubuntu_25 to endorsed list and run all other tests as well.

maddieford: Makes sense

maddieford (approved):

narrieta (approved):

PR #3550, opened 2026-02-03
Improve artifact download failure telemetry

maddieford:
## Description
In a previous Canary release to make HGAP the primary artifact download channel, I noticed some gaps in the download failure telemetry. This PR address two gaps in the telemetry:

1. When HGAP returns a 400 bad request response for an /extensionArtifact request, the response body is read to determine whether it is an InvalidContainerConfiguration response. If it is not an InvalidContainerConfiguration response, the resp is returned to the calling functions with an empty buffer since the body has already been read. This resulted in a lot of telemetry with an empty body, which makes it harder to debug any issues:
`Fetch failed: [ProtocolError] Fetch failed from [http://168.63.129.16:32526/extensionArtifact]: [HTTP Failed] [400: Bad Request] b''`

2. There were many machines that fell back to the secondary channel without any telemetry on the primary channel failure. We should always send telem on the primary channel failure.

Issue # 
---

### PR information

---


azurelinuxagent/common/protocol/wire.py
```diff
@@ -783,8 +783,17 @@ def _fetch_response(self, uri, headers=None, use_proxy=None, retry_codes=None, o
 
             host_plugin = self.get_host_plugin()
 
+            response_error = None
+
+            # If we got a 400 (bad request) because the container id is invalid, it could indicate a stale goal
+            # state. The caller will handle this exception by forcing a goal state refresh and retrying the call.
+            if resp.status == httpclient.BAD_REQUEST:
```
maddieford, line 807: This check was previously done in restutil.http_request on lines 485-491. 

The agent has logic to refresh the goal state on HGAP /extensionArtifact requests that return a 400 status with 'InvalidContainerConfiguration' in the body. This is because the goal state may be stale, and we need to refresh to get the updated containerId before making the artifact download request. 

To handle this case, currently the restutil.http_request function reads the response body for a 400 response and raises an InvalidContainerError exception if the response body contains 'InvalidContainerConfiguration'. If the response body does not contain 'InvalidContainerConfiguration', the restutil.http_request function just returns the resp object. When this calling function (_fetch_response) gets the response object and tries to read the body for a 400 response, the buffer is already empty since restutil.http_request read it already. When this function sends telemetry on the failed response, the body appears to be empty. I saw many instances of this in the telemetry of the canary release to make hgap the primary channel:
`Fetch failed: [ProtocolError] Fetch failed from [http://168.63.129.16:32526/extensionArtifact]: [HTTP Failed] [404: Not Found] b''`

To address this, I moved the check for InvalidContainerConfiguration to this _fetch_response function. This is where we send telemetry for download failures, so it makes sense read the body here. In addition, all artifact downloads go through this fetch_response function, and we only care about handling InvalidContainerError exceptions on artifact download requests.

azurelinuxagent/common/protocol/wire.py
```diff
@@ -942,6 +951,11 @@ def _download_using_appropriate_channel(self, direct_download, hgap_download):
         try:
             return_value = secondary_channel()
 
+            # Send telemetry for primary channel failure
```
maddieford: When monitoring the release in Canary to make HGAP the primary download channel, I noticed that there are some VMs where we don't get telemetry on the primary channel download failure when we fallback to the secondary channel. I'm adding this telemetry here

azurelinuxagent/common/protocol/wire.py
```diff
@@ -783,8 +783,17 @@ def _fetch_response(self, uri, headers=None, use_proxy=None, retry_codes=None, o
...
+            # state. The caller will handle this exception by forcing a goal state refresh and retrying the call.
+            if resp.status == httpclient.BAD_REQUEST:
+                response_error = read_response_error(resp)
+                if INVALID_CONTAINER_CONFIGURATION in response_error:
+                    raise InvalidContainerError(response_error)
+
             if restutil.request_failed(resp, ok_codes=ok_codes):
-                error_response = restutil.read_response_error(resp)
+                error_response = response_error if response_error is not None else restutil.read_response_error(resp)
                 msg = "Fetch failed from [{0}]: {1}".format(uri, error_response)
```
maddieford: When we send this telemetry for the download failure, the artifact URI is not included in the message for HGAP requests since that artifact URI is a header (not in the request URI itself). As a result, the event doesn't show which artifact we're failing to fetch:
Fetch failed: [ProtocolError] Fetch failed from **[http://168.63.129.16:32526/extensionArtifact]**: [HTTP Failed] [404: Not Found] ...

We aren't able to easily see which artifacts are failing to be downloaded across multiple VMs, regions, subs, etc.

Do you think it's worth adding the headers to this telemetry event so we can see which artifact is failing to be downloaded in a single event?

tests/common/protocol/test_wire.py
```diff
@@ -643,7 +660,7 @@ def http_get_handler(url, *_, **kwargs):
             self.assertTrue(os.path.exists(target_directory), 'The extension package was not downloaded')
             self.assertTrue(HostPluginProtocol.is_default_channel, "The host channel should have been set as the default")
 
-    def test_download_zip_package_should_retry_the_host_channel_after_refreshing_host_plugin(self):
+    def test_download_zip_package_should_retry_the_host_channel_after_refreshing_host_plugin_on_resource_gone_error(self):
```
maddieford, line 663: We were only testing that the agent refreshes goal state on ResourceGone responses, but the agent should also refresh the goal state on 400 InvalidContainerConfiguration responses so I added unit tests for that scenario on each download type (vmap, manifest, package)

azurelinuxagent/common/protocol/wire.py
maddieford: For example, it was difficult to determine if a download failure was a failure on the manifest or the zip in the previous canary release

azurelinuxagent/common/protocol/wire.py
narrieta: Today restutil.http_request has error handling logic that applies to all requests, no matter what kind of request it is. It is hard to know if that logic applies to WS, HGAP, storage, or combinations of those.

I wasn't sure if InvalidContainer applied also to WS requests. It seems like we are safe with this change. Looking at the code history, this was added to handle errors from the HGAP (though initially it was reported as ResourceGone - which is what WS returns in similar conditions)

```
            # Map invalid container configuration errors to resource gone in
            # order to force a goal state refresh, which in turn updates the
            # container-id header passed to HostGAPlugin.
            # See #1294.
            if _is_invalid_container_configuration(resp):
                raise ResourceGoneError()
```

The original comment was lost during a piece of refactoring; let's bring it back

narrieta: Initially _fetch_response was meant to be used for calls to storage (see call to call_storage_service at the top). Somehow, it ended up being used for calls to HGAP as well. 

Calls to HGAP should fall a pattern similar to fetch_config, which goes thru call_wireserver. This method enforces behavior, such as never using a proxy, which is critical on calls to WS.

Calls to HGAP should also never use a proxy. It is better to follow a pattern similar to fetch_config than rely on developers passing the correct value in the use_proxy parameter.

narrieta: The call to report_fetch_health @ line 801 is also specific to HGAP (it reports failures in HGAP's artifact downloads API).

Let's mark it with a comment as well

azurelinuxagent/common/protocol/wire.py
narrieta: Yes, but let's add only the header with the URI, so as not to risk exposing anything sensitive. Those URIs are public blobs, so ok to add that header.

azurelinuxagent/common/protocol/wire.py
narrieta: It seems like when InvalidContainerConfiguration was used instead of ResourceGoneError, not all the code was updated.

For example
```
        try:
            host = self.get_host_plugin()
            host.put_vm_status(self.status_blob, extensions_goal_state.status_upload_blob, extensions_goal_state.status_upload_blob_type)
            return
        except ResourceGoneError:
            # refresh the host plugin client and try again on the next iteration of the main loop
            self.update_host_plugin_from_goal_state()
            return
```

(unless the HGAP actually returns 410 here; most likely a code omission, but worth checking with the HGAP devs)

azurelinuxagent/common/protocol/wire.py
narrieta: <deleted previous comment>

narrieta: Should we add the message to the DefaultChannelChange event instead?

nagworld9: Do we want to know always or only when secondary_chaneel() succeed?

tests/common/protocol/test_wire.py
```diff
@@ -685,6 +702,55 @@ def http_get_handler(url, *_, **kwargs):
...
+
+                protocol.client.download_zip_package("Microsoft.FakeExtension-1.0.0.0", [extension_url],
+                                                     target_file, target_directory, use_verify_header=False,
+                                                     signature="", ignore_signature_validation_errors=True)
+
+                urls = protocol.get_tracked_urls()
+                self.assertEqual(len(urls), 4, "Unexpected number of HTTP requests: [{0}]".format(urls))
+                self.assertEqual(urls[0], extension_url,
+                                 "The first attempt should have been over the direct channel")
+                self.assertTrue(self.is_host_plugin_extension_artifact_request(urls[1]),
```
nagworld9: why we check same is_host_plugin_extension_artifact_request for both zip pkg and manifest downloads? Are they not different?

azurelinuxagent/common/protocol/wire.py
maddieford: We If secondary channel fails, we raise exception with the failure on both channels and the caller will handle the exception

azurelinuxagent/common/protocol/wire.py
maddieford: I added these notes as comments

azurelinuxagent/common/protocol/wire.py
maddieford: Added the header with the appropriate URIs to the failure messages

azurelinuxagent/common/protocol/wire.py
```diff
@@ -783,11 +796,27 @@ def _fetch_response(self, uri, headers=None, use_proxy=None, retry_codes=None, o
...
-                msg = "Fetch failed from [{0}]: {1}".format(uri, error_response)
+                error_response = response_error if response_error is not None else restutil.read_response_error(resp)
+                if len(headers_for_failure_msg) > 0:
+                    msg = "Fetch failed from [{0}] with headers [{1}]: {2}".format(uri, json.dumps(headers_for_failure_msg), error_response)
+                else:
+                    msg = "Fetch failed from [{0}]: {1}".format(uri, error_response)
                 logger.warn(msg)
 
+                # TODO: The call to report_fetch_health should be limited to HGAP requests. That method should only
+                #  be used to report failures in HGAP's artifact downloads API
```
narrieta: can we emphasize that this code needs to be moved out of this method along with the above code for BAD_REQUEST?

tests/common/protocol/test_wire.py
narrieta: Before formalizing this behavior in a unit test, we should double-check whether the HGAP ever returns 410.

From my last reading of this code, seems to me that InvalidContainerConfiguration is meant only for HGAP APIs, and ResourceGone was meant initially for WireServer APIs.

narrieta: Reply from HGAP dev. 410 is returned when

/vmsettings -> if hgap's internal container cache has no rcf file path (i.e. it hasn't been made aware of any rcf files by node service)
/extensionArtifact -> for vmap blob download, if it has no vmap url cached (which is only because it has no ccf or rcf related state)
/status -> same as /extensionArtifact but for status blob url

maddieford: In that case I'll keep the Resource Gone unit tests

azurelinuxagent/common/protocol/wire.py
```diff
@@ -987,6 +1033,7 @@ def upload_status_blob(self):
             return
         except ResourceGoneError:
             # refresh the host plugin client and try again on the next iteration of the main loop
+            # TODO: Check if we should also be handling the InvalidContainerConfiguration scenario with a goal state refresh
```
narrieta: Thanks for adding this comment. After all, given the last reply from the HGAP team, ResourceGone is OK here, no need to handle InvalidContainer.

narrieta (dismissed):
LGTM, added 1 minor comment

azurelinuxagent/common/protocol/wire.py
maddieford: Removed

narrieta (approved):

nagworld9 (approved):

PR #3553, opened 2026-02-10
Define 'backslashreplace' error handler for Python <= 3.4

narrieta:
First PR to address 'backslashreplace' on Python <= 3.4

The PR defines an error handler for Unicode encoding/decoding errors on Python <= 3.4. Callers can opt-in to use this implementation using  azurelinuxagent.common.future.BACKSLASH_REPLACE.

BACKSLASH_REPLACE is a temporary mechanism to use the error handler. Once all the code has been updated to use this error handler, I will remove BACKSLASH_REPLACE and use the error handler by default on Python <= 3.4.

### DIFF
--- a/azurelinuxagent/common/future.py
+++ b/azurelinuxagent/common/future.py
@@ -69,6 +69,35 @@ elif sys.version_info[0] == 2:
 else:
     raise ImportError("Unknown python version: {0}".format(sys.version_info))
 
+#
+# Python <= 3.4 doesn't have a 'backslashreplace' error handler for encoding/decoding strings; for those versions, we register
+# a custom implementation as 'waagent_backslashreplace'.
+#
+# BACKSLASH_REPLACE points to the name of this custom implementation; on Python versions where this is not needed, BACKSLASH_REPLACE
+# points to the standard implementation in Python.
+#
+if not (sys.version_info[0] == 2 or sys.version_info[0] == 3 and sys.version_info[1] <= 4):
+    BACKSLASH_REPLACE = 'backslashreplace'
+else:
+    import codecs
+
+    def __waagent_backslashreplace__(unicode_error):
+        obj, start, end = unicode_error.object, unicode_error.start, unicode_error.end
+        replacement = ustr('')
+        if isinstance(unicode_error, UnicodeDecodeError):
+            for c in obj[start:end]:
+                if not isinstance(c, int):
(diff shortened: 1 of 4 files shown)


azurelinuxagent/common/future.py
```diff
@@ -69,6 +69,35 @@
...
+                replacement += ustr('\\x{0:02x}'.format(c))
+        else:  # Encoding error
+            for c in obj[start:end]:
+                c = ord(c)
+                replacement += ustr('\\U{0:08x}'.format(c)) if c >= 0x10000 else ustr('\\u{0:04x}'.format(c))
+        return replacement, end
+
+    BACKSLASH_REPLACE = 'waagent_backslashreplace'
+
+    codecs.register_error(BACKSLASH_REPLACE, __waagent_backslashreplace__)
```
nagworld9, line 100: I read codecs.backslashreplace_errors built-in has this implementation, can we use instead custom handler?

That built-in handler available in <= python 3.4 it seems

narrieta: no luck

```
Python 2.6.9 (unknown, Nov  9 2016, 21:08:27) 
[GCC 4.8.4] on linux3
>>> import codecs
>>> codecs.register_error('backslashreplace', codecs.backslashreplace_errors)
>>> print(unicode(b"Hello\xffWorld", encoding='utf-8', errors="backslashreplace"))
  File "<stdin>", line 1, in <module>
TypeError: don't know how to handle UnicodeDecodeError in error callback

```

narrieta:
```
Key Points for Python 2.6
No codecs.backslashreplace_errors function — you just pass 'backslashreplace' as the errors argument.
Works only for encoding in Python 2.6.
If you need similar behavior for decoding, you’d have to write a custom error handler and register it with codecs.register_error().
```

nagworld9: Ok, i see. Implementation had for encoding only. We could return bulitin for encode and decode for custom logic but that may not be reliable. If other versions removed encode and had for decode, then we are in trouble. I think I'm good with current custom handler
```
odecs.backslashreplace_errors(exception)[¶](https://docs.python.org/2.6/library/codecs.html#codecs.backslashreplace_errors)
Implements the backslashreplace error handling (for encoding only): the unencodable character is replaced by a backslashed escape sequence
```

maddieford (approved):

nagworld9 (approved):

PR #3558, opened 2026-02-17
Move FIPS test to West US 2

narrieta:
The test was previously on the Canary regions waiting for a deployment of FIPS support on CRP/Fabric. The deployment has been completed for several months, so moving the test to our default region (West US 2).

### DIFF
--- a/tests_e2e/test_suites/fips.yml
+++ b/tests_e2e/test_suites/fips.yml
@@ -17,11 +17,14 @@ images:
   - "rhel_95"
 
 #
-# Support for FIPS 140-3 is currently deployed only to a few regions on AzureCloud, and deployment has not started on AzureChinaCloud.
+# The test uses a keyvault located on westus2.
 #
 locations:
-  - "AzureCloud:eastus2euap"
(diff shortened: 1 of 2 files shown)


tests_e2e/test_suites/fips.yml
```diff
@@ -17,11 +17,14 @@ images:
   - "rhel_95"
 
 #
-# Support for FIPS 140-3 is currently deployed only to a few regions on AzureCloud, and deployment has not started on AzureChinaCloud.
+# The test uses a keyvault located on westus2.
 #
 locations:
-  - "AzureCloud:eastus2euap"
+  - "AzureCloud:westus2"
```
nagworld9, line 23: we use default westus2, so no need to specify. Many tests don't specify as well

narrieta: Yes, but if we change the default I still want to stay on West US 2. The test VM needs to be in the same region as the Key Vault.

nagworld9 (approved):

maddieford (approved):

PR #3559, opened 2026-02-18
Add Oracle Linux 9.5 to FIPS test

narrieta:
... and remove workaround for Certificates.pem

### DIFF
--- a/tests_e2e/tests/fips/fips.py
+++ b/tests_e2e/tests/fips/fips.py
@@ -227,22 +227,6 @@ class Fips(AgentVmTest):
                 }
             })
         else:
-            # if Fips._is_red_hat_distro(self._distro):
-            #     #
-            #     # TODO: Remove this workaround once the pre-installed Agent RHEL/Oracle 9.5 is updated to support FIPS 140-3.
-            #     #
-            #     # The current Daemon on RHEL/Oracle 95 (2.7.0.6) has not been updated to support FIPS 140-3 and goes into an infinite loop while trying to fetch the certificates in the goal
-            #     # state. The reason is that, even if it cannot decrypt the response from the WireServer, 2.7.0.6 assumes that Certificates.pem always exists; if it does not, it goes
-            #     # into an infinite retry loop. To prevent this, before deallocating and reallocating, ensure that there is a Certificates.pem file, even if it is empty.
-            #     #
-            #     # The agent may remove the new file if it fetches the goal state certificate (after the VM restart in the previous step) and fails to decrypt it while we create the file below.
-            #     # Therefore, stop the agent service to prevent it from removing the file.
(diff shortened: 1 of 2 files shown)


tests_e2e/tests/fips/fips.py
```diff
@@ -213,22 +227,6 @@ def _force_new_pfx(self):
                 }
             })
         else:
-            if self._distro == 'rhel_95':
```
narrieta, line 216: The workaround is no longer needed, since current agent ensures that Certificates.pem is created.

maddieford: Do you mean the latest ext handler ensures that Certificates.pem is created? 

I need a refresher, does that happen in agent init?

narrieta: Yes, current release has the change. 

We do it while fetching the certificates: https://github.com/Azure/WALinuxAgent/blob/release-2.15.1.0/azurelinuxagent/common/protocol/goal_state.py#L552

tests_e2e/tests/fips/fips.py
```diff
@@ -122,6 +122,15 @@ def _opt_in_to_fips(self) -> None:
...
         log.info("Opt-in completed...")
 
+    @staticmethod
+    def _is_red_hat_distro(distro: str) -> bool:
+        """
+            Returns true if the distro is a Red Hat-based distro (e.g. RHEL or Oracle Linux).
+
+            The 'distro' parameter is expected to be in the format {name}_{version}, as returned by get_distro.py.
+        """
+        return distro in ("rhel_95", "oracle_95")
```
nagworld9, line 132: Missing a change for adding oracle_95 in fips.yml?

narrieta: that is why the tests are passing 100%, huh?

I was testing doing manual runs of the test pipeline. Let me add it.

tests_e2e/tests/fips/fips.py
maddieford: Is there a concern for possible race condition where agent service gets stopped before Certificates.pem is created during test vm setup? Or we get to this point in the test before Certificates.pem is created?

tests_e2e/tests/fips/fips.py
narrieta: added

tests_e2e/tests/fips/fips.py
narrieta: No concern.

This workaround was here because the old agent would delete Certificates.pem on decryption failure, which would happen after enabling FIPS. The new agent does not remove it. Instead, it ensures an empty file is created on decryption failure, or if the file is not created for any other reason.

We needed the workaround here because we are about the deallocate/reallocate the machine.

nagworld9 (approved):

maddieford (approved):

PR #3561, opened 2026-02-24
Improve retry strategy for artifact download

maddieford:
## Description
Current artifact download retries are expensive for VMs where outbound connections are disabled. Since direct downloads are the primary download channel, each request on these VMs takes 10 seconds to timeout. The agent currently retries up to 6 times, with a varying delay between each retry (this takes up to 2.5 minutes), before falling back to the host channel for downloads. This is impacting TDPR for those VMs.

We did a release to Canary where HGAP was the primary download channel in an attempt to address the above issue. We saw an increase in 400 responses on HGAP requests due to a cache issue in HGAP that may produce a BAD_REQUEST failure for valid URIs when using the extensionArtifact API. Retries should help address that cache issue, but the HGAP dev also cited performance concerns w.r.t making HGAP the primary download.

As a result of the above concerns making HGAP the primary download channel, this PR keeps the direct download channel as the primary, but makes some changes to account for those VMs where outbound connections are disabled. For direct artifact downloads, there will be no retries on timed out requests (and the timeout will be decreases to 5 seconds for the first request). As a result, the agent will fallback to the host channel much faster, reducing the delay on those VMs from 2.5 mins to 5 seconds.

In addition, the PR makes a change to make the BAD_REQUEST response retryable on zip downloads via the host channel to address the cache issue described above. BAD_REQUEST is already treated as retryable for manifest and VMAP downloads on Host channel, and HGAP devs confirmed it should be treated as retryable for zip downloads too.

Issue # 
---

### PR information

---

### DIFF
--- a/azurelinuxagent/common/protocol/wire.py
+++ b/azurelinuxagent/common/protocol/wire.py
@@ -607,7 +607,9 @@ class WireClient(object):
         """
         host_ga_plugin = self.get_host_plugin()
 
-        direct_download = lambda uri: self.fetch(uri)[0]
+        # Fail fast on request timeouts when doing direct downloads, as these may indicate no outbound connection on
+        # the VM and should fall back quickly to the host channel.
+        direct_download = lambda uri: self.fetch(uri, fail_fast_on_timeout=True)[0]
 
         def hgap_download(uri):
             request_uri, request_headers = host_ga_plugin.get_artifact_request(uri, use_verify_header=use_verify_header)
@@ -635,11 +637,13 @@ class WireClient(object):
         """
         host_ga_plugin = self.get_host_plugin()
 
-        direct_download = lambda uri: self.stream(uri, target_file, headers=None, use_proxy=True)
+        # Fail fast on request timeouts when doing direct downloads, as these may indicate no outbound connection on
+        # the VM and should fall back quickly to the host channel.
+        direct_download = lambda uri: self.stream(uri, target_file, headers=None, use_proxy=True, fail_fast_on_timeout=True)
 
         def hgap_download(uri):
             request_uri, request_headers = host_ga_plugin.get_artifact_request(uri, use_verify_header=use_verify_header, artifact_manifest_url=host_ga_plugin.manifest_uri)
-            return self.stream(request_uri, target_file, headers=request_headers, use_proxy=False)
+            return self.stream(request_uri, target_file, headers=request_headers, use_proxy=False, retry_codes=restutil.HGAP_GET_EXTENSION_ARTIFACT_RETRY_CODES)
 
         def on_downloaded():
             # If the 'signature' parameter is not an empty string, validate the zip package signature immediately after download.
@@ -733,14 +737,14 @@ class WireClient(object):
             except Exception as exception:
                 logger.warn("Cannot delete {0}: {1}", target_file, ustr(exception))
 
-    def stream(self, uri, destination, headers=None, use_proxy=None):
+    def stream(self, uri, destination, headers=None, use_proxy=None, retry_codes=None, fail_fast_on_timeout=False):
         """
         Downloads the content of the given 'uri' and saves it to the 'destination' file.
         """
         try:
             logger.verbose("Fetch [{0}] with headers [{1}] to file [{2}]", uri, headers, destination)
 
-            response = self._fetch_response(uri, headers, use_proxy)
+            response = self._fetch_response(uri, headers, use_proxy, retry_codes, fail_fast_on_timeout=fail_fast_on_timeout)
             if response is not None and not restutil.request_failed(response):
                 chunk_size = 1024 * 1024  # 1MB buffer
                 with open(destination, 'wb', chunk_size) as destination_fh:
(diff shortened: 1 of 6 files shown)


azurelinuxagent/common/protocol/wire.py
```diff
@@ -635,11 +638,14 @@ def download_zip_package(self, package_name, uris, target_file, target_directory
...
-        direct_download = lambda uri: self.stream(uri, target_file, headers=None, use_proxy=True)
+        # Fail fast on request timeouts when doing direct downloads, as these may indicate no outbound connection on
+        # the VM and should fall back quickly to the host channel. Reconsider this strategy if we switch the primary
+        # download channel to HGAP.
+        direct_download = lambda uri: self.stream(uri, target_file, headers=None, use_proxy=True, fail_fast_on_timeout=True)
 
         def hgap_download(uri):
             request_uri, request_headers = host_ga_plugin.get_artifact_request(uri, use_verify_header=use_verify_header, artifact_manifest_url=host_ga_plugin.manifest_uri)
-            return self.stream(request_uri, target_file, headers=request_headers, use_proxy=False)
+            return self.stream(request_uri, target_file, headers=request_headers, use_proxy=False, retry_codes=restutil.HGAP_GET_EXTENSION_ARTIFACT_RETRY_CODES)
```
maddieford, line 648: We weren't treating BAD_REQUEST response as retryable on zip downloads via HGAP. Rohit confirmed it should be retryable for zip downloads (same as manifest and VMAP downloads)

azurelinuxagent/common/utils/restutil.py
```diff
@@ -372,7 +372,8 @@ def http_request(method,
                  retry_delay=DELAY_IN_SECONDS,
                  throttle_delay=THROTTLE_DELAY_IN_SECONDS,
                  redact_data=False,
-                 return_raw_response=False):
+                 return_raw_response=False,
+                 fail_fast_on_timeout=False):
```
maddieford, line 378: I opted to add an additional option to this method since all of the retry logic is here and I only wanted to change the retry logic for IOErrors on artifact download requests via direct channel

azurelinuxagent/common/utils/restutil.py
```diff
@@ -520,6 +524,11 @@ def http_request(method,
                 msg = '[HTTP Failed] {0} {1} {2} -- IOError {3}'.format(method, clean_url, json.dumps(headers_for_failure_msg), e)
             else:
                 msg = '[HTTP Failed] {0} {1} -- IOError {2}'.format(method, clean_url, e)
+            # If the error was an IOError and fail_fast_on_timeout is True, check if the error message contains
+            # "timed out" or "timeout" and break the loop to fail fast in the case of no outbound connection on the VM.
+            # Otherwise, continue with the retries.
+            if fail_fast_on_timeout and "timed out" in ustr(e) or "timeout" in ustr(e):
```
maddieford, line 532: errno is None for the time out exception so I am using the string here

tests_e2e/tests/scripts/no_outbound_connections-verify_download_channel_fallback.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def main():
+    # 2026-02-20T22:39:22.691020Z INFO ExtHandler ExtHandler Downloading artifacts profile blob
+    # 2026-02-20T22:39:27.714015Z WARNING ExtHandler ExtHandler Fetch failed: [HttpError] [HTTP Failed] GET https://md-hdd-nqxfsnnvnjr3.z27.blob.storage.azure.net/$system/lisa-maddieford-20260220-223218-252-e0-n0.49e33a82-9f6e-4482-86a3-0db71c338ebc.vmSettings -- IOError timed out -- 1 attempts made
+    # 2026-02-20T22:39:27.740385Z INFO ExtHandler ExtHandler Default channel changed to HostGAPlugin channel.
+    # 2026-02-20T22:39:30.733714Z INFO ExtHandler ExtHandler ProcessExtensionsGoalState started [incarnation_1 channel: WireServer source: Fabric activity: 1082ed79-bbc7-4925-a932-aefe84bca6e9 correlation 2ad608f9-327b-4b2a-8673-6b40ca7ee6e7 created: 2026-02-20T22:33:22.800011Z]
+    # Patterns to match
+    downloading_vmap = re.compile(r"Downloading artifacts profile blob")                        # This is the artifact download we expect to see the channel change happen on
+    fetch_failed_on_direct_pattern = re.compile(r"Fetch failed:.*1 attempts made")              # There should only be one attempt on the Direct channel before falling back to HostGAPlugin
```
narrieta:
```suggestion
    fetch_failed_on_direct_pattern = re.compile(r"Fetch failed:.* 1 attempts made")              # There should only be one attempt on the Direct channel before falling back to HostGAPlugin
```

narrieta: (.*1 would match 01, 11, etc)

azurelinuxagent/common/utils/restutil.py
```diff
@@ -458,10 +459,13 @@ def http_request(method,
         attempt += 1
 
         try:
+            # If fail_fast_on_timeout is True, use a shorter timeout for the first attempt to fail fast in the case of
+            # no outbound connection on the VM.
+            req_timeout = timeout if not fail_fast_on_timeout or attempt > 1 else min(timeout, 5)
```
nagworld9: can we add constant for 5 secs default?

narrieta (approved):

nagworld9 (approved):

Issue #3557, opened 2026-02-17
[BUG] property CPUAccounting deprecated in systemd 258+


mirespace:
**Describe the bug: A clear and concise description of what the bug is.**

When preparing the new version of WALInuxagent for Ubuntu Resolute (26.04), the following can be seen in the log when the agent is running:

Feb 17 10:55:50 rwala python3[7874]: 2026-02-17T10:55:50.235937Z INFO ExtHandler ExtHandler [CGW] Error initializing cgroups: Can't find property CPUAccounting of walinuxagent.service
Feb 17 10:55:50 rwala python3[7874]: 2026-02-17T10:55:50.300308Z INFO ExtHandler ExtHandler [HEARTBEAT] Agent WALinuxAgent-2.15.0.1 is running as the goal state agent [DEBUG HeartbeatCounter: 0;HeartbeatId: DE900D47-FC69-4384-A340-DFE079A920EE;UpdateGSErrors: 0;AutoUpdate: 1;UpdateMode: SelfUpdate;]

From  https://manpages.debian.org/unstable/systemd/systemd-system.conf.5.en.html , we can read:

systemd 258
DefaultCPUAccounting= setting is deprecated, because CPU accounting is always available on the unified cgroup hierarchy and such setting has no effect.

**Distro and WALinuxAgent details (please complete the following information):**
 - Distro and Version:  Ubuntu 26.04
 - WALinuxAgent version:
          ubuntu@rwala:~ waagent --version
          WALinuxAgent-2.15.0.1 running on ubuntu 26.04
         Python: 3.13.12
        Goal state agent: 2.15.0.1
- Cgroups: 
     ubuntu@rwala:~$ stat -fc %T /sys/fs/cgroup
     cgroup2fs
- systemd version:
      ubuntu@rwala:~$ systemctl --version
      systemd 259 (259-1ubuntu3)
[...]

PR #3564, opened 2026-02-27
Remove CPU and memory accounting enablement in CgroupV2

nagworld9:
## Description

Fixing https://github.com/Azure/WALinuxAgent/issues/3557

The issue is 
```
systemd 258
DefaultCPUAccounting= setting is deprecated, because CPU accounting is always available on the unified cgroup hierarchy and such setting has no effect.
```
In the agent, we explicitly set CPUAccounting and MemoryAccounting to true using systemd to enable the related resource metrics. This is required for cgroup v1–based distros. When we onboarded cgroup v2 distros, the same logic was applied there as well, even though it is not needed(since cgroup v2 enables CPU and memory accounting by default).

Now, newer distros like Ubuntu 26 are adopting the latest systemd version, where CPU accounting deprecated as mentioned above, as a result agent failing with

`Feb 17 10:55:50 rwala python3[7874]: 2026-02-17T10:55:50.235937Z INFO ExtHandler ExtHandler [CGW] Error initializing cgroups: Can't find property CPUAccounting of walinuxagent.service`

This PR removes the explicit accounting enablement for cgroup v2 distros.

Issue # 
---

### PR information

---

### DIFF
--- a/azurelinuxagent/ga/cgroupconfigurator.py
+++ b/azurelinuxagent/ga/cgroupconfigurator.py
@@ -308,7 +308,8 @@ class CGroupConfigurator(object):
                 files_to_create.append((azure_slice, _AZURE_SLICE_CONTENTS))
 
             # Slice has only accounting properties, so no need explicit set in cgroupv2
-            if not os.path.exists(vmextensions_slice) and not self.using_cgroup_v2():
+            accounting_props, _ = self._cgroups_api.get_accounting_properties()
+            if not os.path.exists(vmextensions_slice) and len(accounting_props) > 0:
                 files_to_create.append((vmextensions_slice, _VMEXTENSIONS_SLICE_CONTENTS))
 
             if fileutil.findre_in_file(agent_unit_file, r"Slice=") is not None:
--- a/azurelinuxagent/ga/cgroupconfigurator.py
+++ b/azurelinuxagent/ga/cgroupconfigurator.py
@@ -56,16 +56,6 @@ Before=slices.target
 CPUAccounting=yes
 MemoryAccounting=yes
 """
-_EXTENSION_SLICE_CONTENTS = """
-[Unit]
-Description=Slice for Azure VM extension {extension_name}
-DefaultDependencies=no
-Before=slices.target
-[Slice]
-CPUAccounting=yes
-CPUQuota={cpu_quota}
-MemoryAccounting=yes
-"""
 LOGCOLLECTOR_SLICE = "azure-walinuxagent-logcollector.slice"
 # More info on resource limits properties in systemd here:
 # https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/resource_management_guide/sec-modifying_control_groups
@@ -277,15 +267,14 @@ class CGroupConfigurator(object):
             # New agent will setup limits for scope instead slice, so removing existing logcollector slice.
             CGroupConfigurator._Impl._cleanup_unit_file(logcollector_slice)
 
-            # Cleanup the old drop-in files, new agent will use systemdctl set-property to enable accounting and limits
+            # Cleanup the old drop-in files, new agent will use systemctl set-property to enable accounting and limits
             CGroupConfigurator._Impl._cleanup_unit_file(agent_drop_in_file_cpu_accounting)
 
             CGroupConfigurator._Impl._cleanup_unit_file(agent_drop_in_file_memory_accounting)
 
             CGroupConfigurator._Impl._cleanup_unit_file(agent_drop_in_file_cpu_quota)
 
-        @staticmethod
-        def _setup_azure_slice():
+        def _setup_azure_slice(self):
             """
             The agent creates "azure.slice" for use by extensions and the agent. The agent runs under "azure.slice" directly and each
             extension runs under its own slice ("Microsoft.CPlat.Extension.slice" in the example below). All the slices for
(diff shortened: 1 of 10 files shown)


azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -56,16 +56,6 @@
 CPUAccounting=yes
 MemoryAccounting=yes
 """
-_EXTENSION_SLICE_CONTENTS = """
```
nagworld9, line 59: leftover code after switch to systemctl --runtime

nagworld9:
FYI @mirespace

azurelinuxagent/ga/cgroupconfigurator.py
```diff
@@ -318,7 +307,8 @@ def _setup_azure_slice():
             if not os.path.exists(azure_slice):
                 files_to_create.append((azure_slice, _AZURE_SLICE_CONTENTS))
 
-            if not os.path.exists(vmextensions_slice):
+            # Slice has only accounting properties, so no need explicit set in cgroupv2
+            if not os.path.exists(vmextensions_slice) and not self.using_cgroup_v2():
```
maddieford: Now that we have the get_accounting_properties() method, should we just check if the result of that is empty instead of explicitly checking if v2 is in use? Just in case we ever need to add other accounting props in v2

azurelinuxagent/ga/cpucontroller.py
```diff
@@ -125,7 +125,7 @@ def get_tracked_metrics(self):
         return tracked
 
     def get_unit_properties(self):
-        return ["CPUAccounting", "CPUQuotaPerSecUSec"]
+        return ["CPUQuotaPerSecUSec"]
```
maddieford, line 128: Does `systemctl show walinuxagent.service --property CPUAccounting` fail for v2 machine if we don't set it explicitly?

tests/ga/test_cgroupconfigurator.py
```diff
@@ -233,6 +233,11 @@ def test_agent_enforcement_enabled_in_v2(self):
             self.assertIn(cmd1, configurator.mocks.commands_call_list, "The command to set CPU quota was not called")
             self.assertIn(cmd2, configurator.mocks.commands_call_list, "The command to set Memory quota was not called")
 
+    def test_accounting_properties_not_set_explicitly_in_cgroupv2(self):
+        with self._get_cgroup_configurator_v2() as configurator:
+            cmd = 'systemctl set-property walinuxagent.service CPUAccounting=yes MemoryAccounting=yes --runtime'
```
maddieford, line 238: I think we should check that neither `CPUAccounting=yes` or `MemoryAccounting=yes` in any cmd, instead of depending on both being called in same command and in same order

tests/ga/test_collect_logs.py
```diff
@@ -63,29 +63,33 @@ def _create_collect_logs_handler(iterations=1, cgroup_version=CgroupVersions.V1,
...
                             with patch.object(cgroups_configurator_singleton, "enabled", return_value=cgroups_enabled):
-                                def run_and_wait():
-                                    collect_logs_handler.run()
-                                    collect_logs_handler.join()
-
-                                collect_logs_handler = get_collect_logs_handler()
-                                collect_logs_handler.get_mock_wire_protocol = lambda: protocol
-                                collect_logs_handler.run_and_wait = run_and_wait
-                                yield collect_logs_handler
+                                with patch.object(cgroups_configurator_singleton, "_cgroups_api", mock_cgroups_api, create=True):
```
maddieford, line 71: Why add this mock here?

tests/ga/test_collect_logs.py
```diff
@@ -63,29 +63,33 @@ def _create_collect_logs_handler(iterations=1, cgroup_version=CgroupVersions.V1,
 
                         # Grab the singleton to patch it
                         cgroups_configurator_singleton = CGroupConfigurator.get_instance()
+                        mock_cgroups_api = MagicMock()
+                        mock_cgroups_api.get_accounting_properties.return_value = ((), ())
```
maddieford, line 67: Should we have different value for v1?

tests/lib/mock_cgroup_environment.py
```diff
@@ -101,6 +102,11 @@
 
     MockCommand(r"^systemctl show (.+) --property MemoryHigh$",
 '''MemoryHigh=infinity
+'''),
+    MockCommand(r"^systemd-run --unit=([^\s]+) --scope ([^\s]+)",
```
maddieford, line 106: Looks like same command in _MOCKED_COMMANDS_V1 and _MOCKED_COMMANDS_V2? Unless Im missing something

azurelinuxagent/ga/cpucontroller.py
nagworld9: Getting value nothing to do with set. We can query properties which are supported by systemd.

But it will fail for distros who has systemd 258+, since they removed CPUAccounting property.

tests/ga/test_collect_logs.py
nagworld9: Because now get_logcollector_unit_properties call the accounting properties from cgroups_api object.

tests/ga/test_collect_logs.py
nagworld9: it deoesn't matter, because we are not testing that. we are just mocking get_accounting_properties call. But to match the original behavior, I could add properties names and values

tests/lib/mock_cgroup_environment.py
nagworld9: There is difference, in v1 we expect something (which is accounting properties) after systemd-run, but in v2 we don't. There is extra regex (.+) in v1

narrieta (dismissed):

tests/ga/test_cgroupconfigurator.py
nagworld9: updated

azurelinuxagent/ga/cgroupconfigurator.py
nagworld9: updated

maddieford (approved):

narrieta (approved):

PR #3572, opened 2026-03-03
Disable signature validation if timeout period exceeded

mgunnala:
## Description

Issue # 
---

To avoid impacting customers during telemetry release, the signature validation feature should be disabled if a certain timeout period is exceeded (10 seconds, by default).

### PR information

---

### DIFF
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -1472,7 +1472,8 @@ class ExtHandlerInstance(object):
                         cleanup_package_with_invalid_signature(package_file)
                         raise
                     if isinstance(ex, SignatureValidationTimeoutError):
-                        SignatureValidationTimeout.mark_exceeded()
+                        # In the case of signature validation timeout, only disable future validations for customers who haven't opted into enforcement.
+                        SignatureValidationTimeout.disable_validation()
                     report_validation_event(op=ex.operation, level=logger.LogLevel.WARNING, message=ustr(ex),
                                             name=self.ext_handler.name, version=self.ext_handler.version, duration=ex.duration)
 
@@ -1515,7 +1516,8 @@ class ExtHandlerInstance(object):
                 if not ignore_signature_validation_errors:
                     raise   # Package has already been cleaned up
                 if isinstance(ex, SignatureValidationTimeoutError):
-                    SignatureValidationTimeout.mark_exceeded()
+                    # In the case of signature validation timeout, only disable future validations for customers who haven't opted into enforcement.
+                    SignatureValidationTimeout.disable_validation()
                 report_validation_event(op=ex.operation, level=logger.LogLevel.WARNING, message=ustr(ex), name=self.ext_handler.name,
                                         version=self.ext_handler.version, duration=ex.duration)
 
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -1446,6 +1446,9 @@ class ExtHandlerInstance(object):
             # TODO: This is temporary behavior for the telemetry release. For production release, remove this
             # if-block so timeout is treated like any other signature validation failure (extension should fail).
             SignatureValidationTimeout.disable_validation()
+            report_validation_event(op=WALAEventOperation.SignatureValidation, level=logger.LogLevel.WARNING,
+                                    message="Signature validation timeout exceeded. Disabling signature validation until agent restart.",
+                                    name=self.ext_handler.name, version=self.ext_handler.version, duration=0)
         report_validation_event(op=ex.operation, level=logger.LogLevel.WARNING, message=ustr(ex),
                                 name=self.ext_handler.name, version=self.ext_handler.version, duration=ex.duration)
 
@@ -1491,7 +1494,7 @@ class ExtHandlerInstance(object):
             if self._unzip_extension_package(package_file, self.get_base_dir()):
                 package_exists = True
             else:
-                msg = "The existing extension package is invalid, will ignore it."
+                msg = "Could not expand existing extension package '{0}', will ignore it.".format(package_file)
                 self.logger.info(msg)
                 add_event(op=WALAEventOperation.Download, message=msg, name=self.ext_handler.name, version=self.ext_handler.version, is_success=True, log_event=False)
                 signature_validation_succeeded = False
--- a/azurelinuxagent/ga/exthandlers.py
+++ b/azurelinuxagent/ga/exthandlers.py
@@ -60,7 +60,7 @@ from azurelinuxagent.common.utils.flexible_version import FlexibleVersion
 from azurelinuxagent.common.version import AGENT_NAME, CURRENT_VERSION
 from azurelinuxagent.ga.signature_validation_util import validate_handler_manifest_signing_info, SignatureValidationError, \
     PackageValidationError, ManifestValidationError, signature_validation_enabled, validate_signature, \
-    cleanup_package_with_invalid_signature, report_validation_event
+    cleanup_package_with_invalid_signature, report_validation_event, SignatureValidationTimeoutError, SignatureValidationTimeout
 
 _HANDLER_NAME_PATTERN = r'^([^-]+)'
 _HANDLER_VERSION_PATTERN = r'(\d+(?:\.\d+)*)'
@@ -1467,10 +1467,12 @@ class ExtHandlerInstance(object):
                     validate_signature(package_file, self.ext_handler.encoded_signature, package_full_name=self.get_full_name())
                     signature_validation_succeeded = True
                 except SignatureValidationError as ex:
-                    # validate_signature() only raises SignatureValidationError.
+                    # validate_signature() only raises SignatureValidationError (and subclasses).
                     if not ignore_signature_validation_errors:
                         cleanup_package_with_invalid_signature(package_file)
                         raise
+                    if isinstance(ex, SignatureValidationTimeoutError):
+                        SignatureValidationTimeout.mark_exceeded()
                     report_validation_event(op=ex.operation, level=logger.LogLevel.WARNING, message=ustr(ex),
                                             name=self.ext_handler.name, version=self.ext_handler.version, duration=ex.duration)
 
@@ -1512,6 +1514,8 @@ class ExtHandlerInstance(object):
                 # validation errors should not be ignored, otherwise report the error and continue.
                 if not ignore_signature_validation_errors:
                     raise   # Package has already been cleaned up
+                if isinstance(ex, SignatureValidationTimeoutError):
+                    SignatureValidationTimeout.mark_exceeded()
                 report_validation_event(op=ex.operation, level=logger.LogLevel.WARNING, message=ustr(ex), name=self.ext_handler.name,
                                         version=self.ext_handler.version, duration=ex.duration)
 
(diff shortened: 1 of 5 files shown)


azurelinuxagent/ga/exthandlers.py
```diff
@@ -1467,10 +1467,12 @@ def download(self, ignore_signature_validation_errors):
                     validate_signature(package_file, self.ext_handler.encoded_signature, package_full_name=self.get_full_name())
                     signature_validation_succeeded = True
                 except SignatureValidationError as ex:
-                    # validate_signature() only raises SignatureValidationError.
+                    # validate_signature() only raises SignatureValidationError (and subclasses).
                     if not ignore_signature_validation_errors:
                         cleanup_package_with_invalid_signature(package_file)
                         raise
+                    if isinstance(ex, SignatureValidationTimeoutError):
```
narrieta: this probably should be done before the check at 1471

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1512,6 +1514,8 @@ def download(self, ignore_signature_validation_errors):
                 # validation errors should not be ignored, otherwise report the error and continue.
                 if not ignore_signature_validation_errors:
                     raise   # Package has already been cleaned up
+                if isinstance(ex, SignatureValidationTimeoutError):
```
narrieta: We are duplicating too much logic for the existing/downloaded extension package. I don't recall: what was the motivation for doing the signature validation as part of self.protocol.client.download_zip_package and not here?

mgunnala: The validation was placed in download_zip_package() so we can validate after download but before extraction for new packages. The "existing package" path doesn't go through download, so it needed its own validation.

I agree there's too much duplication now. For this PR, I could extract a shared helper function for the error handling. A larger refactor to centralize validation would need more thought - happy to discuss in a follow-up.

azurelinuxagent/ga/exthandlers.py
mgunnala: I think this should be after the check at 1471 - when enforcement is enabled (user is opted-in to signature validation), a timeout should fail the extension but not disable future validations. See the UT - test_should_not_disable_future_validation_if_timeout_exceeded_when_enforced()

maddieford: When `ignore_signature_validation_errors=False`, shouldn't a signature validation timeout be considered an extension failure since we were not able to successfully validate the signature of the package?

How is that extension status being reported?

tests/ga/test_extension.py
```diff
@@ -4419,9 +4422,42 @@ def test_should_not_validate_signature_on_non_cvm(self):
...
+
+            # Second run: Process another extension and verify signature validation is skipped
+            with mock_wire_protocol(data_file) as protocol:
+                protocol.aggregate_status = None
+                protocol.report_vm_status = MagicMock()
+                exthandlers_handler = get_exthandlers_handler(protocol)
+                protocol.set_http_handlers(http_get_handler=self._make_http_get_handler(data_file))
+
+                with patch('azurelinuxagent.ga.exthandlers.validate_signature') as mock_validate:
+                    exthandlers_handler.run()
```
maddieford, line 4466: On this invocation of exthandlers_handler.run() VmAccessForLinux is already installed/enable so we never enter the installation flow where validate_signature would be called. As a result, this doesn't actually test that future validations are disabled. 

Adding an unintalled ext to the GS before this invocation of run() should fix that

mgunnala: Thanks for catching this Maddie! I've updated it.

maddieford (dismissed):

azurelinuxagent/ga/exthandlers.py
narrieta: @mgunnala I think you are only focusing on the immediate telemetry release, but after that this should not be the behavior. If this is just a temporary behavior, please add a TODO comment and explain what needs to happen.

azurelinuxagent/ga/exthandlers.py
narrieta: ideally, downloading and expanding the ZIP should be separate operations 

the refactoring may be as simple as exposing the on_downloaded callback one level up

once download and expand are separate, the logic for existing and downloaded ZIPs is practically identical

mgunnala: @narrieta would it be okay if I marked it as TODO for now and implemented that refactoring in a follow-up PR?

azurelinuxagent/ga/exthandlers.py
mgunnala: I've added a TODO - during production release, "ignore_signature_validation_errors" will be False by default, and any validation error will immediately be raised. At that point, this block of code could be removed.

azurelinuxagent/ga/exthandlers.py
narrieta: ok, but for now please do the refactoring you suggested above

mgunnala: @narrieta I added a helper function as discussed above, and added the refactoring as a TODO

maddieford (dismissed):

azurelinuxagent/ga/signature_validation_util.py
```diff
@@ -341,12 +374,15 @@ def _should_delay_signature_validation():
...
-    TODO: Remove the is_confidential_vm() check once signature validation is supported on all VMs.
+    Returns True if all conditions for signature validation are met:
+    - Conf flag 'EnableSignatureValidation' is True
+    - Validation timeout has not been exceeded (TODO: remove after telemetry release)
+    - Initial delay period after agent start has passed (TODO: remove after telemetry release)
+    - OpenSSL version supports required validation parameters (TODO: remove after timestamp validation implemented)
+    - Agent is running on a Confidential VM (TODO: remove when all VMs are supported)
     """
     return conf.get_signature_validation_enabled() and \
+        not SignatureValidationTimeout.is_validation_disabled() and \
```
narrieta, line 385: The exception handled at line 396 should not be a validation error, should it?

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1432,6 +1432,23 @@ def _unzip_extension_package(self, source_file, target_directory):
...
+        If ignore_errors is True, handles timeout behavior and reports the error via telemetry.
+        """
+        if not ignore_errors:
+            if package_file is not None:
+                cleanup_package_with_invalid_signature(package_file)
+            raise ex
+        if isinstance(ex, SignatureValidationTimeoutError):
+            # TODO: This is temporary behavior for the telemetry release. For production release, remove this
+            # if-block so timeout is treated like any other signature validation failure (extension should fail).
+            SignatureValidationTimeout.disable_validation()
```
narrieta, line 1448: we should emit telemetry indicating that signature validation will be disabled until agent restart

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1467,12 +1486,7 @@ def download(self, ignore_signature_validation_errors):
                     validate_signature(package_file, self.ext_handler.encoded_signature, package_full_name=self.get_full_name())
                     signature_validation_succeeded = True
                 except SignatureValidationError as ex:
-                    # validate_signature() only raises SignatureValidationError.
-                    if not ignore_signature_validation_errors:
-                        cleanup_package_with_invalid_signature(package_file)
-                        raise
-                    report_validation_event(op=ex.operation, level=logger.LogLevel.WARNING, message=ustr(ex),
-                                            name=self.ext_handler.name, version=self.ext_handler.version, duration=ex.duration)
+                    self._handle_signature_validation_error(ex, ignore_signature_validation_errors, package_file)
```
narrieta, line 1493: why we do not set signature_validation_succeeded to False when there is a SignatureValidationError?

narrieta, line 1493: @line 1494: "Could not expand existing extension package 'foo.zip', will ignore it

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1508,12 +1522,9 @@ def download(self, ignore_signature_validation_errors):
...
             except SignatureValidationError as ex:
-                # download_zip_package() will propagate a SignatureValidationError if validation fails. Re-raise if
-                # validation errors should not be ignored, otherwise report the error and continue.
-                if not ignore_signature_validation_errors:
-                    raise   # Package has already been cleaned up
-                report_validation_event(op=ex.operation, level=logger.LogLevel.WARNING, message=ustr(ex), name=self.ext_handler.name,
-                                        version=self.ext_handler.version, duration=ex.duration)
+                # download_zip_package() will propagate a SignatureValidationError if validation fails.
+                # Package has already been cleaned up by download_zip_package().
+                self._handle_signature_validation_error(ex, ignore_signature_validation_errors)
```
narrieta, line 1532: why we do not set signature_validation_succeeded to False when there is a SignatureValidationError?

azurelinuxagent/ga/exthandlers.py
mgunnala: signature_validation_succeeded is initialized to False at the top of the method (line 1474) and only set to True after successful validation. If a SignatureValidationError is raised, the = True line is never reached, so it remains False. 

I can explicitly set signature_validation_succeeded to False in the error handling code, if it makes it easier to understand.

azurelinuxagent/ga/signature_validation_util.py
mgunnala: That's true, I've updated the event type to "Download" instead of "SignatureValidation", let me know if that addresses your concern.

azurelinuxagent/ga/exthandlers.py
mgunnala: Added this telemetry.

azurelinuxagent/ga/exthandlers.py
mgunnala: See explanation above

azurelinuxagent/ga/exthandlers.py
mgunnala: Updated

narrieta: the code definitively needs cleanup there are too many branches in this logic, some branches set signature_validation_succeeded to True, some to False, and some don't change the value even if the validation failed.

azurelinuxagent/ga/signature_validation_util.py
narrieta: It is not a download error either, we are just failing to cleanup the zip

mgunnala: Does "ExtensionProcessing" work? It seems to be the best fit of the operations we already have defined.

azurelinuxagent/ga/exthandlers.py
mgunnala: Updated so that every failure branch now explicitly sets the variable. The larger refactor to unify the two code paths (existing vs downloaded package) is tracked by the TODO on line 1478 and will be a follow-up PR. That should help clean up this code / reduce the number of branches.

narrieta: Please prioritize the follow-up PR

azurelinuxagent/ga/signature_validation_util.py
narrieta: Let's define a new operation "ExtensionCleanup"

mgunnala: @narrieta I've changed the operation to ExtensionCleanup

narrieta (approved):

PR #3577, opened 2026-03-09
add support for Azure container linux

mayankfz:
## Description

This adds support Azure Container Linux, which is derived from Azure Linux, and is compatible with it. 

---
### PR information

---

### DIFF
--- a/azurelinuxagent/common/osutil/factory.py
+++ b/azurelinuxagent/common/osutil/factory.py
@@ -149,8 +149,8 @@ def _get_osutil(distro_name, distro_code_name, distro_version, distro_full_name)
 
     # distro_name is same for azurelinux/mariner and Azure Container Linux (ACL).
     # ACL is a flavor of Azure Linux, differentiation is based on distro_full_name.
-    # For Azure Container Linux, distro_name is "azurelinux" and distro_full_name is "Microsoft Azure Container Linux".
-    if distro_name == "azurelinux" and distro_full_name == "Microsoft Azure Container Linux":
+    # For Azure Container Linux, distro_name will be made "azurecontainerlinux".
+    if distro_name == "azurecontainerlinux":
         return AclOSUtil()
 
     if distro_name in ["mariner", "azurelinux"]:
--- a/azurelinuxagent/common/osutil/factory.py
+++ b/azurelinuxagent/common/osutil/factory.py
@@ -19,6 +19,7 @@
 import azurelinuxagent.common.logger as logger
 from azurelinuxagent.common.version import DISTRO_NAME, DISTRO_CODE_NAME, DISTRO_VERSION, DISTRO_FULL_NAME
 from azurelinuxagent.common.utils.distro_version import DistroVersion
+from .acl import AclOSUtil
 from .alpine import AlpineOSUtil
 from .arch import ArchUtil
 from .bigip import BigIpOSUtil
@@ -145,6 +146,9 @@ def _get_osutil(distro_name, distro_code_name, distro_version, distro_full_name)
 
     if distro_name == "iosxe":
         return IosxeOSUtil()
+
+    if "azure container linux" in distro_full_name.lower():
+        return AclOSUtil()
 
     if distro_name in ["mariner", "azurelinux"]:
         return MarinerOSUtil()
--- a/azurelinuxagent/common/osutil/factory.py
+++ b/azurelinuxagent/common/osutil/factory.py
@@ -147,7 +147,10 @@ def _get_osutil(distro_name, distro_code_name, distro_version, distro_full_name)
     if distro_name == "iosxe":
         return IosxeOSUtil()
 
-    if "azure container linux" in distro_full_name.lower():
+    # distro_name is same for azurelinux/mariner and Azure Container Linux (ACL).
+    # ACL is a flavor of Azure Linux, differentiation is based on distro_full_name.
+    # For Azure Container Linux, distro_name is "azurelinux" and distro_full_name is "Microsoft Azure Container Linux".
+    if distro_name == "azurelinux" and distro_full_name == "Microsoft Azure Container Linux":
         return AclOSUtil()
 
     if distro_name in ["mariner", "azurelinux"]:
--- a/azurelinuxagent/common/osutil/mariner.py
+++ b/azurelinuxagent/common/osutil/mariner.py
@@ -17,12 +17,14 @@
 #
 
 from azurelinuxagent.common.osutil.default import DefaultOSUtil
+from azurelinuxagent.common.version import DISTRO_FULL_NAME
 
 
 class MarinerOSUtil(DefaultOSUtil):
     def __init__(self):
         super(MarinerOSUtil, self).__init__()
         self.jit_enabled = True
+        self.distro_full_name = DISTRO_FULL_NAME
 
     @staticmethod
     def get_systemd_unit_file_install_path():
@@ -42,6 +44,10 @@ class MarinerOSUtil(DefaultOSUtil):
         self._run_command_without_raising(["systemctl", "restart", "systemd-networkd"])
 
     def restart_ssh_service(self):
+        # ACL uses sshd.socket for socket-activated SSH (similar to Flatcar/CoreOS).
+        # Restarting sshd.service would conflict with the active sshd.socket.
+        if "azure container linux" in self.distro_full_name.lower():
+            return
         self._run_command_without_raising(["systemctl", "restart", "sshd"])
 
     def stop_dhcp_service(self):
--- a/config/acl/waagent.conf
+++ b/config/acl/waagent.conf
@@ -1,0 +1,36 @@
+#
+# Microsoft Azure Linux Agent Configuration (ACL)
+#
+
+Extensions.Enabled=y
+
+Provisioning.Agent=waagent
+Provisioning.DeleteRootPassword=n
+Provisioning.RegenerateSshHostKeyPair=n
+Provisioning.SshHostKeyPairType=auto
+Provisioning.MonitorHostName=y
+Provisioning.DecodeCustomData=y
+Provisioning.ExecuteCustomData=n
+Provisioning.AllowResetSysUser=n
+
+ResourceDisk.Format=y
+ResourceDisk.Filesystem=ext4
+ResourceDisk.MountPoint=/mnt/resource
+ResourceDisk.EnableSwap=n
+ResourceDisk.SwapSizeMB=0
+ResourceDisk.MountOptions=None
+
+LBProbeResponder=y
+
+Logs.Verbose=n
+# Logs.Console=y
+
+OS.EnableFIPS=n
+OS.SshDir=/etc/ssh
+OS.RootDeviceScsiTimeout=300
+OS.OpensslPath=None
+
+#AutoUpdate.Enabled=n
+
+# Add firewall rules to protect access to Azure host node services
+OS.EnableFirewall=y
--- a/config/acl/waagent.conf
+++ b/config/acl/waagent.conf
@@ -4,7 +4,8 @@
 
 Extensions.Enabled=y
 
-Provisioning.Agent=waagent
+# Auto-detect the provisioning agent (coreos-cloudinit or waagent).
+Provisioning.Agent=auto
 Provisioning.DeleteRootPassword=n
 Provisioning.RegenerateSshHostKeyPair=n
 Provisioning.SshHostKeyPairType=auto
--- a/init/acl/10-waagent-sysext.conf
+++ b/init/acl/10-waagent-sysext.conf
@@ -1,0 +1,2 @@
+[Unit]
+Upholds=waagent.service
--- a/init/acl/10-waagent-sysext.conf
+++ b/init/acl/10-waagent-sysext.conf
@@ -1,2 +1,13 @@
+# Drop-in for multi-user.target on Azure Container Linux (ACL).
+#
+# WALinuxAgent is delivered as a systemd-sysext image.  Sysext units are
+# only visible after systemd-sysext.service merges the overlay, which
+# happens *after* the default target is already queued.  Without an
+# explicit "Upholds=", systemd has no dependency edge to pull
+# waagent.service into the boot transaction.
+#
+# "Upholds=" (systemd 249+) is a soft requirement: it starts the unit
+# whenever multi-user.target is active but does not fail the target if
+# waagent.service cannot start.
 [Unit]
 Upholds=waagent.service
--- a/init/acl/waagent.service
+++ b/init/acl/waagent.service
@@ -3,11 +3,13 @@ Description=Microsoft Azure Linux Agent (ACL)
 Wants=network-online.target sshd.service sshd-keygen.service
 After=network-online.target sshd-keygen.service systemd-sysext.service
 
+ConditionFileIsExecutable=/usr/bin/waagent
+ConditionPathExists=/etc/waagent.conf
+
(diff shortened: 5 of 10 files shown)


mayankfz:
Attaching the waagent logs
[ACL_waagent.log](https://github.com/user-attachments/files/26200150/ACL_waagent.log)

azurelinuxagent/common/future.py
```diff
@@ -235,3 +235,27 @@ def array_to_bytes(buff):
         return buff.tostring()
 
     return buff.tobytes()
+
+def get_variant_id():
```
nagworld9: Why do you need variant id? can we use distro name and version? Its diverging from current implementation. We have requirement that distro module need to be present https://github.com/Azure/WALinuxAgent/blob/master/requirements.txt#L1

azurelinuxagent/ga/persist_firewall_rules.py
```diff
@@ -37,7 +37,7 @@ class PersistFirewallRulesHandler(object):
 # Do not edit.
 [Unit]
 Description=Setup network rules for WALinuxAgent
-After=local-fs.target 
+After=local-fs.target systemd-sysext.service
```
nagworld9: why do we need this change? This change will affect other distros. This particular service needs to be executed before network ready. New change may affect that dependency. Please limit changes to acl distro

mayankfz: Updated, this is not required for read-only fs, as currently there seems to be some limitation on waagent side.

azurelinuxagent/common/future.py
mayankfz: Updated, instead of variant_id using the full name now. Thanks

init/acl/waagent.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Microsoft Azure Linux Agent (ACL)
+Wants=network-online.target sshd.service sshd-keygen.service
+After=network-online.target sshd-keygen.service systemd-sysext.service
+
+[Service]
+Type=simple
+
+Environment=PYTHONUNBUFFERED=x
```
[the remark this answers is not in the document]
nagworld9: We had concern of setting environment on the service file, as it will get inheriteed by extensions which are executed by the agent. We had issue in Suse distro in the past https://github.com/Azure/WALinuxAgent/pull/3286. Just pointing out if you want consider alternative which is python -u

init/acl/waagent.service
```diff
@@ -0,0 +1,16 @@
...
+ExecStart=/usr/bin/waagent -daemon
+
```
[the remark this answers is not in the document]
nagworld9: This is worth considering

setup.py
```diff
@@ -166,12 +166,21 @@ def get_data_files(name, version, fullname):  # pylint: disable=R0912
                           src=["init/clearlinux/waagent.service"])
     elif name in ["mariner", "azurelinux"]:
         set_bin_files(data_files, dest=agent_bin_path)
-        set_conf_files(data_files, dest="/etc",
-                       src=["config/mariner/waagent.conf"])
-        set_systemd_files(data_files, dest=systemd_dir_path,
-                          src=["init/mariner/waagent.service"])
         set_logrotate_files(data_files)
-        set_udev_files(data_files)
+        if fullname == "Microsoft Azure Container Linux":
```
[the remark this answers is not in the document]
nagworld9: Good suggestion

init/acl/waagent.service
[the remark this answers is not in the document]
mayankfz: updated , thanks

init/acl/waagent.service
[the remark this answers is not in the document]
mayankfz: Updated, thanks

setup.py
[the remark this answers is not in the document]
mayankfz: updated, thanks

nagworld9:
lgtm, tested e2e https://dev.azure.com/cplatruntime/WALinuxAgent/_build/results?buildId=14581&view=results

nagworld9 (dismissed):

nagworld9 (dismissed):

azurelinuxagent/common/osutil/mariner.py
```diff
@@ -42,6 +44,10 @@ def restart_if(self, ifname=None, retries=None, wait=None):
         self._run_command_without_raising(["systemctl", "restart", "systemd-networkd"])
 
     def restart_ssh_service(self):
+        # ACL uses sshd.socket for socket-activated SSH (similar to Flatcar/CoreOS).
```
narrieta: Rather than modifying the base class, we should derive a class specialized for ACL and implement the ACL-specific behavior there.

azurelinuxagent/common/osutil/mariner.py
```diff
@@ -17,12 +17,14 @@
...
 
 from azurelinuxagent.common.osutil.default import DefaultOSUtil
+from azurelinuxagent.common.version import DISTRO_FULL_NAME
 
 
 class MarinerOSUtil(DefaultOSUtil):
     def __init__(self):
         super(MarinerOSUtil, self).__init__()
         self.jit_enabled = True
+        self.distro_full_name = DISTRO_FULL_NAME
```
narrieta: Since ACL is derived from Flatcar, shouldn't we be specializing from CoreOSUtil instead?

```
    if distro_name in ("flatcar", "coreos") or distro_code_name in ("flatcar", "coreos"):
        return CoreOSUtil()
```

mayankfz: > Since ACL is derived from Flatcar, shouldn't we be specializing from CoreOSUtil instead?
> 
> ```
>     if distro_name in ("flatcar", "coreos") or distro_code_name in ("flatcar", "coreos"):
>         return CoreOSUtil()
> ```

Deriving from MarinerOSUtil (current approach) seems a better approach. Although ACL is based on Flatcar, CoreOSUtil hard-codes OEM overlay paths (/usr/share/oem/) that ACL does not use — ACL puts its config in waagent.conf and binaries in bin, which matches MarinerOSUtil. We would have to override nearly every CoreOSUtil customization, defeating the purpose of inheriting from it.

The only behavioral overlap with CoreOS is socket-activated SSH (no sshd restart), which is already handled in MarinerOSUtil.restart_ssh_service() via the distro_full_name check. If ACL diverges further, the cleanest path would be dedicated AzureContainerLinuxOSUtil(MarinerOSUtil) subclass, thus avoiding switching to CoreOSUtil.

azurelinuxagent/common/osutil/mariner.py
mayankfz: For a single function deriving seemed a bit overkill, so had modified in the marinerosutil itself. Thanks

azurelinuxagent/common/osutil/mariner.py
narrieta: Azure Linux and Flatcar are fundamentally different from the point of view of WALinuxAgent and VM Extensions due to the immutability of the latter. Since Azure Container Linux is derived from Flatcar and is also immutable, it looks a lot closer to Flatcar than Azure Linux. By deriving from MarinerOSUtils you risk taking future changes in Azure Linux that are not applicable to immutable distros.

You could make a copy of the current MarinerOSUtil and make your changes there.

azurelinuxagent/common/osutil/mariner.py
narrieta: Do create a new class, please. Also, create a new file for it and assign an owner to it in CODEOWNERS

config/acl/waagent.conf
```diff
@@ -0,0 +1,36 @@
+#
+# Microsoft Azure Linux Agent Configuration (ACL)
+#
+
+Extensions.Enabled=y
+
+Provisioning.Agent=waagent
```
narrieta: Provisioning with WALinuxAgent is now deprecated. Most distros have moved to cloud-init or have their own provisioning code. The provisioning code in WALinuxAgent is no longer maintained and changes to it to accommodate any new behavior are very unlikely.

config/acl/waagent.conf
```diff
@@ -0,0 +1,36 @@
+#
```
narrieta, line 1: Assign an owner to the acl directory in CODEOWNERS

init/acl/10-waagent-sysext.conf
```diff
@@ -0,0 +1,2 @@
+[Unit]
```
narrieta, line 12: Please document why this is needed

setup.py
```diff
@@ -166,12 +166,21 @@ def get_data_files(name, version, fullname):  # pylint: disable=R0912
                           src=["init/clearlinux/waagent.service"])
     elif name in ["mariner", "azurelinux"]:
         set_bin_files(data_files, dest=agent_bin_path)
-        set_conf_files(data_files, dest="/etc",
-                       src=["config/mariner/waagent.conf"])
-        set_systemd_files(data_files, dest=systemd_dir_path,
-                          src=["init/mariner/waagent.service"])
         set_logrotate_files(data_files)
-        set_udev_files(data_files)
+        if "azure container linux" in fullname.lower():
```
narrieta: Move this to its own "elif name" rather that doing it in Azure Linux's

config/acl/waagent.conf
```diff
@@ -0,0 +1,36 @@
...
+Provisioning.Agent=waagent
+Provisioning.DeleteRootPassword=n
+Provisioning.RegenerateSshHostKeyPair=n
+Provisioning.SshHostKeyPairType=auto
+Provisioning.MonitorHostName=y
+Provisioning.DecodeCustomData=y
+Provisioning.ExecuteCustomData=n
+Provisioning.AllowResetSysUser=n
+
+ResourceDisk.Format=y
```
narrieta, line 17: I see that the sample waagent.log has some errors in the Resource Disk configuration. Please address those.

narrieta:
> Attaching the waagent logs [ACL_waagent.log](https://github.com/user-attachments/files/26200150/ACL_waagent.log)

We need to address these errors in the log

```
2026-03-24T03:42:50.204451Z ERROR Daemon Daemon Failed to mount resource disk [ResourceDiskError] unable to detect disk topology
...
2026-03-24T03:42:53.032311Z ERROR ExtHandler ExtHandler Unable to setup the persistent firewall rules: [Errno 30] Read-only file system: '/usr/lib/systemd/system/waagent-network-setup.service'
```

config/acl/waagent.conf
mayankfz: Done

init/acl/10-waagent-sysext.conf
mayankfz: Updated.

mayankfz:
> > Attaching the waagent logs [ACL_waagent.log](https://github.com/user-attachments/files/26200150/ACL_waagent.log)
> 
> We need to address these errors in the log
> 
> ```
> 2026-03-24T03:42:50.204451Z ERROR Daemon Daemon Failed to mount resource disk [ResourceDiskError] unable to detect disk topology
> ...
> 2026-03-24T03:42:53.032311Z ERROR ExtHandler ExtHandler Unable to setup the persistent firewall rules: [Errno 30] Read-only file system: '/usr/lib/systemd/system/waagent-network-setup.service'
> ```

Fixed. 
[ACL_waagent_1.log](https://github.com/user-attachments/files/26440663/ACL_waagent_1.log)

config/acl/waagent.conf
mayankfz: I see this error is also in Flatcar.
image

The issue seems to be of choosing the VM that has the support of resource disk. In the latest waagent logs attached for ACL, the error is not seen.

setup.py
mayankfz: Updated

config/acl/waagent.conf
mayankfz: Updated.

azurelinuxagent/common/osutil/mariner.py
mayankfz: Updated

azurelinuxagent/common/osutil/factory.py
```diff
@@ -146,6 +147,9 @@ def _get_osutil(distro_name, distro_code_name, distro_version, distro_full_name)
     if distro_name == "iosxe":
         return IosxeOSUtil()
 
+    if "azure container linux" in distro_full_name.lower():
```
narrieta: we do the rest of the comparisons based on full comparison of distro_name... why do we do this one based on distro_full_name and the in operator?

setup.py
```diff
@@ -164,13 +164,23 @@ def get_data_files(name, version, fullname):  # pylint: disable=R0912
...
+        set_conf_files(data_files, dest="/etc",
+                       src=["config/acl/waagent.conf"])
+        set_systemd_files(data_files, dest=systemd_dir_path,
+                          src=["init/acl/waagent.service"])
+        multi_user_target_drop_in_dir = os.path.join(systemd_dir_path, "multi-user.target.d")
+        set_systemd_files(data_files, dest=multi_user_target_drop_in_dir,
+                          src=["init/acl/10-waagent-sysext.conf"])
     elif name in ["mariner", "azurelinux"]:
         set_bin_files(data_files, dest=agent_bin_path)
+        set_logrotate_files(data_files)
```
narrieta: why this change?

azurelinuxagent/common/osutil/factory.py
mayankfz: distro name is same for azurelinux/mariner and Azure Container Linux, differentiation is based on distro_full_name.
Attaching the os-release file.
Sure, will do the full comparison but did "in comparison" based on Copilot review suggestion. Sure will update it again.
image

setup.py
mayankfz: updated, by mistake

azurelinuxagent/common/osutil/factory.py
```diff
@@ -146,6 +147,9 @@ def _get_osutil(distro_name, distro_code_name, distro_version, distro_full_name)
     if distro_name == "iosxe":
         return IosxeOSUtil()
 
+    if distro_name == "azurelinux" and distro_full_name == "Microsoft Azure Container Linux":
```
narrieta: is there a case where distro_name != "azurelinux" and distro_full_name == "Microsoft Azure Container Linux"?

mayankfz: that shouldn't be the case, but we can make it less strict by just checking the distro_full_name, if you are ok with that ?

narrieta: yes, it's fine. 

i'm just trying to understand if we were planning to support different flavors of "azurelinux", or something like that.

if you base your comparison on full_name, please add a short note on why, to help people reading this code (since the rest of the comparisons are on distro_name)

narrieta: oh, i see why the full name... please do add a comment

narrieta: actually, the current condition would be clearer, just adding a comment that ACL is a flavor of azure linux

elif name == 'azurelinux' and fullname == "Microsoft Azure Container Linux":

mayankfz: Added, thanks

narrieta (commented):
lgtm

mayankfz:
@narrieta Now the waagent --version on ACL will look like - >

image

narrieta (approved):

nagworld9 (approved):

PR #3600, opened 2026-04-22
 Implement graceful thread shutdown during ext_handler exit

nagworld9:
## Description

When ext_handler exit, the background threads started by the extension handler (MonitorHandler, EnvHandler, SendTelemetryHandler, CollectTelemetryEventsHandler, CollectLogsHandler) were not being stopped cleanly. They were simply abandoned when the process exited. This PR ensures all threads complete their in-progress work and shut down gracefully before the ext_handler process exit

Issue # 
---

### PR information

---


azurelinuxagent/ga/env.py
```diff
@@ -309,5 +310,6 @@ def stop(self):
         Stop server communication and join the thread to main thread.
         """
         self.stopped = True
+        self._stop_event.set()
         if self.env_thread is not None:
-            self.env_thread.join()
+            self.env_thread.join(timeout=self._THREAD_JOIN_TIMEOUT)
```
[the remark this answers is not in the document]
nagworld9: fixed

azurelinuxagent/ga/send_telemetry_events.py
```diff
@@ -84,7 +84,7 @@ def stop(self):
 
     def join(self):
         self._queue.join()
-        self._thread.join()
+        self._thread.join(timeout=self._THREAD_JOIN_TIMEOUT)
```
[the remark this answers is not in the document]
nagworld9: fixed

azurelinuxagent/ga/update.py
```diff
@@ -330,6 +336,18 @@ def run(self, debug=False):
...
+            if signum == signal.SIGTERM:
+                if self._handle_sigterm_in_progress:
+                    logger.info("SIGTERM handler already in progress for ext handler, ignoring duplicate signal")
+                    return
+                self._handle_sigterm_in_progress = True
+                logger.info("SIGTERM received for ext handler, shutting down gracefully...")
+                self._shutdown()
+                sys.exit(0)
+
+        signal.signal(signal.SIGTERM, handle_sigterm)
```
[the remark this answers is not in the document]
nagworld9: I dont see why its needed. In the tests, we don't call sigterm and also, every new run need to register sigterm and don't need to keep track of previous handler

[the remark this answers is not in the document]
narrieta: We need tests for the entire stop sequence. End-to-end tests, definitively. Unit tests should also be possible.

So we may need to implement this suggestion as well.

azurelinuxagent/ga/update.py
```diff
@@ -178,6 +178,12 @@ def __init__(self):
 
         self._is_initial_goal_state = not os.path.exists(self._initial_goal_state_file_path())
 
+        # List of thread handlers managed by this UpdateHandler; populated in run() and used during shutdown
+        self._all_thread_handlers = []
```
narrieta, line 185: doesn't need to be a member; it could be passed as argument to _shutdown

azurelinuxagent/ga/update.py
```diff
@@ -178,6 +178,12 @@ def __init__(self):
...
+
+        # Guard to prevent a second SIGTERM from interrupting the first handler mid-execution
```
narrieta: There is a race condition between the check and the assigment

```
                if self._handle_sigterm_in_progress:
                    logger.info("SIGTERM handler already in progress for ext handler, ignoring duplicate signal")
                    return
                self._handle_sigterm_in_progress = True

```

azurelinuxagent/ga/collect_logs.py
```diff
@@ -152,7 +155,7 @@ def init_protocols(self):
     def daemon(self):
         # Delay the first collector on start up to give short lived VMs (that might be dead before the second 
         # collection has a chance to run) an opportunity to do produce meaningful logs to collect.
-        time.sleep(conf.get_log_collector_initial_delay())
+        self._stop_event.wait(conf.get_log_collector_initial_delay())
```
narrieta: if the event is set, we should return immediately

azurelinuxagent/ga/collect_logs.py
```diff
@@ -112,6 +112,7 @@ def __init__(self):
         self.last_state = None
         self.period = conf.get_collect_logs_period()
         self.log_collector_cgroup_path_validation_errors = 0
+        self._stop_event = threading.Event()
```
narrieta: Let's try to capture this synchronization in the base class; right now it is repeated on all threads

azurelinuxagent/ga/interfaces.py
```diff
@@ -21,8 +21,31 @@
...
+    The stop() method follows this sequence:
+      1. Set the should_run/stopped flag so the daemon loop exits on its next check.
+      2. Set the _stop_event to wake up the thread if it is sleeping.
+      3. Call join(timeout=_THREAD_JOIN_TIMEOUT) to wait for the thread to finish its current operation and exit.
+
+    This approach allows threads to complete their in-progress work (e.g. sending telemetry, processing events)
+    before exiting, while ensuring they wake up promptly instead of sleeping through the shutdown window.
     """
 
+    # Default timeout to wait for threads to stop during shutdown
```
narrieta:
```suggestion
    # Default timeout in seconds to wait for threads to stop during shutdown 
```

azurelinuxagent/ga/update.py
```diff
@@ -1068,10 +1086,28 @@ def _initial_goal_state_file_path():
         return os.path.join(conf.get_lib_dir(), INITIAL_GOAL_STATE_FILE)
 
     def _shutdown(self):
-        # Todo: Ensure all threads stopped when shutting down the main extension handler to ensure that the state of
-        # all threads is clean.
         self.is_running = False
 
+        # Stop all thread handlers gracefully. Each handler's stop() signals the thread to exit (by setting its
+        # should_run/stopped flag and its stop event) and then joins the thread, allowing it to complete its current
+        # operation before stopping. The handlers are stopped in reverse order; if a dependency exists between
```
narrieta: what kind of dependencies are we concerned about? these threads should not have dependencies between each other. If they have, what kind are they and how does reverse traversal addresses them?

azurelinuxagent/ga/update.py
```diff
@@ -1068,10 +1086,28 @@ def _initial_goal_state_file_path():
...
+        # handlers, stopping in reverse allows the dependent handlers to stop first, which reduces the chance of
+        # hitting issues during shutdown.
+        for thread_handler in reversed(self._all_thread_handlers):
+            try:
+                thread_name = thread_handler.get_thread_name()
+                logger.info("Stopping {0} thread...", thread_name)
+                thread_handler.stop()
```
narrieta: the call to stop waits for the thread to join; we shouldn't do this... we should signal all threads to stop and then join with all of them

azurelinuxagent/ga/env.py
```diff
@@ -309,5 +311,6 @@ def stop(self):
         Stop server communication and join the thread to main thread.
         """
         self.stopped = True
+        self._stop_event.set()
```
narrieta: In this loop above

```
                    for op in periodic_operations:
                        op.run()
```

we should not start the next operation if the thread has been signaled to stop

azurelinuxagent/ga/update.py
nagworld9: collect_telemetry_event thread depends on send_telemetry_events thread running today. If it stopped, collect throw error. That's why we start send_telemetry_events first and pass that to collect_telemtry. When we stop we should stop collect first and taht will to stop adding more events to queue. Reverse traversing would achieve that

azurelinuxagent/ga/update.py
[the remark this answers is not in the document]
nagworld9: i don't get the copilot comment first place. Unit tests were added.  Will add e2e tests

azurelinuxagent/ga/update.py
narrieta: If we cannot remove that dependency, let's make it explicit instead of relying on the order of the items in the array. Stop collect, then send, then the rest of the threads.

azurelinuxagent/ga/update.py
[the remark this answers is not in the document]
narrieta: The new unit tests don't actually use signals. If possible, we should add some unit tests that do, but if too complicated, the end-to-end tests should be enough.

The comment by copilot is relevant if we had unit tests that send the signal to the current process.  Since the signal handler is global to the process, it should be restored, or tests could be unstable.

azurelinuxagent/ga/update.py
nagworld9: side effect - the _shutdown aslo called in Deamon process and need to define as default parameter with None

narrieta: Thanks; i did not notice the call from the daemon. 

Instead of defaulting to None we could pass an empty array. A class member is OK, though, so no need to change it.

However,  we are calling shutdown from the 2 processes and we have a race condition on the check for the sentinel file.

 We need to be very careful on what kind of cleanup is done in that function, and the order of that cleanup, given that race condition. Your changes look OK, but please add a comment about the race condition.

azurelinuxagent/ga/periodic_operation.py
```diff
@@ -65,10 +65,13 @@ def _operation(self):
         raise NotImplementedError()
 
     @staticmethod
-    def sleep_until_next_operation(operations):
+    def sleep_until_next_operation(operations, stop_event=None):
```
maddieford: Why is stop_event optional? Are there threads that shouldn't use this option?

azurelinuxagent/ga/send_telemetry_events.py
```diff
@@ -83,8 +83,11 @@ def stop(self):
             self.join()
```
maddieford: why no _stop_event.set() in this operation?

azurelinuxagent/ga/periodic_operation.py
nagworld9: Today it's used only in threads. If periodicoperations used outside threads, then stop_event should be optional. I think for now, we can target for threads in scope and make it as mandatory parameter. If the new caller needs differently, they can change i guess.

azurelinuxagent/ga/send_telemetry_events.py
nagworld9: There is no stop_event.wait used in this operation, so no need of _stop_event.set() to wake up the thread

tests_e2e/tests/agent_graceful_shutdown/agent_graceful_shutdown.py
```diff
@@ -0,0 +1,233 @@
...
+    """
+
+    def run(self):
+        ssh_client = self._context.create_ssh_client()
+
+        # One-time setup: enable log collection so the CollectLogsHandler thread is started by the
+        # agent every time it comes up.
+        log.info("Enabling log collection so all threads run on every iteration...")
+        ssh_client.run_command(
+            "sh -c 'agent-service stop && update-waagent-conf Logs.Collect=y'",
```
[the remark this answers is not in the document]
nagworld9: Need && for cmd to fail when we failed to stop the agent

azurelinuxagent/ga/periodic_operation.py
```diff
@@ -77,5 +80,6 @@ def sleep_until_next_operation(operations):
         sleep_seconds = ((sleep_timedelta.days * 24 * 3600 + sleep_timedelta.seconds) * 10.0 ** 6 + sleep_timedelta.microseconds) / 10.0 ** 6
 
         if sleep_seconds > 0:
-            time.sleep(sleep_seconds)
+            stop_event.wait(sleep_seconds)
```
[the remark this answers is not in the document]
nagworld9: updated docstring

azurelinuxagent/ga/periodic_operation.py
```diff
@@ -65,10 +65,13 @@ def _operation(self):
...
-    def sleep_until_next_operation(operations):
+    def sleep_until_next_operation(operations, stop_event):
         """
         Takes a list of operations, finds the operation that should be executed next (that with the closest next_run_time)
         and sleeps until it is time to execute that operation.
+
+        If a stop_event (threading.Event) is provided, the sleep will be interrupted when the event is set, allowing
+        the caller to wake up promptly during shutdown.
         """
         next_operation_time = min(op.next_run_time() for op in operations)
```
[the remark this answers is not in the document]
nagworld9: updated docstring

tests/ga/test_update.py
```diff
@@ -899,6 +899,181 @@ def test_shutdown(self):
...
+        captured = {}
+
+        class _BailOut(Exception):
+            pass
+
+        def fake_signal(signum, handler):
+            if signum == _signal.SIGTERM:
+                captured["handler"] = handler
+                # Exit run() immediately after the handler is installed; any exception from run()'s
+                # initialization is logged and swallowed by run(), so we wrap with a try/except below.
```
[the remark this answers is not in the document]
nagworld9: current comment explanatory

tests_e2e/tests/agent_graceful_shutdown/agent_graceful_shutdown.py
```diff
@@ -0,0 +1,237 @@
...
+        # agent every time it comes up. Also lower Debug.LogCollectorInitialDelay (default 5 minutes)
+        # so the collector exits its initial sleep quickly; otherwise an iteration that stops the
+        # agent before that delay elapses would not see the "Signaling CollectLogsHandler thread to
+        # stop" line because the thread would still be parked in the initial sleep.
+        log.info("Enabling log collection so all threads run on every iteration...")
+        ssh_client.run_command(
+            "sh -c 'agent-service stop && "
+            "update-waagent-conf Logs.Collect=y Debug.LogCollectorInitialDelay=5'",
+            use_sudo=True,
+        )
```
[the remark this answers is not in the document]
nagworld9: Running logcollector in other tests is ok

tests_e2e/test_suites/agent_graceful_shutdown.yml
```diff
@@ -0,0 +1,16 @@
...
+#
+name: "AgentGracefulShutdown"
+tests:
+  - "agent_graceful_shutdown/agent_graceful_shutdown.py"
+images:
+  - "endorsed"
+  - "endorsed-arm64"
+skip_on_images:
+  # Oracle 7.9 is excluded from this test, since agent stop behaving differently and
+  # not waiting for the background threads to stop before exiting, which causes the test to fail.
```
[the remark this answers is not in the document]
nagworld9: fixed

azurelinuxagent/ga/collect_logs.py
```diff
@@ -304,8 +316,13 @@ def __init__(self, controllers):
...
-    def stop(self):
+    def signal_stop(self):
+        # Flag the loop to exit on its next iteration; do not block waiting for the thread.
         self.should_run = False
+        self._signal_stop()
+
+    def stop(self):
+        self.signal_stop()
         if self.is_alive():
             self.join()
```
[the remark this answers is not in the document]
nagworld9: fixed

azurelinuxagent/ga/update.py
```diff
@@ -330,6 +340,35 @@ def run(self, debug=False):
...
+                    self._shutdown()
+                except Exception as e:
+                    exit_code = 1
+                    logger.warn("SIGTERM shutdown for ext handler failed: {0}", ustr(e))
+                    logger.warn(textutil.format_exception(e))
+                finally:
+                    sys.exit(exit_code)
+
+        signal.signal(signal.SIGTERM, handle_sigterm)
+
```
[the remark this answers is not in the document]
nagworld9: Same old comment, not fixing as no tests sends signal

tests_e2e/test_suites/agent_graceful_shutdown.yml
```diff
@@ -0,0 +1,11 @@
+#
+# This test validates that the ext-handler shuts down gracefully when SIGTERM is delivered to it
+# (e.g. by `systemctl stop`). It checks that every background thread the agent starts is signaled
+# to stop, exits within the join timeout.
```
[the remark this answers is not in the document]
nagworld9: fixed

azurelinuxagent/ga/collect_logs.py
```diff
@@ -304,8 +316,12 @@ def __init__(self, controllers):
     def run(self):
         self.start()
 
-    def stop(self):
+    def signal_stop(self):
+        # Flag the loop to exit on its next iteration; do not block waiting for the thread.
         self.should_run = False
+
+    def stop(self):
+        self.signal_stop()
```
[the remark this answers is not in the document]
nagworld9: fixed

azurelinuxagent/ga/interfaces.py
```diff
@@ -46,4 +115,12 @@ def start(self):
...
\ No newline at end of file
+        raise NotImplementedError("stop() not implemented")
+
+    def signal_stop(self):
+        """
+        Signals the thread to stop without waiting for it to exit. This allows callers to broadcast a stop
+        signal to several handlers in parallel and only join afterwards, rather than serializing the join
+        timeouts inside each handler's stop().
+        """
+        raise NotImplementedError("signal_stop() not implemented")
```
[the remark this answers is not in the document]
nagworld9: This is new abstract method, all threads needs to implement

azurelinuxagent/ga/periodic_operation.py
```diff
@@ -77,5 +81,6 @@ def sleep_until_next_operation(operations):
         sleep_seconds = ((sleep_timedelta.days * 24 * 3600 + sleep_timedelta.seconds) * 10.0 ** 6 + sleep_timedelta.microseconds) / 10.0 ** 6
 
         if sleep_seconds > 0:
-            time.sleep(sleep_seconds)
+            stop_event.wait(sleep_seconds)
+
```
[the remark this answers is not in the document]
nagworld9: No caller left behind, all callers passing stop_event

azurelinuxagent/ga/periodic_operation.py
```diff
@@ -65,10 +64,15 @@ def _operation(self):
...
         """
         Takes a list of operations, finds the operation that should be executed next (that with the closest next_run_time)
         and sleeps until it is time to execute that operation.
+
+        stop_event (threading.Event) is required: the sleep is implemented via stop_event.wait(), so setting the
+        event from another thread interrupts the sleep immediately. This is what allows handlers to wake up
+        promptly when shutdown is signaled instead of sleeping through the shutdown window. All call sites must
+        pass a valid threading.Event; passing None will raise AttributeError when wait() is invoked.
         """
         next_operation_time = min(op.next_run_time() for op in operations)
```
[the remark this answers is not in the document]
nagworld9: All callers are in thread handler passing 2-arg required parameter and no other caller outside thread handler using it

azurelinuxagent/ga/update.py
```diff
@@ -330,6 +340,35 @@ def run(self, debug=False):
...
+                try:
+                    self._shutdown()
+                except Exception as e:
+                    exit_code = 1
+                    logger.warn("SIGTERM shutdown for ext handler failed: {0}", ustr(e))
+                    logger.warn(textutil.format_exception(e))
+                finally:
+                    sys.exit(exit_code)
+
+        signal.signal(signal.SIGTERM, handle_sigterm)
```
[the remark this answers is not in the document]
nagworld9: Same old comment, not fixing as no tests sends signal

azurelinuxagent/ga/send_telemetry_events.py
```diff
@@ -74,17 +75,25 @@ def start(self):
         self._thread.name = self.get_thread_name()
         self._thread.start()
 
+    def signal_stop(self):
+        # Signal the daemon loop to exit without blocking on the thread to finish. The processing loop
+        # is queue-driven, so it will wake up on the next queue timeout (or sooner) and observe stopped().
+        self.should_run = False
+
```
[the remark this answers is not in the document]
nagworld9: I don’t think we can implement interruptible waits for the queue. At most, we would get a warning that the "thread didn’t stop," but the thread not processing events as there are none the queue to process. In that regard, we are safe.

In regard to timeout, since both are same seconds, we will not cross that window

tests_e2e/tests/agent_graceful_shutdown/agent_graceful_shutdown.py
```diff
@@ -0,0 +1,199 @@
...
+        self._ssh_client.run_command(
+            "update-waagent-conf Logs.Collect=y Debug.LogCollectorInitialDelay=60",
+            use_sudo=True)
+
+        log.info("Waiting for the agent to self-upgrade to the latest published version...")
+        # _verify_agent_updated_to_latest_version() polls waagent --version until the running
+        # agent matches the latest version on the manifest. This is what guarantees that the
+        # AgentUpgradeExitException path was taken on the original (custom) agent process and
+        # therefore _shutdown() has run.
+        self._verify_agent_updated_to_latest_version()
```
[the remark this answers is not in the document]
nagworld9: updated PR

tests_e2e/tests/agent_graceful_shutdown/agent_graceful_shutdown.py
```diff
@@ -0,0 +1,199 @@
...
+            if m:
+                signaled.add(m.group(1))
+                continue
+            m = _STOPPED_OK_RE.search(line)
+            if m:
+                terminal.add(m.group(1))
+                continue
+            m = _STOPPED_TIMEOUT_RE.search(line)
+            if m:
+                terminal.add(m.group(1))
```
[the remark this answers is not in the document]
nagworld9: there is no guarantee that threads finish with in join time window but we check to ensure that stop called and waited till timeout

tests_e2e/tests/agent_graceful_shutdown/agent_graceful_shutdown.py
```diff
@@ -0,0 +1,206 @@
...
+]
+
+# Started only when log collection is enabled.
+_LOG_COLLECTOR_THREAD = "CollectLogsHandler"
+
+# Emitted by collect_logs.is_log_collection_allowed() during agent startup. The agent prints "[True]"
+# when all preconditions are met and the CollectLogsHandler thread will be started, "[False]" otherwise.
+# We use this to decide whether the collector is expected to appear in the shutdown sequence.
+_LOG_COLLECTION_ALLOWED_RE = re.compile(r"log collection is allowed at this time \[True]")
+
```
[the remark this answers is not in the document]
nagworld9: fixed

azurelinuxagent/ga/interfaces.py
```diff
@@ -17,12 +17,81 @@
...
 
 class ThreadHandlerInterface(object):
     """
-    Interface for all thread handlers created and maintained by the GuestAgent.
+    Interface and shared base for all thread handlers.
+
+    Each handler creates a background thread that runs a daemon loop. To support graceful shutdown, this base class
+    provides a shared threading.Event (self._stop_event) that makes sleep operations in the daemon loop interruptible:
+
+    - Without the event, threads use time.sleep() between loop iterations. time.sleep() is uninterruptible -- if a
```
narrieta: time.sleep is interruptible

console 1 (terminated from console 2):

```
$ python3
Python 3.12.3 (main, Mar 23 2026, 19:04:32) [GCC 13.3.0] on linux
>>> import os
>>> os.getpid()
11278
>>> import time
>>> time.sleep(3600)
Terminated
$

```
console 2:

```
$ kill -15 11278
$ 

```

azurelinuxagent/ga/interfaces.py
```diff
@@ -17,12 +17,81 @@
 # Requires Python 2.6+ and Openssl 1.0+
 #
 
+import threading
+
 
 class ThreadHandlerInterface(object):
     """
-    Interface for all thread handlers created and maintained by the GuestAgent.
+    Interface and shared base for all thread handlers.
```
narrieta: Let's rename the class. It was named "Interface" because it provided only the signature of the methods the derived class must implement; now it is more of an abstract class thaht provides implementation of some of those methods.

azurelinuxagent/ga/collect_logs.py
```diff
@@ -115,22 +115,28 @@ def keep_alive(self):
...
     def stopped(self):
         return not self.should_run
 
-    def stop(self):
+    def signal_stop(self):
+        # Signal the daemon loop to exit without blocking on the thread to finish.
         self.should_run = False
+        self._signal_stop()
+
+    def stop(self):
```
narrieta: it's hard for the reader to understand when to use stop vs signal_stop unless they read and understand the implementation. Consider refactoring this functionality. One single method may be enough.

azurelinuxagent/ga/interfaces.py
```diff
@@ -17,12 +17,81 @@
...
+    - Without the event, threads use time.sleep() between loop iterations. time.sleep() is uninterruptible -- if a
+      thread is sleeping for several minutes, calling stop() would have to wait that entire duration before the
+      thread wakes up and checks its stop flag.
+
+    - With the event, threads call self._interruptible_sleep(timeout) (which delegates to self._stop_event.wait).
+      This behaves identically to time.sleep() during normal operation (blocks for the specified duration), but when
+      stop() is called, self._signal_stop() sets the event, which immediately wakes up any thread waiting on it. The
+      thread then checks its stop flag, sees that it should exit, and terminates promptly.
+
+    The stop() method in subclasses should follow this sequence:
```
narrieta: The "stop" protocol should be implemented in the base class. It seems to be practically the same in each derived class.

There are some differences in the implementation of the derived classes, though, but I think they are not intentional. For example:

```
    def stop(self):
        self.signal_stop()
        if self.is_alive():
            try:
                self.join()
            except RuntimeError:
                pass

```
vs

```
    def stop(self):
        """
        Stop server communication and join the thread to main thread.
        """
        self.signal_stop()
        if self.is_alive():
            self.thread.join(timeout=self._THREAD_JOIN_TIMEOUT)
```

azurelinuxagent/ga/interfaces.py
```diff
@@ -17,12 +17,81 @@
...
+      1. Set the should_run/stopped flag so the daemon loop exits on its next check.
+      2. Call self._signal_stop() to wake up the thread if it is sleeping.
+      3. Call join(timeout=_THREAD_JOIN_TIMEOUT) to wait for the thread to finish its current operation and exit.
+
+    Subclasses that restart the thread should call self._reset_stop_event() in start() so the new thread does not
+    inherit a "stopped" state from a previous run.
```
narrieta: they should also set self.should_run to True (currently some of them don't)

azurelinuxagent/ga/interfaces.py
```diff
@@ -17,12 +17,81 @@
...
+    Each handler creates a background thread that runs a daemon loop. To support graceful shutdown, this base class
+    provides a shared threading.Event (self._stop_event) that makes sleep operations in the daemon loop interruptible:
+
+    - Without the event, threads use time.sleep() between loop iterations. time.sleep() is uninterruptible -- if a
+      thread is sleeping for several minutes, calling stop() would have to wait that entire duration before the
+      thread wakes up and checks its stop flag.
+
+    - With the event, threads call self._interruptible_sleep(timeout) (which delegates to self._stop_event.wait).
+      This behaves identically to time.sleep() during normal operation (blocks for the specified duration), but when
+      stop() is called, self._signal_stop() sets the event, which immediately wakes up any thread waiting on it. The
```
narrieta: This needs some clarification. Calling stop() does not immeadiately stops a thread, since the thread may be busy executing some operation; is't only after that operation completed that the thread checks the event.

```
but when  stop() is called, self._signal_stop() sets the event, which immediately wakes up any thread waiting on it. The  thread then checks its stop flag, sees that it should exit, and terminates promptly.
```

azurelinuxagent/ga/update.py
```diff
@@ -459,24 +463,24 @@ def run(self, debug=False):
...
             # Launch all monitoring threads
-            self._start_threads(all_thread_handlers)
+            self._start_threads(self._all_thread_handlers)
 
             logger.info("Goal State Period: {0} sec. This indicates how often the agent checks for new goal states and reports status.", self._goal_state_period)
 
             while self.is_running:
                 self._check_daemon_running(debug)
-                self._check_threads_running(all_thread_handlers)
+                self._check_threads_running(self._all_thread_handlers)
```
narrieta: do you still need the parameter, if it is a class member now?

tests_e2e/tests/agent_graceful_shutdown/agent_graceful_shutdown.py
```diff
@@ -0,0 +1,205 @@
...
+# limitations under the License.
+#
+
+#
+# Validates that the ext-handler shuts down gracefully on the natural exit paths
+# (AgentUpgradeExitException / ExitException). The agent's UpdateHandler.run() loop catches these
+# exceptions, calls UpdateHandler._shutdown() and then sys.exit(0). _shutdown() must signal every
+# worker thread to stop and wait (with a bounded timeout) for each of them to finish.
+#
+# How the test works:
```
narrieta, line 26: This is a good test.

Can we also add a check that the agent update completer within a reasonable time? (reasonable == within the limits of the synchronization timeouts)

azurelinuxagent/ga/interfaces.py
narrieta: Maybe just don't use "interruptible". When one reads the comment, the first thing that comes to mind is

```
An interruptible function call generally refers to a function (often a system call) that can be prematurely stopped before it finishes, usually because the process received a signal.

In Linux/Unix, this is most often discussed in the context of interrupted system calls.

* What It Means
Many blocking system calls (e.g., read(), write(), accept(), wait()) may pause execution until some condition is met (data arrives, a connection is made, etc.).
If a signal is delivered to the process while it is blocked, the kernel may interrupt the call and return control to user space before completion.
In such cases, the function returns -1 and sets errno to EINTR.
```

azurelinuxagent/ga/interfaces.py
```diff
@@ -17,12 +17,81 @@
...
+import threading
+
 
 class ThreadHandlerInterface(object):
     """
-    Interface for all thread handlers created and maintained by the GuestAgent.
+    Interface and shared base for all thread handlers.
+
+    Each handler creates a background thread that runs a daemon loop. To support graceful shutdown, this base class
+    provides a shared threading.Event (self._stop_event) that makes sleep operations in the daemon loop interruptible:
```
maddieford: Could you update the comment with which threads share the threading.Event? I assume each thread has it's own and all are shared with the main thread?

azurelinuxagent/ga/interfaces.py
maddieford: This seems important. Is it worth adding it to the base class and having all the derived classes inherit?

azurelinuxagent/ga/interfaces.py
```diff
@@ -17,12 +17,81 @@
...
+        # handlers do not have to re-implement the same synchronization boilerplate.
+        self._stop_event = threading.Event()
+
+    def _interruptible_sleep(self, timeout):
+        """
+        Sleeps for the given timeout (in seconds), but returns early if stop() has been signaled.
+        """
+        self._stop_event.wait(timeout)
+
+    def _signal_stop(self):
```
maddieford: should we consider renaming this to something like set  or wake_thread()? Since we have def signal_stop() as well which touches the should_run flag

azurelinuxagent/ga/send_telemetry_events.py
```diff
@@ -69,22 +70,31 @@ def is_alive(self):
...
+        self._reset_stop_event()
         self._thread = threading.Thread(target=self._process_telemetry_thread)
         self._thread.daemon = True
         self._thread.name = self.get_thread_name()
         self._thread.start()
 
+    def signal_stop(self):
+        # Signal the daemon loop to exit without blocking on the thread to finish. The processing loop
+        # is queue-driven, so it will wake up on the next queue timeout (or sooner) and observe stopped().
+        self.should_run = False
```
maddieford: Could you add a comment about why we don't call _signal_stop() here. From my understanding, it's because there's at most 5 seconds wait (SendTelemetryEventsHandler._MAX_TIMEOUT) between each loop, but comment would be helpful

azurelinuxagent/ga/send_telemetry_events.py
maddieford, line 106: Is this comment still accurate? 
`Incase the service is stopped but we have events in queue, ensure we send them out before killing the thread`

maddieford, line 109: why do we keep processing events when self.stopped()?

azurelinuxagent/ga/update.py
```diff
@@ -1129,6 +1184,40 @@ def _shutdown(self):
...
+        dependency-ordered handlers, where the producer must fully exit before the consumer is signaled.
+        """
+        try:
+            logger.info("Signaling {0} thread to stop...", thread_handler.get_thread_name())
+            thread_handler.signal_stop()
+        except Exception as e:
+            logger.warn(
+                u"Exception signaling thread {0} to stop: {1}",
+                thread_handler.get_thread_name(),
+                ustr(e))
```
maddieford: Lines 1193-1200 are duplicates of lines 1162-1169. Could we create signal helper?

azurelinuxagent/ga/send_telemetry_events.py
nagworld9: Otherwise, we lose that telemetry. We still need to push telemetry if queue is not empty

nagworld9: The service is stopped referring to "thread is stopped"

tests_e2e/tests/agent_graceful_shutdown/agent_graceful_shutdown.py
nagworld9: Added

tests/ga/test_monitor.py
```diff
@@ -59,58 +59,66 @@ def _mock_wire_protocol():
...
 
-                                if network_changes:
-                                    expected_operations.append(ReportNetworkConfigurationChanges.__name__)
+                            if 'Linux' in platform.system():
+                                expected_operations.append(MonitorKernelSoftLockup.__name__)
 
-                                if 'Linux' in platform.system():
-                                    expected_operations.append(MonitorKernelSoftLockup.__name__)
+                            monitor_handler = get_monitor_handler()
+                            monitor_handler.run()
```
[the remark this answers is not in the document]
nagworld9: won't fix: This is existing test

azurelinuxagent/ga/env.py
```diff
@@ -254,28 +253,17 @@ def get_thread_name():
...
     def run(self):
-        if not self.stopped:
+        # If a previous worker thread is still alive, stop it cleanly before starting a new one.
+        if self.is_alive():
             logger.info("Stop existing env monitor service.")
-            self.stop()
-
-        self.stopped = False
+            self.stop_and_join()
         logger.info("Starting env monitor service.")
```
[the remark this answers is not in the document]
nagworld9: Keeping existing behavior but I don't see a case where we start new thread while its running

azurelinuxagent/ga/thread_handler_base.py
```diff
@@ -17,33 +17,143 @@
...
+        handler.is_alive()                       # True until the thread has exited
+
+    To stop a handler completely, callers do:    handler.stop(); handler.join()
+    Separating the two lets a shutdown sequence broadcast stop() to many handlers and then join them in
+    parallel, instead of serializing each handler's join timeout.
+
+    Subclasses only need to implement get_thread_name() and daemon() (the thread body). The base class
+    provides everything else.
+
+    Daemon loops should:
```
narrieta, line 42: Consider implementing this protocol in the base class

azurelinuxagent/ga/interfaces.py
```diff
@@ -17,33 +17,143 @@
...
+
+    # ----- subclass contract -----
+
     @staticmethod
     def get_thread_name():
         raise NotImplementedError("get_thread_name() not implemented")
 
+    def daemon(self):
+        """
+        The thread body. Subclasses MUST override this with their main loop.
```
narrieta: update comment: not all subclasses have a main loop

azurelinuxagent/ga/thread_handler_base.py
```diff
@@ -17,33 +17,143 @@
 # Requires Python 2.6+ and Openssl 1.0+
 #
 
+import threading
```
narrieta, line 20: should rename this file to thread_handler_base.py

narrieta (dismissed):
minor comments

azurelinuxagent/ga/thread_handler_base.py
nagworld9: done

azurelinuxagent/ga/thread_handler_base.py
nagworld9: The main logic in each thread is different, putting it in base class is non-trivial change. It required some thought and will defer to another pr

azurelinuxagent/ga/env.py
```diff
@@ -254,28 +253,17 @@ def get_thread_name():
...
         self.start()
```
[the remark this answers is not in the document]
nagworld9: won't fix: old comments

tests/ga/test_monitor.py
```diff
@@ -59,58 +59,66 @@ def _mock_wire_protocol():
...
-                                    SendHostPluginHeartbeat.__name__,
-                                    SendImdsHeartbeat.__name__,
-                                ]
+                            if network_changes:
+                                expected_operations.append(ReportNetworkConfigurationChanges.__name__)
 
-                                if network_changes:
-                                    expected_operations.append(ReportNetworkConfigurationChanges.__name__)
+                            if 'Linux' in platform.system():
+                                expected_operations.append(MonitorKernelSoftLockup.__name__)
```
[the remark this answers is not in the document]
nagworld9: won't fix: old comment

narrieta (approved):

maddieford (approved):

PR #3603, opened 2026-04-29
Implement fallback to WireServer when deploying SSH keys

narrieta:
As part of #3484, the waagent provision handler no longer pulls the entire goal state. As a side effect, when SSH keys are not available in ovfenv.xml, the fallback to the WireServer's Certificates package is no longer available. This PR adds that fallback.


azurelinuxagent/pa/provision/default.py
```diff
@@ -240,9 +241,35 @@ def config_user_account(self, ovfenv):
         logger.info("Configure sshd")
         self.osutil.conf_sshd(ovfenv.disable_ssh_password_auth)
 
+        self._download_ssh_keys(ovfenv)
         self.deploy_ssh_pubkeys(ovfenv)
         self.deploy_ssh_keypairs(ovfenv)
 
+    def _download_ssh_keys(self, ovfenv):
+        #
```
narrieta, line 249: For the reason of this check, see the implementation of  self.deploy_ssh_pubkeys() and self.deploy_ssh_keypairs(). In summary, the former will look for /var/lib/waagent/\*.crt when ovfenv.xml doesn't have a value for the key, and the latter will look for /var/lib/waagent/*.prv when ovfenv.xml contains any key pairs.

For the format of ovfenv.xml, see the unit tests.

azurelinuxagent/pa/provision/default.py
```diff
@@ -240,9 +241,35 @@ def config_user_account(self, ovfenv):
...
+        # We need to download the Certificates package from the Wireserver if any public key in ovfenv.xml has only a thumbprint (i.e. no value for the key) or if any key pairs need to be installed
+        #
+        download_certificates = any(value is None and thumbprint is not None for _, thumbprint, value in ovfenv.ssh_pubkeys) or len(ovfenv.ssh_keypairs) > 0
+        if not download_certificates:
+            return
+
+        #
```
narrieta, line 256: Originally, the PA would retrieve the goal state as part of protocol detection, which retries 360 times, with a 10 second period. If that fails, the process exits and is restarted by the system, which goes again into the 360 retries, etc.

I feel just a few tries are enough, let me know if you think otherwise; thanks.

tests/data/ovf-env-2.xml
```diff
@@ -16,12 +16,6 @@
               <Value>ssh-rsa AAAANOTAREALKEY== foo@bar.local</Value>
             </PublicKey>
           </PublicKeys>
-          <KeyPairs>
```
narrieta, line 19: The key pairs in the unit tests are not used, and with the new code they would trigger the Certificates download. To fix this, quite a few tests need to be rewritten to mock that download. Since the keys are not used by the tests, I decided to remove them instead.

azurelinuxagent/pa/provision/default.py
```diff
@@ -240,9 +241,35 @@ def config_user_account(self, ovfenv):
...
+        # Try to download the certificates a few times but continue execution if all attempts fail. The code that deploys the SSH keys will report that as a provisioning error.
+        #
+        max_attempts = 5
+        for attempt in range(max_attempts):
+            try:
+                protocol = self.protocol_util.get_protocol(init_goal_state=False)
+                _ = GoalState(protocol.client, goal_state_properties=GoalStateProperties.Certificates)
+                logger.info("Downloaded certificates successfully")
+                return
```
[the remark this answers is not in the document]
narrieta: removed log message; failure to download is handled later in the code

azurelinuxagent/pa/provision/default.py
```diff
@@ -240,9 +241,35 @@ def config_user_account(self, ovfenv):
...
+            except Exception as e:
+                if attempt < max_attempts - 1:
+                    logger.warn("Unable to download certificates; will retry after a short delay: {0}", ustr(e))
+                    time.sleep(30)
+                else:
```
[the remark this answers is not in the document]
narrieta: won't fix

tests/pa/test_provision.py
```diff
@@ -377,6 +380,34 @@ def test_get_provision_handler_config_cloudinit(
...
+            handler._download_ssh_keys(ovfenv)
+        self.assertEqual(0, protocol.mock_wire_data.call_counts['certificates'], "The Certificates package should not have been retrieved")
+
+    def test_it_should_download_certificates_when_the_public_key_does_not_have_a_value(self):
+        ovfenv = OvfEnv(load_data("ovf-env_public_key_no_value.xml"))
+        with TestProvision._create_provision_handler_with_mock_protocol() as (handler, protocol):
+            handler._download_ssh_keys(ovfenv)
+        self.assertEqual(1, protocol.mock_wire_data.call_counts['certificates'], "The Certificates package should have been retrieved")
+
+    def test_it_should_download_certificates_when_key_pairs_need_to_deployed(self):
```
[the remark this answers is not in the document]
narrieta: fixed

tests/pa/test_provision.py
```diff
@@ -377,6 +380,34 @@ def test_get_provision_handler_config_cloudinit(
...
+        ovfenv = OvfEnv(load_data("ovf-env_key_pair.xml"))
```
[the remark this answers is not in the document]
narrieta: fixed

tests/pa/test_provision.py
```diff
@@ -377,6 +380,34 @@ def test_get_provision_handler_config_cloudinit(
...
+        with TestProvision._create_provision_handler_with_mock_protocol() as (handler, protocol):
+            handler._download_ssh_keys(ovfenv)
+        self.assertEqual(1, protocol.mock_wire_data.call_counts['certificates'], "The Certificates package should have been retrieved")
+
+    def test_it_should_download_certificates_when_key_pairs_need_to_be_deployed(self):
+        ovfenv = OvfEnv(load_data("ovf-env_key_pair.xml"))
+        with TestProvision._create_provision_handler_with_mock_protocol() as (handler, protocol):
+            handler._download_ssh_keys(ovfenv)
+        self.assertEqual(1, protocol.mock_wire_data.call_counts['certificates'], "The Certificates package should have been retrieved")
+
```
nagworld9, line 465: cam we add test around exception when downloading certs?

narrieta: Those are covered in the unit tests for the goal state.

Are you thinking of a specific scenario?

narrieta: Or do you mean the retry logic in _download_ssh_keys?

azurelinuxagent/pa/provision/default.py
```diff
@@ -240,9 +241,34 @@ def config_user_account(self, ovfenv):
         logger.info("Configure sshd")
         self.osutil.conf_sshd(ovfenv.disable_ssh_password_auth)
 
+        self._download_ssh_keys(ovfenv)
         self.deploy_ssh_pubkeys(ovfenv)
         self.deploy_ssh_keypairs(ovfenv)
 
+    def _download_ssh_keys(self, ovfenv):
```
nagworld9: since its not download always, can we rename to _download_ssh_keys_if_needed or something like that?

tests/pa/test_provision.py
nagworld9: Retry logic in download_ssh_keys. We expect next operation would raise error if we failed to download with all retries. 

can we test that we don't re-raise error here? It would be good to test if next operation rising that provision error if not already present in the unit tests

tests/data/ovf-env_public_key.xml
```diff
@@ -0,0 +1,34 @@
...
+      <LinuxProvisioningConfigurationSet xmlns="http://schemas.microsoft.com/windowsazure" xmlns:i="http://www.w3.org/2001/XMLSchema-instance">
+        <ConfigurationSetType>LinuxProvisioningConfiguration</ConfigurationSetType>
+        <HostName>HostName</HostName>
+        <UserName>UserName</UserName>
+        <UserPassword>UserPassword</UserPassword>
+        <DisableSshPasswordAuthentication>false</DisableSshPasswordAuthentication>
+        <SSH>
+          <PublicKeys>
+            <PublicKey>
+              <Fingerprint>EB0C0AB4B2D5FC35F2F0658D19F44C8283E2DD62</Fingerprint>
```
maddieford, line 14: Is it expected that the customer should only provide one of fingerprint and value, or is it acceptable to provide both?

tests/pa/test_provision.py
narrieta: yes, i can add a test for the retry logic

tests/data/ovf-env_public_key.xml
narrieta: The original sample file from which I copied this one includes both. In practice, I believe only 1 would be populated.

Now, the logic in OSUtil.deploy_ssh_pubkey() prioritizes the value over the thumbprint, so in terms of the unit test that uses this new file, having both is a good test case.

tests/pa/test_provision.py
narrieta: tests added

azurelinuxagent/pa/provision/default.py
narrieta: I checked that cloud-init also retries for a long period of time, so I decided to use the same strategy as protocol detection

tests/pa/test_provision.py
```diff
@@ -377,6 +382,86 @@ def test_get_provision_handler_config_cloudinit(
...
+                if mock_http_get.call_count <= 10:
+                    return Exception("Mock failure")
+            return None
+        mock_http_get.call_count = 0
+
+        with TestProvision._create_provision_handler_with_mock_protocol() as (handler, protocol):
+            protocol.set_http_handlers(http_get_handler=mock_http_get)
+            with patch('azurelinuxagent.pa.provision.default.PROBE_INTERVAL', 0):  # set the delay between retries to 0
+                handler._download_ssh_keys_if_needed(ovfenv)
+        self.assertEqual(11, mock_http_get.call_count, "Expected 11 requests for Certificates (10 failed and retried requests, and 1 successful request")
```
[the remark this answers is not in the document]
narrieta: added )

tests/pa/test_provision.py
```diff
@@ -377,6 +382,86 @@ def test_get_provision_handler_config_cloudinit(
...
+                return Exception("Mock failure")
+            return None
+        mock_http_get.call_count = 0
+
+        with TestProvision._create_provision_handler_with_mock_protocol() as (handler, protocol):
+            protocol.set_http_handlers(http_get_handler=mock_http_get)
+            with patch('azurelinuxagent.pa.provision.default.PROBE_INTERVAL', 0):  # set the delay between retries to 0
+                handler._download_ssh_keys_if_needed(ovfenv)
+        self.assertEqual(2 * MAX_RETRY, mock_http_get.call_count, "Expected maximum number of retries ({0}) to have been attempted".format(MAX_RETRY))  # times 2 since two cyphers are attempted for FIPS support
+        ssh_key_path = os.path.join(conf.get_lib_dir(), '8979F1AC8C4215827BF3B5A403E6137B504D02A4.crt')
```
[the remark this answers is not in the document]
narrieta: won't fix, test takes .3 secs to execute and want to avoid excessive use of mocks

tests/pa/test_provision.py
```diff
@@ -377,6 +382,86 @@ def test_get_provision_handler_config_cloudinit(
...
+        self.assertFalse(os.path.exists(ssh_key_path), 'The SSH key should not have been downloaded, since all requests failed. Got: {0}'.format(ssh_key_path))
+
+    def test_deploy_ssh_pubkeys_should_raise_if_no_keys_have_been_downloaded(self):
+        ovfenv_data = load_data("ovf-env_public_key_no_value.xml")
+        ovfenv_data = ovfenv_data.replace('<UserName>UserName</UserName>', '<UserName>{0}</UserName>'.format(os.getlogin()))
+        ovfenv_data = ovfenv_data.replace('<Path>$HOME/UserName/.ssh/authorized_keys</Path>', '<Path>{0}</Path>'.format(os.path.join(self.tmp_dir, "authorized_keys")))
```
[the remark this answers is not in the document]
narrieta: fixed

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -613,7 +618,7 @@ def _try_download_certificates_pfx(self, wire_client, uri, pfx_file):
 
             return True
 
-        raise Exception("Cannot download certificates using any of the supported cyphers")
+        raise ProtocolError("Cannot download certificates using any of the supported cyphers")
```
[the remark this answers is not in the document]
narrieta: fixed

azurelinuxagent/common/protocol/goal_state.py
```diff
@@ -543,9 +545,12 @@ def __init__(self, wire_client, uri, logger_):
             self._convert_certificates_pfx_to_pem(pfx_file, pem_file)
         except Exception as e:
             # A failure to download the certificates won't necessarily produce an error. Certificates do not change often and they may have
-            # already been saved to disk on a previous goal state. We simply report the error and continue processing the goal_state; later on,
+            # already been saved to disk on a previous goal state. Re-raise the exception only if explicitly requested via the ignore_download_errors
+            # parameter, otherwise simply report the error and continue processing the goal_state; later on,
             # before extensions are processed, the Agent checks whether the required certificates are already on disk and refreshes the goal
-            # state if they are not.
+            # state if they are not
```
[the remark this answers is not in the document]
narrieta: won't fix; use of a boolean should be easy to understand by future maintaners

tests/pa/test_provision.py
```diff
@@ -377,6 +383,86 @@ def test_get_provision_handler_config_cloudinit(
...
+                mock_http_get.call_count += 1
+                return Exception("Mock failure")
+            return None
+        mock_http_get.call_count = 0
+
+        with TestProvision._create_provision_handler_with_mock_protocol() as (handler, protocol):
+            protocol.set_http_handlers(http_get_handler=mock_http_get)
+            with patch('azurelinuxagent.pa.provision.default.PROBE_INTERVAL', 0):  # set the delay between retries to 0
+                handler._download_ssh_keys_if_needed(ovfenv)
+        self.assertEqual(2 * MAX_RETRY, mock_http_get.call_count, "Expected maximum number of retries ({0}) to have been attempted".format(MAX_RETRY))  # times 2 since two cyphers are attempted for FIPS support
```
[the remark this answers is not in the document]
narrieta: fixed

azurelinuxagent/common/protocol/goal_state.py
maddieford, line 602: Why are download failures treated differently than missing data element?

Right now we continue in the case of missing data, but raise in the case of download/decryption errors (when ignore_download_error is False).

narrieta: They would be different: a download error vs a successful request with no data. The way the code looks, it seems that it is (or it was) possible to get a response without any data. I've never seen such a response.

maddieford (approved):

nagworld9 (approved):

PR #3609, opened 2026-05-22
Add expiry check for ext signature validation feature and log any error checking expiry only once per agent execution

maddieford:
## Description
The extension signature validation feature should be disabled after the expiry time has passed. This is to prevent any unexpected long-term consequences of sending telemetry for signature validation on customer VMs where a version of the agent with these changes are baked-in.

I updated the expiry check to only log once per agent execution, since this code is hit in multiple threads (ext handler, collect logs, etc.)  whenever forcing an update of the goal state or fetching a new goal state. 

Issue # 
---

### PR information

---


azurelinuxagent/ga/signature_validation_util.py
```diff
@@ -449,53 +463,63 @@ def _is_signature_validation_telemetry_expired():
         return datetime.datetime.now(UTC) >= expiry_date
     except Exception as ex:
         # Catch any exception (e.g. ValueError from a malformed date string) and treat the feature as expired so we fail safe.
-        logger.error("Failed to parse agent signature validation expiry time. Treating feature as expired. Error: {0}", ustr(ex))
+        # Only report the error once per agent execution to avoid flooding the log, since this function can be called
+        # multiple times for a single goal state.
+        if not _ExpiryCheckErrorReporter.already_reported():
```
maddieford, line 468: I added logic to only log this error once per agent execution. I also added logic to send an event for it. 

If you don't think the logging/event is necessary, I can remove it.  I'm debating whether it's necessary or not since it's a debug flag and customer's shouldn't be using it at all

azurelinuxagent/ga/signature_validation_util.py
```diff
@@ -449,53 +463,63 @@ def _is_signature_validation_telemetry_expired():
...
+            _ExpiryCheckErrorReporter.mark_reported()
+            msg = ("Error while checking signature validation expiry time "
+                   "(Debug.SignatureValidationTelemetryExpiryTime) from conf, will skip signature validation. "
+                   "Error: {0}").format(ustr(ex))
+            logger.warn(msg)
+            add_event(op=WALAEventOperation.SignatureValidation, is_success=False, message=msg, log_event=False)
         return True
 
 
 def ext_signature_validation_enabled():
```
maddieford, line 478: I reordered the checks in ext_signature_validation_enabled()/agent_signature_validation_enabled()/agent_signature_goal_state_telemetry_enabled() to reduce any logging from these checks in case this is not a CVM

azurelinuxagent/ga/signature_validation_util.py
```diff
@@ -449,53 +463,63 @@ def _is_signature_validation_telemetry_expired():
         return datetime.datetime.now(UTC) >= expiry_date
     except Exception as ex:
         # Catch any exception (e.g. ValueError from a malformed date string) and treat the feature as expired so we fail safe.
-        logger.error("Failed to parse agent signature validation expiry time. Treating feature as expired. Error: {0}", ustr(ex))
+        # Only report the error once per agent execution to avoid flooding the log, since this function can be called
+        # multiple times for a single goal state.
+        if not _ExpiryCheckErrorReporter.already_reported():
+            _ExpiryCheckErrorReporter.mark_reported()
+            msg = ("Error while checking signature validation expiry time "
```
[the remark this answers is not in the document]
maddieford: This seems low risk. Worst case scenario, multiple threads send this log/telemetry instead of only one. 

I can add if anyone thinks it is necessary

azurelinuxagent/ga/signature_validation_util.py
narrieta: But any logic on ConfidentialVMInfo.is_confidential_vm()  should be executed only after checking that the feature (signature validation) is enabled.

narrieta: Similar comment at lines 532/533

azurelinuxagent/common/conf.py
```diff
@@ -174,7 +174,12 @@ def load_conf_from_file(conf_file_path, conf=__conf__):
     "ResourceDisk.Filesystem": "ext3",
     "AutoUpdate.GAFamily": "Prod",
     "Policy.PolicyFilePath": "/etc/waagent_policy.json",
-    "Protocol.EndpointDiscovery": "dhcp"
+    "Protocol.EndpointDiscovery": "dhcp",
+    #
+    # "Debug" options are experimental and may be removed in later
+    # versions of the Agent.
+    #
+    "Debug.SignatureValidationTelemetryExpiryTime": "2026-12-01"
```
nagworld9, line 182: PR title says for ext but code changes for both agent and extension. Can you update accordingly?

maddieford: My last PR already added the check for agent signature validation (see agent_signature_validation_enabled() in signature_validation_util), but this PR adds the check to the extension signature validation enablement conditions (def ext_signature_validation_enabled())

azurelinuxagent/ga/signature_validation_util.py
maddieford: Right now the conf flag check is first. What set of conditions indicate signature validation is enabled? Do you mean that the is_confidential_vm() check should come last?

_is_signature_validation_telemetry_expired() and openssl_version_supported_for_signature_validation() have the potential to log on error, so I was trying to prevent that logging if the feature is not even eligible for this VM (i.e. VM is a CVM). 

But if we want to only hit the is_confidential_vm() logic after we verify every other requirement is met first, I can update the ordering

azurelinuxagent/common/conf.py
maddieford: I updated the title with the logging change which affects agent and extension path

azurelinuxagent/ga/signature_validation_util.py
narrieta: My bad. I meant the check for the config flag should come first, but it is already the first check, so all good.

narrieta (approved):

nagworld9 (approved):

PR #3611, opened 2026-05-28
fix: replace deprecated logger.warn() with logger.warning()

MD-Mushfiqur123:
Replaces all instances of deprecated `logger.warn()` with `logger.warning()` to resolve Python 3.13+ deprecation warnings.


maddieford:
The agent doesn't use the python logger, it uses a custom logger

maddieford:
Closed the pull request since the agent uses a custom logger, not the python logger

PR #3622, opened 2026-06-10
Ext signature should be validated when required by policy

maddieford:
## Description
This PR fixes an issue in the extension signature policy implementation. Previously, signature validation would be silently skipped despite being required by customer policy in the following cases:
- Debug.EnableExtSignatureValidation=n
- A previous validation attempt timed out
- Within the first 10 minutes after service start
- The feature expiry time has passed (Debug.SignatureValidationTelemetryExpiryTime)
- openssl version does not support validation 

If the customer requires validation, these conditions should not prevent validation. These conditions were added in attempt to prevent any performance impacts, but customers who explicitly require signature validation via policy accept performance cost. 

If the openssl version is not supported, that error will be reported at the time policy is parsed to prevent any extensions from being processed until the customer fixes this dependency issue.

Issue # 
---

### PR information

---


tests/ga/test_extension.py
```diff
@@ -4589,23 +4589,27 @@ def test_enable_should_fail_if_extension_unsigned(self):
                                     expected_status_code=ExtensionErrorCodes.PluginEnableProcessingFailed,
                                     expected_handler_status='NotReady', expected_ext_count=1)
 
-    def test_enable_should_succeed_for_extension_with_invalid_signature_if_conf_flag_disabled(self):
-        # If 'Debug.EnableExtSignatureValidation' flag is set to false, enable should succeed for an extension with
-        # invalid signature, even with enforcement enabled.
+    def test_enable_should_fail_for_extension_with_invalid_signature_if_validation_disabled(self):
```
maddieford, line 4592: Interestingly, we had a unit test checking that signature validation is skipped (even when enforced) when the debug flag (Debug.EnableExtSignatureValidation) is false. This seems incorrect, so I updated this UT to test the opposite: Signature validation is not skipped when enforced

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -298,8 +299,17 @@ def _check_attributes(object_, object_name, valid_attributes):
             if k not in valid_attributes:
                 raise InvalidPolicyError("unrecognized attribute '{0}' in {1}".format(k, object_name))
 
-        if object_.get("signatureRequired") is True and not ConfidentialVMInfo.is_confidential_vm():
-            raise InvalidPolicyError("setting 'signatureRequired' to true is only supported on confidential virtual machines (CVMs).")
+        if object_.get("signatureRequired") is True:
```
narrieta, line 310: I'd suggest moving this into a helper function and calling it from _parse_extension and _parse_extension_policies. That way the syntactic and semantic rules for each object (individual extension and global extension policies) are more explicit and this check does not apply to any other object.

If you prefer to keep it here, let's update the description of the function.

narrieta (dismissed):
Minor comment

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1470,7 +1475,20 @@ def download(self, ignore_signature_validation_errors):
...
+        # If validation is required (e.g. ignore_signature_validation_errors is False) but signature wasn't provided,
+        # fail immediately rather than silently downloading an unsigned package.
+        if not ignore_signature_validation_errors and self.ext_handler.encoded_signature == "":
+            raise SignatureValidationError(
+                msg="Signature validation is required for extension '{0}', but no signature was provided in the goal state.".format(self.get_full_name()),
+                operation=WALAEventOperation.SignatureValidation, duration=0)
+
+        # Validate signature if the extension is signed and the feature is enabled, OR if the caller requires
+        # validation results (ignore_signature_validation_errors=False). 
+        if not ignore_signature_validation_errors:
```
nagworld9, line 1487: First time checking the signature code in great detail, ignore_signature_validation_errors tells us extension is using signature and agent should validate?

The name ignore_signature_validation_errors sounds different

azurelinuxagent/ga/exthandlers.py
```diff
@@ -1470,7 +1475,20 @@ def download(self, ignore_signature_validation_errors):
...
+            should_validate_ext_signature = True
+        else:
+            should_validate_ext_signature = ext_signature_validation_enabled() and self.ext_handler.encoded_signature != ""
```
nagworld9, line 1490: How does agent know feature is not supported yet vs feature is disabled due to performance reasons.? if its former, we don't want to continue signature validation even customer policy is place right?

azurelinuxagent/ga/policy/policy_engine.py
maddieford: Moved to helper

azurelinuxagent/ga/exthandlers.py
maddieford: ignore_signature_validation_errors is False when the config is False (Debug.IgnoreExtSignatureValidationErrors=n) OR the customer explicitly requires the extension to be signed via their policy (/etc/waagent_policy.json). In both of these cases, the agent should not download the extension without validating its signature.

When ignore_signature_validation_errors is False, two things happen:
 1. should_validate_ext_signature is set to True unconditionally. The agent should attempt to validate the ext signature before unzipping (i.e., do not silently skip signature validation for any reason)
 2. Any signature validation failure blocks the download.

When it's True, validation may still run (if the feature flag is enabled, performance workaround aren't blocking, and a signature is present), but failures are just logged as telemetry and download proceeds.

So not ignore_signature_validation_errors effectively means "signature validation is required" — it forces validation on and makes errors fatal.

azurelinuxagent/ga/exthandlers.py
maddieford: If the customer has a policy which requires the extension signature validation is required, then we do not want to skip signature validation for any reason. By creating a policy file which requires signature validation, they accept the performance cost

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -298,8 +301,23 @@ def _check_attributes(object_, object_name, valid_attributes):
...
+            # Signature validation is currently only supported on CVMs. If a non-CVM user creates a policy with
+            # signatureRequired=true, reject the policy at parse time.
+            # TODO: Remove once signature validation is supported on all VMs.
+            if not ConfidentialVMInfo.is_confidential_vm():
+                raise InvalidPolicyError("setting 'signatureRequired' to true is only supported on confidential virtual machines (CVMs).")
+            # Signature validation requires OpenSSL >= 1.1.0. If the system OpenSSL is too old, reject the policy at
+            # parse time.
+            # TODO: Remove once signature validation no longer depends on the 'no_check_time' flag.
+            if not openssl_version_supported_for_signature_validation():
+                raise InvalidPolicyError("setting 'signatureRequired' to true requires OpenSSL >= 1.1.0; the OpenSSL version on this system does not support signature validation.")
```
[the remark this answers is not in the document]
maddieford: Won't fix. The current error message is straight forward in my opinion

azurelinuxagent/ga/exthandlers.py
nagworld9: what about scenario of feature is not ready and not enabled from the begging in particular agent version but the customer explicitly set the extension to be signed via their policy, do we still validate the extension signature?

narrieta (approved):

narrieta (approved):

azurelinuxagent/ga/exthandlers.py
nagworld9: discussed offline

nagworld9 (approved):

PR #3625, opened 2026-06-11
Ext upgrade should fail when old handler is disallowed by policy

maddieford:
## Description
During upgrade, if the old extension handler is disallowed by policy (i.e. it is not an allowlisted extension, or it's signature wasn't validated when required) then we should fail the upgrade. We should not perform any operations (disable/update/unisntall) on the old handler if it is not allowed by the customer's policy.

This PR also makes a change to distinguish between the following two extension failures:
1. Extension is unsigned, but policy requires signature
2. Extension is already installed and it's signature wasn't validated when it was downloaded, but policy requires signature

Issue # 
---

### PR information

---


azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -358,7 +375,7 @@ def _should_allow_extension(self, extension_name):
         Return whether we should allow extension download based on policy.
         If policy feature not enabled, return True.
         If allowListedExtensionsOnly=true, return true only if extension present in "extensions" allowlist.
-        If allowListedExtensions=false, return true always.
+        If allowListedExtensionsOnly=false, return true always.
```
maddieford, line 398: fixing typo here

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -61,13 +61,30 @@ def __init__(self, msg, inner=None):
...
+    Base error raised when policy requires signature, but the extension is not signed or was not previously validated.
+    Callers should raise one of the more specific subclasses (ExtensionUnsignedError or ExtensionSignatureNotValidatedError)
+    so that the appropriate message can be reported.
     This error does not accept a message.
     """
     def __init__(self):
         super(ExtensionSignaturePolicyError, self).__init__()
 
 
+class ExtensionUnsignedError(ExtensionSignaturePolicyError):
```
maddieford, line 76: these new exceptions allow us to improve the error reporting in extension status

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -399,10 +418,8 @@ def check_extension_policy(self, extension_name, extension_is_signed):
         if not extension_allowed:
             raise ExtensionDisallowedError()      # Caller sets message and error code, based on requested extension operation
 
-        # TODO: Differentiate error handling:
```
maddieford, line 420: The TO-DO suggested raising scenario specific exceptions here, but this class has no knowledge of whether this is fresh install vs enable etc.

Instead I raise the generic exception here, and the caller is responsible for handling it by raising the specific exception for the error

azurelinuxagent/ga/exthandlers.py
```diff
@@ -825,13 +834,28 @@ def handle_enable(self, ext_handler_i, extension):
 
             # Check extension policy and raise error if disallowed (e.g., not in allowlist, unsigned when policy requires signature).
             extension_is_signed = ext_handler_i.ext_handler.encoded_signature != ""
-            self._policy_engine.check_extension_policy(ext_handler_i.ext_handler.name, extension_is_signed)
+            try:
+                self._policy_engine.check_extension_policy(ext_handler_i.ext_handler.name, extension_is_signed)
+            except ExtensionSignaturePolicyError:
```
maddieford, line 839: I wrapped the existing policy checks so that we can raise the appropriate exception in the case of extension signature errors

azurelinuxagent/ga/exthandlers.py
```diff
@@ -825,13 +834,28 @@ def handle_enable(self, ext_handler_i, extension):
...
             if old_ext_handler_i is None:
                 ext_handler_i.install(extension=extension)
             elif ext_handler_i.version_ne(old_ext_handler_i):
+                # Before running any commands on the old handler (disable/update/uninstall), check that the old
+                # handler is allowed by the current policy. The old handler may violate the current policy (e.g.,
+                # it is no longer in the allowlist, or its signature was never validated by the agent). Treat the
+                # old handler as signed if its signature was previously validated on download.
+                old_extension_is_signed = old_ext_handler_i.signature_validated
+                try:
+                    self._policy_engine.check_extension_policy(old_ext_handler_i.ext_handler.name, old_extension_is_signed)
```
maddieford, line 855: We add a policy check here before we do any operations on the old ext handler during upgrade scenario. 

@narrieta We talked about doing this check in ExtHandlersHandler._update_extension_handler_and_return_if_failed conditionally. However, I noticed that we always call uninstall operation on old handler (see line 973), so I decided to do the check here unconditionally. Please give that a review and let me know what you think
`uninstall_exit_code = execute_old_handler_command_and_return_if_succeeds(
            func=partial(old_ext_handler_i.uninstall, extension=extension))`

tests/ga/test_extension.py
```diff
@@ -3791,6 +3791,114 @@ def test_uninstall_should_succeed_for_extension_that_was_not_installed(self):
             vm_status = args[0]
             self.assertEqual(0, len(vm_status.vmAgent.extensionHandlers))
 
+    def test_upgrade_should_succeed_if_old_and_new_extension_versions_are_allowed_by_policy(self):
```
maddieford, line 3794: I added two policy unit tests for upgrade to test the change

azurelinuxagent/ga/policy/policy_engine.py
```diff
@@ -61,13 +61,30 @@ def __init__(self, msg, inner=None):
 
 class ExtensionSignaturePolicyError(PolicyError):
     """
-    Error raised when policy requires signature, but extension is not signed and was not previously validated.
+    Base error raised when policy requires signature, but the extension is not signed or was not previously validated.
+    Callers should raise one of the more specific subclasses (ExtensionUnsignedError or ExtensionSignatureNotValidatedError)
+    so that the appropriate message can be reported.
```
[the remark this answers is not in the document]
narrieta: +1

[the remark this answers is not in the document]
narrieta: maybe the confusion is on who exactly is the "caller" mentioned in this comment. this comment seems to contradict the new comment at line 421

azurelinuxagent/ga/exthandlers.py
```diff
@@ -825,13 +834,28 @@ def handle_enable(self, ext_handler_i, extension):
 
             # Check extension policy and raise error if disallowed (e.g., not in allowlist, unsigned when policy requires signature).
             extension_is_signed = ext_handler_i.ext_handler.encoded_signature != ""
-            self._policy_engine.check_extension_policy(ext_handler_i.ext_handler.name, extension_is_signed)
+            try:
+                self._policy_engine.check_extension_policy(ext_handler_i.ext_handler.name, extension_is_signed)
+            except ExtensionSignaturePolicyError:
+                # New installation path: the downloaded package itself is unsigned.
```
narrieta: I don't quite get why we raise unconditionally, because extension_is_signed could be False.

Likely that is handled in self._policy_engine.check_extension_policy, but that is not clear at all from this code. If this is the case, and refactoring is too expensive, let's add some comments here.

narrieta (commented):
Implementation LGTM, but use and documentation of the exceptions is a little confusing. Added some comments.

azurelinuxagent/ga/policy/policy_engine.py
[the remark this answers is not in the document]
maddieford: I updated the docstring for the exception, hopefully its clearer now

azurelinuxagent/ga/exthandlers.py
```diff
@@ -722,14 +722,23 @@ def handle_ext_handler(self, ext_handler_i, extension, goal_state_id):
...
             ).format(operation, ext_handler_i.ext_handler.name, conf.get_policy_file_path())
             self.__handle_ext_disallowed_error(ext_handler_i, error_code, report_op=WALAEventOperation.ExtensionSignaturePolicy, message=msg,
                                                extension=extension)
+        except ExtensionSignatureNotValidatedError:
+            operation, error_code = _EXT_DISALLOWED_ERROR_MAP.get(ext_handler_i.ext_handler.state)
+            msg = (
+                "Extension will not be processed: failed to {0} extension '{1}' because policy specifies that extension must be signed, "
+                "but the installed extension's signature was not validated by the agent. To {0}, remove and reinstall the extension so that "
+                "its signature can be validated, or set 'signatureRequired' to false in the policy file ('{2}')."
+            ).format(operation, ext_handler_i.ext_handler.name, conf.get_policy_file_path())
```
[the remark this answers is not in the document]
maddieford: Updated

azurelinuxagent/ga/exthandlers.py
maddieford: the policy engine checks if the policy requires signature and only raises if it does and extension is not signed. 
I updated the comments around all the callers

tests_e2e/tests/ext_signature_validation/ext_signature_validation.py
```diff
@@ -111,7 +111,13 @@ def _should_fail_to_enable_extension(extension_case):
         except Exception as error:
             # We exclude the extension name from regex because CRP sometimes installs test extensions with different
             # names (ex: Microsoft.Azure.Extensions.Edp.RunCommandHandlerLinuxTest instead of Microsoft.CPlat.Core.RunCommandHandlerLinux)
-            pattern = r".*Extension will not be processed: failed to run extension .* because policy specifies that extension must be signed, but extension package signature could not be found.*"
+            # The agent reports one of two messages depending on whether the extension is being newly installed
+            # ("extension package signature could not be found") or was previously installed without signature
+            # validation ("the installed extension's signature was not validated by the agent").
```
maddieford:
```suggestion
            # validation ("the installed extension's signature was not previously validated by the agent").
```

tests_e2e/tests/ext_signature_validation/ext_signature_validation.py
```diff
@@ -111,7 +111,13 @@ def _should_fail_to_enable_extension(extension_case):
...
+            pattern = (
+                r".*Extension will not be processed: failed to run extension .* because policy specifies that extension must be signed, "
+                r"but (extension package signature could not be found|the installed extension's signature was not validated by the agent).*"
```
maddieford:
```suggestion
                r"but (extension package signature could not be found|the installed extension's signature was not previously validated by the agent).*"
```

azurelinuxagent/ga/exthandlers.py
narrieta: ok, makes sense

azurelinuxagent/ga/exthandlers.py
```diff
@@ -825,13 +834,31 @@ def handle_enable(self, ext_handler_i, extension):
...
 
             self.__setup_new_handler(ext_handler_i, extension, self.__should_ignore_ext_signature_validation_errors(ext_handler_i))
 
             if old_ext_handler_i is None:
                 ext_handler_i.install(extension=extension)
             elif ext_handler_i.version_ne(old_ext_handler_i):
+                # Before running any commands on the old handler (disable/update/uninstall), check that the old
+                # handler is allowed by the current policy. The old handler may violate the current policy (e.g.,
+                # it is no longer in the allowlist, or its signature was never validated by the agent). Treat the
+                # old handler as signed if its signature was previously validated on download.
```
nagworld9, line 852: Do we track if extension is previously validated and use that here? If previously validated, will we continue with upgrade even old handler violate the current policy?

maddieford: When we validate the signature of an extension, we write a signatureValidated field to the HandlerStatus:
{
  "name": "Microsoft.Azure.Extensions.CustomScript",
  "version": "2.0.7",
  "status": "NotReady",
  "code": 0,
  "message": "Plugin disabled",
  "supports_multi_config": false,
  "extension_status": null,
  **"signature_validated": false**
}

If the policy now requires signature validation, but the old handler's signature was not validated successfully (i.e. signature_validation: false), then we do not perform any operations on the old handler. 

This is true for re-enable/upgrade/uninstall

narrieta (approved):

nagworld9 (approved):

PR #3653, opened 2026-07-15
Add soft lockup trace capture to detection logic

adityagarg0911:
## Description
Add capturing of soft lockup logs/traces in the soft lockup detection logic to enable RCA of different soft lockup issues across the fleet.

Changes:
- Capture a bounded kernel stack trace / dmesg context (50 lines of dmesg ahead of soft lockup log) for soft lockup occurrences (preferring distinct CPUs) so the stuck code path is available for root-cause analysis instead of only aggregate counts.
- Reduce cpuDetails, increase the reporting period, and bound the max stack traces and lines per stack trace so the telemetry event does not exceed the payload limit. Per my calculation the payload stays within ~30 KB, conservative against the 64 KB limit.
- Capture every context line that shares a kernel timestamp: the kernel emits multiple dmesg lines within the same microsecond (register dump, call trace, etc.), so all lines sharing a timestamp within a scan are captured to preserve the complete stack-trace context.
- Add 5 unit tests covering bounded capture, shared-timestamp capture, repeated same-CPU capture, distinct-CPU preference, and separate stack-trace reporting.

Issue # 
---

### PR information

---


azurelinuxagent/ga/kernel_event_monitor.py
```diff
@@ -193,10 +241,11 @@ def _parse_and_aggregate_soft_lockup_events(self, line):
...
-        if kernel_timestamp <= self._last_processed_timestamp:
+        if kernel_timestamp <= self._parse_baseline_timestamp:
             return
 
-        self._last_processed_timestamp = kernel_timestamp
+        self._last_processed_timestamp = max(self._last_processed_timestamp, kernel_timestamp)
+        self._append_line_to_stack_traces(line)
         lockup_match = self._SOFT_LOCKUP_PATTERN.search(line)
         if lockup_match is None:
             return
```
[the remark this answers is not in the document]
adityagarg0911: This is expected behaviour

azurelinuxagent/ga/kernel_event_monitor.py
```diff
@@ -87,9 +87,14 @@ class MonitorKernelSoftLockup(PeriodicOperation):
...
 
     # Maximum number of affected CPUs to include in telemetry details; if exceeded, details will be truncated with a flag.
-    _MAX_CPU_DETAILS = 500
+    _MAX_CPU_DETAILS = 50
 
-    _PERIOD_SECONDS = 21600  # 6 hours
+    # Attach soft lockup logs for a limited number of occurrences, preferring distinct CPUs
+    # but falling back to repeat CPUs when fewer are available.
+    _MAX_CONTEXT_LINES_PER_TRACE = 50
+    _MAX_STACK_TRACES = 4
```
nagworld9, line 95: I'm worried about too much payload in memory and sending it in telemetry. Does this mean you send 200 log lines in single telemetry event in worst case? How much payload size you observed in total?

what is the different between max_stack vs max_cpu. Does cpu details also adds more data to send in telemetry?

adityagarg0911: @nagworld9 
Yes, the most we ever send is 200 lines = 4 stack traces x 50 lines each. 
It never grows past that no matter how many lockups happen.

I measured the actual telemetry event size against the 64 KB limit:
Real event, 14 CPUs affected (4 traces + 14 cpuDetails): ~20 KB (31%)
worst case (4 full traces + 50 cpuDetails): ~23 KB (35%) —> so about 40 KB of headroom to spare.
Even in a far-fetched case, I think it should not cross 30-35KB

The two limits control different things:

_MAX_STACK_TRACES = how many stack traces we attach. 
_MAX_CPU_DETAILS = how many CPUs we list in the cpuDetails summary. Each entry is just 4 small numbers (cpuId, count, maxStuckTimeSec, lastKernelTimestamp)

azurelinuxagent/ga/kernel_event_monitor.py
```diff
@@ -193,10 +241,11 @@ def _parse_and_aggregate_soft_lockup_events(self, line):
             return
 
         kernel_timestamp = float(timestamp_match.group('timestamp'))
-        if kernel_timestamp <= self._last_processed_timestamp:
+        if kernel_timestamp <= self._parse_baseline_timestamp:
             return
 
-        self._last_processed_timestamp = kernel_timestamp
+        self._last_processed_timestamp = max(self._last_processed_timestamp, kernel_timestamp)
+        self._append_line_to_stack_traces(line)
```
nagworld9, line 265: is this expected behavior? Each line adds to each trace, even though line is not related to trace CPU ID

azurelinuxagent/ga/kernel_event_monitor.py
maddieford: There is no limit on the length of each line though. very long lines can increase the size of this event. We may want to add a limit there as well

adityagarg0911: @maddieford, agreed
instead of a per-line character limit, I'm thinking of a combined cap on the stack traces of 200 lines and 32 KB. I'll check its feasibility and push the changes accordingly.

azurelinuxagent/ga/kernel_event_monitor.py
```diff
@@ -183,6 +190,47 @@ def _read_and_parse_dmesg(self):
         except Exception as e:
             logger.warn("KernelSoftLockup: Failed to read dmesg output: {0}".format(ustr(e)))
 
+    def _append_line_to_stack_traces(self, line):
+        # Grow each still-open trace with the next dmesg line until it reaches its line limit.
+        for trace in self._stack_traces:
+            if trace["remaining_after_lines"] > 0:
+                trace["logLines"].append(line)
+                trace["remaining_after_lines"] -= 1
```
maddieford: Instead of keeping this counter, could we just check the length of trace['logLines']? And then we can remove _get_kernel_stack_traces()

azurelinuxagent/ga/kernel_event_monitor.py
```diff
@@ -256,15 +307,20 @@ def _report_events(self):
...
         """
         Main operation: parse dmesg, aggregate new soft lockup events, and report.
         Exceptions are caught and logged by the base class (PeriodicOperation.run).
         """
-        previous_timestamp = self._last_processed_timestamp
+        # Use the previous scan's watermark as a fixed baseline for this scan: each dmesg line is compared
+        # against this baseline rather than the watermark that advances as lines are processed. This ensures
+        # consecutive lines sharing the same kernel timestamp are all captured (not just the first), while the
+        # watermark still advances during the scan and is persisted below to skip these lines on the next scan.
+        self._parse_baseline_timestamp = self._last_processed_timestamp
```
maddieford: Could we update the naming of `_parse_baseline_timestamp` to something like `_curr_scan_baseline_timestamp`?

azurelinuxagent/ga/kernel_event_monitor.py
```diff
@@ -193,10 +241,11 @@ def _parse_and_aggregate_soft_lockup_events(self, line):
             return
 
         kernel_timestamp = float(timestamp_match.group('timestamp'))
-        if kernel_timestamp <= self._last_processed_timestamp:
+        if kernel_timestamp <= self._parse_baseline_timestamp:
             return
 
-        self._last_processed_timestamp = kernel_timestamp
+        self._last_processed_timestamp = max(self._last_processed_timestamp, kernel_timestamp)
```
maddieford: why is max() needed here? If needed, please add a comment with the reasoning

azurelinuxagent/ga/kernel_event_monitor.py
maddieford: nit: we already set the kernel timestamp at line 266, so it's probably not needed here

azurelinuxagent/ga/kernel_event_monitor.py
```diff
@@ -183,6 +190,47 @@ def _read_and_parse_dmesg(self):
...
+        seen_cpu_ids = set()
+        for index, trace in enumerate(self._stack_traces):
+            if trace["cpuId"] in seen_cpu_ids:
+                return index
+            seen_cpu_ids.add(trace["cpuId"])
+        return None
+
+    def _start_log_context_capture(self, cpu_id, line):
+        # Keep up to _MAX_STACK_TRACES traces, preferring distinct CPUs. When full, a new distinct
+        # CPU evicts a repeat-CPU trace (if one exists); otherwise the new occurrence is dropped.
```
maddieford: can you specify in the comment which repeat CPU trace should be evicted (i.e., the first occurrence or the last occurrence)?

azurelinuxagent/ga/kernel_event_monitor.py
```diff
@@ -183,6 +190,47 @@ def _read_and_parse_dmesg(self):
...
 
+    def _append_line_to_stack_traces(self, line):
+        # Grow each still-open trace with the next dmesg line until it reaches its line limit.
+        for trace in self._stack_traces:
+            if trace["remaining_after_lines"] > 0:
+                trace["logLines"].append(line)
+                trace["remaining_after_lines"] -= 1
+
+    def _index_of_first_repeat_cpu_trace(self):
+        # Index of the first captured trace whose CPU already appears earlier in the list,
```
maddieford: Let's say there's 4 lockups on CPU0, this would return the index of the second lockup and the stack trace for that one would be dropped. 

Is that intentional? or should either the first or last occurrence be dropped?

azurelinuxagent/ga/kernel_event_monitor.py
```diff
@@ -183,6 +190,47 @@ def _read_and_parse_dmesg(self):
...
+        # Keep up to _MAX_STACK_TRACES traces, preferring distinct CPUs. When full, a new distinct
+        # CPU evicts a repeat-CPU trace (if one exists); otherwise the new occurrence is dropped.
+        if len(self._stack_traces) >= self._MAX_STACK_TRACES:
+            captured_cpu_ids = set([trace["cpuId"] for trace in self._stack_traces])
+            if cpu_id in captured_cpu_ids:
+                return
+            repeat_index = self._index_of_first_repeat_cpu_trace()
+            if repeat_index is None:
+                return
+            del self._stack_traces[repeat_index]
```
maddieford: nit: let's use pop() here instead of the del keyword

azurelinuxagent/ga/kernel_event_monitor.py
adityagarg0911: Agree, it was redundant. Will remove.

azurelinuxagent/ga/kernel_event_monitor.py
adityagarg0911: There is no specific intent here for eviction.
Ack, I'll switch to evicting last one

azurelinuxagent/ga/kernel_event_monitor.py
adityagarg0911: Yes, I could have removed it. 
But I’m now also using it to mark a stack trace complete when the shared byte limit (size) is reached, preventing later lines from being appended and creating gaps.

azurelinuxagent/ga/kernel_event_monitor.py
```diff
@@ -183,6 +195,60 @@ def _read_and_parse_dmesg(self):
         except Exception as e:
             logger.warn("KernelSoftLockup: Failed to read dmesg output: {0}".format(ustr(e)))
 
+    def _append_line_to_stack_traces(self, line):
+        for trace in self._stack_traces:
+            if trace["remaining_after_lines"] <= 0:
```
maddieford: nit: The dual use of remaining_after_lines (line counter + closed flag) is a bit hard to follow. I would consider simplifying. For example, you could check the byte budget once at the top of append_line_to_stack_traces and return early for all traces, then use len(logLines)  for the per-trace line limit. Another option would be to keep a 'closed' key with bool value to indicate whether the trace can accept more lines, and use len(logLines) for the per-trace line limit to make the intent of the key/value pair clear

Not blocking, but could improve readability

tests/ga/test_kernel_event_monitor.py
```diff
@@ -120,6 +124,126 @@ def test_parse_should_advance_watermark_even_without_lockups(self):
...
+
+        self._feed_dmesg(monitor, "\n".join(dmesg_lines))
+
+        captured_lines = monitor._get_kernel_stack_traces()[0]["logLines"]
+        self.assertLess(len(captured_lines), len(dmesg_lines),
+                 "The byte budget should have truncated the trace for this test to be meaningful")
+        self.assertEqual(captured_lines, dmesg_lines[:len(captured_lines)],
+                 "A size-capped trace should be truncated, never missing lines from the middle")
+
+    def test_parse_should_not_start_a_trace_without_size_budget(self):
```
maddieford: nit: could we split this into two different tests:
1. new lockup should not start a trace if budget will be exceeded
2. existing lockup should not be evicted if the new trace is too big

tests/ga/test_kernel_event_monitor.py
maddieford, line 318: Could we update this unit test to check that stack_traces and stack_trace_bytes are also cleared on failure

maddieford (approved):

nagworld9 (approved):

PR #3671, opened 2026-09-03
Add nftables fallback for missing iptables modules

maddieford:
## Description
Latest EL10 distros (RHEL 10.2 for example) are not packaging the 'kernel-modules-extra' modules which are required to setup firewall rules with the iptables cmd. Since the iptables command, the agent does not fallback to nftables to create the firewall rules. As a result, the firewall rules are not created on those images. The commands to create the firewalld passthrough rules do not fail, but when firewalld eventually loads the rules it will enter a failed state because the required kernel modules (xt_owner, xt_conntrack) to apply the rules aren't packaged.

This PR updates the agent to fallback to nftables if the required kernel modules aren't available (even if iptables cmd is). It also adds cleanup to remove the remaining iptables DNS rule when this fallback happens, and also to cleanup the firewalld passthrough rules when firewalld is running and the firewall manager uses nftables. 

Issue #3510 
---

### PR information

---

### DIFF
--- a/azurelinuxagent/agent.py
+++ b/azurelinuxagent/agent.py
@@ -312,7 +312,7 @@ class Agent(object):
             if isinstance(firewall_manager, IpTables):
                 try:
                     run_command(['systemctl', 'is-enabled', '--type=service', 'firewalld.service']).rstrip()
-                    logger.info("Firewalld is enabled. Will not setup the iptables firewall rules to avoid conflicts.")
+                    logger.info("Firewalld is enabled. Will not setup the firewall rules to avoid conflicts.")
                     sys.exit(0)
                 except CommandError:
                     # Differences across versions of systemd make hard to determine whether the command failed because firewalld is not installed
--- a/azurelinuxagent/agent.py
+++ b/azurelinuxagent/agent.py
@@ -307,18 +307,18 @@ class Agent(object):
         logger.info("Setting up firewall during boot. Endpoint: {0}", ustr(endpoint))
 
         try:
-            run_command(['systemctl', 'is-enabled', '--type=service', 'firewalld.service']).rstrip()
-            logger.info("Firewalld is enabled. Will not setup the firewall rules to avoid conflicts.")
-            sys.exit(0)
-        except CommandError:
-            # Differences across versions of systemd make hard to determine whether the command failed because firewalld is not installed
-            # or for another reason. Assume it is not installed and continue.
-            pass
-
-        try:
             firewall_manager = FirewallManager.create(endpoint)
             firewall_manager.verbose = True
             if isinstance(firewall_manager, IpTables):
+                try:
+                    run_command(['systemctl', 'is-enabled', '--type=service', 'firewalld.service']).rstrip()
+                    logger.info("Firewalld is enabled. Will not setup the iptables firewall rules to avoid conflicts.")
+                    sys.exit(0)
+                except CommandError:
+                    # Differences across versions of systemd make hard to determine whether the command failed because firewalld is not installed
+                    # or for another reason. Assume it is not installed and continue.
+                    pass
+
                 try:
                     #
                     # We execute "iptables -C -m conntrack" to force loading of the conntrack module with the intention of avoiding the
--- a/azurelinuxagent/ga/firewall_manager.py
+++ b/azurelinuxagent/ga/firewall_manager.py
@@ -86,21 +86,46 @@ class FirewallManager(object):
     @staticmethod
     def create(wire_server_address):
         """
-        Creates the appropriate FirewallManager implementation depending on the availability of the underlying command-line tools.
+        Creates the appropriate FirewallManager implementation depending on the availability of the underlying
+        command-line tools and the kernel modules required by iptables.
 
-        NOTE: Currently this method checks only for iptables and nftables, giving precedence to the former.
+        NOTE: Currently this method checks only for iptables and nft, giving precedence to the former. There are
+        some images (Rhel 10.2/Alma 10/Rocky 10), where iptables exists but the required modules (xt_owner and
+        xt_conntrack) to create the agent rules don't exist. This method now handles that case and cleans up the
+        remaining IpTables rules when we fall back to nft on those machines.
         """
         try:
-            manager = IpTables(wire_server_address)
-            event.info(WALAEventOperation.Firewall, "Using iptables [version {0}] to manage firewall rules", manager.version)
-            return manager
+            iptables_manager = IpTables(wire_server_address)
+
+            # Some distros (like RHEL 10.2) ship iptables but not the xt_owner and xt_conntrack kernel modules. As a
+            # result, the iptables commands to create the UID-based allow rule and conntrack-based drop rule will fail.
+            # In this case if nft is available, the agent should try to clean up the leftover iptables rules (since the
+            # DNS rule may exist) and fall back to nft for firewall management.
+            unresolved_modules = iptables_manager.get_unresolved_modules()
+            if len(unresolved_modules) > 0:
+                try:
+                    nftables_manager = NfTables(wire_server_address)
+                    event.warn(WALAEventOperation.Firewall, "Falling back to nftables because required iptables kernel modules are unresolved: {0}", unresolved_modules)
+
+                    try:
+                        iptables_manager.remove()
+                    except Exception as error:
+                        event.warn(WALAEventOperation.Firewall, "Unable to remove existing iptables rules while switching to nft: {0}", ustr(error))
+
+                    event.info(WALAEventOperation.Firewall, "Using nft [version {0}] to manage firewall rules", nftables_manager.version)
+                    return nftables_manager
+                except FirewallManagerNotAvailableError:
+                    event.warn(WALAEventOperation.Firewall, "Required iptables kernel modules are unresolved ({0}) and nft is not available, continuing with iptables as best effort", ", ".join(unresolved_modules))
+
+            event.info(WALAEventOperation.Firewall, "Using iptables [version {0}] to manage firewall rules", iptables_manager.version)
+            return iptables_manager
         except FirewallManagerNotAvailableError:
             pass
 
         try:
-            manager = NfTables(wire_server_address)
-            event.info(WALAEventOperation.Firewall, "Using nft [version {0}] to manage firewall rules", manager.version)
-            return manager
+            nftables_manager = NfTables(wire_server_address)
+            event.info(WALAEventOperation.Firewall, "Using nft [version {0}] to manage firewall rules", nftables_manager.version)
+            return nftables_manager
         except FirewallManagerNotAvailableError:
             pass
 
@@ -389,6 +414,31 @@ class IpTables(_FirewallManagerIndividualRules):
                     raise IptablesInconsistencyError(e.missing_rules, output_chain)
             raise
 
+    @staticmethod
+    def get_unresolved_modules():
+        """
+        Returns a list of the required iptables modules that modinfo cannot resolve. If modinfo is not available for
+        the check, returns an empty list because the modules cannot be confirmed as unresolved.
+
+        The list of modules this helper checks may not be exhaustive. It checks for the modules which are known to be
+        missing on the latest EL10 distros.
+        """
+        try:
+            shellutil.run_command(["modinfo", "--version"])
+        except Exception:
+            return []
+
+        unresolved_modules = []
+        for module_name in ["xt_owner", "xt_conntrack"]:
+            try:
+                shellutil.run_command(["modinfo", module_name])
+            except CommandError:
+                unresolved_modules.append(module_name)
+            except Exception:
+                pass
+
+        return unresolved_modules
+
     def load_conntrack(self):
         """
         Forces the conntrack module to be loaded by executing "iptables -C -m conntrack..."
@@ -450,7 +500,7 @@ class FirewallCmd(_FirewallManagerIndividualRules):
             self._version = shellutil.run_command(["firewall-cmd", "--version"]).strip()
         except Exception as exception:
             if isinstance(exception, OSError) and exception.errno == errno.ENOENT:  # pylint: disable=no-member
-                raise FirewallManagerNotAvailableError("nft is not available")
+                raise FirewallManagerNotAvailableError("firewall-cmd is not available")
             self._version = "unknown"
 
     @property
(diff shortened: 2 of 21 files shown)


azurelinuxagent/ga/firewall_manager.py
```diff
@@ -450,7 +500,7 @@ def __init__(self, wire_server_address):
             self._version = shellutil.run_command(["firewall-cmd", "--version"]).strip()
         except Exception as exception:
             if isinstance(exception, OSError) and exception.errno == errno.ENOENT:  # pylint: disable=no-member
-                raise FirewallManagerNotAvailableError("nft is not available")
+                raise FirewallManagerNotAvailableError("firewall-cmd is not available")
```
maddieford, line 503: This was an existing typo that I am fixing in this PR

tests_e2e/test_suites/agent_firewall.yml
```diff
@@ -12,4 +12,7 @@ tests:
 images:
   - "endorsed"
   - "endorsed-arm64"
+#  - "rocky_10"  # TODO: Rocky 10 does not ship the required kernel modules for iptables or nft by default. Once the distro fixes the dependency issue, this image needs to be tested
```
maddieford, line 15: rocky_10 isn't in our endorsed list. I added it commented out here so that we remember to test it once ciq fixes the dependency issues in the image

tests_e2e/test_suites/agent_firewall.yml
```diff
@@ -12,4 +12,7 @@ tests:
...
 owns_vm: true  # This vm cannot be shared with other tests because it modifies the firewall rules and agent status.
+skip_on_images:
+  - "alma_10"   # TODO: Alma 10 does not ship the required kernel modules for iptables or nft by default. Once the distro fixes the dependency issue, this image needs to be tested
```
maddieford, line 18: We should skip alma_10 to reduce noise in our daily runs until alma publishers fix the dependency issues

tests_e2e/test_suites/images.yml
```diff
@@ -293,11 +296,16 @@ images:
       locations:
          AzureChinaCloud: []
          AzureUSGovernment: []
-   rhel_10:
+   rhel_10_0:
```
maddieford, line 299: The RHEL 10 urn we were using before is specifically 10.0. I added the new urn which takes the latest minor version in 10 major family and added it to our endorsed list

tests_e2e/tests/agent_persist_firewall/agent_persist_firewall.py
```diff
@@ -69,9 +72,20 @@ def _test_setup(self):
...
     def _verify_persist_firewall_service_running(self):
         log.info("Verifying persist firewall service is running")
-        self._run_remote_test(self._ssh_client, "agent_persist_firewall-verify_persist_firewall_service_running.py", use_sudo=True)
+        command = "agent_persist_firewall-verify_persist_firewall_service_running.py"
+        if self._expect_firewalld_running:
+            command += " --expect-firewalld-running"
+        self._run_remote_test(self._ssh_client, command, use_sudo=True)
         log.info("Successfully verified persist firewall service is running\n")
 
+    def _is_firewalld_running(self):
```
maddieford, line 81: I added this check so we can ensure firewalld does not enter a failed state after removing or applying any of agent's passthrough rules

tests_e2e/tests/lib/agent_log.py
```diff
@@ -370,6 +370,32 @@ def get_errors(self) -> List[AgentLogRecord]:
...
+            #
+            {
+                'message': r"(Required iptables kernel modules are unresolved \(xt_owner, xt_conntrack\) and nft is not available, continuing with iptables as best effort)"
+                           "|"
+                           r"((Error initializing firewall|An error occurred while verifying the state of the firewall): .*iptables.*-m (owner|conntrack).*missing kernel module)"
+                           "|"
+                           r"(The firewall rules for Azure Fabric are not setup correctly \(the environment thread will fix it\): The following rules are missing: \['ACCEPT', 'DROP'\])"
+                           "|"
+                           r"(The firewall is not configured correctly. The following rules are missing: \['ACCEPT', 'DROP'\].*Will reset it.)",
+                'if': lambda r: DISTRO_NAME == "almalinux" and FlexibleVersion(DISTRO_VERSION).major == 10
```
maddieford, line 387: We may get noise in alma_10 runs for other test suites. Adding this ignore rule temporarily until the dependency issue for firewall is resolved

tests_e2e/tests/lib/agent_log.py
```diff
@@ -370,6 +370,32 @@ def get_errors(self) -> List[AgentLogRecord]:
...
+            },
+            #
+            # RHEL 10.2 does not include the kernel modules required by the iptables rules. These warnings are expected
+            # when the agent selects nftables for runtime and persistent firewall rules instead.
+            #
+            {
+                'message': r"(Falling back to nftables because required iptables kernel modules are unresolved:)"
+                           "|"
+                           r"(Firewalld service is running, but runtime firewall rules use nftables; trying to set up )",
+                'if': lambda r: r.level == "WARNING" and DISTRO_NAME in ["rhel", "redhat"] and FlexibleVersion(DISTRO_VERSION) == FlexibleVersion("10.2")
```
maddieford, line 397: The agent now adds a warning log whenever we detect the missing module state and fallback to nftables. We only want to ignore that warning on 10.2 right now, so that if this issue starts surfacing in other distros/versions we will have visibility.

tests_e2e/tests/scripts/agent_persist_firewall-access_wireserver
```diff
@@ -117,7 +117,7 @@ fi
 if [[ "$USER" != "root" ]]; then
   echo ""
   echo "checking tcp traffic to wireserver port 53 for non-root user ($USER)"
-  echo -n 2>/dev/null < /dev/tcp/$WIRE_IP/53 && echo 0 || echo 1  # Establish network connection for port 53
+  echo -n 2>/dev/null < /dev/tcp/$WIRE_IP/53  # Establish network connection for port 53
```
maddieford, line 120: This check was flawed before, since 'echo 0' and 'echo 1' are the last commands run and they always have exit code 0, even if the tcp check failed

azurelinuxagent/ga/firewall_manager.py
```diff
@@ -86,21 +86,46 @@ def __init__(self, wire_server_address):
...
     def create(wire_server_address):
         """
-        Creates the appropriate FirewallManager implementation depending on the availability of the underlying command-line tools.
+        Creates the appropriate FirewallManager implementation depending on the availability of the underlying
+        command-line tools and the kernel modules required by iptables.
 
-        NOTE: Currently this method checks only for iptables and nftables, giving precedence to the former.
+        NOTE: Currently this method checks only for iptables and nft, giving precedence to the former. There are
+        some images (Rhel 10.2/Alma 10/Rocky 10), where iptables exists but the required modules (xt_owner and
+        xt_conntrack) to create the agent rules don't exist. This method now handles that case and cleans up the
```
nagworld9, line 94: is it worth checking xt_tcp as well? We may have this case at some point

tests/ga/test_update.py
```diff
@@ -1222,6 +1223,30 @@ def test_it_should_setup_the_firewall(self):
                             mock_firewall_cmd.call_list,
                             "Expected 2 calls for the legacy rule (-C and -D), followed by 3 sets of calls for the current rules (-C and -A)")
 
+    def test_initialize_firewall_should_pass_True_to_persistence_when_nftables_is_selected(self):
```
nagworld9, line 1226: I don't get the test case. What does it means True and False here?

tests_e2e/tests/agent_persist_firewall/agent_persist_firewall.py
```diff
@@ -33,9 +34,11 @@ class AgentPersistFirewallTest(AgentVmTest):
     def __init__(self, context: AgentVmTestContext):
         super().__init__(context)
         self._ssh_client: SshClient = self._context.create_ssh_client()
+        self._expect_firewalld_running = False
 
     def run(self):
         FirewallUtilities.skip_test_if_proxy_agent_is_managing_the_wireserver_endpoint(self._ssh_client)
+        self._expect_firewalld_running = self._is_firewalld_running()
```
nagworld9, line 41: I remember you mentioning that firewalld result into failed state if modules not available but we apply the rules anyway. In that case will your check not fail?

tests_e2e/tests/scripts/agent_persist_firewall-verify_firewalld_rules_readded.py
```diff
@@ -20,19 +20,56 @@
...
+
+def main():
     if not Firewalld.is_service_running():
         log.info("firewalld.service is not running and skipping test")
         return
 
     firewall = Firewalld()
     firewall.log_firewall_state("** firewalld.service is running; initial state of the firewall")
 
+    if isinstance(FirewallManager.create(), NfTables):
```
nagworld9, line 35: we already check in the [-verify_persist_firewall_service_running.py‎] script that if its nft and firewalld running, verfiy firewalld rules not present. Why here?

tests/ga/test_update.py
```diff
@@ -1222,6 +1223,30 @@ def test_it_should_setup_the_firewall(self):
...
+        firewall_manager = MagicMock(spec=NfTables)
+        firewall_manager.check.return_value = True
+        persistence_handler = MagicMock()
```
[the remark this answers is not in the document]
maddieford: isinstance(firewall_manager, NfTables) returns True, which is asserted on line 1236

tests/test_agent.py
```diff
@@ -511,6 +513,44 @@ def test_it_should_parse_setup_firewall_properly(self):
         self.assertEqual(cmd, AgentCommands.Help)
         self.assertEqual(None, wire_server_address)
 
+    def test_setup_firewall_should_return_early_when_firewalld_is_enabled(self):
+        firewall_manager = Mock(spec=IpTables)
```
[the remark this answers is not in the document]
maddieford: isinstance(firewall_manager, Iptables) returns True and the test takes the expected branch

azurelinuxagent/ga/firewall_manager.py
maddieford: I was trying to keep this PR scoped to the two modules known to be missing on the affected images. I checked and xt_tcpudp module is built in on Rhel/Alma/Rocky 10. We could add it to the list, but there are probably other modules needed as well and I don't know if we want to maintain this dependency list in the agent.

tests/ga/test_update.py
maddieford: When we call PersistFirewallRulesHandler().setup() we should pass whether the firewall manager used to setup the runtime rules was NfTables or not. If it is NfTables (True), then we should not use firewalld

tests_e2e/tests/agent_persist_firewall/agent_persist_firewall.py
maddieford: Firewalld will only enter a failed state when the service is restarted or the VM is rebooted. We want to check the state before rebooting the VM on line 49 and then make sure it is still running after reboot if it was running before reboot

tests_e2e/tests/scripts/agent_persist_firewall-verify_firewalld_rules_readded.py
maddieford: verify_persist_firewall_service_running.py checks that after installation and each reboot, incompatible firewalld passthrough rules are not created when nftables is selected. It doesn't necessarily test cleanup of old passthrough rules, since sometimes the passthrough rules are created by an older agent and other times they are not.

This test verifies that the agent cleans up any passthrough rules created by a previous agent. It stops the agent, creates the 3 passthrough rules that an older agent may have created, restarts the agent, and verifies they were removed.

azurelinuxagent/ga/firewall_manager.py
nagworld9: I guess when miss xt_tcp, we likely see other two modules are also missing. But keeping pr scoped to current problem makes sense.

tests/ga/test_update.py
nagworld9: Then we are not testing firewalld used or not but we are checking persistfirewallruleshandler called or not, right?

tests_e2e/tests/agent_persist_firewall/agent_persist_firewall.py
nagworld9: if your test is passing means, our assumption of firewalld is running before boot and failing after reboot is wrong?

tests_e2e/tests/scripts/agent_persist_firewall-verify_firewalld_rules_readded.py
nagworld9: can you add it as separate test(method) instead in the verify_firewall_rules_readded to make it more intentional?

tests/ga/test_update.py
maddieford: That's right, this just checks that we call persistfirewallruleshandler with the correct value for 'runtime_uses_nftables'

The actual tests that we use firewalld or not are in test_persist_firewall_rules.py

tests_e2e/tests/agent_persist_firewall/agent_persist_firewall.py
maddieford: At this point (line 41), we have already installed the test agent which removes the bad passthrough rules, so firewalld shouldn't go into bad state after reboot

tests_e2e/tests/scripts/agent_persist_firewall-verify_firewalld_rules_readded.py
maddieford: Updated

azurelinuxagent/agent.py
```diff
@@ -306,19 +306,19 @@ def setup_firewall(endpoint):
...
-
         try:
             firewall_manager = FirewallManager.create(endpoint)
             firewall_manager.verbose = True
             if isinstance(firewall_manager, IpTables):
+                try:
+                    run_command(['systemctl', 'is-enabled', '--type=service', 'firewalld.service']).rstrip()
+                    logger.info("Firewalld is enabled. Will not setup the firewall rules to avoid conflicts.")
+                    sys.exit(0)
+                except CommandError:
```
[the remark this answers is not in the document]
maddieford: This is acceptable. The only functional difference is that this would exit with code 1 instead of code 0 when firewalld is enabled.

That doesn't have any impact on agent behavior, and the new exit code is more indicative that there is some unexpected behavior happening.

nagworld9 (approved):

